PR #197, opened 2015-01-02
Fix typo.

hanton:
"fadeOutPlaceholder" should be "placeholderFadesOut", which is declared in "ASDisplayNode.h".


rnystrom:
Good catch!

PR #203, opened 2015-01-07
Fix crash in Flow Controller

tanliboy:
The cur.first may out of boundary in the case for all empty sections


secretiverhyme:
Whoops, wrong button!  Merged =]

PR #206, opened 2015-01-08
Only download images when multiplex node is displayed

### LINKED ISSUE #178: ASMultiplexImageNode downloads aren't restricted to the working range
Setting the `imageIdentifiers` property will immediately enqueue downloads, even if the node isn't in the working range.  This should instead be gated by `displayWillStart`.

rnystrom:
fixes #178


secretiverhyme:
Build break!  Looks like we'll need to update some unit tests.

rnystrom:
On it! :+1:

secretiverhyme:
:D

Issue #200, opened 2015-01-06
All nodes are initialized and measured when ASRangeController is setup


rnystrom:
- Open the Kittens example
- Change the [litterSize](https://github.com/facebook/AsyncDisplayKit/blob/master/examples/Kittens/Sample/ViewController.m#L19) to something large (1000 works)
- Observe very long startup (iPhone 4 takes ~30 seconds)

On the Kittens demo, it appears that we are calling [-measure:](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASDisplayNode.mm#L357) on every single node. The `measure:` method also sets up and renders the placeholder. When there is text involved, this spins up an `ASTextNodeRenderer` per node (1000 in this above case).

The issue with `ASRangeControllers` is that it waits for *every single node* to init and measure in the [-insertNodes:](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASDataController.mm#L306-345) method before displaying anything. With a large data set this is problematic.

rnystrom:
@tanliboy would you mind taking a peak at this and seeing if its something we can optimize w/ the way we insert nodes? Either in batches or something. We can jam on some ideas, just grab me :smiley:

tanliboy:
Yeah, we should use the batch update here. I will fix it.

tanliboy:
Fixed in the pending pull request

rnystrom:
:clap: thanks @tanliboy

PR #211, opened 2015-01-09
Batch animation

tanliboy:
Not for 1.1 release.
Add batch update and animation support for UITableView.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -14,6 +14,8 @@
 #import "ASDataController.h"
 #import "ASDisplayNodeInternal.h"
 
+const static NSUInteger kASCollectionViewAnimationNone = 0;
+
 
 #pragma mark -
 #pragma mark Proxying.
@@ -137,7 +139,7 @@ static BOOL _isInterceptedSelector(SEL sel)
   ASDisplayNodePerformBlockOnMainThread(^{
     [super reloadData];
   });
-  [_dataController reloadData];
+  [_dataController reloadDataWithAnimationOption:kASCollectionViewAnimationNone];
 }
 
 - (void)setDataSource:(id<UICollectionViewDataSource>)dataSource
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -14,6 +14,8 @@
 #import "ASDataController.h"
 #import "ASDisplayNodeInternal.h"
 
+const static NSUInteger kASCollectionViewAnimationNone = 0;
```
secretiverhyme: should this be an enum?

secretiverhyme:
> Not for 1.1 release.

Let's hold off on merging this until ASDK 1.1.1 is released — we'll include it in 1.2.

wockerwalker:
When do you all expect that this will be released? It contains a fix for a problem that currently makes using ASTableView/ASCollectionView problematic with large data sources, like one I'm dealing with. Thanks for any info!

secretiverhyme:
> When do you all expect that this will be released?

@wockerwalker — current plan is to push the big red button and release AsyncDisplayKit 1.1.1 later this week.  We'll then merge this into master, along with some other cool stuff we have waiting in the wings, and unofficially open the 1.2 beta.

secretiverhyme:
Looks like 1.1.1 is stable — merging.

wockerwalker:
@secretiverhyme thanks for the info! I just installed 1.1.1 and I'm still seeing every ASCellNode init'ed for every item in the data source...any ideas?

secretiverhyme:
@wockerwalker, are these ASCellNode initialisations blocking content from appearing onscreen?  If so, that's a bug that should've been fixed already (#200).  Otherwise, that's currently expected behaviour — ASTableView and ASCollectionView asynchronously allocate and measure all cell nodes, using the working range and `-reclaimMemory` to make sure memory usage stays low.

wockerwalker:
no, they aren’t blocking the initial content from appearing onscreen promptly. it’s helpful to know that the async allocation and measurement of all nodes is expected behavior. that said, even though initial content propagates without delay, the CPU churns for a long while finishing all the other inits and measurements in the background. for a larger data source (~1000+), it’s unfortunate to be initializing and measuring a cell hundreds or thousands of rows away that the user may never see. this seems like one of the only large conceptual disadvantages of ASTableView over the stock UITableView, which inits and calculates those heights on the fly. (i also consistently get memory warnings when working with a data source this large on ASDK.) this leads me to a few questions:
	- is this behavior expected to change in a future update?
	- in the meantime, are there best practices for keeping ASCellNode -init methods lightweight, and subsequently performing those aspects of initialization not impacting measurement closer to display time?
	- are there any workarounds, such as feeding the ASTableView data in chunks, that you would recommend?

thanks for your help and patience!

secretiverhyme:
@wockerwalker — yes, we're planning changes to this behaviour in order to better handle large data sets (cc @rnystrom).  ASTableView and ASCollectionView currently blindly instantiate and measure *all* available nodes as soon as they're available, but the "a node is never onscreen if it hasn't been measured" invariant can be preserved with incremental loading.  We're thinking of adapting the working range to only instantiate nodes when they're likely to be needed / destroy them as necessary for memory warnings.

For a stopgap measure, your suggestion of feeding ASTableView data in chunks — tail-loading, à la infinite scroll — is your best bet.  It's possible to have your code execute closer to display time using, for example, `-displayWillStart`, but this may actually be detrimental depending on what you're trying to do.

Hope this helps!  =]

wockerwalker:
Very helpful, thanks!

secretiverhyme:
> We're thinking of adapting the working range to only instantiate nodes when they're likely to be needed / destroy them as necessary for memory warnings.

(We're tracking this in #258.)

PR #215, opened 2015-01-12
Actual place holding

### LINKED ISSUE #213: ASTextNode Flashing on Highlight
I'm seeing some weird flashing in an ASTextNode that I made. When I push down on the node to highlight it, the background color changes as expected, but the top left corner flashes a lighter color. Have you guys seen this at all or am I doing something wrong?

I can post the code if that would help, but I thought maybe you would know the issue without it. Also notable, this is a new issue since updating from 1.0.1 to 1.1.

rnystrom:
Only show a placeholder if there is a place that needs holding

fixes #213


secretiverhyme:
Solid.

Issue #217, opened 2015-01-13
Is it possible to create ASCollectionView cells on background thread?


Jercik:
It is possible to configure ASCollectionView in such way, that it will call data source's -collectionView:nodeForItemAtIndexPath: method on background thread? If I remember correctly, early prerelease versions of ASCollectionView called -collectionView:nodeForItemAtIndexPath: on background thread. 
In my application, initialization of ASCellNode on main thread causes stuttering while scrolling. I wrote my ASCellNode subclass assuming initialization will be done on background thread. Is there any way to create ASCollectionView cells on background thread or should I refactor my code to make initialization faster?

tanliboy:
Great question, we do have a lot discussions on this issue before.

Short answer, YES. 

Let us start with your first question:

>If I remember correctly, early prerelease versions of ASCollectionView called -collectionView:nodeForItemAtIndexPath: on background thread. 

Calling "collectionView:nodeForItemAtIndexPath" in async mode is problematic, because the dataSource may be changed after the async delay. For example, if I insert 2 elements, and then delete the 2 elements, in async mode, we could not guarantee the "collectionView:nodeForItemAtIndexPath" will be called before the "deletion". So we have to keep "collectionView:nodeForItemAtIndexPath" to be called while you call "insert/reloading" on CollectionView.

So there are two options for you to make it in background thread:
1) call insert/delete/moving in background thread directly. Yes, it is thread safe. :-)  The "reloadData" is not thread safe (I have abandoned pull request to add the thread-safe method, but abandoned to avoid make the api complicated), but you don't need to worry about it if you only want to "insert/deletion/moving" or "reloadRowsAtIndexPaths".
2) Create and initialize the node in background thread, and "dispatch_sync" to call the "insert/deletion" on   main thread or any other thread you want. 

Help it will be helpful ! Let us know if you still have problem.

Jercik:
Thank you for your suggestion! Option 1 works great for me. I've updated comments for insert/delete/moving methods in ASCollectionView and ASTableView.

tanliboy:
Glad to know it works for you. 
These operations are actually "asynchronous". We fetch the data "synchronous", but update elements in Collection View in async mode.

PR #218, opened 2015-01-14
Updated comments.

Jercik:
Issue #217


tanliboy:
The operations are actually "asynchronous", although we fetched the data "synchronous".
We fetch the data "synchronously", layout in the background thread, and then dispatched to main thread to insert/delete/move the elements in CollectionView.

Jercik:
I see. I've read updated comments again and they don't make much sense now :P. Sorry for trouble.

PR #221, opened 2015-01-14
Removed unintentional conversion.

### LINKED ISSUE #219: ASTextNode misaligned when screen scale is 3.0
I've found bug in ASTextNode when running on iPhone 6 Plus simulator. For some strings calculatedSize is slightly too large, which causes view to stretch. This results in blurry text and (probably) performance decrease.

Here is example project which demonstrates issue: https://github.com/Jercik/ASTextStretch. After enabling "Color Misaligned Images" you can see that first text node is stretched: 

iphone6plus
calculatedSize returned by first node is: {286.33334350585938, 45.666667938232422}. 
Second node's size was set manually to correct value which is: {286.33333333333331, 45.666666666666664}.

Jercik:
Fixes #219.


rnystrom:
Awesome, thank you!

secretiverhyme:
Nice catch @Jercik — thanks!

PR #234, opened 2015-01-19
Tidy ASCollectionView sample project in Swift.

DaniArnaout:
This is a swift version of 'ASCollectionView' sample project that was contributed by @secretiverhyme back in November 18th 2014.


secretiverhyme:
Thanks for the pull request!  It doesn't look like this example project materially differs from the existing ASTableView Swift sample project — is there something it showcases that the other doesn't?

DaniArnaout:
You're right. It's just that I was aiming to have all examples available in both Swift & Objective-C, so I started with this one and was waiting for your response to see if it's worth it to do the rest.
I believe that most new developers are only learning Swift, so I thought it would be nice to give them some samples in Swift. What do you think?

secretiverhyme:
Ooh, gotcha.  In principle I like the idea, but since we're adding and updating sample projects relatively frequently as we expand the framework, the code duplication would quickly become unwieldy.

If you'd like to upgrade the existing Swift sample project to a more comprehensive demo, I think that would be pretty useful for Swift-only developers.  The documentation (asyncdisplaykit.org/guide) is updated quite infrequently, so adding a toggle between Objective-C and Swift code snippets could also be worthwhile.

DaniArnaout:
I think that the best way to contribute right now is by adding an Objective-C/Swift toggle to the documentation. Is there a special Markdown syntax to add the switch?

secretiverhyme:
There isn't currently — the site is generated with Jekyll using a very simple [documentation template](https://github.com/facebook/AsyncDisplayKit/blob/master/docs/_layouts/docs.html).

PR #109, opened 2014-11-15
Add [ASDisplayNode initWithLoadBlock:]

aleclarson:
For customized initialization (on a per-instance basis) of backing view/layer. See #108.

PR #235, opened 2015-01-20
ASBridgeNode: wrapper around an existing UIView with ASDisplayNode semantics

ide:
ASBridgeNode is a node that that wraps a UIView that you want to embed in a view hierarchy created by ASDK. Sometimes a view can't be constructed with `-[initWithViewClass:]` but you want to use it with ASDK.

The API is meant to preserve ASDisplayNode's behavior, so you can still construct and set properties on the node on a background queue before its view is loaded; even though the view was created a priori, it is not considered to be loaded until `node.view` is accessed.

Using ASBridgeNode looks like this:

    // Main queue:
    UIButton *button = [UIButton buttonWithType:UIButtonTypeSystem];
    [button sizeToFit];
    CGRect buttonFrame = button.frame;

    dispatch_async(backgroundQueue, ^{
        ASDisplayNode *node = [ASBridgeNode alloc] initWithView:button];
        // Use `node` as you normally would...
        node.frame = buttonFrame;
    });

ASBridgeNode takes care of managing the view hierarchy so you don't need to worry about it. The main thing it doesn't do (can't do?) is layout. Methods like `-[ASDisplayNode calculateSizeThatFits:]` and `-[ASDisplayNode layout]` cannot delegate to `[UIView sizeThatFits:]` and `[UIView layoutSubviews]` since the UIView methods must run on the main thread (well not always, but some views need it). If ASDK were internally asynchronous and could dispatch its layout calls to different threads (sort of like how ASTableView concurrently computes its cells' layouts) then we could mark ASBridgeNode as having "main-queue affinity" and delegate its layout to UIKit.

Test cases are included and all existing tests pass.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -12,6 +12,8 @@
 #import "ASBaseDefines.h"
 #import "ASDealloc2MainObject.h"
 
+typedef UIView *(^ASDisplayNodeViewBlock)();
+typedef CALayer *(^ASDisplayNodeLayerBlock)();
 
 /**
  * An `ASDisplayNode` is an abstraction over `UIView` and `CALayer` that allows you to perform calculations about a view
@@ -68,6 +70,22 @@
  * The layer instance will be created with alloc/init.
  */
 - (id)initWithLayerClass:(Class)layerClass;
+
+/**
+ * @abstract Alternative initializer with a block to create the backing view.
+ *
+ * @return An ASDisplayNode instance that loads its view with the given block that is guaranteed to run on the main
+ * queue. The view will render synchronously and -layout and touch handling methods on the node will not be called.
+ */
+- (id)initWithViewBlock:(ASDisplayNodeViewBlock)viewBlock;
+
+/**
+ * @abstract Alternative initializer with a block to create the backing layer.
+ *
+ * @return An ASDisplayNode instance that loads its layer with the given block that is guaranteed to run on the main
+ * queue. The layer will render synchronously and -layout and touch handling methods on the node will not be called.
+ */
+- (id)initWithLayerBlock:(ASDisplayNodeLayerBlock)viewBlock;
 
 
 /** @name Properties */
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -181,6 +181,35 @@ void ASDisplayNodePerformBlockOnMainThread(void (^block)())
   return self;
 }
 
+- (id)initWithViewBlock:(ASDisplayNodeViewBlock)viewBlock
+{
+  if (!(self = [super init]))
+    return nil;
+
+  ASDisplayNodeAssertNotNil(viewBlock, @"should initialize with a valid block that returns a UIView");
+
+  [self _initializeInstance];
+  _viewBlock = viewBlock;
+  _flags.synchronous = YES;
+
+  return self;
+}
+
+- (id)initWithLayerBlock:(ASDisplayNodeLayerBlock)layerBlock
+{
+  if (!(self = [super init]))
+    return nil;
+
+  ASDisplayNodeAssertNotNil(layerBlock, @"should initialize with a valid block that returns a CALayer");
+
+  [self _initializeInstance];
+  _layerBlock = layerBlock;
+  _flags.synchronous = YES;
+  _flags.layerBacked = YES;
+
+  return self;
+}
+
 - (void)dealloc
 {
   ASDisplayNodeAssertMainThread();
@@ -249,6 +278,48 @@ void ASDisplayNodePerformBlockOnMainThread(void (^block)())
 
 }
 
+- (UIView *)_viewToLoad
+{
+  UIView *view;
+  ASDN::MutexLocker l(_propertyLock);
+
+  if (_viewBlock) {
+    view = _viewBlock();
+    ASDisplayNodeAssertNotNil(view, @"View block returned nil");
+    ASDisplayNodeAssert(![view isKindOfClass:[_ASDisplayView class]], @"View block should return a synchronously displayed view");
+    _viewBlock = nil;
+    _viewClass = [view class];
+  } else {
+    if (!_viewClass) {
+      _viewClass = [self.class viewClass];
+    }
+    view = [[_viewClass alloc] init];
+  }
+
+  return view;
+}
+
+- (CALayer *)_layerToLoad
+{
+  CALayer *layer;
+  ASDN::MutexLocker l(_propertyLock);
+
+  if (_layerBlock) {
+    layer = _layerBlock();
+    ASDisplayNodeAssertNotNil(layer, @"Layer block returned nil");
+    ASDisplayNodeAssert(![layer isKindOfClass:[_ASDisplayLayer class]], @"Layer block should return a synchronously displayed layer");
+    _layerBlock = nil;
+    _layerClass = [layer class];
+  } else {
+    if (!_layerClass) {
+      _layerClass = [self.class layerClass];
+    }
+    layer = [[_layerClass alloc] init];
+  }
+
+  return layer;
+}
+
 - (void)_loadViewOrLayerIsLayerBacked:(BOOL)isLayerBacked
 {
   ASDN::MutexLocker l(_propertyLock);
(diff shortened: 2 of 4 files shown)


ide:
Thinking about this a bit more I think this API is convenient and could see a lot of use but there's at least one case it doesn't accomodate so well: if you're implementing a custom ASDisplayNode's initializer you can't safely construct UIViews to use with a ASBridgeNode.

```objc
@implementation XXCustomDisplayNode
- (instancetype)init
{
    // Might be on a background queue; not safe...
    UIButton *button = [UIButton buttonWithType:UIButtonTypeSystem];
    [self addSubnode:[[ASBridgeNode alloc] initWithView:button];
    // Needs to be something like: 
    [self addSubnode:[[ASBridgeNode alloc] initWithViewBlock:^UIView *{
        return [UIButton buttonWithType:UIButtonTypeSystem];
    }];
    // More complicated when you don't need it but addresses this case
}
@end
```

secretiverhyme:
Thanks for the pull request!

ASDisplayNode is intended to support synchronous wrapper nodes as first-class nodes — they should be indistinguishable from any other node with asynchronous display disabled.  So, instead of creating an additional wrapper-only node, we should fix these problems in the base node class.

It looks like this is more or less the same line of thought as #109 (which I haven't gotten to yet — sorry @aleclarson!).  @ide, if you'd like to hoist this functionality into the base node and add CALayer support, I'd be happy to merge it in and close both PRs.  =]

secretiverhyme:
> but there's at least one case it doesn't accomodate so well: if you're implementing a custom ASDisplayNode's initializer you can't safely construct UIViews to use with a ASBridgeNode

This is what the `-didLoad` subclass hook is useful for — it's called on the main thread (guaranteed) once the node's backing view has been created, and is where you're supposed to add subviews, gesture recognisers, &c.  It might be worth adding a subclass hook (`-load`, perhaps?) instead of an additional "view block" initialiser for API consistency.

ide:
Thanks for taking a look. I'll see what the implementation ends up looking like with CALayer support. Hopefully it should be simple.

Regarding the API, I find view/layer block approach simpler to think about because the entire node hierarchy can be set up inside of init methods without splitting the code between `init` and `didLoad`.

ide:
Moved the code into ASDisplayNode and squashed the commits. It ended up being pretty reasonable except that we don't know if the view/layer supports synchronous rendering until it has been created. I took a conservative approach of making `isSynchronous` fail if we don't yet know (this is covered in the unit tests).

Diff description:

This adds new initializer methods to ASDisplayNode:
```objc
initWithView:(UIView) *view
initWithLayer:(CALayer *)layer
initWithViewBlock:(ASDisplayNodeViewBlock)viewBlock
initWithLayerBlock:(ASDisplayNodeLayerBlock)layerBlock
```

Sometimes a view can't be constructed with `-[initWithViewClass:]` but you want to use it with ASDK, so these new methods provide a way to wrap an existing view in a node.

The API is meant to preserve ASDisplayNode's behavior, so you can still construct and set properties on the node on a background queue before its view is loaded; even though the view was created a priori, it is not considered to be loaded until `node.view` is accessed.

Using the API looks like this:

    dispatch_async(backgroundQueue, ^{
        ASDisplayNode *node = [ASDisplayNode alloc] initWithViewBlock:^UIView *{
            // Guaranteed to run on the main queue
            UIButton *button = [UIButton buttonWithType:UIButtonTypeSystem];
            [button sizeToFit];
            CGRect buttonFrame = button.frame;
        }];
        // Use `node` as you normally would...
        node.backgroundColor = [UIColor redColor];
    });

The main thing this bridging API doesn't do (can't do?) is layout. Methods like `-[ASDisplayNode calculateSizeThatFits:]` and `-[ASDisplayNode layout]` cannot delegate to `[UIView sizeThatFits:]` and `[UIView layoutSubviews]` since the UIView methods must run on the main thread. If ASDK were internally asynchronous and could dispatch its layout methods to different threads (sort of like how ASTableView computes its cells' layouts) then we could mark nodes with externally provided views/layers as having "main-queue affinity" and delegate its layout to UIKit.

Test cases are included and all existing tests pass.

secretiverhyme:
Looks good — thanks for the follow-up!  A couple of questions:

> except that we don't know if the view/layer supports synchronous rendering until it has been created

I'm a little concerned about adding an assert to the `-isSynchronous` method.  The async display machinery *should* always be checking `_flags` directly for this property, but IIRC it also checks some `_flags`-wrapping methods at times, so it's easy to imagine this tripping up someone later on.

Instead, what about treating all block- and wrapper-constructed nodes as synchronous?  There are very few situations where you'd want a custom view or layer object that happens to subclass `_ASDisplayView` or `_ASDisplayLayer`, and most people using the framework (should) have no idea what either of those classes are, so requiring use of `-initWith{View,Layer}Class:` is fine.

> you can still construct and set properties on the node on a background queue before its view is loaded; even though the view was created a priori, it is not considered to be loaded until node.view is accessed

Hm, interesting.  The duplicate pending state object will cause problems here:

* accessing properties on the node *before* the `.view` property is accessed → default values
* accessing properties on the node *after* the `.view` property is accessed → actual values
* setting properties on the view then accessing the node's `.view` property → clobbered!

In the long run, we'd like to replace the current "becomes main-thread-affined once loaded" behaviour with a "set properties on any thread at any time and we'll make sure it works" system.  Until that's built, though, we can't immediately consider the node loaded without ending any-thread property support…

When is using an existing view rather than creating a new one desirable?  I think it would be safest to avoid supporting `-initWith{View,Layer}:` directly in ASDK until these problems are resolved — API consumers can still easily do

```objective-c
[[ASDisplayNode alloc] initWithViewBlock:^UIView *{
  return self.alreadyExtantView;
}];
```

or

```objective-c
- (void)didLoad {
  [self.view addSubview:alreadyExtantView];
}
```

if they really need to.

ide:
I agree with your comment about isSynchronous and treating all externally provided views/layers synchronously. Do you think it would be worth asserting that views/layers returned from the block actually don't subclass _ASDisplayView/Layer? There are a couple of runtime type checks for the _ASDisplay classes and there might be assumptions that a node renders synchronously iff the backing view/layer is of type _ASDisplayView/Layer.

> When is using an existing view rather than creating a new one desirable?

One case is when you want an ivar that references the view so you can configure it later. The code might look like:

```objc
// in a view controller
- (instancetype)init {
    _webView = [[WKWebView alloc] initWithFrame:... configuration:...];
    _webNode = [[ASDisplayNode alloc] initWithView:_webView];
    ...
    [someOtherNode addSubnode:_webNode];
}

- (void)handleSomeEvent {
  // could happen before the node is rendered
  [_webView loadRequest:...];
}
```

I think it's often not good to use the backing view directly (because of the duplicate-pending-state problem you brought up) but there are a bunch of cases (like a web view's URL) where it makes sense. In any case I'm on board with removing the `initWithView/Layer:` convenience initializers.

secretiverhyme:
> Do you think it would be worth asserting that views/layers returned from the block actually don't subclass _ASDisplayView/Layer?

Yes, definitely.

> One case is when you want an ivar that references the view so you can configure it later.

Hmm.  The standard pattern here is to create an ASDisplayNode subclass and cast its `view` property:

```objective-c
@interface WebNode : ASDisplayNode
@property (nonatomic, readonly, retain) WKWebView *view;
@end

@implementation WebNode
- (instancetype)init {

}

**Comment by secretiverhyme** [comment]
> Do you think it would be worth asserting that views/layers returned from the block actually don't subclass _ASDisplayView/Layer?

Yes, definitely.

> One case is when you want an ivar that references the view so you can configure it later.

Hmm.  The standard pattern here is to create an ASDisplayNode subclass and cast its `view` property:

```objective-c
@interface WebNode : ASDisplayNode
@property (nonatomic, readonly, retain) WKWebView *view;
@end

@implementation WebNode
- (instancetype)init {
    return [super initWithViewClass:WKWebView.class];
}
```

Then, in this example, you'd do

```objective-c
- (void)handleSomeEvent {
    [_webNode.view loadRequest:...];
}
```

The synchronous wrapper node system definitely has room for improvement, but that's outside the scope of this pull request.  Let's go ahead and drop the `-initWith{View,Layer}:` initialisers for now.  Thanks!

ide:
Latest update includes assertions that the returned views/layers are not subclasses of _ASDisplayXX and there's a test case for that. The `-[initWithView/Layer:]` convenience initializers are gone too.

secretiverhyme:
Excellent.  Thanks for pushing this through!

ide:
Thanks for the code review!

Issue #233, opened 2015-01-18
Counterintuitive requirement for ASTableView to use calculated node heights


nolanw:
Hello! I found it counterintuitive that `ASTableView` requires its `asyncDelegate` to be set in order for each cell to take on its node's height. If you never set `asyncDelegate`, the rows just take on their default `UITableView` heights. I guess this requirement could be documented somewhere, but ideally the table view would Just Work even if I never muck with its delegate.

The issue stems from the table view only setting `super.delegate` in its `asyncDelegate` setter, and even then only in the non-`nil` branch. One fix might be to initialize `super.delegate` to be `self`, and to set `super.delegate = self` whenever `asyncDelegate` is `nil`.

aleclarson:
:+1:

secretiverhyme:
Tracking in #231.  Thanks!  =]

nolanw:
Oh, my apologies! I'll take a closer look for duplicates next time.

secretiverhyme:
No worries!  Fix landed ^_^

PR #237, opened 2015-01-20
Default proxy object for ASTableView and ASCollectionView

### LINKED ISSUE #231: ASCollectionView and ASTableView ignore working range if asyncDelegate is not set
ASCollectionView recalculates the visible nodes and working range when `willDisplayCell:`/`didEndDisplayingCell:` are called on the UICollectionViewDelegate, but this delegate is set to the proxy only when asyncDelegate is set. It looks like ASTableView has the same problem too.

rnystrom:
fixes #231 #233


secretiverhyme:
Looks good.  Thanks much!

PR #240, opened 2015-01-22
Fix ASRangeController's delegate method invocations

ide:
"did" should be "will" in a couple of places


secretiverhyme:
Nice catch.  Thanks!

PR #245, opened 2015-01-23
Make ASDataController call rowsInSection with the appropriate index

andyscott:
`rowsInSection` needs to be called with the correct section index. Otherwise crashes occur in multi sectioned collection views (and I expect table views too).


secretiverhyme:
Nice catch.  Thanks!

PR #249, opened 2015-01-24
Fix strong ASNetworkImageNodeDelegate reference.

### LINKED ISSUE #239: ASNetworkImageNode _delegate is not weak reference.
Once I set ASNetworkImageNode's delegate, unless I unset ASNetworkImageNode's delete to nil, it will not be deallocated.

secretiverhyme:
Fixes #239.  (Thanks @rnystrom!)


rnystrom:
Awesome, thank you!

PR #250, opened 2015-01-24
Simple UIScrollView wrapper node.

### LINKED ISSUE #99: Provide a UIScrollView wrapper node.
Hey I know we can do an ASDisplayNode, and have it of class: UIScroll view, but I've got scroll views within ASCellNodes and really want to have those cells laying out asynchronously (which gets screwed up if I use UI anything. Is there are possibility of ASScrollNode?

secretiverhyme:
Closes #99.  This is a quick sketch -- room for improvement includes
considering a less-questionable class name, potentially exposing
additional UIScrollView functionality directly on the node, and using it
in a sample project.


rnystrom:
:laughing: that prefix

Should we be using the new initialization from #235?

@toulouse the gesture recognizer should be exposed through the `view` property here, which is a `UIScrollView`, right?

cfcommando:
Yep

appleguy:
@toulouse — there's a simple answer to this.  Remember FBStoryNode?  It's just one example of many where a node hierarchy needs to contain a scroll view, and unless the UIScrollView is wrapped in a node, asynchronous layout will be interrupted for children of the scrollable region.

We addressed this in Paper by doing an initWithViewClass:.  We now have initWithViewBlock:, which is an improvement, but still—scrollable areas are such a fundamental part of iOS UI, that a scroll node fits right alongside ASControlNode, ASImageNode, ASTextNode, ASTableView, ASCollectionView...

Beyond the unambiguous utility (although it may not save a lot of work for developers, it is definitely going to be used and also does reduce client code while cleaning up style), it will also serve as an example for developers about how to wrap synchronous views.  It'll be referenced in my upcoming presentation as well as documentation as such an example, in fact.

That said, I appreciate you weighing in here and asking these legitimate questions as we are making API changes.  It is really great having more eyes on this.  In the future we might open FBScrollView, but we don't have time to clean it up or expose more functionality before implementing other key classes like ASPageNode.

toulouse:
Ah, cool. I actually didn't delve deeply into FBStoryNode very much at all, and ramped up on ASDK somewhat late, so my recollections of its implementation are not very detailed. Thanks for the context. I wasn't against the wrapper so much as I was looking for the context I was missing.

PR #260, opened 2015-01-26
Added node access methods to ASTableView, ASCollectionView.

dshahidehpour:


rnystrom:
This could be super useful. Thanks for the PR! We'll take a second to review.

nolanw:
I was just about to swing by and ask for this. Thanks y'all!

PR #264, opened 2015-01-27
Compile for extensions

### LINKED ISSUE #98: Support using AsyncDisplayKit in extensions.
`+[UIApplication sharedApplication]` is verboten in app extensions.  It's currently only in use in ASControlNode — we should remove it, and use Travis to ensure ASDK stays extension-safe.

rnystrom:
Set a build setting so that ASDK compiles "extension-safe".

We've removed calls to `-[UIApplication sharedApplication]` which is banned in extensions. Instead, we fall back on the plain ol' responder chain to handle `ASControlNode` events.

This PR also includes some basic tests to account for event propagation for control nodes, as well as minor semantics with what the actual `-[ASDisplayNode nextResponder]` (a private method) is.

fixes #98

### DIFF
--- a/AsyncDisplayKit/ASControlNode.m
+++ b/AsyncDisplayKit/ASControlNode.m
@@ -347,13 +347,18 @@ void _ASEnumerateControlEventsIncludedInMaskWithBlock(ASControlNodeEvent mask, v
         for (NSString *actionMessage in targetActions)
         {
           SEL action = NSSelectorFromString(actionMessage);
+          id responder = target;
 
-          // Hand off to UIApplication to send the action message.
-          // This also handles sending to the first responder is target is nil.
-          if (target == [NSNull null])
-            [[UIApplication sharedApplication] sendAction:action to:nil from:self forEvent:event];
-          else
-            [[UIApplication sharedApplication] sendAction:action to:target from:self forEvent:event];
+          // NSNull means that a nil target was set, so start at self and travel the responder chain
+          if (responder == [NSNull null]) {
+            // if the target cannot perform the action, travel the responder chain to try to find something that does
+            responder = [self.view targetForAction:action withSender:self];
+          }
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Warc-performSelector-leaks"
+          [responder performSelector:action withObject:self withObject:event];
+#pragma clang diagnostic pop
         }
       }
     });
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -241,13 +241,6 @@ void ASDisplayNodePerformBlockOnMainThread(void (^block)())
   _pendingDisplayNodes = nil;
 }
 
-#pragma mark - UIResponder overrides
-
-- (UIResponder *)nextResponder
-{
-  return self.view.superview;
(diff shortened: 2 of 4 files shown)


AsyncDisplayKit/ASControlNode.m
```diff
@@ -347,13 +347,18 @@ - (void)sendActionsForControlEvents:(ASControlNodeEvent)controlEvents withEvent:
...
-          // Hand off to UIApplication to send the action message.
-          // This also handles sending to the first responder is target is nil.
-          if (target == [NSNull null])
-            [[UIApplication sharedApplication] sendAction:action to:nil from:self forEvent:event];
-          else
-            [[UIApplication sharedApplication] sendAction:action to:target from:self forEvent:event];
+          // NSNull means that a nil target was set, so start at self and travel the responder chain
+          if (responder == [NSNull null]) {
+            // if the target cannot perform the action, travel the responder chain to try to find something that does
+            responder = [self.view targetForAction:action withSender:self];
```
rnystrom: No hacks!

secretiverhyme: Fantastic.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -245,7 +245,7 @@ - (void)dealloc
 
 - (UIResponder *)nextResponder
 {
-  return self.view.superview;
+  return [self.view nextResponder];
```
secretiverhyme: Woah.  This is almost as bad as the "not handling a touch? let's forward it up the view hierarchy instead of the class inheritance heirarchy!" bug.  Nice find.

Have you confirmed that we're not depending on the old (wrong) behaviour anywhere?

AsyncDisplayKitTests/ASControlNodeTests.m
```diff
@@ -0,0 +1,118 @@
+/* Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <CoreText/CoreText.h>
```
secretiverhyme: Wait, what?

secretiverhyme:
Looks good.  Couple quick questions, then we're good to merge this in!

AsyncDisplayKit/ASDisplayNode.mm
rnystrom: Ran through unit tests and Kittens. Let me spin up the other sample projects to be extra sure.

AsyncDisplayKitTests/ASControlNodeTests.m
rnystrom: :stuck_out_tongue: copy/pasta from `ASTextNodeTests`

AsyncDisplayKit/ASDisplayNode.mm
rnystrom: Turns out its not used *anywhere*. So I snipped it entirely :laughing:

secretiverhyme:
Awesome.  I wonder where that `-nextResponder` came from…

Issue #243, opened 2015-01-23
Question: ASTextNode single line of text with truncationAttributedString


andyscott:
This is a question, not an issue.

What's the proper way to size and then layout an ASTextNode so that it shows exactly one line of text and truncates with `truncationAttributedString` at a fixed length? I need the calculated height of the text node in order to compute the rest of my layout.

At the moment I calculate the text size with an infinite width so that the text always renders in a single line. When I lay the node out, I use my fixed width instead of the calculated width.

Is there any other way to force the text to render as a single line? Should I calculate the size using a `CGSizeMake(fixedLength, guessedHeightOfOneLine)`? #152 talks of a label node-- one line text seems like the label's job.

secretiverhyme:
Quoting @ocrickard in #152:

> It's trivial to add maximumNumberOfLines property on ASTextNode as-is. The ASTextNodeRenderer would just have to set it on the NSTextContainer. We're open to Pull-requests!

Infinite-width hacks will work, but this is the most proper solution that I'm aware of.  I don't see any reason not to expose `maximumNumberOfLines` on ASTextNode as well as any future ASLabelNode, if you feel like taking a stab at it!  =]

andyscott:
I'll take a stab early next week.

Issue #152, opened 2014-12-11
[ASLabelNode] Create a simpler alternative to ASTextNode


secretiverhyme:
ASTextNode's `attributedString` API is overkill for a simple, uniformly-styled label.  Consider—

``` objective-c
ASLabelNode *label = [[ASLabelNode alloc] init];
label.textColor = [UIColor grayColor];
label.font = [UIFont systemFontOfSize:12.0f];
label.text = @"hello";
```

—versus—

``` objective-c
ASTextNode *label = [[ASTextNode alloc] init];
NSDictionary *attributes = @{ NSForegroundColorAttributeName: [UIColor grayColor],
                                         NSFontAttributeName: [UIFont systemFontOfSize:12.0f] };
NSAttributedString *attributedString = [[NSAttributedString alloc] initWithString:@"hello"
                                                                       attributes:attributes];
```

ocrickard:
I'd be OK with this so long as we don't add it to ASTextNode and do not subclass.  In order to add it to ASTextNode we'd have to add a new prerender stage to wrap the attributed string with the default attributes, something I'm **very** against for performance reasons.

The actual implementation of ASLabelNode could just wrap an ASTextNode or perhaps a further extracted ASTextNodeRenderer/friends.  It could lazily generate an attributed string and pass it into the same backend as ASTextNode.

For the record I believe subclassing of concrete leaf nodes like ASTextNode should be seen as an antipattern.  Composition makes refactors I'm working on much easier.

secretiverhyme:
+1 @ocrickard.  An ASLabelNode that's a stripped-down counterpart to ASTextNode would be a useful addition to the framework, but using ASTextNodeRenderer directly rather than subclassing or expanding ASTextNode is the way to go.

ide:
UILabel's adjustsFontSizeToFitWidth and numberOfLines properties are also sometimes really helpful and seem like it'd be good to offload them to a background thread. I'd actually like ASTextNode to support these (and ASLabelNode, too, by delegating to a text node) since the inconsistent APIs across UILabel/UITextField/UITextView are a pain point of UIKit.

ocrickard:
It's trivial to add maximumNumberOfLines property on ASTextNode as-is.  The ASTextNodeRenderer would just have to set it on the NSTextContainer.  We're open to Pull-requests!

ocrickard:
[...]

PR #265, opened 2015-01-27
Add maximumLineCount to ASTextNode

andyscott:
Per #243 and #152, here's the implementation.

I'd followed the naming conventions in `ASTextNode` rather than naming the property `maximumNumberOfLines` to match the property on `NSTextContainer`.


andyscott:
Forgot to update the tests cases.

secretiverhyme:
Looks good — thanks for including a test case.

ocrickard:
Looks great, thanks!

PR #273, opened 2015-01-29
Add -tableView:editActionsForRowAtIndexPath: to ASCommonTableViewDelegate

nolanw:
Might as well make ASTableView's support for this method explicit.

It was introduced in iOS 8 but I don't think there's a need to gate it with an `#ifdef`. It'll simply never get called on iOS 7.


secretiverhyme:
Sure, sounds reasonable.  Thanks for the pull request!

Issue #134, opened 2014-12-04
[ASDisplayNode] Revisit the thread-affinity API.


secretiverhyme:

toulouse:
also subscribed

appleguy:
@secretiverhyme @rnystrom @toulouse - The community is planning an exciting long term road map for the project and getting organized around how to deliver these feature requests. We are moving to a new issue tracker with more sophisticated planning capabilities (check out www.realArtists.com and their new product: Ship!). 

If you are interested in helping contribute to this component or any other, don’t hesitate to send me an email at AsyncDisplayKit@gmail.com. If you would like to contribute for a few weeks, I can also add you to our Ship bug tracker so that you can see what everyone is working on and actively coordinate with us. 

As always, keep filing issues and submitting pull requests here on Github and we will only move things to the new tracker if they require long term coordination.

PR #274, opened 2015-01-29
Introduce Temporary Internal API to Allow Workaround for Threading Affin...

bdaz:
...ity Violation

See D1809545 in fbobjc.


secretiverhyme:
cc #134

PR #275, opened 2015-01-30
Remove "working range" from flow layout

rnystrom:
This is a step towards a preloading range. We first want to migrate towards arbitrary ranges instead of being coupled to the notion of a single "working range".

This PR is one of several iterative, sequential PRs so that they are broken into smaller, consumable changes.


rnystrom:
Waiting until all work is done and send one PR w/ sequenced commits

PR #276, opened 2015-01-31
Added highlighted image for ASImageNode

Dreddik:


secretiverhyme:
Thanks for the pull request!  Let's please avoid changing ASControlNode -- the `highlighted` property is intended to reflect the UI control's current state and should not be externally modifiable.  (Overriding `-setHighlighted:` in subclasses [will still do the right thing](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASControlNode.m#L115).)

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -155,7 +155,7 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer;
 
   return [[_ASImageNodeDrawParameters alloc] initWithCrop:_cropEnabled
                                                    opaque:self.opaque
-                                                    image:self.image
+                                                    image:!self.highlighted ? self.image : self.highlightedImage
```
secretiverhyme: nit:  this would be clearer as

```objective-c
self.highlighted ? self.highlightedImage : self.image
```

Dreddik:
Well, I guess now should be okay

secretiverhyme:
Not quite, Travis CI reports build failures :)

The `highlighted` property is intended to reflect UI control state — whether the node is currently being highlighted by the user's finger — and is therefore readonly.  That's why the `[super setHighlighted:]` call doesn't compile, and why we shouldn't expose highlighted-ness even on ASImageNode.

It sounds like you'd like to add a programmatically-configurable *selected* state to image nodes.  That's not something we'd like to add to ASImageNode right now (it would make more sense on a generic ASButtonNode) but you could easily add it as a new property in a custom ASImageNode subclass.

PR #279, opened 2015-02-03
Fix ASControlNode Event Dispatch Table Memory Semantics

bdaz:
Summary: `weakToStrong` where the keys are NSNumbers that aren't strongly held by anyone means the events in the table get removed almost immediately. 

Test Plan: Confirm that events are not dropped from the table after they're added. 

Differential Revision: https://phabricator.fb.com/D1819759


secretiverhyme:
Nice catch.  Thanks!

Issue #236, opened 2015-01-20
Batch animation beginUpdates / endUpdates with ASCollectionView


secretiverhyme:
Merging #211 broke the build (fixed in 7542f7ca17626bed2e0eabf53aeea73fea2cfdac) — looks like we're expecting UITableView-style update bracketing as opposed to UICollectionView-style update blocks.  @tanliboy, once you're back, please double-check to make sure this is doing the right thing!  =]

tanliboy:
Oh, thanks for fixing it !

secretiverhyme:
There are still a couple of inconsistencies here.

In ASCollectionView:
```objective-c
/* no -performBatchUpdates:completion: override */
- (void)rangeControllerBeginUpdates:(ASRangeController *)rangeController {
}
- (void)rangeControllerEndUpdates:(ASRangeController *)rangeController {
}
```

ASTableView:
```objective-c
- (void)beginUpdates
{
  [_dataController beginUpdates];
}
- (void)endUpdates
{
  [_dataController endUpdates];
}
/* ... */
- (void)rangeControllerBeginUpdates:(ASRangeController *)rangeController
{
  ASDisplayNodeAssertMainThread();
  [super beginUpdates];
}
- (void)rangeControllerEndUpdates:(ASRangeController *)rangeController
{
  ASDisplayNodeAssertMainThread();
  [super endUpdates];
}
```

Is this intentional?  It sounds like ASRangeController expects begin/end-update methods, which won't work with UICollectionView.  Should we change this to work with update blocks?  We could implement them on ASTableView with something like...

```objective-c
- (void)rangeControllerPerformUpdates:(as_update_block_t)updates
{
  [super beginUpdates];
  updates();
  [super endUpdates];
}
```

tanliboy:
I will check and update shortly.

tanliboy:
Send #295 to fix this one.

secretiverhyme:
Awesome, thanks!  I'll review your pull requests when I've a moment.  =]

secretiverhyme:
Fixed in #295.

PR #289, opened 2015-02-04
Fix end update in group animation

tanliboy:


secretiverhyme:
cc #236.  Thanks for the fix!

secretiverhyme:
@tanliboy, want to delete the `#pragma mark Assertions` comment?

PR #290, opened 2015-02-04
Add property to quickly get the cell view for nodes in tables or collections.

andyscott:
Sometimes it is useful to have access to the cell view that's being displayed in the table or collection view.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -19,6 +19,12 @@
 //@property (atomic, retain) UIColor *backgroundColor;
 @property (nonatomic) UITableViewCellSelectionStyle selectionStyle;
 
+/**
+ * @abstract Returns the UICollectionViewCell or UITableViewCell if the node is currently
+ * visible in either a ASCollectionView or ASTableView, respectively.
+ */
+@property (nonatomic, readonly, retain) UIView *cellView;
+
 @end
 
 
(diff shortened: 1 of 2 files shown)


secretiverhyme:
Thanks for the pull request!  This is unfortunately too brittle to add to ASCellNode upstream — if anyone ever changes the way ASTableView or ASCollectionView are implemented, this property will break without warning.  A slightly safer approach would be to walk the view hierarchy looking for a superview of class `_ASTableViewCell` or `_ASCollectionViewCell` and return the first one found, but that has problems as well.

In general, having to walk up the view hierarchy (whether by a hardcoded number of `.superview` calls, class comparisons, or even `-respondsToSelector:`) is a bad sign.  In what situations have you found this property useful?  This may be a good opportunity to expand ASCellNode or improve ASTV / ASCV.

andyscott:
I had a hunch this would be too brittle. Figured I could maybe prompt the addition one way or another.

At the moment, I'm doing some calculations based off of the cell's position within my collection's scroll view. The calculations run as the user scrolls through the collection. It's kind of awkward to write `superview.superview.frame`.

secretiverhyme:
Hmm.  What about `[node.view convertRect:node.frame toView:_scrollView]`?

andyscott:
:laughing: That worked. Disregard request!

andyscott:
Related--

When the nodes are initially displayed in my collection view, I need to update a property based off of that same position. ASDK controls the regular `dataSource` and `collectionView:cellForItemAtIndexPath:`. Where/how can I get access to the collection/scroll view and the node on initial display?

secretiverhyme:
Glad it worked!  ASCollectionView is a UICollectionView subclass, so it's a scroll view.  You can implement the delegate method `-collectionView:willDisplayNodeForItemAtIndexPath:` and use `-[ASCollectionView nodeForItemAtIndexPath:]` to hook into displays.

andyscott:
That worked. The solution brought up another small issue (extra cast, at least in Swift). I filed a PR for it.

PR #293, opened 2015-02-05
Use specific types for Collection/TableView async delegate params

andyscott:
Small adjustment to make the delegates ASTableView and ASCollectionView send back the right types.


secretiverhyme:
I'd been meaning to fix this.  Thanks!

Issue #292, opened 2015-02-05
[ASTableView] Calling reloadData causing animated flickers


lxcid:
I realize that calling `reloadData` causes animation flickers though. I had this from my app and was able to repro on Kittens sample, attached is a screen capture GIF to illustrate the issue.

![2015-02-06 01_49_38](https://cloud.githubusercontent.com/assets/297156/6065756/be04bc9a-ada2-11e4-856d-3124b937a268.gif)

tanliboy:
Thanks for reporting this issue. It is caused by the multi-step animations while reloading data.
I will think about a way to work around.

lxcid:
Cool! Thanks man!

secretiverhyme:
Fixed in #295.

lxcid:
Unfortunately, it still happens. I'm on cd3a859f8ccf59897236e3ed99ff7f5a6e923fb1.

If you call `reloadData` more than once, you see the animation flicker still. :(

secretiverhyme:
@tanliboy, want to take a look?

tanliboy:
@lxcid , could you try to wrap the `reloadData` with `beginUpdate` and `endUpdate` ? 
In the initialization of ASTableView, the `reloadData` will be invoked. To shorten the initialization cost, we make `reloadData` to use batch update by default. Also, you could always group some operations (even not in single runloop) using `beginUpdate` and `endUpdate`.

lxcid:
Thanks! Wrapping them between `beginUpdates` and `endUpdates` does help solve the issue.

Should it be documented in the readme for future reference? :)

:+1:

tanliboy:
Yeah, it sounds a good idea. I will think about a way to avoid the wrapping firstly. If impossible, I will add it into doc or code comments.

lxcid:
On the latest master (9477114c868b658aed39b887766ce6557e25da60), I realise that this issue is back again even with begin/end updates.

Below is a screen capture I done to illustrate the issue on kittens. I make sure that kitten always generate the same data set.

https://www.dropbox.com/s/vmlr18lki7ll1mc/asdk_reload_issue.mp4?dl=0

weibel:
This happens to me when the datasource is async. When it is sync everything is good.

yiqin:
I have the same issue. How to resolve it?

appleguy:
[...]

PR #295, opened 2015-02-06
Fix group updates/animation

tanliboy:
Fix group updates/animation:
1) the group updates need to be done in single run loop: (fix it by set the pending blocks in the ASDataController)
2) Add performBatchUpdates into the ASCollectionView, and tested it with insertion and deletion in groups

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -98,6 +98,9 @@ static BOOL _isInterceptedSelector(SEL sel)
   ASDataController *_dataController;
   ASRangeController *_rangeController;
   ASFlowLayoutController *_layoutController;
+
+  BOOL _batchUpdateFlag;
+  NSMutableArray *_batchUpdateBlocks;
 }
 
 @end
@@ -127,6 +130,9 @@ static BOOL _isInterceptedSelector(SEL sel)
 
   _proxyDelegate = [[_ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
   super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
+
+  _batchUpdateFlag = NO;
+  _batchUpdateBlocks = [NSMutableArray array];
 
   [self registerClass:[UICollectionViewCell class] forCellWithReuseIdentifier:@"_ASCollectionViewCell"];
   
@@ -210,6 +216,13 @@ static BOOL _isInterceptedSelector(SEL sel)
 }
 
 #pragma mark Assertions.
+
+- (void)performBatchUpdates:(void (^)())updates completion:(void (^)(BOOL))completion
+{
+  [_dataController beginUpdates];
+  updates();
+  [_dataController endUpdatesWithCompletion:completion];
+}
 
 - (void)insertSections:(NSIndexSet *)sections
 {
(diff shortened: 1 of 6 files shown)


rnystrom:
@tanliboy does this fix #292 as well?

tanliboy:
Yes, the issue https://github.com/facebook/AsyncDisplayKit/issues/292 could be work around by adding the "beginUpdate" and "endUpdate" around "reloadData".

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -98,6 +98,9 @@ @interface ASCollectionView () <ASRangeControllerDelegate, ASDataControllerSourc
   ASDataController *_dataController;
   ASRangeController *_rangeController;
   ASFlowLayoutController *_layoutController;
+
+  BOOL _batchUpdateFlag;
```
secretiverhyme: instead of maintaining a flag, could we just check whether `_batchUpdateBlocks.count > 0`?  if not, can we rename this to something like `_performingBatchUpdates` so it's clearer what is being flagged?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -124,63 +109,24 @@ + (BOOL)isSizingQueue {
...
-  });
-
-  return dataUpdatingQueue;
-}
-
-+ (BOOL)isDataUpdatingQueue {
-  return kASDataUpdatingQueueContext == dispatch_get_specific(kASDataUpdatingQueueContext);
-}
-
 - (void)asyncUpdateDataWithBlock:(dispatch_block_t)block {
```
secretiverhyme: since `-asyncUpdateDataWithBlock:` is now a non-blocking call, do we have to dispatch_async() to the sizing queue before invoking it?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -208,25 +154,39 @@ - (void)initialDataLoadingWithAnimationOption:(ASDataControllerAnimationOptions)
...
 
 - (void)endUpdates {
+  [self endUpdatesWithCompletion:NULL];
+}
+
+- (void)endUpdatesWithCompletion:(void (^)(BOOL))completion
+{
   dispatch_async([[self class] sizingQueue], ^{
-    [self asyncUpdateDataWithBlock:^{
+    dispatch_async(dispatch_get_main_queue(), ^{
```
secretiverhyme: for example, here, we're trampolining to the sizing queue and then immediately trampolining to the main queue — is this necessary?

secretiverhyme:
@tanliboy, what do you think about removing the `-asyncUpdateDataWithBlock:` and `-syncUpdateDataWithBlock:` methods instead of explicit `dispatch_async()` and `dispatch_sync()` calls?  It looks like we no longer need to dispatch to the sizing queue before invoking those asynchronous update methods, unless I'm missing something, so we should be able to change

```objective-c
 - (void)beginUpdates
{
  dispatch_async([[self class] sizingQueue], ^{
    [self asyncUpdateDataWithBlock:^{
      _batchUpdateCounter++;
    }];
  });
}
```

to

```objective-c
 - (void)beginUpdates
{
    _batchUpdateCounter++;
  });
}
```

and so on.

AsyncDisplayKit/ASCollectionView.mm
tanliboy: Hmm, I will rename it to be `_performingBatchUpdates.`
The `_batchUpdateBlocks.count > 0` is not enough for the checking considering it is changed after layout in background thread.

PR #296, opened 2015-02-06
Test code (discussion only)

tanliboy:
For discussion only


secretiverhyme:
It looks like we can shelve this given #303.

PR #228, opened 2015-01-15
Let ASMultiplexImageNode skip failed image identifiers

andyscott:
A hopefully better attempt at #159.

I went for the easy route and just kept track of the failed image identifier. The logic that determines which image to load next was updated to consider the failed identifier before using the last loaded identifier.

PR #302, opened 2015-02-10
Revert "Let ASMultiplexImageNode skip failed image identifiers"

secretiverhyme:
Reverts facebook/AsyncDisplayKit#228 (cc @andyscott) — it introduced a bug where, if the best image identifier is unavailable (e.g., 404 error), ASMultiplexImageNode infinite-loops in an attempt to load it.


andyscott:
Would it be appropriate to submit a fix for this and then have the feature merged back in?

secretiverhyme:
@andyscott — sure, go for it!

Issue #288, opened 2015-02-04
ASTableView & ASCollectionView shouldn't block the main thread by default


phatmann:
I have an ASTableView with 1000 nodes. When the table is reloaded, there are several distinct visible phases:

1. No cells showing
1. Cells showing but CPU churning and UI not responsive
1. Cell subnodes showing placeholders while they load
1. Cells fully loaded and visible

I would like to hide the table until phases 1 and 2 are complete because it is not in a genuinely useful state until then. In fact, if I try to remove a cell (via a close button) during phase 2, I often run into bad access errors.

Browsing through the ASTableView code, I did not see any easy and obvious way to see what state the table is in. If anyone can point me in the right direction, I would be happy to put together a PR. My naive approach would be to dig into the node loading code in ASDataController and see what hooks I can add.

I would either add additional methods for ASTableView subclasses to hook into, or add methods to the table delegate. Suggestions welcome.

Being able to detect when phase 3 is done would be nice but is not necessary.

Worth noting: this is not much of a problem on fast modern devices, but could be an issue on older, slower devices. If you run your app in the pokey iOS Simulator you will see what I mean :-(

tanliboy:
Do you have sample code to reproduce the issues you described ? 
The ASDataController will fetch the nodes through "nodeAtIndexPath:" in the calling thread, and then dispatched to background thread in batch for layout. After the layout, the nodes will show up.
With pre-allocation on the nodes in background thread may could remove the stuck in stages 1) or 2). 
It would be great if you have the sample code so that we could dig into and optimize it.

phatmann:
@tanliboy, if you want sample code that shows the loading slowdown, just modify the Kittens sample to load a 1000 nodes instead of 20 and run in the simulator.

Do you want sample code to that illustrates the bad access error? Could be trickier since the problem is sporadic.

tanliboy:
Thanks for reply !
I will update the Kitten project to make the cell allocated in background thread, and try with a demo to illustrate the editing. The inserting/deleting need to go through the ASTableView API, because the we do the layout in async mode and the index path in the UITableView may does not match the index path in the ASTableView.

phatmann:
When you say that deleting needs to go through the ASTableView API, what do you mean? Since ASTableView overrides `deleteRowsAtIndexPaths`, isn't this automatic?

tanliboy:
[...]

Issue #258, opened 2015-01-26
Allocation/Measurement Range


rnystrom:
Instead of initializing every single node when using `ASTableView` and `ASCollectionView`, there should be a (large) range in which nodes are initialized and sized. This will be especially beneficial for large data sets so that we are not initializing and sizing nodes that are waaaaaay off screen, essentially wasting time.

Considerations:

- Nodes that *exit* this range should not be destroyed unless there is a memory warning.
- The working and fetching ranges are subsets of the allocation range.
- Unit tests

secretiverhyme:
#285 is a fantastic step in the right direction.  We're also going to need to break the assumption that all nodes *exist* at all times — currently the allocation and measurement ranges are (1) not explicitly tracked and (2) hardcoded to be "all nodes".

appleguy:
Merging this with #309 (moving them to Ship to track structural upgrades to ASDataController).

PR #303, opened 2015-02-10
Move ASTableView & ASCollectionView data fetching to background thread

tanliboy:
Add "collectionViewLockDataSourceForDataUpdating" and "collectionViewUnlockDataSourceForDataUpdating" for enable data fetching from background thread.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -35,6 +35,11 @@
 @property (nonatomic, assign) ASRangeTuningParameters rangeTuningParameters;
 
 /**
+ * Initializer.
+ */
+- (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout asyncDataFetching:(BOOL)asyncDataFetchingEnabled;
+
+/**
  * Reload everything from scratch, destroying the working range and all cached nodes.
  *
  * @warning This method is substantially more expensive than UICollectionView's version.
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -107,6 +107,26 @@
  */
 - (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtIndexPath:(NSIndexPath *)indexPath;
 
+@optional
+
+/**
+ * Indicator to lock the data source for data loading in asyn mode.
+ * We should not update the data source until the data source has been unlocked. Otherwise, it will incur data inconsistence or exception
+ * due to the data access in async mode.
+ *
+ * @param tableView The sender.
+ */
+- (void)collectionViewLockDataSourceForDataUpdating:(ASCollectionView *)collectionView;
+
+/**
+ * Indicator to unlock the data source for data loading in asyn mode.
+ * We should not update the data source until the data source has been unlocked. Otherwise, it will incur data inconsistence or exception
+ * due to the data access in async mode.
+ *
+ * @param tableView The sender.
+ */
+- (void)collectionViewUnlockDataSourceForDataUpdating:(ASCollectionView *)collectionView;
+
 @end
 
 
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -36,6 +36,14 @@
 
 /**
  * Initializer.
+ *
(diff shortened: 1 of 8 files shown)


phatmann:
This looks great, thanks for the fast turnaround! I will pull this and try and it out tomorrow.

phatmann:
So far this change is working great. No issues and the main thread is not blocked while nodes are being created.

secretiverhyme:
Very cool!  Some thoughts:

`ENABLE_ASYNC_DATA_FETCHING` and `_asyncDataFetchingEnabled` seem redundant, and the way `BEGIN_DATA_FETCHING` and `END_DATA_FETCHING` convert method contents into blocks is hacky.  Can we remove all three `#define`s in favour of a method that explicitly takes a block and processes it asynchronously if async fetching is enabled?  So,

```objective-c
- (void)method
{
  BEGIN_DATA_FETCHING
  /* do stuff */
  END_DATA_FETCHING
}
```

would become

```objective-c
- (void)method
{
  [self performDataUpdate:^{
    /* do stuff */
  }];
}
```

ASTableView asserts that its data source implements lock & unlock methods, but ASCollectionView does not.  Instead of adding a new initialiser for both of these classes and juggling these assertions, can we check whether the data source implements both lock & unlock methods at the time `-setDataSource:` is called, and use that to set ASDataController's `_asyncDataFetchingEnabled` property?  This way simply implementing those two methods will be enough to opt into the new behaviour.

Can we remove `ForDataUpdating` from the new delegate method names?  `-tableViewLockDataSource`, `-collectionViewUnlockDataSource`, and so on are easier to read.

Typo:  `asyncDataSouceLocked` should be `asyncDataSourceLocked`.

Overall this is a fantastic improvement.  We should update the Kittens sample project to use these new delegate methods, and update the documentation to clearly advise that you use these methods ("WARNING:  ASTableView and ASCollectionView performance will be adversely impacted if you don't implement `-...LockDataSource` and `-...UnlockDataSource`!").  We should also consider making these data source methods non-optional, given how critical this change is.

phatmann:
After further testing, I am unhappy to report that this pull request does not solve my original problem described in #288. I have an AKTableView with 1000 nodes using async data loading. Each node has a close button that removes the node from the table. If I tap a node's close button immediately after the nodes appear, the tapped node is not removed for several seconds. If a wait a few seconds after the nodes appear, then tapping the close button removes the node immediately. In either case, subsequent taps on a node's close button remove the node immediately. In short, when the nodes first appear, the table is not responsive. Our assumption has been that this lack of responsiveness is due to the overhead of loading nodes. Either this assumption is wrong, or this pull request does not solve the issue. I will do more profiling to see what is going on.

If we cannot solve this issue in the near term, then I would like to consider re-opening #288, which requested that there be a way to determine when the table is actually fully loaded and responsive.

phatmann:
I just did some profiling on my iPhone 5S. Over 70% of the load time is spent in ASDisplayNode's `measure`. My tap on the close button is recognized immediately, so the UI is responsive during this load time, which is expected since the measuring takes place on a background thread. However, there is a delay of several seconds between the tap and when the node is removed from the table. After all the measuring is done, subsequent taps remove the node immediately, so the measuring is what is causing the delay. My node has a fairly simple layout; more complex nodes would have even more delay.

So the issue is that visible nodes cannot be removed until _all_ nodes are measured. Should I file a new issue for this?

And back to #288 again: if we cannot solve the issue with the measuring cycle blocking node removal, I would like a way to know that all the nodes have been measured, so I can accommodate removal delays in the UI; e.g. show an activity indicator during the slow removal.

tanliboy:
Yes, this change would not solve that issue.
We are doing the editing in async mode.
For the initialization, we inserted 1000 nodes in batches. Supposed we grouped it into 50 operations, each operations will insert 20 nodes into table. These operations are queued in a serialized dispatch queue. If we want to remove a node, it will be pushed into the queue and not be executed directly.

tanliboy:
Yeah, it sounds a good  idea to add a function call to tell the data source that all operations has been take effect.

phatmann:
I filed #309 to cover the node removal delay.

phatmann:
@tanliboy, thanks for the explanation. Can we re-open and rename #288 (or file a new issue) to request a notification that all operations have been completed?

tanliboy:
@phatmann , hmm, if we could solve the issue #309 , we may don't need the notifications any more.
I will think about that problem and update there.

phatmann:
For the readers at home just joining us:

* #309 will be largely taken care of by #258 
* If #309 is solved, then #288 might not be so important

So if we can knock off #258 lots of good things start happening :-)

None of this has much to do with this issue, so I will make a similar post at #309 and continue the discussion there.

PR #308, opened 2015-02-12
Allow touches and long press on ASTextNode

### LINKED ISSUE #262: ASTextNodes in a table view are unhighlighted automatically
The touch within an ASTextNode is cancelled right away causing links to unhighlight, even if the table view does not scroll at all.

rnystrom:
fixes #262


secretiverhyme:
Nice find!

PR #285, opened 2015-02-04
Custom Ranges Mechanism

rnystrom:
This PR introduces the concept of the "Preload Range" which is defaulted to 2 screens trailing and leading of the current visible range (e.g. the nodes on the screen). Included in this update:

- Preload range concept implemented into the `ASLayoutController`, `ASFlowLayoutController`, `ASRangeController`, `ASTableView`, and `ASCollectionView`
- `fetchExternalContent` and `purgeFetchedContent` APIs to fetch and destroy content that is expensive to fetch
- `ASNetworkImageNode` to unload its fetched contents in the new `purgeFetchedContents` method

fixes #257

PR #312, opened 2015-02-13
Batch Fetching API

rnystrom:
This PR introduces a new API for `ASTableView` and `ASCollectionView` to be notified when scrolling near the tail end of the available content. Notifications are limited to vertically scrolling near the bottom of `ASTableView`, and vertical and horizontal scrolling to the bottom and right, respectively, for `ASCollectionView`.

There are 2 new delegate methods for `ASTableView` and `ASCollectionView`:

- `-shouldBatchFetchForCollectionView:` returns a boolean to determine if a collection view should initiate the fetching process.
- `-collectionView:beginBatchFetchingWithContext:` is called when near the end of content.

`ASTableView`'s APIs are identical and differ only by classes and naming conventions.

The `ASBatchContext` object passed to `-collectionView:beginBatchFetchingWithContext:` has the method `-completeBatchFetching:` which **must** be called when finished fetching and processing your content. You can call this method synchronously, or after an asynchronous process such as fetching content from the network. 

If you fail to call `-completeBatchFetching:`, you will never be notified to fetch batch content again. This mechanism exists to prevent over-notifying and over-fetching.

The API adds `@property (assign) CGFloat leadingScreensForBatching` to both `ASTableView` and `ASCollectionView`, which defaults to 1.0. Changing this value will adjust the distance from the end of content that batch notifications will fire.

The Kittens project has been updated with an example of how to batch load content. The ASCollectionView project has been updated to simply demonstrate batch notifications with horizontal scrolling.

Unit tests are included, but this is the first iteration, so YMMV :smile:

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -11,6 +11,7 @@
 #import <AsyncDisplayKit/ASRangeController.h>
 #import <AsyncDisplayKit/ASCollectionViewProtocols.h>
 #import <AsyncDisplayKit/ASBaseDefines.h>
+#import <AsyncDisplayKit/ASBatchContext.h>
 
 
 @class ASCellNode;
@@ -61,6 +62,13 @@
  * The application should not update the data source while the data source is locked, to keep data consistence.
  */
 - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout asyncDataFetching:(BOOL)asyncDataFetchingEnabled;
+
+/**
+ * The number of screens left to scroll before the delegate -collectionView:beginBatchFetchingWithContext: is called.
+ *
+ * Defaults to one screenful.
+ */
+@property (nonatomic, assign) CGFloat leadingScreensForBatching;
 
 /**
  * Reload everything from scratch, destroying the working range and all cached nodes.
@@ -165,6 +173,34 @@
 
 - (void)collectionView:(ASCollectionView *)collectionView willDisplayNodeForItemAtIndexPath:(NSIndexPath *)indexPath;
 - (void)collectionView:(ASCollectionView *)collectionView didEndDisplayingNodeForItemAtIndexPath:(NSIndexPath*)indexPath;
+
+/**
+ * Tell the collectionView if batch fetching should begin.
+ *
+ * @param collectionView The sender.
+ *
+ * @discussion Use this method to conditionally fetch batches. Example use cases are: limiting the total number of
+ * objects that can be fetched or no network connection.
+ *
+ * If not implemented, the collectionView assumes that it should notify its asyncDelegate when batch fetching
+ * should occur.
+ */
+- (BOOL)shouldBatchFetchForCollectionView:(UICollectionView *)collectionView;
+
+/**
+ * Receive a message that the collectionView is near the end of its data set and more data should be fetched if 
+ * necessary.
+ *
+ * @param tableView The sender.
+ * @param context A context object that must be notified when the batch fetch is completed.
+ *
+ * @discussion You must eventually call -completeBatchFetching: with an argument of YES in order to receive future
+ * notifications to do batch fetches.
+ *
+ * UICollectionView currently only supports batch events for tail loads. If you require a head load, consider
+ * implementing a UIRefreshControl.
+ */
+- (void)collectionView:(UICollectionView *)collectionView beginBatchFetchingWithContext:(ASBatchContext *)context;
 
 @end
 
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -185,7 +185,7 @@
  * If not implemented, the collectionView assumes that it should notify its asyncDelegate when batch fetching
  * should occur.
  */
-- (BOOL)shouldBatchFetchForCollectionView:(UICollectionView *)collectionView;
+- (BOOL)shouldBatchFetchForCollectionView:(ASCollectionView *)collectionView;
 
 /**
  * Receive a message that the collectionView is near the end of its data set and more data should be fetched if 
@@ -200,7 +200,7 @@
  * UICollectionView currently only supports batch events for tail loads. If you require a head load, consider
  * implementing a UIRefreshControl.
  */
-- (void)collectionView:(UICollectionView *)collectionView beginBatchFetchingWithContext:(ASBatchContext *)context;
+- (void)collectionView:(ASCollectionView *)collectionView beginBatchFetchingWithContext:(ASBatchContext *)context;
 
 @end
 
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -175,6 +175,21 @@
 - (void)collectionView:(ASCollectionView *)collectionView didEndDisplayingNodeForItemAtIndexPath:(NSIndexPath*)indexPath;
 
 /**
+ * Receive a message that the collectionView is near the end of its data set and more data should be fetched if 
+ * necessary.
+ *
+ * @param tableView The sender.
+ * @param context A context object that must be notified when the batch fetch is completed.
+ *
+ * @discussion You must eventually call -completeBatchFetching: with an argument of YES in order to receive future
+ * notifications to do batch fetches. This method is called on a background queue.
+ *
+ * UICollectionView currently only supports batch events for tail loads. If you require a head load, consider
+ * implementing a UIRefreshControl.
+ */
+- (void)collectionView:(ASCollectionView *)collectionView willBeginBatchFetchWithContext:(ASBatchContext *)context;
+
+/**
  * Tell the collectionView if batch fetching should begin.
  *
  * @param collectionView The sender.
@@ -186,21 +201,6 @@
  * should occur.
  */
 - (BOOL)shouldBatchFetchForCollectionView:(ASCollectionView *)collectionView;
-
-/**
- * Receive a message that the collectionView is near the end of its data set and more data should be fetched if 
- * necessary.
- *
- * @param tableView The sender.
- * @param context A context object that must be notified when the batch fetch is completed.
- *
- * @discussion You must eventually call -completeBatchFetching: with an argument of YES in order to receive future
- * notifications to do batch fetches.
- *
- * UICollectionView currently only supports batch events for tail loads. If you require a head load, consider
- * implementing a UIRefreshControl.
- */
-- (void)collectionView:(ASCollectionView *)collectionView beginBatchFetchingWithContext:(ASBatchContext *)context;
 
 @end
 
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -384,13 +384,13 @@ static BOOL _isInterceptedSelector(SEL sel)
   }
 }
 
-- (BOOL)shouldFetchBatch
+- (BOOL)shouldBatchFetch
 {
-  if ([self.asyncDelegate respondsToSelector:@selector(shouldBatchFetchForCollectionView:)]) {
-    return [self.asyncDelegate shouldBatchFetchForCollectionView:self];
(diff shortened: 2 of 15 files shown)


smeis:
Looks greate! One thing: the new delegate methods have UITableView and UICollectionView as the sender types, those should probably be ASTableView and ASCollectionView if I'm not mistaken.

rnystrom:
Whoops, good catch already!

lxcid:
Just tried and it look awesome. This will be useful! One question, often there's loading indicator at the bottom, how can we add such indicator?

examples/Kittens/Sample/ViewController.m
```diff
@@ -105,4 +112,33 @@ - (BOOL)tableView:(UITableView *)tableView shouldHighlightRowAtIndexPath:(NSInde
...
+  NSLog(@"adding kitties");
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    sleep(1);
+    dispatch_async(dispatch_get_main_queue(), ^{
+      NSArray *moarKittens = [self createLitterWithSize:kLitterBatchSize];
+
+      NSMutableArray *indexPaths = [[NSMutableArray alloc] init];
+      NSInteger existingKittens = _kittenDataSource.count;
+      for (NSInteger i = 0; i < moarKittens.count; i++) {
+        [indexPaths addObject:[NSIndexPath indexPathForRow:existingKittens + i inSection:0]];
```
lxcid: There's a special-case first row taking up the index path {0, 0}. So the starting position (`existingKittens`) is technically off by 1 row.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -340,6 +350,43 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
...
+- (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoint)velocity targetContentOffset:(inout CGPoint *)targetContentOffset
+{
+  [self handleBatchFetchScrollingToOffset:*targetContentOffset];
+
+  if ([_asyncDelegate respondsToSelector:@selector(scrollViewWillEndDragging:withVelocity:targetContentOffset:)]) {
+    [_asyncDelegate scrollViewWillEndDragging:scrollView withVelocity:velocity targetContentOffset:targetContentOffset];
+  }
+}
+
+- (BOOL)shouldFetchBatch
```
secretiverhyme: should fetch batch, should batch fetch, someone call Theodor Geisel

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -120,4 +128,32 @@
...
+ *
+ * @param collectionView The sender.
+ *
+ * @discussion Use this method to conditionally fetch batches. Example use cases are: limiting the total number of
+ * objects that can be fetched or no network connection.
+ *
+ * If not implemented, the collectionView assumes that it should notify its asyncDelegate when batch fetching
+ * should occur.
+ */
+- (BOOL)shouldBatchFetchForCollectionView:(ASCollectionView *)collectionView;
```
secretiverhyme: `-shouldBatchFetchForCollectionView:` versus `-collectionView:beginBatchFetchingWithContext:` is a little confusing, especially since this method is declared before that one — I can imagine someone getting mixed up and actually executing the batch fetch here.  No strong objections, but maybe we could tweak the name to be more unambiguous?

Or, maybe we could remove this method and have people call `[context cancelBatchFetching]` if they're not yet ready to fetch something?  That's clunky, though.

AsyncDisplayKit/ASCollectionView.mm
secretiverhyme: nit:  this method uses `self.asyncDelegate`, the other ones use `_asyncDelegate`

AsyncDisplayKit/ASTableView.mm
```diff
@@ -347,6 +355,43 @@ - (void)tableView:(UITableView *)tableView didEndDisplayingCell:(UITableViewCell
...
+- (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoint)velocity targetContentOffset:(inout CGPoint *)targetContentOffset
+{
+  [self handleBatchFetchScrollingToOffset:*targetContentOffset];
+
+  if ([_asyncDelegate respondsToSelector:@selector(scrollViewWillEndDragging:withVelocity:targetContentOffset:)]) {
+    [_asyncDelegate scrollViewWillEndDragging:scrollView withVelocity:velocity targetContentOffset:targetContentOffset];
+  }
+}
+
+- (BOOL)shouldFetchBatch
```
secretiverhyme: same remarks as above

AsyncDisplayKit/Details/ASBatchContext.m
```diff
@@ -0,0 +1,60 @@
...
+
+typedef NS_ENUM(NSInteger, ASBatchContextState) {
+  ASBatchContextStateFetching,
+  ASBatchContextStateCancelled,
+  ASBatchContextStateCompleted
+};
+
+@interface ASBatchContext ()
+{
+  ASBatchContextState _state;
```
secretiverhyme: we don't guard this with a lock.  ASTableView / ASCollectionView support editing & updates off-main and we don't specify thread-affinity for notifying ASDK code that the batch fetching completed.  will this go :boom:?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -340,6 +350,43 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
...
+{
+  ASDisplayNodeAssert(_batchContext != nil, @"Batch context should exist");
+
+  if (![self shouldFetchBatch]) {
+    return;
+  }
+
+  if (ASDisplayShouldFetchBatchForContext(_batchContext, [self scrollDirection], self.bounds, self.contentSize, targetOffset, _leadingScreensForBatching)) {
+    [_batchContext beginBatchFetching];
+    [self.asyncDelegate collectionView:self beginBatchFetchingWithContext:_batchContext];
```
secretiverhyme: we should either dispatch_async() this to a general concurrent background queue, or warn folks in the header that this method is called on main and that they should trampoline elsewhere

AsyncDisplayKit/Details/ASBatchFetching.h
```diff
@@ -0,0 +1,36 @@
...
+/**
+ @abstract Determine if batch fetching should begin based on the state of the parameters.
+ @param context The batch fetching context that contains knowledge about in-flight fetches.
+ @param scrollDirection The current scrolling direction of the scroll view.
+ @param bounds The bounds of the scrollview.
+ @param contentSize The content size of the scrollview.
+ @param targetOffset The offset that the scrollview will scroll to.
+ @param leadingScreens How many screens in the remaining distance will trigger batch fetching.
+ @return Whether or not the current state should proceed with batch fetching.
+ @discussion This method is broken into a category for unit testing purposes and should be used with the ASTableView and
```
secretiverhyme: move this file to the Private/ directory?  those headers aren't exported

AsyncDisplayKit/Details/ASBatchContext.h
```diff
@@ -0,0 +1,60 @@
...
+ * @discussion Call this method only when something has corrupted the batch fetching process. Calling this method should
+ * be left to the owner of the batch process unless there is a specific purpose.
+ */
+- (void)cancelBatchFetching;
+
+/**
+ * Notify the context object that fetching has started.
+ *
+ * @discussion Call this method only when you are beginning a fetch process. This should really only be called by the 
+ * context object's owner. Calling this method should be complimented with -completeBatchFetching:.
```
secretiverhyme: s/complimented/complemented/ (but maybe you meant "paired"?)

AsyncDisplayKit/Details/ASBatchContext.m
rnystrom: :sob: dumb

I'm updating and using the `ASDN::MutexLocker l(_propertyLock)` pattern that we use in `ASDisplayNode.mm`.

AsyncDisplayKit/ASCollectionView.mm
rnystrom: I'm going to both pop this to a bg queue and note in the docs that it is on a bg queue.

rnystrom:
@secretiverhyme thanks so much for the feedback! Really really good catches. I rebased on master (woof) and made some updates in a separate commit.

jeroenleenarts:
I've noticed that the batching does not work when working with a low batch size. Any thoughts on that? Some API's have very low batch sizes and it would be great if it also works in such a case.

rnystrom:
@jeroenleenarts Can you explain the low batch size issue a little more? Do you mean that if your initial data set is very small (or your cells are small) that the API won't fire? If so I definitely need to look into that.

Any specific examples on reproducing this would be really helpful!

jeroenleenarts:
If you have a batch size so small that all initial data fits into one page of the collection view. Then there is no way to trigger the did end dragging event which is the trigger for the batch size behavior.

jeroenleenarts:
@rnystrom This branch demonstrates the behavior: https://github.com/jeroenleenarts/SyncDisplayKitTrial

Look for the pragma mark "change me to a higer value" and change it's value to a higher value (beyond one page) and all works.

rnystrom:
@jeroenleenarts awesome thanks! I'll look into this. @secretiverhyme I probably need to update this PR if this is the case and figure out an elegant solution that doesn't thrash the API call when dealing with small data sets or rows.

lxcid:
I think a decent workaround for this edge case is to look out for last index path in `- [ASTableView tableView:willDisplayNodeForRowAtIndexPath:]`.

secretiverhyme:
Flagging for revision for the small batch size case.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -49,6 +50,13 @@
 - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout asyncDataFetching:(BOOL)asyncDataFetchingEnabled;
 
 /**
+ * The number of screens left to scroll before the delegate -collectionView:beginBatchFetchingWithContext: is called.
```
secretiverhyme: tiny nit:  it might be clearer to say something like "how many screens away from the end of the collection view's content you have to be for batch fetching to kick in" — I initially read this as "okay, once we've scrolled `leadingScreensForBatching` more screens, we'll call this delegate method"

AsyncDisplayKit/Details/ASBatchContext.mm
```diff
@@ -0,0 +1,68 @@
...
+
+- (BOOL)batchFetchingWasCancelled
+{
+  ASDN::MutexLocker l(_propertyLock);
+  return _state == ASBatchContextStateCancelled;
+}
+
+- (void)completeBatchFetching:(BOOL)didComplete
+{
+  if (didComplete) {
```
secretiverhyme: what was the usecase for calling `[context completeBatchFetching:NO]`?  would it be appropriate to set `_state` to `.Cancelled` if we didn't successfully complete?

AsyncDisplayKit/Details/ASBatchFetching.h
secretiverhyme: flagging this note again, if we don't want ASBatchFetching.h exported let's move it to Private/.  also, nit, this isn't a category =]

secretiverhyme:
Looks good.  We can land this now and make changes to support small batches in a follow-up patch, if you prefer.

secretiverhyme:
#285 merged and it's time to rebase again

rnystrom:
Story of my life!

AsyncDisplayKit/Details/ASBatchContext.mm
lxcid: I am wondering the same as well. calling it with `NO` seems to be a no-op. Couldn't think of a situation where this will be useful.

jeroenleenarts:
I've implemented a reasonable workaround in my trial of display kit. By chaining a couple of network requests in a ascollectionview batch request.

phatmann:
Why is batch fetching implemented in the delegate? Seems like it should be in the data source, since it controls data flow.

PR #323, opened 2015-02-23
Change placeholder boolean to time interval

### LINKED ISSUE #282: Change placeholder API to placeholderFadeInterval
For finer control of the duration of the placeholder animation. Default to 0.0 for no fade animation.

rnystrom:
fixes #282

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -332,11 +332,11 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
 @property (nonatomic, assign) BOOL placeholderEnabled;
 
 /**
- * @abstract Toggle to fade-out the placeholder when a node's contents are finished displaying.
+ * @abstract Set the time it takes to fade out the placeholder when a node's contents are finished displaying.
  *
- * @discussion Defaults to NO.
+ * @discussion Defaults to 0 seconds.
  */
-@property (nonatomic, assign) BOOL placeholderFadesOut;
+@property (nonatomic, assign) NSTimeInterval placeholderFadeDuration;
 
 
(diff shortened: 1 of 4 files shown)


secretiverhyme:
What about having *both* `placeholderFadeEnabled` and `placeholderFadeDuration`?  I'm concerned about requiring folks to set a duration in order to use placeholders, since they may end up doing something ad-hoc (0.1f here, 0.3f there) just to get them working.

Looks like we're now disabling placeholders by default.  Intentional?

rnystrom:
We previously disabled placeholder animations by default. Just keeping it consistent.

Regarding the duration/enabled bit, what if we keep `placeholderFadeEnabled` but just override the setter/getter to adjust the duration value? That way we are mucking w/ a duration & bool when deciding when to animate.

Thoughts?

secretiverhyme:
> We previously disabled placeholder animations by default. Just keeping it consistent.

Oh right, sorry — I misread this as affecting whether *placeholders* were enabled, not their animations.  Carry on, then!

I think keeping to one property is the way to go.  Let's stick with the NSTimeInterval.  We could even make it a class property instead of an instance one, favouring consistency over configurability; normally I wouldn't suggest such a thing, but given how limited the built-in placeholder animations are, perhaps that's not such a bad tradeoff.

secretiverhyme:
Let's ditch the `placeholderFadeEnabled` property entirely and then merge this in =]

rnystrom:
The only reason I'm keeping `placeholderFadeEnabled` currently is to deprecate. I'd say we push this in 1.3 and kill it in 1.4.

secretiverhyme:
Push this in 1.2 and kill it in 1.3, you mean?  Sounds good.  =]

Issue #89, opened 2014-11-05
ASTextNode doesn't render NSAttributedString properly when style.alignment = NSTextAlignmentCenter;


aaronschubert0:
See:
![screen shot 2014-11-05 at 14 18 53](https://cloud.githubusercontent.com/assets/9281756/4919053/ce75bd26-64f6-11e4-87e5-6a354ab5922d.png)

mtitolo:
+1 this is a problem for us.

In addition: When the text is only 1 line, it's not shown at all.

Code to reproduce:
```
    NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
    paragraphStyle.alignment = NSTextAlignmentCenter;
    
    NSDictionary *attributes = @{
                                 NSForegroundColorAttributeName: [UIColor whiteColor],
                                 NSFontAttributeName: [UIFont systemFontOfSize:17],
                                 NSParagraphStyleAttributeName: paragraphStyle,
                                 };
    NSAttributedString *testString = [[NSAttributedString alloc] initWithString:@"This is a test." attributes:attributes];
    text.attributedString = testString;
    [text measure:CGSizeMake(self.view.frame.size.width, 21)];
    text.frame = CGRectMake(200, 300, text.calculatedSize.width, text.calculatedSize.height);
    [self.view addSubview:text.view];
```

ocrickard:
OK, yep, definitely a bug, just repro'd locally and have a fix I'll put up in a minute.  However, we should note that this kills one of our optimizations, but in the name of correctness I'm happy we found it!  Thanks for reporting.

secretiverhyme:
Fixed in #329.

nivanchikov:
@ocrickard, @secretiverhyme 
I faced the alignment issue after the fix was introduced in #329.

![img_0324](https://cloud.githubusercontent.com/assets/1830010/6493524/e6c2dba4-c2c3-11e4-8046-b3caf0e4fe88.PNG)


Here's my setup  

```objc
- (instancetype) initWithUser: (SMUser*) user
{
	if (self = [super init])
	{
		self.displaysAsynchronously = NO;
		self.backgroundColor = [UIColor whiteColor];

		self.nameTextNode = (
		{
			NSAttributedString* text = nil;

			if ([user.fullName length])
				text = [[NSAttributedString alloc] initWithString: user.fullName
													   attributes: [SMDirSpecialistCellNode _nameTextAttributes]];

[...]

PR #329, opened 2015-02-24
Fix text node positioning for non-left alignment

ocrickard:
We were using the cached renderer even if the bounds differed from the constraining size as a performance optimization, and suggesting people measure instead when they need to for performance reasons, but this leads to some pretty broken layouts using simple reproduction code, as reported in https://github.com/facebook/AsyncDisplayKit/issues/89.


secretiverhyme:
Thanks @ocrickard!  I'll merge this in once Travis CI builds pass.

PR #336, opened 2015-02-25
Proposed initial implementation for supporting custom layouts

RCacheaux:
This is by no means a complete implementation. Sharing to get feedback and to use as a starting point for supporting custom layouts.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -497,7 +497,7 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
 - (NSArray *)nodesAtIndexPaths:(NSArray *)indexPaths
 {
   ASDisplayNodeAssertMainThread();
-  return ASFindElementsInMultidimensionalArrayAtIndexPaths(_nodes, indexPaths);
+  return ASFindElementsInMultidimensionalArrayAtIndexPaths(_nodes, [indexPaths sortedArrayUsingSelector:@selector(compare:)]);
 }
 
 @end


rnystrom:
@RCacheaux this is so great! Thank you for putting this up. I've been anxiously hoping your would :wink: 

I'll spin through it and take some notes. If you've got a sample project or anything to share too that would be awesome so we can experiment/tweak/profile things.

And keep us in the loop whenever you think its in a state to be merged. So awesome.

AsyncDisplayKit/Details/ASCollectionViewLayoutController.m
```diff
@@ -0,0 +1,96 @@
...
+  for (UICollectionViewLayoutAttributes *la in layoutAttributes) {
+    [indexPathSet addObject:la.indexPath];
+  }
+  
+  return indexPathSet;
+}
+
+- (CGRect)workingRangeRectWithScrollDirection:(enum ASScrollDirection)scrollDirection viewportSize:(CGSize)viewportSize {
+  if (_layoutDirection == ASFlowLayoutDirectionHorizontal) {
+    ASDisplayNodeAssert(scrollDirection == ASScrollDirectionNone || scrollDirection == ASScrollDirectionLeft || scrollDirection == ASScrollDirectionRight, @"Invalid scroll direction");
```
rnystrom: Wouldn't we want to eventually support up/down?

rnystrom:
Some immediate notes/thoughts:

My custom ranges PR is going to break a lot of things with the [layout controller](https://github.com/facebook/AsyncDisplayKit/pull/285/files?diff=unified#diff-7a321959fd5e24cecb3caa7d2c702046), but mostly just naming and such. Take a look at that and the new `ASFlowLayoutController` to get an idea of what we'd need to update.

This looks like it might be pretty darn simple! Just using `-layoutAttributesForElementsInRect:` could be awesome. We will definitely need to put together some type of sample/demo app to check this out, especially with #285 and #312 being merged soonish. I'm thinking a Pinterest-esque demo would be cool.

RCacheaux:
@rnystrom sounds great; I'll merge those branches into mine and get the layout controller conforming.

Yup, I think the only thing that's missing here is preloading in multiple dimensions. My initial thoughts are to build some type of hallo-like range that will pre-load the leading screenfuls on both axes in the direction of movement.

Agree on the Pinterest-esque demo! Would be sick.

PR #348, opened 2015-03-03
Custom collection view layouts

RCacheaux:
More progress towards fully supporting custom layouts. This adds 2-axis range calculations for render range type. Leaving this up here for reference. It's still in development, but would love any feedback. Still needs some TLC.


rnystrom:
Sorry for the delay in responding to this, we've been pretty slammed lately! This looks really awesome and I'm going to play around with it tomorrow.

Just curious, what would be your *pie in the sky* features that you would add on top of this PR? Mostly wondering what is "still in development", what roadblocks you've run into, etc.

Seriously, thanks a lot for digging into this. Going to be a huge win!

appleguy:
@RCacheaux — checking in here, any thoughts on this work?  I would love to see it brought to master; in fact, I will probably have a use case for it soon, but don't have bandwidth to undertake the implementation myself.  If you do have time to spin up again, I'll review more deeply to see if any feedback would be helpful.

PR #353, opened 2015-03-05
Improve `hitTestSlop` sample code

phatmann:
`hitTestSlop` sample code should make both width and height larger, for those of us that copy and paste such things

### DIFF
--- a/docs/guide/1-introduction.md
+++ b/docs/guide/1-introduction.md
@@ -139,9 +139,10 @@ nice if there were a more elegant way?
   // size all the things
   /* ... */
 
-  // make the tap target taller
+  // make the tap target larger
+  CGFloat extendX = roundf( (44.0f - size.width) / 2.0f );
   CGFloat extendY = roundf( (44.0f - size.height) / 2.0f );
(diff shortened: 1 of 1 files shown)


secretiverhyme:
Thanks for the pull request!  This unfortunately introduces a bug — if the width is less than 44pt, we'll expand the tap target accordingly, but if the width is greater, we'll *shrink* the tap target.  Hit regions aren't visually obvious, and horizontal hit slop isn't usually necessary for horizontally-laid-out text, so this could trip someone up.  What about using a constant instead?

phatmann:
Nice catch on the bug, this is bad and should be fixed. But a few things to note:

* The existing code suffers from the same issue regarding height. That is, the hit region can shrink instead of grow.
* For image buttons with small images, growing the width is just as important as height.

I understand now that the sample code was purely just a way to illustrate how `hiTestSlop` might be used for a textual button. So feel free to close this PR if you feel that my changes are not appropriate in this context.

Or I can submit an amended version that prevents shrinking if you want it. Let me know

secretiverhyme:
Yup, the existing code would suffer from this bug if used with different input.  Let's leave the sample be since it's primarily for illustration purposes.  Thanks!  =]

PR #354, opened 2015-03-05
return the processed image instead of the original image when it's being stretched

343max:


secretiverhyme:
Oh dear.  Nice find!

PR #356, opened 2015-03-06
Fix method documentation in ASCollectionView.

nguyenhuy:


secretiverhyme:
Nice find, thanks @nguyenhuy!

Issue #358, opened 2015-03-06
If you turn on voice over (accessibility), when attempt to create table view.


lxcid:
```
2015-03-07 07:04:26.546 Sample[7073:1162802] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[NSProxy methodSignatureForSelector:] called!'
*** First throw call stack:
(0x184dba59c 0x19550c0e4 0x184dba4dc 0x185c7d304 0x184dbe160 0x184cc2b6c 0x18320b278 0x18320b348 0x18963df00 0x1000fb944 0x100095c34 0x1000970e4 0x1895da6f0 0x1897f1688 0x1897f3e08 0x1897f24a0 0x18d02162c 0x184d72a28 0x184d71b30 0x184d70154 0x184c9d0a4 0x1895d3aac 0x1895ceaa4 0x100097c54 0x195b7aa08)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

lxcid:
This is the causing method https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTableView.mm#L153

lxcid:
When I remove the line (and the related line above), the crash don't happen anymore. I'm not sure if this is what you intended. The problem occurs when proxy delegate target is nil.

rnystrom:
duplicate of #349

Issue #300, opened 2015-02-10
Not setting ASTableView asyncDelegate causes cell's calculatedSize to be ignored


mtitolo:
When I create a new ASTableView, if I don't set the asyncDelegate, the size returned from `calculateSizeThatFits:` in a cell is ignored, and the entire table is drawn at the default row height. Setting the delegate (even if it's not being used) fixes the problem.

The tableview should respect the cellnode's calculatedSize regardless of an asyncDelegate being set.

Gist to reproduce: https://gist.github.com/mtitolo/f07460dfa0c7d1b52982
Screenshot:
![ios simulator screen shot feb 9 2015 3 59 57 pm](https://cloud.githubusercontent.com/assets/763698/6118748/adaeb3e8-b075-11e4-989b-a93e52b6b7fb.png)

secretiverhyme:
Thanks for the report! This was fixed in #237, which will be included in the upcoming 1.2 release.

Issue #299, opened 2015-02-09
ASCellNode sized incorrectly by ASTableView if delegate not set


isair:
In my controller if I set my table view's delegate then the cells are sized correctly, if I remove that line then my cells get sized to the default row height of the table view for some reason. I implemented no functions from the table view delegate protocol in the controller.

secretiverhyme:
Thanks for the report!  This was fixed in #237, which will be included in the upcoming 1.2 release.

RuiAAPeres:
@secretiverhyme I was able to reproduce this again with the current version (if delegate is not set, the cell's height will be 49.0f)

nguyenhuy:
Interesting. I suppose that you are using v1.2.3, right? It's hard for me to believe that such an old bug like this comes back, although it might be possible due to recent changes. Can you reproduce it in one of our examples, like Kitten example? What do your set up code and node(s) look like?

RuiAAPeres:
Hey, I am using carthage and I am pointing to this [commit](https://github.com/facebook/AsyncDisplayKit/commit/c8518c7fa6304f99db84b5480afce8f0b6ddb66e). I will check in one of the examples later.

appleguy:
@RuiAAPeres thanks for confirming the commit you're using — needless to say, that one should be pretty current and we wouldn't expect this issue to be occurring.  Please do open a new task with additional details as soon as you have a chance, as this is the only report I've seen of this for a long time.

RuiAAPeres:
Hey @appleguy you can find the testing project [here](https://dl.dropboxusercontent.com/u/14102938/TestDelegate.zip). It was build with the following:

* Xcode 7.0.1 (7A1001)
* Carthage 0.8.0
* Tested on simulator iPhone6s Plus with iOS 9.0

**Expected results:**

* Independently of line 21 (`self.tableView.asyncDelegate = self`) being commented or not, the cells will have a random size.

**Actual results:**

*  If it's commented, the cells height will use the default one (49points). If not, they will have a random size.

PR #359, opened 2015-03-07
Table/Collection view proxy with a nil target conflict with certain accessibility functions.

### LINKED ISSUE #349: NSInvalidArgumentException on ASCollectionView initialization
Hi there,
On the latest master, I am getting the following exception whenever I try to initialize an `ASCollectionView`.  Here is the stack trace:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[NSProxy methodSignatureForSelector:] called!'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000105e10f35 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x0000000105aa9bb7 objc_exception_throw + 45
	2   CoreFoundation                      0x0000000105e10e6d +[NSException raise:format:] + 205
	3   Foundation                          0x00000001056f1227 -[NSProxy methodSignatureForSelector:] + 43
	4   CoreFoundation                      0x0000000105d6ffc0 ___forwarding___ + 288
	5   CoreFoundation                      0x0000000105d6fe18 _CF_forwarding_prep_0 + 120
	6   UIKit                               0x0000000115ad2e62 -[UIScrollViewAccessibility _accessibilitySetDelegateClearer:] + 150
	7   UIKit                               0x0000000115ad2f00 -[UIScrollViewAccessibility setDelegate:] + 86
	8   UIKit                               0x0000000106f0f759 -[UICollectionView setDelegate:] + 49
	9   MyApp                              0x0000000104b7a64b -[ASCollectionView initWithFrame:collectionViewLayout:asyncDataFetching:] + 1627
	10  MyApp                              0x0000000104b79f53 -[ASCollectionView initWithFrame:collectionViewLayout:] + 163
```

This does not happen in the latest published release, but I am trying to use master so I can use the new `maximumLineCount` property on `ASTextNode`.

Any ideas/workarounds?
Thanks!

lxcid:
Mainly affect voice over, speech and selection control, it cause crashes.
We guard against creating nil targeted proxies similar to how data source is already doing so.

This will resolve #349 and #358.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -147,9 +147,6 @@ static BOOL _isInterceptedSelector(SEL sel)
 
   _leadingScreensForBatching = 1.0;
 
-  _proxyDelegate = [[_ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
-  super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
-
   _asyncDataFetchingEnabled = asyncDataFetchingEnabled;
   _asyncDataSourceLocked = NO;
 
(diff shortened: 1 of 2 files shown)


secretiverhyme:
Thanks for the pull request!  The nil-target proxy object is necessary for ASTableView and ASCollectionView to function correctly when `asyncDelegate` is not set (fixed in #237; see also #231, #233, #299, #300).  As such, we'll need to fix the actual _ASTableViewProxy and _ASCollectionViewProxy implementations to fix #349.

secretiverhyme:
(cc @rnystrom — this would be a good thing to write a unit test for)

lxcid:
I see. That make sense. I made some more diagnostic. Will post on the relevant issue. I hope it will helps! Currently, I needed to avoid the crash, I look forward to new solution. :)

PR #367, opened 2015-03-10
Thread safe creation and canceling of download tasks

### LINKED ISSUE #364: Race condition in -[ASNetworkImageNode fetchRemoteData]
In ASNetworkImageNode:

```objective-c
      } else {
          // NSURLSessionDownloadTask will do file I/O to create a temp directory. If called on the main thread this
          // will cause significant performance issues.
          dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
              [self _downloadImageWithCompletion:finished];
          });
      }
```

The following race condition exists:

1.  image node enters fetching range, this `dispatch_async()` is triggered
2.  image node exits fetching range, `_cancelImageDownload` is synchronously executed
3.  this asynchronous block executes and image downloading continues

Let's move this dispatch out of the image node and into ASBasicImageDownloader, and let the downloader object be responsible for cancelling not-yet-executed download blocks.  We should probably also update the `ASImageDownloaderProtocol` docblock to require that implementations not block the main thread — we shouldn't work around this in individual nodes.  h/t @appleguy

rnystrom:
fixes #364


rnystrom:
Did Travis flake? From the logs it doesn't look like a build issue. FWIW I just rebased (nothing to rebase) and ran tests locally and everything is a-ok.

rnystrom:
There are some internals that still aren't fixed. Found a more suitable test bed for this and will be working out the issues today.

rnystrom:
Ok g2g :+1:

secretiverhyme:
Looking good!

* Let's make sure not to rename `ASImageDownloaderProtocol` to `ASBasicImageDownloaderProtocol`.
* Let's fully revert ebe614b8c90083f70fe93a8121a6806c2b7afa9d and remove both dispatch_async() calls in ASNetworkImageNode's image loading code — we should make sure download requests and cache fetches have the same thread affinity.
* Can we add a comment to ASImageProtocols.h indicating that both methods must not block the calling thread?

rnystrom:
Find & Replace ftw... :-1: 

Good catches, updated!

Issue #315, opened 2015-02-18
[ASDisplayNode] Replace the working range window with manual display system.


rnystrom:
*Capturing enhancement discussions*

Create an enum for `ASDisplayNode` that captures the current state of a node in its lifecycle.

>blank slate -> data loaded (changing this calls -invalidateCalculatedSize, eg the attributed string or image property) -> measurement -> layout -> rendering -> all done

The node can react to what happens at each state change, and other mechanisms (such as ranges) can rely on the state and changes in order to perform certain actions.

Changes should be rolled out into phases:

- Create enum & flag on `ASDisplayNode` that is changed in appropriate methods. Flag reflects state in node lifecycle.
- React to state changes
- Wire up ranges and any other mechanisms that can piggy-back off changes
- Remove Core Animation hooks and instead rely on state changes

Any more thoughts?

appleguy:
Important history / context is in https://github.com/facebook/AsyncDisplayKit/pull/371

appleguy:
This is finally happening: #828

appleguy:
This is completely implemented and has been used in the stable version of the framework for several months. Definitely a huge and very important improvement

PR #371, opened 2015-03-11
Working range window

rnystrom:
This brings back the concept of a window store for nodes that are in the working range (reverting #127). It turns out that due to the system architecture if there are nodes who fetch remote content (e.g. `ASNetworkImageNode`), calls to `-display` will occur before fetching has been completed. The next chance the nodes have to decode and display content is then when they are actually on the screen, thus defeating the purpose of a working range.

With the reintroduction of the working range window, nodes are "stored" in the window and when content is finished being fetched, CA triggers `-display` since they are part of a view hierarchy.

This can be tested in the Kittens project by insuring that before `ASRangeController` adds a node to [a visible view](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASRangeController.mm#L57) that the image node (with remote content) has set its layer's contents.

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeHandlerRender.mm
+++ b/AsyncDisplayKit/Details/ASRangeHandlerRender.mm
@@ -12,64 +12,34 @@
 #import "ASDisplayNode+Subclasses.h"
 #import "ASDisplayNodeInternal.h"
 
-@interface ASDisplayNode (ASRangeHandlerRender)
-
-- (void)display;
-- (void)recursivelyDisplay;
-
-@end
-
-@implementation ASDisplayNode (ASRangeHandlerRender)
-
-- (void)display
-{
-  if (![self __shouldLoadViewOrLayer]) {
-    return;
-  }
-
-  ASDisplayNodeAssertMainThread();
-  ASDisplayNodeAssert(self.nodeLoaded, @"backing store must be loaded before calling -display");
-
-  CALayer *layer = self.layer;
-
-  // rendering a backing store requires a node be laid out
-  [layer setNeedsLayout];
-  [layer layoutIfNeeded];
-
-  if (layer.contents) {
-    return;
-  }
-
-  [layer setNeedsDisplay];
-  [layer displayIfNeeded];
-}
-
-- (void)recursivelyDisplay
-{
-  if (![self __shouldLoadViewOrLayer]) {
-    return;
-  }
-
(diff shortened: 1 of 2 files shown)


secretiverhyme:
Let's move the
```objective-c
  [node recursivelySetDisplaySuspended:NO];
```
call out of the range controller and into ASRangeHandlerRender, then land this!

rnystrom:
Also making a note from @appleguy that thrashing a `UIWindow` like this will likely incur some wasted processing on the main thread. Something we should be aware of and fix soon.

secretiverhyme:
Yup — the hierarchy manipulation thrash is specifically why we wanted to kill the rendering-window hack.  We're going to have to either convert the render range to "continuously re-display everything as needed" (matching the rendering window behaviour) or set up something like the state machine described in #315, so nodes can know whether they're supposed to render themselves once preload data arrives.

AsyncDisplayKit/Details/ASRangeHandlerRender.mm
```diff
@@ -12,64 +12,32 @@
...
-  if (![self __shouldLoadViewOrLayer]) {
-    return;
-  }
-
-  for (ASDisplayNode *node in self.subnodes) {
-    [node recursivelyDisplay];
-  }
-
-  [self display];
+  // we add nodes' views to this invisible window to start async rendering
```
secretiverhyme: just to be *absolutely* safe, can we add an `ASDisplayNodeAssertMainThread()` here?

AsyncDisplayKit/Details/ASRangeHandlerRender.mm
```diff
@@ -12,64 +12,32 @@
...
 {
   ASDisplayNodeAssertMainThread();
   ASDisplayNodeAssert(rangeType == ASLayoutRangeTypeRender, @"Render delegate should not handle other ranges");
 
-  // if node is in the working range it should not actively be in view
-  [node.view removeFromSuperview];
+  [node recursivelySetDisplaySuspended:NO];
 
-  [node recursivelyDisplay];
+  // if node is in the working range it should not actively be in view
```
secretiverhyme: update this comment?

secretiverhyme:
Two tiny nits, then we're good to merge.  =]

rnystrom:
Nits have been picked!

PR #373, opened 2015-03-12
Remove the nil-asyncDelegate proxy.

### LINKED ISSUE #349: NSInvalidArgumentException on ASCollectionView initialization

secretiverhyme:
Fixes #349.

(It's inelegant to require consumers to set an `asyncDelegate` even if it implements no methods; we could set a placeholder NSObject delegate ourselves, but in the case that an `asyncDelegate` is immediately set, we could conceivably run into race conditions in the accessibility subsystem.  This pull request is goofy, but probably the safest option for now.)

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -28,7 +28,7 @@
 @interface ASCollectionView : UICollectionView
 
 @property (nonatomic, weak) id<ASCollectionViewDataSource> asyncDataSource;
-@property (nonatomic, weak) id<ASCollectionViewDelegate> asyncDelegate;
+@property (nonatomic, weak) id<ASCollectionViewDelegate> asyncDelegate;       // must not be nil
 
 /**
  * Tuning parameters for a range.
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -66,7 +66,8 @@ static BOOL _isInterceptedSelector(SEL sel)
   if (!self) {
     return nil;
   }
-  
+
+  ASDisplayNodeAssert(target, @"target must not be nil");
   ASDisplayNodeAssert(interceptor, @"interceptor must not be nil");
   
   _target = target;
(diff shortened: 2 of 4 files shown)


rnystrom:
So the one issue I have w/ this is a possible regression of #231 as this looks like it essentially reverts #237. So w/ this we're basically requiring that you wire up a delegate *and* data source in order to display anything. Should we assert that?

secretiverhyme:
UITableView and UICollectionView already require a data source to function correctly — it wouldn't surprise me if they already have NSInternalInconsistencyExceptions or assertions requiring a non-nil data source.  The *delegate* requirement is new and breaks with UITV and UICV's behaviour; this is indeed a regression of #231, but per the discussion on #349 we don't have much choice.

PR #283, opened 2015-02-03
Add image modification block to render template images w/ tint.

andyscott:
Quick addition to render tinted images for buttons and other UI components.

PR #375, opened 2015-03-12
Stretchy or mod block images

### LINKED ISSUE #374: Fix ASImageNode's handling of stretched images vs. tint colour & modification blocks.
* Revert #354.
* Add an assertion forbidding image modification blocks on stretched images — they're mutually incompatible.
* Figure out how to support tint colour with stretched images, or assert incompatibility.

rnystrom:
Regarding tint, it actually seems like the only place we tint the image during display is [here](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASImageNode.mm#L229) so we can actually only tint greyscale. Should we assert having a tint and a stretchable image as well if its only greyscale?

fixes #374

### DIFF
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -175,11 +175,19 @@
 
   ASDisplayNodeAssert(parameters.contentsScale > 0, @"invalid contentsScale at display time");
 
+  // if the image is resizable, bail early since the image has likely already been configured
+  BOOL stretchable = !UIEdgeInsetsEqualToEdgeInsets(image.capInsets, UIEdgeInsetsZero);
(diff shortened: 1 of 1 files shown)


secretiverhyme:
Hm.  The `tint` property is of type ASImageNodeTint and can be either `.Normal` or `.Greyscale`.  There's a new image modification block, `ASImageNodeTintColorModificationBlock()` — it was added by @andyscott in #283, and an argument could be made that we should delete the `tint` property in favour of it.

I think this is actually a reasonable justification to support image modification blocks on stretchable images, beyond simply maintaining consistency.  Deleting the `tint` option would further simplify this code path.  Thoughts?

rnystrom:
This guy is ready to go!

PR #377, opened 2015-03-12
ASImageNode drawing with background color

rnystrom:


secretiverhyme:
Is this necessary?  Doesn't it duplicate CALayer's handling of the `backgroundColor` property?

Issue #380, opened 2015-03-14
deadlock (Race condition) between reloadData and ASImageView.setImage


eanagel:
I managed to track down a deadlock in the current master build (1278fe2cf08249f9ba8c02927c9da72a4fb27d58.) It is pretty easy for me to reproduce with several runs of my application. The issue appears to happen if a node property is set (in my case `ASImageView`'s `setImage`) while the sizing queue is also measuring the same cell. I would expect the mutex to allow these operations to happen without negative interaction, but in my case we are ending up with a deadlock somehow. When the app has hung, I can break at anytime and I'll find the stacks traces below. I think these are the offending scenarios.

Here is the first stack trace, on the main thread:

	#2	0x0000000104ac5685 in ASDN::Mutex::lock() at .../Headers/Build/AsyncDisplayKit/ASThread.h:124
	#3	0x0000000104ac5653 in ASDN::Locker<ASDN::Mutex>::Locker(ASDN::Mutex&) at .../Headers/Build/AsyncDisplayKit/ASThread.h:69
	#4	0x0000000104ac54ad in ASDN::Locker<ASDN::Mutex>::Locker(ASDN::Mutex&) at .../Headers/Build/AsyncDisplayKit/ASThread.h:70
	#5	0x0000000104ab2b7c in -[ASDisplayNode shouldRasterizeDescendants] at .../AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:520
	#6	0x0000000104aafbc9 in -[ASDisplayNode __rasterizedContainerNode] at .../AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:249
	#7	0x0000000104a9dd3b in -[ASDisplayNode(UIViewBridge) setNeedsDisplay] at .../AsyncDisplayKit/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm:161
	#8	0x0000000104adb7b9 in __24-[ASImageNode setImage:]_block_invoke at .../AsyncDisplayKit/AsyncDisplayKit/ASImageNode.mm:116
	#9	0x0000000104aad9e0 in ASDisplayNodePerformBlockOnMainThread(void () block_pointer) at .../AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:76
	#10	0x0000000104adb6d5 in -[ASImageNode setImage:] at .../AsyncDisplayKit/AsyncDisplayKit/ASImageNode.mm:114

[...]

Issue #57, opened 2014-10-16
Incorrect display node thread affinity


l4u:
'Incorrect display node thread affinity' when ASTableView is used with a UIPageViewController

https://github.com/l4u/AsyncDisplayKit_ASTableView_bug_with_UIPageViewController

To reproduce, swipe for 1 or more pages.

phatmann:
@megakode, if you can confirm that this issue is the cause of your crash, we should either re-open this issue or you should file a new one.

secretiverhyme:
We haven't seen any reports of this particular crash recently, but please do file a new issue if they recur.  The thread-affinity API is something we've been wanting to remove for a while, and it's on our radar for AsyncDisplayKit 1.2 — subscribe to #134 if you'd like to follow along.

PR #389, opened 2015-03-20
Fix ASImageNode -measure: / -setImage: deadlock.

secretiverhyme:
Speculative fix for #380.  Calling `-[ASImageNode setImage:]` on a
background queue will take the image lock, then trampoline to the main
queue to invalidate calculatedSize and set needsDisplay without
releasing the lock.  Any other method call that takes the image lock on
the main queue can block on the trampoline completing -- deadlock.

Note that the trampoline is itself a clowny workaround for thread
affinity (#134) and should be removed.  With this patch applied,
ASImageNode's `-setImage:` and `-calculateSizeThatFits:` can still race:

1.  -setImage changes _image, then trampolines to main
2.  -calculateSizeThatFits measures _image
3.  trampoline completes, invalidates calculated size
4.  ???
5.  ~~profit!~~ :boom:

### DIFF
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -108,6 +108,8 @@
   ASDN::MutexLocker l(_imageLock);
   if (_image != image) {
     _image = image;
+
+    ASDN::MutexUnlocker u(_imageLock);
     ASDisplayNodePerformBlockOnMainThread(^{
       [self invalidateCalculatedSize];
       [self setNeedsDisplay];


secretiverhyme:
For correctness' sake, we should really just revert 72f9cb9d731e8977b3b20e39ee98ef04fc50b798 (#57), but at the time the decision was made to favour ease-of-use over handling edge cases like the above race condition.  As such, this is probably the best we can do until #134 is solved.

rnystrom:
I think merging one of the other PRs broke this. Quick rebase and I'll merge!

PR #407, opened 2015-03-30
Fix handling of pending nodes and placeholder layer

ianolito:
Some fixes to the placeholder API:
* The ```_pendingDisplayNodes``` object was never initialized.
* Placeholder opacity was not reset after being faded out once.
* Small documentation typo.


appleguy:
Sorry for the delay!  @rnystrom, do you know this code best?  If not I'll take a look as soon as I can.

appleguy:
Thanks @ianolito, I'm going to go ahead and take this.  It certainly would be better to lazily create the set because this will add a relevant amount of overhead to leaf nodes that won't display any subnodes — particularly layer-backed nodes that are otherwise quite cheap and may be numerous.  If you get around to another diff to address that, awesome :).

PR #418, opened 2015-04-03
Add new kittens after last row

spicyj:
New kittens were being added before the last kitten instead of after it
because we weren't taking into account the first (header) row.


rnystrom:
Nice catch, thanks!

PR #425, opened 2015-04-14
change ASMutableAttributedStringBuilderTests so that it is safe against ...

onhachoe:
...different MAX macro

when MAX macro is defined in such a way in "sys/params.h", this test will become flaky,
fix so that it doesn't happen


appleguy:
Interesting find...I'm still not totally clear on what MAX definition you found that this fails for, but believe you that it is possible; thanks for submitting the diff and sorry for the delay!

PR #427, opened 2015-04-16
Bug Fix - ASTableView sometimes fails to render cell node when scrolling programmatically.

eanagel:
Root cause is because we assume `indexPathsForVisibleRows` would return the new item to be displayed immediately after `tableView:willDisplayCell:forRowAtIndexPath:` is called. This is not **always** true, especially when we are scrolling programmatically. Fix is to maintain our own list of visible index paths. Most of the code here is actually for debugging (see `DEBUG_VISIBLE_INDEX_PATHS`) and can be removed once we are confident this works in all use cases. I do wonder if there are edge cases where `tableView:didEndDisplayingCell:forRowAtIndexPath:` is not called (such as reloading), but I haven't seen this so far.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -16,6 +16,7 @@
 #import "ASDisplayNodeInternal.h"
 #import "ASBatchFetching.h"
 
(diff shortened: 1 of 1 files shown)


eanagel:
Posting a better solution shortly

PR #430, opened 2015-04-17
Bug Fix - ASTableView sometimes fails to render cell contents when scrolling programmatically.

eanagel:
Root cause is because we assume `UITableView`'s `indexPathsForVisibleRows` returns the new item to be displayed immediately after `tableView:willDisplayCell:forRowAtIndexPath:` is called. This is not always true, especially when we are scrolling programmatically. Fix is to track this "pending" visible index path and return it along with the `UITableView`'s `indexPathsForVisibleRows` in `rangeControllerVisibleNodeIndexPaths:`.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -118,6 +118,8 @@ static BOOL _isInterceptedSelector(SEL sel)
   BOOL _asyncDataFetchingEnabled;
 
   ASBatchContext *_batchContext;
+
+  NSIndexPath *_pendingVisibleIndexPath;
 }
 
 @property (atomic, assign) BOOL asyncDataSourceLocked;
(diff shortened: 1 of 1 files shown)


eanagel:
Note this is an improved version of an earlier pull request I have closed.

eanagel:
I see the tests failed, but it doesn't seem to have anything to do with my PR, maybe someone has a change in flight?
Failure:

         /Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKitTests/ASBasicImageDownloaderContextTests.m:53: (([context isCancelled]) is true) failed - Context should be cancelled

appleguy:
Hmm, this is very interesting.  Thank you for debugging issues like this, as they are quite tricky and thus the fixes are particularly valuable.

1) Are you confident that it is sufficient to have a single index path tracked in this manner, and a set is not required?
2) What code path uses the visible index paths method which ensures display is successful when this one is "injected"?

appleguy:
Pretty frustrated that there doesn't seem to be a better way to ping Travis (at least, the mythical reload button that some show on the interwebs remains elusive...)

eanagel:
I originally had a more complex version of this PR that attempted to track visible cells based on calls to `willDisplayCell:...` and `didEndDisplayingCell...' and then merging that with the list from 'UITableView' (just to be safe.) Turns out this was actually caused problems because didEndDisplayingCell is not always called (not on reload I believe -- for starters.) As I began to investigate I realized the edge case is just when the cell about to be displayed hasn't hit the screen quite when our code is running, thus this fix seems to work well. I haven't seen any problems since adding this fix in our app. If you want to take a look at that solution, check out #427

appleguy:
Thanks for the details!  I'll open a task for deeper investigation, but certainly believe that this is strictly better than before.

Issue #394, opened 2015-03-21
Support exclusion paths on ASTextNode.


tberman:
Without this, it is hard to use it to replace all uses of UITextView, where the NSTextContainer/etc are exposed and usable.

secretiverhyme:
ASTextNode abstracts away its Text Kit components because it relies on Text Kit — for example, client code is not permitted to set exclusion paths, because the text node uses them internally for truncation support.  We can certainly expand ASTextNode's APIs, though.  What use cases did you have in mind?

tberman:
I am using exclusion paths to create empty areas in a text layout and then putting other UIView/ASDisplayNode subclasses into those empty areas.

I could instead use multiple ASTextNodes, which works (but isn't fun to maintain) in some of the cases (when the empty area spans the width of the device), but is pretty awful in others.

appleguy:
@tberman Thanks for filing this task!  We just merged @tonklon's diff to address this.  Wish we got to it sooner, but I expect it will be useful to others with advanced text needs as well.

PR #432, opened 2015-04-20
ASTextNode expose exclusion paths

tonklon:
expose NSTextContainer's exclusionPaths property on ASTextNode to be able to exclude view areas from typesetting.

This was already discussed in #394.

I do not see were exclusion paths are used internally. A comment on `[ASTextNodeRenderer _calculateCharacterIndexBeforeTruncationMessage]` claims that, but I don't see that anywhere in code. Perhaps the comment is outdated?


appleguy:
@tonklon thanks for both the implementation and the tests!  My apologies for the delayed merge; should be more prompt from here out.

@ocrickard, do your improvements to the text stack in CK happen to include exclusionPaths?  If not, perhaps you can apply this there to ensure we don't lose the functionality if the code is shared.

PR #435, opened 2015-04-22
Add waterfall layout into ASCollectionView

xujim:
Implemented a waterfall layout for ASCollectionView, which support three type item append mode. 
However, it has following limitation:
1, The shortest item append mode doesn't support insert node operation
2, Not incorporate into existing ASCollectionView due to its initWithFrame function is hardcoded to support Flowlayout only.

If this changes was approved, we need to extend ASCollectionView to support more type of layout.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -171,6 +171,8 @@
 		D785F6621A74327E00291744 /* ASScrollNode.h in Headers */ = {isa = PBXBuildFile; fileRef = D785F6601A74327E00291744 /* ASScrollNode.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		D785F6631A74327E00291744 /* ASScrollNode.m in Sources */ = {isa = PBXBuildFile; fileRef = D785F6611A74327E00291744 /* ASScrollNode.m */; };
 		DB7121BCD50849C498C886FB /* libPods-AsyncDisplayKitTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = EFA731F0396842FF8AB635EE /* libPods-AsyncDisplayKitTests.a */; };
+		FA9EAC861ADFAF50005FC4D5 /* ASWaterfallLayoutController.h in Headers */ = {isa = PBXBuildFile; fileRef = FA9EAC841ADFAF50005FC4D5 /* ASWaterfallLayoutController.h */; };
+		FA9EAC871ADFAF50005FC4D5 /* ASWaterfallLayoutController.mm in Sources */ = {isa = PBXBuildFile; fileRef = FA9EAC851ADFAF50005FC4D5 /* ASWaterfallLayoutController.mm */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
@@ -347,6 +349,8 @@
 		D785F6601A74327E00291744 /* ASScrollNode.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ASScrollNode.h; sourceTree = "<group>"; };
 		D785F6611A74327E00291744 /* ASScrollNode.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = ASScrollNode.m; sourceTree = "<group>"; };
 		EFA731F0396842FF8AB635EE /* libPods-AsyncDisplayKitTests.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = "libPods-AsyncDisplayKitTests.a"; sourceTree = BUILT_PRODUCTS_DIR; };
+		FA9EAC841ADFAF50005FC4D5 /* ASWaterfallLayoutController.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ASWaterfallLayoutController.h; sourceTree = "<group>"; };
+		FA9EAC851ADFAF50005FC4D5 /* ASWaterfallLayoutController.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; path = ASWaterfallLayoutController.mm; sourceTree = "<group>"; };
 		FB07EABBCF28656C6297BC2D /* Pods-AsyncDisplayKitTests.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-AsyncDisplayKitTests.debug.xcconfig"; path = "Pods/Target Support Files/Pods-AsyncDisplayKitTests/Pods-AsyncDisplayKitTests.debug.xcconfig"; sourceTree = "<group>"; };
 /* End PBXFileReference section */
 
@@ -528,6 +532,8 @@
 		058D09E1195D050800B7D73C /* Details */ = {
 			isa = PBXGroup;
 			children = (
+				FA9EAC841ADFAF50005FC4D5 /* ASWaterfallLayoutController.h */,
+				FA9EAC851ADFAF50005FC4D5 /* ASWaterfallLayoutController.mm */,
 				058D09E2195D050800B7D73C /* _ASDisplayLayer.h */,
 				058D09E3195D050800B7D73C /* _ASDisplayLayer.mm */,
 				058D09E4195D050800B7D73C /* _ASDisplayView.h */,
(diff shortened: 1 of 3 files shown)


appleguy:
@xujim wow, this looks really cool — and your code is quite clean!  We're very interested in expanding layout support in ASDK, and if we can figure out the right way to integrate this, it could be a great addition.  I think most of the work is done, just the last 10% around integration.

Could you share some screenshots and/or videos of this working?  Do you have a test project?

appleguy:
@xujim thanks again for this.  We have some exciting new improvements to ASCollectionView Layout support that I think will make this even better.

The framework should now support using any UICollectionViewLayout, and is no longer limited to only Flow layouts.  I think that should mean that this class no longer needs to conform to the ASLayoutController protocol.

Importantly, I believe the way this was originally implemented was not thread safe.  Because it derives from UICollectionViewLayout, it would mean that the ASLayoutController reference created by ASRangeController would be created and used off the main thread.  This may work, but is not necessarily safe to do since UIKit makes no guarantees about supporting this kind of usage.

I am actually supportive of including a couple UICollectionViewLayout classes with ASDK because it could help developers implement what they need even more easily, although this could technically be a separate github project.  If you can take a pass at cleaning it up to integrate with master / intended for use as a UICollectionViewLayout instead of an ASLayoutController, I think we could more easily evaluate how it should be merged.

appleguy:
@xujim, check out the master branch for these changes.  Namely, there is now ASCollectionViewLayoutController, which allows usage of any UICollectionViewLayout with ASCollectionView (no longer flow layouts only).

In thinking about this, it is possible that your implementation is more efficient than it could otherwise be if contained in a UICollectionViewLayout because it has a more specifically targeted API for usage with ASDK.  That said, I think it's important to better understand that impact; it may be very small, in which case we should prefer providing a UICollectionViewLayout implementation.

Let me know if you plan to spend any further time here.  If you'll be unavailable for a while, we can let this sit or save the code for follow-up in a Github Issue.

erema:
Hello, @appleguy , so, how can I implement a waterfall layout using a current version? I've tried to implement own class of UICollectionViewLayout and pass it to ASCollectionView, but nothing happened. What do I wrong?

appleguy:
@erema — happy to help.  Could you clarify what you mean by "nothing happened"?

Are you getting delegate callbacks, including nodeForRowAtIndexPath:?  Be sure you've set asyncDelegate and asyncDataSource, which supersede / replace the UIKit ones (delegate / dataSource).

erema:
@appleguy thank you for the fast answer. I've tried to reproduce this http://www.raywenderlich.com/107439/uicollectionview-custom-layout-tutorial-pinterest . When I wrote own UICollectionViewLayout and pass it to ASCollectionView I saw that its methods are called like 
- (void)prepareLayout
- (CGSize)collectionViewContentSize
- (NSArray<UICollectionViewLayoutAttributes *> *)layoutAttributesForElementsInRect:(CGRect)rect

but grid wasn't change, this is what I meant.

appleguy:
Hmm, very odd...let's get it fixed!

- What init method are you using for ASCollectionView?
- What version of AsyncDisplayKit are you using?  Please test using the latest master build — this is important, there has been a great deal of divergence since 1.2.3 (master is close to being released as a final 2.0, and is quite stable right now with many improvements).
- Can you share any of the code, even just in snippets or perhaps the text file (.m) of relevant implementation?
--- Can you upload a test project?  This would allow us to fix with certainty, and to do so the most quickly, but I totally understand if you can't do this (I'll still try to help debug).

Note that Pinterest is currently using a UICollectionView with custom layout, and we have a branch of the app that works with ASCollectionView using an identical, unmodified custom UICollectionViewLayout.  It's important to ensure the correct init method & delegate setup is done.

erema:
1. - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout asyncDataFetching:(BOOL)asyncDataFetchingEnabled;
2. I installed AsyncDisplayKit via pod 3 weeks ago
3. 

```objective-c
- (void)viewDidLoad {
    [super viewDidLoad];
UIThumbsLayout *layout = [[UIThumbsLayout alloc] init];
    
    _collectionView = [[ASCollectionView alloc] initWithFrame:CGRectZero collectionViewLayout:layout asyncDataFetching:YES];
    _collectionView.asyncDataSource = self;
    _collectionView.asyncDelegate = self;
    _collectionView.backgroundColor = [UIColor clearColor];
    _collectionView.contentInset = UIEdgeInsetsMake(STATUSBAR_HEIGHT + NAVIGATIONBAR_HEIGHT, 0, 0, 0);
[self.view addSubview:_collectionView];
}
- (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtIndexPath:(NSIndexPath *)indexPath
{
    FCThumbCell *node = [[FCThumbCell alloc] init];
    node.backgroundColor = [UIColor greenColor];
    node.frame = CGRectMake(0, 0, self.view.frame.size.width, self.view.frame.size.height);
    
    return node;
}

- (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSection:(NSInteger)section
{
    return content.count;
}

- (BOOL)collectionView:(UICollectionView *)collectionView shouldSelectItemAtIndexPath:(NSIndexPath *)indexPath
{
    return YES;
}

@interface UIThumbsLayout()
{
    int numberOfColumns;
    CGFloat cellPadding;
    
    NSMutableArray *cache;
    
    CGFloat contentHeight;
    CGFloat contentWidth;
}

@property (nonatomic, strong) NSArray *cellSizes;


@end

@implementation UIThumbsLayout

- (id)init
{
    if (!(self = [super init]))
        return nil;
    
    numberOfColumns = 2;
    cellPadding = 6.0;
    contentHeight = 100.0;
    
    
    cache = [[NSMutableArray alloc] init];
    
    return self;
}
- (void)prepareLayout
{
    [super prepareLayout];
    
    contentWidth = CGRectGetWidth(self.collectionView.bounds) - (self.collectionView.contentInset.left + self.collectionView.contentInset.right);
    
    if (!cache.count)
    {
        CGFloat columnWidth = contentWidth / numberOfColumns;
        
        NSMutableArray *xOffset = [[NSMutableArray alloc] init];
        for (int i = 0; i < numberOfColumns; i++)
            [xOffset addObject:[NSNumber numberWithFloat:i*columnWidth]];
        
        NSMutableArray *yOffset = [[NSMutableArray alloc] init];
        for (int i = 0; i < numberOfColumns; i++)
            [yOffset addObject:[NSNumber numberWithFloat:0]];
        
        NSInteger numberOfSections = [self.collectionView numberOfSections];
        if (numberOfSections == 0) {
            return;
        }
        
        NSInteger itemCount = [self.collectionView numberOfItemsInSection:0];
        
        int column = 0;
        for (int i = 0; i < itemCount; i++)
        {
            CGFloat width = columnWidth - cellPadding * 2;
            
            CGFloat height = cellPadding + cellPadding;
            CGRect frame = CGRectMake([[xOffset objectAtIndex:column] floatValue], [[yOffset objectAtIndex:column] floatValue], columnWidth, height);
            CGRect insetFrame = CGRectInset(frame, cellPadding, cellPadding);
            
            UICollectionViewLayoutAttributes *attr = [UICollectionViewLayoutAttributes layoutAttributesForCellWithIndexPath:[NSIndexPath indexPathForRow:i inSection:0]];
            attr.frame = insetFrame;
            [cache addObject:attr];
            
            contentHeight = MAX(contentHeight, CGRectGetMaxY(frame));
            [yOffset addObject:[NSNumber numberWithFloat:[[yOffset objectAtIndex:i] floatValue] + height]];
            
            column = column >= (numberOfColumns - 1) ? 0 : ++column;
        }
    }
}

- (CGSize)collectionViewContentSize
{
    return CGSizeMake(contentWidth, contentHeight);
}

- (NSArray<UICollectionViewLayoutAttributes *> *)layoutAttributesForElementsInRect:(CGRect)rect
{
    NSMutableArray *layoutAttributes = [[NSMutableArray alloc] init];
    for (UICollectionViewLayoutAttributes *attributes in cache)
    {
        if (CGRectIntersectsRect(attributes.frame, rect))
        {
            [layoutAttributes addObject:attributes];
        }

    }
    return layoutAttributes;
}
@end
```

appleguy:
Will take a close look at this tomorrow morning!

Installing via the pod, if you made no changes to the configuration, means you would be on 1.2.3 (and a likely explanation for a variety of issues — not sure if it could cause this particular one, but there are definitely some improvements to how custom UICV layouts are handled in the master version).  Check out the pods documentation if you haven’t needed to manually point to a project master branch before (I don’t remember the syntax offhand :( )

		— Scott

erema:
I resolved issue, I should sync cell's frame in UIThumbsLayout class and my Cell class. But now I have a new bug(crash), when I add new items into collection view - it crashes!

erema:
Any ideas?

appleguy:
@erema glad you found a resolution!  The crash you indicated shows entirely UIKit APIs / errors, and I'm not sure it is related to ASDK (or at least, I can't from the information provided see how it would be).

One quick thing to try would be to switch to UICollectionView (instead of ASCollectionView), just as a test.  You would have to implement the cellForItemAtIndexPath: and possibly the size method as well, just like a normal UICollectionView, instead of nodeForRowAtIndexPath which replaces a couple of them.  I have a hunch that you will hit the same crash.

If you want to share the project or more of the code, we might be able to debug it!  Let me know if you get to a solution or have any additional questions, happy to help however I can.

appleguy:
@erema did you reach a resolution?  Please do file a GitHub issue if you encounter any problems with ASDK, and I would love to help in more detail.  It's been busy lately, so my apologies if my responses didn't solve your issue — I'll jump on the thread if you make an issue.

@xujim Thanks again for submitting this PR, but I'm going to close it out since we haven't heard from you in quite a while.  I think it would be interesting to include some more first-class collection view layouts with ASDK, but this would require some further development / iteration to do.

PR #409, opened 2015-04-01
Ensure all nodes are dealloc'd in ASDataController teardown

barfoon:
This is a proposed fix for https://github.com/facebook/AsyncDisplayKit/issues/406.

This ensures that all node instances used with an instance of `ASCollectionView` are dealloc'd when the dataController is dealloc'd.

<hr>

Memory consumption using an `ASCollectionView` with `ASCellNode`

**Before**

image

**After**

image

<hr>

**Without changeset applied:**

**Node alloc's**
```
[MyNode:65] node init | 0x15fd49530
[MyNode:65] node init | 0x15fe649d0
[MyNode:65] node init | 0x15fd4a470
[MyNode:65] node init | 0x15fd4dd50
[MyNode:65] node init | 0x15fe6af90
[MyNode:65] node init | 0x15fe6e910
[MyNode:65] node init | 0x15fd507f0
[MyNode:65] node init | 0x15fd52ce0
[MyNode:65] node init | 0x15fd55f40
[MyNode:65] node init | 0x15fe73ae0
[MyNode:65] node init | 0x15fe75a80
[MyNode:65] node init | 0x15fd59aa0
[MyNode:65] node init | 0x15fe7ae90
[MyNode:65] node init | 0x15fe7d0d0
[MyNode:65] node init | 0x15fe80980
[MyNode:65] node init | 0x15fe83a50
```

**Controller dealloc**
```
[MyCollectionViewController:336] dealloc | controller dealloc'd!
```

**Node dealloc's**
```
[MyNode:96] dealloc | 0x15fd52ce0
[MyNode:96] dealloc | 0x15fd507f0
[MyNode:96] dealloc | 0x15fe6e910
[MyNode:96] dealloc | 0x15fe6af90
[MyNode:96] dealloc | 0x15fd4dd50
[MyNode:96] dealloc | 0x15fd4a470
[MyNode:96] dealloc | 0x15fe649d0
[MyNode:96] dealloc | 0x15fd49530
```

<hr>

**With changeset applied:**

**Node alloc's**
```
[MyNode:65] node init | 0x13e548460
[MyNode:65] node init | 0x13e548910
[MyNode:65] node init | 0x13e54b9e0
[MyNode:65] node init | 0x13e54ea40
[MyNode:65] node init | 0x13e552500
[MyNode:65] node init | 0x13e66aff0
[MyNode:65] node init | 0x13e66e080
[MyNode:65] node init | 0x13e670470
[MyNode:65] node init | 0x13e673610
[...]

PR #437, opened 2015-04-22
ASDataController dealloc, ensure cell nodes are not in a superview

eanagel:
For reasons I don't fully understand, this avoids a retain cycle with the currently displayed cells.

This is an alternate implementation of "[Ensure all nodes are dealloc'd in ASDataController teardown #409](https://github.com/facebook/AsyncDisplayKit/pull/409)" that works with multiple sections and IMO should supersede that one. (Thanks @barfoon for working this out in the first place. I was quite mystified until I saw his PR.)

In our application, this fix reduced AsyncDisplayKit's retained memory significantly. (There are still other issues I haven't tracked down, but this one is the biggest by far.)

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -514,8 +514,12 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
   ASDisplayNodeAssertMainThread();
   [_nodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
     [section enumerateObjectsUsingBlock:^(ASCellNode *node, NSUInteger rowIndex, BOOL *stop) {
-      if (node.isNodeLoaded && node.view.superview) {
-        [node.view removeFromSuperview];
+      if (node.isNodeLoaded) {
+        if (node.layerBacked) {
+          [node.layer removeFromSuperlayer];
+        } else {
+          [node.view removeFromSuperview];
+        }
       }
     }];
   }];
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -508,4 +508,17 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
   return ASFindElementsInMultidimensionalArrayAtIndexPaths(_nodes, [indexPaths sortedArrayUsingSelector:@selector(compare:)]);
 }
 
+#pragma mark - Dealloc
+
+- (void)dealloc {
+  ASDisplayNodeAssertMainThread();
+  [_nodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
+    [section enumerateObjectsUsingBlock:^(ASCellNode *node, NSUInteger rowIndex, BOOL *stop) {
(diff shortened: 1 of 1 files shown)


appleguy:
@eanagel Thanks a bunch for investigating this.  We have some known issues here, but unfortunately have been very slow in addressing them.  My preferred solution will be to completely rewrite ASDataController, as it is unnecessarily complicated and incorrectly handles certain edge cases; because of this we have hesitated to invest in modifying the current version.  Despite that, though, it will take sufficiently long to rewrite that having a nearer term fix is tremendously valuable — it will reduce time pressure on the replacement and allow us to ensure it is much better tested, and has a better API (the data lock should be eliminated by simply asserting data modification happens on main, which should be the only sane architecture for most apps anyway; layout, rendering, etc will still be automatically asynchronous)

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -508,4 +508,17 @@ - (NSArray *)nodesAtIndexPaths:(NSArray *)indexPaths
   return ASFindElementsInMultidimensionalArrayAtIndexPaths(_nodes, [indexPaths sortedArrayUsingSelector:@selector(compare:)]);
 }
 
+#pragma mark - Dealloc
+
+- (void)dealloc {
+  ASDisplayNodeAssertMainThread();
+  [_nodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
+    [section enumerateObjectsUsingBlock:^(ASCellNode *node, NSUInteger rowIndex, BOOL *stop) {
+      if (node.isNodeLoaded && node.view.superview) {
```
appleguy: Definitely correct form to check .isNodeLoaded here (I'm sure this is obvious to you, but it's nice to see a well considered patch :) )

I would remove the second part of the condition, as it performs two method calls.  Additionally, it's not compatible with layer-backing.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -508,4 +508,17 @@ - (NSArray *)nodesAtIndexPaths:(NSArray *)indexPaths
...
+        [node.view removeFromSuperview];
```
appleguy: This needs to be modified a little bit to handle layer-backed nodes.  I'm pretty sure an ASCellNode itself can be layer-backed.

We could provide a nicer API to hide this, but I think for now you need:

if (node.isLayerBacked)
   [node.layer removeFromSuperlayer];
else
   [node.view removeFromSuperview];

appleguy:
If you make the fixes I described, and have tested this in a focused way, I'm happy to merge this ASAP!

eanagel:
I didn't bother with supporting layers because it appears `ASRangeController` isn't into layers (see https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASCellNode.m#L47), but it probably makes sense to code this for the future. I'll update the PR in the next few minutes here.

appleguy:
Pinging Travis CI to see if we can get a green build.

appleguy:
Winning, thanks!

ERussel:
I can confirm that the issue is still present. For example, when i call reloadData for collectionView, invisible nodes would not be deallocated. As i understand from ASRangeHandlerRender code, invisible nodes are added to global working window to activate offscreen rendering, but i don't see any workaround to remove them from superview in DELETE_NODES call.

PR #439, opened 2015-04-28
Add option to pass through touches to non-link text in ASTextNode

eanagel:
Adds a new property, `passthroughNonlinkTouches` that will allow touches in non-link areas to pass through to other event handlers event with `userInateractionEnabled` for an `ASTextNode`.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -749,6 +749,33 @@ ASDISPLAYNODE_INLINE CGFloat ceilPixelValue(CGFloat f)
 
 #pragma mark - Touch Handling
 
+-(BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event
+{
+  if (!_passthroughNonlinkTouches) {
+    return [super pointInside:point withEvent:event];
+  }
+
+  NSRange range = NSMakeRange(0, 0);
+  NSString *linkAttributeName = nil;
+  BOOL inAdditionalTruncationMessage = NO;
+
+  id linkAttributeValue = [self _linkAttributeValueAtPoint:point
+                                             attributeName:&linkAttributeName
+                                                     range:&range
+                             inAdditionalTruncationMessage:&inAdditionalTruncationMessage];
+
+  NSUInteger lastCharIndex = NSIntegerMax;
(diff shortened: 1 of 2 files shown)


appleguy:
Ha, this is a really cool idea!  I want to think a little about the right way to test this, but will merge soon.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -749,6 +749,33 @@ - (UIImage *)placeholderImage
...
+  NSRange range = NSMakeRange(0, 0);
+  NSString *linkAttributeName = nil;
+  BOOL inAdditionalTruncationMessage = NO;
+
+  id linkAttributeValue = [self _linkAttributeValueAtPoint:point
+                                             attributeName:&linkAttributeName
+                                                     range:&range
+                             inAdditionalTruncationMessage:&inAdditionalTruncationMessage];
+
+  NSUInteger lastCharIndex = NSIntegerMax;
```
appleguy: Hmm, this doesn't seem right, does it?  Not only does the logic look odd using this value, why use the local variable at all if it is just NSIntegerMax?

appleguy:
@eanagel I'm OK with this being manually tested for now; although I'd like to see something like this have more sophisticated testing, I don't want to block the commit and am happy to merge it.

However, check out my comment on the diff and let me know what you think.  If it is /not/ a bug, perhaps add a comment because I think it's a bit unclear.

AsyncDisplayKit/ASTextNode.mm
lxcid: I think on line 758, I recommending using `NSMakeRange(NSNotFound, 0)` instead which better describe a `null` range?

lxcid:
This is awesome. I have a use case for this, particularly when adding a text node in cell node.

AsyncDisplayKit/ASTextNode.mm
eanagel: Yeah that does look a little suspicious. This is a straight copy and paste from `touchesBegan:withEvent:`, so I didn't really look into how it worked too much. Suggest we merge this PR then fix in both places as a separate PR? (Goal would be to clean-up and merge the logic into one place.)

appleguy:
@eanagel At your request, I'll merge this but certainly appreciate any follow-up you can provide.  Thanks for your patience, just settled in after returning from vacation :).

appleguy:
https://github.com/facebook/AsyncDisplayKit/issues/473

PR #440, opened 2015-05-01
Don’t retain an action’s target in ASControlNode

rlasante:
- addTarget:action:forControlEvents: should not retain the target. If it does
then there is a very high likelihood of a retain cycle.


abaskov:
We've got confirmed memory leaks because of this bug, extremely needing this pull request to be merged into master.

appleguy:
Ouch, thank you guys for debugging and fixing this.  This particular code is quite new and was not deployed in Paper, AKA not nearly as well tested as the overwhelming majority of ASDK.

PR #441, opened 2015-05-01
Changed the references folder name

nscoding:
Following my PR on [ios-snapshot-test-case](https://github.com/facebook/ios-snapshot-test-case/pull/77) I changed the default directory to end with _32 if it's running on 32bits. Updating the folder name to make the tests work again.

Test plan: Run ASImageNodeSnapshotTests


nscoding:
I will modify ios-snapshot-test-case to fallback to `ReferenceImages`

PR #443, opened 2015-05-04
Allow ASTableView to be loaded in a xib/storyboard

eanagel:
This code moves the initialization code to a shared method and uses it to initialize correctly in `-initWithCoder:` so it can be used in a xib or  storyboard. Note that when initializing this way `asyncDataFetching` is defaulted to NO, so it's not possible to use asyncDataFetching in this configuration. (We could do this by making `asyncDataFetching` an `IBInspectable` property but that was a bit more refactoring than I'm comfortable with at the moment.)

It might be worth double-checking `ASCollectionView` to see if it could benefit form a similar refactor. (Next time I'm in that code I'll take a look if someone else doesn't.)

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -129,17 +129,7 @@ static BOOL _isInterceptedSelector(SEL sel)
 #pragma mark -
 #pragma mark Lifecycle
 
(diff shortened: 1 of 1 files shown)


appleguy:
Pinging flaky Travis CI

appleguy:
Cool, thanks for matching the initializer style too :).  Will definitely accept any further diffs for IB compatibility if you do this in ASCollectionView, or find another spot where it is applicable.

PR #444, opened 2015-05-07
Fix typo in ASTableView

levi:


appleguy:
Ouch — thank you!  I just started working in these files and noticed a lot of cleanup like this to do.  Appreciate you submitting this one before I could get around to it in the next few weeks.

PR #445, opened 2015-05-08
Prevent UITableView insertion/deletion animations by default

### LINKED ISSUE #292: [ASTableView] Calling reloadData causing animated flickers

levi:
Calling `[super insertRowsAtIndexPaths:withRowAnimation:]` and its `delete`/`sections` counterparts with `UITableViewRowAnimationNone` seems to do nothing to prevent the cascading animations when data changes in the table view. Using other options like `UITableViewRowAnimationFade` and `UITableViewRowAnimationRight` change the animation completely. With the `UITableViewRowAnimationNone` option not functioning, I have turned off animations during the default insertion/deletion calls.

This fixes #292.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -126,6 +126,16 @@ static BOOL _isInterceptedSelector(SEL sel)
 
 @implementation ASTableView
 
+void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
+  if (withoutAnimation) {
+    [UIView setAnimationsEnabled:NO];
+    block();
+    [UIView setAnimationsEnabled:YES];
+  } else {
+    block();
+  }
+}
+
 #pragma mark -
 #pragma mark Lifecycle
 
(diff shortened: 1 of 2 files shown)


levi:
I have already signed the CLA

AsyncDisplayKit/ASTableView.mm
```diff
@@ -126,6 +126,16 @@ @interface ASTableView () <ASRangeControllerDelegate, ASDataControllerSource> {
 
 @implementation ASTableView
 
+void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
+  if (withoutAnimation) {
+    [UIView setAnimationsEnabled:NO];
```
appleguy: For guaranteed correctness, whenever using this pattern you must:
BOOL animationsEnabled = [UIView areAnimationsEnabled];
...
[UIView setAnimationsEnabled:animationsEnabled];

AsyncDisplayKit/ASTableView.mm
```diff
@@ -467,29 +477,36 @@ - (CGSize)rangeControllerViewportSize:(ASRangeController *)rangeController
...
+    [super insertRowsAtIndexPaths:indexPaths withRowAnimation:(UITableViewRowAnimation)animationOption];
+  });
 }
 
 - (void)rangeController:(ASRangeController *)rangeController didDeleteNodesAtIndexPaths:(NSArray *)indexPaths withAnimationOption:(ASDataControllerAnimationOptions)animationOption
 {
   ASDisplayNodeAssertMainThread();
 
-  [super deleteRowsAtIndexPaths:indexPaths withRowAnimation:(UITableViewRowAnimation)animationOption];
+  ASPerformBlockWithoutAnimation(animationOption == UITableViewRowAnimationNone, ^{
```
appleguy: Personally I read this fine, but for extra clarity I'd suggest either adding extra parentheses (make it more clear that this is evaluating an expression inside an arguments list) or create a local variable like "preventAnimation ="

AsyncDisplayKit/ASTableView.mm
```diff
@@ -126,6 +126,16 @@ @interface ASTableView () <ASRangeControllerDelegate, ASDataControllerSource> {
 
 @implementation ASTableView
 
+void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
```
appleguy: Add a comment above this that describes the UIKit issue being worked around and/or references this PR.

appleguy:
I noticed this recently while doing some testing, but didn't investigate it because the animation wasn't a concern at the time.  Very reassuring to see this submitted already — thank you!

It bothers me that this is happening in the first place.  Presumably the latest iOS versions don't do this with plain UITableView, or is it also blatantly broken?

Make a couple small tweaks to this and I'll get it merged right away. :)

levi:
@appleguy thanks for the feedback. I updated the pull request to reflect your changes.

appleguy:
Fantastic, thank you @levi!

PR #449, opened 2015-05-11
Added iOS framework target for Carthage support

nvh:
This PR adds a separate framework target so AsyncDisplayKit can be used with [Carthage](https://github.com/Carthage/Carthage)

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -168,6 +168,117 @@
 		AC3C4A511A1139C100143C57 /* ASCollectionView.h in Headers */ = {isa = PBXBuildFile; fileRef = AC3C4A4F1A1139C100143C57 /* ASCollectionView.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		AC3C4A521A1139C100143C57 /* ASCollectionView.mm in Sources */ = {isa = PBXBuildFile; fileRef = AC3C4A501A1139C100143C57 /* ASCollectionView.mm */; };
 		AC3C4A541A113EEC00143C57 /* ASCollectionViewProtocols.h in Headers */ = {isa = PBXBuildFile; fileRef = AC3C4A531A113EEC00143C57 /* ASCollectionViewProtocols.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		B31A241E1B0114FD0016AE7A /* AsyncDisplayKit.h in Headers */ = {isa = PBXBuildFile; fileRef = 6BDC61F51978FEA400E50D21 /* AsyncDisplayKit.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		B35061DF1B010EDF0018CF92 /* AsyncDisplayKit-iOS.h in Headers */ = {isa = PBXBuildFile; fileRef = B35061DE1B010EDF0018CF92 /* AsyncDisplayKit-iOS.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		B35061F31B010EFD0018CF92 /* ASCellNode.h in Headers */ = {isa = PBXBuildFile; fileRef = 055F1A3A19ABD43F004DAFF1 /* ASCellNode.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		B35061F41B010EFD0018CF92 /* ASCellNode.m in Sources */ = {isa = PBXBuildFile; fileRef = 055F1A3B19ABD43F004DAFF1 /* ASCellNode.m */; };
+		B35061F51B010EFD0018CF92 /* ASCollectionView.h in Headers */ = {isa = PBXBuildFile; fileRef = AC3C4A4F1A1139C100143C57 /* ASCollectionView.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		B35061F61B010EFD0018CF92 /* ASCollectionView.mm in Sources */ = {isa = PBXBuildFile; fileRef = AC3C4A501A1139C100143C57 /* ASCollectionView.mm */; };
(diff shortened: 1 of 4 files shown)


appleguy:
@secretiverhyme, any concerns for how this could affect the Facebook build systems?  I'll let you make the call on merging here.

clooth:
Any update on this?

appleguy:
@clooth & @nvh, there is some concern that this will affect clients' Xcode configurations or build systems, but I didn't hear back from our build wizard.  I'll take a closer look.

appleguy:
@nvh, @clooth ok, I think we can merge this, but I'm worried the additional target is a little confusing to people not using Carthage.  Does it need to be named with "-iOS"?  Can it be named simply "AsyncDisplayKit"?

clooth:
I don't *think* it needs to be suffixed with iOS.

appleguy:
Thanks for the update.  I see a AsyncDisplayKit-iOS.h, but the GitHub app is throwing an error when I try to check out this PR now (it was working fine last time I tried a few days ago).  AsyncDisplayKit-iOS.h doesn't seem to make sense to me as a name since there is no other supported platform, right?

I still want to land this ASAP if we can get the naming cleaned up.  My late-night development time is aimed at the ASNetworkImageNode fix, but if the naming isn't consistent for this patch by the weekend, I'll aim to fix and land it personally.

Hopefully this is understandable & not too frustrating.  My apologies for prior latency on this PR; we're super close now.

markmark1:
Thanks image node fix will be good will it be in the head branch

markmark1:
This is unrelated is there a way to get in the apple el capitan OS X and
ios 9 private beta list as its not yet available from apple developer site

nickvelloff:
Seems like anything related to `ASDisplayNode+Subclasses.h` is not getting included.
I get a whole bunch of warnings that look like the following and crash the app at runtime. 
```
        ^
/Users/travis/Library/Developer/Xcode/DerivedData/The_App_iOS-bsbsfdfdfdgfbfssfwgfqeecnlm/Build/Intermediates/ArchiveIntermediates/The App/BuildProductsPath/Release-iphoneos/AsyncDisplayKit.framework/Headers/ASDisplayNode+Subclasses.h:11:2: warning: missing submodule 'AsyncDisplayKit._ASDisplayLayer'
```

paulyoung:
That crash went away for us once the framework was copied via a build phase.

There are however 33 warnings all along the lines of:

```
Umbrella header for module 'AsyncDisplayKit' does not include header 'ASDisplayNode+Subclasses.h'
```

and

```
Missing submodule 'AsyncDisplayKit.ASDisplayNode_Subclasses'
```

appleguy:
Thanks @paulyoung  for the information — very helpful.  I would like to land this with clear naming (no -iOS unless it is necessary or clearly explained), and of course, it should not have any warnings that regular usage of the project does not have.

Is anyone volunteering to take this patch the rest of the way?  I will be rolling a 1.2.1 in the next 1-2 days for other fixes, and it would be great to include this, but I won't be able to commit to it myself.

nickvelloff:
@appleguy I would love to help out here I just can't make that deadline.

nickvelloff:
By the way it passes the Xcode upload to app store sniff test so it's not a full blocker FWIW.

appleguy:
I was able to get this checked out locally; it does not apply against master cleanly, apparently.  The -iOS name looks to be in a ton of places and is going to be really hard to rename unless the target is completely recreated.  What process was done to create it, and how easy or difficult is it to recreate?

You guys have my 100% support to add this new target — I simply feel it is important to get the name right the first time so we don't break anyone's build processes.  Once we land it with a redundant name like AsyncDisplayKit-iOS it would be hard to change.

Try using AsyncDisplayKit as the name, and let's get this out to engineers soon!  I was floored by how many people grabbed me at WWDC with thanks and enthusiasm about ASDK, so I think there are more people than I would have expected who would rather manage the framework manually instead of use cocoapods; I know they will appreciate it very much.

nickvelloff:
@appleguy +1 on appreciation for ASDK. Helped UICollectionView performance immensely.

appleguy:
@nickvelloff @nvh although we missed that release, I'd be happy to roll another one shortly after this lands with clean naming.

Anyone going to sign up for this?  Otherwise we should close out this PR until it can be re-developed.

clooth:
@appleguy I might be completely wrong here, but it seems like all targets must have unique names, which means we have to decide on either a -Dynamic suffix or something generic that's useable in the future as well. I only did a little poking around but it seems like so.

paulyoung:
Multiple targets can still have the same product name set in Build Settings.

I do this all the time for OS X and iOS frameworks. MyProject-OSX and MyProject-iOS both create a MyProject.framework

clooth:
@paulyoung Yeah I think that's not the issue that @appleguy means here. He's concerned of the naming of the target itself as it's redundant because iOS is the only supported platform. (Correct me if I'm wrong)

paulyoung:
I see. My two cents - having the iOS suffix is clearer and forwards compatible and doesn't have to affect the product name.

If a decision could be made on naming I could create a new framework target based on that. Probably easier to start over though.

clooth:
@paulyoung I'm with you on the -iOS suffix as well. I've seen it enough in other places for it to be familiar enough. +1

appleguy:
Thanks for weighing in @clooth and @paulyoung.  So be it — I am thrilled to have the community drive this project, and you make a compelling case.  Let's get this merged with the current naming, and say a few prayers that it doesn't break anything :) (I think we'll be good since it is a new target).

Appreciate the background on Carthage and the naming schemes (e.g. that -OSX is the other part of the pair).  There was once a version of ASDK that did work on OS X...but unfortunately I don't see that ever being released :).

nickvelloff:
@appleguy thanks for the merge! this is great news.

nvh:
I'm really sorry that I didn't have the time to participate in this discussion, but am happy to see that this PR is merged now. :tada:

PR #451, opened 2015-05-12
Fix a crash in setAsyncDelegate: that happens randomly (UIScrollViewAccessibility or NSProxy)

eanagel:
change the order we tear things down in `setAsyncDelegate`/`setDataSource` to avoid a random crash in `UIScrollViewAccessibility` or an `NSProxy` exception. I'm not sure exactly why, but it appears the `UIScrollView` will sometimes attempt to call the delegate while it is being nilled out. Previous code was deallocing the proxy before clearing `super.delegate`. 

This can manifest as a null ref in `UIScrollViewAccessibility` or as an illegal call to `NSProx`y, depending on exactly the order things happen.

**Update:** I've made some additional changes that help solve this problem, but it's important to manually nil `asyncDelegate` (and `asyncDataSource`) to avoid crashes. If you depend on ARC bad things will happen currently and I'm not sure exactly how to handle it. Root cause is that it's possible for the `_AS(Table|Collection)ViewProxy` objects to still exist and be wired when the underlying target has been nilled by arc. At this point any calls to the `delegate` or `dataSource` will throw an exception. Yuck. Would love suggestions on how to handle this. (Essentially we need to set `super.delegate|dataSource` to nil when `_proxyDelegate|DataSource` are nilled.)

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -78,11 +78,17 @@ static BOOL _isInterceptedSelector(SEL sel)
 
 - (BOOL)respondsToSelector:(SEL)aSelector
 {
+  ASDisplayNodeAssert(_target, @"target must not be nil"); // catch weak ref's being nilled early
+  ASDisplayNodeAssert(_interceptor, @"interceptor must not be nil");
+
   return (_isInterceptedSelector(aSelector) || [_target respondsToSelector:aSelector]);
 }
 
 - (id)forwardingTargetForSelector:(SEL)aSelector
 {
+  ASDisplayNodeAssert(_target, @"target must not be nil"); // catch weak ref's being nilled early
+  ASDisplayNodeAssert(_interceptor, @"interceptor must not be nil");
+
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -159,6 +165,12 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
   return self;
 }
 
+-(void)dealloc {
+  // a little defense move here.
```
appleguy: Maybe reference the github issue # here, or add a little more substance :).  Great comments overall.

appleguy:
This could be the most subtle issue we've seen a diff for.  Thanks for pushing through it and perfecting the call order of this code for everyone's benefit!

VivinoMorten:
I'm seeing the ASDisplayNodeAssert(_target, @"target must not be nil") failing - is that due to my app misusing AsyncDisplayKit or a bug in AsyncDisplayKit itself?

PR #456, opened 2015-05-19
Adapt Swift sample to Swift 1.2

nguyenhuy:
Got some errors while checking the sample. They are easy to fix though.


nguyenhuy:
Tests failed because Travis is using an old version (iOS 8.1). I tempted to update [the build config](https://github.com/facebook/AsyncDisplayKit/blob/master/build.sh#L3) as well, but found out that 8.1 is the latest version supported by Travis. I guess we have to live with it for now.

PS - they are working on [supporting Swift 1.2](https://github.com/travis-ci/travis-ci/issues/3216)

Issue #261, opened 2015-01-26
[ASTextNode] truncationStyle text overlapping, reproduction case included.


wockerwalker:
setting the `truncationStyle` of ASTextNode doesn't result in truncation. I've repro'd with `NSLineBreakByTruncatingHead/Tail/Clipping`. When set to `NSLineBreakByTruncatingTail`, the text runs to the end and then the final words (which should be truncated) stack on top of one another. Additionally, the `textNode.isTruncated` bool is `NO` when checked after sizing.

secretiverhyme:
Thanks for the report!

rnystrom:
For records sake, here's a screenshot of the repro. Good find :)

![kitten](https://cloud.githubusercontent.com/assets/739696/5925012/648092a0-a615-11e4-8e68-a252837d9682.png)

lxcid:
I'm not sure if this help, it seems that if the last character in the truncated line end with a space character, the rendering goes haywire. In my own project, I get overlapping character like the screenshot above, but in the kittens sample project, The text sometimes truncate too early just by adding a space character behind each new line. The screenshots come with a visual of the diff.

__No space character added at the end of each new line:__
![screenshot 2015-02-25 23 04 07](https://cloud.githubusercontent.com/assets/297156/6372996/be78edfe-bd42-11e4-99b2-69ec636f9b77.png)

__One space character added at the end of each new line:__
![screenshot 2015-02-25 23 03 41](https://cloud.githubusercontent.com/assets/297156/6373014/e3d32b96-bd42-11e4-8f95-81c9955bb50e.png)

appleguy:
I confirmed this is still happening.  Attached a diff.

An easy workaround is to use NSParagraphStyle to set the .lineBreakMode.  Still, this bug is very strange.

@ocrickard, I'm not sure if there is a comparable API in component kit to directly set the truncation mode separate from the paragraph style, or if there are any other known bugs with truncation in the renderer, but please let me know if so!

[TruncationOverlap.txt](https://github.com/facebook/AsyncDisplayKit/files/120540/TruncationOverlap.txt)

![simulator screen shot feb 6 2016 6 19 10 pm](https://cloud.githubusercontent.com/assets/565251/12870528/2a3abdc6-ccfe-11e5-8754-fadba007976c.png)
[...]

PR #462, opened 2015-05-22
Possible fix for text node truncation issue

lxcid:
This fix is related to #261.

First, the issue seems to originate from `ASTextNodeWordKerner` which is set as the `NSLayoutManager`'s delegate. If you prevent it from being the layout manager's delegate, rendering will work correctly.

Digging deeper, it seems like we are checking for the wrong glyph property.

Inside the code, we seems to be trying to search for whitespaces so that we can check its glyph property. The problem is that we are checking for `NSGlyphPropertyControlCharacter`.

In the documentation `NSGlyphPropertyControlCharacter` is described as follows:

> Control character such as tab, attachment, and so on, that has associated special behavior.

It seems like there's a mismatch here. Whitespaces are usually described with `NSGlyphPropertyElastic`. This is checked against the documentation...

> Glyphs with elastic glyph width behavior such as whitespace.

... and check against the original provided glyph property with something like this in the method...

```Objective-C
    NSGlyphProperty glyphProperty = properties[arrayIndex];
    switch (glyphProperty) {
        case NSGlyphPropertyNull: {
            NSLog(@"NSGlyphPropertyNull");
        } break;
        case NSGlyphPropertyControlCharacter: {
            NSLog(@"NSGlyphPropertyControlCharacter");
        } break;
        case NSGlyphPropertyElastic: {
            NSLog(@"NSGlyphPropertyElastic");
        } break;
        case NSGlyphPropertyNonBaseCharacter: {
            NSLog(@"NSGlyphPropertyNonBaseCharacter");
        } break;
    }
```

I'm unsure under what circumstances will whitespaces not describe as `NSGlyphPropertyElastic` by the system, thats something to ponder too.

Hope this help!

### DIFF
--- a/AsyncDisplayKit/Details/ASTextNodeWordKerner.m
+++ b/AsyncDisplayKit/Details/ASTextNodeWordKerner.m
@@ -31,7 +31,7 @@
       continue;
 
     // If we've set the whitespace control character for this space already, we have nothing to do.
-    if (properties[arrayIndex] == NSGlyphPropertyControlCharacter) {
+    if (properties[arrayIndex] == NSGlyphPropertyElastic) {
       usesWordKerning = YES;
       continue;
     }
@@ -43,7 +43,7 @@
     }
 
     // It's a space. Make it a whitespace control character.
-    newGlyphProperties[arrayIndex] = NSGlyphPropertyControlCharacter;
+    newGlyphProperties[arrayIndex] = NSGlyphPropertyElastic;
   }
 
   // If we don't have any custom glyph properties, return 0 to indicate to the layout manager that it should use the standard glyphs+properties.


lxcid:
I could repro the bug with the following changes to Kittens.

screenshot 2015-05-23 03 17 15

lxcid:
Okay, after seeing the test failing. I can maybe understand the rationale of checking whitespaces for `NSGlyphPropertyControlCharacter` property. It help trigger the code path to allow `ASTexTNodeWordKerner` to provide word kerned space width.

But this seems to cause problem for the truncation issue.

appleguy:
@lxcid thanks for researching this, and it was really great to meet you at the Pinterest WWDC party :).

It does look like this is a bit more complicated than this diff can address right now, so I'm going to keep the task open (and reference this diff) but close the PR.  There is a possibility of integrating a slightly newer text layout system from ComponentKit that is likely to fix some corner cases like this.

PR #468, opened 2015-05-28
Flexbox layout system, forked from ComponentKit

nguyenhuy:
## Author: Thanh Huy Nguyen

This is my proposal for the new Flexbox layout system for Texture.


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1374,6 +1362,43 @@ - (void)recursivelyClearFetchedData
...
+      stack.pop();
+    } else {
+      ASDisplayNodeAssertNotNil(context.layout.node, "node is required in calculated ASLayout.");
+      context.visited = YES;
+      
+      if ([context.layout.node isKindOfClass:[ASCompositeNode class]]) {
+        ASDisplayNode *subnode = ((ASCompositeNode *)context.layout.node).displayNode;
+        ASDisplayNodeAssertNotNil(subnode, "displayNode is required in ASCompositeNode.");
+        
+        CGPoint subnodePosition = context.absoultePosition;
```
levi: `absolutePosition`

nguyenhuy: Nice catch! Thanks :+1:

appleguy:
Thanks @nguyenhuy, this is one of the biggest undertakings that anyone in the community has started.  Now that WWDC has finally come and gone, I will prioritize working with you on this.  My co-workers at Pinterest are currently evaluating AsyncDisplayKit, and supporting asynchronous box model layout is something that got almost every developer excited.  I mentioned this PR specifically and it was really a great moment :)

Could you share your current thinking on this, such as:

- API thoughts / questions, areas to focus on
- Known issues
- If you have bandwidth to continue advancing this PR or need to hand it off

Awesome work, will add my diff review comments tonight or tomorrow!

nguyenhuy:
I guess it is also a ridiculously huge undertaking that whenever anyone in the community wants to review, the person immediately gets stack-overflow error :) 

It's also interesting to know that people at Pinterest are considering ASDK. I guess the Pinterest grid would gain a notable performance boost with ASDK. 

Anyway, let me sum up a few things about this PR.
- Pure Objective-C for public interfaces: My main objective for the work is to fork as much code from CK as possible. However, since CK relies heavily on Objective-C++, if the layout code was copied as-is into ASDK, ASDK clients would eventually have to convert all files to Objective-C++ (due to cascade effect when importing headers) and supporting Swift is not straightforward. So most of the work I spent on this PR is to convert all public parts to pure Objective-C (for example, no stacks or vectors in public interfaces, some structs are replaced by objects to be easily used in NSArray, etc). Original Objective-C++ code in private files were reserved.
	- Since some structs are converted to objects, we can no longer use [aggregate initialization](http://componentkit.org/docs/why-cpp.html#aggregate-initialization). Thus, creating a big data object with many default values can be a bit messy. One of the solutions is to provide a builder for such object. I implemented a builder for [ASStackLayoutNodeChild](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-f52e889555f498a5e617f0e17b8b34d4R90).
	- With all the changes in place, I think client code for creating a complex layout ([KittenNode](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-a75ff2e9b9a785e6ef072c54041d7241R134), for example) is quite similar to the [desired one in CK](http://componentkit.org/docs/why-cpp.html). It is still a good idea for some client code to be written in Objective-C++ and make use of aggregate initialization (but definitely not required). 
- Integration with current ASDK code base:
	- Naming convention: New components that are strictly layout related are called layout node, as opposed to display node. Every layout node class must contain *LayoutNode* in its name, to emphasize its nature. The naming convention may be confusing to some people, so let me know what you think.
	- Composition: Layout node inside display node: every display node contains a layout node that describes its layout. The display node measures its layout node and cache the result (`ASLayout`). Calculated size and position of each subnode can be retrieved from the calculated layout. I prefer not to put layout logic within display node itself for 3 reasons:
		- To reduce the complexity of display node.
		- To avoid unnecessary views/layers: There are layout nodes that describe empty spaces (such as inset) or the way to organize its children (for example stack, overlay and ratio layout nodes). Those layout nodes don't need a backing view/layer.
		- To avoid a deep hierarchy may required for complex layouts.
	- Immutable: Layout nodes are immutable. If there is any layout changes, the old layout must be invalidated. This approach actually fits nicely with the existing size caching mechanism.
	- Most of the time, ASDisplayNode can automatically handle laying out its subnodes [by walking through its layout tree](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-230ee3845413c2a11bf13fbdb7823ddfR1370). Custom nodes (like [KittenNode](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-a75ff2e9b9a785e6ef072c54041d7241R160)) can still implement their own logic.
	- [ASCompositeNode](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-643342af09fb0b79dcd06f7df5e79d1eR15): ASDisplayNode and ASLayoutNode are separate entities. ASCompositeNode is used to embed ASDisplayNode subnodes to their parent's layout. That way, each subnode will also be measured while the parent node is measuring, and the parent node knows where its subnodes are within its layout.
- API changes:
	- `-invalidateCalculatedSize` is replaced by `-invalidateCalculatedLayout`.
	- `-calculateSizeThatFits:(CGSize)` is replaced by `-layoutNodeThatFits:(CGSize)`, which usually is the only method custom nodes need to override.
	- Custom nodes do not need to layout its subnodes (in -layout:) anymore. 
- Snapshot test cases: ported directly from CK. Also, they are very good examples of how to use the new layout system.

Known issues/concerns:
	- Swift code for creating layout is not optimal: Creating C structs in Swift is not as sweet as aggregate initialization. It may be good to provide some handy ways for Swift users. It's low priority though.
	- Since the code is mostly from CK, I assume that it is quite fast. However, it would be a good idea to analyse its performance. 
	- Last but not least, any question or feedback on the overall architecture (especially layout node composition) is very welcome. 

Regarding my bandwidth, I'm committed to improving this PR until the feature is merged (and beyond that). The purpose for this PR is to propose and get some initial feedbacks.

nguyenhuy:
Another concern is compatibility with `UIStackView`. After a quick look at it, I think the equivalent (and maybe more powerful) class is [`ASStackLayoutNode`](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-f52e889555f498a5e617f0e17b8b34d4R13). Since this PR was made before WWDC, I was not aware of `UIStackView` and its public interface. I'm wondering whether `ASStackLayoutNode` should be updated to mimic `UIStackView` (as much as possible). Let me know what you think.

appleguy:
@nguyenhuy Thanks for this documentation—it's extremely helpful and very well written.  Seriously, it's not just a comment (you could practically publish it with the code!)  Wish you were on my team at work :)

I'm still reading it, and starting to take a detailed look.  My initial feedback:

- I am very happy to see you are adapting the API to Objective-C, and realize this is a ton of work.  It gives me great confidence that we can develop this into a landable version of FlexBox for ASDK, which honestly is a huge achievement (and I can say probably would be our biggest external contribution!)
- Impressed you implemented Layout Nodes.  Could be a good API solution for this; I was personally debating whether we should go that route, or integrate some form of flag / property options on nodes.  Layout nodes are probably the better choice but seemed possibly harder to get right.

I'll try to complete a review during the week, probably with a few different comments as I go.  If we complete and land this, I will definitely roll ASDK v1.3 for it and try to implement a few extra features on the weekends.

appleguy:
First comments:

- I want to make sure we minimize impact on apps / programmers that do not choose to use the FlexBox support, and write layout code manually.  Indeed, it should be supported to have some node hierarchies in an app use manual layout, and other parts (perhaps even levels of that tree?) use FlexBox.  Is this possible?
- My biggest concern with the implementation as-is would be performance cost of the new ASLayout* components.  Do you see any simple way to eliminate objects?  I believe one might be to remove ASLayoutChild and add a "position" property to ASLayout.

The diff is very comprehensive and quite clean overall!  Thanks for your ideas on API & keeping the performance + simplicity of existing manual layout implementations.  This may even mean that -calculateSizeThatFits: is still supported (as there are many clients currently using this), but that ASDisplayNode would assert only one method implementation is allowed (the other being -calculateLayoutThatFits).

nguyenhuy:
> It gives me great confidence that we can develop this into a landable version of FlexBox for ASDK
> If we complete and land this, I will definitely roll ASDK v1.3 for it and try to implement a few extra features on the weekends.
That would be awesome :)

> I want to make sure we minimize impact on apps / programmers that do not choose to use the FlexBox support, and write layout code manually. 

Absolutely. That's why [`-(ASLayout *)calculateLayoutThatFits:(CGSize)`](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-21f895ebcd9f9959caa2826a33f74315R125) is public for subclasses. Normally custom nodes would implement `-layoutNodeThatFits:(CGSize)`. The returned layout node will then be used to [measure a `ASLayout`](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-230ee3845413c2a11bf13fbdb7823ddfR1271). However if someone loves math and wants to write layout code manually, he/she can override `-calculateLayoutThatFits:` directly. Simple examples of this approach can be found in built-in display nodes, such as [ASImageNode](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-52d8ea287b97179d930f1df8c140394eR112) and [ASTextNode](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-221147908d17673747bdcbd88905332cR189). Again, `ASCompositeNode` should be used to embed subnodes to the `ASLayout`. Even if this approach is used, the programmer will have the benefit of doing math once, since the default implementation of `-layout:` can still do its job.

> Indeed, it should be supported to have some node hierarchies in an app use manual layout, and other parts (perhaps even levels of that tree?) use FlexBox. Is this possible?

Yes, it is. A display node only keeps a layout of its immediate subnodes/children. It doesn't (and shouldn't) know about its sub-subnodes/grandchildren. So it's possible to switch between auto and manual layouts at any level of a layout hierarchy.

> My biggest concern with the implementation as-is would be performance cost of the new ASLayout* components. Do you see any simple way to eliminate objects? I believe one might be to remove ASLayoutChild and add a "position" property to ASLayout.

Well, we can certainly flatten `ASLayout` tree by removing non-composite layout nodes (inset, stack, ratio, etc). Then, as you suggested, we can also store pairs of subnodes (or subnode identifiers) and their positions. I haven't implemented optimizations because the architecture is subject to change (and there is no performance measurements (yet), but I guess it's always a good idea to reduce the number of objects).

> This may even mean that -calculateSizeThatFits: is still supported (as there are many clients currently using this), but that ASDisplayNode would assert only one method implementation is allowed (the other being -calculateLayoutThatFits).

This is actually a great idea. And the implementation should be trivial.

appleguy:
Going through the deeper review this weekend.  More soon.

- I'm a bit concerned about the amount of duplication from ComponentKit.  The code is very clean, a testament to that framework in its adaptability, but it also feels like there are some portions of genuinely core functionality (particularly the internals still using C++ and the layout constraint logic).  I'm going to reach out to my friends who authored CK to see if they have any suggestions for a better way.  This code does seem stable enough that a fork is reasonable to consider, though.
- I think profiling will show that the overhead from using NSArray is material.  I saw a couple areas internal to the framework where we created an array just to pass into a method that unpacks the array into the C++ structures.  We don't have to fix this to land, but it would be a high priority to follow up and provide perhaps an internal API for those areas to use a fast-path.
- I have increasing confidence that we should adopt the -calculateLayoutThatFits: pattern.  I'm still thinking carefully about the right data structure for that to return, as well as the "LayoutNode" nomenclature.  I think it is a nice parallelism and we might use it as-is, but it could be slightly confusing because it is a fundamentally different type of node than a DisplayNode.  We might be better off avoiding the word Node entirely and calling it an ASLayoutCalculator or ASLayoutSpec, or something else.  Any ideas?

Lastly, could you please rebase this onto 1.2.1?  I am confident that we will land an improved form of this patch in the next ~2 weeks, and would like to make sure it doesn't develop conflicts with other changes.

appleguy:
I hit an assertion when running the Kittens example on an iPod touch 5G.  It works great in the simulator.  Is ASDisplayNodeCAssertPositiveReal supposed to permit FLT_MAX, or exclude it?  Looks like it is triggering on a FLT_MAX height.

screen shot 2015-06-20 at 11 23 05 am

AsyncDisplayKit/Layout/ASStackLayoutNode.mm
```diff
@@ -0,0 +1,133 @@
...
+  mutableChild.flexGrow = self.flexGrow;
+  mutableChild.flexShrink = self.flexShrink;
+  mutableChild.flexBasis = self.flexBasis;
+  mutableChild.alignSelf = self.alignSelf;
+  return mutableChild;
+}
+
++ (instancetype)newWithInitializer:(void (^)(ASMutableStackLayoutNodeChild *))initializer
+{
+  ASStackLayoutNodeChild *c = [super new];
```
appleguy: Why does this create a new object if the initializer block is present?  It seems like it will unconditionally overwrite the object without using it.

In using this API, I'm still a little confused about the necessity of the initializer block as opposed to passing in an actual instance of the ASStackLayoutNodeChild.  Anything you can share to explain this?

appleguy:
I noticed there are some methods that say -computeLayout, whereas the current methods say -calculateLayout.  I don't think either name is better — compute might be the best if it were the only one — but we should make these consistent.  To avoid thrashing people using the framework we should probably standardize on -calculate.

appleguy:
Writing some test code, and it's confirming my suspicion that the API setup here is not quite right.  It's close, but at least 2-3 steps away.

1) The two Node types are confusing.  I see ASCompositeNode and have no idea if it's an ASLayoutNode or ASDisplayNode.  Worse, ASStackLayoutNodeChild has a .node property.  Again it's unclear if it is a Layout or Display node, especially in the case that all I want to do is have a stack child that represents a single display node.
2) Wrapping display nodes in ASCompositeNode is cumbersome and annoying.  I don't yet fully understand the importance of ASCompositeNode.  I believe this may be an essential element in a purely declarative design like ComponentKit, but the style of AsyncDisplayKit is somewhat less rigorous and designed primarily to cater to common programmer workflows in structuring an app.  In this case if I want to represent a series of nodes in a stack, right now it seems unnecessarily tough.

For naming, I think we might consider:
- ASLayoutSpec
- ASLayoutCalculator
- ASLayoutGenerator
- ASLayoutSolver

appleguy:
There are an extraordinary number of size types that this introduces.  We need to find a way to reduce or better-organize them, and I'd love your input @nguyenhuy 

To start, of course we have standard types:
- CGSize
- CGFloat
- CGRect, CGPoint, etc

Of the new ones, the most confusing to me is ASLayoutNodeSize.  But look at all the types!
- ASLayoutNodeSize
- ASRelativeDimension
- ASRelativeDimensionType
- ASSizeRange
- ASRelativeSize
- ASRelativeSizeRange

I understand that most, if not all of these are originally from ComponentKit.  It's also definitely true that they provide useful power and utility that we need to specify Flexbox constraints.  However, as I'm creating layout nodes, composite nodes, display nodes, stack children, etc...the types required are often different, and constructing the correct data type can take some work.

As I continue to experiment with writing code against this model, I may have some simplification suggestions.  However, I suspect you have better ideas to offer at this stage and it would probably help me make progress more quickly.  Let me know!

One example of the confusion I mean is when reading ASDimension.h, I'm really not sure what to make of these constants.  They seem really useful and likely needed on a regular basis, but what is the difference between just Auto, Auto and the /RANGE/ from Auto, Auto to Auto, Auto?! :-D

/** type = Auto; value = 0 */
extern ASRelativeDimension const ASRelativeDimensionAuto;

/** min = {0,0}; max = {INFINITY, INFINITY} */
extern ASSizeRange const ASSizeRangeUnconstrained;

/** width = Auto; height = Auto */
extern ASRelativeSize const ASRelativeSizeAuto;

/** min = {Auto, Auto}; max = {Auto, Auto} */
extern ASRelativeSizeRange const ASRelativeSizeRangeAuto;

nguyenhuy:
Thanks for the deeper review round, @appleguy.

Firstly, I admit that the "layout node" naming convention is somewhat confusing. Since changing it is not difficult but a bit time consuming, I would prefer to do it only once :) I vote for ASLayoutSpec.

> I see ASCompositeNode and have no idea if it's an ASLayoutNode or ASDisplayNode
> Wrapping display nodes in ASCompositeNode is cumbersome and annoying. I don't yet fully understand the importance of ASCompositeNode. 

ASCompositeNode is more like a bridge between ASLayoutNode and ASDisplayNode. There are 2 problems I tried to solve using ASCompositeNode is:
- To identify/indicate where a display subnode is within its parent's layout.
- While calculating, a container ASLayoutNode asks its children to return their layouts. If one of the child layout node happens to be the place where a display node will be attached later on, we must figure out a way to get that display node and ask for its size.

The first problem can also be solved by employing a subnode identifier/tag (which may result in a new required API method in ASDisplayNode). The second problem is more difficult to solve without using a bridge/ASCompositeNode.

> I'm still thinking carefully about the right data structure for that to return

I think we can simplify the calculated ASLayout. As you suggested before, we may not need ASLayoutChild, if we have a way to identify each display subnode within its parent. Let me know if you have a better solution in mind.

nguyenhuy:
Regarding new size types, I agree that the learning curve is quite big. All of the new types are in fact ported from CK. We can remove some of them, although some simplifications will result in a less flexible layout model. Some suggestions that I have in mind is:
- Remove ASSizeRange and ASRelativeSizeRange: They contains min and max CGSizes. The min size is actually not very useful in ASDK, since the sizing nature of ASDK is that display nodes try to fit into its constrained size (or max size) as much as possible. So we may be able to replace those size ranges with CGSize.
- ASLayoutNodeSize might be removed for the above reason - max width and height are the only values that make sense in ASDK.
- Reduce the number of ASRelativeDimensionTypes (which has 3 values: Auto, Percent and Points).
	- Points is the only required one. 
	- Auto is the type that I think we can get rid of without any major effect.
	- Percent could be useful for container layouts. For example a stack layout may have 3 children and each of them takes 30%, 50% and 20% of the parent's space. Removing Percent would make this use case quite difficult to achieve. 
	- We will be able to remove ASRelativeDimension and ASRelativeSize if Points is the only type supported. So the real question is, do we want Percent?

nguyenhuy:
> I'm a bit concerned about the amount of duplication from ComponentKit. I'm going to reach out to my friends who authored CK to see if they have any suggestions for a better way. 

That would be tremendously helpful, thanks. Please add me to the conversation if possible, I am not knowledgeable enough in either ASDK or CK to engage in such conversation, but it will be a good learning opportunity for me.

> I saw a couple areas internal to the framework where we created an array just to pass into a method that unpacks the array into the C++ structures. We don't have to fix this to land, but it would be a high priority to follow up and provide perhaps an internal API for those areas to use a fast-path.

Yes, unpacking the array into C++ structures is more like a hack I used to help bridge between the new (pure Objective-C) APIs and original (C++ heavy) internal implementation. I will create follow up issues once this main PR is merged.

> Is ASDisplayNodeCAssertPositiveReal supposed to permit FLT_MAX, or exclude it? Looks like it is triggering on a FLT_MAX height.

I believe FLT_MAX is permitted and can't find any reason to think otherwise. I don't know why the actual assertion code excludes FLT_MAX and have no evidence from CK history (the code was committed in the huge "Initial commit"). Perhaps it is a good idea to reach out to the original author of the code for an explaination?

> I noticed there are some methods that say -computeLayout, whereas the current methods say -calculateLayout. I don't think either name is better — compute might be the best if it were the only one — but we should make these consistent. To avoid thrashing people using the framework we should probably standardize on -calculate.

Will do, real soon :)

AsyncDisplayKit/Layout/ASStackLayoutNode.mm
nguyenhuy: Sure. This builder pattern can be used to create a big data object (ASStackLayoutNodeChild in this case) with most of its properties have default value. It also allows users to create the object in an inline fashion, as opposed to create using a default initializer and then change a few needed properties. 

This pattern is the result of borrowed ideas from [this blog post](http://cocoanuts.mobi/2014/05/15/builder/) and [Pop](https://github.com/facebook/pop/blob/master/pop/POPAnimatableProperty.mm#L1244).

wangii:
`ASLayoutSpec` reminds me Android. How about `ASFlexLayout`?

nguyenhuy:
Maybe we should set up a poll for this?

nguyenhuy:
And the thing I concern most about the final naming convention is: how the client code that can declare a complex, highly nested layout will look like.

appleguy:
Thanks — great questions, will reply to all.

- I've asked some friends at Pinterest to help weigh in on the naming idea, because it really is best to see what feels natural to more people.  I also like ASLayoutSpec, although the Android connection is funny.

- I really don't like ASCompositeNode.  If we do keep it, the name should be changed; maybe even ASAdapterNode or ASContainerNode.  Actually...if we change ASLayoutSpec, maybe we could use ASLayoutNode there :).  Honestly I think it will be confusing no matter what it is named, and we should strive to eliminate it.

- Adding an ID is a fine idea, and in fact, I think we might find some other uses for this that could help efficiency in certain places.

- Can you explain an example of the case where a subnode is not yet present, but the layout must be calculated for it?

- Could you spend some time thinking about how we could better support calculating new / updated layouts for a given hierarchy?  For example, in order to handle the case of the sub-node not yet present, could we leave it out of the initial layout and then trigger a new asynchronous measure: pass with the real node available to put into a new ASLayoutSpec assembly?  If so, then we would want a completion block at the end of the async layout to mark the new elements as un-hidden — or it could be used to trigger an animation from the old ASLayout to the new ASLayout.  I think this could be one of the coolest things enabled by this diff, and would directly solve a problem that two critical ASDK clients need to overcome.

- Please do spend some time on simplifying ASLayout, if it can be done without compromising functionality.  This would materially help us get to a merge-able state.  I'm happy to file a few followup tasks that can be handled over time for a variety of stuff in the PR to get it merged quickly, but focusing especially on these API details will be great.

appleguy:
Now replying to your next comment on size types.

I agree with your assessment of the functionality provided by the additional options and believe we should preserve this functionality.  That includes the Percent type of the relative positioning.

I am glad you mentioned the Auto option, because when I encountered that part of the definitions, it is really when I started to feel like I did not understand the system.  I in fact still don't understand what Auto does.  If you could explain what we lose by removing it (I did not look in detail at its implementation), that would be good, but let's assume we take it out!  Just too many options and not likely to add that much value over the other two.

appleguy:
I did talk with the CK authors.  They are supportive of the effort and agreed that forking is the right decision for now, particularly because of the API differences, but also because of the stability of the code (unlikely to change) and inherent differences that React Native's implementation has (so we couldn't share a version across all three anyway).

The FLT_MAX question I'm still not sure about.  I think we should just change the condition so it allows this.  I did that locally for my weekend experimentations and it worked fine for my purposes.

appleguy:
@wangii Thanks for weighing in with the suggestion!  Unfortunately, the name has to also include manual layout implementations, as that will be one of the common types of subclasses of the base layout node / layout spec object.  So we should avoid putting Flex in the name.

garrettmoon:
I'd vote on ASLayoutSpec as well: layout spec indicates what the class does for the developer (as opposed to calculator or generator which indicates what the class does to achieve the spec). Mostly though, I agree that having node in the name would be confusing.

This is very cool by the way! I'm new to working with ASDK and the duplication of code in calculate and layout was one of the first drawbacks I encountered.

nguyenhuy:
@appleguy Thank you for your answers.

There are some new commits that remove ASLayoutNodeSize, rename -computeLayoutThatFits and fix CGFLOAT_MAX issue. I will post yet another comment on my TODO list and answers to your questions soon, probably within a couple of days.

Quick heads up: I've rebased and forced push the branch. I ended up doing a rebase instead of a merge because it's easier to fix conflicts commit by commit (especially conflicts in `project.pbxproj`). And I'm the only one who is committing to the branch anyway. If you happen to have a local copy, `reset --hard` is your friend :P

nguyenhuy:
@garrettmoon Thank you for your input. It's great to see people like the work :)

appleguy:
OK, so ASLayoutSpec and ASLayout.  ASStackLayoutSpec, ASInsetLayoutSpec...hmmm, it doesn't immediately suggest that they are compose-able, that you could combine multiple specs containing each other even for just one node's layout definition.  Maybe ASLayoutDefinition?  ASStackLayoutDefinition, ASInsetLayoutDefinition.  A lot less nice to type, so perhaps not.

ASLayout itself is a pretty good name as it feels short and struct-like, but I'm not sure it is perfect because it could be unclear exactly how it relates to the Spec / Definition.  Maybe ASComputedLayout, or ASCompletedLayout.  Other ideas?  We should try to communicate that this object is immutable / is the output of one running of the Spec / Definition.

Blueprint is probably not the right word to use, but this does feel like ASLayoutBlueprint (the spec object) and then there needs to be something like ASLayoutInstance / ASLayoutManifestation.  We need the right pair of words to describe the generator / abstract version, and the generated / concrete version.

nguyenhuy:
I pushed new commits that remove Auto size, simplify ASDimension header and update Kittens example (which makes Travis happy).

> Can you explain an example of the case where a subnode is not yet present, but the layout must be calculated for it?
Sorry, I think you misunderstood me because of my bad English. I simply meant we need to somehow ask the right subnode for its size while calculating the parent node layout.

Regarding the Auto option, in short, we don't lose anything by removing it. After having a closer look at the code, I realize that it means slightly different things in different situations:
- In static layout, an Auto width/height can be as big as the child wants, **within its parent bounds** (which is finite).
- In stack layout, an Auto flex basis can take as big as the child wants (in stack dimension), **without limits**.
  - If the returned stack size exceeds parent's stack size, a violation number will be computed and distributed equally to flexShrink-enabled childs, causing them to shrink. Then, if the violation still can't be resolved, justifyContent will be taken into account.
- [Those behaviours are replaced by default values](https://github.com/nguyenhuy/AsyncDisplayKit/commit/93564899fdbe6d0c941f727d08f40db45236e352).

For supporting calculating new/updated layouts, I think there are few things enabled by this diff:
- Animation: we can compare the old and new ASLayouts and figure out insert/resize/delete animations needed.
- Layout optimization: we can compute and only measure dirty parts of the layout assembly, maybe by a flag and/or notification from subnodes. Dirty parts include inserted/resized/deleted layout specs within the root spec.
- Completion callback: we can execute the block once we computed the final layout.

Anyway, here are things on my plate right now:
- ASCompositeNode: I will probably add an ID to display node and may be a measure delegate used to ask subnodes for their sizes.
- Simplify ASLayout: Still thinking. Once I have a display node ID, I think ASLayout will be really simple.
- Rename AS*LayoutNode.

appleguy:
Great, this looks good.  Replacing Auto with the Unconstrained type seems much easier to understand.

I'm really excited about the "relayout" possibilities this enables.  We can first focus on landing the overall diff, but because several of the biggest clients of ASDK really need this relayout feature, it's something I would love to see added afterwards.  The other important part is a completion handler for asynchronous layout, and it probably makes sense to allow ASDisplayNode to start managing a global layout queue instead of requiring classes like ASDataController to do it.  This way, a client app could kick off asynchronous sizing without even referencing a queue, and also get the completion callback :).

Let me know how I can help.

appleguy:
The header definition for this method on ASDisplayNode should presumably be in the *Subclasses file:

- (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize;

The comment also needs to be cleaned up as it mentions calculatedSize.  If I understand the system correctly, a subclasser should implement only ONE of the three methods: 1) returns ASLayoutSpec, 2) returns ASLayout directly, or 3) returns a calculated CGSize.

I think it is an interesting question to consider if we should drop support for -calculateSizeThatFits:.  My preference is that if we can continue supporting it for now without much work (and I think we can), then let's leave it until we can very carefully think about the implications of removal for all of the different use cases of ASDK.  If we do roll an ASDK v2.0 in the next month or two, it would be the best time to make a breaking API change of that nature, but it would be necessary to write a dead-simple migration guide that explains how to quickly and accurately convert a calculate-size method to a calculate-layout method.

xujialiang:
������������������������

------------------ �������� ------------------


    
        ������: appleguy  <notifications@github.com>;
    
    
        ��������: 2015-06-21 09:58:23
    
    
        ������: facebook/AsyncDisplayKit  <AsyncDisplayKit@noreply.github.com>; 
    
    
        ����: (��); 
    
    
        ����: Re: [AsyncDisplayKit] Flexbox layout system, forked from ComponentKit(#468)
    

There are an extraordinary number of size types that this introduces.  We need to find a way to reduce or better-organize them, and I'd love your input @nguyenhuy   To start, of course we have standard types:   CGSize CGFloat CGRect, CGPoint, etc   Of the new ones, the most confusing to me is ASLayoutNodeSize.  But look at all the types!   ASLayoutNodeSize ASRelativeDimension ASRelativeDimensionType ASSizeRange ASRelativeSize ASRelativeSizeRange   I understand that most, if not all of these are originally from ComponentKit.  It's also definitely true that they provide useful power and utility that we need to specify Flexbox constraints.  However, as I'm creating layout nodes, composite nodes, display nodes, stack children, etc...the types required are often different, and constructing the correct data type can take some work.  As I continue to experiment with writing code against this model, I may have some simplification suggestions.  However, I suspect you have better ideas to offer at this stage and it would probably help me make progress more quickly.  Let me know!  One example of the confusion I mean is when reading ASDimension.h, I'm really not sure what to make of these constants.  They seem really useful and likely needed on a regular basis, but what is the difference between just Auto, Auto and the /RANGE/ from Auto, Auto to Auto, Auto?! :-D  /** type = Auto; value = 0 */ extern ASRelativeDimension const ASRelativeDimensionAuto;  /** min = {0,0}; max = {INFINITY, INFINITY} */ extern ASSizeRange const ASSizeRangeUnconstrained;  /** width = Auto; height = Auto */ extern ASRelativeSize const ASRelativeSizeAuto;  /** min = {Auto, Auto}; max = {Auto, Auto} */ extern ASRelativeSizeRange const ASRelativeSizeRangeAuto;  ��Reply to this email directly or view it on GitHub.

nguyenhuy:
@appleguy: It's a great hint, thanks. I solved the compatible issue with recent commit. -calculateLayoutThatFits: (and the old manual layout model) is still supported. This PR should introduce no breaking API changes.

I will fix the naming convention very soon. I guess the final names would be ASLayoutSpec, ASLayout and ASLayoutable (the protocol that both ASLayoutSpec and ASDisplayNode conforms to).

Thinking about ASLayout. I figure the current ASLayout structure can be used as an intermediate product. It then will be flatten into a new, lightweight structure that contains size of the parent node and pairs of position-size of each subnode.

P/S: The current branch have many conflicts, most (if not all) of them are in project.pbxproj. That's annoying. I will just go ahead fixing remaining issues, then make a final rebase right before merging.

markmark1:
project.pbxproj. To fix conflicts in this file

Abort merge copy the original file out 

Do merge and copy  the original pbxproj file back make the changes to it manually git add and commit

nguyenhuy:
@markmark1 Thanks a lot, your trick works. Not sure if I'm breaking anything, but all tests passed locally and Travis is happy.

appleguy:
@markmark1 thanks for weighing in — agree that is a better tactic and glad it was helpful to @nguyenhuy !

@nguyenhuy I like the names you mentioned.  Let's lock them down and move forward with the rename.

Can you better explain what you mean by the changes to the ASLayout structure?  I'm happy with the name, but do think that there could be some improvements to what the structure actually contains (IIRC there's an ASLayoutChild or something which seemed probably avoidable).

AsyncDisplayKit/Layout/ASLayout.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+#import <UIKit/UIKit.h>
+#import <AsyncDisplayKit/ASAssert.h>
+#import <AsyncDisplayKit/ASLayoutable.h>
+
+/** Represents the computed size of a layout node, as well as the computed sizes and positions of its children. */
+@interface ASLayout : NSObject
+
+@property (nonatomic, readonly) id<ASLayoutable> layoutableObject;
```
appleguy: This is really cool!  I am pretty happy with how this should be able to eliminate the ASCompositeNode.

AsyncDisplayKit/Layout/ASLayout.h
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Convenience that does not have any children.
+ */
++ (instancetype)newWithLayoutableObject:(id<ASLayoutable>)layoutableObject size:(CGSize)size;
+
+@end
+
+@interface ASLayoutChild : NSObject
+
+@property (nonatomic, readonly) CGPoint position;
```
appleguy: Let me know what you think about removing ASLayoutChild completely, and adding a .position property to ASLayout itself.  What bad / weird effects would this cause?

I think the main issue is that the .position would be unset / invalid for the top level ASLayout (e.g. for a given node, its ASLayoutSpec could generate an ASLayout that has a valid size and valid "sublayouts" with both size and position, but not a valid position for the top-level ASLayout).  I think this would be OK, but I'm not sure if we need a special constant-type CGPoint to mark this as invalid (sorta like kCGRectNull; is there a similar thing for a CGPoint?)

Not as critical, but I'm also thinking it would be ideal to rename usages of "children:" that came from ComponentKit.  This term isn't used anywhere else in ASDK and it would be better to come up with a more specific term.  subLayouts, subLayoutable (eh, not really great), subnodes would be perfect but does this ever need to contain sub-elements that aren't actual nodes?  I guess that only applies to ASLayoutSpec, and the actual generated ASLayout should only ever describe the qualities of real subnodes, right?  If that's true, let's use subnodeLayouts:

nguyenhuy:
@appleguy Few new changes:
- ASLayoutChild is gone, for good.
- ASLayout now has a position, defaults to CGPointNull (NAN, NAN). I also added some asserts to make sure a valid position is set once (and only once) to sublayouts.
- ASDisplayNode now stores a flattened ASLayout. This should help reduce memory footprint and give -layout a boost.

An uncompressed ASLayout actually contain both layout specs and subnodes. So sublayouts looks like a better name in this case. I will fix all naming issues altogether, which looks like the only task left in this PR. It's exciting to see how this PR evolves and get very close to this merge-able state.

appleguy:
@nguyenhuy great step with what are probably the final structural changes; I appreciate the strategy of keeping the renames separate.  Let's get that big push in and then do a final detailed review.  I will prioritize breaking out future changes into separate improvement diffs so we can land the large one soon, assuming the API is solid as it is harder to change.

nguyenhuy:
I think I'm done fixing naming conventions. It's time for nitpicking :)

AsyncDisplayKit/ASCellNode.m
```diff
@@ -106,21 +105,12 @@ - (instancetype)init
   return self;
 }
 
-- (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
+- (id<ASLayoutable>)layoutSpecThatFits:(ASSizeRange)constrainedSize
```
appleguy: Is there a use case where we would want this method to return something other than ASLayoutSpec itself?  I see the cases where an ASLayoutSpec would want to directly contain an ASDisplayNode, rather than wrapping with the ASCompositeNode/Spec.  But especially because the name of this method is layoutSpecThatFits:, it's a bit confusing to me if it should actually allow returning an ASDisplayNode with nothing else.  I suspect the answer is no...but that we perhaps should support returning an ASLayout as well as ASLayoutSpec, so if you wanted to manually create an ASLayout and set it up to describe the position of subnodes, you could do that without subclassing ASLayoutSpec.

Thanks for any thoughts on this.  Since this API will be very commonly used and indeed is arguably /the most important API/ which ASDK adds on top of UIKit, it's very important that we nail the naming and semantics.  I think we are very close and the new ASLayoutSpec name makes a great deal of sense, but let's think carefully about any related method names and their return types.

appleguy:
Phenomenal, thanks @nguyenhuy!  I will review this in detail as soon as I can — first thing on Saturday morning (possibly Friday night).  Are there any remaining issues you're aware of?  Could you think through any APIs added or changed on ASDisplayNode and consider what it would be like to explain their purpose and functionality to a new developer, and think if that story is easy to explain?

AsyncDisplayKit/ASCellNode.m
nguyenhuy: If a display node want to return a ASLayout directly (kind of semi-auto layout), it can overrides -layoutThatFits: and return a flat ASLayout that describe its size, as well as position and size of subnodes. I think the new APIs added to ASDisplayNode can support a great deal of use cases, including auto, semi auto and manual layout.

nguyenhuy:
@appleguy awesome. Thanks.

Regarding API changes in ASDisplayNode. I think newly-updated documentation of [ASDisplayNode+Subclasses.h](https://github.com/facebook/AsyncDisplayKit/pull/468/files#diff-21f895ebcd9f9959caa2826a33f74315R112) explains the idea quite clearly (or maybe not). Anyway, let me try to explain it again here:

The main APIs added are 2 new method -layoutSpecThatFits: and -calculateLayoutThatFits:. Those methods, together with the existing -calculateSizeThatFits:, help to cover most (if not all) of use cases:
- Manual layout: supported by implementing -calculateSizeThatFits: **and** -layout. This is the model used by existing clients and there is no breaking API changes introduced.
- Auto layout: supported by implementing -layoutSpecThatFits:. This should be the recommended way once this PR is merged.
- Semi-auto layout: supported by implementing -calculateLayoutThatFIts:. This should be considered as an advanced model and should rarely be used. The implementation should construct and return a flat ASLayout that describes its size, as well as position and size of subnodes. As long as a correct ASLayout is returned, it will be safely cached and correctly used by -layout.

There is no high-level issue that I'm aware of. However there are 2 minor concerns I'm having in mind:
- [Static layout spec](https://github.com/nguyenhuy/AsyncDisplayKit/blob/layout_node/AsyncDisplayKit/Layout/ASStaticLayoutSpec.h):
  - It can be helpful to describe a static part of the layout. But it requires a bit of learning to use correctly (new ASStaticLayoutChild and some size types).
  - I'm thinking about eliminating ASStaticLayoutChild by using the same way we eliminate ASStackLayoutChild, but it leads to another concern.
  - Or perhaps we should just remove it completely?
- (Yet another) naming convention issue: ASLayoutSpec calls its sub ASLayoutable objects: "children". I didn't change it to subLayoutables because the new name does sound good either. I think it is not really a critical issue and we can live with it for now?

nguyenhuy:
Other small API changes on ASDiplayNode are:
- New `calculatedLayout` property. (`calculatedSize` is still there, which simply returns the calculatedLayout's size)
- `-invalidateCalculatedSize` is replaced by `-invalidateCalculatedLayout`.
- `constrainedSizeForCalculatedSize` property is replaced by `-constrainedSizeForCalculatedLayout`.

appleguy:
Thanks for this overview — very helpful, definitely what I was interested in.  I need to take a deeper look for better feedback, but I think we're just about ready to merge.  The next steps:

- We absolutely have to get a new release of ASDK out before this lands, because it does introduce some API changes (for example, clients that use -invalidateCalculatedSize will need to change it, which is the right thing rather than keeping both methods).  That release needs to include my ASDataController fixes and I expect to be able to get it out on this weekend.
- We then land this and loudly announce it so that people start looking.  Either leave on master or create a "Beta" release tag like we've done before.  I'm hopeful that we can do this landing & beta release this weekend as well (Sunday).  I think the main limiting factor here is getting the stable release out with adequate testing and no known issues (there are a couple known ones that I haven't had time to fix, e.g. the ASFlowLayoutController problem).
- Quickly follow up with improvements as we start to get experience with using the beta.

AsyncDisplayKit/ASCellNode.m
```diff
@@ -12,6 +12,7 @@
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
 #import <AsyncDisplayKit/ASTextNode.h>
 
+#import "ASInsetLayoutSpec.h"
```
appleguy: Since this is a public header, it should be <AsyncDisplayKit/ASInsetLayoutSpec.h> — right or maybe not?

AsyncDisplayKit/ASCellNode.m
```diff
@@ -132,7 +122,7 @@ - (void)setText:(NSString *)text
   _textNode.attributedString = [[NSAttributedString alloc] initWithString:_text
                                                                attributes:@{NSFontAttributeName: [UIFont systemFontOfSize:kFontSize]}];
 
-  [self invalidateCalculatedSize];
+  [self invalidateCalculatedLayout];
```
appleguy: What do you think about naming this -invalidateLayout?  I would suggest this change later, but anyone who updates to master (a fair number of ASDK clients build against master) would have to change it in their code again.

In thinking about this, consider what we will need to do as the very most important change after this one (even better if we can do it now): more gracefully handling re-layouts, like what happens today if you rotate the Kittens app (text does not reflow, new cells are at the full width but also do not re-wrap when rotating back to portrait).

If that were fixed, we'd need to understand the expected behavior in handling asynchronous layout; fortunately, ASDataController already does this in batches, so if we passed through ASTableView / ASCollectionView -layoutSubviews calls as triggering cell-relayout, things would probably show up at once.  We shouldn't do anything specific for scroll point restoration as the cells change height.

With those changes, I suspect we would make -setNeedsLayout do the bulk of this work: doing that would kick off an asynchronous measurement pass and then automatically trigger -layout afterwards.  I'm not totally sure, but think we might be able to get rid of -invalidateLayout entirely...which would be a huge success enabled by your work, as -invalidateCalculatedSize has always been one more custom API for developers to try to understand.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -65,6 +67,17 @@
  */
 @property (nonatomic, readonly, assign, getter=isInHierarchy) BOOL inHierarchy;
 
+/**
+ * @abstract Return the calculated layout.
+ *
+ * @discussion Ideal for use by subclasses in -layout, having already prompted their subnodes to calculate their size by
```
appleguy: As of this patch, is it still necessary for each node to call -measure: on the subnodes in -calculateLayoutThatFits:?  That is surely necessary for manual layout areas, but for layout spec areas, the -measure: calls will be prompted automatically — right?

Update the comment to be a bit more explicit about this (feel free to add another sentence describing the automatic layout / spec case and clarify the first is for manual / nodes that override -calculateLayoutThatFits:

AsyncDisplayKit/ASCellNode.m
nguyenhuy: `-invalidateLayout` could be a better (and shorter) name. But then we also need to think about renaming `calculatedLayout` property, since `-invalidateLayout` doesn't clearly imply that the `calculatedLayout` will be invalidated. Should we change `calculatedLayout` to `layout`?

I think it's worth investigating re-layout handling. Definitely an interesting follow-up ticket.

AsyncDisplayKit/ASCellNode.m
nguyenhuy: Fixed.

appleguy:
For those arriving here from my group posts about ASDK 1.2.2: we expect to land this diff to master within a day or so.  After discovering a small issue with API calling convention used internal to the ASLayoutSpec implementations included in this diff, we are cleaning it up before landing this, so that pushes back the prior plan to land this weekend.

Nonetheless, this huge feature is coming soon and we'd love input from anyone who checks out this PR and tries out the new layout specs.  The examples/Kittens app includes #define UseAutomaticLayout on by default to demonstrate use cases in both BlurbNode and KittenNode.

nguyenhuy:
@appleguy I've push a new commit that should fix the blocking issues. Please re-check the diff. Thanks.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
nguyenhuy: Comment updated.

AsyncDisplayKit/ASCellNode.m
nguyenhuy: After discussing, we agreed to not change the name.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
markmark1: What is the right way to do rounded corners of images or buttons if border radius is slow in objective c

nguyenhuy: @markmark1 Check out ASImageNodeRoundBorderModificationBlock :)

On 6 Jul 2015 at 10:03:53 pm, markmark1 (notifications@github.com) wrote:

In AsyncDisplayKit/ASDisplayNode+Subclasses.h:

> @@ -65,6 +67,17 @@
>   */
>  @property (nonatomic, readonly, assign, getter=isInHierarchy) BOOL inHierarchy;
>   
> +/**
> + * @abstract Return the calculated layout.
> + *
> + * @discussion Ideal for use by subclasses in -layout, having already prompted their subnodes to calculate their size by
What is the right way to do rounded corners of images or buttons if border radius is slow in objective c
…

markmark1: Thanks huy checking in general outside asdk how do you create rounded corners

nguyenhuy: Similar to the modification block :) Its implementation can be extracted and should work outside of ASDK.
On 6 Jul 2015 at 10:26:00 pm, markmark1 (notifications@github.com) wrote:

In AsyncDisplayKit/ASDisplayNode+Subclasses.h:

> @@ -65,6 +67,17 @@
>   */
>  @property (nonatomic, readonly, assign, getter=isInHierarchy) BOOL inHierarchy;
>   
> +/**
> + * @abstract Return the calculated layout.
> + *
> + * @discussion Ideal for use by subclasses in -layout, having already prompted their subnodes to calculate their size by
Thanks huy checking in general outside asdk how do you create rounded corners
…

markmark1: Can you give a sample as our asdk and non asdk project have  lots of clips to bounds rounded corners

nguyenhuy: Sorry I slept before your latest message arrived. I can help you privately. Feel free to send me an email. My email address is huy.nguyen151190 at gmail dot com.

On 6 Jul 2015 at 10:44:30 pm, markmark1 (notifications@github.com) wrote:

In AsyncDisplayKit/ASDisplayNode+Subclasses.h:

> @@ -65,6 +67,17 @@
>   */
>  @property (nonatomic, readonly, assign, getter=isInHierarchy) BOOL inHierarchy;
>   
> +/**
> + * @abstract Return the calculated layout.
> + *
> + * @discussion Ideal for use by subclasses in -layout, having already prompted their subnodes to calculate their size by
Can you give a sample as our asdk and non asdk project have  lots of clips to bounds rounded corners
…

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -96,8 +118,39 @@
...
+ *
+ * @discussion Though this method does not set the bounds of the view, it does have side effects--caching both the
+ * constraint and the result.
+ *
+ * @warning Subclasses must not override this; it caches results from -calculateLayoutThatFits:.  Calling this method may
+ * be expensive if result is not cached.
+ *
+ * @see [ASDisplayNode(Subclassing) calculateLayoutThatFits:]
+ */
+- (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize;
```
appleguy: I believe this is intended to be in ASDisplayNode.h, or some other category header (is there one for Layout?) - but not the Subclasses header.  I think we should move this before landing if that understanding is true.

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -128,10 +128,12 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
...
- * @warning Subclasses must not override this; it caches results from -calculateSizeThatFits:.  Calling this method may 
+ * @warning Subclasses must not override this; it calls -measureWithSizeRange: with zero min size. 
+ * -measureWithSizeRange: caches results from -calculateLayoutThatFits:.  Calling this method may 
  * be expensive if result is not cached.
  *
- * @see [ASDisplayNode(Subclassing) calculateSizeThatFits:]
+ * @see [ASDisplayNode(Subclassing) measureWithSizeRange:]
+ * @see [ASDisplayNode(Subclassing) calculateLayoutThatFits:]
  */
 - (CGSize)measure:(CGSize)constrainedSize;
```
appleguy: I think we should move the other -measure method here, and consider if updating this name to be more size-specific to clarify its usage compared to the layout-returning version.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -89,9 +70,17 @@ + (void)initialize
...
+  ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(calculatedLayout)), @"Subclass %@ must not override calculatedLayout method", NSStringFromClass(self));
   ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(measure:)), @"Subclass %@ must not override measure method", NSStringFromClass(self));
+  ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(measureWithSizeRange:)), @"Subclass %@ must not override measureWithSizeRange method", NSStringFromClass(self));
   ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(recursivelyClearContents)), @"Subclass %@ must not override recursivelyClearContents method", NSStringFromClass(self));
   ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(recursivelyClearFetchedData)), @"Subclass %@ must not override recursivelyClearFetchedData method", NSStringFromClass(self));
+
+  // At most one of the three layout methods is overridden
+  ASDisplayNodeAssert((ASDisplayNodeSubclassOverridesSelector(self, @selector(calculateSizeThatFits:)) ? 1 : 0)
```
appleguy: This is a fun way to express this condition — I like the style of ternary usage! :)

AsyncDisplayKit/Private/ASInternalHelpers.h
```diff
@@ -0,0 +1,28 @@
...
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <UIKit/UIKit.h>
+#import "ASBaseDefines.h"
+
+@class ASLayoutChild;
```
appleguy: Thanks for creating ASInternalHelpers.h and ASLayoutSpecUtilities.h!  You can probably remove this ASLayoutChild declaration, and do we need the UIKit import?

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: Thanks. Not sure if it's the best way to do, but it's quite easy to understand (at least to me) :)

AsyncDisplayKit/ASDisplayNode.h
nguyenhuy: Will move the other -measure method.

Updating this name will make yet another breaking API change. Maybe we should leave it for now? we may end up remove CGSize methods altogether later on.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
nguyenhuy: Yes, will move it to ASDisplayNode.h

AsyncDisplayKit/Private/ASInternalHelpers.h
nguyenhuy: :+1:  I forgot to clean this up!

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1282,19 +1281,51 @@ - (void)_tearDownPlaceholderLayer
...
+  ASDisplayNodeAssertThreadAffinity(self);
+  if (_methodOverrides & ASDisplayNodeMethodOverrideCalculateSizeThatFits) {
+    CGSize size = [self calculateSizeThatFits:constrainedSize.max];
+    return [ASLayout newWithLayoutableObject:self size:ASSizeRangeClamp(constrainedSize, size)];
+  } else {
+    id<ASLayoutable> layoutSpec = [self layoutSpecThatFits:constrainedSize];
+    ASLayout *layout = [layoutSpec measureWithSizeRange:constrainedSize];
+    // Make sure layoutableObject of the root layout is `self`, so that the flattened layout will be structurally correct.
+    if (layout.layoutableObject != self) {
+      layout.position = CGPointZero;
```
appleguy: This condition is interesting to me.  Do we expect it to happen without programmer error?  Can we add a basic assertion in the condition if we don't expect it to happen, but subclasses might accidentally trigger it from improper implementation?  In this case we still want to keep the fallback behavior because such conditions might only be encountered for the production usage of an app.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1282,19 +1281,51 @@ - (void)_tearDownPlaceholderLayer
...
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   ASDisplayNodeAssertThreadAffinity(self);
   return CGSizeZero;
 }
 
+- (id<ASLayoutable>)layoutSpecThatFits:(ASSizeRange)constrainedSize
+{
+  ASDisplayNodeAssertThreadAffinity(self);
+  return nil;
```
appleguy: I'm not sure if it would be necessary to enter here, but I think if a root class ASDisplayNode instance responded to a -measure: call by returning a layout with some size in between the min and max.  This would make it easier to manage nodes created to show a .backgroundColor layer or highlighting line.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1377,6 +1408,19 @@ - (void)recursivelyClearFetchedData
 - (void)layout
 {
   ASDisplayNodeAssertMainThread();
+  
+  if (!_flags.isMeasured) {
```
appleguy: What should this method do if a superclass calls [super layout]?  Should this logic be run for a class that implements manual layout?

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: This condition actually will happen most (if not all) of the time people if a layout spec is returned (via -layoutSpecThatFits:). For example KittenNode returns a layout spec with ASInsetLayoutSpec as layoutable object of the root layout.

There are 2 solutions for this problem:
- -layoutSpecThatFits: requires users to wrap his layout spec inside an umbrella spec that have valid root layoutable object.
- Automatically wrap the calculated layout, which is what I'm doing.

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: That's actually an interesting idea. I guess returning a layout spec with size equals to min size would make more sense.

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: Class that implements manual layout should have a complete implementation of -layout and should not call this default implementation. 

Class that uses automatic layout can safely rely on this method. (However, it can still override -layout and call [super layout] if it wants to add a custom layout logic, which is a very interesting idea in itself. Example can be found in KittenNode)

AsyncDisplayKit/Layout/ASLayout.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+@property (nonatomic, readonly) NSArray *sublayouts;
+
++ (instancetype)newWithLayoutableObject:(id<ASLayoutable>)layoutableObject
+                                   size:(CGSize)size
+                               position:(CGPoint)position
+                             sublayouts:(NSArray *)sublayouts;
+
+/**
+ * Convenience that has CGPointNull position.
```
appleguy: Is it correct to say that all "root layouts" should have a Null position?  And also that all non-root layouts should have a non-null position?

It would be helpful to clarify the use cases of these methods with changes to the comments, since otherwise it is tougher to get a sense of when these methods should be used.  For example, the conveniences that provide Null position are probably best used by ASDisplayNode subclasses that are manually creating a layout for -calculateLayoutThatFits:, or for ASLayoutSpec subclasses that are referencing the "self" level in the layout tree.

AsyncDisplayKit/Layout/ASStackLayoutChild.h
```diff
@@ -0,0 +1,22 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+/**
+ Each child may override their parent stack's cross axis alignment.
+ @see ASStackLayoutAlignItems
+ */
+typedef NS_ENUM(NSUInteger, ASStackLayoutAlignSelf) {
```
appleguy: Most of the contents of this header was deleted, right?  Does it make sense to put it internal to the stack objects, or is there an unobtrusive header it could move to?  If we keep the separate file, let's name it ...Defines.h instead of ...Child.h.

AsyncDisplayKit/Layout/ASLayout.h
nguyenhuy: Yes, it is correct. I will write documentation for these methods.

AsyncDisplayKit/Layout/ASStackLayoutChild.h
appleguy: Yeah, reading the next file, it will actually make the stack system more clear to put this next to the AlignItems definition.  Improve the comment to explain that the property you set this on is defined on ASDisplayNode as well as ASLayoutSpec, so that the children of the stack layout (whether nodes or sublayouts) can also participate in how the stacking is performed.

AsyncDisplayKit/Layout/ASStaticLayoutSpec.h
```diff
@@ -0,0 +1,46 @@
...
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <AsyncDisplayKit/ASLayoutSpec.h>
+#import <AsyncDisplayKit/ASStaticLayoutSpecDimension.h>
+
+@interface ASStaticLayoutSpecChild : NSObject
```
appleguy: Is this the ideal way to create a manual layout spec that allows full control of positioning the elements, but allows the subclass to have no custom implementation of -layout?  I wonder if there is a shorter / simpler name for this.  It probably makes more sense to name ASManualLayoutSpec, or at least clarify comments referring to some use cases.

AsyncDisplayKit/Layout/ASStackLayoutChild.h
nguyenhuy: But `ASStackLayoutAlignSelf` is required in ASLayoutable, which is imported by layout specs and especially ASDisplayNode. So if we put this type in ASStackLayoutSpec.h, other stack-related types will be imported as well, and they are irrelevant unless stack layout is used.

AsyncDisplayKit/Layout/ASStaticLayoutSpecDimension.h
```diff
@@ -0,0 +1,73 @@
...
+#import <AsyncDisplayKit/ASDimension.h>
+
+/** 
+ Expresses a size with relative dimensions.
+ Used by ASStaticLayoutSpec.
+ */
+typedef struct {
+  ASRelativeDimension width;
+  ASRelativeDimension height;
+} ASRelativeSize;
```
appleguy: Please rename this file to ASRelativeSize.h.  It almost belongs in ASDimension.h, but we can leave it in a separate file for readability — but it seems much more general than ASStaticLayoutSpec.  Nice that the comment mentions that usage area as an example though!

AsyncDisplayKit/Layout/ASStaticLayoutSpec.h
nguyenhuy: There is a [follow-up issue](https://github.com/facebook/AsyncDisplayKit/issues/539) for this. I will probably make some sample usages of this spec.

AsyncDisplayKit/Layout/ASStaticLayoutSpecDimension.h
nguyenhuy: I created this separated header for those size types because they are only used in static layout spec. So if one doesn't use the spec, one won't need to bother!

nguyenhuy: Yeah, the new name is better. Will change :+1:

appleguy:
Thanks for everything, @nguyenhuy !  It's been really awesome working with you on a feature so huge that it forms the basis for what will be AsyncDisplayKit 2.0.  We'll take up the last round of discussions in ongoing PRs and Issues.

nguyenhuy:
Thanks @appleguy for working with me on this. So happy that this PR eventually gets merged! :dancer: :dancers: :tada:

nguyenhuy:
Thanks everyone else as well :+1: :sunglasses:

markmark1:
Any examples in how to use this 😎😇

nguyenhuy:
Great question, @markmark1. Thanks for asking! 

The Kittens sample was updated to include both manual and automatic layout codes, which I think is a good starting point for new users of this box model. In addition, there are many snapshot test cases of layout specs (excepts ASStaticLayoutSpec) that I think are very good examples. That said, improving the documentation and/or examples is high on our priority list. And if you want to help us on this, please do!

markmark1:
thanks huy

PR #478, opened 2015-06-11
Fixed crash in ASNetworkImageNode.

victormayorov:


appleguy:
Thank you for the quick turnaround @victormayorov !

Issue #471, opened 2015-05-31
How to use -shouldRasterizeDescendants?


joostvandijk:
I'm trying to use -shouldRasterizeDescendants, but I keep ending up with an empty node. What properties should I set on my subnodes to get them to correctly rasterize?

This is my current node hierarchy. None of the subnodes change after creation. Text is also immediately set on the textnodes.

```
<<ASDisplayNode> isLayerBacked:1 shouldRasterizeDescendants:1 backgroundColor:white>
 | <<ASDisplayNode> isLayerBacked:1>
 |  | <<ASTextNode> isLayerBacked:1>
 | <<ASDisplayNode> isLayerBacked:1>
 |  | <<ASTextNode> isLayerBacked:1>
 | <<ASDisplayNode> isLayerBacked:1>
 |  | <<ASTextNode> isLayerBacked:1>
```

When I disable rasterization, the node looks like this (as expected):
![ios simulator screen shot 31 mei 2015 12 10 18](https://cloud.githubusercontent.com/assets/8394738/7901543/1ee54f44-078e-11e5-871b-5b207dbbc19a.png)

Any help would be much appreciated. Thanks!

appleguy:
@joostvandijk Thanks for the report — I'm going to test rasterization now with the latest version of ASDK to make sure it is working correctly.  If so, we can start looking more closely at your code or node setup for any issues.  My hope is that once we find the issue, we can make the API more clear or provide explicit alerts / warnings if an important element is left out of the setup.

appleguy:
I was able to reproduce a similar issue, so I doubt it is your configuration.  I'm not sure how this may have failed, but am hoping the fix is straightforward.  With the knowledge that it's likely a framework issue, feel free to dive in, but I'll try to debug as soon as I get a chance.  WWDC week ahead means it might take a little while, though!

appleguy:
Fortunately was able to figure out what is most likely going on.  Background colors take on a particular significance with rasterization, and I believe it is interpreting [UIColor clearColor] as black for an opaque context.  You should be able to work around this by setting [UIColor whiteColor] on a few nodes in your hierarchy.

This behavior is unacceptably non-intuitive, so there is an important follow-up to improve either the behavior or documentation / developer-facing warnings.  Thanks again for your report and I hope to hear if the workaround is successful :)

joostvandijk:
[...]

Issue #318, opened 2015-02-19
shouldRasterizeDescendants / frame issue


wockerwalker:
When creating an `ASDisplayNode` and setting `shouldRasterizeDescendants` to `YES`, any change to the node's origin (that is, any origin not equal to [0, 0]) seems to consequently inset the node's rasterized contents by the new origin's components. For example: when setting the node's frame to [10, 10, width, height] and setting the `backgroundColor` to red, the node displays with black 'margins' on top and left that are 10u each. The red background doesn't kick in until [10, 10] within the node itself.

rnystrom:
It sounds like the node is opaque. Try changing `opaque = NO` and see if the issue still occurs. 

Also it sounds kind of odd that if the node is moved (e.g. changing the origin to {10, 10}) that there are still margins.

If you could provide a sample project with a reproduction I would be more than happy to check this out!

wockerwalker:
Good call! The issue is resolved if I set `opaque = NO`. I would have expected that to resolve the black margin issue, but would have expected the frame to still be messed up (e.g. 'missing' the chunks where the margins used to be). Why does setting the `opaque` flag fix both issues?

rnystrom:
Setting `opaque = NO` will be eventually passed through to the [drawing mechanics](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Private/ASDisplayNode%2BAsyncDisplay.mm#L181). If the backing store is set to be opaque, when drawing happens anything not covered by subnodes/views/layers/rastered content will be black.

secretiverhyme:
The opacity triggering black backgrounds makes sense, but the offset issue is a bug I've seen reports of before.  This isn't super high-priority but we should look into this.

appleguy:
This is now fixed by https://github.com/facebook/AsyncDisplayKit/pull/479.  Sorry for the long delay on this one.  Subtree precompositing has not been used by many folks outside of Facebook, and so this bug that didn't affect Paper's version of ASDK went unresolved for a while.  I'm starting to use precompositing in other projects and will make sure it doesn't regress.

PR #479, opened 2015-06-13
Fix offset issue for rasterized nodes 

joostvandijk:
This fixes an issue that causes rasterized nodes that have a non-zero origin to draw incorrectly, as seen in #471 and #318.

This is a screenshot of the issue. The second box has an offset of about 150 points, which causes the rasterization to inset the subnodes with 150 points.
e2376ef0-0cf7-11e5-8c51-66a886017eb9

I've fixed it by using a frame with a zero origin when the node that's being rasterized is the container node.

schermafbeelding 2015-06-13 om 21 14 22


appleguy:
Really happy to have this.  One of my favorite contributions from the community so far — the fix is simple but I am particularly happy you're using subtree precompositing and debugged this.  This diff will make it into 1.2.1.

PR #485, opened 2015-06-15
Fixed bounds in ASDisplayNode

victormayorov:
Bounds.origin shouldn't be affected by -[ASDisplayNode setFrame:] method. This affects many nodes, for example ASScrollNode scrolls to top after frame changing.


appleguy:
Thank you!  This code looks like it could use some cleanup refactoring to reduce the duplication in the if/else, but I will do that later.

PR #488, opened 2015-06-18
Clean up logic in setFrame: for clarity and compactness

appleguy:
Thanks @victormayorov for finding an issue in this code.  Here's a refactoring that reads more cleanly.  Does it still look correct?


appleguy:
It does pass tests locally.  Thanks for adding tests!  The Kittens app layout also looks correct, but that is hardly a stressful test :)

PR #491, opened 2015-06-20
Custom Collection view layout prep PR

RCacheaux:
Turns ASScrollDirection into Option set in preparation for supporting 2-axis collection view layouts. Clears test host target warnings.


appleguy:
Nice and clean — thanks @RCacheaux!

PR #495, opened 2015-06-21
Cleaned up Kittens sample app and added comments for clarity.

hannahmbanana:
In particular, the NSValue-boxed CGSize array used as a data model for the kittens
is generated with random sizes, for unclear reasons.  It turns out this is a workaround
to convince PlaceKitten to send us unique images.


appleguy:
Thanks @hannahmbanana!  This definitely improves the readability of the code.  Looking at it myself, it was surprisingly unclear before that the reason for generating random sizes was essentially to trick the backend service into providing different images...because we show them at the same size anyway :)

PR #497, opened 2015-06-22
Adds abstract ASLayoutController

RCacheaux:
ASCollectionViewLayoutController will share logic with ASFlowLayoutController, this adds an abstract default implementation for classes conforming to ASLayoutController.


AsyncDisplayKit/Details/ASAbstractLayoutController.mm
```diff
@@ -0,0 +1,97 @@
...
+@implementation ASAbstractLayoutController
+
+- (instancetype)init
+{
+  if (!(self = [super init])) {
+    return nil;
+  }
+  
+  _tuningParameters = std::vector<ASRangeTuningParameters>(ASLayoutRangeTypeCount);
+  _tuningParameters[ASLayoutRangeTypePreload] = {
```
appleguy: These are flipped — preload should have a broader scope than render :)

appleguy:
Thanks for the highly digestible diff sizes!  Makes sense to move the common stuff up in to a superclass.

I'm quite alarmed by the error that existed in the current code, with the render range being larger than preload...fortunately the preload range was only recently added :).  If you could flip those in an upcoming PR, that would be great!

PR #503, opened 2015-06-23
ASCollectionViewLayoutController

RCacheaux:
Replaces FlowLayoutController with CollectionViewLayoutController in ASCollectionView. The new layout controller leverages UICollectionViewLayout's layoutAttributesForElementsInRect in order to calculate index paths for items in AsyncDisplayKit ranges. UICollectionViewFlowLayout is still a requirement in order to test the waters. If this works well with existing clients, we can simply remove the UICollectionViewFlowLayout class assertion.

### DIFF
--- a/AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
+++ b/AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
@@ -1,0 +1,153 @@
+/* Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "ASCollectionViewLayoutController.h"
+
+#include <vector>
+
+#import "ASAssert.h"
+#import "ASCollectionView.h"
+#import "CGRect+ASConvenience.h"
+
+struct ASDirectionalScreenfulBuffer {
+  CGFloat positiveDirection; // Positive relative to iOS Core Animation layer coordinate space.
+  CGFloat negativeDirection;
+};
+typedef struct ASDirectionalScreenfulBuffer ASDirectionalScreenfulBuffer;
+
+ASDirectionalScreenfulBuffer ASDirectionalScreenfulBufferHorizontal(ASScrollDirection scrollDirection,
+                                                                    ASRangeTuningParameters rangeTuningParameters) {
+  ASDirectionalScreenfulBuffer horizontalBuffer = {0, 0};
+  BOOL movingRight = ASScrollDirectionContainsRight(scrollDirection);
+  horizontalBuffer.positiveDirection = movingRight ? rangeTuningParameters.leadingBufferScreenfuls :
+                                                     rangeTuningParameters.trailingBufferScreenfuls;
+  horizontalBuffer.negativeDirection = movingRight ? rangeTuningParameters.trailingBufferScreenfuls :
+                                                     rangeTuningParameters.leadingBufferScreenfuls;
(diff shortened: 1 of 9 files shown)


AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
```diff
@@ -0,0 +1,153 @@
...
+  ASScrollDirection scrollableDirections = [collectionView scrollableDirections];
+  
+  BOOL canScrollHorizontally = ASScrollDirectionContainsHorizontalDirection(scrollableDirections);
+  if (canScrollHorizontally) {
+    ASDirectionalScreenfulBuffer horizontalBuffer = ASDirectionalScreenfulBufferHorizontal(scrollDirection,
+                                                                                           rangeTuningParameters);
+    rangeBounds = asdk_CGRectExpandHorizontally(rangeBounds,
+                                                horizontalBuffer.negativeDirection,
+                                                horizontalBuffer.positiveDirection);
+    // Update bounds is at most 95% of the next/previous screenful and at least half of tuning parameter value.
```
appleguy: I don't completely understand this yet.  Why 95%?  Can this be better-documented?

AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
```diff
@@ -0,0 +1,153 @@
...
+@interface ASCollectionViewLayoutController () {
+  ASCollectionView * __weak _collectionView;
+  std::vector<CGRect> _updateRangeBoundsIndexedByRangeType;
+}
+@end
+
+@implementation ASCollectionViewLayoutController
+
+- (instancetype)initWithCollectionView:(ASCollectionView *)collectionView {
+  if (!(self = [super init])) {
```
appleguy: Thank you for even matching our init style!

AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
```diff
@@ -0,0 +1,153 @@
...
+                                               MIN(verticalBuffer.positiveDirection * 0.5, 0.95));
+  }
+
+  return {rangeBounds, updateBounds};
+}
+
+- (NSSet *)indexPathsForItemsWithinRangeBounds:(CGRect)rangeBounds collectionView:(ASCollectionView *)collectionView {
+  NSMutableSet *indexPathSet = [[NSMutableSet alloc] init];
+  NSArray *layoutAttributes = [collectionView.collectionViewLayout layoutAttributesForElementsInRect:rangeBounds];
+  for (UICollectionViewLayoutAttributes *la in layoutAttributes) {
```
appleguy: This is not the most efficient way.  Unless I'm missing something, just do return [NSSet setWithArray:]

appleguy:
Some small changes here, but I'm happy to merge those in a later diff.  They aren't reason to hold this one up.

Let's avoid deleting the existing flow layout controller until there is time to profile it in comparison to this approach.  I'd be happy to take on that profiling and update the code to swap in the special-case layout controller if it proves to be faster, which I suspect it might (although maybe not!).

AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
RCacheaux: @appleguy Yup, I can document this better and PR it up.

AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
RCacheaux: Dropping note here for future reference. The iteration is necessary because the item's index path is stored within the layout attributes object. 

In the future, we could define a method that layouts can optionally implement that returns just the index paths to avoid the extra iteration.

PR #504, opened 2015-06-23
Custom layout enabled. Removes UICollectionViewFlowLayout requirement.

RCacheaux:
I think we might want to hold off on merging this until we verify that the new layout controller works well with flow layout as a first step.


appleguy:
I'm going to pull this in because it's sufficiently obscure (master-only, not making a post about the support) that anyone who tries it out is very likely to know what they're getting into :).  Literally right now we are spinning up a test with the existing custom Pinterest UICollectionViewLayout, will let you know if we survive any resulting detonations.

PR #507, opened 2015-06-24
Include missing files in iOS Framework target

sebreh:
The newly added iOS framework (added for Carthage support) was missing a few files, so it failed to build. This PR modifies the target membership for those files to make the target build. Fix verified to work with Carthage 0.7.4.


nickvelloff:
Thanks for adding the framework target. When using in Swift it builds but there are many warnings. Perhaps not related and I'll file another issue if it is helpful. Thanks.

screenshot 2015-06-24 19 59 40

PR #511, opened 2015-06-25
Exclude ASDealloc2MainObject.h from main spec

jflinter:
Currently when using Cocoapods to install the latest version of AsyncDisplayKit using the `use_frameworks!` option, `ASDealloc2MainObject.h` is added to the headers build phase twice due to its inclusion in both the main podspec as well as the `ASDealloc2MainObject` subspec. This leads to a "Multiple build commands for output file" warning in Xcode.

In this case the fix is pretty simple - just exclude that header from the main spec (you're already doing this for `ASDealloc2MainObject.m`).

### DIFF
--- a/AsyncDisplayKit.podspec
+++ b/AsyncDisplayKit.podspec
@@ -25,7 +25,10 @@ Pod::Spec.new do |spec|
 
   # ASDealloc2MainObject must be compiled with MRR
   spec.requires_arc = true
-  spec.exclude_files = ['AsyncDisplayKit/Details/ASDealloc2MainObject.m']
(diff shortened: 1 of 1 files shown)


appleguy:
Thanks a bunch — really appreciate your attention to detail here!  I hate build warnings with the fiery passion of a thousand suns, but have not had a chance to look at any of these.  Glad you're part of the community :)

jflinter:
Totally understand - it's so hard to stay on top of everything that can go
wrong across every dependency manager/build tool. Thanks for the merge!

PR #513, opened 2015-06-25
Add UIResponder methods to ASDisplayNode

### LINKED ISSUE #11: [ASDisplayNode] Support UIMenuController
I found two identifiers that were only mentioned in comments but the types themselves aren't defined anywhere:

`ASDisplayNodeView`, seems like a protocol that any view usable in a async display node would need to implement, mentioned in `_ASPendingState.h:25`.

`ASDisplayNodeAsyncView`, seems like a subclass of `UIView` that would too be usable in a node (also the way of adding more async views by subclassing), mentioned in `ASDisplayNode.h:22`.

Could you please add more details on these types? Plus some sort of a documentation of how to add a custom async view class. Thanks!

jflinter:
This adds several `UIResponder` methods to `ASDisplayNode`, further increasing its API-compatibility with `UIView`. These methods include:
- `canBecomeFirstResponder`
- `becomeFirstResponder`
- `canResignFirstResponder`
- `resignFirstResponder`
- `isFirstResponder`
- `canPerformAction:withSender`

These by default just forward along to their underlying `_ASDisplayView`, which should already implement them correctly. The exception to this is `canBecomeFirstResponder` and `canResignFirstResponder`, which by default return `NO` and `YES`, respectively (this is so the `_ASDisplayView` can in turn ask its `ASDisplayNode` if it can become the first responder without having to expose `_ASDisplayView` directly).

I put this together to make it easy to close #11. However, the work's not fully done: it's still up to implementors to actually create a copyable `ASTextNode` subclass. This is intentional - that's pretty much the state of things with `UILabel` as well (see nshipster.com/uimenucontroller/ for an example of how to properly subclass and implement this). If you'd like me to contribute an `ASCopyableTextNode` class, I'd be happy to, although it feels a bit out-of-scope for this library.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -468,6 +468,15 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
 - (void)setNeedsDisplay;    // Marks the view as needing display. Convenience for use whether view is created or not, or from a background thread.
 - (void)setNeedsLayout;     // Marks the view as needing layout.  Convenience for use whether view is created or not, or from a background thread.
 
+// UIResponder methods
+// By default these fall through to the underlying view, but can be overridden.
+- (BOOL)canBecomeFirstResponder;                                            // default==NO
+- (BOOL)becomeFirstResponder;                                               // default==NO (no-op)
+- (BOOL)canResignFirstResponder;                                            // default==YES
+- (BOOL)resignFirstResponder;                                               // default==YES (no-op)
+- (BOOL)isFirstResponder;
+- (BOOL)canPerformAction:(SEL)action withSender:(id)sender;
+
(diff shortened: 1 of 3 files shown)


appleguy:
Ah, this is very interesting — I appreciate the investigation @jflinter .  I'd like to merge this and very much support the general direction, but have a couple concerns:

- Have you carefully considered what will happen for layer-backed or subtree-precomposited (rasterized) nodes?
- Can you think of a way we could evaluate the performance overhead of returning non-default values for the first responder properties?  It is a vague concern right now, but I worry that this could somehow increase the substantial UIKit overhead that is seen with -addSubview: and -removeFromSuperview as those operations scan the hierarchy to update the first responder.

Could you share one or two use cases that you've run into for these features?  I think we should also include at least a couple unit tests for this kind of functionality, to make sure it behaves in a consistent way with layer-backing, pre compositing, and the desired behavior itself.

Issue #458, opened 2015-05-21
Investigate completeness of visible row tracking.


appleguy:
See discussion in https://github.com/facebook/AsyncDisplayKit/pull/430.  Is there a way to avoid this manual tracking entirely, and is it certain that a single pending indexPath is sufficient for all cases?

appleguy:
@eanagel as it turns out, this can cause a very severe wedging of ASLayoutController...it's possible for the pending index to be disjoint from the other indices in a way that makes for an unsolvable set of constraints for the flow layout controller, which causes it to live-lock and burn as much memory and CPU as is available :(.

Do you have a reproduction case for the original issue?  I do fortunately have a reproduction case for the live-lock issue.

Here's some example output alongside the screenshot:

(lldb) p _scrollDirection
(ASScrollDirection) $0 = 4
(lldb) po visibleNodePaths
<__NSArrayI 0x7fc1c2f9a2d0>(
<NSIndexPath: 0xc000000000220056> {length = 2, path = 1 - 68},
<NSIndexPath: 0xc000000000228056> {length = 2, path = 1 - 69},
<NSIndexPath: 0xc000000000230056> {length = 2, path = 1 - 70},
<NSIndexPath: 0xc000000000038116> {length = 2, path = 4 - 7}
)


(lldb) po _delegate
<ASTableView: 0x7fc1ba815c00; baseClass = UITableView; frame = (0 0; 100 500); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x7fc1b9632dc0>; layer = <CALayer: 0x7fc1b962fe70>; contentOffset: {0, 36343}; contentSize: {100, 172800}>

(lldb) po [_delegate indexPathsForVisibleRows]
<__NSArrayM 0x7fc1bb16a380>(
<NSIndexPath: 0xc000000000220056> {length = 2, path = 1 - 68},
<NSIndexPath: 0xc000000000228056> {length = 2, path = 1 - 69},
<NSIndexPath: 0xc000000000230056> {length = 2, path = 1 - 70}
)

![screen shot 2015-06-21 at 4 56 55 pm](https://cloud.githubusercontent.com/assets/565251/8273962/0ad7ef7e-1838-11e5-8a8b-56bc64abc1db.png)

eanagel:
[...]

PR #514, opened 2015-06-25
ASTableView programmatic scrolling fix w/contiguous guarantee

eanagel:
This is a candidate fix for the race condition @appleguy tracked down in #458. This version uses the single pending `NSIndexPath` approach but adds code to validate that the pending index path is always contiguous with respect to `UITableView`'s `indexPathsForVisibleNodes`. (I had previously discovered that the approach of tracking all visible nodes using willDisplay/didEndDisplaying is flawed because `UITableView` doesn't always call didEndDisplaying. Boo. See #427)

In order to do this we may make up to two calls to `-[ASDataController  numberOfRowsInSection:]` but this *only* happens when we have a pending index that is not part of the existing visible range (a rare but important edge case.) From what I've seen this happens pretty rarely and should not impact overall system performance.

I took great pains to validate the logic for the `isAfter(...)` method, so I think we are in good shape there. (It even covers odd situations like sections with no rows.) One caveat, if the pending value is an invalid `NSIndexPath` for our dataSource for some reason, weird things can happen. If we need to cover this edge case as well it can be done, but it would require additional calls to `ASDataController` to validate it.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -522,25 +522,37 @@ void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
   if ( _pendingVisibleIndexPath ) {
     NSMutableSet *indexPaths = [NSMutableSet setWithArray:self.indexPathsForVisibleRows];
 
-    BOOL (^isNextTo)(NSIndexPath *, NSIndexPath *, int) = ^(NSIndexPath *indexPath, NSIndexPath *anchor, int dir) {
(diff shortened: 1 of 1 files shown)


appleguy:
I'm going to take this diff even though I only 80% understand it, because I trust your work :).  I'll be working more in this code this weekend and want to make sure I don't conflict with this, and also that my testing includes the change so it can be more fully validated.

Thanks @eanagel for being one of the very few people in the world willing to dive into the really tough parts of AsyncDisplayKit and help improve the framework for everyone!

PR #529, opened 2015-06-29
Overhaul ASDataController and extensively test ASTableView.

appleguy:
This diff resolves all known consistency issues with ASTableView and ASCollectionView.
It includes significantly more aggressive thrash-testing in ASTableViewStressTest,
which now passes on a variety of device and simulator configurations.  It also updates
the unit tests run on every commit to ensure any regression is caught quickly.

A few of the salient changes in this diff:
- ASTableView now uses Rene's ASCollectionViewLayoutController, and actually uses a
UICollectionViewFlowLayout without any UICollectionView.  This resolves an issue where
ASFlowLayoutController was generating slightly out-of-bounds indicies when programmatically
scrolling past the end of the table content.  Because the custom implementation is likely
faster, I will revisit this later with profiling and possibly returning to the custom impl.
- There is now a second copy of the _nodes array maintained by ASDataController.  It shares
the same node instances, but this does add some overhead to manipulating the arrays. I've
filed a task to follow up with optimization, as there are several great opportunities to
make it faster.  However, I don't believe the overhead is a significant issue, and it does
guarantee correctness in even the toughest app usage scenarios.
- ASDataController no longer supports calling its delegate /before/ edit operations.  No
other class was relying on this behavior, and it would be unusual for an app developer to
use ASDataController directly.  However, it is possible that someone with a custom view
that integrates with ASDataController and ASRangeController could be affected by this.
- Further cleanup of organization, naming, additional comments, reduced code length
wherever possible.  Overall, significantly more accessible to a new reader.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -10,7 +10,7 @@
 
 #import "ASAssert.h"
 #import "ASDataController.h"
-#import "ASFlowLayoutController.h"
+#import "ASCollectionViewLayoutController.h"
 #import "ASLayoutController.h"
 #import "ASRangeController.h"
 #import "ASDisplayNodeInternal.h"
@@ -117,7 +117,7 @@ static BOOL _isInterceptedSelector(SEL sel)
   _ASTableViewProxy *_proxyDelegate;
 
   ASDataController *_dataController;
-  ASFlowLayoutController *_layoutController;
+  ASCollectionViewLayoutController *_layoutController;
 
   ASRangeController *_rangeController;
 
@@ -159,7 +159,8 @@ void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
 
 - (void)configureWithAsyncDataFetching:(BOOL)asyncDataFetchingEnabled
 {
-  _layoutController = [[ASFlowLayoutController alloc] initWithScrollOption:ASFlowLayoutDirectionVertical];
+  UICollectionViewFlowLayout *flowLayout = [[UICollectionViewFlowLayout alloc] init];
+  _layoutController = [[ASCollectionViewLayoutController alloc] initWithScrollView:self collectionViewLayout:flowLayout];
 
   _rangeController = [[ASRangeController alloc] init];
   _rangeController.layoutController = _layoutController;
@@ -412,20 +413,12 @@ void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
 {
   CGPoint scrollVelocity = [self.panGestureRecognizer velocityInView:self.superview];
   ASScrollDirection direction = ASScrollDirectionNone;
-  if (_layoutController.layoutDirection == ASFlowLayoutDirectionHorizontal) {
-    if (scrollVelocity.x > 0) {
-      direction = ASScrollDirectionRight;
-    } else if (scrollVelocity.x < 0) {
-      direction = ASScrollDirectionLeft;
-    }
+  if (scrollVelocity.y > 0) {
+    direction = ASScrollDirectionDown;
   } else {
-    if (scrollVelocity.y > 0) {
-      direction = ASScrollDirectionDown;
-    } else {
-      direction = ASScrollDirectionUp;
-    }
+    direction = ASScrollDirectionUp;
   }
-
+  
   return direction;
 }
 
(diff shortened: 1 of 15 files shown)


AsyncDisplayKit/ASTableView.mm
```diff
@@ -159,7 +159,8 @@ void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
 
 - (void)configureWithAsyncDataFetching:(BOOL)asyncDataFetchingEnabled
 {
-  _layoutController = [[ASFlowLayoutController alloc] initWithScrollOption:ASFlowLayoutDirectionVertical];
+  UICollectionViewFlowLayout *flowLayout = [[UICollectionViewFlowLayout alloc] init];
```
RCacheaux: Super interesting that flow layout works without a collection view! Does this flow layout need to know the sizes of the table view cells? Flow layout does have an itemSize property. I'm wondering if this will only work for table views with uniform cell size since flow layout only supports custom cell sizing via collection view delegate and we don't have a collection view to attach a delegate.

If we find that using flow layout isn't sufficient for ASTableView and we want to use ASCollectionViewLayoutController, I think we could build a custom collection view layout that uses a table view to calculate layout attributes. I imagine most of that logic already lives in ASFlowLayoutController. The trick would be coordinating table view data source changes with the layout.

nguyenhuy:
It's great to see you're using a queue-based approach for edit operations/blocks. I do believe that this approach can guarantee data consistency, even under very high load.

I'm curious to know your though on inter-operations optimization, such as collapsing multiple reloads on the same indices, or drop an insert operation on cells that will be removed later on, etc... I believe such optimization will (only?) be achieved if each operation is atomic. That's said, no optimization is worth doing without hard evidence and if not implemented correctly, the processing overhead would cancel out any performance gain.

Another interesting idea is to prioritize operations. For example, a reload operation on cells within visible and/or working ranges might be added to the front of the pending queue.

nguyenhuy:
And needless to say, your solution to fix these known issues is just elegant :+1:

markmark1:
Thanks when is the next release wanna try it out

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -21,16 +21,16 @@
...
 @interface ASDataController () {
-  NSMutableArray *_nodes;
-  NSMutableArray *_pendingBlocks;
+  NSMutableArray *_completedNodes;            // Main thread only.  External data access can immediately query this.
+  NSMutableArray *_editingNodes;              // Modified on _editingTransactionQueue only.  Updates propogated to _completedNodes.
+  
+  NSMutableArray *_pendingEditCommandBlocks;  // To be run on the main thread.  Handles begin/endUpdates tracking.
+  NSOperationQueue *_editingTransactionQueue; // Serial background queue.  Dispatches concurrent layout and manages _editingNodes.
+  
   BOOL _asyncDataFetchingEnabled;
```
eanagel: When we optimize, maybe use bit flags for these?

appleguy:
@nguyenhuy thanks for your thoughts — helpful!

Indeed this amounts to a first pass architectural improvement for the ASDataController aimed at 100% correctness first, even in the most extreme circumstances, with optimization scaling in later.

The visible-range-aware prioritization would be great; indeed we could adjust the size of the layoutGroups depending on, say, the size of the render range rather than a number determined only by a multiple of the CPU core count.

We should also implement the ability to not lay out all nodes in the data source, and notify UIKit (underlying UITableView / UICollectionView / even the ASRangeController) only of a subset of the total nodes in the asyncDataSource.  This would significantly reduce the number of nodes that are created and sized if an app has a fairly large local data source with hundreds of rows.  In a way, this would be like an "allocation + layout range" — similar in principle to the Render and Preload ranges.  However, it is conceptually different in that a node should not be available to enter any other ASRangeHandler until it has gotten past the allocation + layout step.  After thinking about this for a while I'm convinced that keeping this functionality in ASDataController is the right thing, not extending ASRangeController.  It is however possible that we would want to use or extend the idea of an ASRangeHandler, or even abstract parts of ASRangeController, for ASDataController to drive its decision and processing of which nodes to allocate and lay out.

I think there's a lot of impact to be had pursuing these increasingly sophisticated takes on ASDataController, and they are probably in the top 3 "major projects" that should be undertaken in the next 6 months for ASDK.

appleguy:
Thanks @GSatko for flagging that!  I think it is appropriate to remove the QoS usage here as it is unlikely to be of material importance to how this code executes, even on iOS 8.

PR #535, opened 2015-06-30
Propagate ASCellNode's clipsToBounds value to the __ASTableViewCell

eanagel:
This helps work around a bug we are seeing in some rare cases (selected background view overlaps other cells if size of ASCellNode has changed.)

There is an edge case I ignored for this PR - if you set ASCellNode clipsToBounds while the cell is currently visible it will not propagate to the Cell until it is next displayed. (Just because it doesn't seem worth the added complexity.)

@appleguy I've tested this with my ADKFlicker test and my internal project and it masks the issue in both cases.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.m
+++ b/AsyncDisplayKit/ASCellNode.m
@@ -25,6 +25,7 @@
 
   // use UITableViewCell defaults
   _selectionStyle = UITableViewCellSelectionStyleDefault;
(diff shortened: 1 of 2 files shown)


appleguy:
Thanks @eanagel!  I think this is a safe workaround in the short term for this unusual issue, which I'll dig into with priority after the most important ASDataController & layout architecture changes have landed smoothly.

Issue #541, opened 2015-07-07
Duplicate symbols on ASIndexPath


grigaci:
Global functions implemented in **ASIndexPath.h** causes duplicate symbol errors at compile time.

I have reproduced the problem within a complicated project, but a much simpler way would be to import **ASIndexPath.h** in two different files and just create ASIndexPath instances in both files. A project that demos this case can be found [here](https://www.dropbox.com/s/80lue19uvp0qyun/ASDKExample.zip?dl=0).
<img width="724" alt="asindexpath_duplicate_symbols" src="https://cloud.githubusercontent.com/assets/1185635/8550312/7e48b2b6-24d8-11e5-82bf-3bc972d12e8c.png">

grigaci:
Submitted pull request [542](https://github.com/facebook/AsyncDisplayKit/pull/542) with the fix.

appleguy:
Thanks, merged!  This will only happen when a file outside of ASFlowLayoutController tries to use ASIndexPath, right?  I ask to make sure I understand the scope of this impacting people using the framework — it will be a bit challenging to do a new stable release (e.g. 1.2.3) because we have now landed the flex box layout system, which means we're now focusing on 2.0 Beta.

grigaci:
Actually, I had reproduced the problem on a more complicated scenario. I have two projects, first one written in Objective C & Swift and I integrate it as a pod into the second project. The generated Swift bridging header file imports AsyncDisplayKit as module. I had referenced this file in several other files, which caused duplicated symbol errors.

grigaci:
@appleguy when gets the first beta released?

appleguy:
@grigaci there are a couple small issues being found with the box model layout support, so I'm thinking we should wait until about mid next-week for the developers using master to try out the new features and flag / fix any big problems.  Then we can push out ASDK 2.0 Beta 1.  However, it's not all that important to have a "Beta" release as opposed to just using master; one of the main reasons is as an opportunity to post about the features an goals of 2.0 in discussion forums, and get more people looking at it.  Since you already know, feel free to continue using master to benefit from the latest fixes and we should have more folks joining us next week when the announcement happens.

PR #542, opened 2015-07-07
Issue #541 Fix duplicate symbols on ASIndexPath

grigaci:


appleguy:
Thank you @grigaci !  As you can tell, I was in a rush writing this and threw it into a file after having developed it in-situation for ASFlowLayoutController.  Nice to have this change.

PR #543, opened 2015-07-08
Implemented test for crash in ASCollectionView on reloadData

victormayorov:
Test to show how ASCollectionViewController may crash. Looks like Apple bug. Simplest possible fix in ASDK is to modify ASCollectionView.mm:
```
- (void)reloadDataWithCompletion:(void (^)())completion
{
  ASDisplayNodeAssert(self.asyncDelegate, @"ASCollectionView's asyncDelegate property must be set.");
  ASDisplayNodePerformBlockOnMainThread(^{
    [super reloadData];
    [self layoutIfNeeded]; // Here is the FIX
  });
  [_dataController reloadDataWithAnimationOptions:kASCollectionViewAnimationNone completion:completion];
}
```


appleguy:
Very cool, thanks for creating this new Tests file.  I have several ASCollectionView tests locally that I want to upload soon, both an examples project and adapting some of that code to run as part of the unit test target on every commit.

Any reason you chose to commit this DISABLED_ without including your -layoutIfNeeded fix?  Do you have a sense for what is causing this and if the workaround / fix would need to be put in other methods besides -reloadData, such as -reloadSections or -reloadRows, or any other editing methods?

victormayorov:
@appleguy I don't like the workaround, may be you can find another solution? But this is the only way, which I know for now. It looks like Apple bug, and I haven't checked it on iOS 7. It means that the workaround theoretically can be used only for iOS 8, or all iOS older than 7. I need to investigate it. I'll do it in nearest days, and now you have possibility to look at it also.

PR #545, opened 2015-07-08
fixed retain cycle between display node and its stored layout

ryanfitz:


nguyenhuy:
Thank you for working on this, @ryanfitz. This is definitely a embrassing mistake of mine! I do agree with you that layoutableObject should not be "strong" and second this PR :+1:

appleguy:
Thanks for the fix!  We're really excited to get the community looking at the new changes in master so that we can together converge to a stable and awesome AsyncDisplayKit 2.0.  In fact, I'm going to hold off on cutting the ASDK 2.0 Beta 1 tag / branch until we have at least another week of folks poking around and flushing out a few of the inevitable issues like this!  Really appreciate it :)

PR #548, opened 2015-07-09
Automatically adjust content offset for insertion/deletion

eanagel:
This PR adds a new property for ASTableView (automaticallyAdjustsContentOffset) that, when enabled, will adjust the contentOffset for inserts and deletes before the visible viewport so the users view is maintained. In order to use this effectively, animations must be disabled while performing updates. To facilitate this an option has been added to endUpdates (and also a completion handler that has many uses.) Because this only works well with animations are disabled and it may require some thought to use effectively, it defaults to off.

You can see this in action with a simple demo app at: https://github.com/eanagel/ASDKVirtualizationDemo

There is a performance trade-off for this feature that I'd like anyone to weigh in on. In order to make this work, we need to know the heights of the nodes being inserted or deleted. (well we actually only need to know the heights of the ones before the visible cells.) I've extended our delegates to pass that information own to the ASTableView, but if it's not used it is a waste. These data is pretty easily available for inserting nodes, but in the case of deleting nodes, there is the additional overhead of capturing the deleted items and passing them down. There are alternative approaches I'm sure we could work out of this seems like it could be a big performance hit.

Additionally, this includes a **speculative** fix for a problem where ASDataController appears to be calling it's endUpdates delegate method before the didInsert/Delete delegate methods. The fix included here works for my tests but it does **not** pass the ASTableViewStressTest, so there may be more to the story here. I bow to @appleguy's wisdom on this one. See issue #564 for more details.

I think this handles the common cases for virtualization, but there's more that could be added to it to make it even more robust. Right now, it doesn't handle insertion/deletion of sections and you must explicitly disable animations in endUpdate to make it work correctly. If we wanted to make things more automatic we could separate the items that would be inserted before the visual window and make those insertions/deletions without animations (in a separate begin/end update block) and then follow up with any other updates in an animated fashion. It's considerably more work, but that would make this an "it just works" feature which I like ;)

Attempts to resolve issue #564, which must work correctly for this code to function.


appleguy:
Retrying the build; thanks so much for this work!

AsyncDisplayKit/ASTableView.h
```diff
@@ -133,6 +134,14 @@
  */
 - (NSArray *)visibleNodes;
 
+/**
+ * YES to automatically adjust the contentOffset when cells are inserted or deleted "before"
+ * visible cells, maintaining the users' visible scroll position. Very useful when supporting virtualization.
```
appleguy: Expand comment to describe "virtualization", or let's leave it out.  "Data source virtualization" is probably a good way to refer to it, but even with the extra words, I think more than half of developers will not really know what this means.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -374,6 +396,52 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
...
+  _contentOffsetAdjustmentTop = nil;
+}
+
+- (void)adjustContentOffsetWithNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths dir:(int)dir {
+  ASDisplayNodeAssert(_automaticallyAdjustsContentOffset, @"this method should only be called when _automaticallyAdjustsContentOffset == YES");
+
+  CGFloat adjustment = 0;
+  NSIndexPath *top = _contentOffsetAdjustmentTop ?: self.indexPathsForVisibleRows.firstObject;
+
+  for(int index=0; index<indexPaths.count; index++) {
```
appleguy: Add some spacing around operators and parens to match surrounding style

AsyncDisplayKit/ASTableView.mm
```diff
@@ -374,6 +396,52 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
...
+    NSIndexPath *indexPath = indexPaths[index];
+    if ([indexPath compare:top] <= 0) {
```
appleguy: Worth a brief comment on the logical function of this check

AsyncDisplayKit/ASTableView.mm
```diff
@@ -374,6 +396,52 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
...
+  ASDisplayNodeAssert(_automaticallyAdjustsContentOffset, @"this method should only be called when _automaticallyAdjustsContentOffset == YES");
+  if (_contentOffsetAdjustment != 0) {
+    self.contentOffset = CGPointMake(0, self.contentOffset.y+_contentOffsetAdjustment);
+  }
+
+  _contentOffsetAdjustment = 0;
+  _contentOffsetAdjustmentTop = nil;
+}
+
+- (void)adjustContentOffsetWithNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths dir:(int)dir {
```
appleguy: "dir" property and its possible values are unclear.  I think there are at least a couple existing data types for scroll direction that we should use, even though you would need one more line below before being able to multiply against it.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -335,7 +336,9 @@ - (void)endUpdatesWithCompletion:(void (^)(BOOL))completion
   _batchUpdateCounter--;
 
   if (_batchUpdateCounter == 0) {
-    [_delegate dataControllerBeginUpdates:self];
+    [_editingTransactionQueue addOperationWithBlock:^{
+      [_delegate dataControllerBeginUpdates:self];
```
appleguy: I've been out of this code for a week, but I'm nearly positive this will result in calling the delegate on a background queue, which is not currently intended / is likely to break the threading model currently being used in this class.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -343,7 +346,11 @@ - (void)endUpdatesWithCompletion:(void (^)(BOOL))completion
     }];
     [_pendingEditCommandBlocks removeAllObjects];
 
-    [_delegate dataControllerEndUpdates:self completion:completion];
+    // Queue the endUpdates call in our editing queue so it is guaranteed to happen after our edits.
+
+    [_editingTransactionQueue addOperationWithBlock:^{
+        [_delegate dataControllerEndUpdates:self completion:completion];
```
appleguy: Too much indentation, but same issue as above

appleguy:
Thanks for working on this!  It's really a great feature for any clients that do a material amount of editing with a table view.  I would love to merge the feature as soon as it is in a state where it is unlikely to cause confusion for clients, and it definitely needs to pass the existing tests.

Assuming that it is not blocking the functionality of this diff, let's separate the fix for the delegate method call order in the batched update case.  Does a task currently exist for this (aside our email thread about it?)  I'll try my best to investigate it tomorrow, as I suspect it is a pretty solvable problem with careful inspection.

AsyncDisplayKit/ASTableView.h
eanagel: good point, removed.

AsyncDisplayKit/ASTableView.mm
eanagel: good point, clarified.

AsyncDisplayKit/Details/ASDataController.mm
eanagel: Ahh yes I see that now. I got a little confused because ASRangeController also re-wraps it's delegate calls in a main thread block so I thought ASDataControllerDelegate was a background thing. Fixed.

AsyncDisplayKit/Details/ASDataController.mm
eanagel: derp. got it.

eanagel:
I've made the updates you've suggested, thanks for spending some time really looking through it.

Unfortunately the fix for issue #564 (which I just added) is required in order for this to function correctly. Otherwise the endUpdates call comes before the editing calls, so, well nothing really works correctly. ;) The more I look at this fix, the more comfortable I'm becoming with it as a solution. I wonder if the tests could be wrong (they were passing with a pretty significant bug.) I'll take a look at that and see what I can figure out.

eanagel:
It appears the problem we are hitting in the tests here may be the same as #566, perhaps the begin/end change I made somehow makes it more prevalent?

eanagel:
Recent Commits include an important bug fix for ASRangeController and an improvement to `-endUpdatesAnimated:completion:` that handles an edge case that could leave animations disabled globally.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -326,9 +333,14 @@ - (void)beginUpdates
 
 - (void)endUpdates
 {
-  [_dataController endUpdates];
+  [self endUpdatesAnimated:YES completion:nil];
 }
 
+- (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL completed))completion;
+{
+  _endUpdatesAnimationsEnabled = animated;
```
eanagel: In thinking about this now, it seems it's possible for this value to be overwritten by overlapping calls to endUpdates. The best option would be to pass the animated value through the delegates, @appleguy do you think that's the way to go or do you have another suggestion? Turns out the simple approach of using the completion block is not 100% reliable and can lead to animations becoming disabled globally.

appleguy:
@eanagel Thanks for all the updates here.  It's of material service to the community, and indeed these fixes will be directly valuable to the ASDK adoption underway at Pinterest — that allows me to spend more of my time on new features that I think will be valuable to your team :).

After a busy week unable to focus on the dynamics to this patch (particularly as it relates to the bug in my testing code and resulting problem cases), I'm now ramping up to ensure we get your diffs landed this weekend.  Appreciate your patience as I catch up.

Could you rebase the diff?  Any remaining issues or tricky areas you want to call out, besides the animation value pass-through?

AsyncDisplayKit/ASTableView.h
```diff
@@ -94,6 +94,7 @@
  */
 - (void)beginUpdates;
 - (void)endUpdates;
+- (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL completed))completion;
```
appleguy: I like how you've structured this.  A single more detailed option that provides the completion, even if you're not interested in animation (just pass NO).  Worth an AppleDoc comment with a brief explanation of the semantics of the completion?

AsyncDisplayKit/ASTableView.mm
```diff
@@ -126,6 +126,11 @@ @interface ASTableView () <ASRangeControllerDelegate, ASDataControllerSource> {
   ASBatchContext *_batchContext;
 
   NSIndexPath *_pendingVisibleIndexPath;
+
+  NSIndexPath *_contentOffsetAdjustmentTop;
```
appleguy: This name is slightly unclear but tough to make more clear without being super long.  Could you add a small comment that explains its purpose?  E.g. "Set to the topmost fully-visible index path before executing editing commands, if automaticallyAdjustsContentOffset is enabled. "

eanagel:
Awesome! I'm in the process of making changes to endEditsAnimated:completion: you suggested (and pushing the animated parameter through the delegates). Once I'm done I'll rebase and create a new PR. I may go ahead and split out the bug fixes from the features into separate PR's to make it easier to see the changes.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -79,6 +79,30 @@
 @property (nonatomic, assign) CGFloat leadingScreensForBatching;
 
 /**
+ *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. You can call it from background 
+ *  thread (it is recommendated) and the UI collection view will be updated asynchronously. The asyncDataSource must be updated
```
appleguy: Typo - recommendated :).  Let's just call it UICollectionView.  (or "collection view").

Can this really be called on a background thread?  Because we are aiming for 100% correctness / no test failures in any permutation of client usage, I think we should make sure editing methods are main thread only.

I'm really glad you are adding the batch feature to ASCollectionView, but this might suggest it is time to break up the diff (specifically, adding more newly-designed APIs will make it harder to get this merged with confidence).  Very interested to hear your thoughts on the threading model here and the initial motivation for it to be different than all existing edit methods.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -79,6 +79,30 @@
 @property (nonatomic, assign) CGFloat leadingScreensForBatching;
 
 /**
+ *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. You can call it from background 
+ *  thread (it is recommendated) and the UI collection view will be updated asynchronously. The asyncDataSource must be updated 
+ *  to reflect the changes before this method is called.
+ *
+ *  @param animated   NO to disable animations for this batch
```
appleguy: Could you add some additional documentation description for why the animated property exists?  Wouldn't the individual update blocks specify any animation?  If so, could we delete that variant?

PR #550, opened 2015-07-09
Access to underlying UITextView for more configuration options.

### LINKED ISSUE #549: ASEditableTextNode - add setter/getter for underlying UITextView for customisation
Will be very useful for further customisation - currently unable to change keyboard type / appearance, data detector type etc.

marmelroy:
This allows much more control over the display of ASEditableTextNodes and closes #549

### DIFF
--- a/AsyncDisplayKit/ASEditableTextNode.h
+++ b/AsyncDisplayKit/ASEditableTextNode.h
@@ -19,7 +19,10 @@
 
 #pragma mark - Configuration
 
-//! @abstract Access to underlying UITextView for more configuration options.
+/**
+  @abstract Access to underlying UITextView for more configuration options.
+  @warning This property should only be used on the main thread and should not be accessed before the editable text node's view is created.
+ */
 @property (nonatomic, readwrite, strong) UITextView *textView;
 
 //! @abstract The attributes to apply to new text being entered by the user.
--- a/AsyncDisplayKit/ASEditableTextNode.h
+++ b/AsyncDisplayKit/ASEditableTextNode.h
@@ -19,6 +19,9 @@
 
 #pragma mark - Configuration
 
+//! @abstract Access to underlying UITextView for more configuration options.
+@property (nonatomic, readwrite, strong) UITextView *textView;
+
 //! @abstract The attributes to apply to new text being entered by the user.
 @property (nonatomic, readwrite, strong) NSDictionary *typingAttributes;
 
--- a/AsyncDisplayKit/ASEditableTextNode.h
+++ b/AsyncDisplayKit/ASEditableTextNode.h
@@ -23,7 +23,7 @@
   @abstract Access to underlying UITextView for more configuration options.
   @warning This property should only be used on the main thread and should not be accessed before the editable text node's view is created.
  */
-@property (nonatomic, readwrite, strong) UITextView *textView;
+@property (nonatomic, readonly, strong) UITextView *textView;
 
 //! @abstract The attributes to apply to new text being entered by the user.
 @property (nonatomic, readwrite, strong) NSDictionary *typingAttributes;
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -41,6 +41,7 @@
 {
   @private
   // Configuration.
+  UITextView *_textView;
   NSDictionary *_typingAttributes;
 
   // Core.
@@ -134,7 +135,7 @@
   [self.view addSubview:_placeholderTextKitComponents.textView];
 
   // Create and configure our text view.
-    _textKitComponents.textView = _textView;
+  _textKitComponents.textView = _textView;
   //_textKitComponents.textView = NO; // Unfortunately there's a bug here with iOS 7 DP5 that causes the text-view to only be one line high when scrollEnabled is NO. rdar://14729288
   _textKitComponents.textView.delegate = self;
   _textKitComponents.textView.editable = YES;
@@ -188,6 +189,21 @@
 
 #pragma mark - Configuration
 @synthesize delegate = _delegate;
+
+#pragma mark -
+@dynamic textView;
+
+- (UITextView *)textView{
+  ASDisplayNodeAssertMainThread();
+  return _textView;
+}
+
+- (void)setTextView:(UITextView *)textView
+{
+  ASDisplayNodeAssertMainThread();
+  _textView = textView;
+}
+
 
 #pragma mark -
 @dynamic typingAttributes;
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -196,6 +196,7 @@
 - (UITextView *)textView
 {
   ASDisplayNodeAssertMainThread();
+  ASDisplayNodeAssert(_textView != nil, @"ASEditableTextNode's text view has not been allocated yet — access it in -didLoad or after accessing the .view property instead");
   return _textView;
 }
 
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -193,7 +193,8 @@
 #pragma mark -
 @dynamic textView;
 
-- (UITextView *)textView{
+- (UITextView *)textView
+{
   ASDisplayNodeAssertMainThread();
   return _textView;
 }
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -196,7 +196,9 @@
 - (UITextView *)textView
 {
   ASDisplayNodeAssertMainThread();
-  ASDisplayNodeAssert(_textView != nil, @"ASEditableTextNode's text view has not been allocated yet — access it in -didLoad or after accessing the .view property instead");
+  if (!_textView) {
+    _textView = [[_ASDisabledPanUITextView alloc] initWithFrame:CGRectZero textContainer:_textKitComponents.textContainer];
+  }
   return _textView;
 }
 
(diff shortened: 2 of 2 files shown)


AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -133,7 +134,7 @@ - (void)didLoad
   [self.view addSubview:_placeholderTextKitComponents.textView];
 
   // Create and configure our text view.
-  _textKitComponents.textView = [[_ASDisabledPanUITextView alloc] initWithFrame:CGRectZero textContainer:_textKitComponents.textContainer];
+    _textKitComponents.textView = _textView;
```
appleguy: Please fix indentation here

appleguy:
Awesome; thank you!  If accessing this property before the editable text node's view is created (basically before the node is "loaded") will trigger the creation of that view, please add a comment in the header about that side-effect.  Also worth mentioning in the comment, just to drive the point home, that the property can only be used on the main thread.  This is tricky because a lot of nodes are allocated / configured on background queues and that is a time you might want to set other properties on the UITextView, so we may need to do this more carefully to properly address the broad use cases.

Lastly, please add an accessor / getter implementation that has a main thread assertion in it so that folks aren't accidentally calling the method in the background.

marmelroy:
Excellent points. Pushed an update.

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -189,6 +191,21 @@ - (void)setOpaque:(BOOL)opaque
 @synthesize delegate = _delegate;
 
 #pragma mark -
+@dynamic textView;
+
+- (UITextView *)textView{
```
appleguy: Brace style :(

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -189,6 +191,21 @@ - (void)setOpaque:(BOOL)opaque
...
+  ASDisplayNodeAssertMainThread();
+  return _textView;
```
appleguy: One last thing that worries me here: if this is called before it is initialized, it's likely because a client will be attempting to set properties on the text view in an -init method.  Say that it is on the main thread, like a UIView init directly hosting an ASEditableTextNode.  The problematic pitfall is that nil messaging will mean the calls fail silently, and the developer is left wondering why their configuration is not working...

appleguy:
Thanks for your improvement, @marmelroy!  Super close here, almost hit merge.  Last thing: please add some mechanism to alert developers accessing the text view before it is initialized.  I think this would be sufficient: ASDisplayNodeAssert(_textField != nil, @"ASEditableTextNode's text field has not been allocated yet — access it in -didLoad or after accessing the .view property instead")

Since we now have the main thread assertion (very important), an even better solution would probably be to force creation of the view and text field at the access time of the property if it hasn't been created.  This would make the API more convenient to use where the editable node is being directly hosted, and in the case of a regular node hierarchy, if initialization is being performed off-main then the assertion would catch it.

appleguy:
Sorry for the few days' delay here, busy week :).  I will be working on the project over the weekend and can merge the finalized change quickly.

marmelroy:
No worries :) So I fixed the brace style (oops). Added the warning for premature accessing and then replaced it with lazy creation of the textview in the accessor after the main thread assert.

markmark1:
I would like to use pinch to Zoom to resize all the text in the ascollectionview is there an example or pointers on how to do thanks

appleguy:
@markmark1 that seems like a question that isn't at all related to this PR...feel free to ask it in the Google Group or Facebook Group for discussion, though!  There are a lot more people watching those who might have advice.  It's not totally clear the effect you're going for, because redrawing text is pretty expensive and usually has discrete font sizes (at least integers) that wouldn't feel very natural to control with a continuous gesture like a pinch.  Even a slider might be a more understandable UI.  I could be misunderstanding your idea, though.

appleguy:
@marmelroy thanks!  Rerunning the build to see if we can get a green one, then will merge.

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -77,6 +78,7 @@ - (instancetype)init
   _textKitComponents = [ASTextKitComponents componentsWithAttributedSeedString:nil textContainerSize:CGSizeZero];
   _textKitComponents.layoutManager.delegate = self;
   _wordKerner = [[ASTextNodeWordKerner alloc] init];
+  _textView = [[_ASDisabledPanUITextView alloc] initWithFrame:CGRectZero textContainer:_textKitComponents.textContainer];
```
appleguy: Hmm, unfortunately this isn't right.  We can't allocate views in -init, as it is likely to be called off the main thread.  It's necessary to refactor this so that no text view initialization happens in -init.

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -133,7 +135,7 @@ - (void)didLoad
   [self.view addSubview:_placeholderTextKitComponents.textView];
 
   // Create and configure our text view.
-  _textKitComponents.textView = [[_ASDisabledPanUITextView alloc] initWithFrame:CGRectZero textContainer:_textKitComponents.textContainer];
+  _textKitComponents.textView = _textView;
```
appleguy: Since we're doing this, it looks like we might as well not add the _textView instance variable at all.  Let's just use _textKitComponents.textView, anywhere and everywhere that _textView would otherwise need to be used.  This ensures that we wouldn't have a case where only one of them is initialized but they aren't set up to be the same.

AsyncDisplayKit/ASEditableTextNode.h
```diff
@@ -19,6 +19,12 @@
 
 #pragma mark - Configuration
 
+/**
+  @abstract Access to underlying UITextView for more configuration options.
+  @warning This property should only be used on the main thread and should not be accessed before the editable text node's view is created.
+ */
+@property (nonatomic, readwrite, strong) UITextView *textView;
```
appleguy: This should not be a readwrite property as we don't want arbitrary UITextViews to be able to assigned here, since it needs to be a custom subclass.

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -41,6 +41,7 @@ @interface ASEditableTextNode () <UITextViewDelegate, NSLayoutManagerDelegate>
 {
   @private
   // Configuration.
+  UITextView *_textView;
```
appleguy: Let's just remove this variable completely, because a) it is not needed, and b) increases the potential for bugs where it is out of sync with the _textKitComponents.

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -189,6 +191,25 @@ - (void)setOpaque:(BOOL)opaque
...
+- (UITextView *)textView
+{
+  ASDisplayNodeAssertMainThread();
+  if (!_textView) {
+    _textView = [[_ASDisabledPanUITextView alloc] initWithFrame:CGRectZero textContainer:_textKitComponents.textContainer];
+  }
+  return _textView;
+}
+
+- (void)setTextView:(UITextView *)textView
```
appleguy: Seems like we should get rid of this since presumably the _ASDisabledPanUITextView is important.

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -189,6 +191,25 @@ - (void)setOpaque:(BOOL)opaque
 @synthesize delegate = _delegate;
 
 #pragma mark -
+@dynamic textView;
```
appleguy: No need for either @dynamic or @synthesize, right?  I could be wrong, but with the readonly property and implementing the getter method below, I think we should be good.  Also remove the pragma mark because it doesn't seem to be adding anything (e.g. it doesn't have a name or any real value because it is so close to the other one).

AsyncDisplayKit/ASEditableTextNode.mm
appleguy: Let's refactor this area to do something like this:

UITextView *textView = self.textView; // The textView accessor method should set up _textKitComponents.textView

Then just use textView directly in the following six lines.  This eliminates 6 message sends as well as makes the code more readable.

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -189,6 +191,25 @@ - (void)setOpaque:(BOOL)opaque
...
+
+- (UITextView *)textView
+{
+  ASDisplayNodeAssertMainThread();
+  if (!_textView) {
```
appleguy: This will then just be !_textKitComponents.textView,

and if it doesn't exist, just set _textKitComponents.textView = alloc / init.

Then just quickly check the various areas in the class that use _textKitComponents to ensure that in all cases, this accessor method will be called at least once before the textView is used, so that it is set up properly.

appleguy:
Sorry for all the comments!  This one won't require much code, but it's very careful that the threading & initialization paths are correct.

markmark1:
Ok will do @appleguy

markmark1:
Wanted to know what the right way to do it for all the text uitextviews in asdk as its super expensive  and discrete font sizes 

Hence posted in the uitextviews thread but will check the Facebook and Google groups as well

markmark1:
@appleguy can you please let me know what the links to the facebook group and google group

appleguy:
@markmark1 —

https://groups.google.com/forum/#!forum/asyncdisplaykit
http://fb.me/paperf8

marmelroy:
@appleguy - I pushed a simpler and cleaner implementation this morning. Travis CI stalls during building but doesn't look like it's related to this pull request.

appleguy:
Something is really messed up with travis!  @eanagel I wonder if It could be any changes we made to the tests?

Restarting the build.

appleguy:
We are working on the build issue.  It appears to be the recent Cocoapods version change, which in fact broke the internal Pinterest build as well.  Sorry that's blocking this for now, should be fixed tomorrow.

marmelroy:
@appleguy looks like Travis issues were fixed!

appleguy:
@marmelroy this change is officially badass-ified.  Thank you for surfacing the need.  We actually also added returnKeyType here as a direct passthrough property, FWIW.

PR #551, opened 2015-07-09
releasing downloaded image when multiplex image node exits render range

ryanfitz:
This update improves the memory consumption of `ASMultiplexImageNode` which are no longer on screen, but remain in memory. The downside to this change is that when the node gets re-rendered it will fetch images off the network, hopefully most users are using an image cache so this will not too negatively impact them.

This keeps things a bit more inline with how `ASNetworkImageNode` works, however, I noticed `ASNetworkImageNode` clears itself out in `clearFetchedData`. However, `ASMultiplexImageNode` is currently using `clearContents` for clearing itself out. I kept this logic inside `clearContents` because of that, but let me know if this is more appropriate in  `clearFetchedData`.

### DIFF
--- a/AsyncDisplayKit/ASMultiplexImageNode.mm
+++ b/AsyncDisplayKit/ASMultiplexImageNode.mm
@@ -139,6 +139,13 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
  */
 - (void)_downloadImageWithIdentifier:(id)imageIdentifier URL:(NSURL *)imageURL completion:(void (^)(UIImage *image, NSError *error))completionBlock;
 
+/**
+ @abstract Returns a Boolean value indicating whether the downloaded image should be removed when clearing fetched data
+ @discussion Downloaded image data should only be cleared out if a cache is present
+ @return YES if an image cache is available; otherwise, NO.
+ */
+- (BOOL)_shouldClearFetchedImageData;
+
 @end
 
 @implementation ASMultiplexImageNode
@@ -173,11 +180,15 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   }
 }
 
-- (void)displayWillStart
+- (void)clearFetchedData
 {
-  [super displayWillStart];
-
-  [self fetchData];
+  [super clearFetchedData];
+    
+  if ([self _shouldClearFetchedImageData]) {
+    // setting this to nil makes the node fetch images the next time its display starts
+    _loadedImageIdentifier = nil;
+    self.image = nil;
+  }
 }
 
 - (void)fetchData
@@ -624,6 +635,10 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   // Load our next image, if we have one to load.
   if ([self _nextImageIdentifierToDownload])
     [self _loadNextImage];
+}
+
+- (BOOL)_shouldClearFetchedImageData {
+  return _cache != nil;
 }
 
 @end


appleguy:
Interesting find.  A few comments on the sorts of things we should consider in making this behavior positive for all clients, in the spirit of helping the framework serve many use cases:

- Could you share any of your memory profiling related to this?  Typically, the .image will be a compressed image; for JPEG, it is usually a full 10x smaller than the decoded image.  Particularly for apps not using a cache — I wouldn't be surprised if many ASDK clients don't have one, since there is no "ASBasicImageCache" provided with the framework — it would be quite costly to force redownloads when they may not be experiencing any memory pressure in the app.

- On that note, let's wrap the clearing login in a check for a valid cache object.  Only clear the data if there is a cache.  Feel free to update a header comment about this, if you'd like.

- As a bonus, we could track whether the node is expected to have data loaded (e.g. set a flag in -fetchData, clear the flag in -clearFetchedData).  Then implement an NSNotification for UIKit memory warnings, and on memory warning, purge the fetched data if 1) not inside the fetchData zone, 2) even if there is no cache.  This should make the behavior quite robust for apps not using a cache.

- Please do move to -clearFetchedData.  The kick-off of the download should be in -fetchData as well; can you check that and fix it as well if it's oddly somewhere else?  The network / preload range is adjustable independently from the render range, and so for users that don't have a cache, it would make a lot of sense to be able to adjust the eviction aggressiveness as potentially distinct from the decode / backing store deallocation behavior.

ryanfitz:
I ran some quick profiling on my app with and without this PR. I should first point out that my use case is definitely on the extreme end. My app has a root tab bar controller where 3 of those tabs are using `ASTableView` which render rows of images. Within each tab a user can navigate further down into more view controllers of `ASTableView`. 

Without this PR memory usage slowly, but steadily builds up as a user continues to move further and further into the app. With moderate usage memory grows to over 300megs. However, with this PR memory usage stays very low and stable. Ive tried heavy usage within my app and memory usage peaks to around 100megs, it remains very stable at 100 and quickly drops down depending on where a user navigates. Ive yet to receive a memory warning with this PR even when testing on a 4s. I'm sure its because I'm caching images, but I have yet to notice a difference in user experience when removing the images and they need to get reloaded.

With that said, my use is absolutely on the extreme end and not the typical use case. I would be perfectly fine with having either having to configure a ASMultiplexImage or subclass it in order to clean itself out the way I require it. 

I'll get you an updated pull request with the changes you suggested. Thanks for the feedback.

ryanfitz:
I pushed up a new version of this PR.

* I found that `displayWillStart` was being overridden to invoke `fetchData`, I removed this override. 

* I added a new private method `_shouldClearFetchedImageData` to help clearly convey if / when a previously fetched image will be nilled out. I also added some header comments. 

* I moved the logic of removing the fetched image to `clearFetchedData`. I left the existing `clearContents` code intact as I was not sure if it also should be in `clearFetchedData` or if its more appropriate in `clearContents`

markmark1:
Way to go

markmark1:
Ryan can you share an example of this extreme use case of as table view

ryanfitz:
@markmark1 here are a couple of screenshot that might give a better idea of my use case. Everything other than the navbar & tabbars are ASDK Nodes, rendered in an `ASTableView`. The table view has sticky headers, which are also display nodes. The multiple images you see rendered in the grid are the `ASMultiplexImages`, it fetches a low quality and medium quality images. Low quality images are usually under 10kb and the mediums around 70.

I could also put together an example app pretty quickly which might give a better example.

Explore Image
Profile Image

appleguy:
Awesome, thank you for the attention to detail!  This is a great improvement.

Your app is looking quite beautiful.  Keep up the awesome development :).

markmark1:
Wow looks amazing can you please Share the demo would help a lot

markmark1:
It looks beautiful what's it called thanks I d love to download it

Issue #565, opened 2015-07-13
FBSnapshotCase 2.0+ requires frameworks enabled


eanagel:
Performing a pod update for AsyncDisplayKit project causes an error because FBSnapshotCase 2.0+ requires framework support. Simplest fix is to specify a pod version pre 2.0 (1.8.1). Is there any reason we need 2.0+? I assume no one has used it since this isn't fixed ;) (Alternatively we could enable frameworks form cocoapods and see how well that works.)

eanagel:
Resolved, closing.

PR #567, opened 2015-07-13
Fb snapshot test case version

eanagel:
Update the Podfile to point to FBSnapshotTestCase 1.8.1 because 2.0+ requires framework support which we haven't tried enabling yet. This is a fix for #565.


appleguy:
Thanks!

Issue #566, opened 2015-07-13
Race in ASRangeController updateVisibleNodeIndexPaths causes NSRangeException


eanagel:
A race condition in ASRangeController updateVisibleNodeIndexPaths ultimately causes an NSRangeException while attempting to calculate the list of nodes that have exited the visible range. I haven't been able to figure out the exact scenario that causes this but I am readying an updated test for ASTableView the causes the exception at least sometimes.

eanagel:
Resolved by #575, closing.

PR #568, opened 2015-07-13
Update ASTableView test

eanagel:
ASTableView test wasn't as random as it was likely intended to be. I've made the arguments actually random and upped the number of iterations to 250 (Previously it was looping through with almost no variations in the parameters, except for the ranges.) These changes causes the test to reproduce issue #566 about 2/3 of the time.

I've also added some log statements to aid in debugging that can be removed once we are on stable footing with #566. (We might want to reduce the number of iterations once our confidence is a bit higher too.)


AsyncDisplayKitTests/ASTableViewTests.m
```diff
@@ -12,7 +12,7 @@
 
 #define NumberOfSections 10
 #define NumberOfRowsPerSection 20
-#define NumberOfReloadIterations 50
+#define NumberOfReloadIterations 250
```
appleguy: This takes quite a while to run on some machines and we already have flaky tests on some Travis runs (for example, this run seems to have failed, though not sure if it is this test).  Let's dial this back to no more than 100, maybe 50, because I think the build machines are Mac Minis.

AsyncDisplayKitTests/ASTableViewTests.m
```diff
@@ -155,28 +155,34 @@ - (void)testReloadData
...
-    BOOL letRunloopProceed            = (arc4random_uniform(1) == 0 ? YES : NO);
-    BOOL useBeginEndUpdates           = (arc4random_uniform(2) == 0 ? YES : NO);
+    UITableViewRowAnimation rowAnimation = (arc4random_uniform(2) == 0 ? UITableViewRowAnimationMiddle : UITableViewRowAnimationNone);
+    BOOL animatedScroll               = (arc4random_uniform(2) == 0 ? YES : NO);
+    BOOL reloadRowsInsteadOfSections  = (arc4random_uniform(2) == 0 ? YES : NO);
+    NSTimeInterval runLoopDelay       = ((arc4random_uniform(2) == 0) ? (1.0 / (1 + arc4random_uniform(500))) : 0);
+    BOOL useBeginEndUpdates           = (arc4random_uniform(3) == 0 ? YES : NO);
+
+    // instrument our instrumentation ;)
+    NSLog(@"Iteration %03d: %@|%@|%@|%@|%g", i, (rowAnimation == UITableViewRowAnimationNone) ? @"NONE  " : @"MIDDLE", animatedScroll ? @"ASCR" : @"    ", reloadRowsInsteadOfSections ? @"ROWS" : @"SECS", useBeginEndUpdates ? @"BEGEND" : @"      ", runLoopDelay);
```
appleguy: This should not be unconditionally enabled, right?  Again would add a lot more overhead to the test runs.  Cool to have there commented out, although it would be much more useful in ASTableViewStressTest the sample project where any actual debugging would be done.

AsyncDisplayKitTests/ASTableViewTests.m
```diff
@@ -155,28 +155,34 @@ - (void)testReloadData
...
   
   for (int i = 0; i < NumberOfReloadIterations; ++i) {
-    UITableViewRowAnimation rowAnimation = (arc4random_uniform(1) == 0 ? UITableViewRowAnimationMiddle : UITableViewRowAnimationNone);
-    
-    BOOL animatedScroll               = (arc4random_uniform(1) == 0 ? YES : NO);
-    BOOL reloadRowsInsteadOfSections  = (arc4random_uniform(1) == 0 ? YES : NO);
-    BOOL letRunloopProceed            = (arc4random_uniform(1) == 0 ? YES : NO);
-    BOOL useBeginEndUpdates           = (arc4random_uniform(2) == 0 ? YES : NO);
+    UITableViewRowAnimation rowAnimation = (arc4random_uniform(2) == 0 ? UITableViewRowAnimationMiddle : UITableViewRowAnimationNone);
+    BOOL animatedScroll               = (arc4random_uniform(2) == 0 ? YES : NO);
```
appleguy: Methodology for changing the frequencies of these?  It seems fine to do, but can you bring the changes to ASTableViewStressTest as well?  Sadly there is code duplication between them, but not complete duplication because the test method needs to run in isolation of other things.  The editable data source of the sample project is much more flexible to fully test, and has found issues that this test cannot.  Similarly we need to move to ASCollectionViewStressTest soon and also thrash supplementary views, which are currently known to be crashing but untested.

AsyncDisplayKitTests/ASTableViewTests.m
eanagel: Yeah I def. agree once we have the issue I'm seeing fixed. I wanted to make sure you had something that would produce the problem ;) Let me tighten this up a bit and I'll update the stress test to make sure it can reproduce #566.

AsyncDisplayKitTests/ASTableViewTests.m
eanagel: Yep, was just to help troubleshoot #566. It's also good to help debug the test logic ;) I'll move to stress test and comment out in case it's useful at a future date here.

AsyncDisplayKitTests/ASTableViewTests.m
eanagel: arc4random_uniform(upper_bound) returns a value between 0..upper_bound-1, so arc4random_uniform(1) === 0, which means these were all previously YES. Additionally, the runLoopDelay calculation wasn't using floating point so it was always returning either 0 or (very rarely) 1. So it just wasn't very random, although it looked like it was.

eanagel:
Ok I've updated the test so it's not abusing the test servers unduly.

appleguy:
Awesome, thanks for auditing this!  Since it is occasionally failing, I'll land this only after making the fixes (hopefully this weekend)

appleguy:
I'm going to land this now that I have some time in the next couple days to hopefully resolve the underlying issue(s).  Thanks for not only finding my bugs, but fixing them too!  ha :).  I'll aim to back port this to ASTableViewStressTest.

PR #569, opened 2015-07-13
Actually coalesce calls to updateVisibleNodeIndexPaths

eanagel:


appleguy:
@eanagel I think we discussed this one in person — this should not be necessary due to the boolean flag that is checked earlier in the method, correct?  In what scenario could the performSelector: be queued up multiple times?  Note that calling cancel and perform again incurs some additional cost, so it is net preferred if it can be avoided.

eanagel:
Oh yeah! I see that now. Closing.

Issue #572, opened 2015-07-17
endUpdates (or editing commands) may complete asynchronously when it's no longer valid


eanagel:
Because endUpdates (or any editing command not in a begin/endUpdates block) completes asynchronously, it is possible for this to complete after the ASTableView has been discarded. This causes a crash when calling [super endUpdates], or other editing commands.

Here's a specific example: my app is using paging (or virtualization), so we are constantly inserting and deleting rows as the user is scrolling. If the user starts scrolling and then dismisses the view, it's possible for an async endUpdates (in my case) to be in flight when we dismiss. When the endUpdates completes the UITableView is no longer in a valid state and crashes.

I'm pretty sure this scenario would also apply to reloadData.

I can think of a couple solutions to this problem. The simplest is to check that asyncDataSource is valid in each of the async rangeController methods in ASTableView and fail silently if it is no longer wired. A possibly more better performing approach would be to kill any pending tasks immediately when asyncDataSource becomes nil.

eanagel:
Resolved, closing.

PR #573, opened 2015-07-17
Prevent making async updates to UITable|CollectionView when asycDataSource is cleared

eanagel:
Short-circuit calls that make async updates to UITableView or UIViewController after the asyncDataSource has been set to nil. This can happen when an update is being processed asynchronously and the client begins tearing down an ASTableView or ASCollectionView instance.

This is a proposed fix for #572.


appleguy:
Thank you for debugging this and submitting the patch, covering every edit method!  The scenario immediately makes sense to me as a clear flaw in the implementation — not something that most test cases would run into, but very plausibly relevant for an app dismissing a non-quiesced table or collection.

Issue #564, opened 2015-07-13
ASDataController does not call begin/end Updates delegate methods in correct order


eanagel:
Currently `ASDataControlle`r calls` -dataControllerBeginUpdates:` and `-dataControllerEndUpdates:completion`: directly on the main thread while calls to perform edits (such as` -dataController:didInsertNodes:atIndexPaths:withAnimationOptions:`) are routed through the editingTransactionQueue first. This results in the begin and end updates calls being made immediately followed by the editing delegate calls.

UITableView, for instance will ultimately see calls such as:

 - beginUpdates
 - endUpdates
 - insertRows:...

AgeOfMobile:
I second this. After upgrade to 1.2.2 from 1.2, I faced a lot of problems if I use [ASCollectionView performBatchUpdates:complete. The "complete" block is called before edit methods (insert/reload/deleteRows...) completed.

eanagel:
Resolved by #575, closing

PR #575, opened 2015-07-20
Take 2: Automatically adjust content offset for insertion/deletion

eanagel:
This is a merged and cleaned-up version of PR #548.

This PR adds a new property for ASTableView (automaticallyAdjustsContentOffset) that, when enabled, will adjust the contentOffset for inserts and deletes before the visible viewport so the users view is maintained. In order to use this effectively, animations must be disabled while performing updates. To facilitate this an option has been added to endUpdates (and also a completion handler that has many uses.) Because this only works well with animations are disabled and it may require some thought to use effectively, it defaults to off.

There is a performance trade-off for this feature that I'd like anyone to weigh in on. In order to make this work, we need to know the heights of the nodes being inserted or deleted. (well we actually only need to know the heights of the ones before the visible cells.) I've extended our delegates to pass that information own to the ASTableView, but if it's not used it is a waste. These data is pretty easily available for inserting nodes, but in the case of deleting nodes, there is the additional overhead of capturing the deleted items and passing them down. There are alternative approaches I'm sure we could work out of this seems like it could be a big performance hit.

I think this handles the common cases for virtualization, but there's more that could be added to it to make it even more robust. Right now, it doesn't handle insertion/deletion of sections and you must explicitly disable animations in endUpdate to make it work correctly. If we wanted to make things more automatic we could separate the items that would be inserted before the visual window and make those insertions/deletions without animations (in a separate begin/end update block) and then follow up with any other updates in an animated fashion. It's considerably more work, but that would make this an "it just works" feature which I like ;)

You can see this in action with a simple demo app at: https://github.com/eanagel/ASDKVirtualizationDemo

Additionally, this includes a fix for a problem where ASDataController appears to be calling it's beginUpdate and endUpdates delegate method before the didInsert/Delete delegate methods.. See issue #564 for more details.

There is also a fix for a crash in ASRangeController caused when it doesn't correctly refresh its internal cache of indexPaths when inserts or deletes happen. This is a fix for #566.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -79,9 +79,8 @@
 @property (nonatomic, assign) CGFloat leadingScreensForBatching;
 
 /**
- *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. You can call it from background 
- *  thread (it is recommendated) and the UI collection view will be updated asynchronously. The asyncDataSource must be updated 
- *  to reflect the changes before this method is called.
+ *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. This method must be called from the main thread. 
+ *  The asyncDataSource must be updated to reflect the changes before the update block completes.
  *
  *  @param animated   NO to disable animations for this batch
  *  @param updates    The block that performs the relevant insert, delete, reload, or move operations.
@@ -92,8 +91,8 @@
 - (void)performBatchAnimated:(BOOL)animated updates:(void (^)())updates completion:(void (^)(BOOL))completion;
 
 /**
- *  Perform a batch of updates asynchronously. You can call it from background thread (it is recommendated) and the UI collection 
- *  view will be updated asynchronously. The asyncDataSource must be updated to reflect the changes before this method is called.
+ *  Perform a batch of updates asynchronously.  This method must be called from the main thread.
+ *  The asyncDataSource must be updated to reflect the changes before update block completes.
  *
  *  @param updates    The block that performs the relevant insert, delete, reload, or move operations.
  *  @param completion A completion handler block to execute when all of the operations are finished. This block takes a single
@@ -123,8 +122,7 @@
  *
  * @param sections An index set that specifies the sections to insert.
  *
- * @discussion This operation is asynchronous and thread safe. You can call it from background thread (it is recommendated)
- * and the UI collection view will be updated asynchronously. The asyncDataSource must be updated to reflect the changes
+ * @discussion This method must be called from the main thread. The asyncDataSource must be updated to reflect the changes
  * before this method is called.
  */
 - (void)insertSections:(NSIndexSet *)sections;
@@ -134,8 +132,7 @@
  *
  * @param sections An index set that specifies the sections to delete.
  *
- * @discussion This operation is asynchronous and thread safe. You can call it from background thread (it is recommendated)
- * and the UI collection view will be updated asynchronously. The asyncDataSource must be updated to reflect the changes
+ * @discussion This method must be called from the main thread. The asyncDataSource must be updated to reflect the changes
  * before this method is called.
  */
 - (void)deleteSections:(NSIndexSet *)sections;
@@ -145,8 +142,7 @@
  *
  * @param sections An index set that specifies the sections to reload.
  *
- * @discussion This operation is asynchronous and thread safe. You can call it from background thread (it is recommendated)
- * and the UI collection view will be updated asynchronously. The asyncDataSource must be updated to reflect the changes
+ * @discussion This method must be called from the main thread. The asyncDataSource must be updated to reflect the changes
  * before this method is called.
  */
 - (void)reloadSections:(NSIndexSet *)sections;
@@ -158,8 +154,7 @@
  *
  * @param newSection The index that is the destination of the move for the section.
  *
- * @discussion This operation is asynchronous and thread safe. You can call it from background thread (it is recommendated)
- * and the UI collection view will be updated asynchronously. The asyncDataSource must be updated to reflect the changes
+ * @discussion This method must be called from the main thread. The asyncDataSource must be updated to reflect the changes
  * before this method is called.
  */
 - (void)moveSection:(NSInteger)section toSection:(NSInteger)newSection;
@@ -169,8 +164,7 @@
  *
  * @param indexPaths An array of NSIndexPath objects, each representing an item index and section index that together identify an item.
  *
- * @discussion This operation is asynchronous and thread safe. You can call it from background thread (it is recommendated)
- * and the UI collection view will be updated asynchronously. The asyncDataSource must be updated to reflect the changes
+ * @discussion This method must be called from the main thread. The asyncDataSource must be updated to reflect the changes
  * before this method is called.
  */
 - (void)insertItemsAtIndexPaths:(NSArray *)indexPaths;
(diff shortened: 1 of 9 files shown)


eanagel:
Also should resolve #566.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -600,7 +615,7 @@ - (void)rangeController:(ASRangeController *)rangeController didInsertNodesAtInd
   }
 }
 
-- (void)rangeController:(ASRangeController *)rangeController didDeleteNodesAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
+- (void)rangeController:(ASRangeController *)rangeController didDeleteNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
```
appleguy: I appreciate you explicitly calling out the performance cost of adding this.  Because I have plans to do a detailed and focused optimization effort for these code paths in the next few months (probably ASDK 2.1, Oct or Nov), I'm not concerned about this.  The impact is likely small for most use cases, although it is also possible to construct cases where the overhead could be significant.  The feature is worth the anticipated cost in this case.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -79,6 +79,30 @@
 @property (nonatomic, assign) CGFloat leadingScreensForBatching;
 
 /**
+ *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. You can call it from background 
+ *  thread (it is recommendated) and the UI collection view will be updated asynchronously. The asyncDataSource must be updated
```
appleguy: Review my final comments on the old PR — there is a typo here and I'm also worried about the explicit advice to call on a background thread.  Very interested to hear your thoughts on this, but I was pretty confident we should not be even /supporting/ edits on a background thread — so probably not recommend them :)

AsyncDisplayKit/ASTableView.mm
```diff
@@ -126,6 +128,9 @@ @interface ASTableView () <ASRangeControllerDelegate, ASDataControllerSource> {
   ASBatchContext *_batchContext;
 
   NSIndexPath *_pendingVisibleIndexPath;
+
+  NSIndexPath *_contentOffsetAdjustmentTopVisibleRow;
```
appleguy: Awesome name improvement

AsyncDisplayKit/ASTableView.mm
```diff
@@ -374,6 +385,57 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
...
+      adjustment += cellNode.calculatedSize.height * dir;
+      if (indexPath.section == top.section) {
+        top = [NSIndexPath indexPathForRow:top.row+dir inSection:top.section];
+      }
+    }
+  }
+
+  if (_contentOffsetAdjustmentTopVisibleRow) { // true of we are in a begin/end update block (see beginAdjustingContentOffset)
+    _contentOffsetAdjustmentTopVisibleRow = top;
+     _contentOffsetAdjustment += adjustment;
```
appleguy: Extra space, fix indentation

AsyncDisplayKit/ASTableView.mm
```diff
@@ -374,6 +385,57 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
...
+  // before the visible area. If in a begin/end updates block this will update _contentOffsetAdjustment, otherwise it will
+  // update self.contentOffset directly.
+
+  ASDisplayNodeAssert(_automaticallyAdjustsContentOffset, @"this method should only be called when _automaticallyAdjustsContentOffset == YES");
+
+  CGFloat dir = (inserting) ? +1 : -1;
+  CGFloat adjustment = 0;
+  NSIndexPath *top = _contentOffsetAdjustmentTopVisibleRow ?: self.indexPathsForVisibleRows.firstObject;
+
+  for (int index=0; index<indexPaths.count; index++) {
```
appleguy: Should have space between operators to match the de facto style

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -318,6 +325,7 @@ - (void)_populateFromEntireDataSourceWithMutableNodes:(NSMutableArray *)nodes mu
 
 - (void)beginUpdates
 {
+  [_editingTransactionQueue waitUntilAllOperationsAreFinished];
   // Begin queuing up edit calls that happen on the main thread.
   // This will prevent further operations from being scheduled on _editingTransactionQueue.
   // It's fine if there is an in-flight operation on _editingTransactionQueue,
```
appleguy: Update this comment / delete at least the lines referring to in-flight operations.  I assume the -wait call turned out to be necessary here (I remember adding it in my patch and believe it did help test behavior), and thus we are not actually incrementing it with operations in flight.

appleguy: I believe this is important so that the delegate itself isn't proceeding with edits to its own state (e.g. that which is returned in the asyncDataSource/Delegate protocols) before any prior ones have finished.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -327,23 +335,38 @@ - (void)beginUpdates
...
-- (void)endUpdatesWithCompletion:(void (^)(BOOL))completion
+- (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
 {
   _batchUpdateCounter--;
 
   if (_batchUpdateCounter == 0) {
-    [_delegate dataControllerBeginUpdates:self];
+    LOG(@"endUpdatesWithCompletion - beginning");
+
+    [_editingTransactionQueue addOperationWithBlock:^{
```
appleguy: In this patch or a followup, it would be worth adding short comments describing here and in the endUpdates delegate call case why it is important to queue them.

Here, the idea is that a runloop turn needs to take place (is this for UIKit or some other reason?)  For the end call, it's necessary to allow the running of main thread delegate call blocks that were scheduled by calling the _pendingEditCommandBlocks.  Since the current method is operating on main, it would otherwise always be able to call the delegate method before the queued ones completed — even if execution waited on each block.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -417,6 +446,9 @@ - (void)reloadSections:(NSIndexSet *)sections withAnimationOptions:(ASDataContro
       
       [_editingTransactionQueue addOperationWithBlock:^{
         NSArray *indexPaths = ASIndexPathsForMultidimensionalArrayAtIndexSet(_editingNodes, sections);
+        
+        LOG(@"Edit Transaction - reloadSections: updatedIndexPaths: %@, indexPaths: %@, _editingNodes: %@", updatedIndexPaths, indexPaths, ASIndexPathsForMultidimensionalArray(_editingNodes));
```
appleguy: I'm pretty sure the macro you wrote will precompile to "literally nothing" here, but just hoping this particular argument doesn't get executed when disabled — it would be a big performance cost if I happened to be wrong in my interpretation of the macro.

eanagel: The macro would translate this to a no-op, only a semicolon will remain

AsyncDisplayKit/Details/ASDataController.mm
eanagel: Missed this in my updates, next time?

AsyncDisplayKit/ASTableView.mm
eanagel: ThankGoodnessLongVariableNamesAreOk

PR #579, opened 2015-07-21
Add vertical text alignment property to ASTextNode

jarodl:


AsyncDisplayKit/ASTextNode.h
```diff
@@ -12,6 +12,14 @@
 @protocol ASTextNodeDelegate;
 
 /**
+ * Alignment
+ */
+typedef NS_ENUM(NSUInteger, ASTextNodeVerticalAlignment) {
+    ASTextNodeVerticalAlignmentDefault = 0,
```
appleguy: Could you add a comment here describing what the Default behavior actually is?  In fact, it's probably better to rename this to a description of its behavior (e.g. ...AlignmentBottom = 0,  // Default)

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -406,10 +408,22 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
   // Offset the text origin by any shadow padding
   UIEdgeInsets shadowPadding = [self shadowPadding];
   CGPoint textOrigin = CGPointMake(self.bounds.origin.x - shadowPadding.left, self.bounds.origin.y - shadowPadding.top);
-  return [[ASTextNodeDrawParameters alloc] initWithRenderer:[self _renderer]
-                                                       shadower:[self _shadower]
-                                                     textOrigin:textOrigin
-                                                backgroundColor:self.backgroundColor.CGColor];
+
+  switch (_verticalAlignment)
+  {
```
appleguy: Code style has the opening brace on the same line as the control statement.  Although the switch would make sense if we had more than one option for this, could you fall back to a simple if (_verticalAlignment == ASTextNodeAlignmentCenter) check?

Check the return line below this — it is indented too far.  Thanks for fixing the alignment of the method arguments, as they were misaligned before!

AsyncDisplayKit/ASTextNode.h
```diff
@@ -12,6 +12,14 @@
 @protocol ASTextNodeDelegate;
 
 /**
+ * Alignment
+ */
+typedef NS_ENUM(NSUInteger, ASTextNodeVerticalAlignment) {
```
appleguy: Let's actually make this type ASTextNodeAlignment.  That way, we can use the same enum for horizontal alignment later.  The main reason I'd prefer this is because it will make the autocomplete much easier to find; ASTextNodeAl [tab] would work, whereas guessing / remembering the necessity of Ver[tab] is tougher.

appleguy:
@jarodl thanks for making this, I'd love to merge the feature!  Could you make the small changes requested?  I'll merge as soon as I see the update.  If you disagree with any of my suggestions, feel free to describe why and I'll almost certainly agree with you.

appleguy:
We are working on the build issue.  It appears to be the recent Cocoapods version change, which in fact broke the internal Pinterest build as well.  Sorry that's blocking this for now, should be fixed tomorrow.

appleguy:
After thinking carefully about this, we've decided to move in the direction of baseline alignment controls as the primary mechanism for adjusting vertical positioning of text: https://github.com/facebook/AsyncDisplayKit/pull/611

In this patch, my understanding is that the only way these properties could have effect is if the bounds / frame size of the text node is made larger than necessary for its content, and thus there would be vertical slack space to position within.  This condition should be avoided pretty strictly anyway, because it consumes additional memory with the larger backing store (and also more pixel area to composite).

Please weigh in with any ideas / use cases on #611 so we don't miss them!

PR #581, opened 2015-07-24
Add completion callbacks to indexPath modifying functions

frankenthumbs:


eanagel:
Well from this PR I can see we have traveled down some similar paths! I've added similar capabilities to endUpdates and reloadData. There are great use cases for this kind of thing (where you need to know the state of your change have been reflected in the underlying UITableView.) When I added the completion handler for endUpdates I was thinking that you could use it whenever you needed to execute tasks on the completion of updates and avoid adding completion handlers for all the update methods.

I'm just curious, are there use cases where that approach wouldn't work, or do you see these methods more as an added convenience?

appleguy:
@frankenthumbs thanks for submitting this!  Echoing @eanagel comments — could you check out the latest master, which includes his changes merged only a couple days ago?  It seems likely that you might not have seen those changes when developing this one.

I am going to be focusing some time on ASDK this weekend and will think more carefully about this in the next two days.  I'm certainly in favor of a simplified API, so in some sense believe we should rely on begin/endUpdates wrapping these calls whenever the developer needs the callback.  On the other hand, having the callback is one way to remind developers of the mental model of using ASTableView and ASCollectionView—the edit changes may not show up onscreen synchronously, e.g. the user could scroll around a bit instead of their interaction being blocked until the edit is processed.  Even for developers who pass NULL for the block, perhaps in some cases that reminder would be useful...

frankenthumbs:
you guys are right. i missed @eanagel endUpdates merge... overriding performBatchUpdates was also a great call imo. using UICollectionView's was what caused me to create this PR.

appleguy:
ok, cool.  Let's close this one out for now, but we can revive it if the next wave of API improvements to ASCollectionView suggest it would be the right direction.  There are definitely a lot of tweaks coming to the API.

Issue #539, opened 2015-07-07
Reconsider static layout spec


nguyenhuy:
ASStaticLayoutSpec can be helpful in some situations. But it's learning curve might be steep (new ASStaticLayoutChild and some size types). So there are a few questions worth discussing:
- Is it worth it? Should we remove the spec completely?
  - [ ] Make some sample usages of this spec, so that it will be easier for others to follow-up.
- In case the spec proves to be useful:
  - [x] Try to simplify the spec: 
    - [x] Eliminate ASStaticLayoutChild, probably by adding position and size to ASLayoutable, which might cause confusion.
    - [x] Revisit new size types.
  - [x] Add tests.

Issue #328, opened 2015-02-24
It looks like ASImageNode ignores constrained size. Is this intended?


jonahneugass-ck:
Looking at ASImageNode.mm, I am seeing this:

    - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
    {
      ASDN::MutexLocker l(_imageLock);
      if (_image)
        return _image.size;
      else
        return CGSizeZero; 
    }

It looks like the constrained size is being completely ignored. Is this intended?

secretiverhyme:
Thanks for the report!  Yes, this is intentional.

Images are typically positioned and sized independently of their contents' actual size — for example, ASNetworkImageNode doesn't even *know* its image's size at display-time, and ASMultiplexImageNode starts out with a much smaller image size than its final state.

As such, it's a standard pattern to not call `-measure:` on image nodes, instead using static layouts like…
* profile picture:  always 150⨉150pt
* photo in newsfeed story:  always a square, with width equal to "supernode's width minus padding"
* fullscreen photo:  always the size of the device's screen

PR #590, opened 2015-07-30
Add expectedSize to ASNetworkImageNode

rcancro:
ASNetworkImageNode defers to ASImageNode to return its calculatedSize. ASImageNode returns the size of its image. There is a good chance that ASNetworkImageNode hasn't downloaded its image yet when calculatedSize is called, so it returns a size of CGSizeZero.

On top of that, it is possible that the size of the image is not actually the size that we wish to display in our node.

I've added an "expectedImageSize" property that can be used to determine the calculatedSize of an ASNetworkImageNode.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -186,8 +186,10 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
 @property (nonatomic, readonly, assign) ASSizeRange constrainedSizeForCalculatedLayout;
 
 /**
- * @abstract Used by some implementations of measureWithSizeRange: to provide an intrisic content size
- * to a node when one cannot be computed from its subnodes
+ * @abstract Provides a default intrinsic content size for calculateSizeThatFits:. This is useful when laying out
+ * a node that either has no intrinsic content size or should be laid out at a different size than its intrinsic content
+ * size. For example, this property could be set on an ASImageNode to display at a size different from the underlying
+ * image size.
  *
  * @return The preferred frame size of this node
  */
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -185,6 +185,14 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
  */
 @property (nonatomic, readonly, assign) ASSizeRange constrainedSizeForCalculatedLayout;
 
+/**
+ * @abstract Used by some implementations of measureWithSizeRange: to provide an intrisic content size
+ * to a node when one cannot be computed from its subnodes
+ *
+ * @return The preferred frame size of this node
+ */
+@property (atomic, assign, readwrite) CGSize preferredFrameSize;
+
 /** @name Managing the nodes hierarchy */
 
 
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -112,7 +112,7 @@
 {
   ASDN::MutexLocker l(_imageLock);
   if (CGSizeEqualToSize(self.preferredFrameSize, CGSizeZero) == NO)
-    return CGSizeMake(MIN(constrainedSize.width, self.preferredFrameSize.width), MIN(constrainedSize.height, self.preferredFrameSize.height));
+    return [super calculateSizeThatFits:constrainedSize];
   else if (_image)
     return _image.size;
   else
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -111,6 +111,7 @@
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   ASDN::MutexLocker l(_imageLock);
+  // if a preferredFrameSize is set, call the superclass to return that instead of using the image size.
   if (CGSizeEqualToSize(self.preferredFrameSize, CGSizeZero) == NO)
     return [super calculateSizeThatFits:constrainedSize];
   else if (_image)
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -111,7 +111,9 @@
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   ASDN::MutexLocker l(_imageLock);
-  if (_image)
+  if (CGSizeEqualToSize(self.preferredFrameSize, CGSizeZero) == NO)
+    return CGSizeMake(MIN(constrainedSize.width, self.preferredFrameSize.width), MIN(constrainedSize.height, self.preferredFrameSize.height));
+  else if (_image)
     return _image.size;
   else
     return CGSizeZero;
--- a/examples/Kittens/Sample/KittenNode.mm
+++ b/examples/Kittens/Sample/KittenNode.mm
@@ -134,9 +134,7 @@ static const CGFloat kInnerPadding = 10.0f;
 #if UseAutomaticLayout
 - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
 {
-  ASRatioLayoutSpec *imagePlaceholder = [ASRatioLayoutSpec newWithRatio:1.0 child:_imageNode];
-  imagePlaceholder.flexBasis = ASRelativeDimensionMakeWithPoints(kImageSize * (!_isImageEnlarged ? 1 : 2));
-  
+  _imageNode.preferredFrameSize = _isImageEnlarged ? CGSizeMake(2.0 * kImageSize, 2.0 * kImageSize) : CGSizeMake(kImageSize, kImageSize);
   _textNode.flexShrink = YES;
   
   return
@@ -148,7 +146,7 @@ static const CGFloat kInnerPadding = 10.0f;
       .direction = ASStackLayoutDirectionHorizontal,
       .spacing = kInnerPadding
     }
-    children:!_swappedTextAndImage ? @[imagePlaceholder, _textNode] : @[_textNode, imagePlaceholder]]];
+    children:!_swappedTextAndImage ? @[_imageNode, _textNode] : @[_textNode, _imageNode]]];
 }
 
 // With box model, you don't need to override this method, unless you want to add custom logic.
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -47,6 +47,7 @@
 @synthesize flexShrink = _flexShrink;
 @synthesize flexBasis = _flexBasis;
 @synthesize alignSelf = _alignSelf;
+@synthesize preferredFrameSize = _preferredFrameSize;
 
 BOOL ASDisplayNodeSubclassOverridesSelector(Class subclass, SEL selector)
 {
@@ -135,6 +136,7 @@ void ASDisplayNodePerformBlockOnMainThread(void (^block)())
   _methodOverrides = overrides;
   
   _flexBasis = ASRelativeDimensionUnconstrained;
+  _preferredFrameSize = CGSizeZero;
 }
 
 - (id)init
@@ -1330,7 +1332,7 @@ static NSInteger incrementIfFound(NSInteger i) {
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   ASDisplayNodeAssertThreadAffinity(self);
-  return CGSizeZero;
+  return _preferredFrameSize;
 }
 
 - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
@@ -1357,6 +1359,17 @@ static NSInteger incrementIfFound(NSInteger i) {
   return _constrainedSize;
 }
 
+- (void)setPreferredFrameSize:(CGSize)preferredFrameSize
+{
+  ASDN::MutexLocker l(_propertyLock);
+  _preferredFrameSize = preferredFrameSize;
+}
+
+- (CGSize)preferredFrameSize
+{
+  ASDN::MutexLocker l(_propertyLock);
+  return _preferredFrameSize;
+}
 - (UIImage *)placeholderImage
 {
   return nil;


appleguy:
Thanks @rcancro !  This is indeed very useful, and now quite essential in the world of ASLayoutSpec. @nguyenhuy to also review and consider — I think it is likely that we should push this down to ASDisplayNode, next to the flexBasis and other properties.

nguyenhuy:
Thanks for your contribution, @rcancro.

I agree with @appleguy that, in many cases, this new property is not only very useful for network image nodes but also for display nodes. flexBasis and other properties are declared in `ASLayoutable.h` which is conformed by both ASDisplayNode and layout specs. If I understand "next to the flexBassis and other properties" correctly, I think we should **not** push it down to `ASLayoutable.h`. Because the main, if not the only, reason for the existence of each layout spec is that the spec can measure its sized based on its unique layout logic and its children. So layout specs should not have a predefined size. We should push this down to ASDisplayNode, but **no** further.

Also, since calculated size must be within the given constrained size range, I think there are 2 things we must consider:
- Name this new property according to this rule. To me, "expected" implies too much certainty. Perhaps "suggested" and "proposed" are better?
- In -calculateSizeThatFits:, the predefined size should be clamped with the constrained size range. This will help to enforce the rule in some edge cases (although it will introduce a small run-time overhead because layout specs also clamp the calculated size of each child).

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -111,6 +113,18 @@ - (UIImage *)defaultImage
   return _defaultImage;
 }
 
+-(void)setExpectedImageSize:(CGSize)expectedImageSize
+{
+    ASDN::MutexLocker l(_lock);
```
nguyenhuy: 2 spaces only, please :) Same for the rest of this file.

examples/Kittens/Sample/KittenNode.mm
```diff
@@ -131,8 +132,6 @@ - (NSDictionary *)textStyle
 #if UseAutomaticLayout
 - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
 {
-  ASRatioLayoutSpec *imagePlaceholder = [ASRatioLayoutSpec newWithRatio:1.0 child:_imageNode];
-  imagePlaceholder.flexBasis = ASRelativeDimensionMakeWithPoints(kImageSize);
```
nguyenhuy: Way easier to understand :+1:

appleguy:
@nguyenhuy very good point, I agree this does not belong in ASLayoutable and forgot that flexBasis and the related properties were in that protocol.

Indeed it probably makes more sense to have a property of this nature exposed directly on ASDisplayNode.  Presumably it should go near -calculatedSize / -calculatedLayout and/or -constrainedSize

@rcancro feel free to update this diff with that direction when you have time.  As far as naming, I'm not too sure yet.  I think expectedSize is OK for now but we may make a change just before landing once we see how it comes together.  This is an important API change for the framework so we'd better get it right :)

@nguyenhuy, is there any reason this property should be an ASSizeRange instead of just a CGSize?  I think a CGSize is probably best, but haven't thought about it enough to be sure.

nguyenhuy:
I think CGSize is sufficient because it's the calculated result. That said, I think having an expected layout can be very interesting (and even preferred, since the result from -calculateSizeThatFits: is wrapped into an ASLayout anyway). A static, pre-computed layout can be useful in many other cases.

rcancro:
Just a thought, but do you think it makes sense to do this work in a layout spec instead? I worry about putting a property in a core class like ASDisplayNode when it will only have meaning if the user is taking advantage of the new layout specs.

A new spec could be super simple -- maybe ASStaticSizeLayoutSpec?

```
@implementation ASStaticSizeLayoutSpec

- (instancetype)initWithSize:(CGSize)size child:(id<ASLayoutable>)child
{
    self = [super init];
    if (self) {
        _child = child;
        _size = size;
    }
    return self;
}

- (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
{
    return [ASLayout newWithLayoutableObject:self.child size:ASSizeRangeClamp(constrainedSize, self.size)];
}

@end
```

nguyenhuy:
That's a very good question. Thanks for bringing up! Indeed, there already is a [ASStaticLayoutSpec](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASStaticLayoutSpec.h). The main issue with that layout spec is that it's a little bit complex and takes time to learn.

I think this PR offers a great opportunity for us to discuss an important question, that is how should static/pre-defined/expected layout be supported? An expected size/layout property is simple to implement and understand. On the other hand, ASStaticLayoutSpec is much more powerful but has a steep learning curve. Or maybe we should simplify that spec? Please share your thoughts!

Some other questions related to ASStaticLayoutSpec can be found in #539.

rcancro:
I guess the expectedSize (recommendedSize? desiredSize?) would work in both layoutSpec land and in the previous layout calculation land. I'll put it in ASDisplayNode and update the PR so we can see what that would look like.

I looked at ASStaticLayoutSpec but didn't see how to return a size (just a position). Maybe that is part of the learning curve you are talking about! :) In more general terms, I think that creating ASDisplayNodes that implement a layoutSpec would help with the general learning curve (for example, a ASStackNode that has a similar interface to the new UIStackView).

nguyenhuy:
@rcancro Actually ASStaticLayoutSpec does supports [a size](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASStaticLayoutSpec.h#L26). And it's not a simple CGSize but rather a powerful [ASRelativeSizeRange](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASRelativeSize.h#L24). But again, it takes a bit of time to figure out how to use such power :)  

In KittenNode, a working static spec that wraps the image node and returns an exact size looks like this:
```objective-c
ASRelativeSizeRange imageSize = ASRelativeSizeRangeMakeWithExactCGSize(CGSizeMake(kImageSize, kImageSize));
ASStaticLayoutSpecChild *imageStaticLayoutSpecChild = [ASStaticLayoutSpecChild newWithPosition:CGPointZero
                                                                                layoutableObject:_imageNode
                                                                                            size:imageSize];
ASStaticLayoutSpec *imagePlaceholder = [ASStaticLayoutSpec newWithChildren:@[imageStaticLayoutSpecChild]];
```

Well, not exactly "working". I've just found a bug in the spec itself that prevents it from wrapping its content. A hot fix is commenting out [this line](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASStaticLayoutSpec.mm#L77) and [this line](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASStaticLayoutSpec.mm#L84). I will try to figure out how to best fix this bug later on. In fact, this layout spec is the only one that isn't well tested. So if we decide to keep it, tests will be written :)

nguyenhuy:
> In more general terms, I think that creating ASDisplayNodes that implement a layoutSpec would help with the general learning curve (for example, a ASStackNode that has a similar interface to the new UIStackView).

Do you mean more samples? Yes I think we definitely need more samples (besides existing snapshot test cases).

appleguy:
@nguyenhuy @rcancro haha, funny, I have not looked at all of the latest API names for constraints-based UIKit auto layout...but they may provide some great names to blatantly copy!  Such as "intrinsicContentSize": https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIView_Class/index.html#//apple_ref/occ/instm/UIView/intrinsicContentSize

Let's have some of these names in mind, particularly heading into the 2.0 final version, because we can still change any current API names.  We want to choose names as intuitive as possible to people, and switchers from auto layout to box model will be a pretty common case (I've had a lot of developer friends run into the performance problems of UIKit auto layout, but not want to completely give it up without a substitute — this will provide a nice opportunity for them to get both performance and ease of programming).

appleguy:
Another reference point:

https://developer.apple.com/library/ios/documentation/GraphicsImaging/Reference/CALayer_class/index.html#//apple_ref/occ/instm/CALayer/preferredFrameSize

rcancro:
Oh I like preferredFrameSize. I'll make that update.

@nguyenhuy I meant that it could be useful for us to create an ASStackNode that had an interface like:
```
@interface ASStackNode : ASDisplayNode

- (instancetype)initWithArrangedNodes:(NSArray *)nodes;

- (void)addArrangedSubnode:(ASDisplayNode *)node;
- (void)removeArrangedSubnode:(ASDisplayNode *)node;

- (void)insertArrangedSubnode:(ASDisplayNode *)node atIndex:(NSUInteger)stackIndex;

@property (nonatomic,readonly,copy) NSArray *arrangedNodes;
@property (nonatomic) ASStackLayoutDirection axis;

@property (nonatomic) ASStackLayoutJustifyContent distribution;

@property (nonatomic) ASStackLayoutAlignItems alignment;
@property (nonatomic) CGFloat spacing;

@property (nonatomic, assign) UIEdgeInsets insets;
@end
```
A user could take advantage of this without ever writing any code using `ASStackLayoutSpec`.

nguyenhuy:
I also prefer "preferredFrameSize" :)

@rcancro Sorry for misunderstanding your idea. Here is my 2 cents on the drawbacks of an ASStackNode:
1. It has all the unnecessary properties and methods of ASDisplayNode, most noticeably the backing view or layer.
2. Instead of constructing an ASStack**LayoutSpec**, the root node now has an ASStackNode subnode, which in turn holds some subnodes. For example, KittenNode node has an ASStackNode that has an image and a text node. So, 1 more level added to the node tree (and probably the calculated ASLayout tree).
3. I'm not really sure, but I think the root node still has to be a custom node to correctly describes its layout.
4. Due to above reasons, it would be hard (at least for me) to figure out how to correctly construct the nodes. I mean I want image and text nodes to be stacked horizontally within the KittenNode, but I can't add them directly to KittenNode. Instead, I have to add those nodes to an ASStackNode, then add that stack node to KittenNode. Then, I have to set correct stack properties to the stack node (although they actually describe how those subnodes should be laid out within KittenNode). Oh, and need to override layout method(s) of KittenNode.
5. I think writing layout specs is not that difficult, although it can be because I'm pretty familiar with it.

Anyway, I admit that I might be a bit bias towards the idea of having layout nodes - display nodes that are purely responsible for layout. So, another pair of eyes is very helpful :) And code talks, so if you have the full implementation of ASStackNode with sample usage(s), I would love to read!

rcancro:
@appleguy @nguyenhuy How does the update look?

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -111,7 +111,9 @@ - (instancetype)initWithViewBlock:(ASDisplayNodeViewBlock)viewBlock
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   ASDN::MutexLocker l(_imageLock);
-  if (_image)
+  if (CGSizeEqualToSize(self.preferredFrameSize, CGSizeZero) == NO)
+    return CGSizeMake(MIN(constrainedSize.width, self.preferredFrameSize.width), MIN(constrainedSize.height, self.preferredFrameSize.height));
```
nguyenhuy: I think all ASDisplayNodes must respect the constrained size, so it might make sense to move this logic to ASDisplayNode. Then this method will look like this:

```
{
  ASDN::MutexLocker l(_imageLock);
  if (_image)
    return _image.size;
  else
    return [super calculateSizeThatFits:constrainedSize];
}
```

nguyenhuy:
Despite the above comment, I think this PR is good to land. 

There are 2 small issues I think we might encounter later on:
- After #599 is landed, min size must also be respected. [Here](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASDisplayNode.mm#L1330) should be the best place to do so.
- If a custom node overrides neither -calculateSizeThatFits nor -layoutSpecThatFits, preferredFrameSize should be used. But it won't be because [automatic layout is the default mode](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASDisplayNode.mm#L1329). So with preferredFrameSize, manual layout should be the last resort.

Anyway, these issues can be fixed in another PR.

AsyncDisplayKit/ASImageNode.mm
rcancro: Hmmm...I worry this may not work for the original use case I had. If we have an image node that we set the preferredFrameSize to be 200x200 but we have a local image that is 400x400, the image view will never respect our preferredFrameSize.

What if we flip the logic:

```
{
  ASDN::MutexLocker l(_imageLock);
  if (CGSizeEqualToSize(self.preferredFrameSize, CGSizeZero) == NO)
    return [super calculateSizeThatFits:constrainedSize];

  return _image.size;
}
```

Depending on how network images are cached the same thing could happen.

nguyenhuy: You are right then. The current code should be good to land.

That's said, now that we have preferredFrameSize, I'm wondering if we want to return the image's size anyway. I'm not the author of this piece of code, but I found the discussion in #328 very helpful. Please let me know what you think.

rcancro: Interesting...I never thought this little change would cause such a long discussion! :)

Does this still apply in the new layout scheme? Using a "static layout" (I assume that means setting the node's frame), will have no effect with the new layout spec.

nguyenhuy: haha, I can be annoying sometimes and you are very patient!

Thanks for this very interesting question. Short answer: I believe you have to set the node's frame in -layout method of its parent node. Long answer: With the new layout scheme, there are different ways you can use to force a size on an image node:
- Use ASRatioLayoutSpec and flexBasis. Example [here](https://github.com/facebook/AsyncDisplayKit/blob/master/examples/Kittens/Sample/KittenNode.mm#L138).
- Use ASStaticLayoutSpec.
- Manual layout: reserve enough space for the image node in -calculateSizeThatFits and set its frame in -layout. Example [here](https://github.com/facebook/AsyncDisplayKit/blob/master/examples/Kittens/Sample/KittenNode.mm#L180).

I cannot think of a use case in which the image node has different sizes based on its image's size. Even if there such use case, it's not a common one and should be handled by the user - custom image node. From the framework point of view, I think returning preferredFrameSize should be the default behaviour.

nguyenhuy: Also I think there is no need to clamp preferredFrameSize with constrained size, because it is already done [here](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASDisplayNode.mm#L1331). Sorry, I forgot about it :(

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -185,6 +185,14 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
  */
 @property (nonatomic, readonly, assign) ASSizeRange constrainedSizeForCalculatedLayout;
 
+/**
+ * @abstract Used by some implementations of measureWithSizeRange: to provide an intrisic content size
+ * to a node when one cannot be computed from its subnodes
```
appleguy: As a developer, I literally can't understand what this means in practice.  Consider any improvements to the language here, but I think it would be clear if an example is provided -- add a sentence describing the applicability to images in a grid.  You might say "For example, ASImageNode uses this property to define a default size for itself, without a developer needing to provide a custom layout spec"

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -111,7 +111,9 @@ - (instancetype)initWithViewBlock:(ASDisplayNodeViewBlock)viewBlock
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   ASDN::MutexLocker l(_imageLock);
-  if (_image)
+  if (CGSizeEqualToSize(self.preferredFrameSize, CGSizeZero) == NO)
+    return [super calculateSizeThatFits:constrainedSize];
```
appleguy: Could you add a comment here?  I'm still not totally sure why this works.  You're not passing in preferredFrameSize as the argument to the super method, so I'm not sure where the actual value of preferredFrameSize is used in the ASImageNode case.  Is an alternative to just change the default behavior to lay out at the constrainedSize if there is no image set yet?

appleguy:
This is really cool, thanks for doing all the polish work to land a change to such a core component!

PR #592, opened 2015-07-31
Fix ASTextNodeWordKernerTests for Xcode 7/iOS 9

shannonma:
On Xcode 7 passing nil to a non-null annotated param will error.


shannonma:
Changes are still in progress. Will re-open the PR once everything is ready.

PR #593, opened 2015-07-31
Fixes for ASTextNode test, setNeedsDisplay and font loading failures

shannonma:
This PR fixes three issues:

(f7a4ca4) Compiling the test on Xcode 7 will error as we're passing nil to a non-null annotated param. Fix by passing in the system font.

(15133fd) Crash/assert in ASTextNode on setNeedsDisplay when setting its properties from a background thread after the node has been loaded. Fix by having ASTextNode respect thread-affinity when calling setsNeedsDisplay.

(d5de2a2) Text node to gracefully fail if its unable to load a font specified in the attributed string.


appleguy:
Thanks @shannonma!  Great to see you here, and some nice fixes for the community.

I'm ready to merge this, but need a quick change:
- Leave out the Cocoapod version update.  We had some chronic build failures with version 38, which also broke the internal Pinterest build system.  We're going to wait a while before updating it.
- Thoughts on overriding -setNeedsDisplay on ASTextNode to do the block helper method, rather than wrapping each call to it?  The downside I see here is that it changes the semantics of another class calling -setNeedsDisplay, which could be pretty bad.  I'm leaning towards your solution as-implemented, but curious what you think.

Hopefully, thread affinity will be eliminated before too long!  That could well take until next year, though, so these are important changes.

shannonma:
Hey Scott, great to hear from you! Will revert the Cocoapod version update. Yup I agree with your comments regarding -setNeedsDisplay, I was thinking it could bite us down the road if we changed the semantics. Thanks for the review!

appleguy:
Thanks again, @shannonma !

PR #594, opened 2015-08-02
Support internal relayout

nguyenhuy:
This PR is the first in a series of PRs that aim to support relayout. Relayouts can be categorized into 2 types: external and internal.
- External relayouts are caused by a new constrained size applied to the root node (like orientation change). This can be easily supported by calling -measure on every nodes.
- Internal relayouts are not caused by a different constrained size used on the root node. Instead, they are caused by internal layout changes, like subnodes re-arrangement and/or subnodes' size change.

Internal relayouts are triggers by calling -setNeedsLayout. The method does following things:
- Invalidates calculated layout of the display node.
- If the node is not a root node, -setNeedsLayout will be recursively called on higher-level nodes, until the root node gets the message. This will cause a certain branch in the node tree to be marked as "dirty".
- If the node is the root one, a full measurement pass is executed on the current thread, using the old constrained size. Dirty nodes will then re-measure their layout. [Depends on each scenario](https://goo.gl/photos/N8HsgSsvYf9DVzKL7), clean nodes might reuse their calculated layout.

A layout pass will then be trigger automatically. But depends on the new layout, 2 scenarios will happen:
- If the root node has the same size as before, everything should be fine. A good example is the way image and text nodes in KittenNode are swapped (see -toggleNodesSwap).
- If the root node has a different size and it is attached to a container view (table or collection view, for example), the container view must be notified to redraw. In case of table or collection view, calling beginUpdates and endUpdates is enough (see -toggleImageEnlargement in Kittens example).

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -506,7 +506,18 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
 @interface ASDisplayNode (UIViewBridge)
 
 - (void)setNeedsDisplay;    // Marks the view as needing display. Convenience for use whether view is created or not, or from a background thread.
-- (void)setNeedsLayout;     // Marks the view as needing layout.  Convenience for use whether view is created or not, or from a background thread.
+
+/**
+ * Marks the view as needing layout. Convenience for use whether view is created or not, or from a background thread.
+ * 
+ * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
+ * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
+ * 
+ * Note: If the relayout causes a change in size of the root node that is attached to a container view
+ * (table or collection view, for example), the container view must be notified to redraw. 
+ * In case of table or collection view, calling -beginUpdates and -endUpdates is enough.
+ */
+- (void)setNeedsLayout;
 
 @property (atomic, retain)           id contents;                           // default=nil
 @property (atomic, assign)           BOOL clipsToBounds;                    // default==NO
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -505,10 +505,13 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
  */
 @interface ASDisplayNode (UIViewBridge)
 
-- (void)setNeedsDisplay;    // Marks the view as needing display. Convenience for use whether view is created or not, or from a background thread.
+/**
+ * Marks the view as needing display. Convenience for use whether the view / layer is loaded or not. Safe to call from a background thread.
+ */
+- (void)setNeedsDisplay;
 
 /**
- * Marks the view as needing layout. Convenience for use whether view is created or not, or from a background thread.
+ * Marks the node as needing layout. Convenience for use whether the view / layer is loaded or not. Safe to call from a background thread.
  * 
  * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
  * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -517,8 +517,8 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
  * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
  * 
  * Note: If the relayout causes a change in size of the root node that is attached to a container view
- * (table or collection view, for example), the container view must be notified to redraw. 
- * In case of table or collection view, calling -beginUpdates and -endUpdates is enough.
+ * (table or collection view, for example), the container view must be notified to relayout.
+ * For ASTableView and ASCollectionView, an empty batch editing transaction must be triggered to animate to new row / item sizes.
  */
 - (void)setNeedsLayout;
 
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -560,6 +560,28 @@ void ASDisplayNodePerformBlockOnMainThread(void (^block)())
   [[self asyncLayer] displayImmediately];
 }
 
+- (void)__setNeedsLayout
+{
+  ASDisplayNodeAssertThreadAffinity(self);
+  ASDN::MutexLocker l(_propertyLock);
+  
+  if (!_flags.isMeasured) {
+    return;
+  }
+  
+  ASSizeRange oldConstrainedSize = _constrainedSize;
+  [self invalidateCalculatedLayout];
+  
+  if (_supernode) {
+    // Cause supernode's layout to be invalidated
+    [_supernode setNeedsLayout];
+  } else {
+    // This is the root node. Trigger a full measurement pass on *current* thread. Old constrained size is re-used.
+    [self measureWithSizeRange:oldConstrainedSize];
+    self.frame = CGRectMake(0.0f, 0.0f, _layout.size.width, _layout.size.height);
+  }
+}
+
 // These private methods ensure that subclasses are not required to call super in order for _renderingSubnodes to be properly managed.
 
 - (void)__layout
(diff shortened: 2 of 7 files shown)


AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -506,7 +506,18 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
 @interface ASDisplayNode (UIViewBridge)
 
 - (void)setNeedsDisplay;    // Marks the view as needing display. Convenience for use whether view is created or not, or from a background thread.
-- (void)setNeedsLayout;     // Marks the view as needing layout.  Convenience for use whether view is created or not, or from a background thread.
+
+/**
+ * Marks the view as needing layout. Convenience for use whether view is created or not, or from a background thread.
```
appleguy: Let's rephrase this to be "Marks the node as needing layout.  Convenience for use whether the view / layer is loaded or not.  Safe to call from a background thread."

The original comment likely dates to before layer-backing support.  Now with the expanded abstractions related to layout in AsyncDisplayKit, we can start referring to "node" layout as a more distinct concept than just normal view / layer layout.

nguyenhuy: Sure thing, will update the comment :+1:

nguyenhuy: Done.

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -506,7 +506,18 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
...
+ * 
+ * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
+ * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
+ * 
+ * Note: If the relayout causes a change in size of the root node that is attached to a container view
+ * (table or collection view, for example), the container view must be notified to redraw.
```
appleguy: "redraw" here should probably be "layout".

I am interested in the idea of formalizing the "container view" concept.  We should probably create a delegate API that the root node may implement to manage its interaction with the container.  This could then be used by ASTableView <-> ASCellNode, a new class like ASViewController <-> ASDisplayNode, etc.  If you haven't already done something like this, that's fine because it belongs as a separate effort (and I can probably take on that work).

Maybe revise the next line to say "For ASTableView and ASCollectionView, an empty batch editing transaction will be triggered to animate to new row / item sizes"  (because ASCollectionView doesn't actually have begin/end updates)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -560,6 +560,28 @@ - (void)displayImmediately
   [[self asyncLayer] displayImmediately];
 }
 
+- (void)__setNeedsLayout
+{
+  ASDisplayNodeAssertThreadAffinity(self);
+  ASDN::MutexLocker l(_propertyLock);
+  
+  if (!_flags.isMeasured) {
+    return;
```
appleguy: Is this really the right behavior?  If I call -setNeedsLayout on an un-measured node, it should probably result in a valid layout, even if it must do the measurement on the main thread.  I'm not confident about this, interested in hearing your considerations.

examples/Kittens/Sample/KittenNode.mm
```diff
@@ -29,6 +29,8 @@ @interface KittenNode ()
   ASNetworkImageNode *_imageNode;
   ASTextNode *_textNode;
   ASDisplayNode *_divider;
+  BOOL _isImageEnlarged;
+  BOOL _isNodesSwapped;
```
appleguy: _areNodesSwapped or better yet, _nodesSwapped or _swappedTextAndImage

examples/Kittens/Sample/KittenNode.mm
```diff
@@ -181,4 +184,16 @@ - (void)layout
 }
 #endif
 
+- (void)toggleImageEnlargement
```
appleguy: Great example cases, love it!

examples/Kittens/Sample/ViewController.m
```diff
@@ -102,6 +102,16 @@ - (BOOL)prefersStatusBarHidden
 #pragma mark -
 #pragma mark ASTableView.
 
+- (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
+{
+  [_tableView deselectRowAtIndexPath:indexPath animated:YES];
+  [_tableView beginUpdates];
+  // Assume only kitten nodes are selectable (see -tableView:shouldHighlightRowAtIndexPath:).
```
appleguy: Nice comment, thanks for including it.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -560,6 +560,28 @@ - (void)displayImmediately
...
+  ASSizeRange oldConstrainedSize = _constrainedSize;
+  [self invalidateCalculatedLayout];
+  
+  if (_supernode) {
+    // Cause supernode's layout to be invalidated
+    [_supernode setNeedsLayout];
+  } else {
+    // This is the root node. Trigger a full measurement pass on *current* thread. Old constrained size is re-used.
+    [self measureWithSizeRange:oldConstrainedSize];
+    self.frame = CGRectMake(0.0f, 0.0f, _layout.size.width, _layout.size.height);
```
appleguy: I'm pretty sure this frame setter needs to preserve the origin.  This should work fine in table views, but imagine someone embedding a node in a different view hierarchy.  This would make it jump to the upper left.

Additionally, if there is a transform being applied, setting frame will be undefined.  So let's change this to a bounds setter.

Even the bounds setter needs to preserve the origin, in case it is a scrollable content area.  So if you change this to first read the bounds and only change the size, and set the bounds, I think we'll be good.

appleguy: Important: my most key questions on this diff relate to this method.

- What impact will it have to not call the view /or/ layer's -setNeedsLayout method?
- I think we must include that call.  If the view or layer has not been created yet, then -setNeedsLayout should be set in the pending state.
- If we did not include this call, it seems like nodes that are visited by the -measure pass may never get the new layout actually applied to them.  Particularly if they are not the root node, and if the -measure pass does not touch anything else that may trigger a CoreAnimation relayout.
- I believe omitting the call may break some existing code, so if it is actually essential that we /not/ call it, share your thinking on why.

appleguy:
This is a really exciting change!  Thank you for the two diffs on layout — big developments for the framework :)

AsyncDisplayKit/ASDisplayNode.h
nguyenhuy: Yes, as I already mentioned in my notes, a delegate is best fit for this case. It will facilitate an automatic and flexible relayout model. I haven't implemented it yet and will create a follow-up ticket once this PR is merged. I'm not even sure about how the new API should be, so please take on that work if you have time.

Very good points on the comment. I will update it very soon. 1 tidbit though, I'm gonna write "an empty batch editing transaction **must** be triggered" instead of "will", because without the delegate we still need this to be done in client code.

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: I agree that -setNeedsLayout should always result in a valid layout. But the main problem is that without a prior-measurement, the root node doesn't know its (soon-to-be-old) constrained size. It would be easily solved with a delegate, but as of now, this is the best we can do. (Another way is to can use an unconstrained size, but the calculated layout will be useless most of the time)

examples/Kittens/Sample/KittenNode.mm
nguyenhuy: Thanks :dancer:

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: Nice catch on the bounds setter. I didn't consider it!

Indeed, view/layer -setNeedsLayout is called as always, in [ASDisplayNode+UIViewBridge.mm](https://github.com/facebook/AsyncDisplayKit/pull/594/files#diff-4c96d795186d237f899d33db7bbf4a1aL174).

What do you mean by "then -setNeedsLayout should be set in the pending state"? Do we still need it?

nguyenhuy:
@appleguy I've push a new commit. Please do another round.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -174,6 +174,7 @@ - (void)setNeedsDisplay
 - (void)setNeedsLayout
 {
   _bridge_prologue;
+  [self __setNeedsLayout];
   _messageToViewOrLayer(setNeedsLayout);
```
appleguy: I missed this the first time around.  Now I understand this will still call CA / UIKit.  Whew, we're good to land.

Issue #536, opened 2015-07-03
layout() of ASCellNode not being called when orientation change(traitCollectionChange)


tiepvuvan:
I'm using an ASTableView to display custom ASCellNode sub-class, I override calculateSizeThatFits: and layout()
But when I rotate device, I expect layout() and calculateSizeThatFits: get called to re-layout but they aren't called. Am I missing something or it's a bug of AsyncDisplayKit?

appleguy:
@tiepvuvan thanks for reporting this.  It turns out this is intentionally not done automatically, because relayout passes can be expensive, and are often not what the app developer wants (it is common for re-layouts to occur but result in the same positioning of elements, which can be hard to detect without careful profiling).

To deal with this situation, you should be able to call -invalidateCalculatedSize (if you start using the AsyncDisplayKit 2.0 Beta coming out soon, it is -invalidateLayout).  Then you can call -measure: and -layoutIfNeeded in succession.  One of the easier ways to do this is probably calling -reloadData on the table view.

This is a use case that a number of community members are interested in seeing better support for, so you're not alone!  In particular, PR #468 lays the groundwork for this and I would like to see transparent support for relayouts be part of AsyncDisplayKit 2.0.  Although the Beta is coming very soon (it may make it for today!), the final version may take until late August / shortly before my talk at NSSpain.

appleguy:
@nguyenhuy is going to investigate this within the next couple weeks

appleguy:
@tiepvuvan - thank you for contributing to the AsyncDisplayKit community by asking your question in this forum! I hope you have been able to use the framework! Due to the length of time it’s been since this thread was active, I’m going to mark this issue as closed. Hopefully, I was able to answer your question, but if not, feel free to reopen this.

Issue #512, opened 2015-06-25
Cells do not relayout when edit mode is used in ASTableView


jaichaudhry:
When I try to set the edit mode of ASTableView to true, then the contentView size does not get layout again due to which it shifts to right, and when the minus button of edit mode is pressed the delete  view that comes on the right does not have proper layout.
The reason I think this is happening is because in the following method
- (CGSize)dataController:(ASDataController *)dataController constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
the size that is returned does not considers the change in layout when edit mode is enabled. In actual tableView the contentView width is reduced.
![ios simulator screen shot jun 25 2015 11 53 49 pm](https://cloud.githubusercontent.com/assets/6908626/8362273/806d67a4-1b95-11e5-8378-bdbe2a69c35d.png)
![ios simulator screen shot jun 25 2015 11 53 51 pm](https://cloud.githubusercontent.com/assets/6908626/8362282/8603c1c2-1b95-11e5-9ee4-646e47d1bcc8.png)

jaichaudhry:
I tried this in AsyncDisplayKit 1.2.2 in the Kitten Sample code. Still not working the labels overlapping over the delete button show in edit mode
![ios simulator screen shot jul 6 2015 12 38 54 pm](https://cloud.githubusercontent.com/assets/6908626/8517221/fd0e08ce-23db-11e5-967b-3181d0f435ab.png)

appleguy:
Thanks for reporting this!  We will use this as an example case to improve with the "relayout" work that @nguyenhuy is working on for AsyncDisplayKit 2.0.  The current master branch is now "2.0 Beta" because it includes his extensive changes to the layout system, which will significantly help us deal with this situation.

That said, this is an area where AsyncDisplayKit 1.x requires some manual work from the programmer to set -invalidateCalculatedSize and perform a new -measure / -layoutIfNeeded pass.  One reason is that the relayout could either be done synchronously or asynchronously, and the original framework didn't make an assumption about what is best for all apps when the frame changes (so this behavior forces the app developer to explicitly implement the desired behavior — after calling -invalidate, you can either call -measure on the main thread to block synchronously, or you can call it on a background queue to prepare the new layout asynchronously).

appleguy:
[...]

PR #595, opened 2015-08-02
Support external relayout

nguyenhuy:
External relayout (as defined in #594) can be handled by re-measure all relevant nodes with the new constrained size. This PR is about adding correct hooks that trigger re-measurements in response to some events, such as:
- [x] Orientation change in iOS 8 (#536).
- [x] Orientation change in iOS 7. Since we want ASTableView and ASCollectionView to handle orientation change automatically, I make these views to observe to UIDeviceOrientationDidChangeNotification and schedule a relayout for all loaded cell nodes.
- [x] Cell enters editing mode (#512).

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -15,6 +15,7 @@
 #import "ASDisplayNodeInternal.h"
 #import "ASBatchFetching.h"
 #import "UICollectionViewLayout+ASConvenience.h"
+#import "ASInternalHelpers.h"
 
 const static NSUInteger kASCollectionViewAnimationNone = UITableViewRowAnimationNone;
 
@@ -121,6 +122,8 @@ static BOOL _isInterceptedSelector(SEL sel)
   BOOL _implementsInsetSection;
 
   ASBatchContext *_batchContext;
+  
+  BOOL _pendingRelayoutForAllRows;
 }
 
 @property (atomic, assign) BOOL asyncDataSourceLocked;
@@ -168,6 +171,11 @@ static BOOL _isInterceptedSelector(SEL sel)
   
   [self registerClass:[UICollectionViewCell class] forCellWithReuseIdentifier:@"_ASCollectionViewCell"];
   
+  if (ASSystemVersionLessThan8()) {
+    [[UIDevice currentDevice] beginGeneratingDeviceOrientationNotifications];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(deviceOrientationDidChange) name:UIDeviceOrientationDidChangeNotification object:nil];
+  }
+  
   return self;
 }
 
@@ -177,6 +185,11 @@ static BOOL _isInterceptedSelector(SEL sel)
   // This bug might be iOS 7-specific.
   super.delegate  = nil;
   super.dataSource = nil;
+  
+  if (ASSystemVersionLessThan8()) {
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+    [[UIDevice currentDevice] endGeneratingDeviceOrientationNotifications];
+  }
 }
 
 #pragma mark -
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -168,6 +171,11 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
   
   [self registerClass:[UICollectionViewCell class] forCellWithReuseIdentifier:@"_ASCollectionViewCell"];
   
+  if (ASSystemVersionLessThan8()) {
```
appleguy: Using the math term "less" is harder than the temporal terms "Before" / "After".  Maybe ASSystemVersionBefore8()

Perhaps even better, ASSystemVersionIsBefore(8.0) ?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -463,6 +476,32 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
 
 
 #pragma mark -
+#pragma mark Orientation Change Handling
+
+- (void)traitCollectionDidChange:(UITraitCollection *)previousTraitCollection
+{
+  _pendingRelayoutForAllRows = YES;
```
appleguy: Let's name this _needsRowRelayout

AsyncDisplayKit/ASTableView.mm
```diff
@@ -789,4 +842,23 @@ - (NSUInteger)dataControllerNumberOfSections:(ASDataController *)dataController
...
+  
+  CGSize constrainedSize = [self dataController:_dataController constrainedSizeForNodeAtIndexPath:indexPath];
+  if (state != UITableViewCellStateDefaultMask) {
+    // Edit control or delete confirmation was shown and size of content view was changed.
+    // The new size should be taken into consideration.
+    constrainedSize.width = MIN(cell.contentView.frame.size.width, constrainedSize.width);
+  }
+  
+  [node measure:constrainedSize];
+  node.frame = CGRectMake(0, 0, node.calculatedSize.width, node.calculatedSize.height);
```
appleguy: In an ideal world, this might be done in the UITableViewCell subclass itself — even in the -layoutSubviews method, perhaps.  Not something to worry about now, maybe it would break something

AsyncDisplayKit/Details/ASDataController.h
```diff
@@ -154,6 +154,12 @@ typedef NSUInteger ASDataControllerAnimationOptions;
 
 - (void)reloadRowsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions;
 
+/**
+ * Re-measures all loaded nodes. Used for external relayout (relayout that is caused by a change in constrained size of each and every cell node,
```
appleguy: Rephrase:

"Used to respond to a change in size of the containing view (e.g. ASTableView or ASCollectionView after an orientation change)"

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -555,6 +555,26 @@ - (void)reloadRowsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDat
...
 
+- (void)relayoutAllRows
+{
+  [self performEditCommandWithBlock:^{
+    ASDisplayNodeAssertMainThread();
+    LOG(@"Edit Command - relayoutRows");
+    [_editingTransactionQueue waitUntilAllOperationsAreFinished];
+    
+    NSArray *indexPaths = ASIndexPathsForMultidimensionalArray(_completedNodes);
+    NSArray *loadedNodes = ASFindElementsInMultidimensionalArrayAtIndexPaths(_completedNodes, indexPaths);
```
appleguy: These are quite expensive calls.  Can we avoid them and use iteration over _completedNodes directly?  See other methods in ASDataController that use NSMutableIndexPath, or NSIndexSet

AsyncDisplayKit/Private/ASInternalHelpers.mm
```diff
@@ -61,3 +61,18 @@ CGFloat ASRoundPixelValue(CGFloat f)
 {
   return roundf(f * ASScreenScale()) / ASScreenScale();
 }
+
+BOOL ASSystemVersionLessThan8()
+{
+  static BOOL lessThan8;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    lessThan8 = ASSystemVersionLessThanVersion(@"8");
```
appleguy: Thank you for optimizing this with the static caching.

examples/Kittens/Sample/ViewController.m
```diff
@@ -183,4 +195,19 @@ - (void)tableView:(UITableView *)tableView willBeginBatchFetchWithContext:(ASBat
   });
 }
 
+- (BOOL)tableView:(UITableView *)tableView canEditRowAtIndexPath:(NSIndexPath *)indexPath
+{
+  // Enable editing for Kitten nodes
+  return [_blurbNodeIndexPath compare:indexPath] != NSOrderedSame;
+}
+
+- (void)tableView:(UITableView *)tableView commitEditingStyle:(UITableViewCellEditingStyle)editingStyle forRowAtIndexPath:(NSIndexPath *)indexPath
```
appleguy: Cool!  Really nice to have this added feature to the example.

appleguy:
This is great, @nguyenhuy.  I would like you to make the changes requested here, but will merge the diff now so it is easier for you to continue your development and integrate the changes in any further diff.

AsyncDisplayKit/ASTableView.mm
nguyenhuy: In such case _ASTableViewCell must keep references to the cell node and the table view (to ask for constrained size). More dependencies, I don't like that :)

AsyncDisplayKit/ASCollectionView.mm
nguyenhuy: Yes, you are right. Should check for any changes before setting this flag. Fixed in the new PR.

PR #352, opened 2015-03-04
Section UIEdgeInsets for ASCollectionView

nivanchikov:
`ASCollectionView` uses `self.bounds.size` when providing constrained size for its node cells.

This pull request adds support for section `UIEdgeInsets`, which can be returned in corresponding `UICollectionViewDelegateFlowLayout` method:

```objc
- (UIEdgeInsets) collectionView: (UICollectionView*) collectionView
						 layout: (UICollectionViewLayout*) collectionViewLayout
		 insetForSectionAtIndex: (NSInteger) section
{
	return UIEdgeInsetsMake(30.0f, 15.0f, 30.0f, 15.0f);
}
```

As a result the one of the constrained size dimensions for cell nodes will be reduced by sum of:
- `bottom + top` for height when layout is horizontal
- `left + right` for width when layout is vertical

PR #596, opened 2015-08-03
Add support for sectionInset in ASCollectionView in property and delegate form

levi:
This is something we added internally that I want to bring back upstream. I saw that there was a similar implementation at #352, so I followed the feedback given to that patch and brought it up to speed.

Added a small inset to the example project so people could see it in a action.


appleguy:
Thanks a lot, @levi — this is a nice improvement!  I'm also excited to close the other PR :-D

nguyenhuy:
I promised to take over #352 but haven't had a chance to do it. Thanks for actually doing it, @levi :)

levi:
NP, @nguyenhuy. Thanks for catching that issue with the delegate definition.

Issue #591, opened 2015-07-31
iOS framework target broken


jflinter:
The iOS framework target (and thus, building with Carthage) is currently broken on master. It looks like it's broken for a few reasons:
- `AsyncDisplayKit/Info.plist` was moved to `AsyncDisplayKit-iOS/Info.plist` without updating the `.xcodeproj`
- abe98d5b09b37a70a406a2c8b5e07fb9bcfe66f6 appears to have introduced several compiler errors - I'm sorry, but I don't have time to dig further at the moment.

To reproduce:
`xcodebuild build -project AsyncDisplayKit.xcodeproj/ -target AsyncDisplayKit-iOS`

nickvelloff:
This was merged and had it right: https://github.com/paulyoung/AsyncDisplayKit/commit/8fa6667d93fe164051f6afab70e134c9bb0b0313 
But somehow it subsequently got reverted.

nguyenhuy:
Sorry, I think I messed this up while trying to fix conflicts in #468. Really appreciate if someone familiar with Carthage helps me to fix this.

appleguy:
Thanks for reporting this.  This was the biggest diff landed to ASDK since it was publicly released, and it's a bummer to hear it broke the Carthage target.

I'll eagerly merge any fix.  Right now I don't have cycles to investigate, particularly because I've never used Carthage and don't have a test project, and am rushing to prepare for the ASDK 2.0 launch at NSSpain in a little over a month.

If you'd like to include an examples/ project that uses Carthage, it would be a great way to keep the target working.  The examples/ projects are built by Travis, too.

marmelroy:
I looked into the build issue reported here and just pushed a pull request that should fix it. 

There were several reasons for the framework target failing but it was mostly a matter of the framework missing some recently merged changes (that were only added to the library target).

nguyenhuy:
Awesome. Thank you!

nickvelloff:
@marmelroy thanks for this!

marmelroy:
No worries. 

As @appleguy suggested, an example that uses the framework will help keep the target working in the future. I will write one a little later this week (using Carthage too so that this use case is fully covered).

appleguy:
Excited to see an example or test case!

PR #597, opened 2015-08-03
Fixing build issues for framework target - Issue #591

marmelroy:
I looked into and fixed the recent build issues with AsyncDisplayKit's dynamic framework target (Issue #591).

There were a variety of reasons for this but it turned out to be mostly a matter of the framework missing some recently merged changes (that were only added to the library target).


appleguy:
Thanks!  I'll aim to check the target for future PRs that may affect it.  Any thoughts on how we can automate this would be awesome, too.

marmelroy:
Excellent. To keep the target working, I like your suggestion from #591 for creating an example project that uses the framework (via Carthage so that this use case is fully covered).  

I will write one later this week.

PR #598, opened 2015-08-05
Fix _implementsInsetSection of ASCollectionView being check before the asyncDelegate is set

nguyenhuy:


appleguy:
LOL good catch.

PR #599, opened 2015-08-05
Support different constrained size range per node

nguyenhuy:
Since the return type is ASSizeRange, the min size, ASStackLayoutAlignItems and ASStackLayoutAlignSelf can now be fully utilized. 

One particular use case of min size can be found [here](https://github.com/nguyenhuy/AsyncMessagesViewController/commit/1d409ca8395054caa2cf03acf383ebc6d81a5ac1). In short, I need each cell node to have a positive min width. The min width acts as the min cross size of a vertical stack layout spec. Without a big enough min cross, I won't be able to do right alignment (via ASStackLayoutAlignItemsEnd), because the stack will [hug its content](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Private/ASStackPositionedLayout.mm#L46). (And yes, I've completely replaced manual layout with box-model layout in my AsyncMessagesViewController lib).

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -121,6 +121,7 @@ static BOOL _isInterceptedSelector(SEL sel)
   BOOL _asyncDataFetchingEnabled;
   BOOL _asyncDelegateImplementsInsetSection;
   BOOL _collectionViewLayoutImplementsInsetSection;
+  BOOL _asyncDataSourceImplementsConstrainedSizeForNode;
 
   ASBatchContext *_batchContext;
   
@@ -234,6 +235,7 @@ static BOOL _isInterceptedSelector(SEL sel)
     super.dataSource = nil;
     _asyncDataSource = nil;
     _proxyDataSource = nil;
+    _asyncDataSourceImplementsConstrainedSizeForNode = NO;
   } else {
     _asyncDataSource = asyncDataSource;
     // TODO: Support supplementary views with ASCollectionView.
@@ -242,6 +244,7 @@ static BOOL _isInterceptedSelector(SEL sel)
     }
     _proxyDataSource = [[_ASCollectionViewProxy alloc] initWithTarget:_asyncDataSource interceptor:self];
     super.dataSource = (id<UICollectionViewDataSource>)_proxyDataSource;
+    _asyncDataSourceImplementsConstrainedSizeForNode = ([_asyncDataSource respondsToSelector:@selector(collectionView:constrainedSizeForNodeAtIndexPath:)] ? 1 : 0);
   }
 }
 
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -511,9 +514,21 @@ - (ASCellNode *)dataController:(ASDataController *)dataController nodeAtIndexPat
   return node;
 }
 
-- (CGSize)dataController:(ASDataController *)dataController constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
+- (ASSizeRange)dataController:(ASDataController *)dataController constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
 {
-  CGSize restrainedSize = self.bounds.size;
+  ASSizeRange restrainedSize;
```
appleguy: Call it constrainedSize?

AsyncDisplayKit/ASTableView.h
```diff
@@ -272,6 +272,17 @@
 @optional
 
 /**
+ * Provides the constrained size range for measuring the node at the index path.
+ *
+ * @param tableView The sender.
```
appleguy: Delete these two @param lines and the extra whitespace, so there is just a single empty line between the "Provides ..." and "@returns ..." line.  These particular comments don't add value and make an already large header even bigger.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -108,17 +108,19 @@ - (void)_layoutNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAni
   for (NSUInteger j = 0; j < nodes.count && j < indexPaths.count; j += kASDataControllerSizingCountPerProcessor) {
     NSArray *subIndexPaths = [indexPaths subarrayWithRange:NSMakeRange(j, MIN(kASDataControllerSizingCountPerProcessor, indexPaths.count - j))];
     
-    // TODO: The current implementation does not make use of different constrained sizes per node.
     // There should be a fast-path that avoids all of this object creation.
     NSMutableArray *nodeBoundSizes = [[NSMutableArray alloc] initWithCapacity:kASDataControllerSizingCountPerProcessor];
     [subIndexPaths enumerateObjectsUsingBlock:^(NSIndexPath *indexPath, NSUInteger idx, BOOL *stop) {
-      [nodeBoundSizes addObject:[NSValue valueWithCGSize:[_dataSource dataController:self constrainedSizeForNodeAtIndexPath:indexPath]]];
+      ASSizeRange constrainedSize = [_dataSource dataController:self constrainedSizeForNodeAtIndexPath:indexPath];
```
appleguy: We should definitely optimize this in the near future, for the common case when there is not a different constrainedSize per node.

In most cases, developers will want to use e.g. the width of the ASTableView + CGFloatMax as their constrainedSize coming into measure: of their ASCellNode.  Boxing NSValues and adding to an array, plus all the method calls, is pretty inefficient.

appleguy:
@nguyenhuy this one has merge conflicts, so If you can fix those while addressing the feedback, we're good to land.  Great feature to have — I know a few people (including at Pinterest) who have wanted this!

AsyncDisplayKit/ASCollectionView.mm
nguyenhuy: I was just trying to keep the old name and thus minimize the diff. But constrainedSize is better indeed.

AsyncDisplayKit/ASTableView.h
nguyenhuy: But appledoc will complain :(

nguyenhuy:
@appleguy Fixed conflicts and things according to your feedback. This and #601 whichever lands first will cause new conflicts in the other one. Must be easy to fix though.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -849,14 +858,15 @@ - (void)tableViewCell:(UITableViewCell *)cell atIndexPath:(NSIndexPath *)indexPa
...
   ASCellNode *node = [_dataController nodeAtIndexPath:indexPath];
   
-  CGSize constrainedSize = [self dataController:_dataController constrainedSizeForNodeAtIndexPath:indexPath];
+  ASSizeRange constrainedSize = [self dataController:_dataController constrainedSizeForNodeAtIndexPath:indexPath];
   if (state != UITableViewCellStateDefaultMask) {
     // Edit control or delete confirmation was shown and size of content view was changed.
     // The new size should be taken into consideration.
-    constrainedSize.width = MIN(cell.contentView.frame.size.width, constrainedSize.width);
+    constrainedSize.min.width = MIN(cell.contentView.frame.size.width, constrainedSize.min.width);
+    constrainedSize.max.width = MIN(cell.contentView.frame.size.width, constrainedSize.max.width);
```
nguyenhuy: Oh and it's worth to mention that both min and max sizes must be resolve here. I'm pretty about the math, but I might be wrong!

PR #601, opened 2015-08-09
Clean up the previous external relayout PR

nguyenhuy:
#595


nguyenhuy:
The current approach for doing relayout after orientation change is wrong. Gonna make a new PR with better solution.

PR #602, opened 2015-08-10
Fix race in ASDisplayNodeRespectThreadAffinityOfNode

shannonma:
We must hold the lock for the duration of the block.
Failure to do this can result in the node loading while the block is inflight.


shannonma:
Sorry about the spacing changes, seems my text editor did that behind my back. If its an issue let me know and I can revert them :).

appleguy:
Cool, thank you!  Spacing changes look good :).

PR #603, opened 2015-08-11
Improve relayout when constrained size of all nodes is changed

nguyenhuy:
- In -layoutSubviews of table and collection views, detect changes that cause a different constrained size for nodes, and trigger relayout immediately.
- Orientation change can be handled by this solution. So, no need to observe to its events.
- Clean up #595.
- Update Kittens example to support iPad (easier to catch bugs on these devices) and add a title to navigation bar (looks a bit nicer).

There is 1 bug left that hopefully I will be able to fix in the next few days:
- Invisible nodes are not re-measured after entering editing mode.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -564,15 +564,14 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
     LOG(@"Edit Command - relayoutRows");
     [_editingTransactionQueue waitUntilAllOperationsAreFinished];
     
-    NSArray *indexPaths = ASIndexPathsForMultidimensionalArray(_completedNodes);
-    NSArray *loadedNodes = ASFindElementsInMultidimensionalArrayAtIndexPaths(_completedNodes, indexPaths);
-    
-    for (NSUInteger i = 0; i < loadedNodes.count && i < indexPaths.count; i++) {
-      ASSizeRange constrainedSize = [_dataSource dataController:self constrainedSizeForNodeAtIndexPath:indexPaths[i]];
-      ASCellNode *node = loadedNodes[i];
-      [node measureWithSizeRange:constrainedSize];
-      node.frame = CGRectMake(0.0f, 0.0f, node.calculatedSize.width, node.calculatedSize.height);
-    }
+    [_completedNodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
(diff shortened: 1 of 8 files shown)


nguyenhuy:
Conflicts fixed.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -564,15 +564,14 @@ - (void)relayoutAllRows
...
-    NSArray *indexPaths = ASIndexPathsForMultidimensionalArray(_completedNodes);
-    NSArray *loadedNodes = ASFindElementsInMultidimensionalArrayAtIndexPaths(_completedNodes, indexPaths);
-    
-    for (NSUInteger i = 0; i < loadedNodes.count && i < indexPaths.count; i++) {
-      ASSizeRange constrainedSize = [_dataSource dataController:self constrainedSizeForNodeAtIndexPath:indexPaths[i]];
-      ASCellNode *node = loadedNodes[i];
-      [node measureWithSizeRange:constrainedSize];
-      node.frame = CGRectMake(0.0f, 0.0f, node.calculatedSize.width, node.calculatedSize.height);
-    }
+    [_completedNodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
```
appleguy: This looks a lot nicer.  I'd actually be really curious to profile both versions on a device.  Thinking about this, I'm not 100% confident which will be faster.  My bet is on the new version, but because of the individual creation of NSIndexPaths, it might not be by much (main win is avoiding the array allocations).

Someday in the magical distant future, it would be sweet to banish NSIndexPath and use ASIndexPath everywhere :).  Not important now, but it may well be in a few months with the new use cases starting to expose this code to large data sets.

appleguy:
Love this diff, thanks!

AsyncDisplayKit/Details/ASDataController.mm
nguyenhuy: While I should profile these versions, I actually didn't :( My main motivation for this diff was about readability/maintainability because of its infrequent usage. Regarding performance, my assumption was that enumerate existing array should be faster than array allocations and finding elements.

I've just done profiling and the new version is actually much faster. I ran -relayoutAllRows 10,000 times on the same constrained size, so cached layouts were used and there was no actual layout calculation. The old version took 610ms. Almost 40% of that time was spent on ASIndexPathsForMultidimensionalArray and ASFindElementsInMultidimensionalArrayAtIndexPaths. The new version took 419ms. And creating NSIndexPaths only took 0.2% of that time. On both versions, one of the most expensive operations is ASDisplayNode's setFrame, which is unavoidable I guess.

old_version

new_version

PR #608, opened 2015-08-13
Fix small bugs related to preferredFrameSize

nguyenhuy:
- Fix bug that causes preferred size to not be used if it is set after a layout pass.
- Fix bug that causes preferred size to not be used if -calculateSizeThatFits is not overridden.


rcancro:
Thanks! :)

nguyenhuy:
You're welcome :)

Issue #606, opened 2015-08-13
ASStackLayoutSpec ASTextNode with nil string returning non-zero height


SRandazzo:
Hi ASDK!

I have an implementation question about using `ASStackLayoutSpec` in a vertical context

I'm stacking a few `ASTextNode`s, and was expecting that if one of the text node's attributedText was nil, the height should calculate as zero, resulting in a "stack" with no spaces in between elements.

Instead, the label is returning a width of zero, but height of 14, leaving a 14 px gap in between my resulting layout

Was wondering if this was expected or not?

using the latest master

Thanks!

<img width="88" alt="screen shot 2015-08-12 at 8 49 58 pm" src="https://cloud.githubusercontent.com/assets/307819/9240394/b60196f6-4133-11e5-8e48-4afc8e08cebf.png">

SRandazzo:
I should add that when you switch the layout to `Horizontal`, it lays out as I would expect, since the size for the label is calculated to be `{ 0, 14 }`

<img width="208" alt="screen shot 2015-08-12 at 9 38 57 pm" src="https://cloud.githubusercontent.com/assets/307819/9240941/a35041e0-413a-11e5-8463-446671d3691d.png">

appleguy:
@SRandazzo — this is a curious finding.  I would expect this in a way, in that the height is presumably determined by font size (you could test this by changing the font size.  However, that seems more like an implementation detail than really the correct behavior for this case.

It would be an easy thing to fix, particularly for a nil string (even a @"" empty string might be more plausibly nonzero height, but that's harder to decide!)  Could you run a quick test with UILabel and see what its sizeToFit does with a nil string, with an explicit font set?  I assume it is 0, 0.

Feel free to submit a diff to check for a zero-length string (which would catch nil, too) in the size/layout methods of ASTextNode.  If you don't have a chance to do this, that's fine; I'll make sure it gets into an upcoming release.

SRandazzo:
@appleguy Thanks!

 Threw this into a playground to test what `sizeToFit` does on a label with an empty string text

The result is that calling sizeToFit on a UILabel with an empty strings results in CGSizeZero

<img width="870" alt="screen shot 2015-08-13 at 2 44 59 pm" src="https://cloud.githubusercontent.com/assets/307819/9258510/f106a7f0-41c9-11e5-8add-497e8370af2d.png">

SRandazzo:
[...]

PR #610, opened 2015-08-13
606 sizezero nil text

SRandazzo:
@appleguy In reference to issue #606 

This modifies the ASTextNodeRenderer to set `_calculatedSize` to `CGSizeZero` when the attributedString is either nil or an empty string

I've also added test coverage for these expectations :sunglasses: 

I thought this was the most logical place, but if you think it belongs else where please let me know!

### DIFF
--- a/AsyncDisplayKit/Details/ASTextNodeRenderer.mm
+++ b/AsyncDisplayKit/Details/ASTextNodeRenderer.mm
@@ -173,8 +173,12 @@ static const CGFloat ASTextNodeRendererTextCapHeightPadding = 1.3;
 {
   ASDN::MutexLocker l(_textKitLock);
 
+  if (_attributedString.string.length == 0) {
+    _calculatedSize = CGSizeZero;
+    return;
+  }
+    
   [self _initializeTextKitComponentsIfNeeded];
-
 
   // Force glyph generation and layout, which may not have happened yet (and
   // isn't triggered by -usedRectForTextContainer:).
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Details/ASTextNodeRenderer.mm
```diff
@@ -173,9 +173,13 @@ - (void)_calculateSize
 {
   ASDN::MutexLocker l(_textKitLock);
 
+  if (_attributedString.string.length == 0) {
```
appleguy: Small change, but could you make sure this checks _attributedString.length directly?

I'm not completely sure if the .string property always has an NSString allocated & ready (it might), but because this is an essential codepath that every text label in ASDK apps will hit, it's worth being cautious and minimal.

Thanks for removing the extraneous space :)

appleguy:
@SRandazzo thank you!  This is a great find, and the fix is very well done.  It's awesome to see test coverage continue to increase around the framework, and it wouldn't be happening without changes like this.

I'm happy to merge this as-is, but if you have 2 minutes to upload the change with .length called directly on the attributed string, that would be a little better.

SRandazzo:
@appleguy Thanks for taking a look! I agree with your suggestion to use the attributedString's length directly. It did give me a chance to read through the docs which says that length returns the underlying string object's length! But of course that's Apple's implementation details, so using the attributedString's length makes more sense imo :+1: 

thanks again!

appleguy:
Success.  Thanks again, glad to have you in the community @SRandazzo :).  Don't hesitate to let me know if I can help in any way.

SRandazzo:
Woo! Thanks a ton!

PR #611, opened 2015-08-15
Added baseline alignment for ASStackLayoutSpec

rcancro:
As `ASLayoutable` was beginning to get crowded with stack-specific properties, I created a new protocol `ASStackLayoutable` that implements `ASLayoutable` and moved all stack-specific properties to this protocol.

I also created a 3rd layoutable protocol called `ASBaselineStackLayoutable`. It contains two properties that allow it to layout horizontally and vertically on a baseline:

```
/**
 *  @abstract Used for baseline alignment. The distance from the top of the object to its baseline.
 */
@property (nonatomic, readwrite) CGFloat ascender;

/**
 *  @abstract Used for baseline alignment. The distance from the baseline of the object to its bottom.
 */
@property (nonatomic, readwrite) CGFloat descender;
```

Finally I created a new layout spec `ASBaselineStackLayoutSpec` which lays out children that conform to `ASBaselineStackLayoutable`, which are currently only `ASBaselineStackLayoutSpec` and `ASTextNode`.

Here are some screenshots of it in action:

Baseline Alignment in horizontal stack:
align baseline

Baseline Alignment with wrapping text aligning to first baseline:
align first baseline

Baseline Alignment with wrapping text aligning to last baseline:
align last baseline

Vertical baseline (the red boxes end at the baseline. The spacing between the bottom of the red box to the top of the next red box are consistent):
screen shot 2015-08-15 at 10 01 37 am

Vertical baseline alignment with 2 stack layouts (the spacing measurement is between the two text nodes with gray backgrounds):
screen shot 2015-08-15 at 9 32 41 am

Notice that the vertical alignment works even with 2 horizontal stack layouts. If a stack spec contains a node with an ascender/descender, I pass this value on to the spec's ascender/descender. Then, if this layout spec is used as a child in another layout spec the baseline will still be respected. 

Please provide feedback on the implementation! I'm looking forward to merging this in!!


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -355,6 +355,9 @@ - (void)setAttributedString:(NSAttributedString *)attributedString {
       self.isAccessibilityElement = YES;
     }
   });
+  
+  self.ascender = round([[attributedString attribute:NSFontAttributeName atIndex:0 effectiveRange:NULL] ascender] * [UIScreen mainScreen].scale)/[UIScreen mainScreen].scale;
```
appleguy: Definitely cache the scale:

CGFloat scale = [UIScreen mainScreen].scale;

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
```diff
@@ -47,6 +47,10 @@ typedef NS_ENUM(NSUInteger, ASStackLayoutAlignItems) {
   ASStackLayoutAlignItemsCenter,
   /** Expand children to fill cross axis */
   ASStackLayoutAlignItemsStretch,
+  /** Children align to their first baseline. Only available for horizontal stack nodes */
+  ASStackLayoutAlignItemsBaselineFirst,
```
appleguy: To be clear, the first / last thing could only affect text nodes, right?  ASLayoutable has the baseline defined in a direct / literal sense, without a concept of multiple lines or wrapping.  Would it make sense to simplify this to AlignItemsBaseline, and then have logic / properties on ASTextNode that controls whether it a) sets nothing, b) sets to the first baseline, c) sets to the last baseline?

appleguy:
This is really exciting, especially how the baseline adjustment can propagate up to the stack spec for higher level usage.  I think we're pretty close here.  Would like Huy to take a look, and consider my comments.

AsyncDisplayKit/ASTextNode.mm
nguyenhuy: FYI, there is [ASScreenScale()](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Private/ASInternalHelpers.h#L19) that is already used in other places.

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
nguyenhuy: Yes, I second this. Simplifying these things to AlignItemsBaseline will probably make the API a bit easier to learn. Also, the implementation of ASStackPositionedLayout might be more generic and a bit simpler.

What I'm thinking about is this:
- Instead of ascender and descender properties, ASLayoutable would have a baselineDistance. If it holds a positive value, it equals to ascender. A negative value means descender. (Or maybe the other way around, depends on which way you think is easier to understand)
- Similar to ascender and descender, baselineDistance is 0 for all objects excepts ASTextNode.
- ASTextNode has ascender, descender and a first/last alignment flag. Its baselineDistance can be computed from these properties.
- ASStackPositionedLayout can simply ask for baselineDistance of each item and position it accordingly.

There must be some problems that I overlooked, so please take this comment with a grain of salt :)

appleguy:
Yeah actually, I agree with Huy.  The ascender / descender thing in ASLayoutable didn't sit perfectly with me.  I think the right abstraction is some form of verticalOffset / baselineOffset property.  verticalOffset is the most general, which is great especially for something that applies to ALL nodes and specs...not just text nodes, and not just stack specs.

verticalOffset does tempt a different direction, though, something like a UIEdgeInsets.  That would capture the ascender / descender model, and it could be layoutOffset or something.

Interested in your thoughts Ricky — is there a compelling case for ascender and descender separately?  I know you're a Swift fan; it would be nice to have protocol extensions to support ASLayoutable evolving into more domain-specific areas without too much pollution of the common protocol.

A key reason to avoid putting too much stuff into it is that engineers new to the framework will often balk / fail to understand the abstraction if it has a ton of stuff, which is already a problem with ASLayoutable considering the stack-specific elements.  @nguyenhuy let us know if you have any ideas for how this could be improved (even if you can't do the change), because we only have about a month before we can't easily change it.

nguyenhuy:
Well if I understand FlexBox baseline correctly, it is determined from the content. So it makes sense to allow only text nodes to have non-zero baselineOffset.

verticalOffset is also very interesting. I guess we can implement it in another PR because the purpose of this PR is supporting baseline (or maybe not?).

Also, I think having protocol extensions is a great idea. I can put stack-specific elements to a new extension of ASLayoutable, should I? Please let me know anything else I can help to improve the API. I will do as much as I can :)

nguyenhuy:
After thinking again, I figured that the existing ASInsetLayoutSpec can be used to provide vertical and horizontal offsets, either by wrapping the whole stack spec or its children. So we may not need to implement verticalOffset, unless there is a special use case that inset spec can't help (and I doubt that).

On the other hand, baselineOffset is more difficult to calculate, since you need to compute the max baseline and position children accordingly.

rcancro:
Hey guys,

I just copied the BaselineFirst/BaselineLast idea from UIStackView. For anything but text nodes it seems pretty worthless to me (imagine an image node that has its top aligned with the baseline of a text node) but I wasn't sure that was a good enough reason to not include it. I'm not married to ascender/descender, and can give it a shot with `distanceToBasline` or something similar.

To tell you the truth, for what I need the layout to do, a layoutInsets would be the most helpful. I  actually have another branch where I'd already started working on that idea. ASInsetSpec didn't work for me because I wanted text node subclasses (it is a long, long story), but couldn't return an inset layout spec because a subclass can only implement 1 of the 3 layout methods (is that something we can change?). So I would still be in favor of layoutInsets in ASLayoutable.

But I'm not sure that one needs to replace the other, and contend that both are useful. First of all, if I were looking for a way to align stack children to a baseline I would not immediately think to use layout insets. Secondly, the baseline may not be where you'd want to measure vertical spacing. Finally, I've already implemented ASCenterLayout and ASInsetLayout to honor the layoutInsets so it could be something that is valid for more than just a stack spec (where as baseline would not be).

I agree with @appleguy that we don't want to scare people off by making ASLayoutable too large, but I also think that once people start using stack specs the extra options will be appreciated. So perhaps making an ASStackLayoutable protocol to separate all these options is the right solution. Would we end up changing all classes that implement ASLayoutable then implement ASStackLayoutable? Unless there is an obvious layout spec that should never be a child of a stack layout?

Let me know and I'll give these changes a go and get an updated PR up soon! Thanks for the input!!

nguyenhuy:
> For anything but text nodes it seems pretty worthless to me... but I wasn't sure that was a good enough reason to not include it.

I think it is good enough. From users point of view, it would be quite confusing to see both BaselineFirst and BaselineLast options for, say, an image node. And both work the same in this case.

I'm very interested in your story actually. The only use case I can think of in which ASInsetLayoutSpec doesn't work is that it's hard to compute the insets outside of the text node.

> a subclass can only implement 1 of the 3 layout methods (is that something we can change?)

Well, I think the absolutely strict rule is that a node shouldn't implement both -calculateSizeThatFits: and -layoutSpecThatFits:, because -calculateSizeThatFits: (or rather any method that accepts CGSize constrained size) doesn't take min size into account and should be considered deprecated. Maybe it is the right time to migrate ASTextNode to 1 of the 2 ASSizeRange layout methods? 1 possible way is to rename ASTextNode's -calculateSizeThatFits to -renderSizeThatFits:, and make it available to ASTextNode+subclasses only. ASTextNode's -calculateLayoutThatFits will simply return a layout with the render size, and it is open for extension. 

So, assume that the insets can only be computed within your custom text node, I can make a quick PR to do the above thing and hopefully it will facilitate your custom inset logic.

Anyway, I think that baseline alignment and ASInsetLayoutSpec are enough for most use cases. And to keep the API simple, it should have only those 2 features. Any other special use cases should be implemented using a custom node and inset logic. 

Please let me know if I'm talking non-sense and should just shut up haha.

rcancro:
I'm not wild about putting a flag in ASTextNode to determine baseline alignment since it only gets honored by ASStackLayoutSpec. I going to try keeping baseline first and last as ASStackLayoutAlignItems enums and support both with just the distanceToBaseline property in ASLayoutable.

rcancro:
Ok, big update. 

I split `ASLayoutable` into `ASLayoutable` and `ASStackLayoutable`. I created another protocol  `ASStackTextLayoutable` that is based on `ASStackLayoutable`.  All of the stack-only properties from `ASLayoutable` are now in `ASStackLayoutable`. `ASStackTextLayoutable` adds the `ascender` `descender` properties. 

I created a new layout spec called `ASStackTextLayoutSpec` that only lays out `ASStackTextLayoutable` children. Currently the only classes that implement this protocol are `ASTextNode` and `ASStackTextLayoutableSpec`. 

I'd like to look at `ASStackTextPositionedLayout::compute` and see if I can remove some of the duplicated layout logic from `ASStackPositionedLayout::compute`, but for the time being I was anxious  to get eyes on this. I think it is a pretty great compromise -- we get baseline alignment without polluting `ASLayoutable`, `ASLayoutSpec` and `ASDisplayNode` with properties that don't make sense.

rcancro:
@nguyenhuy @appleguy thoughts?

nguyenhuy:
I guess we can remove some code in ASBaselineStackPositionedLayout. But overall, I think this looks good. I will take a closer look tomorrow morning, it's almost midnight over here :)

appleguy:
Thanks @rcancro!  Lots of great patches recently.  I will review this one over the weekend, most likely tomorrow.  Interested in @nguyenhuy's thoughts too, particularly because I'm not as expert in this code as either of you :).

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
```diff
@@ -58,6 +20,10 @@ typedef struct {
   ASStackLayoutJustifyContent justifyContent;
   /** Orientation of children along cross axis */
   ASStackLayoutAlignItems alignItems;
+  /** 
+   If YES the vertical spacing between two views is measured from the last baseline of the
+   top view to the top of the bottom view*/
+  BOOL baselineRelativeArrangement;
```
nguyenhuy: Do we still need this? I don't think it's used anywhere. Even if it is needed, I think it belongs to ASBaselineStackLayoutSpec.

AsyncDisplayKit/Layout/ASBaselineStackLayoutSpec.mm
```diff
@@ -0,0 +1,81 @@
...
+  
+  const CGSize finalSize = directionSize(stackStyle.direction, unpositionedLayout.stackDimensionSum, baselinePositionedLayout.crossSize);
+  
+  NSArray *sublayouts = [NSArray arrayWithObjects:&baselinePositionedLayout.sublayouts[0] count:baselinePositionedLayout.sublayouts.size()];
+  
+  ASDN::MutexLocker l(_propertyLock);
+  _ascender = baselinePositionedLayout.ascender;
+  _descender = baselinePositionedLayout.descender;
+  
+
```
nguyenhuy: Remove extra line?

AsyncDisplayKit/Private/ASBaselineStackPositionedLayout.mm
```diff
@@ -0,0 +1,118 @@
...
+
+#import "ASBaselineStackPositionedLayout.h"
+
+#import "ASLayoutSpecUtilities.h"
+#import "ASStackLayoutSpecUtilities.h"
+
+static CGFloat baselineForItem(const ASBaselineStackLayoutSpecStyle &style,
+                               const ASLayout *layout) {
+  
+  __weak id<ASBaselineStackLayoutable> textChild = (id<ASBaselineStackLayoutable>) layout.layoutableObject;
```
nguyenhuy: "textChild"?

AsyncDisplayKit/Private/ASBaselineStackPositionedLayout.mm
```diff
@@ -0,0 +1,118 @@
...
+  const CGFloat crossSize = MIN(MAX(minCrossSize, largestChildCrossSize), maxCrossSize);
+  
+  // find the child with the largest height. Use that child's descender as the descender to pass back to the ASBaselineStackLayoutSpec.
+  const auto descenderIt = std::max_element(stackedChildren.begin(), stackedChildren.end(), [&](const ASLayout *a, const ASLayout *b){
+    return  a.position.y + a.size.height <  b.position.y + b.size.height;
+  });
+  const CGFloat minDescender = descenderIt == stackedChildren.end() ? 0 : ((id<ASBaselineStackLayoutable>)(*descenderIt).layoutableObject).descender;
+
+  return {stackedChildren, crossSize, maxAscender, minDescender};
+}
```
nguyenhuy: Not gonna pretend that I fully understand the logic here, but it looks like you managed to reuse ASStackPositionedLayout and compute new positions and size based on its result. That's really smart! :+1:

AsyncDisplayKit/Private/ASBaselineStackPositionedLayout.mm
```diff
@@ -0,0 +1,118 @@
...
+static CGFloat maxDimensionForLayout(const ASLayout *l,
+                                     const ASStackLayoutSpecStyle &style)
+{
+  CGFloat maxDimension = crossDimension(style.direction, l.size);
+  style.direction == ASStackLayoutDirectionVertical ? maxDimension += l.position.x : maxDimension += l.position.y;
+  return maxDimension;
+}
+
+ASBaselineStackPositionedLayout ASBaselineStackPositionedLayout::compute(const ASStackPositionedLayout &positionedLayout,
+                                                                 const ASBaselineStackLayoutSpecStyle &textStyle,
```
nguyenhuy: "textStyle"?

AsyncDisplayKit/Private/ASBaselineStackPositionedLayout.mm
```diff
@@ -0,0 +1,118 @@
...
+                                                                 const ASSizeRange &constrainedSize)
+{
+  ASStackLayoutSpecStyle stackStyle = textStyle.stackLayoutStyle;
+  
+
```
nguyenhuy: Remove extra line, maybe?

AsyncDisplayKit/Private/ASStackPositionedLayout.mm
```diff
@@ -27,9 +27,12 @@ static CGFloat crossOffset(const ASStackLayoutSpecStyle &style,
     case ASStackLayoutAlignItemsStart:
     case ASStackLayoutAlignItemsStretch:
       return 0;
+    default:
+      return 0;
```
nguyenhuy: Looks like you don't need default value because all cases are already handled.

AsyncDisplayKit/Private/ASStackPositionedLayout.mm
```diff
@@ -27,9 +27,12 @@ static CGFloat crossOffset(const ASStackLayoutSpecStyle &style,
...
   }
 }
 
+
```
nguyenhuy: Why?

nguyenhuy:
Besides some nits, I think the diff is great. I'm now a strong advocate of the new ASBaselineStackLayoutable and ASBaselineStackLayoutSpec :)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -453,6 +453,11 @@ - (BOOL)isLayerBacked
 
 #pragma mark -
 
+- (CGFloat)distanceToBaseline:(ASStackLayoutAlignItems)baselineAlignmentType
+{
+  return 0;
+}
+
```
nguyenhuy: Probably a legacy of the old commits :)

AsyncDisplayKit/Private/ASBaselineStackPositionedLayout.mm
rcancro: I added lots of comments. Hopefully it makes more sense now! :)

rcancro:
Ugh, I did something to anger git. 

Also changed all the ASBaselineStack* names to ASBaseline*.

appleguy:
Damn, there is a ton of work here.  Thanks so much for the considerate attention to detail, @rcancro!

We may revisit some specifics on how the ASLayoutable protocol is handled, but these can be done in the next couple weeks leading up to the 2.0 announcement.

AsyncDisplayKit/Private/ASBaselineStackPositionedLayout.mm
nguyenhuy: I'm sorry for this late response. But the comments are helpful indeed. Thanks!

robmontesinos:
@rcancro I love the screenshots and baseline alignment is very exciting for a nicer looking UI. Can you post some sample code of how to apply? For example I have a button with a label next to it - I would like to align these two nodes at center line - code below:

```ObjectiveC
    ASStackLayoutSpec *favoriteHStack = [[ASStackLayoutSpec alloc] init];
    favoriteHStack.direction = ASStackLayoutDirectionHorizontal;
    favoriteHStack.children = @[self.favoriteButtonNode, self.favoriteCountNode];
```

Would I be replacing ASStackLayoutSpec  with ASBaseline and apply the alignment desired?

robmontesinos:
I got my aligning issues solved with alignSelf and alignItems

'''ObjectiveC
    ASStackLayoutSpec *favoriteHStack = [[ASStackLayoutSpec alloc] init];
    favoriteHStack.direction = ASStackLayoutDirectionHorizontal;
    favoriteHStack.children = @[self.favoriteButtonNode, self.favoriteCountNode];
    favoriteHStack.alignSelf = ASStackLayoutAlignSelfCenter;
    favoriteHStack.alignItems = ASStackLayoutAlignItemsCenter;

appleguy:
@robmontesinos awesome!  Yes, these are pretty powerful capabilities from a text system perspective!

PR #613, opened 2015-08-17
ASDisplayNode: Perform one-time initializations in +initialize when possible

eanagel:
This PR aims to improve the performance of `ASDisplayNode -init` by offloading invariant tasks to `+initialize`. This is done in a thread-safe manner by adding a method to each subclass of `ASDisplayNode` (and `ASDisplayNode` itself) which sets `_flags` and `_methodOverrides` with pre-calculated values. (Previously we were doing this work on each init.) Because `+initialize` is guaranteed to run before any other methods in the class, I believe it is thread-safe and reasonably goof-proof.

Of course due to the dynamic nature of Objective C, this can be broken. In fact, my initial implementation was broken immediately by our own tests which use a dynamic implementation of respondsToSelector for it's tests. I've added code that handles this case (degrading to the old functionality if it detects that `+respondsToSelector` or `-respondsToSelector` has been overridden.), but there are still a couple ways a really creative developer could break things: by adding methods we are testing using `class_addMethod` and friends, or by changing the return value of the respondsToSelector results after an instance has been created (which would break the current implementation as well.) These seem like pretty unlikely scenarios to me.

Since this does impact each and every instance of a node, it probably deserves some close inspection before we merge it in. I've tried to structure things so future developers can add to it without getting into the weeds too much.

Note: @appleguy I' did a trial merge into 1.2.3 to test in my application and found it pretty challenging because the code bases are shifting apart a bit here. It's doable, but I found in the end it's a little closer to a re-implementation than a merge ;) Happy to help with that if and when the time comes.


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -85,25 +87,94 @@ void ASDisplayNodeRespectThreadAffinityOfNode(ASDisplayNode *node, void (^block)
...
+ *  @param c        the class, required
+ *  @param instance the instance, which may be nil. (If so, the class is inspected instead)
+ *
+ *  @return ASDisplayNode flags.
+ */
+static struct ASDisplayNodeFlags GetASDisplayNodeFlags(Class c, ASDisplayNode *instance) {
+  struct ASDisplayNodeFlags flags = {0};
+
+  flags.isInHierarchy = NO;
+  flags.displaysAsynchronously = YES;
```
appleguy: Should we have assertions about the class being non-nil?  What if I passed in an ASScrollNode or something else that does not display asynchronously?  How about a developer creating something with initWithViewBlock/LayerBlock:, which also results in NO for displaysAsynchronously?

eanagel: I'll add a nil check for c just to be safe. (I didn't bother since I made it a static function and knew the context where it could be called.)

The values here are only defaults that were originally set in _initializeInstance and then overridden in either a subclasses' init method or after _initializeInstance runs (as in initWithViewBlock/LayerBlock) so I think we are good there.

appleguy:
@eanagel Thanks again for this :).  What did the numbers look like, even if you only measured before the patch?  Roughly what magnitude of costs are you seeing when scrolling (ms, % of main thread)?

I'm curious because it could help prioritize the work to move node initialization off the main thread, which is where it has always belonged!  Of course, regardless of the thread, optimization is appreciated.

I'll take a closer look at this over the weekend and try to get it merged.

appleguy:
Wow, it's passing now.  How odd.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -85,25 +87,98 @@ void ASDisplayNodeRespectThreadAffinityOfNode(ASDisplayNode *node, void (^block)
...
 
-+ (void)initialize
-{
-  if (self == [ASDisplayNode class]) {
-    return;
+/**
+ *  Returns ASDisplayNodeFlags for the givern class/instance. instance MAY BE NIL.
+ *
+ *  @param c        the class, required
+ *  @param instance the instance, which may be nil. (If so, the class is inspected instead)
```
appleguy: It is unclear to me why the instance parameter is ever desired.  Can you remove it, or update the comment to be clearer on why it exists?  Since the class is required and the behavior seems to be correct in the case that no instance is provided, why would we ever provide an instance?  Namely, -instancesRespondToSelector: seems to be sufficient.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -123,38 +198,15 @@ + (Class)layerClass
 
 #pragma mark - Lifecycle
 
+- (void)_staticInitialize {
```
appleguy: Match surrounding brace style.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -85,25 +87,98 @@ void ASDisplayNodeRespectThreadAffinityOfNode(ASDisplayNode *node, void (^block)
...
-    return;
+/**
+ *  Returns ASDisplayNodeFlags for the givern class/instance. instance MAY BE NIL.
+ *
+ *  @param c        the class, required
+ *  @param instance the instance, which may be nil. (If so, the class is inspected instead)
+ *
+ *  @return ASDisplayNode flags.
+ */
+static struct ASDisplayNodeFlags GetASDisplayNodeFlags(Class c, ASDisplayNode *instance) {
```
appleguy: These two functions should have open brace on newline.

appleguy:
@eanagel This is really, really cool.  Although far from the largest, it is one of the most detailed and technical patches we have received from the community, and I really appreciate your thoughtful analysis and optimization.

I'm comfortable landing this as-is, and will do so if you don't have time to make the tweaks requested in the next couple days, but I do have a couple style / documentation asks.

PR #614, opened 2015-08-17
Node init performance: Use shares UIColor/NSAttributedString instances

eanagel:
Re-use global instances of UIColor and NSAttributedString (since they are immutable) for a couple key init methods.


appleguy:
Yes, please!  Thank you.  Any change of this nature is warmly welcomed.

PR #618, opened 2015-08-20
Static layout doesn't properly set a height for collection view cells

garrettmoon:
because their size.width / size.height isn't nan, but FLT_MAX.


garrettmoon:
I'm not sure if this is the correct solution, open to other suggestions.

nguyenhuy:
Looks great to me. Thanks for actually using this spec :)

Issue #617, opened 2015-08-19
New constrained size changes to ASCollectionView providing wrong cell measurements


wpK:
With the new changes from master I noticed one of my ASCollectionViews was not displaying cells. After some debugging, I noticed that the changes made [here](https://github.com/facebook/AsyncDisplayKit/commit/7249661990085606a11ccdfb30b1b1abdfe2a38e) cause the cell node to get passed a height of 0 to measure. I'm using a flowlayout with the direction set to horizontal.

My ASCollectionView is initialized with a CGRectZero frame. It appears when the ASCollectionView layoutSubviews() is called, it detects the change in frame (from CGRectZero to the new size) and calls relayoutAllRows on the data controller. When relayoutAllRows is called, completedNodes is still empty causing none of the nodes to get the updated frame. The end result is on a horizontal flowlayout, the cell node measure call receives width FLT_MAX and height 0.

_layoutNodes is called in the data controller with the nodes, before layoutSubviews but on another thread. When _insertNodes is called from _layoutNodes, it will update the _completedNodes variable on the main thread, but this is executed after layoutSubviews.

nguyenhuy:
@wpK I've been able to pinpoint the problem thanks to your detailed report. I'm trying to fix it and will make a PR once I'm confident with the diff. Thanks again, you're awesome :+1:

wpK:
@nguyenhuy glad you were able to pinpoint it so quickly! I'm not comfortable enough with the internals to try to fix it myself yet.

On a side note, all the contributions you've made are really awesome.

nguyenhuy:
@wpK Thanks for the kind words! I created a new PR that hopefully can fix this issue. Please check it out when you have time :)

wpK:
#620 fixed the issue. Great work!

nguyenhuy:
:dancer: :dancers:

PR #620, opened 2015-08-20
Fix new constrained size is not applied to loading nodes

nguyenhuy:
Reported in #617.

The old version of relayoutAllRows triggers re-measurement on already loaded nodes, but doesn't consider the ones that were measured using the old constrained size and still waiting to be added to _completedNodes (on main thread).

This PR fixes the problem by adding a relayout block to the end of the editing queue. The block will then be executed on main thread, thus making sure all loaded nodes are there to be handled. A new test case is also included.

@wpK: Please help me to test this PR. Thanks!


wpK:
This solved the issue I was having, great job!

nguyenhuy:
Awesome, thanks :)

PR #621, opened 2015-08-21
New min constrained width for cell nodes in table view

nguyenhuy:
Cell nodes in table view should almost always fill its width. To enforce this behaviour, this PR updates the default min constrained width. This new behaviour requires each cell node to have at least 1 flexible object that can fill available space (for example, a flexible shrink and grow text node, or a center layout spec). Users can always op-out of this behaviour by overriding data source's -tableView:constrainedSizeForNodeAtIndexPath:.

Here are screenshots of Kittens sample before and after this diff.

ios simulator screen shot 21 aug 2015 5 41 35 pm

ios simulator screen shot 21 aug 2015 5 42 21 pm

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -791,7 +791,8 @@ void ASPerformBlockWithoutAnimation(BOOL withoutAnimation, void (^block)()) {
   }
   
   // Default size range
-  return ASSizeRangeMake(CGSizeZero, CGSizeMake(_maxWidthForNodesConstrainedSize, FLT_MAX));
+  return ASSizeRangeMake(CGSizeMake(_maxWidthForNodesConstrainedSize, 0),
+                         CGSizeMake(_maxWidthForNodesConstrainedSize, FLT_MAX));
 }
 
 - (void)dataControllerLockDataSource
--- a/examples/Kittens/Sample/KittenNode.mm
+++ b/examples/Kittens/Sample/KittenNode.mm
@@ -136,6 +136,7 @@ static const CGFloat kInnerPadding = 10.0f;
 {
   _imageNode.preferredFrameSize = _isImageEnlarged ? CGSizeMake(2.0 * kImageSize, 2.0 * kImageSize) : CGSizeMake(kImageSize, kImageSize);
   _textNode.flexShrink = YES;
+  _textNode.flexGrow = YES;
   
   return
   [ASInsetLayoutSpec


appleguy:
@nguyenhuy does a typical developer need to know about the flexGrow and flexShrink properties for this to work?  What if there is no stack layout?  If those flex properties are not set, does an assertion or exception get triggered?

Sorry, don't completely understand the implications of this change.  It does seem reasonable to set the default minimum size to be the table view width, though.

nguyenhuy:
Well, I probably used a wrong word. A cell node doesn't need to be "flexible" in the sense of stack layout's flexible. Rather the node should be "responsive". Let me elaborate.

Firstly, for an unresponsive node, it probably has a fixed size and thus can't adapt to different minimum constrained size. Its layout spec would be like this:
```objective-c
{
  _imageNode.preferredFrameSize = CGSizeMake(kImageSize, kImageSize);
  return [ASInsetLayoutSpec newWithInsets:UIEdgeInsetsMake(kOuterPadding, kOuterPadding, kOuterPadding, kOuterPadding)
                                    child:_imageNode];
}
```

The image will be stretched (again, to fill table view width) and look like this:
ios simulator screen shot 22 aug 2015 9 33 56 pm

This is the worst case. We can do better though. For example, we can have a square image:
```objective-c
{
  return [ASInsetLayoutSpec
          newWithInsets:UIEdgeInsetsMake(kOuterPadding, kOuterPadding, kOuterPadding, kOuterPadding)
          child:[ASRatioLayoutSpec newWithRatio:1 child:_imageNode]];
}
```

ios simulator screen shot 22 aug 2015 10 01 54 pm

Or a square image that fills 50% of the table view width:
```objective-c
{
  // 50% of table view width, including insets
  // Height is unconstrained to reserve the ratio
  ASRelativeSize imageMinSize = ASRelativeSizeMake(ASRelativeDimensionMakeWithPercent(0.5), ASRelativeDimensionMakeWithPercent(0));
  ASRelativeSize imageMaxSize = ASRelativeSizeMake(ASRelativeDimensionMakeWithPercent(0.5), ASRelativeDimensionMakeWithPercent(1));
  ASRelativeSizeRange imageSizeRange = ASRelativeSizeRangeMake(imageMinSize, imageMaxSize);
  
  ASStaticLayoutSpecChild *imagePlaceholder =
  [ASStaticLayoutSpecChild
   newWithPosition:CGPointZero
   layoutableObject:
   [ASInsetLayoutSpec
    newWithInsets:UIEdgeInsetsMake(kOuterPadding, kOuterPadding, kOuterPadding, kOuterPadding)
    child:[ASRatioLayoutSpec newWithRatio:1 child:_imageNode]]
   size:imageSizeRange];
  
  return [ASStaticLayoutSpec newWithChildren:@[imagePlaceholder]];
}
```

ios simulator screen shot 22 aug 2015 10 15 10 pm

Or prevent the image from being stretched by using horizontal alignment:
```objective-c
{
  _imageNode.preferredFrameSize = CGSizeMake(kImageSize, kImageSize);
  return [ASCenterLayoutSpec
          newWithCenteringOptions:ASCenterLayoutSpecCenteringX
          sizingOptions:ASCenterLayoutSpecSizingOptionMinimumY
          child:
          [ASInsetLayoutSpec
           newWithInsets:UIEdgeInsetsMake(kOuterPadding, kOuterPadding, kOuterPadding, kOuterPadding)
           child:_imageNode]];
}
```

ios simulator screen shot 22 aug 2015 9 36 52 pm

That I don't use stack layout and still have flexible/responsive cells. And regardless of the cell's actual layout, it should fill table view width :)

PR #622, opened 2015-08-21
Move ASTextNode to use ASLayout

rcancro:
* Renamed `calculateSizeThatFits:` to `renderSizeThatFits:`
* Implemented `calculateLayoutThatFits:`
* Created `ASTextNode+Subclasses.h` that exposes `renderSizeThatFits:`

### DIFF
--- a/AsyncDisplayKit/ASTextNode+Subclasses.h
+++ b/AsyncDisplayKit/ASTextNode+Subclasses.h
@@ -1,0 +1,27 @@
+/* Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <AsyncDisplayKit/ASTextNode.h>
+
+/**
+ * The subclass header _ASTextNode+Subclasses_ defines methods to be
+ * overridden by custom nodes that subclass ASTextNode.
+ *
+ * These methods should never be called directly by other classes.
+ */
+
+@interface ASTextNode (Subclasses)
+
+/**
+ @abstract Calculates a size that will fit all of the text in this text node.
+ @param constrainedSize The maximum available space to render into
+ @discussion The size returned from this method is used to create an ASLayout of the appropriate size.
+ */
+- (CGSize)renderSizeThatFits:(CGSize)constrainedSize;
+
+@end
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -217,6 +217,8 @@
 		509E68651B3AEDC5009B9150 /* CGRect+ASConvenience.h in Headers */ = {isa = PBXBuildFile; fileRef = 205F0E1F1B376416007741D0 /* CGRect+ASConvenience.h */; };
 		509E68661B3AEDD7009B9150 /* CGRect+ASConvenience.m in Sources */ = {isa = PBXBuildFile; fileRef = 205F0E201B376416007741D0 /* CGRect+ASConvenience.m */; };
 		6BDC61F61979037800E50D21 /* AsyncDisplayKit.h in Headers */ = {isa = PBXBuildFile; fileRef = 6BDC61F51978FEA400E50D21 /* AsyncDisplayKit.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		9C204A701B8789CB00313849 /* ASTextNode+Subclasses.h in Headers */ = {isa = PBXBuildFile; fileRef = 9C204A6E1B8789CB00313849 /* ASTextNode+Subclasses.h */; };
+		9C204A711B8789CB00313849 /* ASTextNode+Subclasses.h in Headers */ = {isa = PBXBuildFile; fileRef = 9C204A6E1B8789CB00313849 /* ASTextNode+Subclasses.h */; };
 		9F06E5CD1B4CAF4200F015D8 /* ASCollectionViewTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 9F06E5CC1B4CAF4200F015D8 /* ASCollectionViewTests.m */; };
 		AC21EC101B3D0BF600C8B19A /* ASStackLayoutDefines.h in Headers */ = {isa = PBXBuildFile; fileRef = AC21EC0F1B3D0BF600C8B19A /* ASStackLayoutDefines.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		AC3C4A511A1139C100143C57 /* ASCollectionView.h in Headers */ = {isa = PBXBuildFile; fileRef = AC3C4A4F1A1139C100143C57 /* ASCollectionView.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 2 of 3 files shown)


rcancro:
@nguyenhuy Is this what you had in mind?

nguyenhuy:
Exactly. Thanks for doing this. Does this help your use case?

rcancro:
I'm looking at that now. What I wanted to be able to do in my text node subclass is something like:

```
{
    return [ASInsetLayoutSpec newWithInsets:[self boundingBoxSlop] child:self];
}
```

But that gives me the `'Subclass PITextNode must override at most one of the three layout methods: calculateLayoutThatFits, layoutSpecThatFits or calculateSizeThatFits'` assert.

Do I need to override `- (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize` in my subclass instead?

appleguy:
@rcancro yes, I think you would have to do that.  Let me start by saying that this feels like a confusion that clients are going to experience, so we should think about simplifying the API if at all possible.  I think we may deprecate -calculateSizeThatFits: entirely, although I would like to leave support for it in the framework for quite a while because it used to be the only way (and there is a lot of production code at some pretty big companies that depends on it).

So looking at just -calculateLayout... and -layoutSpec..., the -calculate version basically bypasses ASLayoutSpec entirely.  The ASLayoutSpec version just provides a middleman to handle the calculation step later.

For your use case, although I'm not totally sure if I understand:
- The text node wouldn't have any subnodes, meaning that ASLayoutSpec as an abstraction is presumably useless; the only thing being calculated is our own size, with no children.  The power / utility of ASLayoutSpec is in its ability to calculate not only the self size, but both size and position of children.
- If your subclass used ASLayoutSpec, then the framework has two incompatible things to call calculate on: the superclasses' direct implementation, as well as the ASLayoutSpec itself.  This is probably not what you wanted as the ASLayoutSpec you provide probably can't even accurately calculate the text's size.
- If your subclass used the -calculate method directly, it would be able to call the super implementation very naturally and then modify its return value.  This would remove the ambiguity at the framework level as well.

AsyncDisplayKit/ASTextNode+Subclasses.h
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <AsyncDisplayKit/ASTextNode.h>
+
+/**
+ * The subclass header _ASTextNode+Subclasses_ defines methods to be
```
appleguy: I appreciate the forethought in creating this file, but I actually find +Subclasses files hard to discover / work with unless it is a very complex class that is commonly subclassed (UIGestureRecognizer is well known for this, and also well documented, unlike ASTextNode subclassing).

Since we only have one method, let's put it in the regular ASTextNode interface towards the bottom, and declare prominently in the comments that it is provided as a subclass override hook.  If you can give an example of a use case (e.g. implementing a 3D projection effect or something, to provide additional backing store space), that would be illustrative to clients.

AsyncDisplayKit/ASTextNode+Subclasses.h
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+@interface ASTextNode (Subclasses)
+
+/**
+ @abstract Calculates a size that will fit all of the text in this text node.
+ @param constrainedSize The maximum available space to render into
+ @discussion The size returned from this method is used to create an ASLayout of the appropriate size.
+ */
+- (CGSize)renderSizeThatFits:(CGSize)constrainedSize;
```
appleguy: Thinking about this, I'm wondering if there should be a standard name for it.  Obviously we already have calculateSizeThatFits:, which in a way, feels like a better name.  Do you really need this override point right now, or should we just put the logic into the -calculateLayout method for now?

What's tripping me up is the use of the word render.  It feels like it almost has more significance / meaning than it should.  This is almost more like layoutSizeThatFits:, or sizeThatFits:.

If we want this to be ASTextNode-specific, and it does need to be added now, how about -textBoundingSizeThatFits: ?   Or even just -textSizeThatFits:.

rcancro: I think you are right. I'm not sure I'm really gaining anything by having ASTextNode implement `calculateLayoutThatFits:` instead of `calculateSizeThatFits:` if I've overriding `calculateLayoutThatFits:` anyway. Thinking about it now, what I'm trying to do with these subclasses is probably something that does not need to live in ASDK, but in subclasses in my own project. Let me play around with that and see if I can get it to work, and perhaps we can abandon this PR.

rcancro:
@appleguy That description makes a lot of sense -- thank you! I think a lot of confusion would be solved if `layoutSpecThatFits:` had a name more like `layoutSpecThatFitsSubnodes:` or `layoutSpecForSubnodes:` or just `layoutSubnodes:`. Just as I wouldn't try to do anything in UIView's `layoutSubviews` if I had no subviews, I'd know that I don't have any business making a layout spec if I didn't have any subnodes.

We could also steal the autolayout terminology and instead of `calculateLayoutThatFits:` name it something like `calculateIntrinsicContentSize:` or `calculateLayoutForIntrinsicContentSize:`. That would alert me to the fact that this is for a node with no subnodes and this is how it gets its size. But I'm also used to working in autolayout, so that may be why this name makes sense to me.

nguyenhuy:
Thanks for your explanation, @appleguy. It describes what I had in mind (and many more), with way better language. Definitely something I can learn from.

@rcancro: yes, -calculateLayoutThatFits is what you need to override. But thinking again, if your custom node has already overrides -calculateSizeThatFits, you don't gain any advantage here. However, I hope you don't abandon this PR, because it is exactly what we have to do to drop support for -calculateSizeThatFits. If we do this now, there will be 1 less breaking change for ASTextNode subclasses in the future.

I've been playing around with a custom text node that has built-in insets. My current implementation looks like this:
```objective-c
- (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
{
  UIEdgeInsets insets = UIEdgeInsetsMake(50, 50, 50, 50);
  //TODO: gratefully handle the case in which there is not enough space for insets (one of the two dimensions of renderConstrainedSize is 0).
  CGSize renderConstrainedSize = CGSizeMake(MAX(0, constrainedSize.max.width - insets.left - insets.right),
                                            MAX(0, constrainedSize.max.height - insets.top - insets.bottom));
  CGSize renderSize = [self renderSizeThatFits:renderConstrainedSize];
  CGSize finalSize = CGSizeMake(renderSize.width + insets.left + insets.right,
                                renderSize.height + insets.top + insets.bottom);
  // Do this to be extra sure that finalSize respects constrainedSize, especially min size.
  finalSize = ASSizeRangeClamp(constrainedSize, finalSize);
  return [ASLayout newWithLayoutableObject:self size:finalSize];
}
```

My text node has correct size, but the text is rendered incorrectly because I haven't figured out the way to command the text renderer to draw within the rendering bounds.

And I agree that we need better names. I'm pretty bad at naming things anyway, so it's you call :)

rcancro:
@appleguy and I spoke at this once, and we ultimately decided that a view probably shouldn't be adding its own padding. Even classes like `UIButton` that have `contentEdgeInsets` property are (I assume) adding this padding to a subview.

Ultimately, when I add a text node to a spec I can just wrap it in an `ASInsetLayoutSpec` to get the behavior I want. I'm just looking for a way for this to happen automatically as I know that people won't remember to do it. 

I've thought about adding an optional method to `ASLayoutable` like `willBeAddedToLayoutSpec` that returns an `id<ASLayoutable>` that would allow a child to wrap itself. 

I've also thought about adding a new protocol with this method that a child can conform to. When adding the child, a spec can call this method if the child implements the protocol.

This is all separate from this PR which is about moving ASTextNode to `calculateLayoutThatFits`. So if we can agree on names then we can still merge this. :)

nguyenhuy:
Sounds great! I can't agree more :+1:

appleguy:
@rcancro the idea to provide an auto-wrapping layout spec is pretty interesting.  The fact that ASDisplayNode conforms to ASLayoutable actually breaks this a bit, though.  Imagine for a moment that ASDisplayNode itself could not be added as a child of an ASLayoutSpec, and you /had/ to access a spec.  Then we'd have standard -[ASDisplayNode layoutSpec] method, or something similar, to allow adding individual nodes to the layout tree.  This would potentially serve as an override point to introduce automatic insets and such.  One riff on this would be an -[ASDisplayNode layoutable] method, but if this just returned self, clients could still add display nodes directly and then defeat the override point.

This would be a fairly big change.  It elegantly handles the case we desire, but does add boilerplate and one additional level of understanding required to use layout specs at all (since you wouldn't be able to directly add nodes to specs).  It also makes it a little harder to understand layout specs in general, because in addition to handling layout-manipulating tasks, they would now also have to handle representing an individual node (with no other purpose than standing in for a single node).

Very much in the spirit of ASDK, there might be a fancy way to make this work seamlessly.  What if when you added an ASDisplayNode directly, we had a mechanism built into ASLayoutSpecs to check for a method that provides an optional wrapping ASLayoutSpec?  This seems tricky to get right, as it would be critical that the layout spec returned does contain the node in question, or the user behavior observed would be wildly confusing.

Lastly: I agree we can still land this.  -calculateSizeThatFits: has to move towards deprecation or our API surface area is just too large in this space.  Although ASLayoutSpecs should probably only be used for nodes with children, that is NOT true for ASLayout, as returned by -calculateLayout...  ASLayout specifies self's size, and optionally, any children's position /and/ size.  Every node should have a mechanism of creating an ASLayout, and there are exactly two places they can come from: -calculateLayout called on an ASLayoutSpec, or -calculateLayout called directly on the node.

PR #631, opened 2015-08-27
When relayout a node, preserve its frame origin

nguyenhuy:
Context of the current implementation can be found [here](https://github.com/facebook/AsyncDisplayKit/pull/594#discussion_r36050170).

Since the origin is computed from the node's bounds size, anchorPoint and position properties (see [frame setter](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Private/ASDisplayNode%2BUIViewBridge.mm#L131)), if the position is not adjusted, the node will have a wrong origin within its superview:
ios simulator screen shot 27 aug 2015 9 06 50 pm

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -600,9 +600,27 @@ void ASDisplayNodeRespectThreadAffinityOfNode(ASDisplayNode *node, void (^block)
   } else {
     // This is the root node. Trigger a full measurement pass on *current* thread. Old constrained size is re-used.
     [self measureWithSizeRange:oldConstrainedSize];
-    CGRect bounds = self.bounds;
-    bounds.size = CGSizeMake(_layout.size.width, _layout.size.height);
-    self.bounds = bounds;
+
+    CGSize oldSize = self.bounds.size;
+    CGSize newSize = _layout.size;
+    
+    if (! CGSizeEqualToSize(oldSize, newSize)) {
+      CGRect bounds = self.bounds;
+      bounds.size = newSize;
+      self.bounds = bounds;
+      
+      // Frame's origin must be preserved. Since it is computed from bounds size, anchorPoint
+      // and position (see frame setter in ASDisplayNode+UIViewBridge), position needs to be adjusted.
+      BOOL useLayer = (_layer && ASDisplayNodeThreadIsMain());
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -600,9 +600,27 @@ - (void)__setNeedsLayout
...
+      // and position (see frame setter in ASDisplayNode+UIViewBridge), position needs to be adjusted.
+      BOOL useLayer = (_layer && ASDisplayNodeThreadIsMain());
+      CGPoint anchorPoint = (useLayer ? _layer.anchorPoint : self.anchorPoint);
+      CGPoint oldPosition = (useLayer ? _layer.position : self.position);
+
+      CGFloat xDelta = (newSize.width - oldSize.width) * anchorPoint.x;
+      CGFloat yDelta = (newSize.height - oldSize.height) * anchorPoint.y;
+      CGPoint newPosition = CGPointMake(oldPosition.x + xDelta, oldPosition.y + yDelta);
+
+      useLayer ? _layer.position = newPosition : self.position = newPosition;
```
garrettmoon: What is the reason to check on layer and thread affinity? Don't the properties do that? I.e. self.anchorPoint will check thread affinity and return the layer's anchorPoint if the layer has been loaded and otherwise the stored state in the pendingViewState?

nguyenhuy: @garrettmoon Thanks for pointing this out. I actually was not confident on my understanding of the logic behind loaded layer and pendingViewState. So I ended up do the same layer and thread affinity checks as [-setFrame](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Private/ASDisplayNode%2BUIViewBridge.mm#L136), which are originated from this [commit](https://github.com/facebook/AsyncDisplayKit/commit/b4aaf509537f9c70a7d4c5c1fc303289898873b5) and the huge ["Initial commit"](https://github.com/facebook/AsyncDisplayKit/commit/15565873c93af939ad2abb5fee8ce0285385047e#diff-4c96d795186d237f899d33db7bbf4a1aR123).

After taking a closer look, I think you are right that we can use the properties' getter and setter directly. I will make a new PR to fix this thing (and probably -setFrame as well).

garrettmoon: Great! Thanks!

PR #636, opened 2015-09-02
Use getters and setters while adjusting position of a node

nguyenhuy:
[Here](https://github.com/facebook/AsyncDisplayKit/pull/631/files#r38378312) is the motivation for this PR.

@garrettmoon Please also give feedback on this diff. Thanks!

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -601,25 +601,20 @@ void ASDisplayNodeRespectThreadAffinityOfNode(ASDisplayNode *node, void (^block)
     // This is the root node. Trigger a full measurement pass on *current* thread. Old constrained size is re-used.
     [self measureWithSizeRange:oldConstrainedSize];
 
-    CGSize oldSize = self.bounds.size;
+    CGRect oldBounds = self.bounds;
+    CGSize oldSize = oldBounds.size;
     CGSize newSize = _layout.size;
     
     if (! CGSizeEqualToSize(oldSize, newSize)) {
-      CGRect bounds = self.bounds;
-      bounds.size = newSize;
-      self.bounds = bounds;
+      self.bounds = (CGRect){ oldBounds.origin, newSize };
       
       // Frame's origin must be preserved. Since it is computed from bounds size, anchorPoint
       // and position (see frame setter in ASDisplayNode+UIViewBridge), position needs to be adjusted.
-      BOOL useLayer = (_layer && ASDisplayNodeThreadIsMain());
-      CGPoint anchorPoint = (useLayer ? _layer.anchorPoint : self.anchorPoint);
-      CGPoint oldPosition = (useLayer ? _layer.position : self.position);
-
+      CGPoint anchorPoint = self.anchorPoint;
+      CGPoint oldPosition = self.position;
       CGFloat xDelta = (newSize.width - oldSize.width) * anchorPoint.x;
       CGFloat yDelta = (newSize.height - oldSize.height) * anchorPoint.y;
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -601,25 +601,20 @@ - (void)__setNeedsLayout
     // This is the root node. Trigger a full measurement pass on *current* thread. Old constrained size is re-used.
     [self measureWithSizeRange:oldConstrainedSize];
 
-    CGSize oldSize = self.bounds.size;
+    CGRect bounds = self.bounds;
```
garrettmoon: can this be named oldBounds?

nguyenhuy: oldBounds it is :)

appleguy:
@nguyenhuy I'm happy to merge this, once it builds; do you know what the issue is?

[Joining you from long-range WiFi via 14dB panel antenna @ Burning Man!]

nguyenhuy:
Looks like Travis failed to update homebrew. Not my fault I swear :)

nguyenhuy:
Travis is happy now :)

[Have fun at Burning Man! :dancer: :dancers: :sunglasses: ]

appleguy:
Taking a closer look here, I'm actually not completely sure this is safe...I remember getting pissed at this code and writing those ternary operators.  I am not sure offhand why it was important, possibly only for performance of going to the layer because the node accessors do have overhead (we found and needed to optimize a lot of this in Paper).

Thanks for posting this, as the code is definitely cleaner afterwards.  I'll probably just accept it later, but will think about how we could test what the original intent of the code was.

nguyenhuy:
Sure thing. Please keep me updated :)

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -143,22 +143,12 @@ - (void)setFrame:(CGRect)rect
   ASDisplayNodeAssert(CATransform3DIsIdentity(self.transform), @"Must be an identity transform");
 #endif
 
-  BOOL useLayer = (_layer && ASDisplayNodeThreadIsMain());
-  
-  CGPoint origin      = (useLayer ? _layer.bounds.origin : self.bounds.origin);
-  CGPoint anchorPoint = (useLayer ? _layer.anchorPoint   : self.anchorPoint);
```
appleguy: I've taken a closer look and remember that I made this change after detailed profiling of the Paper app.  setFrame: is a very hot method and additional method calls actually do matter on the main thread, certainly for the iPhone 4S.  I'd like to revisit this nice clean-up once more of Pinterest is running with ASDK and I have a Paper-like target to profile these differences.

Can you revert the changes in this file, and we will land the other file?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -601,25 +601,20 @@ - (void)__setNeedsLayout
     // This is the root node. Trigger a full measurement pass on *current* thread. Old constrained size is re-used.
     [self measureWithSizeRange:oldConstrainedSize];
 
-    CGSize oldSize = self.bounds.size;
+    CGRect oldBounds = self.bounds;
+    CGSize oldSize = oldBounds.size;
     CGSize newSize = _layout.size;
     
     if (! CGSizeEqualToSize(oldSize, newSize)) {
```
appleguy: I forgot about the logic duplication in this file.  Because this code path will be called dramatically less frequently than setFrame:, I agree the cleanup is a good idea.

Just revert the other file.

appleguy:
Sorry for the delay, looking forward to getting this PR integrated & closed out :).  Thanks @nguyenhuy!

nguyenhuy:
Reverted :)

PR #637, opened 2015-09-03
This fixed the cutoff on the 'g' or 'y' character in ASTextNode

lorixx:
For some reason, `NSLayoutManager` did not layout the text correctly for our ASTextNode.
After some experiment, the origin is always off by 1 pixel value.

Hacking that with 1 pixel value seems to fix this annoying bug.

# Before:
ios simulator screen shot sep 2 2015 6 32 50 pm

# See the cutoff from the 'g' character:
screen shot 2015-09-02 at 6 33 17 pm

# After:
ios simulator screen shot sep 2 2015 6 33 05 pm

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -287,11 +287,11 @@ ASDISPLAYNODE_INLINE CGFloat ceilPixelValue(CGFloat f)
   if (_renderer == nil) {
     CGSize constrainedSize = _constrainedSize.width != -INFINITY ? _constrainedSize : self.bounds.size;
     _renderer = [[ASTextNodeRenderer alloc] initWithAttributedString:_attributedString
-                                                        truncationString:_composedTruncationString
-                                                          truncationMode:_truncationMode
-                                                        maximumLineCount:_maximumLineCount
-                                                          exclusionPaths:_exclusionPaths
-                                                         constrainedSize:constrainedSize];
+                                                    truncationString:_composedTruncationString
+                                                      truncationMode:_truncationMode
+                                                    maximumLineCount:_maximumLineCount
+                                                      exclusionPaths:_exclusionPaths
+                                                     constrainedSize:constrainedSize];
   }
   return _renderer;
 }
@@ -421,9 +421,9 @@ ASDISPLAYNODE_INLINE CGFloat ceilPixelValue(CGFloat f)
   UIEdgeInsets shadowPadding = [self shadowPadding];
   CGPoint textOrigin = CGPointMake(self.bounds.origin.x - shadowPadding.left, self.bounds.origin.y - shadowPadding.top);
   return [[ASTextNodeDrawParameters alloc] initWithRenderer:[self _renderer]
-                                                       shadower:[self _shadower]
-                                                     textOrigin:textOrigin
-                                                backgroundColor:self.backgroundColor.CGColor];
+                                                   shadower:[self _shadower]
+                                                 textOrigin:textOrigin
+                                            backgroundColor:self.backgroundColor.CGColor];
 }
 
 #pragma mark - Attributes
--- a/AsyncDisplayKit/Details/ASTextNodeRenderer.mm
+++ b/AsyncDisplayKit/Details/ASTextNodeRenderer.mm
@@ -14,6 +14,7 @@
 #import "ASTextNodeTextKitHelpers.h"
 #import "ASTextNodeWordKerner.h"
 #import "ASThread.h"
+#import "ASInternalHelpers.h"
 
 static const CGFloat ASTextNodeRendererGlyphTouchHitSlop = 5.0;
 static const CGFloat ASTextNodeRendererTextCapHeightPadding = 1.3;
@@ -607,6 +608,10 @@ static const CGFloat ASTextNodeRendererTextCapHeightPadding = 1.3;
 
 - (void)drawInRect:(CGRect)bounds inContext:(CGContextRef)context
 {
+  // NSLayoutManager often missed one pixel at the last line of the text.
+  // We need to offset this extra pixel to make 'g' or 'y' rendered perfectly at the last line within ASTextNode.
+  bounds.origin.y -= 1.0/ASScreenScale();
+
   ASDisplayNodeAssert(context, @"This is no good without a context.");
   UIGraphicsPushContext(context);
   CGContextSaveGState(context);


ocrickard:
Can you please provide the code you used to generate these snapshots?  ASTextNode (if memory serves) returns the non-rounded exact text sizing so we would not accumulate float math error in Paper's complex layouts resulting in phantom mis-alignment creeping in.  As a result it's actually fairly easy to cause this behavior where you accidentally clip the bottom pixel due to how CoreAnimation renders.

lorixx:
```objectiveC
diff --git examples/Kittens/Sample/KittenNode.mm examples/Kittens/Sample/KittenNode.mm
index 0afd654..a050f9c 100644
--- examples/Kittens/Sample/KittenNode.mm
+++ examples/Kittens/Sample/KittenNode.mm
@@ -93,6 +93,15 @@ static const CGFloat kInnerPadding = 10.0f;
   _textNode = [[ASTextNode alloc] init];
   _textNode.attributedString = [[NSAttributedString alloc] initWithString:[self kittyIpsum]
                                                                attributes:[self textStyle]];
+
+    UILabel *label = [[UILabel alloc] initWithFrame:CGRectZero];
+    label.attributedText = _textNode.attributedString;
+    CGSize size = [label sizeThatFits:CGSizeMake(FLT_MAX, FLT_MAX)];
+    [self.view addSubview:label];
+    label.frame = CGRectMake(200, 16, size.width, size.height);
+
+  });
   [self addSubnode:_textNode];
 
   // hairline cell separator
@@ -116,12 +125,12 @@ static const CGFloat kInnerPadding = 10.0f;
     [string appendString:placeholders[i]];
   }
 
-  return string;
+  return @"gyfPji";
 }
 
 - (NSDictionary *)textStyle
 {
-  UIFont *font = [UIFont fontWithName:@"HelveticaNeue" size:12.0f];
+  UIFont *font = [UIFont fontWithName:@"HelveticaNeue" size:30.0f];
 
   NSMutableParagraphStyle *style = [[NSParagraphStyle defaultParagraphStyle] mutableCopy];
   style.paragraphSpacing = 0.5 * font.lineHeight;

ocrickard:
Hmm, you're still not showing how you're setting the frame on the text node.

lorixx:
It's inside the `examples/Kittens` Kitten project that comes with ASDisplayKit. That should be easy to find. 
```objectiveC
- (void)layout
{
  [super layout];
  
  // Manually layout the divider.
  CGFloat pixelHeight = 1.0f / [[UIScreen mainScreen] scale];
  _divider.frame = CGRectMake(0.0f, 0.0f, self.calculatedSize.width, pixelHeight);
}

{
  CGSize imageSize = CGSizeMake(kImageSize, kImageSize);
  CGSize textSize = [_textNode measure:CGSizeMake(constrainedSize.width - kImageSize - 2 * kOuterPadding - kInnerPadding,
                                                  constrainedSize.height)];

  // ensure there's room for the text
  CGFloat requiredHeight = MAX(textSize.height, imageSize.height);
  return CGSizeMake(constrainedSize.width, requiredHeight + 2 * kOuterPadding);
}

- (void)layout
{
  CGFloat pixelHeight = 1.0f / [[UIScreen mainScreen] scale];
  _divider.frame = CGRectMake(0.0f, 0.0f, self.calculatedSize.width, pixelHeight);

  _imageNode.frame = CGRectMake(kOuterPadding, kOuterPadding, kImageSize, kImageSize);

  CGSize textSize = _textNode.calculatedSize;
  _textNode.frame = CGRectMake(kOuterPadding + kImageSize + kInnerPadding, kOuterPadding, textSize.width, textSize.height);
}

ocrickard:
Got it yes, that should be correct.  I had forgotten we ceil to pixel values in the calculateSizeThatFits: method...

lorixx:
@ocrickard Initially I thought it was the https://github.com/lorixx/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L363-L364

That causing the off-pixel since round could get tricky instead of ceilf, but I realized those did not participate on the layout for the text too. Hence I suspected that the villain might be the way we do layout, which leads to the origin point we pass into the `NSLayoutManager`.

ocrickard:
My guess is that the real issue is somewhere before here:

https://github.com/lorixx/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASTextNodeRenderer.mm#L195

What's the value of these rects' origins?  Are they negative?  'cause we're effectively setting them to zero here:

https://github.com/lorixx/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L422

maybe instead of a "size" param on ASTextNodeRenderer we need a "bounds" rect that can be used for maintaining this state throughout.

ocrickard:
lol, yeah even at https://github.com/lorixx/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASTextNodeRenderer.mm#L195 we're truncating the rect through that intersection call.  #magic.  I bet we didn't know what to do about the negative values here.

lorixx:
>What's the value of these rects' origins? Are they negative? 'cause we're effectively setting them to zero 

The rect's origin is always CGPointZero, including the result came back from the `NSLayoutManager`: https://github.com/lorixx/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASTextNodeRenderer.mm#L189

I think what https://github.com/lorixx/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASTextNodeRenderer.mm#L195 does is basically constraining the bounds so that the drawing happens within the expected width. It did document that NSLayoutManager sometimes return an incorrect width which is bigger than the constraining width, which is `surprise`.

Regarding https://github.com/lorixx/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L422

The origin are still {0, 0} and there was no shadowOffset to affect.

ocrickard:
Interesting, well I still don't think this is the right fix for whatever the problem is, but I'll leave this to you guys to accept or further investigate.

lorixx:
So I dig into some code and wrote some sample code using pure NSLayoutManager, and I found out the exact cutoff on the bottom of the 'g' glyph. I think it's an Apple bug.
ios simulator screen shot sep 2 2015 10 54 21 pm


```objectiveC
- (void)viewDidLoad {
  [super viewDidLoad];
  UIImageView *imageView = [[UIImageView alloc] initWithFrame:CGRectZero];

  NSLayoutManager *layoutManager = [[NSLayoutManager alloc] init];
  layoutManager.usesFontLeading = NO;

  NSAttributedString *string =
  [[NSAttributedString alloc]
   initWithString:@"gggggg"
   attributes:@{ NSFontAttributeName : [UIFont fontWithName:@"HelveticaNeue" size:30.0] }];

  NSTextStorage *textStorage = [[NSTextStorage alloc] initWithAttributedString:string];
  [textStorage addLayoutManager:layoutManager];

  NSTextContainer *textContainer = [[NSTextContainer alloc] initWithSize:CGSizeMake(200, FLT_MAX)];
  textContainer.lineFragmentPadding = 0;
  textContainer.lineBreakMode = NSLineBreakByWordWrapping;

  [layoutManager addTextContainer:textContainer];
  [layoutManager ensureLayoutForTextContainer:textContainer];
  CGRect boundingRect = [layoutManager usedRectForTextContainer:textContainer];

  UIGraphicsBeginImageContext(boundingRect.size);
  NSRange glyphRange = [layoutManager glyphRangeForTextContainer:textContainer];
  [layoutManager drawGlyphsForGlyphRange:glyphRange atPoint:boundingRect.origin];
  UIImage *viewImage = UIGraphicsGetImageFromCurrentImageContext();
  imageView.image = viewImage;
  UIGraphicsEndImageContext();

  [self.view addSubview:imageView];
  imageView.frame = CGRectMake(150, 200, boundingRect.size.width, boundingRect.size.height);
}

appleguy:
@ocrickard - thanks for taking a look here, really appreciate your expertise in this area! cc @rcancro.

I agree this is not the right fix, as clearly this adjustment wouldn't necessarily apply to other languages / scripts.

Does this happen on all recent iOS versions?  Have you tried iOS 9?

lorixx:
@appleguy Yea, the bug also happens in iOS 9 as well. I do think it's also affecting other locales as well since it's purely a layout issue within -[NSLayoutManager drawGlyphsForGlyphRange:atPoint:].

Another temporary solution is let the client side increase 1 pixel in the bounds height returned from `-[ASTextNodeRenderer size]`.

appleguy:
@lorixx — Thanks for your analysis, it's much appreciated!  Wish we could be working together :).

@rcancro should take a look at this as he has been pixel-pushing text layout recently.  I want to make sure he can trivially reproduce the problem, at least as a starter.

@ocrickard do you have reason to believe that if we adopted the newer text layout engine (revised for ComponentKit), that it would fix this issue?  If you could test that for me and confirm, it would be really helpful.

lorixx:
Actually found out that it only happens on certain fonts see following, especially on HelveticaNeue and HelveticaNeue-Light: (left is rendered using NSLayoutManager(a.k.a ASTextNode), right is rendered using UILabel.)

ios simulator screen shot sep 6 2015 12 09 45 am
ios simulator screen shot sep 6 2015 12 09 47 am

lorixx:
Abandoning this change.

PR #643, opened 2015-09-08
Avoid doing relayout during initial configuration

nguyenhuy:
First size change occurs during initial configuration. An expensive relayout pass is unnecessary at that time and should be avoided.

This PR proposes a simple fix for this problem, that is adding a new boolean flag. With the fix in place, my ~~manual~~ tests showed that main thread was not blocked at startup by an expensive relayout call. I also tested the use case in which a rotation occurred after initial configuration. The new size was picked up correctly via a relayout.

Another alternative fix would be to completely ignore size changes to *and* from a zero size. With this solution, we will also gain some optimizations in case the scroll view size is changed from a non-zero size (A) to zero, then back to A or another non-zero size (B), for example as the result of hide and unhide animations. The implementation of that solution is more complex. So it may not be a good idea to implement now, because we are trying to push 2.0 out and there are many bugs need to be handled (I suspect most of them ~~are~~ were introduced by me :(

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -126,6 +126,7 @@ static BOOL _isInterceptedSelector(SEL sel)
   ASBatchContext *_batchContext;
   
   CGSize _maxSizeForNodesConstrainedSize;
+  BOOL _ignoreMaxSizeChange;
 }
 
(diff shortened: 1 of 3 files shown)


appleguy:
Great — and especially thrilled to see the tests!  Thanks @nguyenhuy :)

nguyenhuy:
Found a grammar mistake in the first commit message. But it's too late to fix I guess haha.

PR #644, opened 2015-09-08
Added ASLayoutOptions class to store all options for ASLayoutSpecs

rcancro:
`ASLayoutable` is now based on the protocol `ASLayoutablePrivate`. The private protocol contains both the `finalLayoutable` method as well as a `ASLayoutOptions` class.

The family of layoutable protocols (`ASStackLayoutable`, `ASBaselineLayoutable`, etc) still exist, but instead of being based on `ASLayoutable`, `ASLayoutable` implements all of the other layoutable protocols.

Since both `ASDisplayNode` and `ASLayoutSpec` conform to `ASLayoutable` they both contain an `ASLayoutOptions` member. There is a private file, `ASLayoutOptionsPrivate.h/.m`, that implements the private protocol for these two classes. Now that `ASLayoutable` implements the layoutable protocols, `ASDisplayNode` and `ASLayoutSpec` also implement them via forwards to the internal `ASLayoutOptions`. 

Another change is that `ASLayoutSpec` now holds all children for all subclasses. This allowed for the `finalLayoutable` method to only be called in one place throughout the app.

One question I have is whether we should move the baseline alignment back into ASStackLayoutSpec instead of it being its own layout spec. It seems strange to live off on its own to me.


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -155,7 +151,6 @@ - (void)_initializeInstance
   }
   _methodOverrides = overrides;
 
-  _flexBasis = ASRelativeDimensionUnconstrained;
```
appleguy: Are you confident this isn't needed?  I could be wrong, but I vaguely recall @nguyenhuy adding this after we encountered a certain bug.

appleguy:
Two notes:

- We should carefully consider alternative methods to terminate the recursion (provide a base case for) calling -finalLayoutable.  Ideally we could remove the Parent argument, and ASDisplayNode / ASLayoutSpec base classes would return "self".  This may not be possible, but failing to do so means that the abstraction is in some way impure; the base class implementations would quite literally be false if they return nil.  This may inhibit curious parties from fully understanding the system as they explore the code (it's the most likely method to get looked at, as subclassers override it and want to look at the superclass implementation)
- We should think about alternatives to the NSCopying / merging of ASLayoutOptions.  I am fairly confident that it would be better for the ASLayoutSpec subclass to bear the responsibility of exploring the layout tree to whatever depth it requires to operate properly.  In the case of ASStackLayoutSpec, it should probably traverse to the leaf nodes in its search for baseline-alignment hints, and then we can document this behavior next to the definition of those properties.
- Lastly, some way to bring visibility to the @dynamic properties will likely be important.  Right now they're hidden in a separate file, and may not be observed by users of ASDisplayNode or ASStackLayoutSpec.

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: Sorry but I don't have a very good memory. Looking at the commit history, I think I added this to make sure _flexBasis always has a correct default value. The same thing might be done in the new ASLayoutOptions class [here](https://github.com/facebook/AsyncDisplayKit/pull/644/files#diff-753574fd0cd2c18cdce87fd5393d7734R100).

rcancro: This may have been a mistake on my part. I'll put it back in.

AsyncDisplayKit/Layout/ASBaselineLayoutSpec.h
```diff
@@ -29,7 +29,7 @@ typedef NS_ENUM(NSUInteger, ASBaselineLayoutBaselineAlignment) {
  If the spec is created with a vertical direction, a child's vertical spacing will be measured from its
  baseline instead of from the child's bounding box.
 */
-@interface ASBaselineLayoutSpec : ASLayoutSpec <ASBaselineLayoutable>
+@interface ASBaselineLayoutSpec : ASLayoutSpec
```
appleguy: Can this be a subclass of ASStackLayoutSpec, or can it be eliminated by folding in the extra functionality into ASStackLayoutSpec directly?

AsyncDisplayKit/Layout/ASBaselineLayoutable.h
```diff
@@ -6,9 +6,9 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-#import "ASStackLayoutable.h"
+#import <UIKit/UIKit.h>
```
appleguy: Remove if not required.

AsyncDisplayKit/Layout/ASLayoutOptions.mm
```diff
@@ -0,0 +1,252 @@
...
+  copy.ascender = self.ascender;
+  copy.descender = self.descender;
+  
+  copy.sizeRange = self.sizeRange;
+  copy.layoutPosition = self.layoutPosition;
+}
+
+
+#pragma mark - Defaults
+- (void)setupDefaults
```
appleguy: I'm thinking we should kill this method and have people do their configuration in -init.  A subclass will likely override -init out of habit anyway, and would naturally put initialization of any new data structures there, so I'm not sure there is value in -setupDefaults.

More importantly, let's comment out any of these lines that set already-default values.  Since this object will be created / very / frequently, and currently will be created on the main thread, we need to cut down on method calls; there are at least 8 below that will always match the calloc / zeroed-out state that Objective-C objects are created with.  I'm not sure how the enum values are defined, but assume they are also equal to 0.

So you could add a comment explaining that, and keep the lines (commented out) for explicitness if desired, or just remove them.

AsyncDisplayKit/Layout/ASLayoutOptions.mm
```diff
@@ -0,0 +1,252 @@
...
+  self.alignSelf = ASStackLayoutAlignSelfAuto;
+
+  self.ascender = 0;
+  self.descender = 0;
+  
+  self.sizeRange = ASRelativeSizeRangeMake(ASRelativeSizeMakeWithCGSize(CGSizeZero), ASRelativeSizeMakeWithCGSize(CGSizeZero));
+  self.layoutPosition = CGPointZero;
+}
+
+// Do this here instead of in Node/Spec subclasses so that custom specs can set default values
```
appleguy: Could you refine this comment?  I don't understand it...if I have a custom spec, how would that help me set default values, with this method implemented in the Options (not spec)?

AsyncDisplayKit/Layout/ASLayoutOptions.mm
```diff
@@ -0,0 +1,252 @@
...
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if ([layoutable isKindOfClass:[ASTextNode class]]) {
+    ASTextNode *textNode = (ASTextNode *)layoutable;
+    if (textNode.attributedString.length > 0) {
+      self.ascender = round([[textNode.attributedString attribute:NSFontAttributeName atIndex:0 effectiveRange:NULL] ascender] * ASScreenScale())/ASScreenScale();
+      self.descender = round([[textNode.attributedString attribute:NSFontAttributeName atIndex:textNode.attributedString.length - 1 effectiveRange:NULL] descender] * ASScreenScale())/ASScreenScale();
+    }
+  }
+  if ([layoutable isKindOfClass:[ASDisplayNode class]]) {
```
appleguy: Do a slight refactoring: contain the ASTextNode check inside of the ASDisplayNode check.  Thus you can do [displayNode isKindOfClass:[ASTextNode class]].  This pattern will avoid an isKindOfClass: check for all ASLayoutSpec arguments (worthwhile / desirable as class checks are not the fastest, although arguably not essential)

AsyncDisplayKit/Layout/ASLayoutOptions.mm
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// Do this here instead of in Node/Spec subclasses so that custom specs can set default values
+- (void)setValuesFromLayoutable:(id<ASLayoutable>)layoutable
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if ([layoutable isKindOfClass:[ASTextNode class]]) {
+    ASTextNode *textNode = (ASTextNode *)layoutable;
+    if (textNode.attributedString.length > 0) {
+      self.ascender = round([[textNode.attributedString attribute:NSFontAttributeName atIndex:0 effectiveRange:NULL] ascender] * ASScreenScale())/ASScreenScale();
```
appleguy: For both readability and performance, /definitely/ cache textNode.attributedString in a local variable.  I'd also do it for screenScale, although less important.

AsyncDisplayKit/Layout/ASLayoutOptionsPrivate.h
```diff
@@ -0,0 +1,34 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <AsyncDisplayKit/ASDisplayNode.h>
+#import <AsyncDisplayKit/ASLayoutSpec.h>
+
+
+@interface ASDisplayNode()
```
appleguy: We should consider moving these into the primary header files, although it's fine here for now.  I'm worried the existence of these properties is a bit hidden.  I haven't seen the ASDN header yet, though, which maybe enumerates all the properties in the primary header.

AsyncDisplayKit/Layout/ASLayoutOptionsPrivate.mm
```diff
@@ -0,0 +1,133 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "ASLayoutOptionsPrivate.h"
+#import <AsyncDisplayKit/ASDisplayNodeInternal.h>
+
+
+#define ASLayoutOptionsForwarding \
```
appleguy: Very cute.  As an incremental improvement to readability, put a short comment above this calling attention to the strategy used here.  E.g. we are exposing a passthrough API on both ASDisplayNode and ASLayoutSpec for the convenience of the majority of developers, who will never need to customize ASLayoutOptions (and thus don't need to be aware of the object type's existence).  In order to avoid repetition, #defining the entire implementation allows it to be included in category extensions on both classes very easily.

AsyncDisplayKit/Layout/ASLayoutSpec.h
```diff
@@ -22,4 +22,12 @@
 
 - (instancetype)init;
 
+- (void)setChild:(id<ASLayoutable>)child;
```
appleguy: Perhaps worth some comments on children:
- All valid layout specs should have at least one child (is this true?)
- Base class implements handling of multiple children, and named children, as it is commonly required by subclasses (give a couple examples)
- However, not all layout specs may handle multiple children.  (Examples: inset node has one, background node has exactly two, which are specially identified)

Docs like this may not feel that critical, and it's hard to know the exact impact that they will have...but both for our team and others, I'm pretty sure they will help folks confirm their guesses and build confidence in their understanding as they explore the headers during normal implementation work.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -42,4 +49,60 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
   return [ASLayout layoutWithLayoutableObject:self size:constrainedSize.min];
 }
 
+- (ASLayoutSpec *)finalLayoutableWithParent:(ASLayoutSpec *)parentSpec;
+{
+  return nil;
```
appleguy: Please follow up on this after landing this initial version.  I'd like to aim for a solution that would expect / handle a return value of "self" here.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -42,4 +49,60 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
+  [self setChild:child forIdentifier:kDefaultChildKey];
+}
+
+- (id<ASLayoutable>)layoutableToAddFromLayoutable:(id<ASLayoutable>)child
+{
+  ASLayoutOptions *layoutOptions = [child layoutOptions];
+  
+  id<ASLayoutable> finalLayoutable = [child finalLayoutableWithParent:self];
+  if (finalLayoutable) {
+    [layoutOptions copyIntoOptions:finalLayoutable.layoutOptions];
```
appleguy: Create layoutOptions local variable only inside this conditional, as it appears there is a case where it is not needed (e.g. finalLayoutable is nil).  I don't have an intuitive understanding of this code, though, and think it deserves a comment above or at the start of the method.  In what scenario is it called, and what are its semantics?  Is this the thing that currently handles propagating up properties such as baseline alignment for inset-wrapped stack children?

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -42,4 +49,60 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
+  self.layoutChildren[identifier] = [self layoutableToAddFromLayoutable:child];;
+}
+
+- (void)setChildren:(NSArray *)children
+{
+  ASDisplayNodeAssert(self.isMutable, @"Cannot set properties when layout spec is not mutable");
+  
+  NSMutableArray *finalChildren = [NSMutableArray arrayWithCapacity:children.count];
+  for (id<ASLayoutable> child in children) {
+    [finalChildren addObject:[self layoutableToAddFromLayoutable:child]];
```
appleguy: OK, now I see that is just the primary mechanism to call finalLayoutable.  I think we could get rid of it completely, and just call finalLayoutable in-place, if we did not need to do the options merging.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -42,4 +49,60 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
+    [layoutOptions copyIntoOptions:finalLayoutable.layoutOptions];
+    return finalLayoutable;
+  }
+  return child;
+}
+
+- (void)setChild:(id<ASLayoutable>)child forIdentifier:(NSString *)identifier
+{
+  ASDisplayNodeAssert(self.isMutable, @"Cannot set properties when layout spec is not mutable");
+  self.layoutChildren[identifier] = [self layoutableToAddFromLayoutable:child];;
```
appleguy: I think we need assertions to protect against the case of a user calling setChild: and setChildren: on the same spec.  Right now, it would store both, and be a sort of undefined behavior.  It should probably trigger an assertion or exception, and as fallback runtime behavior (e.g. if debug-only assertion), overwrite the last action ("last one wins").

This is fairly important for robustness as a relatively innocent product development mistake could trigger this condition, e.g. imagine experiment-driven code paths that are hard to test.  However, feel free to notate it as a separate followup for a couple weeks from now.

AsyncDisplayKit/Layout/ASLayoutable.h
```diff
@@ -9,16 +9,23 @@
...
+
+#import <AsyncDisplayKit/ASLayoutablePrivate.h>
 
 @class ASLayout;
+@class ASLayoutSpec;
 
 /** 
  * The ASLayoutable protocol declares a method for measuring the layout of an object. A class must implement the method
  * so that instances of that class can be used to build layout trees. The protocol also provides information 
  * about how an object should be laid out within an ASStackLayoutSpec.
```
appleguy: Update this comment, probably referencing ASLayoutOptions as the container of metadata required for layout, giving Stack as an example but referencing the possibility of custom extensions ("see ASLayoutOptions")

AsyncDisplayKit/Layout/ASLayoutable.h
```diff
@@ -9,16 +9,23 @@
...
  */
-@protocol ASLayoutable <NSObject>
+@protocol ASLayoutable <ASLayoutablePrivate, ASStackLayoutable, ASBaselineLayoutable, ASStaticLayoutable>
```
appleguy: Nit, but put ASLayoutablePrivate at the end of the list.

AsyncDisplayKitTests/ASStackLayoutSpecSnapshotTests.mm
```diff
@@ -41,8 +42,8 @@ - (void)setUp
                         ];
   for (ASStaticSizeDisplayNode *subnode in subnodes) {
     subnode.staticSize = subnodeSize;
-    subnode.flexGrow = flex;
-    subnode.flexShrink = flex;
+    subnode.layoutOptions.flexGrow = flex;
```
appleguy: Can you delete every addition of .layoutOptions in this file?  Aren't these supposed to be forwarded automatically?

AsyncDisplayKit/Private/ASStackUnpositionedLayout.mm
```diff
@@ -14,18 +14,19 @@
...
-static ASLayout *crossChildLayout(const id<ASStackLayoutable> child,
+static ASLayout *crossChildLayout(const id<ASLayoutable> child,
                                   const ASStackLayoutSpecStyle style,
                                   const CGFloat stackMin,
                                   const CGFloat stackMax,
                                   const CGFloat crossMin,
                                   const CGFloat crossMax)
 {
-  const ASStackLayoutAlignItems alignItems = alignment(child.alignSelf, style.alignItems);
+  const ASStackLayoutAlignItems alignItems = alignment(child.layoutOptions.alignSelf, style.alignItems);
```
appleguy: Hopefully the .layoutOptions is not needed in this file?

appleguy:
Thanks for the varied exploration and implementation work, @rcancro!  Let's land this as a step forward, and continue evolving it rapidly.

PR #645, opened 2015-09-09
First thoughts on fixing the finalLayoutable method.

rcancro:
A final layoutable method would look like:

```
- (id<ASLayoutable>)finalLayoutable
{
  ASInsetLayoutSpec *insetSpec = [[ASInsetLayoutSpec alloc] init];
  insetSpec.isFinalLayoutable = YES;
  insetSpec.insets = UIEdgeInsetsMake(4, 4, 4, 4);
  [insetSpec setChild:self];
  return insetSpec;
}
```

While I'm not wild about this for a couple of reasons (1: we require the user to set a flag instead of doing something more seamless; 2: it makes it impossible to use the convenience methods on the layoutSpecs in finalLayoutable), I think it is better than before. We can probably do better though, so let's use this as a starting point to discuss.

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayoutablePrivate.h
+++ b/AsyncDisplayKit/Layout/ASLayoutablePrivate.h
@@ -12,8 +12,29 @@
 
 @class ASLayoutSpec;
 @class ASLayoutOptions;
+@protocol ASLayoutable;
 
 @protocol ASLayoutablePrivate <NSObject>
-- (ASLayoutSpec *)finalLayoutableWithParent:(ASLayoutSpec *)parentSpec;
+
+/**
+ *  @abstract A display node cannot implement both calculateSizeThatFits: and layoutSpecThatFits:. This
+ *  method exists to give the user a chance to wrap an ASLayoutable in an ASLayoutSpec just before it is
+ *  added to a parent ASLayoutSpec. For example, if you wanted an ASTextNode that was always inside of an
+ *  ASInsetLayoutSpec, you could subclass ASTextNode and implement finalLayoutable so that it wraps
+ *  self in an inset spec.
+ *
+ *  Note that any ASLayoutable other than self that is returned MUST set isFinalLayoutable to YES BEFORE
+ *  adding a child.
+ *
+ *  @return The layoutable that will be added to the parent layout spec. Defaults to self.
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Layout/ASLayoutablePrivate.h
```diff
@@ -12,8 +12,29 @@
 
 @class ASLayoutSpec;
 @class ASLayoutOptions;
+@protocol ASLayoutable;
 
 @protocol ASLayoutablePrivate <NSObject>
-- (ASLayoutSpec *)finalLayoutableWithParent:(ASLayoutSpec *)parentSpec;
+
+/**
+ *  @abstract A display node cannot implement both calculateSizeThatFits: and layoutSpecThatFits:. This
```
appleguy: The comments you added are overall AWESOME!!

The first line of this one confuses me, and may be a copy/paste error.  If we intended to mention those methods, please add to the description to make it more clear why they're brought up.

AsyncDisplayKit/Layout/ASLayoutablePrivate.h
```diff
@@ -12,8 +12,29 @@
...
+ *  method exists to give the user a chance to wrap an ASLayoutable in an ASLayoutSpec just before it is
+ *  added to a parent ASLayoutSpec. For example, if you wanted an ASTextNode that was always inside of an
+ *  ASInsetLayoutSpec, you could subclass ASTextNode and implement finalLayoutable so that it wraps
+ *  self in an inset spec.
```
appleguy: nit: Sentence reads more smoothly with "itself" instead of "self"

AsyncDisplayKit/Layout/ASLayoutablePrivate.h
```diff
@@ -12,8 +12,29 @@
...
+ *
+ *  Note that any ASLayoutable other than self that is returned MUST set isFinalLayoutable to YES BEFORE
```
appleguy: Here and in the implementation file, add a comma after YES — it looks a little weird to have them run together :)

appleguy:
OK, this is definitely an improvement.  Thanks for iterating quickly on these important APIs :).

PR #647, opened 2015-09-11
Removed ASBaselineLayoutSpec and made baseline alignment part of ASStackLayoutSpec

rcancro:
ASBaselineLayoutSpec really didn't make much sense on its own. Now that ASLayoutable is implementing all of the layoutable protocols, I moved the ascender/descender properties needed for baseline alignment back into ASStackLayoutable.


appleguy:
Great cleanup!  It's been a long and complex path, and breaking this out was an essential part of our exploration.  Thanks for both creating and destroying it :).

PR #648, opened 2015-09-11
Addressed comments to LayoutOptions PR

rcancro:
Addressed comments from the PR: https://github.com/facebook/AsyncDisplayKit/pull/644

Note: I did not change how the ASLayoutOptions are propagated. I did change the "copyInto" method to `propagateOptionsFromLayoutOptions:` (I even spelled it right on my second try). The reasoning behind this is I worry that traversing children to get a valid value for an option may be error prone. 

For example, imagine a stack of 3 text nodes wrapped in inset specs. You want the first 2 to align on the baseline but not the third. You set the 3rd nodes inset spec to have ascender/descender to 0.

While laying out the textNodes the stack spec starts traversing the 3rd node. It sees an inset spec with an ascender/descender of 0. This is the default value. What does it do? Does it assume that since this is a default value, it should keep traversing until it finds a value for ascender/descender? That would be wrong. Does it take whatever is in the top node? If we aren't propagating options then that method will be wrong for the first two text nodes. It seems to me the best way to keep the options unambiguous is to propagate them up each time. Then any user overrides are explicit.

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayoutOptions.h
+++ b/AsyncDisplayKit/Layout/ASLayoutOptions.h
@@ -23,7 +23,7 @@
 - (void)setValuesFromLayoutable:(id<ASLayoutable>)layoutable;
 
 #pragma mark - Subclasses should implement these!
-- (void)propogateOptionsFromLayoutOptions:(ASLayoutOptions *)layoutOptions;
+- (void)propagateOptionsFromLayoutOptions:(ASLayoutOptions *)layoutOptions;
 
 #pragma mark - ASStackLayoutable
 
--- a/AsyncDisplayKit/Layout/ASLayoutOptions.h
+++ b/AsyncDisplayKit/Layout/ASLayoutOptions.h
@@ -12,8 +12,7 @@
 
 @protocol ASLayoutable;
 
-#import <AsyncDisplayKit/ASStackLayoutable.h>
-#import <AsyncDisplayKit/ASStaticLayoutable.h>
+#import <AsyncDisplayKit/ASLayoutSpec.h>
 
 @interface ASLayoutOptions : NSObject <ASStackLayoutable, ASStaticLayoutable, NSCopying>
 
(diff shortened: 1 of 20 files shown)


AsyncDisplayKit/Layout/ASLayoutOptions.h
```diff
@@ -24,8 +23,7 @@
 - (void)setValuesFromLayoutable:(id<ASLayoutable>)layoutable;
 
 #pragma mark - Subclasses should implement these!
-- (void)setupDefaults;
-- (void)copyIntoOptions:(ASLayoutOptions *)layoutOptions;
+- (void)propagateOptionsFromLayoutOptions:(ASLayoutOptions *)layoutOptions;
```
appleguy: How would you describe this behavior?  Is it more like Merge?  Or even Replace?

AsyncDisplayKit/Layout/ASLayoutOptions.mm
```diff
@@ -73,59 +86,57 @@ - (instancetype)initWithLayoutable:(id<ASLayoutable>)layoutable;
 - (id)copyWithZone:(NSZone *)zone
 {
   ASLayoutOptions *copy = [[[self class] alloc] init];
-  [self copyIntoOptions:copy];
+  [copy propagateOptionsFromLayoutOptions:self];
   return copy;
 }
 
-- (void)copyIntoOptions:(ASLayoutOptions *)copy
+- (void)propagateOptionsFromLayoutOptions:(ASLayoutOptions *)layoutOptions
```
appleguy: Yeah, this looks like adopt/overwrite/something.  In fact the original name but flipped might be OK: copyFromOptions:

AsyncDisplayKit/Layout/ASLayoutSpec.h
```diff
@@ -22,10 +22,37 @@
 
 - (instancetype)init;
 
+/**
+ *  Set child methods
```
appleguy: These (and a few others) are really great comments.  Can you look at the Apple Doc format and make sure these will show up in the documentation when we regenerate it?  @nguyenhuy, could you regenerate the docs so we have an updated set before NSSpain?

appleguy:
Moving forward!

AsyncDisplayKit/Layout/ASLayoutSpec.h
nguyenhuy: Sure thing :)

PR #651, opened 2015-09-15
added a callback for initWithViewBlock/initWithLayerBlock

rcancro:


AsyncDisplayKit/ASCellNode.m
```diff
@@ -31,13 +31,13 @@ - (instancetype)init
   return self;
 }
 
-- (instancetype)initWithLayerBlock:(ASDisplayNodeLayerBlock)viewBlock
+- (instancetype)initWithLayerBlock:(ASDisplayNodeLayerBlock)viewBlock layerDidLoadBlock:(ASDisplayNodeLayerLoadedBlock)layerLoadedBlock
```
appleguy: I'm excited to have this feature!  API simplification request:

- Can we name both of these new arguments didLoadBlock: ?
- Can the block type be the same, ASDisplayNodeLoadedBlock or ASDisplayNodeDidLoadBlock? (latter is probably better..)

The argument to the block should simply be the node itself.  It's trivially easy to access the view and layer on the node, but fewer people know how to access the node from the view or layer, so I think it will provide possible utility in addition to the API simplicity.  I agree with your original design in that it is probably more convenient for the common case, but think it is important to minimize duplication between view & layer-backed APIs (as one of the primary values of layer-backing is its seamless compatibility when enabled on previously view-backed nodes)

appleguy:
Just thought of something else...we should probably provide an initWithDidLoadBlock:, too!

appleguy:
Maybe just initWithLoadBlock:.  I dunno.  Anyway, we could skip the one for ASDisplayNode alone, but I could imagine a couple contrived (probably not important) cases where it could avoid subclassing (imagine creating an image node and wanting to install a gesture recognizer on it upon loading, without subclassing)

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -15,16 +15,24 @@
...
  */
 typedef UIView *(^ASDisplayNodeViewBlock)();
+
 /**
  * CALayer creation block. Used to create the backing layer of a new display node.
  */
 typedef CALayer *(^ASDisplayNodeLayerBlock)();
 
 /**
+ * ASDisplayNode loaded callback block. Used for any additional setup after node's layer/view has been loaded
```
appleguy: Not essential now, but we should 1) mention that this is always called on the main thread, and 2) maybe we should AppleDoc these :)

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -66,14 +74,36 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
...
+/**
  * @abstract Alternative initializer with a block to create the backing layer.
  *
  * @param viewBlock The block that will be used to create the backing layer.
+ * @param didLoadBlock The block that will be called after the layer created by the layerBlock is loaded
  *
  * @return An ASDisplayNode instance that loads its layer with the given block that is guaranteed to run on the main
  * queue. The layer will render synchronously and -layout and touch handling methods on the node will not be called.
  */
-- (id)initWithLayerBlock:(ASDisplayNodeLayerBlock)viewBlock;
```
appleguy: lol that the name was incorrect

PR #655, opened 2015-09-15
Relayout table view cell nodes if there is a mismatch between content view size and the node's constrained size

nguyenhuy:
- Above is the generic case. Correctly handling it means relayout when the table view enters or leaves editing mode is taken care of as well.
- Async data source API removal: In a table view, cell nodes should always fill its content view and table view widths. Thus async data source can no longer provide custom constrained size for cell nodes. This removal allows table view to better handle relayout.
- Some more tests are added to ASTableViewTests to check against use cases handled in this diff.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -128,6 +128,7 @@ static BOOL _isInterceptedSelector(SEL sel)
 - (void)didTransitionToState:(UITableViewCellStateMask)state
 {
   [self setNeedsLayout];
+  [self layoutIfNeeded];
   [super didTransitionToState:state];
 }
 
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -16,6 +16,7 @@
 #import "ASDisplayNodeInternal.h"
 #import "ASBatchFetching.h"
 #import "ASInternalHelpers.h"
+#import "ASLayout.h"
 
 //#define LOG(...) NSLog(__VA_ARGS__)
 #define LOG(...)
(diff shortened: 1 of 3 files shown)


nguyenhuy:
With this, I'm confident that #512 is finally solved.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -104,22 +105,30 @@ - (id)forwardingTargetForSelector:(SEL)aSelector
...
+- (void)layoutSubviews
+{
+  [_delegate willLayoutSubviewsOfTableViewCell:self];
+  [super layoutSubviews];
+}
+
 - (void)didTransitionToState:(UITableViewCellStateMask)state
 {
+  [self setNeedsLayout];
   [super didTransitionToState:state];
```
appleguy: I wonder, should we actually be calling didTransitionToState: before the layout has been completed?  Maybe we should add a call to -layoutIfNeeded here.

AsyncDisplayKitTests/ASTableViewTests.m
```diff
@@ -290,4 +294,138 @@ - (void)triggerSizeChangeAndAssertRelayoutAllRowsForTableView:(ASTableView *)tab
   }];
 }
 
+- (void)testRelayoutVisibleRowsWhenEditingModeIsChanged
```
appleguy: Love the new tests — thanks!  Code coverage in ASDK continues to increase, and I think more of it is due to your tests than anyone else.

AsyncDisplayKit/ASTableView.mm
nguyenhuy: Done.

Issue #654, opened 2015-09-15
Assertion failure in -[ASCollectionView _endItemAnimationsWithInvalidationContext:tentativelyForReordering:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/UIKit/UIKit-3505.4/UICollectionView.m:3994


tettoffensive:
I am often getting this exception thrown with my ASCollectionView when reloadData is called. I have seen this pop up pretty regularly, but not all the time. Seems to happen more often when I have a breakpoint set in Xcode.

```
- (void)rangeController:(ASRangeController *)rangeController didDeleteSectionsAtIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
{
  ASDisplayNodeAssertMainThread();

  if (!self.asyncDataSource) {
    return; // if the asyncDataSource has become invalid while we are processing, ignore this request to avoid crashes
  }

  if (_performingBatchUpdates) {
    [_batchUpdateBlocks addObject:^{
      [super deleteSections:indexSet];
    }];
  } else {
    [UIView performWithoutAnimation:^{
      [super deleteSections:indexSet]; // !!!!! <NSMutableIndexSet: 0x1355b4310>[number of indexes: 1 (in 1 ranges), indexes: (0)]
    }];
  }
}
```

appleguy:
Thanks for the report--lets get it fixed.

What version of ASDK are you running?  Please try it against Master and report back. I'm cutting the 2.0 Beta branch (which is actually quite stable) in the next 24-48hrs.

tettoffensive:
I'm using the default for Cocoapods - looks like Using AsyncDisplayKit (1.2.3)

appleguy:
OK, makes sense.  Crashes are never justified, so I’d love to help get this resolved as soon as possible.

I’ve been working on 2.0 for a number of months now, and it includes a vast array of improvements.  In about 36hrs I’m giving the first talk on it (at NSSpain), and so until then and when I get back to the US, I’ll be busier than usual.

Definitely let me know what your test against master yields.  I think it’s likely that it will fix it, but it will be a high priority bug if not.

Cheers,

		— Scott

tettoffensive:
```
pod 'AsyncDisplayKit' , :git => 'https://github.com/facebook/AsyncDisplayKit.git' , :branch => 'master'
```

[...]

PR #656, opened 2015-09-16
As layout options private

rcancro:
Should fix this issue:
https://github.com/facebook/AsyncDisplayKit/issues/654


appleguy:
Thanks Ricky!

PR #657, opened 2015-09-16
Fix appledoc warnings in ASLayoutSpec.h

nguyenhuy:
No more warnings after this diff :) The main problem is that each method must have its own comment.


appleguy:
Great success!

PR #660, opened 2015-09-16
Add ASViewController

nguyenhuy:
I also update Multiplex sample to use the new ASViewController.


nguyenhuy:
This is more like a proof of concept. It's ok to not merge it :P

AsyncDisplayKit/ASViewController.m
```diff
@@ -0,0 +1,47 @@
...
+- (void)loadView
+{
+  ASDisplayNodeAssertTrue(!_node.layerBacked);
+  self.view = _node.view;
+}
+
+- (void)viewWillLayoutSubviews
+{
+  ASSizeRange constrainedSize = ASSizeRangeMake(self.view.bounds.size, self.view.bounds.size);
+  [_node measureWithSizeRange:constrainedSize];
```
appleguy: Cool, I think this effectively does synchronous measurement here and means that just barely in time for -layout, there will be cached sizes.

What would happen if measureWithSizeRange were called on a background thread (on the node) and were still running at this point?  Interestingly there haven't been other cases in usage (either framework or app) where I have needed to consider that.

Lastly, small nit, but always create a local variable for a structure that needs to be accessed twice.  CGSize boundsSize = self.view.bounds.size;

examples/Multiplex/Sample/ScreenNode.h
```diff
@@ -0,0 +1,19 @@
...
+
+@interface ScreenNode : ASDisplayNode
+
+@property (nonatomic, strong) ASMultiplexImageNode *imageNode;
+@property (nonatomic, strong) ASTextNode *textNode;
+
+- (void)start;
+- (void)reload;
+
+@end
```
appleguy: Extraneous characters on this line

examples/Multiplex/Sample/ScreenNode.m
```diff
@@ -0,0 +1,156 @@
...
+  NSAttributedString *string = [[NSAttributedString alloc] initWithString:text
+                                                               attributes:attributes];
+  _textNode.attributedString = string;
+  [self setNeedsLayout];
+}
+
+- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
+{
+  return
+  [ASInsetLayoutSpec
```
appleguy: I would put this first line on the same line as the return.  More importantly, indent the following lines by 2 spaces per level, rather than just one.

AsyncDisplayKit/ASViewController.m
nguyenhuy: Yes, this is synchronous measurement. If it is instead asynchronous and still running at this point, the (backing) view will have the correct size (set previously) but look empty, because the node doesn't layout its subnodes. Once layout is done, we can call setNeedsLayout and subviews will appear correctly.

examples/Multiplex/Sample/ScreenNode.m
nguyenhuy: Yeah, this is CK's convention. Old habits die hard :)

AsyncDisplayKit/ASViewController.m
appleguy: Right — it's totally fine for this to be synchronous.  Later, we can add the API to allow a user to trigger preparation of the view in advance, which would include layout.  My question is - what if a user dispatch_async'd a call to measure:, and then pushed the view controller?

The answer may be that we get bad behavior.  That's also not a big deal right now, as it would be unusual to manually call measure: in that way and then push the view — but it will be a key part of the final version of 2.0 to support kicking off asynchronous layout, and then having a call that is able to block on the remainder of it (e.g. "-ensureMeasurementComplete" or something — that is probably not a good name).  It would return immediately if done, otherwise wait on the background operation.

nguyenhuy: Sorry for misunderstood your question. That's indeed a tricky question. Yeah I think the behaviour would be undefined, at least for the current implementation. My main reason is that the async layout pass can finished right before or after any of these events: VC is pushed, loadView, viewWillLayoutSubviews, etc. If we have block on the remainder in, say, viewDidAppear, we may be able to do final set up there in case the background operation missed these events.

(There is a high chance that I'm talking non-sense here, please take it with a grain of salt haha)

AsyncDisplayKit/ASViewController.m
```diff
@@ -0,0 +1,47 @@
...
+
+@implementation ASViewController
+
+- (instancetype)initWithNode:(ASDisplayNode *)node
+{
+  if (!(self = [super init])) {
+    return nil;
+  }
+  
+  ASDisplayNodeAssertNotNil(node, @"Node must not be nil");
```
SRandazzo: You can now modify your method signature + parameters to indicate that `node` must not be nil by doing the following:
`- (instancetype)initWithNode:(ASDisplayNode *_Nonnull)node` ( XCode 7+)
`- (instancetype)initWithNode:(ASDisplayNode * nonnull)node` (Xcode 6.4)

https://developer.apple.com/swift/blog/?id=25

nguyenhuy:
Nits picked.

AsyncDisplayKit/ASViewController.m
nguyenhuy: That's great. Thanks!

P/S: actually `nonnull` must be right after open parenthesis.
And why is this non*null* instead of non*nil*?

examples/Multiplex/Sample/ScreenNode.h
nguyenhuy: :confused: :question:

PR #665, opened 2015-09-17
Clean up Cocoa Touch framework target

jflinter:
This PR cleans up the Cocoa Touch framework target, used by Carthage etc. Specifically, it adds several missing headers to the headers build phase, as well as adds a `module.modulemap` file. We have to create this file (instead of using the auto-generated one that Xcode will produce for you) because of `ASDisplayNode+Subclasses.h` and `ASControlNode+Subclasses.h`, which aren't included in the umbrella header and thus can't be imported directly into Swift projects.

(There are a lot of changes to the `.pbxproj` file, but all I did was alphabetize the headers and compile sources build phases for easier comparison.)


appleguy:
Very interesting, thanks @jflinter.  I would be willing to reconsider the currently-intentional non inclusion of the +Subclasses header.  The main reason for this was to hide methods that should not need to be called outside of a subclass, but technically we could use runtime assertions to defend (and educate) against this.  Any ideas are welcome, depending on the impact / benefit.

adly-holler:
@appleguy If we have to pick between:

- Can't subclass `ASDisplayNode` in Swift
or
- Reveal internal `ASDisplayNode` methods to other consumers

I think the latter is way less restrictive, and kind of a no-brainer. Using AsyncDisplayKit under Swift but having to use Objective-C to subclass `ASDisplayNode` would be awful I bet

jflinter:
So, sorry for not being clearer - this PR now allows you to import those headers into Swift projects. In other words, if you want to subclass `ASDisplayNode` in Swift, you now just add
```swift
import AsyncDisplayKit
import AsyncDisplayKit.ASDisplayNode_Subclasses
```
to the top of your file.

adly-holler:
Oh! OK great

PR #667, opened 2015-09-17
Update Podspec to support Cocoapods 0.38.x

### LINKED ISSUE #659: Update Podspec to support 0.38.2 and later
Something to do with do with setting `user_target_xcconfig`.

Related Facebook project: https://github.com/facebook/pop/issues/264
Related Cocoapods discussion: https://github.com/CocoaPods/CocoaPods/issues/3813#issuecomment-127218342

levi:
Fixes #659

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: objective-c
 before_install:
   - brew update
   - brew reinstall xctool
-  - gem install cocoapods -v 0.37.2
+  - gem install cocoapods -v 0.38.2
   - gem install slather
   - xcrun simctl list
 install: echo "<3"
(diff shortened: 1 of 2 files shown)


nguyenhuy:
Hm, Travis is still using CocoaPods 0.37.2 because it is explicitly specified [here](https://github.com/facebook/AsyncDisplayKit/blob/master/.travis.yml). You may want to change that as well.

appleguy:
I'm a little intimidated by this Cocoapods version because it broke so much stuff...Not just ASDK, but also Pinterest.  I appreciate you looking into it, though, because it has to be done eventually.  Once Travis is succeeding, I'll hit the button and pray in several languages.

levi:
@appleguy the same thing happened to us a while ago. We recently upgraded to 0.38.2 and seem to be running fine. I just updated travis—let's see what happens.

appleguy:
Thanks for forging the path.  I will now promptly flee my computer and pretend success is inevitable, as is not too hard to do with such competent help from folks like you :).

PR #668, opened 2015-09-18
Fill image context with background color if image node is opaque

adly-holler:
This way you can get a performance boost by marking your image nodes as opaque. Before this PR, no matter what the background color, opaque image nodes with partially-transparent images always show a black background.

### DIFF
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -231,7 +231,7 @@
   // if view is opaque, fill the context with background color
   if (parameters.opaque && parameters.backgroundColor) {
     [parameters.backgroundColor setFill];
-    [[UIBezierPath bezierPathWithRect:{ .size = backingSize }] fill];
+    UIRectFill({ .size = backingSize });
   }
 
   [image drawInRect:imageDrawRect];
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -227,6 +227,12 @@
   // Use contentsScale of 1.0 and do the contentsScale handling in boundsSizeInPixels so ASCroppedImageBackingSizeAndDrawRectInBounds
   // will do its rounding on pixel instead of point boundaries
   UIGraphicsBeginImageContextWithOptions(backingSize, parameters.opaque, 1.0);
+  
+  // if view is opaque, fill the context with background color
+  if (parameters.opaque && parameters.backgroundColor) {
+    [parameters.backgroundColor setFill];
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASImageNode.mm
```diff
@@ -227,6 +227,12 @@ + (UIImage *)displayWithParameters:(_ASImageNodeDrawParameters *)parameters isCa
   // Use contentsScale of 1.0 and do the contentsScale handling in boundsSizeInPixels so ASCroppedImageBackingSizeAndDrawRectInBounds
   // will do its rounding on pixel instead of point boundaries
   UIGraphicsBeginImageContextWithOptions(backingSize, parameters.opaque, 1.0);
+  
+  // if view is opaque, fill the context with background color
+  if (parameters.opaque && parameters.backgroundColor) {
+    [parameters.backgroundColor setFill];
+    [[UIBezierPath bezierPathWithRect:{ .size = backingSize }] fill];
```
appleguy: Let's use a more efficient method to do this.  Instead of creating a path and requiring CG to understand the shape of the path (which it may be able to determine is a rectangle, and short-circuit...but more likely, is not able to) — let's just call CGContextFillRect() :).

appleguy:
Great find, thanks!  Did you discover this by 1) using an image that is actually transparent in areas, 2) over a background that is a solid non-black color?  Just curious.  Opaque drawing is great, not just because of the GPU compositing savings — but the solid 25% memory savings :)  (helps both allocation time, and obviously footprint)

Sorry for the delay in reviewing.  Naturally, this is close to merge-ready, but I'd like to see it changed to use the most direct CG method first.

adly-holler:
@appleguy Yeah, both actually! Transparent image over white background (a little compose icon in a table cell)

Also thanks for helping me find a new function today ^^

appleguy:
Retried now-/will see the result in the morning.  Thanks for the quick turnaround!

Really impressed to see the diff for the collection view issue, and excited to look at that too.  Thanks so much for jumping in with fix-guns blazing!!

adly-holler:
Thanks man! I have a lot of respect for your work so I'm glad to help out!

appleguy:
@adly-holler haha, thanks!!  I wonder if we need to check for alpha == 1.0 here on the color.  I guess theoretically we should, although it is probably a fine behavior if your partial-alpha color gets drawn into a non-alpha backing store (it should theoretically just make it opaque?)

adly-holler:
Good point. Testing in a Playground shows that it premultiplies the color components, so it renders as if it were blended onto a black background, which I'd say is reasonable. So if you use transparent blue as the background color for an opaque node you'll see a dark blue background.

PR #664, opened 2015-09-16
ObjectiveC: Nullability, Lightweight Generics

Adlai-Holler:
Killer framework! This PR updates the public API Objective-C for Xcode 7 so the framework integrates with Swift better plus adds some compiler checks.

PR #671, opened 2015-09-19
Travis use Xcode 7

Adlai-Holler:
@appleguy Pursuant to https://github.com/facebook/AsyncDisplayKit/pull/664#issuecomment-141697512 , here's a branch that just has the Travis updates. On my local machine using Xcode 7A220 I still get the same unrecognized selector errors for category properties like `CALayer.asyncdisplaykit_node`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: objective-c
+osx_image: xcode7
 before_install:
   - brew update
   - brew reinstall xctool
--- a/build.sh
+++ b/build.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 # **** Update me when new Xcode versions are released! ****
-PLATFORM="platform=iOS Simulator,OS=8.1,name=iPhone 6"
-SDK="iphonesimulator8.1"
+PLATFORM="platform=iOS Simulator,OS=9.0,name=iPhone 6"
+SDK="iphonesimulator9.0"
 
 
 # It is pitch black.
--- a/examples/Swift/Sample.xcodeproj/project.pbxproj
+++ b/examples/Swift/Sample.xcodeproj/project.pbxproj
@@ -129,6 +129,7 @@
 		050E7C6619D22E19004363C2 /* Project object */ = {
 			isa = PBXProject;
 			attributes = {
+				LastSwiftUpdateCheck = 0700;
 				LastUpgradeCheck = 0600;
 				ORGANIZATIONNAME = Facebook;
 				TargetAttributes = {
--- a/examples/Swift/Sample/ViewController.swift
+++ b/examples/Swift/Sample/ViewController.swift
@@ -27,7 +27,7 @@ class ViewController: UIViewController, ASTableViewDataSource, ASTableViewDelega
     self.tableView.asyncDelegate = self
   }
 
-  required init(coder aDecoder: NSCoder) {
+  required init?(coder aDecoder: NSCoder) {
     fatalError("storyboards are incompatible with truth and beauty")
   }
 


appleguy:
Interesting, so it is an Xcode 7 specific issue with our test harness I would suppose?  Testing often results in test-specific issues :(.

This code has been working since the late versions of Xcode 4.6 or so, and thus although it may need a change to handle the Xcode changes, I don't think it has been recently broken.

Can you confirm running the tests locally reproduces the issue?  I'm away from a computer but if no one has figured it out by the time the plane boards, I will debug.

adly-holler:
Yep, running it locally does reproduce the issue. 

The issue isn't test-specific – any code that calls `CALayer.asyncdisplaykit_node` would get the same `unrecognized selector` error.

appleguy:
Spent some time looking into this, but have run out.  I have a number of other priorities that I can't bump in the next couple weeks, and am worried I just won't have time to get the testing system running on Xcode 7 in September.

If anyone wants to take a deeper look, the help would be thoroughly appreciated!  I got past several of the issues, but it remains a mystery why they are popping up at all (e.g., the implementation counterpart to the CALayer @dynamic for asyncdisplaykit_node indeed seems to be missing, and adding the implementation fixes that issue, but uncovers a bunch of similar ones).  Before forging ahead to blindly "fix" anything "missing", we should go back to 6.4 to understand what might have changed / how it was working before.  cc @nguyenhuy, @levi

nguyenhuy:
Thanks for cc'ing me, Scott. I strongly agree that we need to understand the root cause first. I can spend time on this issue later this week, or even early next week. I've been insanely unproductive these days due to my serious procrastination problem and/or having a cold. Anyway, if anyone can work on this before me, please do.

adly-holler:
Note: I figured out how it's working normally – `CALayer` just lets you do arbitrary key-value storage, and it'll generate accessor implementations for any properties you add in categories with `@dynamic`.

appleguy:
Aha!  Right, somehow I’d forgotten that.  I wonder why the hell that is no longer working with Xcode 7, only in test targets…

appleguy:
I spent more time investigating this.  I'm actually pretty worried at this point, because I don't have a lead or a good idea for the debugging approach.

The tests are definitely running within a UIKit app.  The stack traces that show the calling of the @dynamic methods (which crash for doesNotRespondToSelector) clearly show UIApplicationMain at the bottom of the stack.  I don't think there should be any difference between the way CALayer handles the @dynamic methods in a regular running session vs the test app, but there quite clearly is.

There is at least one place where we may depend on the fact that CA makes the properties KVC-compliant when it generates them itself.  I'm not sure if this would be supported if we manually implemented these methods, and would really like to avoid doing that until someone is able to help us understand the root cause for the behavior change.  The biggest mystery here is not only why the same code behaves differently under Xcode 6.4 vs 7.0, but also why it only affects the test application. 

I studied the project file / build configurations, but didn't see anything suspicious in the setup for the test environment.

I'm downloading Xcode 7.1 Beta 2 to try, although I don't expect it to fix the issue.

appleguy:
Xcode 7.1 Beta 2 has the same issue, as does 7.0.1.

appleguy:
I've opened a Radar and pinged the wizards at Apple to help investigate this unusual issue.

Hoping to unblock the wonderful https://github.com/facebook/AsyncDisplayKit/pull/664 as soon as possible!

appleguy:
Sadly there is zero traction whatsoever on the Apple issues, so it is going to take months to land this.  I am going to close the PR until we are on Xcode 7, and then reopen.

appleguy:
I worked around Apple's frustrating, still-unfixed bug with @dynamic properties on CALayer by avoiding use of them.  It's a suboptimal solution as we lose certain capabilities, but I'm excited to get this in.  A truly happy Christmas 2015, lol!

https://github.com/facebook/AsyncDisplayKit/pull/994

PR #672, opened 2015-09-19
Fix collection view data inconsistency

### LINKED ISSUE #654: Assertion failure in -[ASCollectionView _endItemAnimationsWithInvalidationContext:tentativelyForReordering:], /BuildRoot/Library/Caches/com.apple.xbs/Sources/UIKit/UIKit-3505.4/UICollectionView.m:3994

adly-holler:
This resolves #654

We could remove the flag and just call `[self layoutIfNeeded]` immediately after `[super reloadData]`
  - + Reduces complexity of ASCollectionView
  - - Relies on implementation details of `UICollectionView`
  - - Performance cost

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -131,6 +131,15 @@ static BOOL _isInterceptedSelector(SEL sel)
 
 @property (atomic, assign) BOOL asyncDataSourceLocked;
 
+/**
+ If YES, the `UICollectionView` will reload its data on next `layoutSubviews` so we should
+ not forward any updates to it.
+ 
+ Rationale:
+ In `reloadData`, a collection view invalidates its data and marks itself as needing reload, and waits until `layoutSubviews` to requery its data source. This can lead to data inconsistency problems. Say you have an empty collection view. You call `reloadData`, then immediately insert an item into your data source and call `insertItemsAtIndexPaths:[0,0]`. You will get an assertion failure saying "Invalid number of items in section 0. The number of items after the update (1) must be equal to the number of items before the update (1) plus or minus the items added and removed (1 added, 0 removed)." The collection view never queried your data source before the update to see that it actually had 0 items.
+ */
+@property (nonatomic) BOOL pendingReloadData;
+
 @end
 
 @implementation ASCollectionView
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -638,7 +648,7 @@ - (void)rangeControllerBeginUpdates:(ASRangeController *)rangeController {
 - (void)rangeController:(ASRangeController *)rangeController endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion {
   ASDisplayNodeAssertMainThread();
 
-  if (!self.asyncDataSource) {
+  if (!self.asyncDataSource || _pendingReloadData) {
     if (completion) {
       completion(NO);
```
adly-holler: Should we call `completion(YES)` if we suppressed the update due to a pending reload?

eanagel:
@adly-holler, this is some great detective work! The idea of simply calling layoutIfNeeded after our reloadData is very attractive indeed. (Would that actually drop this down to a 1-line commit?) You mention performance cost as a drawback to this approach, could you expand on that a bit?

eanagel:
Oh yes and another question, is this a potential problem for ASTableView as well? Have you explored that side of the equation?

adly-holler:
@eanagel Thanks! It would indeed make this a 1-line commit, assuming that `reloadData` -> `setNeedsLayout` which I'd bet it does. The performance cost comes from laying out the cells when they don't necessarily need to be laid out. For a simple example, consider if you called `reloadData` 5 times during the same frame. The collection view cells would get laid out 5 times for one frame.

appleguy:
Will analyze more, merging as it does fix the test project submitted recently.

adly-holler:
Note: Ported the sample project to table view and I'm not seeing the same issue.

PR #673, opened 2015-09-21
Align ASEditableTextNode to pixel boundaries

adly-holler:


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -212,9 +200,9 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
   CGSize renderSizePlusShadowPadding = UIEdgeInsetsInsetRect(CGRect{CGPointZero, rendererSize}, shadowPadding).size;
   ASDisplayNodeAssert(renderSizePlusShadowPadding.width >= 0, @"Calculated width for text with shadow padding (%f) is too  narrow", constrainedSizeForText.width);
   ASDisplayNodeAssert(renderSizePlusShadowPadding.height >= 0, @"Calculated height for text with shadow padding (%f) is too short", constrainedSizeForText.height);
-
```
appleguy: We should actually port the -calculateSize method to be -calculateLayout(...); it's probably not appropriate for -layoutSpecThatFits:.  I think we'll formally mark -calculateSize as deprecated soon.

PR #675, opened 2015-09-22
added ASViewController to the dynamic framework target to fix carthag…

crylico:
…e support

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -140,6 +140,8 @@
 		205F0E211B376416007741D0 /* CGRect+ASConvenience.h in Headers */ = {isa = PBXBuildFile; fileRef = 205F0E1F1B376416007741D0 /* CGRect+ASConvenience.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		205F0E221B376416007741D0 /* CGRect+ASConvenience.m in Sources */ = {isa = PBXBuildFile; fileRef = 205F0E201B376416007741D0 /* CGRect+ASConvenience.m */; };
 		242995D31B29743C00090100 /* ASBasicImageDownloaderTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 242995D21B29743C00090100 /* ASBasicImageDownloaderTests.m */; };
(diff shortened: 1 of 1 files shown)


crylico:
Looks like CI failed because of network problems?

appleguy:
Rebuilding.

appleguy:
@crylico thanks for the fix!  Sorry that various commits keep missing this.  I wonder what the easiest way is to ensure Carthage is tested during the unit test cycle.

crylico:
It's a hard issue - at first thought, maybe running a script that checks the Compile Sources of the static lib target and the dynamic framework target are the same?

This seems a little hacky, though

crylico:
Another possibility would be to include some kind of demo project that is also tested that builds AsyncDisplayKit via Carthage, then make sure the demo project tests pass as well?

appleguy:
@crylico what is the failure mode?  Yeah, I think the example project is the right idea.  I believe right now if you do a simple Duplicate in Finder (cmd-D) on any project in the examples/ directory, and then just give the project a name an modify it, you have a new example project.

We should check, but I believe the build system automatically compiles all of them.  So if we put one in with Carthage, theoretically it could catch this?

crylico:
So, checking out the commit just before this merge and creating a brand new sample project, then following the iOS Getting Started instructions at [Carthage](https://github.com/Carthage/Carthage), then doing `@import AsyncDisplayKit;`, the build error (in Xcode, not Carthage) is `'AsyncDisplayKit/ASViewController.h' file not found`.

Can you think of any other issues that could arise?

appleguy:
@crylico I've actually never used Carthage before, or even the @import flag :).  There are a few other people who have contributed to Carthage support (and filed both Issues and PRs), and it would be great to loop them in for any suggestions.

I know at least a couple were hoping to create a sample project as well, and might actually have a specific strategy in mind for the simplest way to create the sample.  I think duplicating an example is a fine starter approach for most sample projects, although it may be the case that with Carthage, starting from clean build settings (rather than a project / workspace originally managed by Cocoapods) is the better idea?

crylico:
@appleguy if there are already people with some ideas in mind, great! I'd love to hear their inputs :) If not, I'd be more than happy to make the example project with the bare minimum to trigger an error if there are files missing.  

I agree that starting with a clean project would be a good idea, though the README sounds a bit like all projects in the examples folder get "pod install" for CI, which obviously would error out for a carthage project. Is it safe to just create a new project that doesn't depend on cocoapods?

PR #680, opened 2015-09-23
Placeholder bug in ASEditableTextNode

rcancro:
When setting a default placeholder and an attributedString before `_textKitComponents.textView` was created, the placeholder and the string were both appearing in the text node.

`_updateDisplayingPlaceholder` is called in `setAttributedString`, but since  `_textKitComponents.textView` is nil the placeholder was not hidden. Calling `_updateDisplayingPlaceholder` after `_textKitComponents.textView` loads fixes this.

### DIFF
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -143,6 +143,7 @@
   _textKitComponents.textView.accessibilityHint = _placeholderTextKitComponents.textStorage.string;
   configureTextView(_textKitComponents.textView);
   [self.view addSubview:_textKitComponents.textView];
+  [self _updateDisplayingPlaceholder];
 }
 
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize


AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -143,6 +143,7 @@ - (void)didLoad
   _textKitComponents.textView.accessibilityHint = _placeholderTextKitComponents.textStorage.string;
   configureTextView(_textKitComponents.textView);
   [self.view addSubview:_textKitComponents.textView];
+  [self _updateDisplayingPlaceholder];
```
appleguy: Thanks for the fix!  In general I think it would be better to call this method before addSubview:, as it's possible UITextView does some work upon being added to the hierarchy that is affected by the placeholder stuff (although unlikely, just as a general practice it's good to have stuff set up before adding to hierarchy).

appleguy:
Build break is my fault, re-running it and will merge afterwards.

PR #689, opened 2015-09-28
Implementation of Synchronous Concurrency features for AsyncDisplayKit 2.0

appleguy:
This provides features that allow clients to choose if they would prefer to block
on the completion of unfinished rendering, rather than allow a placeholder state to
become visible.

Internal-only features that facilitate the implementation of public API:
-[_ASAsyncTransaction waitUntilComplete]
-[ASDisplayNode recursivelyEnsureDisplay]

Also provided are two public APIs that are powered by those:
-[ASCellNode setNeverShowPlaceholders:], which integrates with both Tables and Collections
-[ASViewController setNeverShowPlaceholders:], which should work with Nav and Tab controllers.

Lastly, on ASDisplayNode, a new property .shouldBypassEnsureDisplay allows individual node types
to exempt themselves from blocking the main thread on their display.  This is on by default for ASNetworkImageNode and ASMultiplexImageNode, for reasons explained in the comments.

By implementing the feature at the ASCellNode level rather than ASTableView & ASCollectionView,
developers can retain fine-grained control on display characteristics.  For example, certain
cell types may be appropriate to display to the user with placeholders, whereas others may not.

The diff overall is extensively commented, including areas that were otherwise unchanged.

Essential follow-up work I have planned includes unit tests, revisiting names, and finalizing the header locations of definitions.


appleguy:
Feel free to share comments on this work, including API / naming ideas or feature requests!

Issue #684, opened 2015-09-25
Collection/Table view changes should be sorted


Adlai-Holler:
`UITableView` and `UICollectionView` allow you to make changes in any order during batch updates. When `endUpdates` is called, the changes are sorted and applied in a safe order. The ASDK versions don't, so if your changes (say, from an `NSFetchedResultsController`) aren't submitted in a specific order, you'll get an index-out-of-bounds crash sooner or later.

Here's the order changes must be applied in, so we should update `ASDataController` to accommodate this:

1. Updated/reloaded items
1. Deleted items, descending order
1. Deleted sections, descending order
1. Inserted sections, ascending order
1. Inserted items, ascending order

EDIT: Oh and moves are tricky. Best bet is to break them up into delete/insert, because the `fromIndexPath` is pre-update while the `toIndexPath` is post-update. So tell the `UITableView` about the move whenever you want, but update the array by breaking the move into two operations that'll get sorted separately.

smyrgl:
This has been a seriously problem for my app and I'm not sure your ordering changes are entirely correct (I suspect it is the moves that are causing the problem as I tend to have a lot of these) but I'm trying to make your workaround function and I'll update if I can point to specific steps that are failing.

smyrgl:
Wait one thing about your solution is confusing to me.  

> So tell the UITableView about the move whenever you want, but update the array by breaking the move into two operations that'll get sorted separately.

So in UITableView you update your backing datastore (I assume that's what you mean by array) before you start telling the table view about the batch updates.  Are you suggesting that you sort the changes and then update a backing datastore incrementally as the batch updates are occurring (using the lock/unlock calls on the ASTableViewDatasource to ensure consistency)?  I could try that but it would require adding a storage object which mirrors what the table view state is at any given time.

Adlai-Holler:
Thanks for engaging on this @smyrgl! As a _user_ of ASDK, there's currently no safe way to do moves intermingled with other changes in a batch update. This is because `ASTableView` immediately applies both the delete and insert for the move, which is unsafe. **This is a pretty major problem.**

For example, say you have 1 section with 7 items `ABCDEFG`. You want to do a batch update with `(move 5 to 1), (insert X at 0)`, so you expect the end result to be `XFABCDEG`.
If you tell ASTableView about the changes in that order you'll end up with `XAFBCDEG` which is wrong. If you tell ASTableView about the changes in reverse order you'll end up with `XEABCDFG`

[...]

PR #693, opened 2015-09-29
Data Controller Batch Change Sorting

Adlai-Holler:
:warning: Not ready to merge :warning: 

I may be way off in the woods here, so I'm looking for comments on the general direction before I continue to testing and refining. Don't hold back!

- Create a new `_ASHierarchyChangeSet` class that stores and processes changes. The intention here is, you submit your changes with their animation options in arbitrary order, and when you're done it generates sorted changes, coalesced by animationOptions. This way we have the minimum possible number of updates but still in a safe order for use by ASDataController.
- In ASDataController our enqueued edit command blocks are now `(_ASHierarchyChangeSet -> Void)`. During `endUpdates` we create a change set, run all the blocks to load it up with data, then call `markCompleted` so the changeset does its logic, then we do `accessDataSourceWithBlock` and go through the sorted, coalesced changes, applying them using the same code that we USED to use in the edit command block.

I copied & pasted the edit command code pretty directly, so I'm not sure if e.g. we can get rid of some of the `[editTransactionQueue waitUntilAllOperationsAreFinished]` or what the fastest safe behavior we can achieve there is.

@nguyenhuy @appleguy

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -374,6 +374,10 @@
 		B350625D1B0111740018CF92 /* Photos.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 051943141A1575670030A7D0 /* Photos.framework */; };
 		B350625E1B0111780018CF92 /* AssetsLibrary.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 051943121A1575630030A7D0 /* AssetsLibrary.framework */; };
 		B350625F1B0111800018CF92 /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 058D09AF195D04C000B7D73C /* Foundation.framework */; };
+		CCE3AEC41BBB02DA00AFAB54 /* _ASHierarchyChangeSet.h in Headers */ = {isa = PBXBuildFile; fileRef = CCE3AEC21BBB02DA00AFAB54 /* _ASHierarchyChangeSet.h */; settings = {ASSET_TAGS = (); }; };
+		CCE3AEC51BBB02DA00AFAB54 /* _ASHierarchyChangeSet.h in Headers */ = {isa = PBXBuildFile; fileRef = CCE3AEC21BBB02DA00AFAB54 /* _ASHierarchyChangeSet.h */; settings = {ASSET_TAGS = (); }; };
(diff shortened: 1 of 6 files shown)


Adlai-Holler:
#684

nguyenhuy:
Great progress, @Adlai-Holler. Thanks a ton.

A quick look at the diff gives me the impression that there are just too many changes in ASDataController. That class is very... hot, mainly because it heavily deals with concurrency. Is there any way we can minimize the changes there? 

Not sure this is a good idea. What I'm having in mind is that ASDataController doesn't know about change set at all. Instead, there is a new subclass of ASDataController that intercepts edit calls during a transaction and put them to a change set. Then in endUpdates (of the subclass), it loops through the sorted change set and forward each edit call to ASDataController. That way, the new class encapsulates the knowledge about change set and ASDataController remains relatively untouched. What do you think?

appleguy:
This is really awesome, thanks so much @Adlai-Holler for your large-scale contributions!  I will review in detail this weekend.

Let me start by saying that I'm not afraid of a change that hugely changes ASDataController — obviously it would need to be rigorously tested, but let's face it, the class needs some upgrades.  The critical piece: we must get the architecture RIGHT.  It isn't worth the testing pain of a really big change if we can spot any difficult limitations of its structure that wouldn't be resolvable with later improvements moving in the same direction.

That said, it does seem likely that we could solve this problem with a more surgical change.  I'd buy it that this direction may be BETTER despite the higher initial risk, if we believe it results in code that is easier to debug, read, maintain, extend, or more performant — or any combination of a few of those.

@Adlai-Holler, what's your general philosophy on this?  @levi, @eanagel - interested to hear your thoughts.  Don't worry guys, I won't merge something like this without personally helping to test it heavily with stressful use cases in Pinterest and upgraded thrash testers, and I will also not merge something that breaks any client's app — so let's critique and brainstorm what the ideal architectural approach is before worrying about the risk of any given change.

Adlai-Holler:
@nguyenhuy I really dig your idea! It's a very natural way to implement this buffering/sorting.

@appleguy You pretty much said what I'd say, that size of change is less important than suitability of the architecture and that something as fiddly as this demands an insane amount of tests. 

As I look back on it, this PR is compromised – it goes too far for a newcomer to just drop in, but not far enough to justify it. I'll close it and create a new PR that has a better benefit:fear ratio. Sometime in the future we'll update the architecture together and I can discharge my obsession with this problem at that time. Cheers guys!

Issue #669, opened 2015-09-18
How to add a ASCollectionView header


duxinfeng:
Hi,I added a header in the following manner，but crash

error info:
*** Assertion failure in -[ASCollectionView setAsyncDataSource:]
AsyncDisplayKit/AsyncDisplayKit/ASCollectionView.mm:224

```
 _collectionView = [[ASCollectionView alloc] initWithFrame:CGRectZero collectionViewLayout:layout asyncDataFetching:YES];
    _collectionView.asyncDataSource = self;
    _collectionView.asyncDelegate = self;
    [_collectionView registerClass:[ToolsHeader class] forSupplementaryViewOfKind:UICollectionElementKindSectionHeader withReuseIdentifier:@"header"];
```
```
- (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView viewForSupplementaryElementOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath
{
    ToolsHeader *header = [_collectionView dequeueReusableSupplementaryViewOfKind:kind withReuseIdentifier:@"header" forIndexPath:indexPath];
    header.title.text = @"ABCD";
    return header;
}
```
Thank you for your answer！

appleguy:
Supplementary views aren't supported by ASCollectionView right now; there was an assertion added for this a while back that I believe is still in 1.2.3 and Master (if you have 1.2.2, it may not be there).

@levi has been working on implementing support for supplementary views, but also has a full-time job.  Levi, are there any updates in terms of the strategy and path from here to having support?

levi:
@appleguy, @duxinfeng: I have a pretty clear picture as where to head with adding supplementary view support for flow layouts. I'm currently sketching out additions to ASDataController specific to ASCollectionView. I'm planning to have a roughed-out data controller implementation in a couple of days and will put up a WIP pull request to begin community feedback and iteration.

appleguy:
Really great, thanks @levi for the update!  Can't wait to see, don't hesitate to share something super early even if the initial diff doesn't actually work with a real UI.  I'm very interested to start considering what would be necessary to go from that to a solution that supports non-flow layouts too (may not be feasible though, as we've discussed some...)

Issue #691, opened 2015-09-28
[ASCollectionView] Use UICollectionViewLayout invalidation events to relayout


levi:
Issue #690 brought up a good point I've been thinking about as I was putting together the supplementary view architecture briefly this weekend. The current data controller integration with ASCollectionView reiles upon layout invalidation from reloading/inserting/deleting/moving items or @nguyenhuy's new relayout items based upon the collection view's bounds changes. These behaviors closely model the specific implementation of a table view, but it's clear as we have begun to mature the behavior of our collection view most of the sizing events and values (namely constrainedSizeForNode:) have been built around the powerful Collection view layouts.

I've been doing a bit of research into the `invalidateLayout` and `invaldateLayoutWithContext:` methods. While the provided context is not as granular as I initially hoped, I believe these events are great candidates for us to use as hooks for relayout events in the collection view's data controller.

Moreover, as we query the collection view for constrained sizes I think we should start asking the collection view layout for specific values. This will move the layout (constrainedSizeForNode:) code we currently have in ASCollectionView into a separate place that is specific to the current collection view layout.

With that, a couple of solutions have come to mind:
1. Make ASDataController an abstract class that table view and collection view have concrete implementations of. This will give each view the ability to add custom behavior such as supplementary views for collection views and header/footer nodes for tables views.
2. Once that we have a concrete ASCollectionDataController with supplementary view support, the data controller interfaces with a strategy object that provides all the necessary answers to questions from the data controller about the current layout. For example, "how many supplementary views are in this section?", "what's the constrained size for this node?". This strategy is provided by the client of ASCollectionView and either set directly as a property on ASCollectionDataController or is composed in ASViewController and its values returned at specific points of the data controllers data source method (I'm leaning toward the latter). The default limitation would be a generic strategy for flow layouts.
3. We migrate ASCollectionView to start node invalidation and relayout on ‘invalidateLayout‘ events. I already have the proxying of invalidateLayout working in a branch, but it's beyond the scope of implementing supplementary views into ASCollectionView. This will allow us to support ‘itemSize‘ changes in a flow layout, for example, and respect the layout invalidation events that can at times use animation transitions to change the item's sizes. 
#1 and #2 is where I'm planning to go this week to get my branch building and running. I feel that it should be a flexible enough solution to test and also to continue composing as custom layout authors investigate using their layouts with supplementary views. #3 is where I'd like to start discussion, as it's a little more involved and not something I'm planning to do this week.

[...]

Issue #588, opened 2015-07-28
collectionView supplementary view crashes with version 1.2.3


wieseljonas:
Works fine with 1.2.2.

With 1.2.3 I get an error like so

   'UICollectionView is planned to support supplementary views by September 2015. You can work around this issue by using standard items.'

This is how I add the supplementary view:

this in 'viewDidLoad'

    let headerlayout = UICollectionViewFlowLayout()
        headerlayout.headerReferenceSize = CGSizeMake(self.view.frame.size.width, 238)
        collectionView.collectionViewLayout = headerlayout
        
        /* Register the header's nib */
        collectionView.registerNib(UINib(nibName: "ProfileHeaderView", bundle: nil), forSupplementaryViewOfKind: UICollectionElementKindSectionHeader, withReuseIdentifier: "ProfileHeaderView")

and this in the 'viewForSupplementaryElementOfKind'

    func collectionView(collectionView: UICollectionView!, viewForSupplementaryElementOfKind kind: String!, atIndexPath indexPath: NSIndexPath!) -> UICollectionReusableView! {
        if kind == UICollectionElementKindSectionHeader {
            let headerView = collectionView.dequeueReusableSupplementaryViewOfKind(kind,
                withReuseIdentifier:"ProfileHeaderView", forIndexPath: indexPath) as! ProfileHeaderView
            return headerView
            
        } else {
            return UICollectionReusableView()
        }
        
    }

appleguy:
@wieseljonas Thanks for reporting this.  I'm interested to hear that supplementary views have been working in your app with prior versions, as they have never been properly supported.

Have you seen any crashes in your app prior to this assertion being added in 1.2.3?  I'd be happy to remove it if it turns out there are some use cases that are working properly, even though other uses cases (especially those that edit the collection contents, such as loading more from the network) are not.

I do hope to properly implement supplementary view support in the next 2-4 weeks, and will roll a 1.2.4 that either removes this exception or adds proper support.

wieseljonas:
@appleguy Thanks frou your reply. No I didn't notice any issues prior to 1.2.3. I reverted to 1.2.2 for the moment.

I do manage to add network content to the supplementary view under 1.2.2.

Please not don't use a `ASNode` for these supplementary views but a classic `UICollectionReusableView`

levi:
[...]

PR #694, opened 2015-09-30
Add support for supplementary views in ASCollectionView

levi:
:+1: Ready for landing :+1:

This patch is a bit of a big one, but something I've needed for my own usage for some time now. What I'm introducing is an extensible addition to `ASCollectionView` and `ASDataController` that supports supplementary views in any `UICollectionViewLayout` (with a bit of custom implementation) and out-of-the-box support for node-backed supplementary views in `UICollectionViewFlowLayouts`.

The core concept of this patch's "extensibility" is the introduction of the `layoutDelegate` property on `ASCollectionView` that receives `ASCollectionViewLayoutInspecting` conforming objects. An implementation of this protocol provides insight into the behavior of a collection view layout such that the data controller can query the existence of supplementary views and their index paths, something that is control entirely by the collection view's layout.

`ASDataController` has been updated to allow subclasses to hook into editing events. Its internal `completedNodes` and `editingNodes` stores have been exposed as key-value structures, allowing subclasses to register non-item/row nodes managed by the data controller. `ASCollectionDataController` is the only subclass of the data controller for now, but shows how supplementary nodes of different kinds (node types) are registered in the completed and editing backing stores and manipulated through event hooks. The benefit here is the potential of extending the data controller to support collection view and table view behaviors independently (section header/footer nodes in ASTableView, perhaps?).

Moreover, I have moved sizing-specific delegate methods from `ASCollectionViewDelegate` to `ASCollectionViewDelegateFlowLayout`, bridging the ASDK API closer to the `UICollectionViewDelegateFlowLayout` original and making `ASCollectionView` more layout-independent. Anyone relying on `collectionView:layout:insetForSectionAtIndex:` will need to update their delegate to conform to this protocol.

-- Original body --

Might be best to read through #691 before skimming through this code, as I briefly outlined the direction I'm embarking on in that issue.

TODO:
- [x] Finish up abstracting some of the internal insert/update/delete behavior in `ASDataController` to be exposed in such a way that a subclass like `ASCollectionDataController` can insert its supplementary view measure/layout passes.
- [x] Get this running in a sample project and battle test the various issues that arise
- [x] Handle section insertion/update/deletion and item insertion/update/deletion from collection view
- [x] Write tests for flow layout inspector implementation
- [x] Implement initial data loading `ASCollectionDataController` hook
- [x] Update collection modifications to support main thread layout passes
- [x] Test stored element kinds in `ASCollectionView`
- [x] Audit public method naming for consistency

Would love feedback on this. I don't have much time during the week, but have been making this a priority nights/weekends so we can have this ready for 2.0 in November.

Fixes: #669, #588, #737


appleguy:
Fantastic, thanks @levi!  I'll review as soon as I can, likely this weekend.  I'm pretty sure others will have feedback sooner.

This direction only supports Flow Layouts, correct?  Any comments to share with the community here on findings re: the difficulty / infeasibility of supporting other types of layouts?  There is some talk at Pinterest of creating a "floating header" system that will work with all layouts, and not attempt to integrate with UICollectionViewLayout directly, instead providing more limited / specific functionality to handle the most common use of supplementary views (floating section headers).  This is not a top priority for us right now though, unclear if it would be attempted before the end of the year.

Adlai-Holler:
@levi This PR is awesome! Hoping to use `AAPLCollectionViewLayout` as well as flow layout with it. @appleguy The Pinterest folks might want to look at that layout, which is contained in some [awesome sample code](https://developer.apple.com/sample-code/wwdc/2015/downloads/Advanced-Collection-View.zip) since it supports pinned headers pretty much natively although there are a couple minor bugs in it.

levi:
Yo guys, my goal is to build this extensible enough that it could be customized for a custom collection view layout. I'm only building concrete support for flow layouts right now, but this can be changed by implementing a custom implementation of `ASCollectionViewLayoutInspecting` and setting it on the `ASCollectionView#layoutDelegate` property. The idea here is the inspector object gives insight to the data controller about the internal implementation of the layout so it can generate nodes correctly. If you check out `ASCollectionViewFlowLayoutInspector`, you can see the general heuristic implementation of an inspector for flow layouts. This is a good place to start for this patch.

appleguy:
@levi that sounds like a pretty awesome architectural approach!  Thanks for the details.  Can't wait to take a look.

levi:
Architectural decision for y'all:

I'm thinking of turning `_completedNodes` and `_editingNodes` into mutable dictionaries that store groups of nodes by a key. The default implementation of ASDataController will retain the same behavior that it does now, but write to a `ASRowNodesKey` key, (ie. `_completedNodes[ASRowNodesKey]`). The reason I'm thinking about this change is I need a place to store supplementary view nodes. Currently, I'm just duplicating a bunch of functionality that `ASDataController` already implements and I think a central store that subclasses can write to will provide flexibility moving forward—including header/footer section view support for a table view data controller subclass. With this change, I'm planning on exposing public versions of `_insertNodes:atIndexPaths:withAnimationOptions:` and its associated delete/section counterparts to support introspection into specific node groups. Something like: `insertNodes:ofKind:atIndexPaths:completion:` for subclasses to utilize on the backbone of our editing/completion transaction logic.

Thoughts?

appleguy:
@levi thanks for the continued work on this — really phenomenal!  My detailed review is coming later tonight.

Could you update the Carthage project for any new files you added?  It looks like the recently added test (a fourth build type on Travis) has started failing.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -342,18 +353,34 @@
...
  * https://developer.apple.com/library/ios/documentation/UIKit/Reference/UICollectionViewDelegateFlowLayout_protocol/index.html#//apple_ref/occ/intfm/UICollectionViewDelegateFlowLayout/collectionView:layout:insetForSectionAtIndex:
  *
  */
-- (UIEdgeInsets)collectionView:(UICollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout insetForSectionAtIndex:(NSInteger)section;
+- (UIEdgeInsets)collectionView:(ASCollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout insetForSectionAtIndex:(NSInteger)section;
+
+/**
+ * Asks the delegate for the size of the header in the specified section.
+ */
+- (CGSize)collectionView:(ASCollectionView *)collectionView layout:(UICollectionViewLayout *)collectionViewLayout referenceSizeForHeaderInSection:(NSInteger)section;
```
appleguy: If the header and footer are supplementary nodes, can we eliminate these methods and rely on calling _measure on the returned nodes?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -202,6 +204,17 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
...
+  // Register the default layout inspector delegate for flow layouts, custom layouts
+  // will need to roll their own ASCollectionViewLayoutInspecting implementation.
+  if ([layout asdk_isFlowLayout]) {
+    _flowLayoutInspector = [[ASCollectionViewFlowLayoutInspector alloc] init];
+    _flowLayoutInspector.layout = (UICollectionViewFlowLayout *)layout;
+    _flowLayoutInspector.collectionView = self;
+    _layoutDelegate = _flowLayoutInspector;
+  }
+  
+  _registeredSupplementaryKinds = [NSMutableArray array];
```
appleguy: Simple optimization, small impact but easy — lazy initialize this on first call of register.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -261,10 +274,6 @@ - (void)setAsyncDataSource:(id<ASCollectionViewDataSource>)asyncDataSource
     _asyncDataSourceImplementsConstrainedSizeForNode = NO;
   } else {
     _asyncDataSource = asyncDataSource;
-    // TODO: Support supplementary views with ASCollectionView.
-    if ([_asyncDataSource respondsToSelector:@selector(collectionView:viewForSupplementaryElementOfKind:atIndexPath:)]) {
-      ASDisplayNodeAssert(NO, @"ASCollectionView is planned to support supplementary views by September 2015.  You can work around this issue by using standard items.");
```
appleguy: I knew it was risky to call out September - but this is a high-value feature man, people are going to be stoked to use this!

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -405,6 +421,11 @@ - (void)moveItemAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)
   [_dataController moveRowAtIndexPath:indexPath toIndexPath:newIndexPath withAnimationOptions:kASCollectionViewAnimationNone];
 }
 
+- (NSString *)__reuseIdentifierForKind:(NSString *)kind
+{
+  return [NSString stringWithFormat:@"_ASCollectionSupplementaryView_%@", kind];
```
appleguy: Also small optimization, but it is close to 10x faster to do something like [@"_ASCollectionSupplementaryView_%@" stringByAppendingString:kind].

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -357,6 +366,13 @@ - (void)performBatchUpdates:(void (^)())updates completion:(void (^)(BOOL))compl
   [self performBatchAnimated:YES updates:updates completion:completion];
 }
 
+- (void)registerSupplementaryNodeOfKind:(NSString *)elementKind
+{
+  [_registeredSupplementaryKinds addObject:elementKind];
```
appleguy: Add an assert and runtime safety check for nil (and zero-length?  maybe that's valid?)

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -605,7 +635,7 @@ - (ASSizeRange)dataController:(ASDataController *)dataController constrainedSize
   }
   
   if (_asyncDelegateImplementsInsetSection) {
-    sectionInset = [_asyncDelegate collectionView:self layout:self.collectionViewLayout insetForSectionAtIndex:indexPath.section];
+    sectionInset = [(id<ASCollectionViewDelegateFlowLayout>)_asyncDelegate collectionView:self layout:self.collectionViewLayout insetForSectionAtIndex:indexPath.section];
```
appleguy: Missed any related stuff to this, why are we casting to FlowLayout here?  Presumably it is not always a flow layout, so is there different protocol naming / structure that would fix?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -630,7 +660,7 @@ - (NSUInteger)dataController:(ASDataController *)dataController rowsInSection:(N
   return [_asyncDataSource collectionView:self numberOfItemsInSection:section];
 }
 
-- (NSUInteger)dataControllerNumberOfSections:(ASDataController *)dataController {
+- (NSUInteger)numberOfSectionsInDataController:(ASDataController *)dataController {
```
appleguy: Nice, definitely like this.  The original API had a couple slightly off names from the original engineer :)

AsyncDisplayKit/ASTableView.mm
```diff
@@ -639,7 +637,7 @@ - (void)handleBatchFetchScrollingToOffset:(CGPoint)targetOffset
 - (void)rangeControllerBeginUpdates:(ASRangeController *)rangeController
 {
   ASDisplayNodeAssertMainThread();
-  LOG(@"--- UITableView beginUpdates");
+  ASLOG(@"--- UITableView beginUpdates");
```
appleguy: I think this is probably a good idea, but I'm worried about not being able to control logging on a per-file basis.  It will get too verbose pretty quickly.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -0,0 +1,91 @@
...
+#import "ASCollectionDataController.h"
+
+#import "ASLog.h"
+#import "ASAssert.h"
+#import "ASMultidimensionalArrayUtils.h"
+#import "ASDisplayNode.h"
+#import "ASDisplayNodeInternal.h"
+#import "ASDataController+Subclasses.h"
+
+@interface ASCollectionDataController ()
```
appleguy: This class seems quite well-factored - great work!

AsyncDisplayKit/Details/ASDataController+Subclasses.h
```diff
@@ -0,0 +1,53 @@
...
+ * An opportunity for a subclass to access the data source before entering into the editing queue
+ */
+- (void)prepareForReloadData;
+
+/**
+ * Subclasses can override this to reload data after the abstract data controller deletes its old data and before it reloads the new.
+ */
+- (void)willReloadData;
+
+- (NSArray *)completedNodesOfKind:(NSString *)kind;
```
appleguy: Stuff from this point on appears fairly specific to the Collection subclass - fine for now, maybe add a comment calling out the boundary line.  It will probably be a little while before we see another ASDataController subclass (or is @eanagel planning to do ASListNode for sure?)

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -52,9 +54,12 @@ - (instancetype)initWithAsyncDataFetching:(BOOL)asyncDataFetchingEnabled
     return nil;
   }
   
-  _completedNodes = [NSMutableArray array];
-  _editingNodes = [NSMutableArray array];
+  _completedNodes = [NSMutableDictionary dictionary];
```
appleguy: This is a pretty big change.  It's the only thing I've seen so far that definitely needs to be profiled.  We will almost surely land without changing it, but sometime in the future, we should consider moving to a C++ type or another solution faster than a large dictionary with main-thread-hot key access.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -112,7 +144,7 @@ - (void)_layoutNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAni
     //TODO: There should be a fast-path that avoids all of this object creation.
     NSMutableArray *nodeBoundSizes = [[NSMutableArray alloc] initWithCapacity:kASDataControllerSizingCountPerProcessor];
     [subIndexPaths enumerateObjectsUsingBlock:^(NSIndexPath *indexPath, NSUInteger idx, BOOL *stop) {
-      ASSizeRange constrainedSize = [_dataSource dataController:self constrainedSizeForNodeAtIndexPath:indexPath];
+      ASSizeRange constrainedSize = constraintedSizeBlock(indexPath);
```
appleguy: A little worried about this, wondering if there is a better way than churning through so many block objects.

AsyncDisplayKit/Details/ASDataController+Subclasses.h
levi: I was thinking after landing this patch adding an `ASTableDataController` subclass that has extra support for header/footer nodes.

AsyncDisplayKit/Details/ASDataController.mm
levi: Thanks for the feedback. I updated the diff to reflect a subclassing approach that I think will work much better with the addition of node kinds. @Adlai-Holler's addition this morning is a nice cherry on top of this whole thing.

appleguy:
@levi regarding your earlier post on the architectural consideration — I did a double-take at the new mutable dictionaries, but I think that and the other details of the architecture you spelled out are solid and a reasonable direction for us to take this.  Specifically because the nodes are still stored in inner arrays, not in the dictionary directly, I think the performance cost will be acceptable (storing the nodes directly would have a very measurable impact).

Can you write up some details on what you think would be necessary to enable supplementary nodes without a flow layout?  I say that having not finished reviewing the diff, but I think it is an important question — as we add complexity to these classes, I think it really proves its value if we can get this feature more broadly exposed, while also achieving the secondary goals of supporting further ASDataController customizations.  Even if the non-flow-layout support depended on new delegate methods (or methods that need to be called at certain times by external code), like some kind of invalidation signal, that's probably fine.

appleguy:
I am so freaking excited whenever I load the PR page and see the little "3 of 5" bar advance...lol!  Let me know how I can be of help here.

levi:
I'm pretty close to moving this out of WIP. I want to clean up a few things and write some unit tests. No. 4 in the list is a change that I think could be done in a separate diff, since we don't support any sort of layout invalidation handling in `ASCollectionView` currently it might be best to simply support the `reloadData` and `layoutSubviews` callbacks as of now and extend it out next.

@appleguy (re: your message yesterday) I'll find some time tonight to write out a better description of the architecture. To validate the extensibility of this implementation I think we should integrate a custom layout in an example and show how it can be done. Do you have one in mind that would be pretty straight forward?

AsyncDisplayKit/ASCollectionView.mm
levi: I want to move our reliance on cell size modification (like sectionInsets) to that of the layout. UICollectionView only exposes this delegate method on `UICollectionViewDelegateFlowLayout`, which essentially makes this functionality valuable only when working with flow layouts.

I could see the issue where a custom layout author provides a custom delegate that implements `collectionView:layout:insetForSectionAtIndex`, hoping that ASDK will see the implementation and use it for the internal section inset calculation. However, this could easily be supported by composing `ASCollectionViewDelegateFlowLayout` into the custom delegate protocol. It's a bit of an edge case, but if you're interested in supporting this pathway we could create a simple protocol with just this method definition that would be composed into `ASCollectionViewDelegateFlowLayout`.

AsyncDisplayKit/ASCollectionView.mm
levi: We just missed it ;)

AsyncDisplayKit/ASCollectionView.h
levi: This is an interesting idea. I was going for API-parity with UICollectionViewFlowLayout just to make this familiar with what people know, but I could see how inferring the size from the node itself  being a valuable result. I'd have to give this more thought, especially as I'm integrating this in our production app and trying out the API in a real-world implementation.

appleguy: Let me know as your thinking evolves on this.  In particular, I'd like to understand from your use case what the practical adoption considerations are for an app that is in fact using ASDK (or will be using ASDK) — does it provide any advantage to have these methods besides the familiarity?  Perhaps it does, if there is some kind of custom size logic that would otherwise have to be moved into a node subclass, but I am completely without experience using supplementary views and making nontrivial UICollectionViewLayout subclasses.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -202,6 +204,12 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
   // and should not trigger a relayout.
   _ignoreMaxSizeChange = CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, CGSizeZero);
   
+  // Register the default layout inspector delegate for flow layouts, custom layouts
+  // will need to roll their own ASCollectionViewLayoutInspecting implementation.
+  _layoutDelegate = [self flowLayoutInspector];
```
appleguy: So that we can do more final reviews, can you prioritize refactoring this to be how you'd suggest landing it?  Naturally this doesn't make much sense for collections created without a flow layout; we should have some mechanism to ensure this is only done for collections with a flow layout.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -291,6 +314,8 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
     super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
     _asyncDelegateImplementsInsetSection = ([_asyncDelegate respondsToSelector:@selector(collectionView:layout:insetForSectionAtIndex:)] ? 1 : 0);
   }
+
+  _flowLayoutInspector.collectionView = self;
```
appleguy: Shouldn't this just be done in -flowLayoutInspector?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -357,6 +382,14 @@ - (void)performBatchUpdates:(void (^)())updates completion:(void (^)(BOOL))compl
   [self performBatchAnimated:YES updates:updates completion:completion];
 }
 
+- (void)registerSupplementaryNodeOfKind:(NSString *)elementKind
+{
+  ASDisplayNodeAssert(elementKind != nil, @"A kind is needed for supplementary node registration");
+  [_registeredSupplementaryKinds addObject:elementKind];
+  [self registerClass:[UICollectionReusableView class] forSupplementaryViewOfKind:elementKind
```
appleguy: I could be wrong but I thought we had a custom subclass of UICollectionReusableView.  It may not actually do anything, although probably best to use it here.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -416,7 +454,7 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
 
   ASCellNode *node = [_dataController nodeAtIndexPath:indexPath];
   
-  [_rangeController configureContentView:cell.contentView forCellNode:node];
+  [_rangeController configureContentView:cell.contentView forNode:node];
```
appleguy: I'm a little worried about these new ones not being cells.  It means they'll miss out on the fairly useful / important property .neverShowPlaceholders, for one thing; it will also defeat the universality of an override point that ASCellNode currently provides for content in collection views.

The name is definitely not ideal (cell), but from a functionality perspective, would there be a downside to using cell nodes for the supplementary nodes?  Perhaps the best solution is to create a subclass of ASCellNode, for the primary (sole?) purpose of giving it a better name for this use case (from the developer-facing part of the API, so it is not confusing to people to see "cell" for supplementary elements that are perhaps smaller and secondary to real cells; clearly the naming is less critical for this internal stuff)

AsyncDisplayKit/ASCollectionView.mm
levi: Yeah, I was originally storing a reference of the collectionView in the inspector but it seems like I'm now just using this to reset the delegate property cache. Will clean this up so that it's just that.

AsyncDisplayKit/ASCollectionView.mm
levi: If you follow the trail of this the flow layout inspector returns nil if the layout is indeed not a flow layout. You bringing this up shows that this isn't super clear. Will clarify this with a conditional in the method body.

AsyncDisplayKit/ASCollectionView.mm
appleguy: I mostly understand, but not sure if completely understand.  This is still calling out to the _asyncDelegate; so are you saying that you plan to keep this as a delegate method, just generally deprecate / discourage its use and attempt to make its specificity to flow layouts more clear by creating a new sub-protocol?  That would make sense to me, if the protocol in question derives from ASCollectionViewDelegate.

The name doesn't really feel right because it doesn't end in Delegate, although I do see the logic behind it.  We should consider ASCollectionViewFlowLayoutDelegate but I trust your choice here.

AsyncDisplayKit/ASCollectionView.mm
levi: I don't see one for `UICollectionReusableView`, but you are correct there is a `_ASTableViewCell` subclass. It looks like it's being used to bubble relayout events out of individual cells. Great idea. Not sure how valuable it is to add an `_ASCollectionReusableView` as of right now. I would like to think about it a bit more post this diff, since a supplementary view's bounds is often defined by the layout outside of the node's scope.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -658,8 +705,40 @@ - (void)dataControllerUnlockDataSource
...
 
-#pragma mark -
-#pragma mark ASRangeControllerDelegate.
+#pragma mark - ASCollectionViewDataControllerSource Supplementary view support
+
+- (ASDisplayNode *)dataController:(ASCollectionDataController *)dataController supplementaryNodeOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath
+{
+  ASDisplayNode *node = [_asyncDataSource collectionView:self nodeForSupplementaryElementOfKind:kind atIndexPath:indexPath];
+  ASDisplayNodeAssert(node != nil, @"A node must be returned for a supplementary node");
+  ASDisplayNodeAssert(!node.nodeLoaded, @"The supplementary node must not be loaded");
```
appleguy: Is this assertion necessary?  I know that a loaded node would require main thread measurement (for now, before I can get to the thread affinity system), but ideally we could support it, in the event that the app has its own cache of nodes.  If there are a couple good reasons why this assertion must hold for now (e.g. nontrivial to support loaded nodes even in a future diff done by another contributor), go ahead and leave it.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -658,8 +705,40 @@ - (void)dataControllerUnlockDataSource
...
+
+- (ASSizeRange)dataController:(ASCollectionDataController *)dataController constrainedSizeForSupplementaryNodeOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath
+{
+  ASDisplayNodeAssert(_layoutDelegate != nil, @"ASCollectionView must have a layoutDelegate for layout inspection");
+  return [_layoutDelegate collectionView:self constrainedSizeForSupplementaryNodeOfKind:kind atIndexPath:indexPath];
+}
+
+- (NSUInteger)dataController:(ASCollectionDataController *)dataController supplementaryViewsOfKind:(NSString *)kind inSection:(NSUInteger)section
+{
+  ASDisplayNodeAssert(_layoutDelegate != nil, @"ASCollectionView must have a layoutDelegate for layout inspection");
```
appleguy: Improve all three assertion messages to clarify that the layout delegate is required specifically for usage of supplementary nodes

examples/ASCollectionView/Sample/SupplementaryNode.m
```diff
@@ -0,0 +1,54 @@
...
+    _textNode.attributedString = [[NSAttributedString alloc] initWithString:text
+                                                                 attributes:[self textAttributes]];
+    [self addSubnode:_textNode];
+  }
+  return self;
+}
+
+- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
+{
+  ASCenterLayoutSpec *center = [[ASCenterLayoutSpec alloc] init];
```
appleguy: I'm not actually certain, but is the Center spec of value / does it do anything in this use case where the Inset is also present?

AsyncDisplayKit/ASCollectionView.mm
levi: Beyond naming, there is the touches forwarding and selection style properties. How about we just use cells and no need to worry about introducing indirection with another name for a node.

AsyncDisplayKit/ASCollectionView.mm
levi: I added it simply because it took me some time to track down why a loaded node was causing data inconsistencies in my own app. Figuring here that calling it out ahead of time will save others debugging time in the future while we figure out how to handle loaded nodes in the data controller.

AsyncDisplayKit/ASCollectionView.mm
levi: Good catch

examples/ASCollectionView/Sample/SupplementaryNode.m
levi: This is something I still need to clean up. Still very new to the spec API and I'm pretty sure this doesn't work right now, but thanks for bringing it up. Will take a look at a right path.

AsyncDisplayKit/Details/ASDataController+Subclasses.h
```diff
@@ -0,0 +1,99 @@
...
+ * When `asyncDataFetching` is enabled, the block is executed on a background thread.
+ */
+- (void)accessDataSourceWithBlock:(dispatch_block_t)block;
+
+- (void)willPerformInitialDataLoading;
+
+/**
+ * An opportunity for a subclass to access the data source before entering into the editing queue
+ */
+- (void)prepareForReloadData;
```
appleguy: This seems vague - is it specific to reloadData, or any edit operation?  If any edit operation, it should be renamed.  If specific to reloadData, the comment seems too broad and should be edited to clarify it is only related to reload data.

AsyncDisplayKit/Details/ASDataController+Subclasses.h
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Subclasses can override this to reload data after the abstract data controller deletes its old data and before it reloads the new.
+ *
+ * @discussion Invoked on the editing transaction queue.
+ */
+- (void)willReloadData;
+
+- (void)prepareInsertSections:(NSIndexSet *)sections;
```
appleguy: How does this differ from willInsertSections?  Below it says prepareFORreloadSections, so should this be prepareForInsertSections?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -273,6 +377,23 @@ - (void)reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animati
   }];
 }
 
+#pragma mark - Initial & Reload Data Hooks (Subclass API)
+
+- (void)willPerformInitialDataLoading
+{
+  // Implemented by subclasses
```
appleguy: Maybe // Optional hook for subclasses (might be more clear that ASDataController is a concrete class, does not require subclassing & implementation of these methods)

appleguy:
@levi awesome body of work here.  I've done as complete a review as I can—realistically with this large of diff, both unit / integration testing and real-world app testing are going to be the only way to completely verify correctness here.

I suggest we land this sooner rather than later in order to be able to iterate on additional improvements with smaller diffs.  Additionally, there are other key changes needed in this code (both the ordering of execution of batch transactions and layout tweaks to ensure reused nodes are not asynchronously measured), so we need to clear the conflict risk soon.  What do you think we should target?

AsyncDisplayKit/ASCollectionView.mm
appleguy: Sounds good, leave it in!  @Adlai-Holler's diff just landed (which will create a small conflict with this one, should be easy to revolve) and should support loaded nodes now.

There are some great ways to improve the behavior of loaded nodes, to allow async measure and such.  Hope to get to those in the coming months.

AsyncDisplayKit/ASCollectionView.mm
appleguy: Agreed, let's not create one now.

AsyncDisplayKit/ASCollectionView.mm
appleguy: Yeah, I think we should refactor this to be = [self layoutInspector] (rename the method) at the least, even if it only ever returns nil or a flow layout inspector.  The comment should also be slightly tweaked to indicate this is the behavior, just making clear that it only has a default for flow layouts.

appleguy:
This is looking really great.  Thanks for all the reply comments.  Message me any time of day if you have ideas / questions to discuss or a new revision to review.

levi:
@appleguy: Still looking into why these tests aren't passing in travis. But in the meantime, I've discovered a strange issue while running ASViewController/Sample (1 out of 10 runs). Seems like it's accessing `completedNodes` via `ASDataController#nodesAtIndexPaths:` and failing to insert into the `deletedElements` collection. I've posted a screenshot below. @nguyenhuy has done some work on switching between a completedNodes copy to provide safe accessing at all time. I was wondering if this was related? I've not touched the range controller much in this patch and curious why it's coming up now.

asmultidimensionalarrayutils_mm

appleguy:
@levi looking great!  Thanks for updating your initial post with the new items included and detailed overview.  I think this is on the right track and should land as soon as your current investigation completes (and any other small items you feel are necessary).

As discussed, I'll roll 1.9 without this, although based on my current schedule we will probably end up landing this before I can post the release.  I'll just point at an earlier commit, so it is no trouble to do so.

nguyenhuy:
I took a look at the crash reported. It occurs when visible range is updated. I guess there is something wrong with the updated visible index paths. Do index paths of header and footer nodes included? If yes then there might be a problem because completedNodes setter returns [normal rows only](https://github.com/facebook/AsyncDisplayKit/pull/694/files#diff-c03b807fdd44839fd88691746d9dcc27R914), no?

I could reproduce the crash on this branch, but not master. I suspect it's a new bug, but nothing for sure right now. I will spend more time on it later on.

> @nguyenhuy has done some work on switching between a completedNodes copy to provide safe accessing at all time.

My previous patch allows ASDataController to switch to a completedNodes copy, but only during a transaction.

levi:
@nguyenhuy was able to provide a fix for this crash last night. Turns out the range controller was using `layoutAttributesForElementsInRect:` which returns all types of collection view items (cells, supplementary views, and decoration views). Check out `23134c4` for the fix, which simply filters out non-cell items when building the visible nodes list in the range controller. Were you experiencing this crash with that commit included? I haven't been able to reproduce it since.

nguyenhuy:
I probably did the test before that commit. Great to know about it. Will do another round of testing tomorrow.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -0,0 +1,217 @@
...
+- (id<ASCollectionDataControllerSource>)collectionDataSource;
+
+@end
+
+@implementation ASCollectionDataController {
+  NSMutableDictionary *_pendingNodes;
+  NSMutableDictionary *_pendingIndexPaths;
+}
+
+- (void)prepareForReloadData
```
Adlai-Holler: @levi Could you add some documentation around `prepareForX` vs `willReloadX` to help describe the architecture here?

Adlai-Holler:
@levi This feature is tremendous! I'm gonna start using it as soon as it lands. Thanks so much. :100:

AsyncDisplayKit/ASCollectionView.h
appleguy: I'm still skeptical about this.  There could be a small performance win from not having to create nodes, but I think the downside of the API surface area is greater — particularly because this is crossing concepts that developers may already understand (e.g. it is not normally on UICollectionView, but rather UICollectionViewFlowLayout).

I could fix the performance cost entirely by providing a +staticSize or similar method on ASCellNode, which, if it returns a non-CGSizeZero, would be used instead of allocating the node at that time.  The only bummer is that these methods will be tough to remove later.

I guess in the interest of time, the best plan is to just make a breaking API change in the future when they're removed.  

@levi - Is it already the case that if they are not implemented, the node IS instantiated and consulted for size?  Or does it not give any space for the header / footer in that case?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -153,6 +153,8 @@ @interface ASCollectionView () <ASRangeControllerDelegate, ASDataControllerSourc
   CGSize _maxSizeForNodesConstrainedSize;
   BOOL _ignoreMaxSizeChange;
   
+  NSMutableArray *_registeredSupplementaryKinds;
```
appleguy: @levi maybe this should be a set if duplicates aren't allowed?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -224,6 +226,14 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
   // and should not trigger a relayout.
   _ignoreMaxSizeChange = CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, CGSizeZero);
   
+  // Register the default layout inspector delegate for flow layouts only, custom layouts
+  // will need to roll their own ASCollectionViewLayoutInspecting implementation and set a layout delegate
+  if ([layout asdk_isFlowLayout]) {
+    _layoutDelegate = [self flowLayoutInspector];
```
appleguy: Can you evaluate the names _layoutDelegate vs _flowLayoutInspector?  Especially since they are the same object in the flow layout case, and because the _layoutDelegate is referred to as a layout inspector here, can we name them congruently?

How about _layoutInspector or _flowLayoutDelegate?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -313,6 +332,17 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
     super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
     _asyncDelegateImplementsInsetSection = ([_asyncDelegate respondsToSelector:@selector(collectionView:layout:insetForSectionAtIndex:)] ? 1 : 0);
   }
+
+  [_flowLayoutInspector cacheSelectorsForCollectionView:self];
```
appleguy: This is a little subtle for my tastes.  It would in fact be a programming error to call [self flowLayoutInspector] here, whereas I was about to recommend we do that to make sure it's allocated.

Is here no applicability of this call point for likely other layoutInspector implementations?  Can we name it something more generic and call it on _layoutDelegate instead?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -239,6 +249,19 @@ - (void)dealloc
   super.dataSource = nil;
 }
 
+/**
+ * A layout inspector implementation specific for the sizing behavior of UICollectionViewFlowLayouts
+ */
+- (ASCollectionViewFlowLayoutInspector *)flowLayoutInspector
+{
+    if (_flowLayoutInspector == nil) {
+        UICollectionViewFlowLayout *layout = (UICollectionViewFlowLayout *)self.collectionViewLayout;
```
appleguy: Please add an assertion checking the class type / that this cast is valid.

 (don't want runtime cost, do want immediate development safety-checking even internal to the framework)

AsyncDisplayKit/ASCollectionView.mm
appleguy: @levi in the meantime, an AS subclass of UICollectionReusableView has been created.  Please adopt it as it would be required to ensure setSelected / setHighlighted (added recently) are not broken.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -313,6 +332,17 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
...
+
+  [_flowLayoutInspector cacheSelectorsForCollectionView:self];
+}
+
+- (void)setCollectionViewLayout:(UICollectionViewLayout *)collectionViewLayout
+{
+  [super setCollectionViewLayout:collectionViewLayout];
+  if ([collectionViewLayout asdk_isFlowLayout]) {
+    _flowLayoutInspector = nil;
+    _layoutDelegate = [self flowLayoutInspector];
```
appleguy: I think what's not feeling right about this pattern is the fact that _flowLayoutInspector is set simply by calling the method...but you cannot call the method in all of the use cases that you need to access that object, as it will crash if you are not currently associated with a flow layout.

Please refactor -flowLayoutInspector to be something like -createInspectorForFlowLayout:()  (that's probably a bad name), explicitly passing in the flow layout and setting the _flowLayoutInspector property from the return value.  Then in this use case you will not need to nil out the property in advance.

Overall though I really feel like we need to consolidate or eliminate one of these properties.  It does not seem like there are many cases where _flowLayoutInspector is uniquely needed as opposed to calling _layoutDelegate, and having them be set to the same object makes that more unusual.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -313,6 +332,17 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
     super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
     _asyncDelegateImplementsInsetSection = ([_asyncDelegate respondsToSelector:@selector(collectionView:layout:insetForSectionAtIndex:)] ? 1 : 0);
   }
+
+  [_flowLayoutInspector cacheSelectorsForCollectionView:self];
+}
+
+- (void)setCollectionViewLayout:(UICollectionViewLayout *)collectionViewLayout
```
appleguy: Are you totally sure that this behavior will work?  Have you tested setting a different collection layout after the ASCollectionView is in full swing operation?  I'd say we shouldn't provide the method at all until it has been tested, for example switching the collection layout while the current collection is scrolling at high speed.  In this implementation I don't see a call to -reloadData or anything, and beyond that there probably needs to be an even more explicit "destroy all cached data" in the ASDataController or something...really not sure, just anxious :)

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -0,0 +1,217 @@
...
+{
+  _pendingNodes = [NSMutableDictionary dictionary];
+  _pendingIndexPaths = [NSMutableDictionary dictionary];
+
+  [[self supplementaryKinds] enumerateObjectsUsingBlock:^(NSString *kind, NSUInteger idx, BOOL *stop) {
+    LOG(@"Populating elements of kind: %@", kind);
+    NSMutableArray *indexPaths = [NSMutableArray array];
```
appleguy: Please make make the dictionaries and arrays created in this file use a capacity-aware initializer, e.g. arrayWithCapacity, wherever possible.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -0,0 +1,217 @@
...
+    
+    NSArray *editingNodes = [self editingNodesOfKind:kind];
+    NSMutableIndexSet *indexSet = [[NSMutableIndexSet alloc] initWithIndexesInRange:NSMakeRange(0, editingNodes.count)];
+    [self deleteSectionsOfKind:kind atIndexSet:indexSet completion:nil];
+
+    // Insert each section
+    NSUInteger sectionCount = [self.collectionDataSource dataController:self numberOfSectionsForSupplementaryNodeOfKind:kind];
+    NSMutableArray *sections = [NSMutableArray arrayWithCapacity:sectionCount];
+    for (int i = 0; i < sectionCount; i++) {
+      [sections addObject:[[NSMutableArray alloc] init]];
```
appleguy: Tiny nit, but might as well be [NSMutableArray array] as the rest of the file uses.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -0,0 +1,121 @@
...
+  return self;
+}
+
+- (void)cacheSelectorsForCollectionView:(ASCollectionView *)collectionView
+{
+  if (collectionView == nil) {
+    _delegateImplementsReferenceSizeForHeader = nil;
+    _delegateImplementsReferenceSizeForFooter = nil;
+  } else {
+    _delegateImplementsReferenceSizeForHeader = [[self delegateForCollectionView:collectionView] respondsToSelector:@selector(collectionView:layout:referenceSizeForHeaderInSection:)];
```
appleguy: Local variable for the delegateForCollectionView: call, particularly because the lines are long anyway.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -0,0 +1,121 @@
...
+  BOOL _delegateImplementsReferenceSizeForFooter;
+}
+
+#pragma mark - Accessors
+
+- (instancetype)initWithCollectionView:(ASCollectionView *)collectionView flowLayout:(UICollectionViewFlowLayout *)flowLayout;
+{
+  self = [super init];
+
+  if (flowLayout == nil) {
```
appleguy: Probably ASDisplayNodeAssert(NO, @"Should never create a layout inspector without a layout") ?  Great of course to have the return nil for safer production runtime behavior.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -0,0 +1,121 @@
...
+- (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSizeForSupplementaryNodeOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath
+{
+  CGSize constrainedSize;
+  CGSize supplementarySize = [self sizeForSupplementaryViewOfKind:kind inSection:indexPath.section collectionView:collectionView];
+  if (_layout.scrollDirection == UICollectionViewScrollDirectionVertical) {
+    constrainedSize = CGSizeMake(collectionView.bounds.size.width, supplementarySize.height);
+  } else {
+    constrainedSize = CGSizeMake(supplementarySize.height, collectionView.bounds.size.height);
+  }
+  return ASSizeRangeMake(CGSizeZero, constrainedSize);
```
appleguy: @levi - This is important...I think it may not be quite correct.  I believe this had to be changed for ASTableView recently.

@nguyenhuy, @rcancro — was the discovery / conclusion that we needed to provide a minimum size of the table width + 0.0 height in order for certain layout spec calculations to work correctly?  If so, we should probably do something similar here where depending on the scroll direction, the appropriate dimension of the minimum size is pegged.

Also important: I'm pretty sure that scrollDirection for collectionView can be BOTH vertical and horizontal at the same time.  Presumably in this case the minimum size of CGSizeZero is correct.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -0,0 +1,121 @@
...
+- (NSUInteger)collectionView:(ASCollectionView *)collectionView supplementaryNodesOfKind:(NSString *)kind inSection:(NSUInteger)section
+{
+  return [self layoutHasSupplementaryViewOfKind:kind inSection:section collectionView:collectionView] ? 1 : 0;
+}
+
+#pragma mark - Private helpers
+
+- (CGSize)sizeForSupplementaryViewOfKind:(NSString *)kind inSection:(NSUInteger)section collectionView:(ASCollectionView *)collectionView
+{
+  if ([kind isEqualToString:UICollectionElementKindSectionHeader]) {
```
appleguy: ak, kinda kills me to see an isEqualToString: in a method like this :).  I wish Apple designed this a bit better, even an enum would be nice!

I think the implementation of this method is the most notable and non-ASDK approach in the patch, and we should make sure to support cell/supplementary-specified sizing soon.

Adlai-Holler: Would using isEqual: offer a little performance bump by trying an identity comparison before a deep compare? IIRC that's how the isEqual:/isEqualToBanana: methods work.

AsyncDisplayKit/Details/ASDataController+Subclasses.h
```diff
@@ -0,0 +1,159 @@
...
+ * Deletes the given sections of the specified kind in the backing store, calling completion on the main thread when finished.
+ */
+- (void)deleteSectionsOfKind:(NSString *)kind atIndexSet:(NSIndexSet *)indexSet completion:(void (^)(NSIndexSet *indexSet))completionBlock;
+
+#pragma mark - Data Manipulation Hooks
+
+/**
+ * Notifies the subclass to perform any work needed before the data controller is reloaded entirely
+ *
+ * @discussion This method will be performed before the data controller enters its editing queue, usually on the main
```
appleguy: Is this /usually/ on main, or always?  If it is not a guarantee of the API, can we eliminate this method and have only -willReloadData?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -152,83 +184,147 @@ - (void)_layoutNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAni
   // Block the _editingTransactionQueue from executing a new edit transaction until layout is done & _editingNodes array is updated.
   dispatch_group_wait(layoutGroup, DISPATCH_TIME_FOREVER);
   free(nodeBoundSizes);
-  // Insert finished nodes into data storage
-  [self _insertNodes:nodes atIndexPaths:indexPaths withAnimationOptions:animationOptions];
+
+  if (completionBlock)
```
appleguy: I don't really like the style of mandatory braces for one-line control statements, but it is the style :(

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -607,41 +785,46 @@ - (void)reloadRowsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDat
...
-      }];
-    };
 
     // Can't relayout right away because _completedNodes may not be up-to-date,
     // i.e there might be some nodes that were measured using the old constrained size but haven't been added to _completedNodes
     // (see _layoutNodes:atIndexPaths:withAnimationOptions:).
     [_editingTransactionQueue addOperationWithBlock:^{
       ASDisplayNodePerformBlockOnMainThread(^{
-        relayoutNodesBlock(_completedNodes);
+        for (NSString *kind in [_completedNodes keyEnumerator]) {
```
appleguy: I think using -allKeys would be fractionally more efficient than keyEnumerator.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -607,41 +785,46 @@ - (void)reloadRowsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDat
...
+  if (!nodes.count) {
+    return;
+  }
+  
+  [self accessDataSourceWithBlock:^{
+    [nodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
+      [section enumerateObjectsUsingBlock:^(ASCellNode *node, NSUInteger rowIndex, BOOL *stop) {
+        NSIndexPath *indexPath = [NSIndexPath indexPathForRow:rowIndex inSection:sectionIndex];
+        ASSizeRange constrainedSize = [self constrainedSizeForNodeOfKind:kind atIndexPath:indexPath];
+        [node measureWithSizeRange:constrainedSize];
```
appleguy: Create a local variable out of the return value of this method, CGSize calculatedSize = [...].size, and use it below.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -662,6 +845,23 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
   }];
 }
 
+#pragma mark - Data Querying (Subclass API)
+
+- (NSArray *)indexPathsForEditingNodesOfKind:(NSString *)kind
+{
+  return _editingNodes[kind] != nil ? ASIndexPathsForMultidimensionalArray(_editingNodes[kind]) : [NSArray array];
```
appleguy: Do we need the [NSArray array]?  Using ASIndexPathsForMultidimensionalArray is pretty inefficient, so I am concerned we might want to eliminate this convenience or audit its usage carefully so the cost is not hidden.

AsyncDisplayKit/Details/ASDataController.mm
Adlai-Holler: Would you be willing to elaborate? Wouldn't allKeys require a potentially big heap allocation and copy?

examples/ASCollectionView/Sample/SupplementaryNode.m
appleguy: @levi - reminder, fix the Supplementary Node example code as applicable.

AsyncDisplayKit/Details/ASDataController.mm
appleguy: @Adlai-Holler you could be right!  As it turns out, there are only a very small number of keys in this dictionary anyway, but the thought I had is that NSFastEnumeration is designed to work directly with NSArray & other collections preferentially over enumerator objects.

Let's leave this code as-is, particularly because either option should not make a measurable difference.  Very plausible that for a large dictionary, the enumerator could save memory.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
appleguy: I'm not confident, but I would get behind using ASObjectIsEqual() for this (which does the pointer comparison and falls back to isEqual).

appleguy:
I'm hitting the button.  There are a number of followups to do here, but they are best handled in a different PR.

Congratulations, and sincere thanks @levi for taking great care to craft one of the largest contributions in ASDK's history.

AsyncDisplayKit/Details/ASDataController.mm
Adlai-Holler: Cool, thanks!

Adlai-Holler:
Absolutely awesome! @levi ! @levi ! @levi !

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
Adlai-Holler: That's what's up

levi:
Glad to see this get in. Will respond to your feedback, @appleguy this week in a separate diff.

AsyncDisplayKit/ASCollectionView.mm
levi: Great idea.

AsyncDisplayKit/Details/ASDataController+Subclasses.h
levi: Usually on main, as long as asyncDataFetching is false. The main priority here is to perform setup work before the data controller enters into the editing transaction queue. Will clarify this in this doc.

AsyncDisplayKit/Details/ASDataController.mm
levi: My impression of using an NSEnumerator over an `allObjects`, `allKeys` is the benefit of lazy loading through fast enumeration. Is this not so internally?

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
nguyenhuy: I think it is a (common?) requirement that cells' layout should fill the table view width. Also, failing to do so would cause cell separators to not extend to the edge of the table view. (#621)

I have no idea about how to do "something similar here", but will try :)

levi: Let me know if there should be tweaks to this value to better support layout specs. I was following the range defined for each cell in the original data controller implementation.

examples/ASCollectionView/Sample/SupplementaryNode.m
levi: Going to leave it as is, since the center spec allows the node to fill the entire constrained sized of the supplementary view. Otherwise, the node just wraps to the size of the text.

AsyncDisplayKit/ASCollectionView.mm
levi: The subclass you're mentioning was created for `UICollectionViewCell` in order to forward the selected and highlighted properties to `ASCellNode`. There isn't really any need to subclass `UICollectionReusableView` at the moment, since this view isn't selectable like `UICollectionViewCell`. Its main job is handling layout change events at the cell level, something that would likely be useful for re-measuring cells when needed. Not something I want to implement for the sake of it right now without a clear goal of how `invalidateLayout` events are being handled from the `ASCollectionView` itself. Definitely a conversation to have in a new issue.

PR #699, opened 2015-10-02
Static layout bug?

rcancro:
I noticed when creating a static layout that the sizeRange of the children was being ignored. The case was:

I had an image as a child of a static layout
The image was set to have an exact range of 90x90
When the static layout was measured, the constrained size came in as 375xInf (the width of the containing node and unbounded)
The static layout computed its final size as 375x90

According to the comments in component kit's header file the static layout  "[c]omputes a size that is the union of all childrens' frames." It appeared that we are only doing that in the unbounded direction. My fix is to do it in both directions.

@nguyenhuy Please take a look and let me know what you think.

### DIFF
--- a/AsyncDisplayKit/Layout/ASStaticLayoutSpec.mm
+++ b/AsyncDisplayKit/Layout/ASStaticLayoutSpec.mm
@@ -58,18 +58,14 @@
     [sublayouts addObject:sublayout];
   }
   
-  if (isnan(size.width) || size.width >= FLT_MAX - FLT_EPSILON) {
-    size.width = constrainedSize.min.width;
-    for (ASLayout *sublayout in sublayouts) {
-      size.width = MAX(size.width, sublayout.position.x + sublayout.size.width);
-    }
+  size.width = constrainedSize.min.width;
+  for (ASLayout *sublayout in sublayouts) {
+    size.width = MAX(size.width, sublayout.position.x + sublayout.size.width);
   }
 
-  if (isnan(size.height) || size.height >= FLT_MAX - FLT_EPSILON) {
-    size.height = constrainedSize.min.height;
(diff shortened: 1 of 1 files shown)


appleguy:
Thanks for finding this!  Definitely interested in Huy's thoughts as well.

nguyenhuy:
Thanks for bringing this up. I agree that this diff does make the spec works accordingly to its description. So it looks good to me.

Also, I am glad that the checks for unbounded directions are removed, because I think it's actually hard to be 100% sure whether a direction is bounded or not (--i.e how large should it be to be considered "unbounded"? FLT_MAX is unbounded, but how about FLT_MAX - 10? Any value near FLT_MAX is tricky because it's too large to be considered "bounded", and hard to detect otherwise. Ultimately we would need to define a threshold, but even such threshold is tricky itself).

This current wrap-content behaviour is good and should work for most use cases. In addition to that, we may add a flag that indicates how the spec should determine its size, either by wrapping its content or filling its parent (for each direction). Such flag already exists [in center spec](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASCenterLayoutSpec.h#L26), but let's do this once there is a demand.

appleguy:
@rcancro, @nguyenhuy eager to merge this — but could you take a half hour or so to devise a test that fails before the fix, and passes after?  Maybe a snapshot test that evaluates against the colored rectangle images?

nguyenhuy:
Not sure if I can push to `rcancro:staticLayout`branch. But I will write a bunch of tests for ASStaticLayoutSpec in the coming days.

appleguy:
Thanks a lot, @nguyenhuy!  I'll merge this now, and feel free to put a test into any future commit.  As long as you're tracking it, it isn't an urgent need, just didn't want it to be forgotten.

PR #704, opened 2015-10-04
Fixing dynamic framework target and adding an example for Travis

marmelroy:
The introduction of ASPhotosFrameworkImageRequest in ee0c027ba68e0299c2ccb803f84de2ed5bef0d9c broke the AsyncDisplayKit dynamic framework target. Again.

This is now fixed.

To prevent it from breaking again in the future, I created a new dynamic framework smoke test (based on the swift example). Travis will now build it and warn if there are any errors.


appleguy:
@marmelroy thanks man!  Here I missed the file add once again.  This is most definitely a job for a machine :).

Sorry it took so much extra effort to get running with Carthage in a stable way.  I'm very relieved to have this fixed once and for all.

PR #708, opened 2015-10-05
Faster Layout Nodes

Adlai-Holler:
@appleguy @nguyenhuy Been majorly frustrated at work lately so it feels super good to have a nice, clean improvement to offer.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -116,7 +116,7 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
     dispatch_group_async(layoutGroup, dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
       for (NSUInteger k = j; k < j + batchCount; k++) {
         ASCellNode *node = nodes[k];
-        ASSizeRange constrainedSize = nodeBoundSizes[k - j];
+        ASSizeRange constrainedSize = nodeBoundSizes[k];
         [node measureWithSizeRange:constrainedSize];
         node.frame = CGRectMake(0, 0, node.calculatedSize.width, node.calculatedSize.height);
       }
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
Also for sugar on top, I was hoping to use `QOS_CLASS_SOMETHING` to specify the global queue there. The docs for `dispatch_get_global_queue` say `You may specify the values QOS_CLASS_USER_INTERACTIVE, QOS_CLASS_USER_INITIATED, QOS_CLASS_UTILITY, or QOS_CLASS_BACKGROUND.`. This feels like `QOS_CLASS_UTILITY` to me but I want to know what y'all think

levi:
Awesome stuff, @Adlai-Holler!

Adlai-Holler:
Thanks man!

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -105,31 +105,27 @@ - (void)_layoutNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAni
...
-        ASCellNode *node = nodes[j + idx];
-        ASSizeRange constrainedSize;
-        [nodeBoundSizes[idx] getValue:&constrainedSize];
+      for (NSUInteger k = j; k < j + batchCount; k++) {
+        ASCellNode *node = nodes[k];
+        ASSizeRange constrainedSize = nodeBoundSizes[k];
         [node measureWithSizeRange:constrainedSize];
-        node.frame = CGRectMake(0.0f, 0.0f, node.calculatedSize.width, node.calculatedSize.height);
-      }];
+        node.frame = CGRectMake(0, 0, node.calculatedSize.width, node.calculatedSize.height);
```
appleguy: Create a local variable, CGSize calculatedSize = [node measureWithSizeRange:constrainedSize];

Adlai-Holler: Ah! Good point

appleguy:
Very nice, thank you @Adlai-Holler!  Some of @levi's changes are a possible performance regression in this code, but hopefully we can clean that up in the time before it lands — and having this as a higher bar will inspire us to do so :).

I would suggest we not touch queue priorities for now.  The dispatch headers actually talk about the risk of priority inversion and in at least one place recommend not using priorities unless absolutely necessary.  The main thread already runs at a higher priority, so there is no real risk of dropping frames from this background thread work.  More importantly, if we run at lower priority, it's likely that arbitrary developers' own background tasks (compressing analytics data to send over the network, JSON parsing, etc) could be taking CPU time away from layout, which the user may well be waiting on (at the end of the scrollable area, awaiting for new cells to finish layout and be added).

PR #710, opened 2015-10-05
Fix shouldRasterizeSubnodes skipping setting offset and clipping.

garrettmoon:


AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -131,8 +131,12 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
       CGContextTranslateCTM(context, frame.origin.x, frame.origin.y);
 
       //support cornerRadius
-      if (rasterizingFromAscendent && cornerRadius && clipsToBounds) {
-        [[UIBezierPath bezierPathWithRoundedRect:bounds cornerRadius:self.cornerRadius] addClip];
+      if (clipsToBounds) {
+        if (cornerRadius) {
+          [[UIBezierPath bezierPathWithRoundedRect:bounds cornerRadius:self.cornerRadius] addClip];
```
appleguy: Might as well use the local variable instead of self.cornerRadius

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -131,8 +131,12 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
       CGContextTranslateCTM(context, frame.origin.x, frame.origin.y);
 
       //support cornerRadius
-      if (rasterizingFromAscendent && cornerRadius && clipsToBounds) {
```
appleguy: Curious what rasterizingFromAscendent was here for, if we know?  It's definitely not needed?

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -117,8 +117,8 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
   // Get the display block for this node.
   asyncdisplaykit_async_transaction_operation_block_t displayBlock = [self _displayBlockWithAsynchronous:NO isCancelledBlock:isCancelledBlock rasterizing:YES];
 
-  // We'll display something if there is a display block and/or a background color.
-  BOOL shouldDisplay = displayBlock || backgroundColor;
+  // We'll display something if there is a display block, clipping, translation and/or a background color.
+  BOOL shouldDisplay = displayBlock || backgroundColor || CGPointEqualToPoint(CGPointZero, frame.origin) == NO || clipsToBounds;
```
appleguy: Thanks for fixing this!

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
garrettmoon: Indeed, it actually would cause that dang thread affinity thing if I use self, good catch!

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
garrettmoon: Didn't mean to take it out, I'll put it back in.

PR #711, opened 2015-10-05
Adds support for disabling and re-enabling should rasterize.

garrettmoon:


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -331,6 +331,40 @@ - (void)dealloc
...
+  _view = nil;
+  if (_flags.layerBacked)
+    _layer.delegate = nil;
+  [_layer removeFromSuperlayer];
+  _layer = nil;
+}
+
+- (void)__loadNode
+{
+  [self layer];
```
appleguy: This seems a little sparse, but smart :).  Multiple loads and unloads of a node is a first for the public ASDK, something we had some challenges implementing a while back when building prototypes of view reuse.  I assume each time will use layerBlock or viewBlock accordingly?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -653,6 +698,15 @@ - (void)displayImmediately
   [[self asyncLayer] displayImmediately];
 }
 
+- (void)recursivelyDisplayImmediately
```
appleguy: Check out -recursivelyEnsureDisplay; will it do what you need?  If so (I am not confident with this quick review), it would be notably faster.

appleguy:
This is looking good, should be able to merge soon...any ideas on testing this?  A couple unit tests would be warranted particularly for a change of this size.  We're tracking unit test coverage now, and with the big new methods I think that will drop unless we add a couple, aside from the legitimate reasons to have them.  Snapshot test may be appropriate using colored boxes.  Try enabling shouldRasterize on the layout tests, which are already running against images of colored boxes?

AsyncDisplayKit/ASDisplayNode.mm
garrettmoon: If I replace recursivelyDisplayImmediately with recursivelySetNeedsDisplay and recursivelyEnsureDisplay it doesn’t seem to complete synchronously, resulting in the UI flashing.

appleguy: Hmm, that's exceedingly odd.  I'm OK with landing this approach, but will consider it a followup for me to debug that issue and consolidate the display mechanisms.  Thank you for trying it.

AsyncDisplayKit/ASDisplayNode.mm
garrettmoon: It *appears* so.

appleguy:
This diff is definitely on the scarier side for me, but I really trust your work!

PR #715, opened 2015-10-06
More public API fast pathing

Adlai-Holler:
Couple more of these.


appleguy:
@Adlai-Holler love this ASObjectIsEqual() thing, encouraging use of the pointer equality fast path at scale! hehe.  Thank you.

PR #719, opened 2015-10-07
Fixes to baseline stack alignment

rcancro:
1) Set the ascender/descender of an ASTextNode when the attributeString is set. Previously ascender/descender were only being computed in `setValuesFromLayoutable` and only when the attribute string was not nil. May make sense to remove the computation from `setValuesFromLayoutable` entirely.
2) Remove ability to allow different children of a stack spec to aling to different baselines. This wasn't working before and I'm not convinced it is possible to do properly/useful enough to invest the time.
3) Have all stack spec run `ASStackBaselinePositionedLayout::compute` to compute the stack's ascender and descender. Even if the stack isn't aligning its children to a baseline, the stack itself may be a child of another stack that IS aligning to a baseline.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -17,6 +17,7 @@
 #import <AsyncDisplayKit/ASTextNodeTextKitHelpers.h>
 #import <AsyncDisplayKit/ASDisplayNodeExtras.h>
 
+#import "ASInternalHelpers.h"
 #import "ASTextNodeRenderer.h"
 #import "ASTextNodeShadower.h"
 #import "ASEqualityHelpers.h"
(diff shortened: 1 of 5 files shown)


appleguy:
@rcancro cool, solid reasoning.

1. Could you please add some tests for the baseline support that remains?  Not urgent this week, but it is pretty important.  If you invest in building a test case for text, we could probably put in a bunch of checks pretty easily & extend its coverage / functionality over time.  Snapshot test would make a lot of sense.

2. It's actually not clear to me, seeing the lines deleted, how I would enable baseline alignment for a given stack; is there still a per-item property in ASStackLayoutable?

rcancro:
1. Absolutely. I agree it is extremely important.
2. I deleted the per-item Self baseline options: ASStackLayoutAlignSelfBaselineFirst, ASStackLayoutAlignSelfBaselineLast. The stack spec ASStackLayoutAlignBaselineFirst and ASStackLayoutAlignBaselineLast still exists and can be set the stack spec's .alignItems property.

Issue #718, opened 2015-10-06
[ASLayoutSpec] ASNetworkImageNode with unknown height don't update ASCellNode's height


vitalybaev:
Hi! There is ASCellNode with vertical ASStackLayoutSpec with 3 children:
- ASTextNode - Title
- ASNetworkImageNode - Some internet image with various size
- ASTextNode - Some text

When image not yet downloaded I'd like to place gray placeholder (horizontal rectangle, like in Twitter app), but when image downloaded it had to be full size.

Here is test project https://github.com/vitalybaev/Test-ASDK of this

When image downloaded, ASCellNode's size doesn't changed. So if image height greater than placeholder's height - there is white gap, if less than placeholder - some content doesn't visible;

Got:
![screen shot 2015-10-06 at 20 36 53](https://cloud.githubusercontent.com/assets/724423/10316840/08a59c16-6c6a-11e5-9f29-0bb27759dbfb.png)
Expect 
![screen shot 2015-10-06 at 20 36 59](https://cloud.githubusercontent.com/assets/724423/10316861/1dd8ebc4-6c6a-11e5-879a-5c6ac4014a40.png)

appleguy:
Quick response — have you tried using preferredFrameSize?  Especially if you are using ASLayoutSpec to position everything, this will ensure the image is shown at the desired size even when the placeholder state (e.g. background color, or a default image you set for .placeholderImage) is shown.

appleguy:
@vitalybaev ah, so are you asking about animating the size of the cells?  I am very happy you provided images as it REALLY helps understand your use case.  However, I don't completely understand these because neither of them have an image placeholder state, and the first one also doesn't show the "some text at the bottom".  So I'm not sure what the desired behavior is, that you would like it to have, to get from the placeholder state to the final state.  Happy to help figure it out once I better understand the intended behavior!

Look at the examples/Kittens app for some examples of cell resizing and relayout.  When you tap on a cell, it does a relayout, which involves calling -invalidateCalculatedLayout and calling -reloadRowAtIndexPath:.

@nguyenhuy sorry to ask again, but could you remind me if there were any other method calls required here?  The ASTableView /SHOULD/ be able to detect that the cell has an invalidated layout, and requires a synchronous measure pass as part of the row reload, but does it do that right now?  Lastly, can you think of an API improvement we can make for this to be a one-line call?  Two ideas for that:
[...]

PR #722, opened 2015-10-07
Add relayout item/row APIs to ASTableView and ASCollectionView

nguyenhuy:
#718

### DIFF
--- a/AsyncDisplayKit/ASCellNode.m
+++ b/AsyncDisplayKit/ASCellNode.m
@@ -123,7 +123,6 @@ static const CGFloat kFontSize = 18.0f;
   _textNode.attributedString = [[NSAttributedString alloc] initWithString:_text
                                                                attributes:@{NSFontAttributeName: [UIFont systemFontOfSize:kFontSize]}];
 
-  [self invalidateCalculatedLayout];
 }
 
 @end
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -192,6 +192,16 @@
 - (void)reloadItemsAtIndexPaths:(NSArray *)indexPaths;
 
 /**
+ * Relayouts the specified item.
+ *
+ * @param indexPath The index path identifying the item to relayout.
+ *
+ * @discussion This method must be called from the main thread. The relayout is excuted on main thread.
+ * The node of the specified item must be updated to cause layout changes before this method is called.
+ */
+- (void)relayoutItemAtIndexPath:(NSIndexPath *)indexPath;
+
+/**
  * Moves the item at a specified location to a destination location.
  *
  * @param indexPath The index path identifying the item to move.
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -556,9 +556,13 @@ typedef void (^ASDisplayNodeDidLoadBlock)(ASDisplayNode *node);
  * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
  * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
  * 
- * Note: If the relayout causes a change in size of the root node that is attached to a container view
- * (table or collection view, for example), the container view must be notified to relayout.
- * For ASTableView and ASCollectionView, an empty batch editing transaction must be triggered to animate to new row / item sizes.
+ * Note: If the relayout causes a change in size of the root node that is attached to a container view, 
+ * the container view must be notified to relayout. 
+ * For ASTableView and ASCollectionView, instead of calling this method directly, 
+ * it is recommended to call -relayoutRowAtIndexPath:withRowAnimation and -relayoutItemAtIndexPath: respectively.
+ *
+ * @see [ASTableView relayoutRowAtIndexPath:withRowAnimation:]
+ * @see [ASCollectionView relayoutItemAtIndexPath:]
  */
 - (void)setNeedsLayout;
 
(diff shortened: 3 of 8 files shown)


AsyncDisplayKit/ASTableView.mm
```diff
@@ -455,6 +455,14 @@ - (void)reloadRowsAtIndexPaths:(NSArray *)indexPaths withRowAnimation:(UITableVi
   [_dataController reloadRowsAtIndexPaths:indexPaths withAnimationOptions:animation];
 }
 
+- (void)relayoutRowAtIndexPath:(NSIndexPath *)indexPath withRowAnimation:(UITableViewRowAnimation)animation
+{
+  ASDisplayNodeAssertMainThread();
+  ASCellNode *node = [self nodeForRowAtIndexPath:indexPath];
+  [node setNeedsLayout];
```
appleguy: I wonder if we should also call -layoutIfNeeded here?  Mainly thinking that most users will want to do animated expansions of cells, etc...if we force layout here then they could capture the relayout's actions in an animation block placed around the call to this method.  Perhaps there are other side effects of that technique that suggest we should avoid it?  Unless we can think of any, let's change it to go that route.  Sometime soon we should create an example app with an animated row or item height change & internal relayout.

appleguy: I guess this also brings up the question of how a developer could force display of the new destination positions.  If we do layoutIfNeeded, then they could call -displayImmediately on the applicable nodes right afterwards; otherwise they would have to rely on toggling isSynchronous, which is cumbersome since it doesn't reset itself after the next display...

nguyenhuy: -layoutIfNeeded is triggered by UITableView's -reloadRowsAtIndexPath:withRowAnimation (which is the next line). In the case of collection view, -layoutIfNeeded is triggered by -reloadItemsAtIndexPath. Do you think we still should call it anyway, or above assumptions are safe to make? Also, since -layoutIfNeeded is actually triggered with the current implementation, users can indeed capture the relayout's actions in an animation block and it should work.

This PR also adds animation to Kitten node's height change. I'm gonna push a commit that animates Kitten node's internal relayout as well.

AsyncDisplayKit/ASCellNode.m
```diff
@@ -123,7 +123,6 @@ - (void)setText:(NSString *)text
   _textNode.attributedString = [[NSAttributedString alloc] initWithString:_text
                                                                attributes:@{NSFontAttributeName: [UIFont systemFontOfSize:kFontSize]}];
 
-  [self invalidateCalculatedLayout];
```
nguyenhuy: -invalidateCalculatedLayout is removed here because calling -invalidateCalculatedLayout here removes the current constrained size and therefore any -setNeedsLayout calls in the future won't have a valid constrained size to proceed. Instead, cell nodes should be relaid-out using the new APIs introduced in ASTableView and ASCollectionView. 

The commit that makes this removal (91f3ba1) has the same explanation, in case people do a git blame later on :)

nguyenhuy: Added animation for Kitten node's internal relayout (image and text nodes swapped) in 210a89e.

AsyncDisplayKit/ASTableView.mm
nguyenhuy: Added animation for Kitten node's internal relayout (image and text nodes swapped) in 210a89e.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -192,6 +192,16 @@
 - (void)reloadItemsAtIndexPaths:(NSArray *)indexPaths;
 
 /**
+ * Relayouts the specified item.
+ *
+ * @param indexPath The index path identifying the item to relayout.
+ *
+ * @discussion This method must be called from the main thread. The relayout is excuted on main thread.
```
appleguy: You should probably mention here that it does perform a re-measurement, with the same constrained size (the width or height of the view in the direction orthogonal to scrolling, plus CGFLOAT_MAX).

If you could also explain whether this does work when called inside of an animation block (to animate all of the changes), I think many people reading the comments will be trying to do that.  Really clean implementation!

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -556,9 +556,13 @@ typedef void (^ASDisplayNodeDidLoadBlock)(ASDisplayNode *node);
  * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
  * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
  * 
- * Note: If the relayout causes a change in size of the root node that is attached to a container view
- * (table or collection view, for example), the container view must be notified to relayout.
- * For ASTableView and ASCollectionView, an empty batch editing transaction must be triggered to animate to new row / item sizes.
+ * Note: If the relayout causes a change in size of the root node that is attached to a container view, 
+ * the container view must be notified to relayout. 
+ * For ASTableView and ASCollectionView, instead of calling this method directly,
```
appleguy: may want to say "for ASCellNodes being used with ASTable ...".

AsyncDisplayKit/ASTableView.mm
appleguy: Thanks for adding that animation, @nguyenhuy!  I think it is safe to rely on these behaviors you described (very helpful for me to understand).  The main action would be to update the header documentation to make an API guarantee that we intend the reloadRow/Item methods to be callable within an animation block.

I'm going to land this as-is, so take your time and do a small followup when convenient.

Issue #700, opened 2015-10-03
ASLayoutSpecs should use bounds and position to layout as opposed to frame if transform is non-identity


garrettmoon:
Currently, ASLayoutSpecs use frame to layout layers and views, however if laying out a layer with a non-identity transform, setting the frame isn't valid (and in fact asserts in ASDisplayNode+UIViewBridging), instead bounds and position should be set.

Any chance you'd be interesting in looking at this @nguyenhuy?

nguyenhuy:
Well, firstly, just to be 100% precise, ASLayoutSpecs don't layout backing stores (--i.e layers/views). If you are talking about laying out subnodes, it's actually done by the parent node ([here](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASDisplayNode.mm#L1640)). 
But yeah, frame is used while laying out subnodes. And it has been like that since the beginning. I don't have the knowledge of why it was decided that way in the first place (perhaps @appleguy or someone else does?). That's said, I agree with you that setting frame isn't safe in some cases. And since frame is just a computed value (calculated from position, bounds and anchorPoint), I think we can instead use bounds and position.

Do you have a diff ready to go? I can give it a try if you don't have time for this :)

appleguy:
Right, makes sense that we are just computing ASLayout objects and then it is actually an ASDisplayNode task to set the values on subnodes.

Huy, if you have time to draft a patch, it would be appreciated--we haven't attempted yet.  I have a few other things higher on my ASDK list.  I'm hoping this change is simple because it should have essentially that one override point, and not require changing each layout spec.

nguyenhuy:
Yes, the change should be pretty simple. I will work on this in the next couple of days.

PR #723, opened 2015-10-07
Use bounds and position to layout subnodes

nguyenhuy:
#700

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -744,6 +744,26 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   }
 }
 
+- (void)__setSafeFrame:(CGRect)rect
+{
+  BOOL useLayer = (_layer && ASDisplayNodeThreadIsMain());
+  
+  CGPoint origin      = (useLayer ? _layer.bounds.origin : self.bounds.origin);
+  CGPoint anchorPoint = (useLayer ? _layer.anchorPoint   : self.anchorPoint);
+  
+  CGRect bounds       = (CGRect){ origin, rect.size };
+  CGPoint position    = CGPointMake(rect.origin.x + rect.size.width * anchorPoint.x,
+                                    rect.origin.y + rect.size.height * anchorPoint.y);
+  
+  if (useLayer) {
+    _layer.bounds = bounds;
+    _layer.position = position;
+  } else {
+    self.bounds = bounds;
+    self.position = position;
+  }
+}
+
 // These private methods ensure that subclasses are not required to call super in order for _renderingSubnodes to be properly managed.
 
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1691,10 +1711,10 @@ - (void)layout
   // Assume that _layout was flattened and is 1-level deep.
   for (ASLayout *subnodeLayout in _layout.sublayouts) {
     ASDisplayNodeAssert([_subnodes containsObject:subnodeLayout.layoutableObject], @"Cached sublayouts must only contain subnodes' layout.");
-    ((ASDisplayNode *)subnodeLayout.layoutableObject).frame = CGRectMake(subnodeLayout.position.x,
-                                                                         subnodeLayout.position.y,
-                                                                         subnodeLayout.size.width,
-                                                                         subnodeLayout.size.height);
+    [((ASDisplayNode *)subnodeLayout.layoutableObject) __setSafeFrame:CGRectMake(subnodeLayout.position.x,
+                                                                             subnodeLayout.position.y,
```
appleguy: More indentation required to align subnodeLayout vertically

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -143,22 +143,7 @@ - (void)setFrame:(CGRect)rect
   ASDisplayNodeAssert(CATransform3DIsIdentity(self.transform), @"-[ASDisplayNode setFrame:] - self.transform must be identity in order to set the frame property.  (From Apple's UIView documentation: If the transform property is not the identity transform, the value of this property is undefined and therefore should be ignored.)");
```
appleguy: Looks great!  Thanks for keeping the assertion here — this is not technically the behavior of setFrame: for UIView so I don't want to enable people to call setFrame: with a transform set.

nguyenhuy:
I was about to fix the indentation :(

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -744,6 +744,26 @@ - (void)__setNeedsLayout
   }
 }
 
+- (void)__setSafeFrame:(CGRect)rect
+{
+  BOOL useLayer = (_layer && ASDisplayNodeThreadIsMain());
```
garrettmoon: Should this lock the property lock? Otherwise should it be documented that it must always be called within the lock?

garrettmoon:
@nguyenhuy Thanks so much for doing this!

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: Indeed. Thanks!

nguyenhuy:
You're welcome :)

PR #725, opened 2015-10-07
Clean up #723

nguyenhuy:
#723


appleguy:
@nguyenhuy thanks!  Please also address @garrettmoon's comment - we should take the property lock at the beginning of the method.  It's recursive, so it should be safe to do.

nguyenhuy:
Done :)

PR #727, opened 2015-10-08
Fix regression on the nil delegate #706

RuiAAPeres:
As discussed here #724


RuiAAPeres:
The cell sizing is working (with a nil delegate) along side with the `table:didSelectRowAtIndexPath:` (with a non-nil delegate).

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -253,10 +259,10 @@ - (void)setAsyncDataSource:(id<ASCollectionViewDataSource>)asyncDataSource
   // the (common) case of nilling the asyncDataSource in the ViewController's dealloc. In this case our _asyncDataSource
   // will return as nil (ARC magic) even though the _proxyDataSource still exists. It's really important to nil out
   // super.dataSource in this case because calls to _ASTableViewProxy will start failing and cause crashes.
+  _asyncDataSource = nil;
```
appleguy: Is this change actually necessary, given that it will be replaced / overwritten below anyway?  If it is necessary, could you add a comment that explains why?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -278,10 +284,11 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
   // will return as nil (ARC magic) even though the _proxyDelegate still exists. It's really important to nil out
   // super.delegate in this case because calls to _ASTableViewProxy will start failing and cause crashes.
 
+  // order is important here, the delegate must be callable while nilling super.delegate to avoid random crashes
+  // in UIScrollViewAccessibility.
+  super.delegate = nil;
```
appleguy: Same question here...

appleguy:
@RuiAAPeres Thank you for looking at this!  I would love to land this but am not yet comfortable as I don't understand the significance of one of the core changes here, moving the = nil outside the conditionals.  Once I understand that, and if applicable (e.g. it is subtle reasoning and critical) it is commented, we'll get it in.

appleguy:
@RuiAAPeres ping - I'm going to branch for the next release soon (looks like we'll call it 1.9 — representing confidence in improvements for all 1.x surface area, but some fluidity remaining for 2.x APIs).  This is not /essential/ for the branch, but nice to have.

RuiAAPeres:
@appleguy sorry for the delay. About the changes, they allowed me to be able to fix both the cell height's issue (when no delegate is assigned), enable `table:didSelectRowAtIndexPath:` and make the tests pass. I am unsure how/why setting `_asyncDataSource = nil;` before assigning it to a proper delegate makes it work, but the reality is that it does (I am suspecting there is a bug in UIKit). On my current project I am using [my own branch](https://github.com/RuiAAPeres/AsyncDisplayKit/commit/34b982ada2bb09a16a25da73054021fa79cc0a24), and things are working as expected. 

I do understand, that you need a more convincing proof, but this week, I don't have time to create a project that:

1) Is able to reproduce [this crash](https://github.com/facebook/AsyncDisplayKit/pull/724#issuecomment-146320395) without any of my changes and then with those changes makes it not crash.
2) With a nil delegate, cells will have the proper size defined (according to `calculateSizeThatFits`).
3) With a non-nil delegate,  `table:didSelectRowAtIndexPath:` will work.
4) The tests pass.

PR #734, opened 2015-10-10
Don't Set Cell Node Frames During Measure

Adlai-Holler:
Small change, but in a delicate part of the framework.

This behavior confused me when I first dug into the framework, and now that I'm currently adding another point where we measure nodes, I think it's worth it to separate out what's essential to this measurement pass rather than duplicate the frame-setting.


Adlai-Holler:
Hmm. On second thought, it's better that this frame-setting occurs in the background if possible. Never mind!

PR #735, opened 2015-10-10
Measure cell nodes on main

### LINKED ISSUE #707: Intermittent Crash in ASDataController
I've got an ASCollectionView in a cell of an ASTableView, and sometimes when that collection view loads its initial data I get an assertion failure in ASDataController. The data controller attempts to access `ASDisplayNode.calculatedSize` on the default-priority dispatch queue so it fails the node thread affinity assertion.

I _believe_ this is the chain of events:
- UICollectionView calls reloadData to load initial data
- That calls ASDataController.reloadDataWithAnimationOptions:completion: on the main thread
- Which dispatches `_layoutNodes:atIndexPaths:withAnimationOptions:` on the editing transaction queue
- Which listens for all the constrained sizes on the default-priority global queue and then updates the frame there. 

Shouldn't that last `dispatch_group_async` probably target the main queue? Can we ever safely measure and `setFrame:` all these cell nodes in the background?

screen shot 2015-10-05 at 11 05 28 am
screen shot 2015-10-05 at 11 05 55 am
screen shot 2015-10-05 at 11 05 50 am

Adlai-Holler:
This fixes #707. The only possible risk I see is if a node's view is loaded after the main thread layout pass but before the editing transaction queue layout pass. That shouldn't happen unless the user abuses the cell node.


Adlai-Holler:
Hmm looks like the test host is failing to connect.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -245,6 +267,8 @@ - (void)reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animati
       NSMutableArray *updatedIndexPaths = [NSMutableArray array];
       [self _populateFromEntireDataSourceWithMutableNodes:updatedNodes mutableIndexPaths:updatedIndexPaths];
       
+      [self _layoutNodesWithMainThreadAffinity:updatedNodes atIndexPaths:updatedIndexPaths];
```
appleguy: I thought about this for a while, but can't see what requires it.  Wouldn't the insertion / reload of nodes trigger a -layoutSubviews pass where the cell layout code will run?  That should happen on the same runloop, so I think theoretically the main thread cells will be handled before the next transaction flush occurs — so I'm not confident why we need these.

Could you add a comment above these calls indicating their purpose?  This file is full of comments, but I think they are important since the cody is tricky.

appleguy:
@Adlai-Holler this is a great improvement, and I am going to land it now anyway, but would def. appreciate understanding the need for the other lines.  That's the most important thing, but secondarily if you can drop in some comments in a future diff it would be nice :)

AsyncDisplayKit/Details/ASDataController.mm
Adlai-Holler: We can finish syncing up on this in the new #736 but in short, I think the problem here is naming. The methods are named `_layoutNodes:` but what we're really doing is measuring them, so that we will be able to return the correct heightForRowAtIndexPath: or sizeForItemAtIndexPath:. Before this PR, we always measured nodes in the background before we reported them to the UIKit class. But if the node's view is already loaded when data controller gets told about it then it's unsafe to remeasure it in the background so we do it before we leave the main thread.

PR #736, opened 2015-10-11
Document Cell Layout Passes

Adlai-Holler:
Pursuant to #735, here's some more documentation! I added some questions in `FIXME` comments. 

I also replaced two needless calls to `enumerateObjects:` with NSFastEnumeration instead. I reckon every bit of performance counts!


appleguy:
Per your comments in the other diff, I'd also support renaming the methods to be more -measure-centric, but the comments added here make it pretty clear.

appleguy:
It looks like the FIXMEs here are squarely within the scope of the batch processing fixes that we need.  I'm not sure who will be able to champion that work, but I think a couple folks were considering it.

PR #738, opened 2015-10-14
Demote Photos.framework image request QoS

Adlai-Holler:
I found that sometimes iOS was inferring `UserInteractive` for these, which would degrade UIKit performance.

### DIFF
--- a/AsyncDisplayKit/ASMultiplexImageNode.mm
+++ b/AsyncDisplayKit/ASMultiplexImageNode.mm
@@ -606,6 +606,10 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
       }
     }];
   }];
+  if (AS_AT_LEAST_IOS8) {
+    // If you don't set this, iOS will sometimes infer NSQualityOfServiceUserInteractive and promote the entire queue to that level, damaging system responsiveness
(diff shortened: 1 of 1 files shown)


appleguy:
@Adlai-Holler interesting.  Do you have a sense of when the system sets it to UserInteractive?  Perhaps an example case you found where it was doing so, and one where it wasn't?  This could have implications for other parts of the framework.

Adlai-Holler:
I didn't take the time to find out when it happens one way and when the other – it was really late and I was trying to dial in a PR on a client project – but from what I've read in the documentation it looks at the QoS of where the operation was created.

So without knowing all the details of ASDK yet, I'm thinking if something running on a user-interactive queue/operation like `scrollViewDidScroll:` called into `ASRangeController` which called into `ASMultiplexImageNode.fetchData` and we create the operation there, the system might think something roughly like "Oh! This is descended from `scrollViewDidScroll:` which is UserInteractive." From the docs:

```
NSOperation has no QoS assigned:
The operation infers the QoS of the parent operation, queue,
[NSProcessInfo performActivityWithOptions:reason:usingBlock:]
block, or thread, if any. In a situation where an operation is created
on the main thread, a QoS of NSQualityOfServiceUserInitiated is inferred.
```

appleguy:
Cool, makes sense that it would mirror the QoS of the queue that launched it.  I think some of the most important queues won't be affected by this, but will audit them when I get a chance.

PR #745, opened 2015-10-15
ASCellNode selected/highlighted properties

vitalybaev:
Selected/highlighted properties on ASCellNode can be useful for doing some work when table view cell is selected or highlighted (e.g. change node's background color).

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -135,6 +135,15 @@ static BOOL _isInterceptedSelector(SEL sel)
   [super didTransitionToState:state];
 }
 
+- (void)setSelected:(BOOL)selected
+{
+  _node.selected = selected;
+}
+
+- (void)setHighlighted:(BOOL)highlighted {
+  _node.highlighted = highlighted;
+}
+
 @end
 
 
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASTableView.mm
```diff
@@ -135,6 +135,15 @@ - (void)didTransitionToState:(UITableViewCellStateMask)state
   [super didTransitionToState:state];
 }
 
+- (void)setSelected:(BOOL)selected
+{
+  _node.selected = selected;
+}
+
+- (void)setHighlighted:(BOOL)highlighted {
```
appleguy: Bump the { down to the next line for consistency.

appleguy:
@vitalybaev this is a nice extension, thank you!  If you are able to hook it up for ASCollectionView too — which also makes use of ASCellNode — that would be great, and allow me to land the diff.

vitalybaev:
@appleguy I'm sorry, but I'm not familiar with ASCollectionView. Quick look showing me, that we need to subclass UICollectionViewCell, but I don't want break smth.

vitalybaev:
@appleguy check my code please. Now it works with UICollectionView, but I'm not sure if it's right way. Thanks

appleguy:
@vitalybaev looks awesome, thank you!

PR #751, opened 2015-10-19
Ascii art visual debugging for ASLayoutSpecs

rcancro:
Created a class called `ASAsciiArtBoxCreator` that, given a parent name and an array of string representations of the children, will create an ascii representation of the parent. 

Objects that want to be rendered in ascii art should implement the `ASLayoutableAsciiArtProtocol` protocol. The two methods are:
```
- (NSString *)asciiArtString;
- (NSString *)asciiArtName;
```

I have updated `ASDisplayNode` and `ASLayoutSpec` so that they conform to this new protocol. If you are working on a layout, you can put a breakpoint at the end of `- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize` and type `po [layoutSpec asciiArtString]` to see a visual representation of your layout.

For example, this is what the layout for `KittenNode` looks like:
```
(lldb) po [insetSpec asciiArtString]
-------------ASInsetLayoutSpec------------
|  ----------ASStackLayoutSpec---------  |
|  |  ASNetworkImageNode  ASTextNode  |  |
|  ------------------------------------  |
------------------------------------------
```

Here is an example of a more complicated example:
```
---------------------ASStackLayoutSpec-----------------------
|               -----ASStackLayoutSpec------                | 
|               |  --ASInsetLayoutSpec--   |                | 
|               |  |     ASTextNode    |   |                | 
|               |  ---------------------   |                | 
|               |  --ASInsetLayoutSpec--   |                | 
|  ASImageNode  |  |     ASTextNode    |   |  ASButtonNode  | 
|               |  ---------------------   |                | 
|               |  --ASInsetLayoutSpec--|  |                | 
|               |  |    ASStackNode     |  |                | 
|               |  ----------------------  |                | 
|               ----------------------------                | 
------------------------------------------------------------- 
```

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2074,6 +2074,19 @@ static void _recursivelySetDisplaySuspended(ASDisplayNode *node, CALayer *layer,
   return self;
 }
 
+
+#pragma mark - ASLayoutableAsciiArtProtocol
+
+- (NSString *)asciiArtString
+{
+  return [ASLayoutSpec asciiArtStringForChildren:@[] parentName:[self asciiArtName]];
+}
+
+- (NSString *)asciiArtName
+{
+  return NSStringFromClass([self class]);
+}
+
 @end
 
 @implementation ASDisplayNode (Debugging)
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2074,19 +2074,6 @@ static void _recursivelySetDisplaySuspended(ASDisplayNode *node, CALayer *layer,
   return self;
 }
 
-
-#pragma mark - ASLayoutableAsciiArtProtocol
-
-- (NSString *)asciiArtString
-{
-  return [ASLayoutSpec asciiArtStringForChildren:@[] parentName:[self asciiArtName]];
-}
-
-- (NSString *)asciiArtName
-{
-  return NSStringFromClass([self class]);
-}
-
 @end
 
 @implementation ASDisplayNode (Debugging)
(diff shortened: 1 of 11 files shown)


nguyenhuy:
This is gold :+1: :+1: :+1:

SRandazzo:
omg. :+1:

wmcbain:
😍WOW 😍

Adlai-Holler:
yesjacknicholson

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2137,6 +2137,18 @@ - (NSString *)_recursiveDescriptionHelperWithIndent:(NSString *)indent
   return subtree;
 }
 
+#pragma mark - ASLayoutableAsciiArtProtocol
+
+- (NSString *)asciiArtString
+{
+    return [ASLayoutSpec asciiArtStringForChildren:@[] parentName:[self asciiArtName]];
```
appleguy: Just to make sure I understand — this only prints out one box with the node's name in it?  Or does it somehow print the hierarchy?

AsyncDisplayKit/Layout/ASAsciiArtBoxCreator.m
```diff
@@ -0,0 +1,185 @@
...
+  }
+  return [string copy];
+}
+
+- (NSString *)debugbox_stringByAddingPadding:(NSString *)padding count:(NSUInteger)count location:(PIDebugBoxPaddingLocation)location
+{
+  NSString *paddingString = [NSString debugbox_stringWithString:padding repeatedCount:count];
+  switch (location) {
+    case PIDebugBoxPaddingLocationFront:
+      return [NSString stringWithFormat:@"%@%@", paddingString, self];
```
appleguy: Just because it's an easy optimization & there can be large layouts to print from device debugging... consider using -stringByAppendingString:, as it's surprisingly much faster than the format initializer.

appleguy:
This is just so cool.  Thanks @rcancro for one of the most delightful patches ASDK has ever seen :).

If someone wants to write a few tests for this, it will keep our code coverage number up.  It should be easy to add one or two in an existing test suite, just printing the node tree or spec tree of some existing hierarchy in the tests.

AsyncDisplayKit/ASDisplayNode.mm
rcancro: This should just print the node's name. If someone wanted to make ascii art of the node hierarchy they'd pass in self.subnodes as the children.

PR #760, opened 2015-10-20
Call UIViewController's designated initializer in ASViewController

levi:
ASViewController currently cannot be subclassed in Swift, as the original call to `[super init]` directs to calling the designated initializer of UIViewController `[self initWithNibName:bundle:]`. Because of Swift's initializer inheritance behavior, this designated initializer will not implicitly be available on the subclassed ASViewController without some extra boilerplate. Adding an explicit call to the designated initializer fixes this issue.


appleguy:
Good call, thanks!

PR #762, opened 2015-10-22
Respond to supplementary node feedback

levi:
Went through @appleguy's post-merge review on #694 and cleaned up various aspects of the supplementary node implementation.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -90,7 +90,7 @@
  * collection view layout subclasses will need to provide their own implementation of an inspector object for their
  * supplementary views to be compatible with `ASCollectionView`'s supplementary node support.
  */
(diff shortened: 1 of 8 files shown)


appleguy:
OK, I don't really know the correct workflow here for fixing a conflict for the author.  I've tried a bunch of different things over time :).  Please teach me if someone knows.

In this case I was able to do it, but "publishing" the rebase (even for the directly-checked-out PR) just created a new PR.  Either way I'm going to try to land that one.

The conflict was from another PR that also set the =nil to =NO, which I hadn't realized this diff also did, because I reviewed it second.

appleguy:
the other PR is https://github.com/facebook/AsyncDisplayKit/pull/771

appleguy:
Haha, ironically I had to revert the diff that conflicted with this one because it broke tests, and that caused my alternate PR to be the one that won't apply cleanly!  Now we can land this one :).

Thanks for these refinements, I really like them!

Issue #761, opened 2015-10-21
Way to synchronously update ASDataController


zintus:
It would be nice to block main thread in case of very first data load, so we can run animations on initial ViewController's content.

Are there any ways to implement this on client code side, or library changes are required? If got this right, core of changes are adding `async` property on ASDataController and then conditional scheduling of tasks to `_editingTransactionQueue`.

levi:
@zintus this will require some updates internally to ASDataController such that anything pushed to a background queue (ie. node batch loading and editing transaction queue work) is done on main. While this is indeed a possible feature we could add to the data controller, I would caution against this approach if you're dealing with a lot of cells. The up front rendering cost could potentially block main for a longer period than desired and result in poor initial performance.

appleguy:
@zintus funny that you bring this up, because @garrettmoon just encountered a use case for this earlier today :).  This does make it a priority for the project overall.

The most important thing is to come up with the right API for doing this.  I think in many use cases, what is desired is a single method call that can do a one-shot synchronous load (akin to a -reloadDataImmediately), although a persistent mode that enables always-synchronous behavior is something to think about (I think there are more risks / considerations with this...we'd probably want to force on neverShowPlaceholders for all cell nodes in such a case).

appleguy:
The actual implementation change here shouldn't be terribly challenging.  The core change is to make sure the layout work for the cells, which is indeed done on the editing transaction queue, is handled synchronously.

Important: I suggest that we /block on the completion of that work/ rather than actually move it to the main thread.  This will allow multiple "chunks" of cell layout to be run concurrently, even though main is synchronously waiting on them, and will vastly reduce the amount of time spent waiting on the layout / height calculations.

zintus:
Totally agree on having only one synchronous method to do all job related to rendering visible portion of table (taking huge tables in consideration, as pointed by @levi).

Here is how such method could look like:
1) query nodes batch of predefined size on main thread;
2) do batch layout on them, blocking main thread;
3) repeat 1,2 if batch is too small to fill the table;
4) proceed as usual, relying on `neverShowPlaceholders` to render nodes content.

Step 1 implies support for partially updated internal state for table's data, which maybe non-trivial to implement.

Maybe it's worth thinking of this process as of pipeline with heavily parallel stages: query data source for nodes, layout nodes, render nodes. We can easily block main thread, waiting for any stage to complete, similarly to current implementation of batch layout.

appleguy:
[...]

PR #765, opened 2015-10-23
Add synchronous reloadData methods to ASTableView and ASCollectionView

levi:
This is a start on the concepts discussed in #761. The underlying data controller has been modified to allow synchronous construction of the underlying nodes on the main thread on `reloadDataImmediately`, while batch layout of the results still benefit from background thread parallelization.

The current implementation loads all the cells in the data controller's data source, so not fit for large amounts of data. #761 discusses the idea of loading a given number of cells, possibly based upon a provided bounds—like screen size. Investigating this next.


levi:
Hrm, not sure why Huy's commit is in here. Likely bad rebase.

levi:
Gonna open another one.

Issue #758, opened 2015-10-20
Periodic test failure: check for number of relayouts is sometimes 0 instead of 1.


appleguy:
This is a better problem to have than multiple relayouts...but it's unusual and a bit concerning that the failure seems to happen only on some runs.

@nguyenhuy here is one of the most recent failures: https://travis-ci.org/facebook/AsyncDisplayKit/jobs/86258557

Failures:
  0) -[ASTableViewTests testRelayoutAllRowsWithNonZeroSizeInitially] (AsyncDisplayKitTests.xctest)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━���━━━
/Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKitTests/ASTableViewTests.m:283: ((node.numberOfLayoutsOnMainThread) equal to (1)) failed: ("0") is not equal to ("1"):
280       for (int row = 0; row < NumberOfRowsPerSection; row++) {
281         NSIndexPath *indexPath = [NSIndexPath indexPathForRow:row inSection:section];
282         ASTestTextCellNode *node = (ASTestTextCellNode *)[tableView nodeForRowAtIndexPath:indexPath];
283         XCTAssertEqual(node.numberOfLayoutsOnMainThread, 1);
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
284         XCTAssertEqual(node.constrainedSizeForCalculatedLayout.max.width, newSize.width);
285       }
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  1) -[ASTableViewTests testRelayoutAllRowsWithZeroSizeInitially] (AsyncDisplayKitTests.xctest)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
/Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKitTests/ASTableViewTests.m:283: ((node.numberOfLayoutsOnMainThread) equal to (1)) failed: ("0") is not equal to ("1"):
280       for (int row = 0; row < NumberOfRowsPerSection; row++) {
281         NSIndexPath *indexPath = [NSIndexPath indexPathForRow:row inSection:section];
[...]

PR #753, opened 2015-10-19
Sort edit commands during batch updating of table and collection views

nguyenhuy:
#684, #693.

@levi, @Adlai-Holler: Please also let me know your thoughts on this PR. Thanks.

- [x] Implement ASChangetSetDataController 
- [x] Failing ASTableViewTests (#758)
- [x] Write documentation for ASChangetSetDataController
- [x] Remove duplicate steps while coalescing changes in _ASHierarchyChangeSet
- [ ] Write tests for ASChangetSetDataController

PR #770, opened 2015-10-25
Fix table view tests

nguyenhuy:
There is a bug in ASTableView that causes unnecessary relayout when the table view width is zero initially. This PR includes the fix for this bug and some minor refactorings.

#758, #753


appleguy:
Awesome, thank you!

PR #773, opened 2015-10-26
Revert "Make ASDisplayNode.name Thread Safe"

appleguy:
Unfortunately this is causing tests to fail on master, 100% of the time.  It's a bit of a weird issue and sadly I don't have time to debug it...sorry for reverting the entire stack!


appleguy:
@Adlai-Holler could you look at the test failure in your original PR and make any changes necessary to ensure tests pass?  @nguyenhuy already fixed the flaky table view test so we should be in a reliable test state now...with the exception that tests can only be run under Xcode 6.4, due to a tooling bug that Apple has acknowledged.

Adlai-Holler:
Huh– sorry for the bum PR man! I'm downloading Xcode 6.4 and I'll investigate in the next couple days.

nguyenhuy:
I debugged the failures on master and figured out the root cause right before falling asleep at around 2AM LOL. Gonna submit a PR to revert this revert and fix the bug.

Adlai-Holler:
Yes!! That rocks, you rock, thank you so much @nguyenhuy

Issue #782, opened 2015-10-28
CGColor leaks in _ASPendingState


eanagel:
If you take a look at _ASPendingState we correctly release/retain CGColors in our setters but we fail to release them on dealloc (in fact there is no dealloc.) The end result is that any CGColor set in an _ASPendingState will live on forever.

@appleguy We should probably hit this before release. I won't have time to actually take this on myself for a couple days yet.

appleguy:
@eanagel hmm, this must be an ARC bug I guess?  Do we take the CGColor out of a UIColor in order to standardize on CGColor for layer-backing?

I'm not super familiar with how to deal with these edge cases under ARC (antiquated MRR guy here...), so I'll let this hang out until there is more time to investigate.  The retain cycle with ASBasicImageDownloader is a far more serious memory issue anyway, for any users of that component.

eanagel:
No problem, the fix is easy (just release the items on dealloc.) I have that integrated into my branch, I'll see if I can push it out your way easily.

eanagel:
Fixed.

PR #786, opened 2015-10-28
release any CGColor's set while the node is in a pending state

eanagel:
fix for #782


levi:
Thanks for this, @eanagel. I've been seeing these leaks in instruments recently. Nice optimization!

PR #789, opened 2015-10-28
Avoid overwriting areAnimationsEnabled if app changes it again during animation

### LINKED ISSUE #788: animationsEnabled not getting set back to yes?
We are seeing an issue where after playing with the app for a while animations aren't working properly and someone suggested `[UIView animationsEnabled:NO]` not getting set back to YES. I've searched our project and have found that the only place where that happens is in the ASCollectionView and ASTableView.

Not sure if this is actually the case, but am going to start investigating. Could it be possible that these aren't getting set back properly?

Adlai-Holler:
We already had this `ASPerformBlockWithoutAnimation` convenience function, so I elevated it from `ASTableView.mm -> ASInternalHelpers.h` and reimplemented it using `performWithoutAnimation`. Now we never call `setAnimationsEnabled:` which should resolve #788

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -789,26 +789,14 @@ static BOOL _isInterceptedSelector(SEL sel)
     }
     return; // if the asyncDataSource has become invalid while we are processing, ignore this request to avoid crashes
   }
-
-  BOOL animationsEnabled = NO;
-
-  if (!animated) {
-    animationsEnabled = [UIView areAnimationsEnabled];
-    [UIView setAnimationsEnabled:NO];
-  }
-
-  [super performBatchUpdates:^{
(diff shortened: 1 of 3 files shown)


adcooley:
:+1: fixes my issue of losing animations when loading ASCollectionViews inside of ASTableViews

appleguy:
Why are all the tests failing?! arg.

levi:
This patch fixes our animation issues, also.

appleguy:
Thanks @Adlai-Holler :).  Is there any equivalent usage of this needed in ASTableView, or do the differences in the batch API mean it isn't necessary?

Adlai-Holler:
Of course @appleguy! I think all our bases are covered – the ASTableView change came in the form of switching our `ASPerformBlockWithoutAnimation()` function away from `setAnimationsEnabled:`. We no longer use `setAnimationsEnabled:` at all so we should be all good :fist:

PR #790, opened 2015-10-29
Delegate nil [WIP]

RuiAAPeres:


RuiAAPeres:
@appleguy I tested this with a `nil` delegate an a non-nil delegate. The `height` behaves as expected. The `didSelectRowAtIndexPath` is also being called as expected for a non-nil delegate. 

I need to figure out what's going with the tests yet.

cc @levi

levi:
@RuiAAPeres looks like the interceptor is nil for those few failing tests. Might have to do with how the tests themselves are setup. They're relatively new and postdate this patch.

RuiAAPeres:
I am yet unsure about the the scenario where you are switching delegates: `nil`, `aDelegate1`, `nil`,`aDelegate2`, etc. I had such case recently and I was bitten.

RuiAAPeres:
So I did a quick test and this scenario: `nil`, `aDelegate1`, `nil`, `aDelegate2`, etc is not working right now.

RuiAAPeres:
@appleguy @levi you can find a test project [here](https://dl.dropboxusercontent.com/u/14102938/DelegateNil.zip), which describes the scenario I described above. I am build AsyncDisplayKit with this `075843addc6af1a6020481075ae6a120df93d227` commit.

appleguy:
@RuiAAPeres Thank you!  Let me know if you get stuck, and I'll try to pick up the investigation this weekend.  Would like to land this before the 1.9 release, but if I keep holding the release for each fix, then it will never go out :-P.

appleguy:
@RuiAAPeres Any idea why this is failing tests, even with re-run attempts?

RuiAAPeres:
@appleguy not yet. Haven't had the time to thoroughly test it.

appleguy:
@RuiAAPeres If you have time to investigate this before the weekend, that would be awesome!  If not, no worries — I will try to get it in over the next couple weekends.  I want to be sure it is in the 2.0 release, which I'm hoping to wrap up in the next 3-4 weeks.

RuiAAPeres:
I will try to check it on Saturday. Can't do it earlier. I have been using this on my build btw.

appleguy:
@RuiAAPeres checking in - would you be able to resolve the merge conflicts?  I'll adopt this PR and get it landed if you don't have cycles (totally understand, I know you're very busy with something awesome ;) ).  If you don't have time for a couple weeks, that is OK too, but I may close this PR until then for that case.

RuiAAPeres:
@appleguy sure, give me a sec!

RuiAAPeres:
@appleguy let me re-test with this merge.

RuiAAPeres:
The [test](https://dl.dropboxusercontent.com/u/14102938/new_TestDelegate.zip) project is still working after the merge.

appleguy:
@RuiAAPeres awesome, thanks!  It looks like there is a legit failure in the test suite, though?

2015-12-08 22:07:36.596 AsyncDisplayKitTestHost[1640:7586] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'interceptor must not be nil'

Issue #792, opened 2015-10-29
Can't add or remove nodes intuitively using new ASLayoutSpec


aaronschubert0:
Hi, I've been playing round with the new layout using the Kittens example project. The new layout api is great but coming from ComponentKit, something that has stump me immediately is the ability to easily add/remove nodes. I had assumed that I could just alter the array that is passed to `ASStackLayoutSpec` like so (KittenNode.mm):
```
  [stackSpec setChildren:!_swappedTextAndImage ? @[_imageNode, _textNode] : @[_textNode, _imageNode, _secondImageNode]];
```
This works in ComponentKit, but in ASDK the node is either not removed or added but the layout isn't recalculated properly. Are there other methods to achieve this or is this supposed to work?

nguyenhuy:
You probably needs to call `-setNeedsLayout` on the node to trigger a new layout pass. If your root node is a cell node and the new layout has a different size, you can simply call [`-relayoutRowAtIndexPath:withRowAnimation`](https://github.com/facebook/AsyncDisplayKit/blob/2c8626e1f6e99468d5fe2baa6489d4fba0ec0c0c/AsyncDisplayKit/ASTableView.mm#L480) on the table view (or `-relayoutItemAtIndexPath:` on the collection view).

aaronschubert0:
Thanks @nguyenhuy i'm using the `-toggleNodesSwap` method in `KittenNode.mm` that calls `-setNeedsLayout` like so:
```
- (void)toggleNodesSwap
{
  _swappedTextAndImage = !_swappedTextAndImage;
  [UIView animateWithDuration:0.5 delay:0 usingSpringWithDamping:0.85 initialSpringVelocity:0 options:UIViewAnimationOptionCurveEaseInOut animations:^{
    [self setNeedsLayout];
    [self.view layoutIfNeeded];
  } completion:^(BOOL finished) {}];
}
```
Would this not be enough to trigger the layout pass?

nguyenhuy:
It is enough. But if the new layout has a different size, the result won't be picked up and rendered until the containing table view/collection view updates. In that case you need to notify the table view, by making an empty batch update, or calling -relayoutItemAtIndexPath.

Also, how deep is your node tree? Are you adding/removing direct subnodes of the cell node? Have you put a breakpoint in -layoutSpecThatFits: and are sure that the method is called?

[...]

Issue #616, opened 2015-08-18
ASTableView, when reloadData to refresh the tableView, it causes splash screen


nannanIT:
I use ASTableView and ASTableViewCell, the cell contains texts, buttons, images(fetch from network, i have image cache). Everything is ok but when i reloadData to refresh the tableView(pull down to refresh
), it causes splash screen, it does't work smoothly, which is bad for the user experience. I have tried many ways to optimize it, but no one comes to be satisfied. Even though i can add some animations to cover the splash, it is not the best solution I think.
So, anyone else meet this problem and can anyone help me? thank you very much...

nguyenhuy:
As explained in #600, reloadData is expensive and you probably know that :) For a "pull to refresh", most of the time, the data set is not completely new, i.e some objects are inserted or removed, but the rest are untouched. Sometimes there is even no change at all (happens frequently to a FB addict like me :P).

That's said, assume that you are completely sure reloadData is what you want, there are some tricks that can be used to hide the splash screen effect. I guess the easiest one is to show a snapshot view of the entire table view on top of it. After reloading, you can just remove the snapshot view, with a nice fade out animation. The code looks like this:

```objective-c
- (void)reloadData
{
  UIView *snapshot = [_tableView snapshotViewAfterScreenUpdates:NO];
  [self.view insertSubview:snapshot aboveSubview:_tableView];
  
  [_tableView beginUpdates];
  [_tableView reloadData];
  [_tableView endUpdatesAnimated:NO completion:^(BOOL completed) {
    [UIView animateWithDuration:1 animations:^{
      snapshot.alpha = 0;
    } completion:^(BOOL finished) {
      [snapshot removeFromSuperview];
    }];
  }];
}
```

And the result can be found [here](https://www.dropbox.com/s/r8g455r4493nnv5/ASDK-reloadData.mov?dl=0).

ryanfitz:
Instead of calling `reloadData`, I would insert, remove and move rows around as needed. For example when a user pulls to refresh, simply add the new rows at the top and move the existing rows down. I have a working [solution](https://github.com/ryanfitz/AsyncDisplayKitHeaderNodeExample/blob/master/AsyncDisplayKitHeaderNodeExample/ASTableViewManager.swift). This works well in my app, which supports both pull to refresh and infinite scroll.

With that said, @nguyenhuy solution is extremely clever

nguyenhuy:
[...]

PR #793, opened 2015-10-29
Improve cell node relayout

nguyenhuy:
As discussed in #792:
- Introduce ASCellNodeDelegate. Cell node automatically notifies the delegate after a relayout (caused by -setNeedsLayout). Confirming to ASCellNodeDelegate; ASTableView and ASCollectionView reload the calling cell upon notifications. These views automatically set themselves as delegate of every node.
- The result is that ASCellNode subclasses don't need to manually notify the containing view. Thus, `-relayoutItemAtIndexPath` and `-relayoutRowAtIndexPath` are removed (#722, #718).
- Kittens example is updated to reflect the change.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -8,6 +8,24 @@
 
 #import <AsyncDisplayKit/ASDisplayNode.h>
 
+@class ASCellNode;
+
+typedef NSUInteger ASCellNodeAnimation;
+
+@protocol ASCellNodeDelegate <NSObject>
+
+/**
+ * Notifies the delegate that the specified cell node has done a relayout that results in a new size. 
+ * The notification is done on main thread.
+ *
+ * @param node A node informing the delegate about the relayout.
+ *
+ * @param newSize A new size that is resulted in the relayout.
+ *
+ * @param suggestedAnimation A constant that indicates how the delegate should animate. See UITableViewRowAnimation.
+ */
+- (void)node:(ASCellNode *)node didRelayoutToNewSize:(CGSize)newSize suggestedAnimation:(ASCellNodeAnimation)animation;
+@end
 
 /**
  * Generic cell node.  Subclass this instead of `ASDisplayNode` to use with `ASTableView` and `ASCollectionView`.
@@ -54,6 +72,18 @@
 @property (nonatomic, assign) BOOL highlighted;
 
 /*
+ * A delegate to be notified (on main thread) after a relayout that results in a new size.
+ */
+@property (nonatomic, weak) id<ASCellNodeDelegate> delegate;
+
+/*
+ * A constant that is passed to the delegate to indicate how a relayout is to be animated.
+ * 
+ * @see UITableViewRowAnimation
+ */
+@property (nonatomic, assign) ASCellNodeAnimation relayoutAnimation;
+
+/*
  * ASCellNode must forward touch events in order for UITableView and UICollectionView tap handling to work. Overriding
  * these methods (e.g. for highlighting) requires the super method be called.
  */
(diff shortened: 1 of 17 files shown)


AsyncDisplayKit/ASCellNode.m
```diff
@@ -27,6 +28,7 @@ - (instancetype)init
   // use UITableViewCell defaults
   _selectionStyle = UITableViewCellSelectionStyleDefault;
   self.clipsToBounds = YES;
+  _relayoutAnimation = UITableViewRowAnimationAutomatic;
```
levi: What's the purpose of having a `suggestedAnimation` param when this is always automatic?

nguyenhuy: It can be set to another value, by another object or a subclass in init.

Adlai-Holler:
@nguyenhuy Does this mean cells will flash when they resize? In client apps I specifically avoid calling `reloadRowsAtIndexPaths` in favor of `beginUpdates() / endUpdates()` because UITableView seems to flash when reloading rows with animation no matter what you pass.

nguyenhuy:
@Adlai-Holler: 

> Does this mean cells will flash when they resize?

No. "Reload" destroys the existing nodes and loads new ones. "Relayout" (as in this PR) means the existing node is reused and only its layout is updated. 

> In client apps I specifically avoid calling reloadRowsAtIndexPaths in favor of beginUpdates() / endUpdates() because UITableView seems to flash when reloading rows with animation no matter what you pass.

Yeah, that's because of the delete + insert behaviour. Though you can do some tricks to hide the flash like [this](https://github.com/facebook/AsyncDisplayKit/issues/616#issuecomment-133422166). I would recommend avoiding reload methods, as discussed [here](https://github.com/facebook/AsyncDisplayKit/issues/616#issuecomment-133422166).

AsyncDisplayKit/ASCellNode.h
```diff
@@ -62,6 +90,17 @@
 - (void)touchesEnded:(NSSet *)touches withEvent:(UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 - (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 
+/**
+ * Marks the node as needing layout. Convenience for use whether the view / layer is loaded or not. Safe to call from a background thread.
```
appleguy: It's only safe to call from the background if the node isn't loaded, right?  We should probably just remove that.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -907,4 +899,15 @@ - (void)rangeController:(ASRangeController *)rangeController didDeleteSectionsAt
...
 }
 
+#pragma mark - ASCellNodeDelegate
+
+- (void)node:(ASCellNode *)node didRelayoutWithSuggestedAnimation:(ASCellNodeAnimation)animation
+{
+  ASDisplayNodeAssertMainThread();
+  NSIndexPath *indexPath = [self indexPathForNode:node];
+  if (indexPath != nil) {
+    [super reloadItemsAtIndexPaths:@[indexPath]];
```
appleguy: I wonder if there's a way to coalesce these without limiting the ability to wrap in animation blocks.  Maybe it's not an issue, but do you see visually sequential loading / flashing if you set multiple cells as needing layout?  Even if there were a longer delay due to main thread measurement, it would be best for all of those to happen together.

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -559,14 +559,6 @@ typedef void (^ASDisplayNodeDidLoadBlock)(ASDisplayNode *node);
  * 
  * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
  * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
- * 
- * Note: If the relayout causes a change in size of the root node that is attached to a container view,
```
appleguy: I would replace this with something like "Note: ASCellNode has special behavior in that calling this method will automatically notify the containing ASTableView / ASCollectionView that the cell should be resized, if necessary."

appleguy:
Thanks for the naming and formatting improvements.  This is really a great change and reflects the development philosophy of ASDK, aiming for a smaller API surface that integrates functionality where it makes sense for the developer (e.g. ASCellNode itself bridging table and collection, layout being a concept that should extend to cells' height, etc)

AsyncDisplayKit/ASCellNode.h
nguyenhuy: Done in #812.

AsyncDisplayKit/ASCollectionView.mm
nguyenhuy: After the changes in #812, it's up to the developer the animate the layout changes in multiple cells, i.e they can wrap multiple `-setNeedsLayout` calls inside a single animation block, or don't animate at all.

AsyncDisplayKit/ASDisplayNode.h
nguyenhuy: Added in #812. Thanks.

PR #794, opened 2015-10-29
Make textContainerInset configurable for ASEditableTextNode placehold…

samhsiung:
Make textContainerInset configurable for ASEditableTextNode placeholder and typed textViews

Previously, it was only possible to configure the textContainerInset of the typed textView by accessing the textView property on didLoad. This would cause the placeholder and typed textViews to become desynced however, so in this commit we add the ability to configure both.

### DIFF
--- a/AsyncDisplayKit/ASViewController.h
+++ b/AsyncDisplayKit/ASViewController.h
@@ -20,4 +20,13 @@
 
 - (instancetype)initWithNode:(ASDisplayNode *)node;
 
+/**
+ * The constrained size used to measure the backing node.
+ *
+ * @discussion Defaults to providing a size range that uses the view controller view's bounds as
+ * both the min and max definitions. Override this method to provide a custom size range to the
+ * backing node.
+ */
+- (ASSizeRange)nodeConstrainedSize;
+
 @end
--- a/AsyncDisplayKit/ASViewController.m
+++ b/AsyncDisplayKit/ASViewController.m
@@ -40,9 +40,7 @@
 - (void)viewWillLayoutSubviews
 {
   [super viewWillLayoutSubviews];
-  CGSize viewSize = self.view.bounds.size;
-  ASSizeRange constrainedSize = ASSizeRangeMake(viewSize, viewSize);
-  [_node measureWithSizeRange:constrainedSize];
+  [_node measureWithSizeRange:[self nodeConstrainedSize]];
 }
 
 - (void)viewDidLayoutSubviews
@@ -59,6 +57,14 @@
   [super viewWillAppear:animated];
   _ensureDisplayed = YES;
   [_node recursivelyFetchData];
+}
+
+// MARK: - Layout Helpers
+
+- (ASSizeRange)nodeConstrainedSize
+{
+  CGSize viewSize = self.view.bounds.size;
+  return ASSizeRangeMake(viewSize, viewSize);
 }
 
 @end


AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -122,7 +123,7 @@ - (void)didLoad
       textView.backgroundColor = nil;
       textView.opaque = NO;
     }
-    textView.textContainerInset = UIEdgeInsetsZero;
+    textView.textContainerInset = self.textContainerInset;
```
levi: The value of the new property will only be recognized on first load. I would recommend extending `setTextContainerInset:` to support value changes after the view has loaded.

samhsiung: Nice catch, added setter logic and tested that it works

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -175,6 +176,15 @@ - (void)setBackgroundColor:(UIColor *)backgroundColor
   _placeholderTextKitComponents.textView.backgroundColor = backgroundColor;
 }
 
+- (void)setTextContainerInset:(UIEdgeInsets)textContainerInset
+{
+  ASDN::MutexLocker l(_textKitLock);
+
+  _textContainerInset = textContainerInset;
+  _textKitComponents.textView.textContainerInset = textContainerInset;
```
appleguy: Would accessing textView here cause the view to be created?  Naturally that would be bad...it may be best to wrap the last two lines in a check of [self isNodeLoaded] (I may have the exact property name wrong)

appleguy:
Thanks @samhsiung, this is a great extension to a lesser-used class.  It is awesome to have some attention applied to these classes because it really helps round out the framework and provide a more confident developer experience when they are needed.

AsyncDisplayKit/ASEditableTextNode.mm
samhsiung: Thanks for the review. Would that be the case if we were accessing `self.textView`. In this case we're accessing the `textView` property from `_textKitComponents` which seems to be generated on `self.textView` instead.

Also I just accidentally nuked my master branch so I'm going to try to reopen this :)

samhsiung:
Moved this PR here: https://github.com/facebook/AsyncDisplayKit/pull/802

JoSpalding:
@samhsiung, I need textContainerInset configurable for ASTextNode but I can barely read obj-c since I'm learning Swift. Would you kindly make this change for ASTextNode also?

samhsiung:
@jospalding have you tried using ASInsetLayoutSpec to wrap your text node to give it padding?

JoSpalding:
@samhsiung yes, ASInsetLayoutSpec partially solves my issue but my node has a backgroundColor and the inset does not respect it (not sure if this is expected). It adds the inset but the inset is transparent.
I did quite a lot of research and imho the most elegant way of doing what I need is using textContainerInset.
Thank you and please let me know if you have any other suggestion that I should try.

samhsiung:
@JoSpalding have you tried setting the background color of the parent node that has the inset instead of the ASTextNode?

JoSpalding:
@samhsiung I can't set a background for the parent because it is also the parent of other nodes.
This is my layoutSpecThatFits for the parent (it's an ASCellNode):

```swift
        return ASInsetLayoutSpec(
            insets: UIEdgeInsetsMake(10, 20, 30, 20),
            child: ASStackLayoutSpec(
                direction: .Vertical,
                spacing: 5,
                justifyContent: .Start,
                alignItems: .Start,
                children: [
                    channelNode,
                    titleNode,
                    ASRatioLayoutSpec(
                        ratio: 2/3,
                        child: imageNode
                    )
                ]
            )
        )
    }
```

And this is how it renders:

untitled

I need the channelNode to have the padding/inset. I could wrap it inside another ASDisplayNode but this looks like a workaround for me.

samhsiung:
For now you can subclass `ASTextNode` and override `- (id<ASLayoutable>)finalLayoutable` to add the inset. I could work on adding an `inset` property to `ASTextNode` that does the same:

- (id<ASLayoutable>)finalLayoutable 
{
    ASInsetLayoutSpec *finalLayoutable = [[ASInsetLayoutSpec alloc] init];
    finalLayoutable.insets = UIEdgeInsetsMake(....);
    finalLayoutable.isFinalLayoutable = YES;

    // isFinalLayoutable must be set before calling setChild:. setChild: will ask the child
    // for its finalLayoutable which will cause infinte recursion.
    [finalLayoutable setChild:self];
    return finalLayoutable;
}
```

samhsiung:
@JoSpalding could you create a separate issue for adding an `inset` property to `ASTextNode` for this so others in the community could add their thoughts?

JoSpalding:
@samhsiung nice, I will try to subclass it! I will open the issue and let you know! Thank your very much!

JoSpalding:
@samhsiung the `finalLayoutable` suggestion worked but does not ended up with what I was trying to achieve. Actually, the result was the same as wrapping the `ASTextNode` inside an `ASInsetLayoutSpec`, adding "margins" (as in CSS) to the `ASTextNode`.

I ended up with something relatively simple, but I'm still not sure if it's the best soluiton.
Please take a look at https://github.com/facebook/AsyncDisplayKit/issues/1138 and give your thoughts.

Thanks again!

PR #796, opened 2015-10-30
adds convenience methods to return a vertical and horizontal stack spec

samhsiung:
It felt cumbersome to repeatedly specify the direction of an ASStackLayoutSpec so I thought it'd be nice to provide these convenience classes. I also ran into cases where I named an ASStackLayoutSpec as a verticalStack or horizontalStack but forgot to set it's direction causing me to waste time debugging.


levi:
Might be a better alternative to add an extra initializer to define the direction or a class method on ASStackLayoutSpec.

samhsiung:
Good idea, added class methods instead.

appleguy:
I agree with @levi's suggestion, so thanks for jumping on that @samhsiung.  This is implemented cleanly — even if the stack is subclassed by a public client, using [self alloc] and instancetype will ensure correct behavior.  Cool!

Issue #797, opened 2015-10-30
Discussion - How to fix ASRangeController cell node leaks


eanagel:
I've been tracking leaking ASCellNodes in my application and, thanks to reading up on #791, I think I'm on the right track. 

The first issue seems pretty simple to fix - we don't always exit all our ranges in tear down. 

The second issue is deeper and is more of a structural problem. This has to do when we perform updates (inserts/deletes, etc). The ASRangeController is maintaining an array of index paths for each range (Render and Preload) and as the range of visible index paths change we update these arrays and call enter or exit for each underlying node. This all works fins until we insert or delete items. At that point our index paths become invalid. Currently we set a flag (_rangeIsValid = NO) and discard all the old index paths. The problem is that we don't exit these index paths, so it's possible we will either enter them twice or never exit them at all. Failing to exit nodes in the Rendered range will cause the cell nodes to be retained in memory forever (they will accumulate in the ASRangeHandlerRender's workingWindow.)

I could totally be wrong about this (maybe all the issues I'm seeing are entirely from the first issue), I'd love to get some additional eyes on this. I'm looking specifically at  [ASRangeController.mm:117](https://github.com/facebook/AsyncDisplayKit/blob/a1dee22a0959b81835706caef510ae62df26c30e/AsyncDisplayKit/Details/ASRangeController.mm#L117). Funny thing about that part of the code base (_rangeIsValid handling), I think I was the last one to touch it. @appleguy, @levi and @nguyenhuy, what do you think, am I right about this being a problem?

I've been considering a couple fixes for this:

1. exiting all ranges/nodes when we first have changes is probably safest but would likely cause other issues and have a significant performance cost.

2. Somehow tracking the nodes either instead of or in addition to the indexPaths for each range. This could allow us to handle this edge case, but we would have to be careful not to retain the cellNodes too long.

3. A hybrid approach. While everything is valid we work off of index paths the way we do today but we track nodes during changes. So it would look something like the following

 - When we first start making changes (begin updates), capture the nodes associated with all current ranges.
 - When we are done making changes, grab the current visible index paths and get the nodes for each. Compare the previous nodes with the new ones and make sure we exit/enter the deltas.

ERussel:
@eanagel 

Thank you for brief issue description. This exactly what i wanted to explain in original task, but unfortunately my English needs more practice :(. Sorry for this.

appleguy:
[...]

Issue #791, opened 2015-10-29
[ASRangeController] Some nodes may be orphaned after insertions and deletions


ERussel:
I can confirm that the issue is still present. For example, when i call reloadData for collectionView, invisible nodes would not be deallocated. As i understand from ASRangeHandlerRender code, invisible nodes are added to global working window to activate offscreen rendering, but i don't see any workaround to remove them from superview in DELETE_NODES call.

appleguy:
@ERussel Thanks for reporting this!  Just to confirm, you're not using ASNetworkImageNode or ASBasicImageDownloader, correct?

This should be an easy fix if the cause you highlighted is correct.  @levi, @nguyenhuy could this have been impacted by recent refactoring?  I would be pretty surprised if no one else had reported this if it is a longstanding issue, but it's certainly possible.

levi:
@appleguy indeed could be a regression from the recent changes. I don't have much time this week to take a look, Sunday at the earliest. @nguyenhuy if you're able to check this out earlier that would be awesome!

nguyenhuy:
I can take a look tomorrow :)

ERussel:
Thank you for attention to this issue.

In fact, it is very old. I can confirm it on 1.1.1 version. I made temporarily fix for my project

- (void)clearInvisibleNodes{
    UIWindow *window = [[ASRangeHandlerRender class] performSelector:NSSelectorFromString(@"workingWindow")];
    for (UIView *view in [window subviews]) {
        if ([view isKindOfClass:[_ASDisplayView class]]) {
            ASDisplayNode *node = [view performSelector:@selector(asyncdisplaykit_node)];
            if ([node isKindOfClass:[GiphyCollectionViewNode class]]) {
                if (node.isNodeLoaded) {
                    if (node.layerBacked) {
                        [node.layer removeFromSuperlayer];
                    } else {
                        [node.view removeFromSuperview];
                    }
                }
            }
        }
    }
}

I combine call to this method with reloadData. I checked with Profile Leak Instrument and it seems that invisible node release properly now.

P.S. I know about leak related to ASNetworkImage when used with ASBasicImageDownloader and i am confident that bugs are different.

appleguy:
Great clarifying information, much appreciate it.On Oct 29, 2015, at 10:36 AM, Ruslan <notifications@github.com> wrote:Thank you for attention to this issue.In fact, it is very old. I can confirm it on 1.1.1 version. I made temporarily fix for my project


[...]

PR #798, opened 2015-10-31
ensure all nodes are exited on dealloc

eanagel:
This is a short-term fix for #797 and #791 suggested by @appleguy. I'm using a tweaked build of 1.2.3 and this seems to be working well for me. It clears any ASCellNodes that were in the Render state when the ASRangeController is dealloced (ultimately.) It is still possible for ASCellNodes to "leak" temporarily while the ASRangeController is alive. The good news is that this code will make sure it is cleaned up when the parent ASTable/CollectionView is dealloced.

I haven't tested this on master though (I'm not really set up for that at the moment), can someone give it a spin and make sure it plays well with all the amazing code in 1.9+?


appleguy:
This is now superseded by #830

PR #801, opened 2015-10-31
Add "Push Another Copy" to ASViewController demo for memory testing.

appleguy:
So far unable to reproduce the reports of rows or items not being released.

@eanagel, @levi, @nguyenhuy, @ERussel — anyone plz help!  I'm more than happy to finish testing & refining the fix that @eanagel has put up, but multiple attempts have never reproduced the issue.  I've tried more with ASTableView than ASCollectionView, but in this example with ASCollectionView I still haven't seen the behavior.  Here's a memory trace showing about 5 copies of the controller being pushed, scrolled, and popped with no leaks or obvious latent memory shown.

screen shot 2015-10-31 at 4 44 49 pm

### DIFF
--- a/examples/ASCollectionView/Sample.xcodeproj/project.pbxproj
+++ b/examples/ASCollectionView/Sample.xcodeproj/project.pbxproj
@@ -7,8 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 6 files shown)


appleguy:
Tried adding a reload button as well.  Still can't get any leaks to show up, whether reloading a view that's scrolled down and then popping it off the stack, or reloading repeatedly while scrolling around...

nguyenhuy:
I basically tried similar things to reproduce the leaks. But after reading @eanagel comment in #797, I think it could be possible that the number of leaked nodes after *each* reload/delete is small, but after a while the total number is material. I'm thinking about writing some test cases to *assert* whether backing stores are correctly removed from the working window.

PR #802, opened 2015-10-31
Make textContainerInset configurable for ASEditableTextNode

samhsiung:
Previously, it was only possible to configure the textContainerInset of the typed textView by accessing the textView property on didLoad. This would cause the placeholder and typed textViews to become desynced however, so in this commit we add the ability to configure both.

### DIFF
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -78,7 +78,8 @@
   _textKitComponents.layoutManager.delegate = self;
   _wordKerner = [[ASTextNodeWordKerner alloc] init];
   _returnKeyType = UIReturnKeyDefault;
-
+  _textContainerInset = UIEdgeInsetsZero;
+  
   // Create the placeholder scaffolding.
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -175,6 +176,15 @@ - (void)setBackgroundColor:(UIColor *)backgroundColor
   _placeholderTextKitComponents.textView.backgroundColor = backgroundColor;
 }
 
+- (void)setTextContainerInset:(UIEdgeInsets)textContainerInset
+{
+  ASDN::MutexLocker l(_textKitLock);
+
+  _textContainerInset = textContainerInset;
+  _textKitComponents.textView.textContainerInset = textContainerInset;
```
appleguy: In the future I'd still prefer we be more explicit about this — a comment is fine if it is logically certain that an if (nodeLoaded) check is not needed.  As a future reader / author in this code, though, it isn't immediately obvious that these latter two lines are intended to be no-ops and should not trigger the views to be created.

appleguy:
Thanks again for the nice refinement!

PR #812, opened 2015-11-03
Improve cell node layout delegate

### LINKED ISSUE #808: Relayout Cell Node Causes Row to Be Reloaded
Post-#793, having a cell node with an editable text node inside, and resizing that node as the user types, causes the table view cell to get reloaded which dismisses the keyboard and causes the view to flash.

Is there any way I can update the layout of a cell node without the dismiss & flash? Maybe bypass the `[super reloadRowsAtIndexPaths:withRowAnimation:]` call that comes from `ASTableView` in the `ASCellNodeDelegate` method?

nguyenhuy:
- Instead of reloading, layout delegates (i.e ASTableView and ASCollectionView) trigger an empty update transaction after a cell relayout (fix #808).
- Since delegates don't animate layout changes anymore, revert the animation example in KittenNode.
- Only set a default layout delegate to ASCellNode if it is not provided by the async data source. This allow developers to set their own delegate.
- Address comments in #793.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -820,7 +820,9 @@ static BOOL _isInterceptedSelector(SEL sel)
 {
   ASCellNode *node = [_asyncDataSource tableView:self nodeForRowAtIndexPath:indexPath];
   ASDisplayNodeAssert([node isKindOfClass:ASCellNode.class], @"invalid node class, expected ASCellNode");
-  node.layoutDelegate = self;
+  if (node.layoutDelegate == nil) {
+    node.layoutDelegate = self;
+  }
   return node;
 }
 
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -895,15 +895,14 @@ static BOOL _isInterceptedSelector(SEL sel)
   }
 }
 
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/ASTableView.mm
```diff
@@ -820,7 +820,9 @@ - (ASCellNode *)dataController:(ASDataController *)dataController nodeAtIndexPat
 {
   ASCellNode *node = [_asyncDataSource tableView:self nodeForRowAtIndexPath:indexPath];
   ASDisplayNodeAssert([node isKindOfClass:ASCellNode.class], @"invalid node class, expected ASCellNode");
-  node.layoutDelegate = self;
+  if (node.layoutDelegate == nil) {
```
appleguy: @nguyenhuy for my understanding, why is the check necessary before setting?

appleguy:
Great patch!  Thank you for carefully watching community feedback for your changes and iterating on them — a real dedication to polish and proper API design in the period of time where we can easily make changes.  Very important and a great example to set.

AsyncDisplayKit/ASTableView.mm
nguyenhuy: I want to allow developers to set their own layout delegate (in -tableView:nodeForRowAtIndexPath). This checks make sure if developer indeed provides such delegate, it is not overridden here.

PR #816, opened 2015-11-05
Add a ASTableNode class, similar to ASCollectionNode

steveram:
Should I have another initializer that drops the asyncDataFetching arg?


steveram:
Should I have another initializer that drops the asyncDataFetching arg?

appleguy:
@steveram if you do a follow up I would make it purely init, maybe with style argument I guess.  Async fetching is off for the foreseeable future and we can add back the initializer later.  But even with views, allocating with a frame is an anti pattern, so we should rely on layout to set that and not have initializers with frame.

PR #818, opened 2015-11-05
Modify ASTableNode initializer

steveram:


appleguy:
Beautiful, thanks @steveram!

PR #821, opened 2015-11-06
Fix ASTableViewCell's node selected/highlighted property

vitalybaev:
When user highlight/select cell, ```- (void)setSelected:(BOOL)selected animated:(BOOL)animated``` and ```- (void)setHighlighted:(BOOL)highlighted animated:(BOOL)animated``` are fired instead of ```- (void)setSelected:(BOOL)selected``` and ```- (void)setHighlighted:(BOOL)highlighted```

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -136,12 +136,12 @@ static BOOL _isInterceptedSelector(SEL sel)
   [super didTransitionToState:state];
 }
 
-- (void)setSelected:(BOOL)selected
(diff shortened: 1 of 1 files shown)


appleguy:
@vitalybaev does this apply on all iOS versions back to 7, our minimum supported one?

BTW, most importantly...how does a developer disable the selected / highlighted state on the cell if they are programming an ASCellNode subclass?  Don't we need a back-pointer from the ASCellNode to the UITableViewCell / UICollectionViewItem?

appleguy:
@vitalybaev closing because this is included in #823

PR #823, opened 2015-11-06
Social app example

vitalybaev:


appleguy:
@vitalybaev so cool, thank you!!  This seems to have the same change as your other diff to fix setSelected / setHighlighted, so I'll either merge only this PR or you can remove it from this one.

vitalybaev:
@appleguy yes, I realized that after sending this PR, cause I'm not so familiar with git, you can merge only this PR

appleguy:
This is beautiful!  Super funny choice of filler content — but I definitely approve (@nguyenhuy, check it out).

I have some style nits, but let's address those in followups.  Probably the biggest one is that the indent width should be 2 spaces.  @vitalybaev if you want to do that change (with Xcode's re-indent feature, it should be fast), then the code will correctly "blame" to you after the change.  Otherwise I'm completely happy to do it, just don't want to take credit for all your code!

@rcancro @garrettmoon this should probably be the new go-to app for teaching people about ASDK, and we should probably look at it in more depth to see if we can clean it up for further enhanced readability (more comments, style consistency, etc).  We could expand it with different cell types that implement various common things with ASLayoutSpec so that folks who take a look and say "I want something like that!" can see the implementation.

nguyenhuy:
haha, love the content as well. Thanks for contributing this, @vitalybaev.

vitalybaev:
@appleguy @nguyenhuy thanks a lot! I have no time right now, cause I'm in Paris at dotGo conf, and I could make all code styling fixes when I'll arrive in Moscow at home

PR #828, opened 2015-11-08
[ASRangeController] Totally remove the "working window"

appleguy:
This commit is still undergoing development (permanently-installed NSLogs, etc) — but it does actually work with no failed display events scrolling around the Kittens app (including when images load from the network).


appleguy:
@eanagel @ERussel I'm happy to report that this does resolve the issue of some nodes being stranded in a deletion / insertion use case, as reproduced by @eanagel's virtualization demo.  Now that I can reproduce it, we'll go ahead and merge something close to #798 as a short-term fix, as this diff is at least a full weekend away from being possible to land (and even then, would be prudent to test for a couple weeks / roll into 2.0 rather than something like 1.9.1).

appleguy:
Historical context on this:
https://github.com/facebook/AsyncDisplayKit/issues/315
https://github.com/facebook/AsyncDisplayKit/pull/371
https://github.com/facebook/AsyncDisplayKit/pull/127

appleguy:
This is superseded by #905

Issue #819, opened 2015-11-06
Crash when Reading ASCII Art String for Static Layout Spec


Adlai-Holler:
@rcancro It looks like the string method does `@[ self.child ]` and the static layout spec returns nil for its child, so we get an exception in `NSArray`. Here's my code that breaks it:

```swift
	override func layoutSpecThatFits(constrainedSize: ASSizeRange) -> ASLayoutSpec {
		let content = ASInsetLayoutSpec(
			insets: insets,
			child: ASStackLayoutSpec(
				direction: .Horizontal,
				spacing: Style.spacing.horizontal,
				justifyContent: .Center,
				alignItems: .Center,
				children: [
					textNode,
					spinner
				]))
		// Fill width, set height
		content.sizeRange = ASRelativeSizeRangeMakeWithExactRelativeDimensions(
			ASRelativeDimensionMakeWithPercent(1),
			ASRelativeDimensionMakeWithPoints(height))
		
		let spec = ASStaticLayoutSpec(children: [content])
		NSLog("Placeholder layout: \(spec.asciiArtString())")
		return spec
	}
```

appleguy:
Haha, good find — thanks!  @samhsiung @rcancro @garrettmoon let me know if you can fix soon and I'll roll into the 1.9.1 release (hope to release one tomorrow, but we'll see if I make it).

PR #831, opened 2015-11-09
Fix crash in ASCII art creator for one-element children arrays.

appleguy:
The API lets people create things with the children array even if they only have one child.


appleguy:
@rcancro @nguyenhuy - should we make it so that accessing self.child checks if the self.children array is count == 1, and then actually return the object?  That would seem pretty reasonable to me...

appleguy:
This is a fix for #819.

rcancro:
Thanks for doing this, I completely forgot :/

PR #837, opened 2015-11-11
Fix selection inconsistency in ASTableView/ASCollectionView cell implementation

### LINKED ISSUE #807: ASCellNode setSelected: called multiple times while scrolling
Here is details: using AsyncDisplayKit 1.9

1. **first issue:**  when I select a cell, ```setSelected:``` of target cell gets called triple times. but I have tapped on the target cell only once!

2. **second issue:** when I deselect the target cell, by tapping again on it, cell gets deselected, but when I scroll I see it gets selected again and it called multiple times with value ```YES```!!!

you can see the issue with provided log in the sample project 
here is [sample project](https://dl.dropboxusercontent.com/u/45698608/ASCollectionView%20With%20Selection%20issue.zip)

levi:
Seems like some of the state wasn't being correctly passed through to the ASCellNode on cell population.

Fixes #807

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -115,13 +115,22 @@ static BOOL _isInterceptedSelector(SEL sel)
 
 @implementation _ASCollectionViewCell
 
+- (void)setNode:(ASCellNode *)node
+{
+  _node = node;
+  node.selected = self.selected;
+  node.highlighted = self.highlighted;
+}
+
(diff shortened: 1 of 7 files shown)


appleguy:
@levi this looks like an improvement.

Question: how does one cause an un-selection from the ASCellNode?  If I get setSelected: called on my ASCellNode instance, how would I trigger an un-selection?  Don't we need some kind of back-pointer to the cell view object?

levi:
Good question, @appleguy. Will take a look at this other case.

PR #838, opened 2015-11-11
Adds ASVisualEffectNode which is a wrapper around UIVisualEffectView

samhsiung:
Future plans could be to add child nodes to UIVisualEffectView's contentView and not it's subview which Apple recommends: 
```After you add the visual effect view to the view hierarchy, add any subviews to the contentView property of the visual effect view. Do not add subviews directly to the visual effect view itself.```
https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIVisualEffectView/

### CODE REVIEW

File: AsyncDisplayKit/ASVisualEffectNode.h [hunk 1]
```diff
@@ -0,0 +1,41 @@
+//
+//  ASVisualEffectNode.h
```
levi: You'll want to include this header in AsyncDisplayKit.h so that it can be used outside of the framework. Also, update the copyright preamble to reflect the one used in the other classes.

Podfile.lock
```diff
@@ -1,13 +0,0 @@
-PODS:
```
levi: You'll want to keep Podfile.lock.

AsyncDisplayKit/ASVisualEffectNode.m
```diff
@@ -0,0 +1,31 @@
...
+@implementation ASVisualEffectNode
+
+- (instancetype)initWithEffect:(UIVisualEffect*)visualEffect
+{
+  if (self = [super initWithViewBlock:^UIView *{
+    _visualEffectView = [[UIVisualEffectView alloc] initWithEffect:visualEffect];
+    _visualEffectView.autoresizingMask = (UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight);
+    return _visualEffectView;
+  }]) {
+    self.userInteractionEnabled = NO;
```
levi: What's the purpose of disabling interaction here?

samhsiung: oops, it got pulled in from somewhere else. I'll remove it, thanks for the catch.

appleguy:
Cool idea!  We may want to install some checks or robustness around this to make sure that people aren't confused if they try to do shouldRasterizeDescendents and find that the effect is missing.  Might be able to throw an assertion if rasterization encounters one of these nodes...

samhsiung:
Going to test contentView children functionality some more

Issue #59, opened 2014-10-16
[ASMapNode] Create MapKit snapshot node


jbejean:
It's possible to use AsyncDisplayKit with MKOverlayRenderer (MKPolygon or MKPolyline)?
It might be awesome!

appleguy:
@aaronschubert0 ran across this old task...we should do it soon!

appleguy:
We may even be able to support tapping (or at least a method call) to toggle to a live MKMapView.  The default state, and most useful one, would be for it to use -fetchData to trigger loading a snapshot — basically a convenient hook into the intelligent preloading system for map snapshots.  But instead of naming it ASMapSnapshotNode, calling it ASMapNode will give some option value in potentially supporting interaction later.

aaronschubert0:
Ha, good find @appleguy ! This is would be a really awesome node to be honest. The live MKMapView sounds even better! I think I might actually go ahead and give this a shot! Any design considerations I should be aware of, before I start?

appleguy:
@aaronschubert0 Thanks so much for implementing this.  It was an awesome "present" to hear from you on Nov 2nd (my birthday) that you would build this for the community.  Congratulations on it launching in 1.9.3!  Double win for closing one of our earliest feature requests.

aaronschubert0:
Thanks @appleguy! That's awesome, I didn't know that. Yeah it's great to be contributing to the project. It's a really great environment to be part of. Hopefully more contributions to come in the future :)

PR #842, opened 2015-11-12
Add ASMapNode

aaronschubert0:
I've started implementing an ASMapNode as discussed here #59 this node uses the `MKMapSnapshotter` API to provide a snapshot of a map. It then supports turning that snapshot into a `MKMapView` when the user taps on it which fully supports the `MKMapView` delegate, as well as annotations.

I'm still in two minds whether the snapshot should use an `ASImageNode` or whether it'd be better to just draw the image directly to the node.

The main issue I'm still facing is a stutter on scrolling when objects are being dealloc'ed in `clearFetchedData` I'm certain this can be improved, and would love community help on this.

cc @appleguy


appleguy:
@aaronschubert0 fantastic, thank you!  @garrettmoon, what is Bin's github handle?

garrettmoon:
@binl

AsyncDisplayKit.podspec
```diff
@@ -21,7 +21,7 @@ Pod::Spec.new do |spec|
       'Base/*.{h,m}'
   ]
 
-  spec.frameworks = 'AssetsLibrary'
+  spec.frameworks = 'AssetsLibrary','MapKit'
   spec.weak_frameworks = 'Photos'
```
appleguy: I don't actually know much about this, but can we make this a Weak framework?  @Adlai-Holler I think you'd set up the Photos one for us.  Curious if we can even do that with AssetsLibrary itself.

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,40 @@
...
+#import <MapKit/MapKit.h>
+@interface ASMapNode : ASControlNode
+- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate NS_DESIGNATED_INITIALIZER;
+
+@property (nonatomic, readonly) ASImageNode *mapImage;
+@property (nonatomic, readonly) ASDisplayNode *liveMap;
+/**
+ Whether the map snapshot should turn into a MKMapView when tapped on. Defaults to YES.
+ */
+@property (nonatomic, assign) BOOL hasLiveMap;
```
appleguy: Can you make this default NO?  It's an important feature to have, but live maps are /very/ costly and it's also important to remove them once a scroll view starts moving again, so it should be an opt-in extra cost / functionality.

Any other name ideas?  Maybe becomesLiveMap, or even just showingLiveMap / similar.  Since it is an ASControlNode, it would be reasonable to expect developers to do addTarget:action ... and just enable the live map directly.  That's related to the need to be able to disable it too.

AsyncDisplayKit.podspec
aaronschubert0: @appleguy happy to change this to weak if that's an option, will wait to hear back from @Adlai-Holler

AsyncDisplayKit/ASMapNode.h
aaronschubert0: Yeah that sounds good, I will change it to default to NO. I prefer changing it to becomeLiveMap since that makes it clear that it's an action at that point rather than it becoming a live map later on. This would make it clear that it would show/hide depending on the developer setting YES/NO. That way we can remove any internal actions so that the developer is free to handle the touch in anyway they choose (they may want to open a separate view to see the map in full screen) making it much more flexible.

appleguy: @aaronschubert0 ok, sounds good, let's remove the default touch handling.  Leave it as an ASControlNode subclass so it's easy for a developer to implement.  Let's actually call it "liveMap / getter = isLiveMap" - then you could do something like -setLiveMap:YES or check mapNode.isLiveMap.

The alternative would be a method like -becomeLiveMap, but then we need one for -removeLiveMap too.  The benefit there would be you could create a direct target / action pair where the action literally is just -becomeLiveMap.  However, I think it will be necessary to change the target / action afterwards anyway or another tap on the live map would disable it, which is probably not desired, as a user might just be trying to double tap to zoom in or something.  So requiring a developer to create their own action method (in which they'll set _mapNode.liveMap = YES) makes sense.  I'm open to different name ideas for sure!

AsyncDisplayKit.podspec
appleguy: Looks like we might not hear back in time for 1.9.2, which I hope to ship at the end of tomorrow.  We don't have to include this in that release, although I'd like to, as we'll start using it in Pinterest :).

If you have time to research the Cocoapods feature itself, please do.  I bet this is possible, but I'm wondering how this works for a given user.  Maybe you could try adding it to weak frameworks locally, and add a map node to one of the example projects that doesn't link MapKit, and see what happens.  Presumably you'll get a linker error, and I think that would be totally fine to expect users to have to manually have linked MapKit if they want to use this feature.

appleguy:
I'm ready to merge this if we can make the framework weak, because API tweaks / refinements can easily be done in followup diffs (whereas I don't want anyone to ship an app with a new framework linkage as large as MapKit if they don't use this feature).  Thanks a lot @aaronschubert0 for grabbing this task and contributing a new component to the community!

AsyncDisplayKit.podspec
aaronschubert0: Hey @appleguy  will investigate myself, hopefully we can change it to weak if it's not needed. I have time today from 16:30 GMT so will get back to you in time for 1.9.2, then we can make the getter/setter changes in a separate diff (probably tomorrow)

aaronschubert0:
Hey @appleguy change the pod spec to weak and seems to work on my local, would be good to check once it's merged. Still worried about the scroll stutter when `clearFetchedData` is called, but I guess we can fix this in future diffs, if it's a problem?

AsyncDisplayKit/ASMapNode.h
aaronschubert0: @appleguy I prefer the first suggestion  "liveMap / getter = isLiveMap" this way it's super clear to the developer what is happening. Will have a look at this, this afternoon.

aaronschubert0: Made those changes in 863156b

aaronschubert0:
@appleguy Anything stopping this from being merged?

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,45 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <AsyncDisplayKit/AsyncDisplayKit.h>
+#import <MapKit/MapKit.h>
+@interface ASMapNode : ASControlNode
```
appleguy: By no means a blocker, but let's add a return before and after the @interface line to match the style of most other ASDK headers.

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,45 @@
...
+- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate NS_DESIGNATED_INITIALIZER;
+/**
+ This is the snapshot shot image node, this will be hidden (but not nil) when .liveMap = YES
+ */
+@property (nonatomic, readonly) ASImageNode *mapImage;
```
appleguy: What use cases do you envision for exposing this?  If it is needed, we should probably call it mapImageNode (for a moment I assumed it was a UIImage).

I think it might be more useful to have a UIImage called mapImage with no access to the ASImageNode, but it might be better to provide no access at all.  Providing access requires that we also provide a means to tell when the image is done loading / complete, although I would argue that's one of the most important things we could provide (moreso than the image actually).  Technically it may be possible to determine this with subnodeDidFinishDisplay: or some similarly named method, but I think as-is this is tricky because the ASImageNode is an internal subnode of the ASControlNode.

One nifty idea: how about the ASMapNode subclasses ASImageNode directly, and we remove this entirely?  Then you'd have incidental access to the ASImageNode (although, still wondering what use cases exist for that), we'd have fewer nodes / views overall, and you could still create the live map to put inside of it / cover any image being displayed.  ASImageNode is an ASControlNode, too.

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ This is the ASDisplayNode that backs the MKMapView. This will be nil if .liveMap = NO. To access the underlying MKMapView, in order to set a delegate for example, use (MKMapView *)mapView.view;
+ */
+@property (nonatomic, readonly) ASDisplayNode *mapView;
```
appleguy: I think we indeed should expose the MKMapView, but let's call it what it is and provide a pointer directly to an MKMapView.  I don't think there is value in providing the ASDisplayNode wrapper and then requiring a typecast, especially since that would be required to set the delegate.  We also want to remind users that the MKMapView is not thread-safe, so it being exposed here directly as an MKMapView might make that a bit more clear.

As API feedback, if we were to leave this as a node, it would be important to name it with "Node" in the name (although the fact that "mapNode" would be confusingly recursive I think highlights the fact that the ASDisplayNode wrapper isn't the ideal approach).  Maybe a better name is liveMapView - but mapView seems nice too.

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ Set this to YES to turn the snapshot into an interactive MKMapView and vice versa. Defaults to NO.
+ */
+@property (nonatomic, assign, getter=isLiveMap) BOOL liveMap;
+/**
+ @abstract Explicitly set the size of the map and therefore the size of ASMapNode. Defaults to CGSizeMake(constrainedSize.max.width, 256).
```
appleguy: Why does it default to 256 high?  Is that the right size on an iPad Pro? :)

I think we should default to returning a -layoutSpecThatFits: of an ASRatioLayoutSpec, perhaps.  Maybe there's a better approach? @nguyenhuy  As for the ratio, maybe just a square.  People will have to customize it anyway.

Importantly, the mapView (and if you keep the separate imageNode) will have to be sized to be the full size of the ASMapNode.  The user may choose to set the frame / bounds of the ASMapNode directly to anything, even something that is not the same as what the -layoutSpecThatFits: asks for, and as a framework class we should handle this.

In a way, for this very unusual class, it might not be important to return anything specific for -layoutSpecThatFits: because the map is inherently scalable and has no fixed "form" (unlike text, or even an image with an aspect ratio).  So we might just want to return an empty ASLayoutSpec, and implement a manual -layout method that just sets self.mapView.frame = self.bounds;

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,45 @@
...
+ @discussion If the mapSize width or height is greater than the available space, then ASMapNode will take the maximum space available.
+ @result The current size of the ASMapNode.
+ */
+@property (nonatomic, assign) CGSize mapSize;
+/**
+ @abstract Whether ASMapNode should automatically request a new map snapshot to correspond to the new node size. Defaults to YES.
+ @discussion If mapSize is set then this will be set to NO, since the size will be the same in all orientations.
+ */
+@property (nonatomic, assign) BOOL automaticallyReloadsMapImageOnOrientationChange;
```
appleguy: Maybe this should be needsMapReloadOnBoundsChange?  Similar to needsDisplayOnBoundsChange in the Apple frameworks?  Main point / question: does this specifically apply to orientation changes, or should it also apply if something else changed the size / shape / position of the map? (e.g. the Yelp "Mo' Map" button)

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,45 @@
...
+/**
+ Set the delegate of the MKMapView.
+ */
+@property (nonatomic, weak) id <MKMapViewDelegate> mapDelegate;
```
appleguy: Aha, this is nice to have.  Extend the comment to explain that this can be set even before the mapView is created, but will be set on the map in the case that the liveMap mode is engaged.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+  if (!(self = [super init])) {
+    return nil;
+  }
+  self.backgroundColor = ASDisplayNodeDefaultPlaceholderColor();
+  _automaticallyReloadsMapImageOnOrientationChange = YES;
+  _liveMap = NO;
+
+  _options = [[MKMapSnapshotOptions alloc] init];
+  _options.region = MKCoordinateRegionMakeWithDistance(coordinate, 1000, 1000);;
+  _mapImage = [[ASImageNode alloc]init];
```
appleguy: Add a return above this; add a space after the alloc].

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+  _mapImage.clipsToBounds = YES;
+  [self addSubnode:_mapImage];
+  _maxSize = self.bounds.size;
```
appleguy: self.bounds.size is always CGSizeZero here, right?  If so, go ahead and move this up with the other "scalar" property sets and just set it to CGSizeZero.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+  return self;
+}
+
+- (void)addAnnotations:(NSArray *)annotations
```
appleguy: I wonder, should this be setAnnotations:?  With only an add method, an annotation could never be removed.  Maybe that's not as common, but it seems like a property that allows reading the current set of annotations and doing setAnnotations would have more flexibility - IF it is not difficult to correctly support.  If it is challenging to cover all the edge cases with that route, we should keep this.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+- (BOOL)isLiveMap
+{
+  ASDN::MutexLocker l(_propertyLock);
+  return _liveMap;
+}
+
+- (void)setLiveMap:(BOOL)liveMap
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if (liveMap == _liveMap) return;
```
appleguy: Though I like this style, let's match the framework standard here and use explicit { } brackets with the return on its own line.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+- (CGSize)mapSize
+{
+  ASDN::MutexLocker l(_propertyLock);
+  return _mapSize;
+}
+
+- (void)setMapSize:(CGSize)mapSize
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if (CGSizeEqualToSize(mapSize,_mapSize)) {
```
appleguy: Here's an example of the expected brace style.  Add a space after the comma, though.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+}
+
+- (void)setAutomaticallyReloadsMapImageOnOrientationChange:(BOOL)automaticallyReloadsMapImageOnOrientationChange
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if (_automaticallyReloadsMapImageOnOrientationChange == automaticallyReloadsMapImageOnOrientationChange) {
+    return;
+  }
+  _automaticallyReloadsMapImageOnOrientationChange = automaticallyReloadsMapImageOnOrientationChange;
+
```
appleguy: Can remove the last return in this method before the brace.  Since this method has no side effects, you could also remove the if statement / early return and just unconditionally set the ivar after obtaining the lock.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+  [super fetchData];
+  [self setUpSnapshotter];
+  [self takeSnapshot];
+}
+
+- (void)clearFetchedData
+{
+  [super clearFetchedData];
+  if (_mapView) {
+    [_mapView removeFromSupernode];
```
appleguy: It's a little unusual to remove the live map view in -clearFetchedData, but I think it may behave well in practice.  However, what happens when the map scrolls back into view?  Won't _liveMap still be set to YES, but when fetchData is called we will simply create a map snapshot image and not restore the interactive map?  I think that behavior could be OK (reverting to the non-interactive map after it is scrolled away), but we'd have to set _liveMap = NO...

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+        _mapImage.image = finalImage;
+      }
+    }];
+  }
+}
+
+- (void)resetSnapshotter
+{
+  if (!_snapshotter.isLoading) {
+    _options.size = _nodeSize;
```
appleguy: Wouldn't _nodeSize be incorrect if the creator of the ASMapNode positions it using its -layoutSpecThatFits method?  They could even use an ASStaticLayoutSpec, and give specific / fixed values.  Assuming this method is called on the main thread, I think you can just use self.bounds.size here - but let me know if you see a reason we shouldn't do that!

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+    _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
+  }
+}
+
+#pragma mark - Action
+- (void)addLiveMap
+{
+  if (self.isNodeLoaded && !_mapView) {
```
appleguy: Should this either check, or set, _isLiveMap?

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+      [mapView setRegion:_options.region];
+      [mapView addAnnotations:_annotations];
+      return mapView;
+    }];
+    [self addSubnode:_mapView];
+    _mapImage.hidden = YES;
+  }
+}
+
+- (void)removeLiveMap
```
appleguy: Should this method be called from -clearFetchedData so that there is a single override point for tearing down the live map, including resetting of the boolean if needed?

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,236 @@
...
+  [super layout];
+  if (_mapView) {
+    MKMapView *mapView = (MKMapView *)_mapView.view;
+    mapView.frame = CGRectMake(0.0f, 0.0f, self.calculatedSize.width, self.calculatedSize.height);
+  }
+  else {
+    _mapImage.frame = CGRectMake(0.0f, 0.0f, self.calculatedSize.width, self.calculatedSize.height);
+    if (!CGSizeEqualToSize(_maxSize, self.bounds.size)) {
+      _mapImage.preferredFrameSize = self.bounds.size;
+      _maxSize = self.bounds.size;
```
appleguy: Ah, I see _maxSize is set here...not sure what _nodeSize is for but I think that one was used in another method.  If we can eliminate some or all of these instance variables and just use self.bounds.size, that might be clearer / more reliable.

README.md
```diff
@@ -22,7 +22,7 @@ pod 'AsyncDisplayKit'
 
 (ASDK can also be used as a regular static library:  Copy the project to your
 codebase manually, adding `AsyncDisplayKit.xcodeproj` to your workspace.  Add
-`libAsyncDisplayKit.a`, AssetsLibrary, and Photos to the "Link Binary With
+`libAsyncDisplayKit.a`, MapKit, AssetsLibrary, and Photos to the "Link Binary With
```
appleguy: is it accurate to say that you can "optionally" add these other frameworks to use certain ASDK classes that need them?

appleguy:
@aaronschubert0 hey there - sorry it took me a long time to get to the review.  There have been a lot of PRs waiting in the queue as I've simply fallen behind on stuff.  I've mostly caught up, and hope to release 1.9.2 with several improvements including my ASInterfaceState feature soon!

Though a lot of my feedback is just small stylistic nits, there are a handful of API design things to work through.  I understand fully if this ended up being more than you bargained for, and so if you don't have a chance to revisit it, rest assured that the code is still very valuable to the project and it will be tweaked and landed soon!

Thanks for your efforts, and let me know if you have any questions.

aaronschubert0:
Hey @appleguy this feedback is awesome, will go through it and respond to each comment, definitely want to get this right and have it clear for all developers. Pretty busy at work this week but will tend to this very soon.

AsyncDisplayKit/ASMapNode.h
nguyenhuy: I think ASMapNode may not provide any (meaningful) layout at all, so neither -layoutSpecThatFits: nor -calsizeSizeThatFits: needs to be implemented. The result is that its parent node must decide its size/layout. For example, the parent node can wrap it inside an ASRatioLayoutSpec, or set a preferredFrameSize to it, or make it flexible inside a stack.

AsyncDisplayKit/ASMapNode.h
aaronschubert0: @appleguy Yeah I've always wrestled with exposing this property, the main use case why I wanted to expose this property was allow developers to manipulate the image if they needed. Subclassing ASImageNode directly would solve both these problems so I'm in favour of going that route :)

AsyncDisplayKit/ASMapNode.h
aaronschubert0: Agreed makes more sense to expose the MKMapView directly.

AsyncDisplayKit/ASMapNode.h
aaronschubert0: @appleguy The reason it defaults to 256 high is because the MKMapSnapshotOptions used in MKMapSnapshotter defaults to 256x256 so I took this as a sensible default. I agree with you and @nguyenhuy that ASMapNode should have no fixed form/size so any solution that accomplishes this would be best, I prefer for ASMapNode to take the height that the parent node gives it, rather than to calculate a intrinsic size for ASMapNode

AsyncDisplayKit/ASMapNode.h
aaronschubert0: Yeah that sounds good. This was specifically built with rotation changes in mind but it would make more sense to abstract it to size changes (maybe shape). Obviously `-layout` is called on size changes but I don't want to load this with bloat. Introducing a `-refresh` method seems quite cumbersome and adds another complexity for the developer. Are there any other good access points that ASDK provides that I could implement this in? Will take a deeper look at `ASDisplayNode` to see if anything stands out.

AsyncDisplayKit/ASMapNode.h
aaronschubert0: Will extend the comment. Would be nice for this to be the single access point to MKMapView in ASMapNode but I suspect that developers will find it more helpful to have access to MKMapView directly, for e.g. to do point-coordinate conversions.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: :+1:

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: Good point, will clean that up.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: I agree setAnnotations: makes more conceptual sense. This is basically what the method already does, since it just takes the new annotations and redraws. So more of a naming change. Possibly will support passing `nil` or an empty array to clear annotations, does ASDK have any guidelines on this?

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: Sure thing.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: That makes sense! Here I was more following previous examples in ASDK but I prefer your suggestion. The less code the better!

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: To be honest this was what I was most unsure about but seemed to work in practice. Maybe there are better methods to call this code in? The current default behaviour is to revert back to the non-interactive map since if we didn't do this then the memory would spike, and it would be wasteful to have a MKMapView offscreen. I do like the idea of restoring the interactive map, if _liveMap = YES, sort of like autoplaying videos in FB. I think we should take this route because either the developer or the user has interacted in a way to turn the snapshot into a interactive map, so it would make sense to recreate the same experience that they last saw.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: From what I've gather from the layout spec format, is that it recursively (or eventually fallback) to -calculateSizeThatFits: so _nodeSize would be the correct size. Having said that, I don't like having another scalar property to track the node's size so I'm going to look to refactor this out in favour of self.bounds.size (or similar in built properties)

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: -addLiveMap is only called from -setLiveMap: when _isLiveMap = YES. So it is set beforehand and assumes that _isLiveMap is set to YES. Since this is an internal method, I would say that would suffice. Or do you prefer to be more defensive in this regard?

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: Yes absolutely!

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: Yeah totally agree, quite messy this way. I was using them mainly to accurately detect orientation changes. Any ideas on that? Although if we abstract to reloading the image on size changes this will be superseded.

README.md
aaronschubert0: I think it's accurate to say that you can optionally add MapKit & Photos. From the pod spec AssetsLibrary is still a 'required' framework, so not sure about that one.

aaronschubert0:
Hey @appleguy implemented most of the feedback that you gave. The changes from the previous iteration are:
- ASMapNode now inherits from ASImageNode, rather than ASControlNode. The mapImage property has been removed because of this change. 
- The mapView property is now a MKMapView rather than a ASDisplayNode
- The mapSize property has been removed. If developers wish to set an explicit size for ASMapNode, then they can set the preferredFrameSize property.
- "automaticallyReloadsMapImageOnOrientationChange" has been renamed to "needsMapReloadOnBoundsChange"
- "-addAnnotations:" has been renamed "setAnnotations:"
- All scalar size properties (_maxSize & _nodeSize) have been removed (yay!) in favour of `self.bounds.size` or `self.calculatedSize`
- "-calculateSizeThatFits:" has been removed as suggested by @nguyenhuy 
- ASMapNode now automatically fetches a new map snapshot if it's size is changed, as opposed to only on orientation change.
- A new internal "CLLocationCoordinate2D" property has been introduced. This is to support an ASMapNode (that has liveMap = YES) to return to the correct liveMap view after the user has scrolled from the node and then returned to it at a later occasion. 
- Syntax cleanup.

P.S. I don't think Travis CI is running against these commits?

nguyenhuy:
There are some conflicts against master. I bet Travis will run once they are resolved. Awesome work, can't wait to try it out :+1:

Also, it would be great if we have a working sample of this new node. Definitely not a blocking issue though.

aaronschubert0:
Thanks @nguyenhuy ! That's a good idea! Will need to come up with a good example. I'll work on that once this has been finalised :)

appleguy:
Big progress, thanks @aaronschubert0!  I'll review as soon as I have stabilized a few issues that have come up with the 1.9.2 release.  Definitely by Sunday.

aaronschubert0:
@appleguy Awesome, looking forward to it!

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,43 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <AsyncDisplayKit/AsyncDisplayKit.h>
+#import <MapKit/MapKit.h>
+
+@interface ASMapNode : ASImageNode
+
+- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate NS_DESIGNATED_INITIALIZER;
```
appleguy: Aha, doesn't this need to be an MKMapRegion or something else that fully specifies the visible extent?

Later - not necessary for you to do personally @aaronschubert0, more like feature request - we could add amazing helper functions that aid a user who has a CLLocationCoordinate2D and perhaps a radius as an expression of zoom scale, and we could create a region for them.

I think I will merge this diff now, but only because 2.0 is coming up and I feel confident to make a breaking API change in that release.  We really should change this though because right now the zoom level seems unspecified, and it would force creation of the mapView if set through there.

AsyncDisplayKit/ASMapNode.h
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ This is the MKMapView that is the live map part of ASMapNode. This will be nil if .liveMap = NO. Note, MKMapView is *not* thread-safe.
```
appleguy: Nice comment - thanks for the very brief callout reminder on thread safety.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,210 @@
...
+  if (!_snapshotter.isLoading) {
+    _options.size = self.calculatedSize;
+    _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
+  }
+}
+
+#pragma mark - Action
+- (void)addLiveMap
+{
+  if (self.isNodeLoaded && !_mapNode) {
```
appleguy: I think we should improve this method in a followup diff.

- Does it work correctly to set isLiveMap before the node is loaded?  This code path bails, and I couldn't readily find what else creates the _mapView on node loading if not this (e.g. there is no -didLoad method).
- Add spaces after the "alloc]" calls below.
- Let's test the memory behavior here.  It is possible there is a strong reference cycle.  You may need to create a "weakSelf" local variable and access things like mapDelegate on it; there is an implicit retain / reference to self in order to access the instance variables directly within the block.  Even after the view load block is called, it will be retained by the node.  We should test a sample app with an ASMapNode that goes to the fully liveMap state, and then deallocate the ASMapNode, and check with Instruments Leaks + Allocations that it is destroyed.
- It might be best to just init with CGRectZero and call setNeedsLayout on weakSelf at the end.  It might be equally good to access weakSelf.bounds (make a local variable CGSize boundsSize = weakSelf.bounds.size in this case, to avoid calling bounds property twice) - and replace calculatedSize with that - because nothing requires the node to be positioned / sized equal to the calculatedSize.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,210 @@
...
+    _mapNode = [[ASDisplayNode alloc]initWithViewBlock:^UIView *{
+      _mapView = [[MKMapView alloc]initWithFrame:CGRectMake(0.0f, 0.0f, self.calculatedSize.width, self.calculatedSize.height)];
+      _mapView.delegate = _mapDelegate;
+      [_mapView setRegion:_options.region];
+      [_mapView addAnnotations:_annotations];
```
appleguy: Maybe we can implement a setAnnotations method, and if it is called after the mapView is created, pass it on correctly.  If annotations already exist, we should ask the map to remove the existing _annotations and then add the new set (and save it in the ivar).

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,210 @@
...
+
+- (void)removeLiveMap
+{
+  if (_mapNode) {
+    _centerCoordinateOfMap = _mapView.centerCoordinate;
+    [_mapNode removeFromSupernode];
+    _mapView = nil;
+    _mapNode = nil;
+  }
+  self.image = nil;
```
appleguy: Is this necessary?  We may want to kill the image if some properties change like the region, although I actually think it could be better to leave it there until the new one loads.  May provide an option like clearsMapWhileLoading; if we later added this, it should be verified to work correctly with a .placeholderImage set (return to showing the placeholder if the map is cleared to load another one).

Anyway in this method let's remove self.image = nil for now.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,210 @@
...
+// Layout isn't usually needed in the box model, but since we are making use of MKMapView which is hidden in an ASDisplayNode this is preferred.
+- (void)layout
+{
+  [super layout];
+  if (_mapView) {
+    _mapView.frame = CGRectMake(0.0f, 0.0f, self.calculatedSize.width, self.calculatedSize.height);
+  }
+  else {
+    // If our bounds.size is different from our current snapshot size, then let's request a new image from MKMapSnapshotter.
+    if (!CGSizeEqualToSize(_options.size, self.bounds.size)) {
```
appleguy: don't we need to reset _options.size to self.bounds.size in this case?

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -0,0 +1,210 @@
...
+      if (_needsMapReloadOnBoundsChange && self.image) {
```
appleguy: should we remove the check for self.image here?  What if I ask the map to reload on bounds change, and then I put it onscreen (and it starts loading) - then change its size before it finishes loading?

appleguy:
@aaronschubert0 thanks again - really happy to be merging this.  If you have time, I do think one more round of changes is needed - but this should prepare the class to be included as a primary API in the 2.0 release!

aaronschubert0:
@appleguy Woohoo! Thanks a lot for the thorough code review. I'll definitely give it another pass to give it that polish.

AsyncDisplayKit/ASMapNode.h
aaronschubert0: Hey @appleguy yeah I've been wanting to make this change since the zoom level is currently at 1km. This will make it unambiguous and available to control for everyone. It's going to be a `MKCoordinateRegion`. In terms of the helper functions, what use cases do you envision for this? I'm just asking since you only need a `CLLocationCoordinate2D` and a radius to create a `MKCoordinateRegion` so not sure how helpful it would end up being or what advantages it would have over the built in method?

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: @appleguy 
- It doesn't work to set isLiveMap before the node is loaded since it's protected by the if condition checking if the node is loaded. I think to implement this behaviour we would have to implement -didLoad and check if isLiveMap = YES then call -addLiveMap. Having thought about this, I think we should support this functionality since otherwise there isn't any other way to start of with a live map. The problem is that since ASMapNode inherits from ASControlNode (indirectly) then it doesn't respond to touches until a target/action is added, so I need to set `self.userInteractionEnabled = YES;` directly to achieve this but I think this is acceptable.
- I've ran the current version of ASMapNode through Instruments and they seem to be getting destroyed properly, however I presume it would be better to have weakSelf anyway so I will do that.
- I've went with init with CGRectZero and calling setNeedsLayout since this feels cleaner.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: Hey @appleguy, We already have the setAnnotations method on line 49 so I've extend this to include the above behaviour you mentioned.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: Good point, will remove it for now.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: I had this in -resetSnapshotter but have moved it here to make it more clear.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: Yeah let's do that. Hadn't thought of that case.

PR #850, opened 2015-11-16
Introduce ASInterfaceState on ASDisplayNode.

appleguy:
Later, this will be added to ASViewController.  Right now it is not hooked up to drive anything.
It will become the codepath that drives calls like -fetchData and -clearContents.


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1683,6 +1683,18 @@ - (void)recursivelyClearFetchedData
...
 }
 
+- (ASInterfaceState)interfaceState
+{
+  return _interfaceState;
+}
+
+- (void)setInterfaceState:(ASInterfaceState)interfaceState
+{
+  if (interfaceState != _interfaceState) {
```
levi: Shouldn't this need to be guarded by the mutex?

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -33,6 +33,27 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
...
+
+typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
+{
+  /** The element is not predicted to be onscreen soon and preloading should not be performed */
+  ASInterfaceStateNone          = 1 << 0,
+  /** The element may be added to a view soon that could become visible.  Measure the layout, including size calculation. */
+  ASInterfaceStateMeasureLayout = 1 << 1,
+  /** The element is likely enough to come onscreen that disk and/or network data required for display should be fetched. */
+  ASInterfaceStateFetchData     = 1 << 2,
+  /** The elemant is very likely to become visible, and concurrent rendering should be executed for any -setNeedsDisplay. */
```
levi: elemant!

AsyncDisplayKit/ASDisplayNode.mm
levi: Might be helpful to introduce a callback method: `-[ASDisplayNode interfaceStateDidChange:(ASInterfaceState)previousState]` for subclassing.

appleguy: Yes, absolutely.  I may finish the initial internal implementation without this, but once proven to be the correct API, this will definitely be developer-facing and could conceivably even replace the fetch/clear methods

AsyncDisplayKit/ASDisplayNode.h
appleguy: LOL.  Fixed locally.

AsyncDisplayKit/ASDisplayNode.mm
appleguy: Added the locking, but decided to wait on interfaceStateDidChange: until we know what the protocol + adoption with ASViewController / table / collection will look like.  The main known use case involves simply reading self.interfaceState (which is accessible) on setURL: or similar to check if a load should be kicked off.  It is also possible to override -setInterfaceState: and call super if needed, but I agree this is not ideal for long term usage outside the framework.

appleguy:
@Adlai-Holler - check it out.  Can you try this with a quick modification to ASMultiplexImageNode to see if you can have it kick off the network load if self.interfaceState & ASInterfaceStateFetchData?

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -33,6 +33,27 @@ typedef CALayer *(^ASDisplayNodeLayerBlock)();
...
 /**
+ Interface state is available on ASDisplayNode and ASViewController, and
+ allows checking whether a node is in an interface situation where it is prudent to trigger certain
+ actions: measurement, data fetching, display, and visibility (the latter for animations or other onscreen-only effects).
+ */
+
+typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
+{
+  /** The element is not predicted to be onscreen soon and preloading should not be performed */
+  ASInterfaceStateNone          = 1 << 0,
```
Adlai-Holler: Could we have this be raw value 0? That way when it's imported into swift, both `[]` and `.None` will correspond to the same value.

Adlai-Holler:
@appleguy Sure, this looks great!

Issue #855, opened 2015-11-17
How use LayoutSpec for specific task


bimawa:
Hi all!
I have intresting task for CollectionView Header. I want to add ASNetworkImageNode with width as screen and height as loaded Image. ContentMode must be UIViewContentModeScaleAspectFit. Can i make this task with LayoutSpec, or no? 
![untitled](https://cloud.githubusercontent.com/assets/1393809/11207063/cdba46ba-8d34-11e5-8e41-d88316a7b610.png)

smiLLe:
try wrapping your ASNetworkImageNode into an ASRatioLayoutSpec

bimawa:
@smiLLe thx for advice, i go try it )

bimawa:
@smiLLe  Yes it what i need. Thx a lot

bimawa:
@smiLLe  i have another trouble, mb u can help me.
Now my HeaderSection View Height dependence at ImageNode height, and in Header View i have 2 addition ASLayoutSpec (like at image)  
![untitled](https://cloud.githubusercontent.com/assets/1393809/11209724/c5c22472-8d44-11e5-8042-e7ecefe333fa.png)
How, if i can, set ImageNode height priority highest then i receive in receiver `constrainedSize` constraints, of method `- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize`? And i now must get new size of HeaderView and update all collectionView?

bimawa:
I understand how Rate works its not what i want unfortunately :(

nguyenhuy:
Does wrapping the image node and 2 other specs inside a vertical stack work for you?

nguyenhuy:
@smiLLe Thanks for suggesting ratio spec. It is indeed the ideal solution I think.

bimawa:
@nguyenhuy Yes I do. Ratio rly help for save proportion for imageNode. But i need a little bit  else. If Image horizontale, as on screenshot, i want get maximum heigh for node, depend of width screen.
![1447764903901](https://cloud.githubusercontent.com/assets/1393809/11212014/167b5ff4-8d55-11e5-8e62-86cec20a2c4e.jpg)
And if Image Verticale pref. it must scale by width screen. And offset other LayoutSpec down by hight
(Example without RatioSpec)
![1447766116587](https://cloud.githubusercontent.com/assets/1393809/11212398/6174093c-8d57-11e5-9dcc-b99d9a3c5d82.jpg)

[...]

PR #857, opened 2015-11-17
Trigger a relayout once the media image of PostNode in SocialAppLayout is loaded

nguyenhuy:
I was looking for a sample code to answer [@bimawa's question in #855](https://github.com/facebook/AsyncDisplayKit/issues/855#issuecomment-157397779) when I found out that, although a reserve-image-ratio logic is available in PostNode of SocialAppLayout example, it isn't used. No use no more :)

### DIFF
--- a/examples/SocialAppLayout/Sample/PostNode.m
+++ b/examples/SocialAppLayout/Sample/PostNode.m
@@ -14,6 +14,9 @@
 #import "LikesNode.h"
 #import "CommentsNode.h"
 
+@interface PostNode() <ASNetworkImageNodeDelegate>
+@end
+
 @implementation PostNode
 
 - (instancetype)initWithPost:(Post *)post {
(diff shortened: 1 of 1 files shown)


bimawa:
@nguyenhuy I'm biMawa (Maxim) :P . And yea i see that too i forgot say about that.

nguyenhuy:
Oops, sorry for the mistake :(

bimawa:
No problem ;)

appleguy:
@nguyenhuy Thanks for the fix.  This code is a little hard to understand, however - certainly as an author, it would be a bit of a struggle to write.  Is there any way you can think of to trigger this automatically?  Should the image node simply mark itself as needing layout, because I am nearly certain that once the layout pass runs, that will make the supernode perform a layout?  This would be ideal solution, if feasible.

Issue #861, opened 2015-11-18
[ASRangeController] Range callback methods aren't always called for replacing nodes at existing paths.


Adlai-Holler:
**EDIT: Issue is more general than this, see next comment.** I noticed a bug in ASTableView. I have a table view that has two very different modes, and I use `reloadDataImmediately` to switch between them. I noticed that when I switch to a mode that has more rows, `fetchData` is generally called on the new nodes, but when I switch to a mode with fewer rows, it isn't. 

Digging deeper I saw that even after `reloadData`, `ASRangeController` still regards the old index paths as handled, so post-update, in `updateVisibleNodeIndexPaths` I would expect `addedIndexPaths` to contain all the index paths in the range but instead it subtracts the index paths that were in the range pre-update.

I'm about to submit a pull request that fixes this issue for me. Not being a range controller expert, it may be flawed, but there you have it.

Adlai-Holler:
Actually, I think this issue is more general than just `reloadData` – something like the range controller only cares about index paths and not their contents.

I've currently got a setup where I use a section with 1 cell to hold my "Loading…" placeholder. Then when the content comes in, I submit a batch update that contains "remove section 0, insert section 0". The first cell in the new section never gets a `fetchData` call. All other cells do. As if the range controller said "ah! I've already handled (0,0)" but it didn't realize that (0,0) had its contents replaced. Going to attempt to reduce this down into a test case, shouldn't be too hard.

1nput0utput:
@Adlai-Holler this explains very well my issue, I have duplicate nodes and sometimes incorrect indexPaths after having fetched a second page of data.

appleguy:
@Adlai-Holler solid debugging here — a very clear explanation of the problem.  I think this is strong support for the impression that ASRangeController needs its first big refactor, as today its core logic is from the same era as the original ASDataController (!!!).

@1nput0utput could you share any details on the issue you encountered?  Was your case specific to reloadData (and thus fixed by the patch linked above), or is the editing scenario described later the one you are seeing?

As per usual with these issues, not having a test case is one inhibition to fixing with confidence.  I'm actively mucking about in ASRangeController (and @levi too, who did a sweeping cleanup of naming and organization of the file in prep for deeper changes).  But with the weekend dwindling, I may not be able to fix this in time for 1.9.2...

[...]

PR #862, opened 2015-11-18
Add a Reset Method to RangeController, Use In reloadData

Adlai-Holler:
Pursuant to #861, here's my attempt at a fix. It resolves the issues I'm having with nodes not always getting `fetchData` post-`reloadData` but I'm hardly an expert on ASRangeController. I threw in the main thread assertion kind of willy-nilly to be honest.


Adlai-Holler:
Suspect the problem is more general than this, see #861 for discussion.

PR #868, opened 2015-11-19
Clean up syntax and organization of ASRangeController

levi:
I'm currently reading through range controller and saw a few things that needed some syntax and organization clean up:

- Removed deprecated layout controller methods that were long longer needed.
  - Now that we're closer to a 2.0, we should start the discussion of removing the remaining deprecated methods that live in ASDisplayNode—there are only three.
- Cleaned up syntax usage in ASRangeController
- Separated data and callback methods into data source and delegate protocols, respectively. It's now a little clearer what data is required for the range controller to work. In the process, I reordered some of the naming of the methods to make them a bit more readable.


appleguy:
Looks great, no changes requested before landing!

Issue #849, opened 2015-11-16
[ASLayoutSpec] Provide convenience wrapper around alignItems and justifyContent.


appleguy:
Though commonly used with the Web / CSS / other Flexbox implementations, the most consistent feedback about the ASDK 2.0 API is that alignItems & justifyContent (along with a few other Stack properties) are hard to understand.  I absolutely agree.

I don't want us to remove these properties as they do have a logic behind them, and the components that are implemented are very well structured and tested.

However, I believe we should be able to implement something like:

- .horizontalAlignment & .verticalAlignment
- ASAlignmentLeft, ASAlignmentCenter (reuse for vertical, or call this one Middle?), ASAlignmentRight
- ASAlignmentTop, ASAlignmentCenter, ASAlignmentBottom

We may be able to combine these into a single property called "alignment" and use bitfields or something, but I don't know if that would be an improvement.

The KEY would be to ensure that both individual items, as well as the overall stack, support this.  Importantly, I think as implemented today, it's either justify or align that doesn't work on a per-item basis.  How difficult would this be to implement, I wonder?

Though there are a few other changes I consider necessary for 2.0, I think this is the most important one in that it is probably the most work and the most impactful.

appleguy:
Landed - earlier than expected, included for 1.9.2!

PR #873, opened 2015-11-20
[ASLayoutSpec] Implement horizontal and vertical alignments

nguyenhuy:
My experiment on #849.

The idea is to provide easier-to-reason alignment properties that are applied to all items in stack layout. Whether an horizontal alignment means "align items" or "justify content" depends on the direction of the stack and is resolved ~~immediately~~ automatically. Same thing for vertical alignment.

I don't provide per-item horizontal and vertical alignments because, depends on the stack direction, only one of them can be resolved. This is because align can be applied on a per-item basis (i.e alignSelf) while justify cannot. For example, in a horizontal stack, vertical alignment equals to alignSelf while horizontal alignment is meaningless.

Another issue is that changing a stack direction *after* setting horizontal and/or vertical alignments will result in unexpected behaviours. Few ideas on how to handle this:
- Store alignments and resolve them right before measurements.
- Automatically swap alignments after a direction change.
- Make direction read-only and can only be set via init.
- Assert and warn user (in direction setter).
- Make it clear in the documentation that after a direction change, horizontal and vertical alignments must be revisited.

I added some test cases to show how the new alignments can be used. Please let me know your thoughts.

EDIT: Added another (and probably the best) idea on handling direction change, hence it's listed first.


rcancro:
My hero!!! 

Do you think it will be possible to create a justifySelf property on ASLayoutable?

nguyenhuy:
I think it is possible, but would require quite a large change in the way stack layout distributes violation (ASStackPositionedLayout). More importantly, what is the behaviour of justifySelf that you are having in mind?

rcancro:
I was mostly just thinking about being able to set horizontal/vertical alignment on a per-child basis.

appleguy:
@nguyenhuy I think we should prevent the stack direction from being changed.  Because a layout spec locks down as immutable after it is returned anyway, there is no valid use case for changing the stack direction.

This would be a bit unusual, but let's keep the direction property as writable, but add comments that it can only be set once — and throw an assertion if it is changed after it has already been set.

Implementing justifySelf would be a huge development, as it would allow us to extend this valuable pattern to each item.  I think that would be a "holy grail" of layout in ASDK, to be honest :).  We should do more research on whether a) CSS / Flexbox support this, b) React or CK have evaluated supporting this (or do now?), and c) make a final decision on whether we'll aim to do this.  If it looks like it will be too complicated to do, that decision itself is useful as we should look at better documenting alignSelf.

nguyenhuy:
@appleguy I agree that we should keep the direction property as writable. However, I propose a slightly different implementation approach (e61f828), because throwing an assertion if the direction is changed more than once doesn't handle all the edge cases. For example, below code would run just fine with no assertion thrown:
```objective-c
ASStackLayoutSpec *stack = [[ASStackLayoutSpec alloc] init];
// Horizontal and vertical alignments are determined using the default direction (horizontal)
[stack setHorizontalAlignment:ASAlignmentRight];
[stack setVerticalAlignment:ASAlignmentBottom];
// Changing the direction now doesn't throw an assertion because it is the first (external) change
// But above alignments will be invalid and need to be set again
stack.direction = ASStackLayoutDirectionVertical;
```

Regarding justifySelf:
- I think CK doesn't supported it as of now, though I'm not sure if CK has evaluated supporting. Not sure about React either. 
- justifySelf is defined in CSS Box spec but not in Flexbox spec. This is probably the reason:

> Flex Items:
> This property does not apply to flex items, because there is more than one item in the main axis. See flex for stretching and justify-content for main-axis alignment. [CSS3-FLEXBOX]

(Source: http://www.w3.org/TR/css-align-3/#justify-self-property)
- I think it will be cool if we decide to implement justifySelf, but we will be pretty much on our own. That's why I asked @rcancro about his idea of its behaviour, because that's the first thing we need to clearly define and I have absolutely no idea.

AsyncDisplayKit/Layout/ASStackLayoutDefines.h
```diff
@@ -67,3 +67,23 @@ typedef NS_ENUM(NSUInteger, ASStackLayoutAlignSelf) {
...
 };
+
+/** Orientation of children along horizontal axis */
+typedef NS_ENUM(NSUInteger, ASHorizontalAlignment) {
+  /** Left aligned */
+  ASAlignmentLeft,
+  /** Right aligned */
+  ASAlignmentRight,
+  /** Center aligned */
+  ASAlignmentMiddle,
```
appleguy: Small detail, but let's put this in the middle of the three just like Center below :)

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
```diff
@@ -35,7 +35,7 @@
  */
 @interface ASStackLayoutSpec : ASLayoutSpec
 
-/** Specifies the direction children are stacked in. */
+/** Specifies the direction children are stacked in. It becomes immutable after either horizontal or vertical alignment was set */
```
appleguy: Can we change the behavior so this property can only be set once?  Then the comment would also be more clear: "The stack direction may only be set once, and is then immutable" (or could be even shorter).

I think there is no valid use case for changing the stack direction, and this restriction helps ensure the correctness of the other features with less complexity.

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
```diff
@@ -48,6 +48,12 @@
 
 - (instancetype)init;
 
+/** Specifies how children are aligned horizontally. The stack direction will become immutable after this call. */
+- (void)setHorizontalAlignment:(ASHorizontalAlignment)alignment;
```
appleguy: These should be a property, if possible.  Although it is not likely to be a common use case to read the value, I think the same is true of alignItems / justifyContent, and the stylistic ability to use the property syntax to set the value will help the initialization of stacks look neater.

We should format the options near, but above, the alignItems / justifyContent properties and try to make it visually (and in description) clear that 1) these options are generally preferred to the others, and 2) if used, they supersede / replace the others, and 3) setting either of them will impact the other.

Implementing that 3) aspect would increase the complexity of this, so I am actually happy to land this diff without that change, but wanted to call it out as critical for the final 2.0.  @rcancro can do that piece if you're not available, @nguyenhuy - just let us know!

AsyncDisplayKitTests/ASStackLayoutSpecSnapshotTests.mm
```diff
@@ -63,6 +65,24 @@ - (void)testStackLayoutSpecWithJustify:(ASStackLayoutJustifyContent)justify
   [self testStackLayoutSpecWithStyle:style sizeRange:sizeRange subnodes:subnodes identifier:identifier];
 }
 
+- (void)testStackLayoutSpecWithDirection:(ASStackLayoutDirection)direction
```
appleguy: As always, thank you for implementing tests!  It is extra effort — especially to create snapshot tests — but truly valuable to the community and sets a great example.

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
nguyenhuy: While I agree that there is no valid use case for changing the stack direction, implementing the restriction actually does not ensure the correctness of alignments, because any number of direction changes would cause problem if they are done after alignments were set. Please check [this new test case](https://github.com/facebook/AsyncDisplayKit/pull/873/files#diff-2e90ad10896b4231329ca61a0e743c63R567) if you couldn't understand the above (poorly constructed) sentence.

nguyenhuy: And since the restriction is not related to the problem at hand, I think it deserves another PR.

nguyenhuy:
I've pushed a new diff that I think is very close to a landing state. Some new improvements compare to the old diff:
- After a direction change (including the first time), horizontal and vertical alignments are resolved again. Thus they always remain valid.
- Assertions are added to ensure that once users opt in for alignments, they can no longer set alignItems and justifyContent directly.
- More test cases are added to verify above behaviours.

AsyncDisplayKit/Layout/ASStackLayoutDefines.h
nguyenhuy: Done :)

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
nguyenhuy: Sorry but I don't get what you meant by 3). Could you please elaborate it?

Btw, other comments were addressed.

AsyncDisplayKit/Layout/ASStackLayoutSpec.h
```diff
@@ -35,10 +35,25 @@
...
+ Specifies the direction children are stacked in. If horizontalAlignment and verticalAlignment were set, 
+ they will be resolved again, causing justifyContent and alignItems to be updated accordingly
+ */
 @property (nonatomic, assign) ASStackLayoutDirection direction;
 /** The amount of space between each child. */
 @property (nonatomic, assign) CGFloat spacing;
+/** 
+ Specifies how children are aligned horizontally. Depends on the stack direction, setting the alignment causes either
+ justifyContent or alignItems to be updated. The alignment will remain valid after future direction changes.
+ Thus, it is preferred to those properties
```
appleguy: Very minor - this sentence wasn't completed.  Let's patch it up in a future diff (it doesn't have to be specific to this, could even be a change to an example app).  Is this format readable by AppleDoc?

appleguy:
@nguyenhuy extremely excited about this patch - thank you!

PR #879, opened 2015-11-23
Add custom collection view layout example project

levi:
Wrote a simple Pinterest-style custom CollectionViewLayout and integrated it with ASCollectionView in a sample project to show how to use the layout inspector API. The layout supports supplementary header views, which receive a constrained size from the custom layout inspector.

This patch also introduces the ability for the layout inspector to define the constrained size of the collection view item. By default, ASCollectionView measures its item constrained size by some heuristics that make the assumption that all cells should be measured with flow layout behavior. From the perspective of UICollectionView, all items receive a constrained size from the UICollectionViewLayoutAttributes#frame property in the layout and this behavior can vary in a custom layout implementation the one in the sample project. 

To increase the flexibility of the constrained size, the layout inspector is now consulted for the size of the item at the given index path. In the included example, the custom inspector queries the internals of the layout for the size. In the case of a regular flow layout, the default behavior of measuring based on the scroll direction still remains. In a future patch, I would like to move the logic of this behavior in the flow layout inspector (left a few TODOs in the source). However, the logic also generically supports defining an item size for non-flow layouts. I like this idea of having the support for custom non-supplementary view custom layouts without the need to implement extra APIs and I want to take a deeper look into retaining this behavior when cleaning up the delegation.


smiLLe:
Why is MosaicCollectionViewLayout just in example project? Wouldn't it make sense to have this layout in ASDK since it is a common layout these days?

levi:
It's a pretty simple layout implementation that doesn't have any sort robust supplementary view support or optimized lookup of elements. IMO it's up to the developer using ASDK to choose what kind custom layout implementation to use and there are plenty of existing open source versions that do a better job than this layout. Implementing a custom layout is at the advanced end of the spectrum of ASDK and this example is to simply give a layout author the taste of how to do that with a common layout problem that's well understood amongst UI developers.

Just noticed the failing test here while writing this. Looks like an issue with Travis. Going to rebase to fix it.

appleguy:
@smiLLe - you raise a reasonable point.  I have debated with myself about exactly what you're talking about — should ASDK include things such as UICollectionViewLayouts?

One reason it is not an obvious fit is that ASDK currently supports standard UICollectionViewLayouts, without any modification for ASDK (there is no special AS-subclass).  This means that a variety of Cocoapods can be directly imported, and even without modification, they can be kept up to date by following the version changes of the cocoa pod.  That's a compelling advantage.

@levi is also absolutely right that efficiently doing this layout is more work than we have here, and there are libraries to do it.

Perhaps most importantly, I think only about 10% or even less of ASDK apps would use this particular layout.  Though a small impact, the compiled size of a complex, efficient layout may be several dozen KB and would increase the build time of the framework for everyone, so it's something to think about.

All that said...I am thinking we will create a custom AS-based layout class soon, because it will provide important efficiency advantages over UICollectionViewLayout.  In that world, it becomes more interesting to include a library of built-in layouts.  I appreciate very much having your vote on this as it really does influence what we decide to do here.

Note that aside from cocoapods you can search for, this PR was put up a while back that has some pretty hardcore code to do this, and looks like it might be very efficient.  It would probably need some adaptation to use, but may form the base of a future built-in layout.  Thanks @xujim for writing it!!  https://github.com/facebook/AsyncDisplayKit/pull/435

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -664,7 +664,14 @@ - (ASCellNode *)dataController:(ASDataController *)dataController nodeAtIndexPat
 
 - (ASSizeRange)dataController:(ASDataController *)dataController constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
 {
-  ASSizeRange constrainedSize;
+  ASSizeRange constrainedSize = [_layoutInspector collectionView:self constrainedSizeForNodeAtIndexPath:indexPath];
```
appleguy: It is possible for _layoutInspector to be nil, right?  Quite subtly, it is actually not safe to message nil for non-scalar return types.  Though it nearly always returns a properly-zero'd struct in the simulator, it is prone to returning garbage data in some fields when run on the device.  This is one of the most common causes of device-only crashers and other weird behaviors that can't be reproduced in the simulator.

examples/CustomCollectionView/Sample/MosaicCollectionViewLayout.m
```diff
@@ -0,0 +1,230 @@
...
+  NSMutableArray *_itemAttributes;
+  NSMutableDictionary *_headerAttributes;
+  NSMutableArray *_allAttributes;
+}
+
+- (instancetype)init
+{
+  self = [super init];
+  if (self != nil) {
+    self.numberOfColumns = 3;
```
appleguy: Later, we might use UIDevice or UIScreen to adjust the number of columns depending on device and/or orientation.

examples/CustomCollectionView/Sample/MosaicCollectionViewLayout.m
```diff
@@ -0,0 +1,230 @@
...
+{
+  CGFloat height = [[[_columnHeights lastObject] firstObject] floatValue];
+  return CGSizeMake(self.collectionView.bounds.size.width, height);
+}
+
+- (NSUInteger)_tallestColumnIndexInSection:(NSUInteger)section
+{
+  __block NSUInteger index = 0;
+  __block CGFloat tallestHeight = 0;
+  [_columnHeights[section] enumerateObjectsUsingBlock:^(NSNumber *height, NSUInteger idx, BOOL *stop) {
```
appleguy: Though this class is far from optimized, I'd like to start preferring the fast enumeration "for ... in" style; it's a bit more efficient than using the block.  ASDataController has a lot of block-style enumeration to remove too.

examples/CustomCollectionView/Sample/SupplementaryNode.h
```diff
@@ -0,0 +1,18 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * FACEBOOK BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
+ * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#import <AsyncDisplayKit/AsyncDisplayKit.h>
+
+@interface SupplementaryNode : ASCellNode
```
appleguy: You may be able to use ASTextCellNode for this.  If you want or need the custom class, suggest naming it SupplementaryTextNode or something less generic.

appleguy:
@levi looks really great!  The only thing preventing landing is the nil messaging with structure return; very quick fix to check, could use the ternary operator or a one-line if with pre declared local variable.

I really need to figure out how to add a diff to someone else's PR, but somehow each time I've tried to do this it has just created problems and not actually worked LOL.

AsyncDisplayKit/ASCollectionView.mm
levi: Good catch

examples/CustomCollectionView/Sample/MosaicCollectionViewLayout.m
levi: Agreed, but I would like to optimize this layout for readability since it's just an example. Can use fast enumeration here, but `enumerateObjectsUsingBlock` is just so damn pretty :smile:

examples/CustomCollectionView/Sample/SupplementaryNode.h
levi: Good idea. Will need to extract the `kVerticalPadding` and `kHorizontalPadding` attributes into an `ASTextCellNode` ivar before moving in this direction.

appleguy:
@levi - thanks for the fix!  Although ironically, I think we still need to set the size to something in the local variable declaration - only Objective-C ivars are guaranteed to be zero'd memory, as far as I have known. 

It is possible that a change in LLVM or some other part of iOS may have made this safe, but I bet the static analyzer would warn.

appleguy:
screen shot 2015-11-29 at 9 27 05 pm

levi:
Good catch! Swift has made me lazy :sweat_smile:

appleguy:
@levi one last thing! :-D

/Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKit/ASCollectionView.mm:673:15: error: no viable conversion from 'const CGSize' to 'ASSizeRange'
  ASSizeRange constrainedSize = CGSizeZero;

levi:
Ouch!

appleguy:
So glad to merge this - it is one of our best example projects!  We will definitely be able to extend it for even greater educational value, as well as testbed utility for developing the framework.  I wonder if we could build integration tests with it in the future.

PR #881, opened 2015-11-23
[Critical] Don't Cancel Image Downloads in ASMultiplexImageNode.clearContents

Adlai-Holler:
Before this PR, if the node leaves the render range, but stays in the preload range, its image download will be canceled and another won't be enqueued so the user may see a lower-quality image or no image.


appleguy:
Thanks for the fix!  This code predated the network range, which was added early in 2015, and quite clearly was not updated correctly as there weren't many clients of ASMultiplexImageNode using the latest ASDK at the time (Paper did not have a network range).

My ASInterfaceState project has been paused for a week, but I'm picking it up this week and it will involve auditing the current clear / fetch methods.  Fortunately for me, you may have just fixed the most serious issue in those methods, as there aren't that many of them in the framework!

Adlai-Holler:
Awesome, thanks for the quick merge! Really glad to have this one in the bag!

PR #894, opened 2015-11-30
Use ASInterfaceState to Fetch Image Node Data when Image Source Changes

Adlai-Holler:
The first commit is the actual relevant change. The second commit is for consistency with the rest of the framework plus mutexes are more flexible than spin locks. The third commit fixes two issues 1) we shouldn't copy the array in the getter and 2) we should deep copy the array in the setter.

This is pursuant to the mention in #850 @appleguy

### DIFF
--- a/AsyncDisplayKit/ASMultiplexImageNode.mm
+++ b/AsyncDisplayKit/ASMultiplexImageNode.mm
@@ -63,7 +63,7 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
 
   // Image flags.
   BOOL _downloadsIntermediateImages; // Defaults to NO.
-  OSSpinLock _imageIdentifiersLock;
+  ASDN::Mutex _imageIdentifiersLock;
   NSArray *_imageIdentifiers;
   id _loadedImageIdentifier;
   id _loadingImageIdentifier;
@@ -270,23 +270,21 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
 
 - (NSArray *)imageIdentifiers
 {
-  OSSpinLockLock(&_imageIdentifiersLock);
-  NSArray *imageIdentifiers = [_imageIdentifiers copy];
-  OSSpinLockUnlock(&_imageIdentifiersLock);
-  return imageIdentifiers;
+  ASDN::MutexLocker l(_imageIdentifiersLock);
+  return [_imageIdentifiers copy];
 }
 
 - (void)setImageIdentifiers:(NSArray *)imageIdentifiers
 {
-  OSSpinLockLock(&_imageIdentifiersLock);
+  ASDN::MutexLocker l(_imageIdentifiersLock);
 
   if (ASObjectIsEqual(_imageIdentifiers, imageIdentifiers)) {
-    OSSpinLockUnlock(&_imageIdentifiersLock);
     return;
   }
-  
-  _imageIdentifiers = [imageIdentifiers copy];
-  OSSpinLockUnlock(&_imageIdentifiersLock);
+
+  _imageIdentifiers = [_imageIdentifiers copy];
+  _imageIdentifiersLock.unlock();
+
   if (self.interfaceState & ASInterfaceStateFetchData) {
     [self fetchData];
   }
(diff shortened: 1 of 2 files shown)


appleguy:
@Adlai-Holler woah, super fast - thanks!

BTW I now have a plan to quite robustly fix ASRangeController with relatively little work, reaching a stable state from which we can survey any longer term directions.  There are two key steps:

1. Finish the elimination of the workingWindow.  This patch already functions, but it needs testing and especially validation of display cancellation / multiple display calls before the previous display has completed.  This will make sure the ranges don't rely on external state that isn't controlled in the setInterfaceState: codepath.

2. Replace the -update method in ASRangeController to simply request the full set of index paths within each range, on every update.  Create a 4th set by taking the union of the 3 range sets (or N), and iterate over each indexPath.  For each, test if it exists in each of the individual ranges, and set the corresponding ASInterfaceState bitfield accordingly.  Finally, obtain the ASCellNode - and crucially - first check if cell.interfaceState != interfaceState.  Only if the root / cell itself doesn't match, then proceed to make a recursive call to setInterfaceState: on all children, each of which will take action only on the elements of the state which have changed.

This will resolve a great deal of issues, including the one you discovered related to existing indexPaths changing to different cells, because it will be up to the cells to understand / "remember" what states they have transitioned through and identify which states are changes or no-ops.

We will need one new method to go alongside enterInterfaceState: and exitInterfaceState: -- in fact, we should consider eliminating / making private those two methods.  The new one is basically just recursivelySetInterfaceState:, since with this model the controller won't actually know which states are being entered vs exited (but the node will).

The last detail here I haven't thought through is how to exit cells from /all/ ranges.  I think this could be pretty straightforward, although an efficient solution that doesn't require visiting / checking the interfaceState of every completedNode may be nontrivial.  I'd be supportive of an "inefficient" but highly robust solution as our next step, especially because even the inefficient solution is likely efficient enough since we're talking about a large number of individual property accesses with a single integer comparison.

Adlai-Holler:
Sounds right – using that functional-style "get all your new data" approach in ASRangeController seems like the best balance between speed, simplicity, and robustness. 

Cutting down the internal API to `recursivelySetInterfaceState:` sounds right, but long-term do we want to expose override points that correspond to `enter/exitInterfaceState:` to replace `fetchData`? If, say, `ASViewController` wanted to manage the interface state for its node hierarchy, would we want it to need to know that `ASInterfaceStateFetchData` -> `fetchData`? Not urgent at all but I want to float that idea.

As for exiting cells, do we want to have nodes reset their states in `dealloc`? As it stands today, multiplex image nodes don't cancel their image downloads on dealloc which means that we might be too harsh to demand that clients who write custom handlers for `fetchData`, or god forbid future custom working ranges, should bear the burden of duplicating their logic between ~`clearFetchedData` and `dealloc` =D.

What is the main situation you're concerned about when it comes to exiting cells from all ranges? If I do something `reloadDataImmediately`? If the table view `dealloc`s? I'm with you, though, that efficiency here isn't worth fretting over while we try to make other fundamental changes – doing some `rectForRowAtIndexPath:` math, even for a thousand nodes, wouldn't be catastrophic as a rare operation.

AsyncDisplayKit/ASMultiplexImageNode.mm
```diff
@@ -63,7 +63,7 @@ @interface ASMultiplexImageNode ()
 
   // Image flags.
   BOOL _downloadsIntermediateImages; // Defaults to NO.
-  OSSpinLock _imageIdentifiersLock;
```
appleguy: lol - didn't realize this had been implemented with a Spin Lock.  Kill that beast!

AsyncDisplayKit/ASMultiplexImageNode.mm
```diff
@@ -270,23 +270,25 @@ - (void)setDataSource:(id <ASMultiplexImageNodeDataSource>)dataSource
...
-  OSSpinLockUnlock(&_imageIdentifiersLock);
-  return imageIdentifiers;
+  ASDN::MutexLocker l(_imageIdentifiersLock);
+  return _imageIdentifiers;
 }
 
 - (void)setImageIdentifiers:(NSArray *)imageIdentifiers
 {
-  OSSpinLockLock(&_imageIdentifiersLock);
+  _imageIdentifiersLock.lock();
```
appleguy: I think ideally we would use the brace-bounding "visual scope" style that is typically used with MutexLocker in this case; then the explicit unlock before the return can also disappear.  (I assume it is important to unlock before calling -fetchData)

appleguy:
@Adlai-Holler going to land this now, and release 1.9.2 - unless you would suggest holding for anything?  I have several other changes planned, but I will be sparse until the weekend.

If master is good enough for 1.9.2, then I would land the CK text engine after branching, and start to make risky changes to the workingWindow / ASRangeController.  Ideally, this becomes "2.0 Beta" and may see one release as such before GM.  I've missed by previous November deadline, but there is an outside chance that December will work!

Adlai-Holler:
@appleguy Nice! Hold up for a minute while I put up this PR to use the lock scoping pattern.

PR #898, opened 2015-12-01
Reduce Frequency of beginUpdates/endUpdates Due to Node Relayout

Adlai-Holler:
@nguyenhuy @appleguy Experimenting with Kittens shows that when you enlarge the image this still correctly updates the table view, but when you swap the image and text it correctly does not bother the table view.

- Change delegate method `nodeDidRelayout:` to `nodeDidRelayoutWithSizeChange:` to filter layouts that don't affect height.
- In `ASTableView`, `ASCollectionView`, enqueue `beginUpdates()/endUpdates()` in the run loop. These calls can be expensive – querying 100 nodes for their heights and enqueuing an animation – so this should be worth the effort.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -15,13 +15,13 @@ typedef NSUInteger ASCellNodeAnimation;
 @protocol ASCellNodeLayoutDelegate <NSObject>
 
 /**
- * Notifies the delegate that the specified cell node has done a relayout
- * that resulted in a change of `calculatedSize`.
+ * Notifies the delegate that the specified cell node has done a relayout.
  * The notification is done on main thread.
  *
  * @param node A node informing the delegate about the relayout.
+ * @param sizeChanged `YES` if the node's `calculatedSize` changed during the relayout, `NO` otherwise.
  */
-- (void)nodeDidRelayoutWithSizeChange:(ASCellNode *)node;
+- (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged;
 @end
 
 /**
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASCellNode.h
```diff
@@ -15,12 +15,13 @@ typedef NSUInteger ASCellNodeAnimation;
...
 /**
- * Notifies the delegate that the specified cell node has done a relayout.
+ * Notifies the delegate that the specified cell node has done a relayout
+ * that resulted in a change of `calculatedSize`.
  * The notification is done on main thread.
  *
  * @param node A node informing the delegate about the relayout.
  */
-- (void)nodeDidRelayout:(ASCellNode *)node;
+- (void)nodeDidRelayoutWithSizeChange:(ASCellNode *)node;
```
levi: I would suggest leaving this naming as `didRelayoutNode:`, as it can be used by other future layout invalidations other than size changes. If you need specifics for your current usage of the API, might I suggest adding a boolean flag: `didRelayoutNode:sizeChange:`.

Adlai-Holler: Thanks, that's way better.

nguyenhuy:
@Adlai-Holler Thanks for bringing this up! It's absolutely true that after a relayout that doesn't result in a size change, the container view (i.e table view or collection view) doesn't need to be bothered. The reason it is still notified is that I want to make use of built-in reload animations (fb18e76). But after we replaced reloads with empty transactions in 5701cbb, that reason is no longer valid! Unnecessary empty transactions are inefficient. Sorry for triggering them in the first place :)

Adlai-Holler:
Ah! That makes sense @nguyenhuy. To be clear, this hasn't become a practical issue for me yet but I do it so I can sleep better =)

appleguy:
Let us all achieve deeper sleep in contemplation of new efficiencies. <3

Adlai-Holler:
Amen. :smile:

PR #901, opened 2015-12-01
ASTableView: Use Fancy dequeueReusableCellWithIdentifier:forIndexPath: Method

Adlai-Holler:
UITableView claims that `this method uses the index path to perform additional configuration based on the cell’s position in the table view.` I reckon it can't hurt to pass that information along. It may save us some view hierarchy swapping if it makes it more likely UITableView will give us the cell previously assigned to that index path. Plus it's less code and more consistent. So fresh and so clean, clean.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -22,6 +22,7 @@
 
 static const NSUInteger kASCollectionViewAnimationNone = UITableViewRowAnimationNone;
 static const ASSizeRange kInvalidSizeRange = {CGSizeZero, CGSizeZero};
+static NSString * const reuseIdentifier = @"_ASCollectionViewCell";
 
 #pragma mark -
 #pragma mark Proxying.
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -22,6 +22,7 @@
 
 static const NSUInteger kASCollectionViewAnimationNone = UITableViewRowAnimationNone;
 static const ASSizeRange kInvalidSizeRange = {CGSizeZero, CGSizeZero};
+static NSString * const reuseIdentifier = @"_ASCollectionViewCell";
```
appleguy: We should name this with a leading "k" or at least a capital letter to signify its special status as a constant.

appleguy:
Curious, I didn't know about this method — seems good to me!

Adlai-Holler:
Scott doesn't know about a UIKit method?! Well I'll be. Glad you dig it, and great point about the name. Fixed.

appleguy:
:-D there's also a whole world of weird and (wonderful?) autolayout-related methods that @rcancro taught me about.  We got useful name inspiration from them, for things I'd never heard of like preferredFrameSize!

PR #910, opened 2015-12-03
Allow User to Move While Touching Link As Long As They Stay In Range

Adlai-Holler:
Before we were canceling the highlight as soon as we got `touchesMoved:` but some of our users are sloppy organic humanoids, and they couldn't figure out why taps weren't working. :smiley:


appleguy:
Cool beans!

Issue #907, opened 2015-12-03
-[ASTextNode linkAttributeValueAtPoint:] Interacts with -shouldHighlight: Delegate Method


Adlai-Holler:
I called `linkAttributeValueAtPoint` but was always getting return value `nil`. On checking into the implementation, I saw that the attribute is being detected but then discarded because I don't implement 
the `shouldHighlightLinkAttribute` method. That doesn't seem quite right. Here's the code in ASTextNode:

```objectivec
// inside -[ASTextNode linkAttributeValueAtPoint:]
for (NSString *attributeName in _linkAttributeNames) {
  NSRange range;
  id value = [attributedString attribute:attributeName atIndex:characterIndex longestEffectiveRange:&range inRange:visibleRange];
  NSString *name = attributeName;

  if (value == nil || name == nil) {
    // Didn't find anything
    continue;
  }

  // Check if delegate implements optional method, if not assume NO.
  // Should the text be highlightable/touchable?
  if (![_delegate respondsToSelector:@selector(textNode:shouldHighlightLinkAttribute:value:atPoint:)] ||
      ![_delegate textNode:self shouldHighlightLinkAttribute:name value:value atPoint:point]) {
    value = nil;
    name = nil;
  }
```

appleguy:
@Adlai-Holler Interesting.  Seems a bit silly.  This is probably a vestigial behavior from Paper where, perhaps, there were some views that may have linkified text returned from the server but we didn't want any part of the gesture system to react to this.

Do you think the fix is as simple as deleting the if() at the end of the quoted code segment?  That's the fix I would suggest, but this is an area I am not deeply familiar with in the codebase.  @bdaz did work in this area a long time ago, as did @ocrickard, but I think it's a longshot that they would know the answer so let's hypothesize a solution proactively!

ocrickard:
lol yeah, I vaguely remember something like @appleguy is describing here. I'd be in favor of just deleting the delegate method.

appleguy:
Thank you @Adlai-Holler !

PR #911, opened 2015-12-04
Reduce Reliance on -textNode:shouldHighlightLinkAttribute:value:atPoint:

Adlai-Holler:
This changes the default value to `YES`, which falls in line with many other `thing:shouldActWithThing:` patterns. Also, the text node no longer consults the delegate when handling external calls to `linkAttributeValueAtPoint:`.

I reckon this could be released immediately. Users who don't implement the method obviously aren't using highlighting.


Adlai-Holler:
Resolves #907.

appleguy:
Agree, this is certainly a better default.

PR #913, opened 2015-12-04
Provide Basic Interface State Support for Nodes Outside of Cells

### LINKED ISSUE #897: [ASInterfaceState] Regression: plain view-hosted nodes don't automatically fetch data anymore
I just updated the pod to the latest master to get the fix to the table view #884 and then started seeing some of my ASNetworkImage nodes that are directly in UIViews not displaying they just continue to show the defaultImage. The source of the error seems to be this change in `- (void)setURL:(NSURL *)URL resetToDefault:(BOOL)reset`

New code:
```
  if (self.interfaceState & ASInterfaceStateFetchData) {
    [self fetchData];
  }
```

Old code:
```
 if (self.nodeLoaded && self.layer.superlayer)
    [self _lazilyLoadImageIfNecessary];
```

Adlai-Holler:
This resolves #897. I added two test cases, and also posted a toy project https://github.com/Adlai-Holler/ASDKPlainNodeTest.

##### Changes

- Add new interface state `ASInterfaceStateInHierarchy` that unites all others. This is the interface state for non cell-hosted nodes that are in the window hierarchy. I don't like the name – I considered `Unknown` but that seems like none more than all, and I considered `All` but that's not very informative.

##### Questions

- What's the safest way to push the `inCellNode`ness down the hierarchy? It's sort of unsafe to walk up the node hierarchy to check during `willEnterHierarchy` and `didExitHierarchy`
- I added `checkViewHierarchy:` param to `_supernodeWithClass` but honestly I wish I had just removed that behavior altogether. We don't call `_supernodeWithClass` anywhere in our code. But I wanted to ask because I haven't thought through the possible benefits of walking up the view hierarchy.

@appleguy What do you think?

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -366,7 +366,7 @@
 
 
 /**
- * Called just before the view is added to a superview.
+ * Called just before the view is added to a window.
  */
 - (void)willEnterHierarchy ASDISPLAYNODE_REQUIRES_SUPER;
 
@@ -426,9 +426,13 @@
 @end
 
 @interface ASDisplayNode (ASDisplayNodePrivate)
-// This method has proven helpful in a few rare scenarios, similar to a category extension on UIView,
-// but it's considered private API for now and its use should not be encouraged.
-- (ASDisplayNode *)_supernodeWithClass:(Class)supernodeClass;
+/**
+ * This method has proven helpful in a few rare scenarios, similar to a category extension on UIView,
+ * but it's considered private API for now and its use should not be encouraged.
+ * @param checkViewHierarchy If YES, and no supernode can be found, method will walk up from `self.view` to find a supernode.
+ * If YES, this method must be called on the main thread and the node must not be layer-backed.
+ */
+- (ASDisplayNode *)_supernodeWithClass:(Class)supernodeClass checkViewHierarchy:(BOOL)checkViewHierarchy;
 
 // The two methods below will eventually be exposed, but their names are subject to change.
 /**
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -41,17 +41,31 @@ typedef void (^ASDisplayNodeDidLoadBlock)(ASDisplayNode *node);
 typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
 {
   /** The element is not predicted to be onscreen soon and preloading should not be performed */
-  ASInterfaceStateNone          = 1 << 0,
+  ASInterfaceStateNone          = 0,
   /** The element may be added to a view soon that could become visible.  Measure the layout, including size calculation. */
-  ASInterfaceStateMeasureLayout = 1 << 1,
+  ASInterfaceStateMeasureLayout = 1 << 0,
   /** The element is likely enough to come onscreen that disk and/or network data required for display should be fetched. */
-  ASInterfaceStateFetchData     = 1 << 2,
+  ASInterfaceStateFetchData     = 1 << 1,
   /** The element is very likely to become visible, and concurrent rendering should be executed for any -setNeedsDisplay. */
-  ASInterfaceStateDisplay       = 1 << 3,
+  ASInterfaceStateDisplay       = 1 << 2,
   /** The element is physically onscreen by at least 1 pixel.
    In practice, all other bit fields should also be set when this flag is set. */
-  ASInterfaceStateVisible       = 1 << 4,
+  ASInterfaceStateVisible       = 1 << 3,
 };
+
+/**
+ * Currently we only set `interfaceState` for
+ * nodes contained in table views or collection views.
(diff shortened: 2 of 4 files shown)


appleguy:
@Adlai-Holler I would like to introduce a new type, ASHierarchyState, that includes things such as inRasterizedTree or inCellNode (those latter names aren't exactly excellent, open to ideas on a more pattern-friendly nomenclature).

I think there would have to be a mechanism to recursively flow updates to the ASHierarchyState down the tree when there is a change, such as enabling or disabling rasterization.

This is actually my preference for how we resolve the deadlock, but I haven't yet studied it in enough detail to be confident this can be done without another possible deadlocking moment.  However, I think it can be.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -366,7 +366,7 @@
 
 
 /**
- * Called just before the view is added to a superview.
+ * Called just before the view is added to a window.
```
appleguy: Sadly I think the behavior of this method is subtle.  It actually varies a bit for layer-backed nodes, which rely on kCAOrderIn (CAAction) rather than the UIView-unique hierarchy notifications.  This needs better documentation, and in fact, I'm pretty sure a Github issue is filed for this purpose.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -426,9 +426,13 @@
...
-// This method has proven helpful in a few rare scenarios, similar to a category extension on UIView,
-// but it's considered private API for now and its use should not be encouraged.
-- (ASDisplayNode *)_supernodeWithClass:(Class)supernodeClass;
+/**
+ * This method has proven helpful in a few rare scenarios, similar to a category extension on UIView,
+ * but it's considered private API for now and its use should not be encouraged.
+ * @param checkViewHierarchy If YES, and no supernode can be found, method will walk up from `self.view` to find a supernode.
+ * If YES, this method must be called on the main thread and the node must not be layer-backed.
+ */
+- (ASDisplayNode *)_supernodeWithClass:(Class)supernodeClass checkViewHierarchy:(BOOL)checkViewHierarchy;
```
appleguy: Can we walk the layer hierarchy instead, or does that not achieve the same result?

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -41,18 +41,28 @@ typedef void (^ASDisplayNodeDidLoadBlock)(ASDisplayNode *node);
 typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
 {
   /** The element is not predicted to be onscreen soon and preloading should not be performed */
-  ASInterfaceStateNone          = 1 << 0,
+  ASInterfaceStateNone          = 0,
```
appleguy: Thanks, that's a bit better :)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1635,6 +1644,20 @@ - (void)recursivelyClearFetchedData
...
+/**
+ * We currently only set interface state on nodes
+ * in table/collection views. For other nodes, if they are
+ * in the hierarchy we return `Unknown`, otherwise we return `None`.
+ *
+ * TODO: Avoid traversing up node hierarchy due to possible deadlock.
+ * @see https://github.com/facebook/AsyncDisplayKit/issues/900
+ * Possible solution is to push `isInCellNode` state downward on `addSubnode`/`removeFromSupernode`.
+ */
+- (BOOL)supportsInterfaceState {
```
appleguy: This is good as a prototype implementation, but we indeed need to improve efficiency with something like ASHierarchyState to avoid the repeated recursion or measurable performance impact will occur.  I think we'll land as-is once I am able to start on that change, but it would be important to fix before releasing (or even having in master very long, definitely increases deadlock probability)

AsyncDisplayKitTests/ASDisplayNodeTests.m
```diff
@@ -1663,6 +1680,48 @@ - (void)testBackgroundColorOpaqueRelationshipNoLayer
   [self checkBackgroundColorOpaqueRelationshipWithViewLoaded:NO layerBacked:YES];
 }
 
+// Check that nodes who have no cell node (no range controller)
+// do get their `fetchData` called, and they do report
+// the fetch data interface state.
+- (void)testInterfaceStateForNonCellNode
```
appleguy: Awesome!!!  Thanks for testing!

appleguy:
I am happy to land this as soon as I can start on the deadlock alleviation measures, worst case should be tomorrow.

appleguy:
@Adlai-Holler I agree we need to rethink the name for the combined state, but this is easy to rename after 1.9.3 has stabilized some of the active issues.

I think _supernodeWithClass is used by internal FB extensions to ASDK that were broken out during the prep for open sourcing.  However, they're almost certainly irrelevant now.  Don't worry about this method for the moment and I will look into possible usages / consider removing it in the next couple months.

PR #914, opened 2015-12-05
Set _interfaceState Before Calling fetchData et al.

Adlai-Holler:


Adlai-Holler:
Simulator fail.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1643,14 +1643,20 @@ - (ASInterfaceState)interfaceState
 
 - (void)setInterfaceState:(ASInterfaceState)interfaceState
 {
-  ASDN::MutexLocker l(_propertyLock);
-  if (interfaceState != _interfaceState) {
-    if ((interfaceState & ASInterfaceStateMeasureLayout) != (_interfaceState & ASInterfaceStateMeasureLayout)) {
+  ASInterfaceState oldValue;
+  {
+    ASDN::MutexLocker l(_propertyLock);
+    oldValue = _interfaceState;
```
appleguy: Great idea to minimize the critical section.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1643,14 +1643,20 @@ - (ASInterfaceState)interfaceState
...
+    _interfaceState = interfaceState;
+  }
+
+  if (interfaceState != oldValue) {
```
appleguy: Small naming suggestion.  How about we name the argument newState, and the local variable oldState?  Probably a bit more clear than oldValue, and it would shorten the comparisons a little.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1670,15 +1676,14 @@ - (void)setInterfaceState:(ASInterfaceState)interfaceState
...
+    if ((interfaceState & ASInterfaceStateVisible) != (oldValue & ASInterfaceStateVisible)) {
       if (interfaceState & ASInterfaceStateVisible) {
         // Consider providing a -didBecomeVisible.
       } else {
         // Consider providing a -didBecomeInvisible.
       }
     }
-    
-    _interfaceState = interfaceState;
+
```
appleguy: Probably don't need this line.

appleguy:
Nice improvement.  Clearly it makes sense to have the new value available during subclass handling of the various methods called - as well as have the property lock free.

PR #915, opened 2015-12-06
[ASDisplayNode] Introduced ASHierarchyState, eliminating upward traversal and locking.

appleguy:
This fixes the recently-reported deadlock, at least to the extent of my testing so far (100% repro with @Adlai-Holler 's test app before, 0% after).

It also adds a new header, ASDisplayNode+FrameworkPrivate.h, to avoid over-usage of *Internal and *+Subclasses.  I cleaned up several such usages.  Having wanted to do this for quite some time, I'm pretty happy to have the file established and know there is more cathartic refactoring to come!

ASHierarchyState itself is obviously similar to ASInterfaceState, but currently has only two state types and does /not/ currently perform operations when the state changes.  Instead, the state is checked by several key codepaths.  I would like to consider expanding this in the future to integrate things like isInHierarchy, and possibly finer-grained definitions of that such as "in window" vs "in subtree / any hierarchy".  Would love ideas both on the current behavior and future possibilities for refactoring + unification around a feature such as this.

The main thing remaining to do here is implement some tests, especially including the combination of the Rasterization and RangeManaged options in the same tree, with runtime enabling + disabling of Rasterization at different levels of the hierarchy.  Should also test that RangeManaged is not set on an ASCellNode used in isolation, but is in both ASTableView and ASCollectionView.  If anyone wants to give this a shot, please do - we need improved test harnesses too, and it may be necessary to do that to get strong coverage of this particular set of behaviors.

I'd love to have a couple community reviewers on this diff before landing it, so please comment!  At the same time, I want to land it quickly and get some testing from at least one or two users who have seen the deadlock.  Lastly, I need to ship this with 1.9.3 by end-of-day tomorrow (Sunday) or I will fall behind during the week.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1585,7 +1585,7 @@ void recursivelyEnsureDisplayForLayer(CALayer *layer)
   ASDisplayNodeAssert(_flags.isEnteringHierarchy, @"You should never call -willEnterHierarchy directly. Appearance is automatically managed by ASDisplayNode");
   ASDisplayNodeAssert(!_flags.isExitingHierarchy, @"ASDisplayNode inconsistency. __enterHierarchy and __exitHierarchy are mutually exclusive");
 
-  if (![self supportsInterfaceState]) {
+  if (![self supportsRangeManagedInterfaceState]) {
     self.interfaceState = ASInterfaceStateInHierarchy;
   }
 }
@@ -1596,7 +1596,7 @@ void recursivelyEnsureDisplayForLayer(CALayer *layer)
   ASDisplayNodeAssert(_flags.isExitingHierarchy, @"You should never call -didExitHierarchy directly. Appearance is automatically managed by ASDisplayNode");
   ASDisplayNodeAssert(!_flags.isEnteringHierarchy, @"ASDisplayNode inconsistency. __enterHierarchy and __exitHierarchy are mutually exclusive");
   
-  if (![self supportsInterfaceState]) {
+  if (![self supportsRangeManagedInterfaceState]) {
     self.interfaceState = ASInterfaceStateNone;
   }
 }
@@ -1647,18 +1647,12 @@ void recursivelyEnsureDisplayForLayer(CALayer *layer)
 }
 
 /**
- * We currently only set interface state on nodes
- * in table/collection views. For other nodes, if they are
- * in the hierarchy we return `Unknown`, otherwise we return `None`.
- *
- * TODO: Avoid traversing up node hierarchy due to possible deadlock.
- * @see https://github.com/facebook/AsyncDisplayKit/issues/900
- * Possible solution is to push `isInCellNode` state downward on `addSubnode`/`removeFromSupernode`.
+ * We currently only set interface state on nodes in table/collection views. For other nodes, if they are
+ * in the hierarchy we enable all ASInterfaceState types with `ASInterfaceStateInHierarchy`, otherwise `None`.
  */
-- (BOOL)supportsInterfaceState
+- (BOOL)supportsRangeManagedInterfaceState
 {
-  return ([self isKindOfClass:ASCellNode.class]
-      || [self _supernodeWithClass:ASCellNode.class checkViewHierarchy:NO] != nil);
+  return (_hierarchyState & ASHierarchyStateRangeManaged);
 }
 
 - (ASInterfaceState)interfaceState
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -7,8 +7,9 @@
  */
 
 #import "ASDisplayNode.h"
-#import "ASDisplayNode+Subclasses.h"
 #import "ASDisplayNodeInternal.h"
+#import "ASDisplayNode+Subclasses.h"
+#import "ASDisplayNode+FrameworkPrivate.h"
 #import "ASLayoutOptionsPrivate.h"
 
 #import <objc/runtime.h>
@@ -37,6 +38,9 @@
 - (void)_staticInitialize;
 
 @end
+
+//#define LOG(...) NSLog(__VA_ARGS__)
+#define LOG(...)
 
 // Conditionally time these scopes to our debug ivars (only exist in debug/profile builds)
 #if TIME_DISPLAYNODE_OPS
@@ -323,17 +327,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 #pragma mark - Core
 
-- (void)__tearDown:(BOOL)tearDown subnodesOfNode:(ASDisplayNode *)node
-{
-  for (ASDisplayNode *subnode in node.subnodes) {
-    if (tearDown) {
-      [subnode __unloadNode];
-    } else {
-      [subnode __loadNode];
-    }
-  }
-}
-
 - (void)__unloadNode
 {
   ASDisplayNodeAssertThreadAffinity(self);
@@ -357,22 +350,9 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   [self layer];
 }
 
-- (ASDisplayNode *)__rasterizedContainerNode
-{
-  ASDisplayNode *node = self.supernode;
-  while (node) {
-    if (node.shouldRasterizeDescendants) {
-      return node;
-    }
-    node = node.supernode;
-  }
-  
-  return nil;
-}
-
 - (BOOL)__shouldLoadViewOrLayer
 {
-  return ![self __rasterizedContainerNode];
+  return !(_hierarchyState & ASHierarchyStateRasterized);
 }
 
 - (BOOL)__shouldSize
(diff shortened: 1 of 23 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1664,23 +1661,23 @@ - (ASInterfaceState)interfaceState
...
+- (void)setInterfaceState:(ASInterfaceState)newState
 {
-  ASInterfaceState oldValue;
+  ASInterfaceState oldState;
   {
     ASDN::MutexLocker l(_propertyLock);
-    oldValue = _interfaceState;
-    _interfaceState = interfaceState;
+    oldState = _interfaceState;
+    _interfaceState = newState;
```
nguyenhuy: This maybe a little bit better?

```objective-c
{
    if (_interfaceState == newState) {
        return;        
    }
    _interfaceState = newState;
}

// Continue the rest of the method
```

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1724,6 +1720,40 @@ - (void)exitInterfaceState:(ASInterfaceState)interfaceState
...
+{
+  ASHierarchyState oldState;
+  {
+    ASDN::MutexLocker l(_propertyLock);
+    oldState = _hierarchyState;
+    _hierarchyState = newState;
+  }
+  
+  if (newState != oldState) {
+    LOG(@"setHierarchyState: oldState = %lu, newState = %lu", (unsigned long)oldState, (unsigned long)newState);
```
nguyenhuy: Same as in `setInterfaceState`, although in this case there is not much difference because this block is short and clean.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -969,7 +963,7 @@ - (void)_insertSubnode:(ASDisplayNode *)subnode atSubnodeIndex:(NSInteger)subnod
   [_subnodes insertObject:subnode atIndex:subnodeIndex];
 
   // Don't bother inserting the view/layer if in a rasterized subtree, becuase there are no layers in the hierarchy and none of this could possibly work.
```
levi: Typo: "because"

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -84,7 +85,7 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
     return;
   }
     
-  BOOL rasterizingFromAscendent = [self __rasterizedContainerNode] != nil;
+  BOOL rasterizingFromAscendent = (_hierarchyState & ASHierarchyStateRasterized);
 
   // if super node is rasterizing descendents, subnodes will not have had layout calls becase they don't have layers
```
levi: Typo: `because`

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1345,7 +1339,15 @@ - (ASDisplayNode *)_deallocSafeSupernode
...
 {
   ASDN::MutexLocker l(_propertyLock);
-  _supernode = supernode;
+  if (_supernode != supernode) {
+    ASHierarchyState oldHierarchyState = _supernode.hierarchyState;
+    _supernode = supernode;
+    if (_supernode) {
+      [self enterHierarchyState:_supernode.hierarchyState];
+    } else {
+      [self exitHierarchyState:oldHierarchyState];
```
Adlai-Holler: Maybe it's worth releasing the lock before changing the hierarchy state?

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
nguyenhuy: Eagle eyes!

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1645,17 +1647,12 @@ - (void)recursivelyClearFetchedData
...
- * TODO: Avoid traversing up node hierarchy due to possible deadlock.
- * @see https://github.com/facebook/AsyncDisplayKit/issues/900
- * Possible solution is to push `isInCellNode` state downward on `addSubnode`/`removeFromSupernode`.
+ * We currently only set interface state on nodes in table/collection views. For other nodes, if they are
+ * in the hierarchy we enable all ASInterfaceState types with `ASInterfaceStateInHierarchy`, otherwise `None`.
  */
-- (BOOL)supportsInterfaceState {
-  return ([self isKindOfClass:ASCellNode.class]
-      || [self _supernodeWithClass:ASCellNode.class checkViewHierarchy:NO] != nil);
+- (BOOL)supportsRangeManagedInterfaceState
```
levi: What's the future plan for this method? Will it be removed and all nodes support interface state (considering that range managed will be one part of the interface state feature set)?

AsyncDisplayKit/Private/ASDisplayNode+FrameworkPrivate.h
```diff
@@ -0,0 +1,109 @@
...
+{
+  /** The node may or may not have a supernode, but no supernode has a special hierarchy-influencing option enabled. */
+  ASHierarchyStateNormal       = 0,
+  /** The node has a supernode with .shouldRasterizeDescendants = YES.
+      Note: the root node of the rasterized subtree (the one with the property set on it) will NOT have this state set. */
+  ASHierarchyStateRasterized   = 1 << 0,
+  /** The node or one of its supernodes is managed by a class like ASRangeController.  Most commonly, these nodes are
+      ASCellNode objects or a subnode of one, and are used in ASTableView or ASCollectionView.
+      These nodes also recieve regular updates to the .interfaceState property with more detailed status information. */
+  ASHierarchyStateRangeManaged = 1 << 1,
```
levi: When ASViewController sends interface state changes, will you add another bit option or rename `RangeManaged` to be more high level?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -645,28 +625,42 @@ - (BOOL)shouldRasterizeDescendants
...
 {
   ASDisplayNodeAssertThreadAffinity(self);
   ASDN::MutexLocker l(_propertyLock);
   
-  if (_flags.shouldRasterizeDescendants == flag)
+  if (_flags.shouldRasterizeDescendants == shouldRasterize)
     return;
   
-  _flags.shouldRasterizeDescendants = flag;
+  _flags.shouldRasterizeDescendants = shouldRasterize;
```
Adlai-Holler: Maybe worth releasing the lock at this point? We call out to subclasses a lot after this point.

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: :+1:

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -645,28 +625,42 @@ - (BOOL)shouldRasterizeDescendants
...
-    }
     
-    [self recursivelyDisplayImmediately];
+    ASDisplayNodePerformBlockOnEverySubnode(self, ^(ASDisplayNode *node) {
+      if (shouldRasterize) {
+        [node enterHierarchyState:ASHierarchyStateRasterized];
+        [node __unloadNode];
+      } else {
+        [node exitHierarchyState:ASHierarchyStateRasterized];
+        [node __loadNode];
```
Adlai-Holler: This will only call `enterHierarchyState` if the node is already loaded. Shouldn't we call it no matter what?

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: Also two questions about this method:

appleguy:
@Adlai-Holler so I implemented this in such a way that ASCellNode will behave like a regular node if it is not actually ingested by the ASTableView/ASCollectionView machinery.  The idea is that ASCellNode is a convenient superclass for things that may be used both standalone and inside a table / collection.  I wouldn't necessarily recommend this widely, but if it can be supported easily, it seems like more intuitive behavior rather than mysteriously breaking things like image loading.

The upshot is that this test you added fails, as it does go into the InHierarchy state after being added to a view.  What do you think: if an interface state is set on the node manually, should that trigger us to avoid automatically manipulating the state?  Or should we "blow it away" with the automatic behavior?  I think blowing it away (and changing this assert to check for InHierarchy) may be the best option for now.

screen shot 2015-12-06 at 1 56 28 pm

Adlai-Holler:
@appleguy I totally agree, let's change the test. No need to be so strict with this.

appleguy:
@levi, @Adlai-Holler, @nguyenhuy thanks so much for the feedback!  Indeed you guys found a couple clear issues that reflect the fact that I hadn't yet seriously tested this, but by calling them out, you definitely saved me time and have bettered 1.9.3 because of it :).

I've been head-desk-banging a bit because the tests are still not passing.  In fact the output of the snapshot tests is blank for many ASLayoutSpec tests, even though ASLayoutSpec-based apps are running fine.  I haven't found an example app with a visible bug yet, which is strange, so I'm suspecting something about the highly artificial force-display mechanisms used by the test harnesses.  Still I don't want to make a change that breaks those, at least until they're working again (at which point I'd like to replace the hackily... method signature with the properly-supported -recursivelyEnsureDisplay or similar).

Let me know if you have any intuition on why layer snapshot comparisons are failing, but I will continue investigating.

AsyncDisplayKit/ASDisplayNode.mm
appleguy: @levi I think this will remain, although the name is probably unclear and needs to be cleaned up.  It will always be possible to add a node directly to a UIWindow or other non-ASDK hierarchy, and it seems to me like we'd either have to require developers to manually setInterfaceState: for even basic functionality or have something like this to enable some stuff automatically.

I may have missed your point and want to make sure to understand, because this area is not fully thought-through and your intuition / suggestions could be key.

For what it's worth, I do eventually want to enable interfaceState as a settable property (I think it is readonly right now) so that advanced custom users can set the different states manually.  Both to implement their own range controller-like patterns (hello, Paper!), but also simple to prompt a hierarchy to do async measurement or display ahead of insertion in the hierarchy.

AsyncDisplayKit/Private/ASDisplayNode+FrameworkPrivate.h
appleguy: Ah, very good question.  Yeah, I'm pretty unhappy with the name RangeManaged.  I think the best idea would be to rename it, making it more high level, but it's possible that there are reasons to do otherwise.  If we need the granularity, it may be best to add two new bits to specify those states but still have a high level one that indicates whether anything will be "driving" .interfaceState.

AsyncDisplayKit/ASDisplayNode.mm
appleguy: Yes - thanks!  I refactored this code a few times and ended up in a state that is incorrect.  Now fixed, although I don't like how it is factored, but I'm going to revisit it later (since it should work correctly now) to make it shorter / more elegant — running low on time tonight.

PR #917, opened 2015-12-07
Ensure that newly inserted section supplementary nodes are added to backing store

### LINKED ISSUE #893: [Supplementary Nodes] ASCollectionView crashed due to nil supplementary node?
I'm not sure why this is occurring, but when layoutSubviews is called on my ASCollectionViewController, it calls 
```objc 
- (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView viewForSupplementaryElementOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath
```
and the node that is provided from 
```objc
ASCellNode *node = [_dataController supplementaryNodeOfKind:kind atIndexPath:indexPath];
```
is nil. Therefore it is not stored in `_completedNodes`.

I've been scrutinizing this section of code in `ASCollectionDataController.mm`:

```objc
- (void)willInsertSections:(NSIndexSet *)sections
{
  [_pendingNodes enumerateKeysAndObjectsUsingBlock:^(NSString *kind, NSMutableArray *nodes, BOOL *stop) {
    NSMutableArray *sectionArray = [NSMutableArray arrayWithCapacity:sections.count];
    for (NSUInteger i = 0; i < sections.count; i++) {
      [sectionArray addObject:[NSMutableArray array]];
    }
    
    [self insertSections:sectionArray ofKind:kind atIndexSet:sections completion:nil];
    [self batchLayoutNodes:nodes ofKind:kind atIndexPaths:_pendingIndexPaths[kind] completion:nil];
    _pendingNodes[kind] = nil;
    _pendingIndexPaths[kind] = nil;
  }];
}
```

This is supposed to insert the pendingSections into the completed sections, however I may be missing something but the nodes never actually get added so I wonder if the completedNodes for the supplementary views ever get populated? It doesn't seem to be a problem with the rowNodes, only kind UICollectionElementKindSectionHeader at this point.

If this code works fine, then perhaps there is a race condition where layoutSubviews is being called when the completedNodes are being populated? This event seems to occur when I scroll to the bottom of my collection view, when the first UICollectionElementKindSectionHeader view is about to be displayed. It leads me to believe the supplementary view is not being correctly cached or I have a broken implementation.

levi:
Fixes #893

### DIFF
--- a/AsyncDisplayKit/Details/ASCollectionDataController.mm
+++ b/AsyncDisplayKit/Details/ASCollectionDataController.mm
@@ -28,22 +28,30 @@
   NSMutableDictionary *_pendingIndexPaths;
 }
 
+- (instancetype)initWithAsyncDataFetching:(BOOL)asyncDataFetchingEnabled
+{
+  self = [super initWithAsyncDataFetching:asyncDataFetchingEnabled];
+  if (self != nil) {
+    _pendingNodes = [NSMutableDictionary dictionary];
+    _pendingIndexPaths = [NSMutableDictionary dictionary];
+  }
+  return self;
+}
+
 - (void)prepareForReloadData
 {
-  _pendingNodes = [NSMutableDictionary dictionary];
-  _pendingIndexPaths = [NSMutableDictionary dictionary];
-
-  [[self supplementaryKinds] enumerateObjectsUsingBlock:^(NSString *kind, NSUInteger idx, BOOL *stop) {
+  NSArray *kinds = [self supplementaryKinds];
+  for (NSString *kind in kinds) {
     LOG(@"Populating elements of kind: %@", kind);
     NSMutableArray *indexPaths = [NSMutableArray array];
     NSMutableArray *nodes = [NSMutableArray array];
     [self _populateSupplementaryNodesOfKind:kind withMutableNodes:nodes mutableIndexPaths:indexPaths];
     _pendingNodes[kind] = nodes;
     _pendingIndexPaths[kind] = indexPaths;
-
+    
     // Measure loaded nodes before leaving the main thread
     [self layoutLoadedNodes:nodes ofKind:kind atIndexPaths:indexPaths];
-  }];
+  }
 }
 
 - (void)willReloadData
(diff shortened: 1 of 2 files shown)


appleguy:
Fantastic, thanks for the ultra fast turnaround @levi!  Getting this into 1.9.3 because of your Sunday night work.  Community dedication right there :)

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -28,22 +28,30 @@ @implementation ASCollectionDataController {
...
+  self = [super initWithAsyncDataFetching:asyncDataFetchingEnabled];
+  if (self != nil) {
+    _pendingNodes = [NSMutableDictionary dictionary];
+    _pendingIndexPaths = [NSMutableDictionary dictionary];
+  }
+  return self;
+}
 
-  [[self supplementaryKinds] enumerateObjectsUsingBlock:^(NSString *kind, NSUInteger idx, BOOL *stop) {
+- (void)prepareForReloadData
```
appleguy: Doesn't this potentially get called multiple times, and need to recreate the _pendingNodes and _pendingIndexPaths dictionaries?

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -28,22 +28,30 @@ @implementation ASCollectionDataController {
...
+{
+  NSArray *kinds = [self supplementaryKinds];
+  for (NSString *kind in kinds) {
```
appleguy: Thanks for using fast enumeration!  No need for block overhead.  In this case I think it would be OK (although not much better) to call [self supplementaryKinds] in the for (), as it will only be called once.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -97,25 +106,29 @@ - (void)willInsertSections:(NSIndexSet *)sections
...
     [self deleteSectionsOfKind:kind atIndexSet:sections completion:nil];
-  }];
+  }
 }
 
 - (void)prepareForReloadSections:(NSIndexSet *)sections
 {
-  [[self supplementaryKinds] enumerateObjectsUsingBlock:^(NSString *kind, NSUInteger idx, BOOL *stop) {
+  NSArray *kinds = [self supplementaryKinds];
+  for (NSString *kind in kinds) {
```
appleguy: Now that I see you did the whole file - which is badass!! - it probably would be better to save the lines and do the self call inline with the for ().

AsyncDisplayKit/Details/ASCollectionDataController.mm
levi: The pending dictionaries exist to keep the pending state between the `prepare` and `will` methods. I've set them up to be created at initialization such that they'll remain throughout the lifetime of the collection data controller. Their contents are cleared out at the end of the `will` methods.

AsyncDisplayKit/Details/ASCollectionDataController.mm
levi: Yup good idea.

PR #918, opened 2015-12-07
Inline supplementary kind iteration in data controller

levi:


appleguy:
Nice refinement, thanks!

PR #924, opened 2015-12-08
Improve ASMapNode to get ready for 2.0

aaronschubert0:
Hey @appleguy, here are changes based on your latest feedback
Notable changes:
- Changed init method to `-initWithRegion:` instead of `initWithCoordinate:`
- You can now call setLiveMap before ASMapNode has been loaded to start off with a live map view.
- Smaller syntax changes and clean up.

Just had a thought though, would it be useful for people to be able to change the region after the initialisation of ASMapNode, should we provide a method for this?

### DIFF
--- a/AsyncDisplayKit/ASMapNode.h
+++ b/AsyncDisplayKit/ASMapNode.h
@@ -11,7 +11,7 @@
 
 @interface ASMapNode : ASImageNode
 
-- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithRegion:(MKCoordinateRegion)region NS_DESIGNATED_INITIALIZER;
 
 /**
  This is the MKMapView that is the live map part of ASMapNode. This will be nil if .liveMap = NO. Note, MKMapView is *not* thread-safe.
--- a/AsyncDisplayKit/ASMapNode.h
+++ b/AsyncDisplayKit/ASMapNode.h
@@ -11,7 +11,10 @@
 
 @interface ASMapNode : ASImageNode
 
-- (instancetype)initWithRegion:(MKCoordinateRegion)region NS_DESIGNATED_INITIALIZER;
+/**
+ The current region of ASMapNode. This can be set at any time and ASMapNode will animate the change.
+ */
+@property (nonatomic, assign) MKCoordinateRegion region;
 
 /**
  This is the MKMapView that is the live map part of ASMapNode. This will be nil if .liveMap = NO. Note, MKMapView is *not* thread-safe.
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -36,11 +36,11 @@
   }
   self.backgroundColor = ASDisplayNodeDefaultPlaceholderColor();
   self.clipsToBounds = YES;
-
+  
   _needsMapReloadOnBoundsChange = YES;
   _liveMap = NO;
   _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
-
+  
   _options = [[MKMapSnapshotOptions alloc] init];
   _options.region = region;
   
@@ -49,29 +49,28 @@
 
 - (void)didLoad
 {
-    [super didLoad];
-    if (self.isLiveMap && !_mapNode) {
-        self.userInteractionEnabled = YES;
-        [self addLiveMap];
-    }
+  [super didLoad];
+  if (self.isLiveMap && !_mapNode) {
+    self.userInteractionEnabled = YES;
+    [self addLiveMap];
+  }
 }
 
 - (void)fetchData
 {
-    [super fetchData];
-    if (_liveMap && !_mapNode) {
-        [self addLiveMap];
-    }
-    else {
-        [self setUpSnapshotter];
-        [self takeSnapshot];
-    }
+  [super fetchData];
+  if (_liveMap && !_mapNode) {
+    [self addLiveMap];
+  } else {
+    [self setUpSnapshotter];
+    [self takeSnapshot];
+  }
 }
 
 - (void)clearFetchedData
 {
-    [super clearFetchedData];
-    [self removeLiveMap];
+  [super clearFetchedData];
+  [self removeLiveMap];
 }
 
 #pragma mark - Settings
@@ -148,11 +147,11 @@
 
 - (void)setUpSnapshotter
 {
-    if (!_snapshotter) {
-        ASDisplayNodeAssert(!CGSizeEqualToSize(CGSizeZero, self.calculatedSize), @"self.calculatedSize can not be zero. Make sure that you are setting a preferredFrameSize or wrapping ASMapNode in a ASRatioLayoutSpec or similar.");
-        _options.size = self.calculatedSize;
-        _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
-    }
+  if (!_snapshotter) {
+    ASDisplayNodeAssert(!CGSizeEqualToSize(CGSizeZero, self.calculatedSize), @"self.calculatedSize can not be zero. Make sure that you are setting a preferredFrameSize or wrapping ASMapNode in a ASRatioLayoutSpec or similar.");
+    _options.size = self.calculatedSize;
+    _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
+  }
 }
 
 - (void)resetSnapshotter
@@ -195,16 +194,16 @@
 
 - (void)setAnnotations:(NSArray *)annotations
 {
-    ASDN::MutexLocker l(_propertyLock);
-    _annotations = [annotations copy];
-    if (annotations.count != _annotations.count) {
-        // Redraw
-        [self setNeedsDisplay];
-        if (_mapView) {
-            [_mapView removeAnnotations:_mapView.annotations];
-            [_mapView addAnnotations:annotations];
-        }
+  ASDN::MutexLocker l(_propertyLock);
+  _annotations = [annotations copy];
+  if (annotations.count != _annotations.count) {
+    // Redraw
+    [self setNeedsDisplay];
+    if (_mapView) {
+      [_mapView removeAnnotations:_mapView.annotations];
+      [_mapView addAnnotations:annotations];
     }
+  }
 }
 
 
@@ -215,13 +214,12 @@
   [super layout];
   if (_mapView) {
     _mapView.frame = CGRectMake(0.0f, 0.0f, self.calculatedSize.width, self.calculatedSize.height);
-  }
-  else {
+  } else {
     // If our bounds.size is different from our current snapshot size, then let's request a new image from MKMapSnapshotter.
     if (!CGSizeEqualToSize(_options.size, self.bounds.size) && _needsMapReloadOnBoundsChange) {
-        _options.size = self.bounds.size;
-        [self resetSnapshotter];
-        [self takeSnapshot];
+      _options.size = self.bounds.size;
+      [self resetSnapshotter];
+      [self takeSnapshot];
     }
   }
 }
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -17,19 +17,18 @@
   MKMapSnapshotter *_snapshotter;
   MKMapSnapshotOptions *_options;
   NSArray *_annotations;
-  ASDisplayNode *_mapNode;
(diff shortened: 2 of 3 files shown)


appleguy:
@aaronschubert0 super awesome, thank you!  I was just thinking about this earlier in the morning.

It would indeed be ideal to proved a property for region, and allow it to be set.  More likely than anything else it may be used to read the value before the node is loaded, but it seems useful to be able to set it too.

I might even suggest that we kill initWithRegion entirely (I know, I know ... I had suggested it), and just rely on the property.  This way if a user's app doesn't know the region at init time (seems plausible), they could set it in -fetchData or wherever else is appropriate.

aaronschubert0:
@appleguy Hmm this is a tricky one. I've been thinking about this for a few days now. I'm not sure how comfortable I would be in having no custom init method that clearly specifies an action. If we did get rid of initWithRegion then we would have to provide a suitable default so that ASMapNode works out of the box. Although [[ASMapNode alloc] init] doesn't produce anything great either at the moment. I'm still 50/50. The layout spec (as well as ASImage/Text/ControlNode etc) conforms to a standard -init and then setting properties so maybe we should actually go this way to keep in line with the framework. Should we provide a default in this case?

appleguy:
@aaronschubert0 I think it's similar to UIView initWithFrame:, which actually is a bad practice / pattern to use, and so that's why ASDisplayNode does not have initWithFrame: even though it is basically a superset of the UIView API.

Let's go ahead and make it a plain -init with the region property.  Defaulting the region to "whole world" would make sense (not US, please!) - simply so that people creating a map node for the first time can see something show up if they don't have a region to set yet.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -41,30 +42,40 @@ - (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate
...
+- (void)fetchData
 {
-  if (!_snapshotter) {
-    ASDisplayNodeAssert(!CGSizeEqualToSize(CGSizeZero, self.calculatedSize), @"self.calculatedSize can not be zero. Make sure that you are setting a preferredFrameSize or wrapping ASMapNode in a ASRatioLayoutSpec or similar.");
-      _options.size = self.calculatedSize;
-      _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
-  }
+    [super fetchData];
+    if (_liveMap && !_mapNode) {
+        [self addLiveMap];
```
appleguy: Make sure all indentation throughout the file is 2 spaces, to match ASDK code style.  You can set this in your editor and then use the "re-indent" function if you select all in the file.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -150,23 +146,33 @@ - (void)takeSnapshot
...
 - (void)addLiveMap
 {
   if (self.isNodeLoaded && !_mapNode) {
-    _mapNode = [[ASDisplayNode alloc]initWithViewBlock:^UIView *{
-      _mapView = [[MKMapView alloc]initWithFrame:CGRectMake(0.0f, 0.0f, self.calculatedSize.width, self.calculatedSize.height)];
+    __weak ASMapNode *weakSelf = self;
+    _mapNode = [[ASDisplayNode alloc] initWithViewBlock:^UIView *{
+      _mapView = [[MKMapView alloc] initWithFrame:CGRectZero];
       _mapView.delegate = _mapDelegate;
       [_mapView setRegion:_options.region];
```
appleguy: I think there is still a retain cycle here, but probably best to test it with a sample app and Instruments allocation trace to make sure everything gets torn down correctly.  accessing ivars here, every one of them including _options, _mapDelegate, etc will create a strong retain of self.

Also, get rid of the reference to _mapView unless it is needed, and access this through the _mapNode.  I'm a little split on whether we should have _mapNode at all to be honest.

How about we just kill _mapNode and work with _mapView exclusively when dealing with the live map?

In any case, make sure to add ASDisplayNodeAssertMainThread() at the top of this method.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -185,9 +191,23 @@ - (void)removeLiveMap
...
     _mapNode = nil;
   }
-  self.image = nil;
 }
 
+- (void)setAnnotations:(NSArray *)annotations
+{
+    ASDN::MutexLocker l(_propertyLock);
+    _annotations = [annotations copy];
+    if (annotations.count != _annotations.count) {
```
appleguy: This condition could never trigger, right?  Since you are setting _annotations = annotations above?  Either way, comparing the count is not the same as isEqual:, because you could have two entirely different sets of annotations that happen to have the same number of them.

If we have a live map, we probably shouldn't call setNeedsDisplay.  Since the user may scroll around the _mapView the snapshot will become out of date anyway, so I think we shouldn't spend resources fetching map tiles and rendering them while the live map is active.

If _mapView is non-nil /if and only if/ we have live map mode set, maybe we should just get rid of the boolean entirely and have - (BOOL)isLiveMap return (_mapView != nil);   It's a bit strange to check if (_mapView) here where some places check live map.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -198,11 +218,10 @@ - (void)layout
   }
   else {
```
appleguy: } else {

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: @appleguy I've ran this through the kittens sample app and it doesn't seem to cause a retain cycle (not one I can spot anyway) Although this isn't an area of expertise for myself. I'd be happy to make all those ivars have a weak local reference but it doesn't look clean visual. Is this the route you would propose? 

In terms of _mapNode. I've decided to kill it and work with _mapView exclusively as you've suggested. Works better this way. I've also added that assert to the top

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: @appleguy I've rewritten this so that it works correctly, both when we have a liveMap and also when only the snapshot is being displayed.

In terms of getting rid of the boolean, I've got rid of _liveMap and now use self.isLiveMap exclusively, rather than mixing and matching _liveMap && !_mapView. It's a lot better that way, thanks for that!

aaronschubert0:
@appleguy I've made the specified changes. The major breaking change is getting rid of the custom init method in favour of altering the region property. Internally, I got rid of _mapNode to work directly with the MKMapView, change isLiveMap to be a 'computed' property and generally improved the code quality. The only snag left is the possibility of a retain cycle in the -addLiveMap method, I've commented on previous comments asking how this may be solved, happy to change anything regarding this in a follow up diff.

AsyncDisplayKit/ASMapNode.h
```diff
@@ -11,7 +11,10 @@
 
 @interface ASMapNode : ASImageNode
 
-- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate NS_DESIGNATED_INITIALIZER;
+/**
+ The current region of ASMapNode. This can be set at any time and ASMapNode will animate the change.
+ */
+@property (nonatomic, assign) MKCoordinateRegion region;
```
appleguy: Add to the comment - "this property may be set from a background thread before the node is loaded, and will automatically be applied to define the region of the static snapshot (if liveMap = NO) or the internal MKMapView (otherwise)."

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -17,70 +17,80 @@ @interface ASMapNode()
   MKMapSnapshotter *_snapshotter;
   MKMapSnapshotOptions *_options;
   NSArray *_annotations;
-  ASDisplayNode *_mapNode;
```
appleguy: This is significantly better - thank you for all the changes, especially removing / limiting state!

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -17,70 +17,80 @@ @interface ASMapNode()
...
-@synthesize liveMap = _liveMap;
 @synthesize needsMapReloadOnBoundsChange = _needsMapReloadOnBoundsChange;
 @synthesize mapDelegate = _mapDelegate;
+@synthesize region = _region;
 
-- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate
+#pragma mark - Lifecycle
+- (instancetype)init
 {
   if (!(self = [super init])) {
```
appleguy: Thanks for noticing / using the unusual init style of ASDK.  Some newer components don't use it, and it's not enforced, but I still prefer this one and use it for my code ;).

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -17,70 +17,80 @@ @interface ASMapNode()
...
   self.backgroundColor = ASDisplayNodeDefaultPlaceholderColor();
   self.clipsToBounds = YES;
-
+  
   _needsMapReloadOnBoundsChange = YES;
-  _liveMap = NO;
+  
   _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
-
+  _region = MKCoordinateRegionMake(CLLocationCoordinate2DMake(43.432858, 13.183671), MKCoordinateSpanMake(0.2, 0.2));
```
appleguy: Add a comment above this with a few-word description of where this is like "world-scale view including several continents" or something

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -17,70 +17,80 @@ @interface ASMapNode()
...
 {
-  if (!_snapshotter) {
-    ASDisplayNodeAssert(!CGSizeEqualToSize(CGSizeZero, self.calculatedSize), @"self.calculatedSize can not be zero. Make sure that you are setting a preferredFrameSize or wrapping ASMapNode in a ASRatioLayoutSpec or similar.");
-      _options.size = self.calculatedSize;
-      _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
+  [super fetchData];
+  if ([self wasLiveMapPreviously]) {
+    [self addLiveMap];
+  } else {
+    [self setUpSnapshotter];
```
appleguy: maybe just call -setUpSnapshotter from inside -takeSnapshot if it is necessary

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -17,70 +17,80 @@ @interface ASMapNode()
...
+    [self takeSnapshot];
   }
 }
 
+- (void)clearFetchedData
```
appleguy: rename this (and the superclass call) to -clearContents

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -17,70 +17,80 @@ @interface ASMapNode()
...
 {
-  ASDN::MutexLocker l(_propertyLock);
-  if (liveMap == _liveMap) {
-    return;
-  }
-  _liveMap = liveMap;
   liveMap ? [self addLiveMap] : [self removeLiveMap];
 }
 
+- (BOOL)wasLiveMapPreviously
```
appleguy: Hmm, the need for this might challenge my original thought of eliminating the liveMap BOOL...I think we may need it after all.  Most particularly, what happens if you allocate the map node (on a background thread as nodes are supposed to be initialized) and then immediately setLiveMap:YES, because you want your map node to always be live once it comes onscreen?  We shouldn't allocate the node then - it should be done AT EARLIEST in -fetchData - but this would require the prior implementation that you had with the _propertyLocked BOOL.

appleguy:
@aaronschubert0 this is really getting there!  Did we add a test case for this, at least an example app?  I wonder if we could add some unit tests to make sure coverage stays high and we don't break it?

I'm going to land this one because it looks like a definitive improvement / step forward.  I do think we should bring back the BOOL and look for any other ways to make sure the internal logic / state is consistently handled for all the possible ordering of calls and use cases that nodes can go through.  One followup there is correct handling of layerBacked mode — at a minimum we'd need to add the mapView's sublayer as a sublayer of the node's layer when the node is layerBacked, but it's very possible that won't work correctly.  Yet we should support layerBacking for the default snapshotter mode.  So I'd suggest overriding setLayerBacked: and setLiveMap: for each to check for the other property being set and throw an assertion if at any point they are both enabled.

aaronschubert0:
@appleguy Yeah!! So much better than when we started. I haven't added test cases yet but will add those along with an example app soon. Good point on the layer backing, can't forget about that! I'll add this in a future diff as well.

In terms of the BOOL, yeah I agree, it's tricky trying to minimise the internal state but will keep working on that and bringing back the BOOL does make it clearer.

Due to work I'll probably be picking this up in the new year, once we've had some rest.

PR #925, opened 2015-12-08
Implement simple horizontal paging node

levi:
This is a pretty simple implementation to get the ball rolling on the API design and control behavior. I'm currently developing this control in tandem at the NFL for an update on how we navigate between weeks in a given season. Let's get a discussion going on how we could develop this in a meaningful way for other groups to use.


AsyncDisplayKit/ASPagerNode.m
```diff
@@ -0,0 +1,72 @@
...
+@end
+
+@implementation ASPagerNode
+
+- (instancetype)init
+{
+  _flowLayout = [[UICollectionViewFlowLayout alloc] init];
+  _flowLayout.scrollDirection = UICollectionViewScrollDirectionHorizontal;
+  _flowLayout.minimumInteritemSpacing = 0;
+  _flowLayout.minimumLineSpacing = 0;
```
binl: We probably wanna expose things like this so people can get spacing between pages

appleguy:
@levi thanks for building this!  Several folks are quite excited to start using it, both at Pinterest and beyond :).

When you get a chance, can you fix the issue that the build server is flagging?

levi:
@appleguy added a few helper methods locally that I'd like to begin adding after this patch lands. You ok landing this?

AsyncDisplayKit/ASPagerNode.m
appleguy: Yes, definitely agreed.  Next changes!

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -0,0 +1,74 @@
...
+  
+  self = [super initWithCollectionViewLayout:_flowLayout];
+  if (self != nil) {
+    self.view.asyncDataSource = self;
```
appleguy: It is essential that we set these in -didLoad, not in -init.  Some big perf problems in Pinterest were addressed by ensuring we didn't poke the creation of ASCollectionViews too early.

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -0,0 +1,74 @@
...
+    self.view.asyncDelegate = self;
+    
+    self.view.pagingEnabled = YES;
+    self.view.allowsSelection = NO;
+    self.view.showsVerticalScrollIndicator = NO;
+    self.view.showsHorizontalScrollIndicator = NO;
+    
+    ASRangeTuningParameters tuningParams = { .leadingBufferScreenfuls = 1.0, .trailingBufferScreenfuls = 1.0 };
+    [self setTuningParameters:tuningParams forRangeType:ASLayoutRangeTypePreload];
```
appleguy: Cool, good idea to have new defaults for these.  However, please set the Preload range to 2.0 both leading and trailing.

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -0,0 +1,74 @@
...
+    [self setTuningParameters:tuningParams forRangeType:ASLayoutRangeTypeRender];
+  }
+  return self;
+}
+
+- (void)reloadData
```
appleguy: Maybe just implement this on ASCollectionNode?  Redeclaring it in the header for clarity so people don't have to look at the superclass is fine for now, but we should make sure that anything relevant to ASCollectionNode is put there.  Same is probably true for setTuningParameters.

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -0,0 +1,74 @@
...
+{
+  [self.view reloadData];
+}
+
+- (void)setTuningParameters:(ASRangeTuningParameters)tuningParameters forRangeType:(ASLayoutRangeType)rangeType
```
appleguy: It's not obvious that this will trigger view creation.  Suggest putting it on ASCollectionNode - and creating a new data type similar to ASPendingViewState that both ASCollectionNode and/or ASTableNode (possibly different data types for each) can use to store properties like the delegate / dataSource / tuningParameters before the view is created, and apply them in -didLoad.

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -0,0 +1,74 @@
...
+
+- (NSInteger)collectionView:(UICollectionView *)collectionView numberOfItemsInSection:(NSInteger)section
+{
+  ASDisplayNodeAssert(self.dataSource != nil, @"ASPagerNode must have a data source to load paging nodes");
+  return [self.dataSource numberOfPagesInPagerNode:self];
+}
+
+- (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
+{
+  return ASSizeRangeMake(CGSizeZero, self.view.bounds.size);
```
appleguy: It might actually make more sense for the minimum size to also be the bounds size, because we probably don't want pages laying out at a size less than the page, even if they have to flex to add more space.

appleguy:
This is a great first start, thanks @levi !  If you can address comments in a followup diff, that would be awesome and almost surely enough for me to adopt this in Pinterest this weekend!

AsyncDisplayKit/ASPagerNode.m
levi: Totally agree with this suggestion. I think we need to introduce a broader data type which allows any custom node to defer application of a value until the view has loaded. I think we could go about this by upgrading ASPendingViewState to support custom properties.

AsyncDisplayKit/ASPagerNode.m
levi: I wanted to keep the restriction of having one node per a page a loose one. With an undefined minimum size, it keeps the possibility open of showing pages that have multiple nodes.

PR #926, opened 2015-12-08
Adds ASTextNode vertical text alignment

nivanchikov:
This pull request adds an option to align text vertically in ASTextNode. May be useful when creating a button node (since ASTextNode is a subclass of ASControlNode), when button title should be centered both vertically and horizontally.

Supported alignment options: top, bottom, center

Screenshot below showcases the updated example from ASCollectionView demo app
img_1091

### DIFF
--- a/AsyncDisplayKit/ASTextNode.h
+++ b/AsyncDisplayKit/ASTextNode.h
@@ -26,6 +26,26 @@ typedef NS_ENUM(NSUInteger, ASTextNodeHighlightStyle) {
 };
 
 /**
+ * The vertical alignment of text within a node.
+ */
+typedef NS_ENUM(NSUInteger, ASTextNodeVerticalTextAlignment) {
+  /**
+   * Aligns the content vertically at the top in the node (the default).
+   */
+  ASTextNodeVerticalTextAlignmentTop,
+  
+  /**
+   * Aligns the content vertically in the center of the node.
+   */
+  ASTextNodeVerticalTextAlignmentCenter,
+  
+  /**
+   * Aligns the content vertically at the bottom in the node.
+   */
+  ASTextNodeVerticalTextAlignmentBottom
+};
+
+/**
  @abstract Draws interactive rich text.
  @discussion Backed by TextKit.
  */
@@ -69,6 +89,12 @@ typedef NS_ENUM(NSUInteger, ASTextNodeHighlightStyle) {
  @default 0 (No limit)
  */
 @property (nonatomic, assign) NSUInteger maximumNumberOfLines;
+
+/**
+ @abstract The vertical alignment of text within the receiver.
+ @discussion Defaults to ASTextNodeVerticalTextAlignmentTop
+ */
+@property (nonatomic, assign) ASTextNodeVerticalTextAlignment verticalTextAlignment;
 
 /**
  @abstract The number of lines in the text. Text must have been sized first.
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -114,6 +114,8 @@ static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncation
 
     _truncationMode = NSLineBreakByWordWrapping;
     _composedTruncationString = DefaultTruncationAttributedString();
+    
+    _verticalTextAlignment = ASTextNodeVerticalTextAlignmentTop;
 
     // The common case is for a text node to be non-opaque and blended over some background.
     self.opaque = NO;
@@ -380,6 +382,18 @@ static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncation
   // Offset the text origin by any shadow padding
   UIEdgeInsets shadowPadding = [self shadowPadding];
   CGPoint textOrigin = CGPointMake(self.bounds.origin.x - shadowPadding.left, self.bounds.origin.y - shadowPadding.top);
+  
+  switch (_verticalTextAlignment)
+  {
+    case ASTextNodeVerticalTextAlignmentCenter:
+      textOrigin.y += (self.bounds.size.height - [self _renderer].size.height) / 2.0f;
+      break;
+    case ASTextNodeVerticalTextAlignmentBottom:
+      textOrigin.y = self.bounds.size.height - shadowPadding.top - shadowPadding.bottom - [self _renderer].size.height;
+    default:
+      break;
+  }
+  
   return [[ASTextNodeDrawParameters alloc] initWithRenderer:[self _renderer]
                                                  textOrigin:textOrigin
                                             backgroundColor:self.backgroundColor.CGColor];
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -381,10 +381,13 @@ static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncation
 {
   // Offset the text origin by any shadow padding
   UIEdgeInsets shadowPadding = [self shadowPadding];
-  CGPoint textOrigin = CGPointMake(self.bounds.origin.x - shadowPadding.left, self.bounds.origin.y - shadowPadding.top);
+  CGPoint textOrigin = CGPointMake(self.bounds.origin.x - shadowPadding.left, 0.0f);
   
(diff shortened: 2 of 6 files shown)


appleguy:
@nivanchikov fantastic, thank you!  @rcancro can help us review this important new addition.  I know that @nguyenhuy / @levi and others are extremely interested in the possibility of enabling a per-item alignment in stacks as well (which in one of the orientations would translate to a per-item vertical alignment).  So my main question is how, if at all, we should integrate this with that.

Could you share a couple examples of the use cases you have found for this?  I'll note that it seems to me like this option will only take effect if the user sets the frame of the text node to be larger than it calculates for itself during the measurement pass.  This is something that should be avoided whenever convenient, because more memory is consumed by the blank areas in the larger backing store (and also, more blending / texture upload cost for the GPU).  Still, I think it is a worthwhile flexibility to have in the framework.

nivanchikov:
@appleguy thanks for your feedback!

It seems like I can't think of more than one example you already mentioned - when node bounds are larger that its content. And although I second your thoughts on memory consumption, this approach served me well when building small nodes that required this option, e.g. buttons or other relatively small controls.

nguyenhuy:
Given the limited use case of this feature and its inefficiency, I'm wondering if we should make it internal and only expose it in relevant nodes (ASButtonNode, for example)?

examples/ASCollectionView/Sample/TextNode.m
```diff
@@ -0,0 +1,46 @@
...
+
+- (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
+{
+  CGSize size = [_textNode measure:constrainedSize];
+  return CGSizeMake(constrainedSize.width, size.height + 20.0f);
+}
+
+- (void)layout
+{
+  _textNode.frame = self.bounds;
```
appleguy: Very important to call [super layout], at least in general (likely not necessary in this specific example, but we should ensure others don't see it as an indication not to)

examples/ASCollectionView/Sample/TextNode.h
```diff
@@ -0,0 +1,15 @@
...
+//  TextNode.h
+//  Sample
+//
+//  Created by Nikita Ivanchikov on 12/8/15.
+//  Copyright © 2015 Facebook. All rights reserved.
+//
+
+#import <AsyncDisplayKit/AsyncDisplayKit.h>
+
+@interface TextNode : ASCellNode
```
appleguy: Do we need this as opposed to using ASTextCellNode?  I think it is defined at the bottom of ASCellNode.h.  If that version isn't good enough for testing, e.g. you can't access the internal text node to set the alignment property, let me know and let's improve it in this diff so it is good enough for you to use here / delete this

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -379,7 +381,22 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
...
-  CGPoint textOrigin = CGPointMake(self.bounds.origin.x - shadowPadding.left, self.bounds.origin.y - shadowPadding.top);
+  CGPoint textOrigin = CGPointMake(self.bounds.origin.x - shadowPadding.left, 0.0f);
+  
+  switch (_verticalTextAlignment)
+  {
+    case ASTextNodeVerticalTextAlignmentTop:
+      textOrigin.y = self.bounds.origin.y - shadowPadding.top;
+      break;
+    case ASTextNodeVerticalTextAlignmentCenter:
+      textOrigin.y += (self.bounds.size.height - [self _renderer].size.height) / 2.0f;
```
appleguy: I'm wondering if we should use the renderer here or self.calculatedSize.size.height.  Probably not, I guess the renderer is the right thing?

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -379,7 +381,22 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
...
+      break;
+    case ASTextNodeVerticalTextAlignmentBottom:
+      textOrigin.y = self.bounds.size.height - shadowPadding.top - shadowPadding.bottom - [self _renderer].size.height;
```
appleguy: Missing a break statement on this one!  Technically doesn't cause a bug, but needs to be added.

AsyncDisplayKit/ASTextNode.mm
appleguy: Why is this one += and the others are = ?  Use = if it is the same functionally, which it seems like it is due to 0.0f initialization

AsyncDisplayKit/ASTextNode.h
```diff
@@ -26,6 +26,26 @@ typedef NS_ENUM(NSUInteger, ASTextNodeHighlightStyle) {
 };
 
 /**
+ * The vertical alignment of text within a node.
+ */
+typedef NS_ENUM(NSUInteger, ASTextNodeVerticalTextAlignment) {
+  /**
+   * Aligns the content vertically at the top in the node (the default).
+   */
+  ASTextNodeVerticalTextAlignmentTop,
```
appleguy: @nivanchikov my most important feedback for this diff is to find a way for this feature to be a bit less prominent, and not push down the rest of the functionality in this critical header as much.

Can you remove the block comments for each of the enum values, so there is one per line, and then on the same line as the first one do:

...Top,   // Top is the default value

AsyncDisplayKit/ASTextNode.h
```diff
@@ -26,6 +26,26 @@ typedef NS_ENUM(NSUInteger, ASTextNodeHighlightStyle) {
 };
 
 /**
+ * The vertical alignment of text within a node.
```
appleguy: Add to the end: "See .verticalTextAlignment property for more details"

AsyncDisplayKit/ASTextNode.h
```diff
@@ -71,6 +91,12 @@ typedef NS_ENUM(NSUInteger, ASTextNodeHighlightStyle) {
 @property (nonatomic, assign) NSUInteger maximumNumberOfLines;
 
 /**
+ @abstract The vertical alignment of text within the receiver.
```
appleguy: Use the documentation flag "default" to indicate the default value.  In the discussion, say something like:

"This property should not be used as the primary mechanism to vertically position text, which should be achieved with layout specs or the -layout method.  However, in cases where it is necessary to alter the position at which text draws within the text node's backing store, this can be used.  Note: this scenario only occurs if the text node's bounds / frame are set to a size greater than specified by calculatedLayout (or return value from -measureWithSizeRange:).  Over-sizing the text node results in extra memory usage and compositing, so consider alternative solutions before relying on .verticalTextAlignment"

appleguy:
@nivanchikov I apologize for the delay in my review - there are a few issues to address as this is a highly critical API surface for AsyncDisplayKit.  I was not sure if I'd be willing to take this diff, but I've decided to do so, if you can tweak the areas requested.  Thanks again for your work on this; I really appreciate it!

AsyncDisplayKit/ASTextNode.h
```diff
@@ -71,6 +91,12 @@ typedef NS_ENUM(NSUInteger, ASTextNodeHighlightStyle) {
...
+ @discussion Defaults to ASTextNodeVerticalTextAlignmentTop
+ */
+@property (nonatomic, assign) ASTextNodeVerticalTextAlignment verticalTextAlignment;
```
appleguy: Let's move this to be the very last property.  Add a #pragma mark - Miscellaneous before it just to separate it from Placeholders (not so sure those pragmas are useful in the first place, but we'll leave them for now).

appleguy:
@nivanchikov could you describe the use case one more time?  If this is actually expected to be used with ASButtonNode fairly regularly, we should think of some way to call it out in that header file so people know about it.  Probably better would be to provide some kind of .titleEdgeInsets (UIEdgeInsets) or something though.

appleguy:
@nivanchikov I know we're going into the holidays, so let me know if you plan to work on this soon - otherwise we can plan to revisit in January perhaps and I won't worry about it much :)

nivanchikov:
@appleguy sorry guys, I'm on vacation myself so lets get back to it in Jan if it's OK

appleguy:
@nivanchikov No rush here - if you don't have a chance to wrap this up in the next week or two, let's close it and we can easily reopen it again later.

PR #931, opened 2015-12-10
[ASCellNode] Support for wrapping UIViewControllers

lappp9:
</img>

This is a bit of a work in progress, but I'm taking a crack and initializing an ASCellNode with a UIViewControllerBlock.  This example image shows the collection view example except the views shown in the cells are generated from an ImageViewController that has an imageView as a subview.

Right now it grabs the vc's view, makes a subnode out of it and adds it as a subnode of the cell.   I'm resizing the vc's view based on what comes through in layoutSpecThatFits but that might be a little off or not the right way to do it.  

Random thought, would it be smart to take the vc's view, calculate the aspect ratio of the size that was calculated for the cell by layoutSpecThatFits and then apply a transform to that view to make it fit in that aspect ratio?

Also touch handling is lost right now but I assume you could forward the taps to the saved vc if there is one?


lappp9:
Also the sample app is called ```CollectionViewWithViewControllerCells``` btw

appleguy:
@lappp9 awesome head start on this feature - a highly desired one for a number of use cases, including @levi's ASPagerNode!  Will review / give feedback soon.

lappp9:
Side note: I noticed that viewWillAppear and viewDidAppear seemed to already work on the view controllers I threw in there.

lappp9:
Oh also, View Controllers do have a preferredContentSize that's intended to be used for popovers.  If we want to go the other direction and have the view controllers inform the cell size I assume we could use that?

https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIViewController_Class/#//apple_ref/occ/instp/UIViewController/preferredContentSize

AsyncDisplayKit/ASCellNode.h
```diff
@@ -93,6 +93,29 @@ typedef NSUInteger ASCellNodeAnimation;
...
+/**
+ * @abstract Initializes a cell with a given viewControllerBlock.
+ *
+ * @param viewControllerBlock The block that will be used to create the view controller whose root view will be used to create the backing view.
+ *
+ * @return An ASCellNode created using the root view of the view controller provided by the viewControllerBlock.
+ * The view controller's root view is resized to match the calcuated size produced during layout.
+ *
+ */
+- (instancetype)initWithViewControllerBlock:(ASDisplayNodeViewControllerBlock)viewControllerBlock;
```
appleguy: Just to reduce API surface area, I think we should probably kill this variant.  Passing nil as the didLoadBlock is pretty reasonable especially since this will not be a very common call in people's apps.

AsyncDisplayKit/ASCellNode.m
```diff
@@ -18,6 +18,14 @@
 #pragma mark -
 #pragma mark ASCellNode
 
+@interface ASCellNode (){
```
appleguy: Open brace on the next line

AsyncDisplayKit/ASCellNode.m
```diff
@@ -32,6 +40,35 @@ - (instancetype)init
...
+}
+
+- (instancetype)initWithViewControllerBlock:(ASDisplayNodeViewControllerBlock)viewControllerBlock didLoadBlock:(ASDisplayNodeDidLoadBlock)didLoadBlock
+{
+  if (!(self = [super init]))
+    return nil;
+  
+  ASDisplayNodeAssertNotNil(viewControllerBlock, @"should initialize with a valid block that returns a UIViewController");
+  
+  _viewController = viewControllerBlock();
```
appleguy: Even though we assert, we need to explicitly check if the block is nil before calling it, so that a production app would not crash (showing blank is much better)

AsyncDisplayKit/ASCellNode.m
```diff
@@ -32,6 +40,35 @@ - (instancetype)init
...
+  
+  _viewControllerNode = [[ASDisplayNode alloc] initWithViewBlock:^UIView *{
+        return _viewController.view;
```
appleguy: I am not certain, but I think this probably creates a retain cycle.  It's probably reasonable to put "return viewController.view;" without accessing the instance variable, oddly - then the block will capture the view controller directly, rather than capturing self so that it can access specifically the value of that instance variable at the time the block is run (rather than just the current value of viewController, the argument)

AsyncDisplayKit/ASCellNode.m
```diff
@@ -32,6 +40,35 @@ - (instancetype)init
...
+      } didLoadBlock:didLoadBlock];
+  
+  [self addSubnode:_viewControllerNode];
+  
+  _nodeLoadedBlock = didLoadBlock;
+  return self;
+}
+
+- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize {
```
appleguy: Brace on next line :)

It is very important not to set a frame in this method.  This method is called off the main thread, and as written today, you can't setFrame: on a node after it is loaded (as it sends it to UIView after the UIView exists).

You should instead not actually implement layoutSpecThatFits:, but instead implement -layout.  Call super in the method, then you can set _viewControllerNode.frame = self.bounds;  Fortunately -layout is always guaranteed to be called on the main thread.

This seems to me like the best implementation as it could be possible for a subclass to implement an actual -layoutSpecThatFits, and it's a bit tough to add the viewControllerNode to an existing spec.  Honestly though it could be done using ASOverlayLayoutSpec, but it doesn't seem necessary.

appleguy:
@lappp9 ok, I will land this tomorrow if you are able to do another round on feedback!  I'd like to build off this patch this weekend with some customizations for ASViewController-based wrapping (setting interfaceState automatically etc), as I am currently adopting ASPagerNode in Pinterest and want it to page through ASVCs (but it currently expects ASCellNodes, so this is directionally what is necessary).  Thanks for your work across the project, it is really cool!

AsyncDisplayKit/ASCellNode.m
lappp9: Alright, I went through your comments and fixed everything except for this so far.  For some reason the constrainedSize.max returns the correct size for each cell but when I try to just set the frame in -layout or -layoutDidFinish every cell returns a size of (width = 355, height = 44).

Seemingly unrelated, once you get to those methods _viewControllerNode is nil for some reason.

appleguy:
Nice improvements!  I'll go ahead and land this now, but it is /very/ important that we ensure the view controller as well as its view are not created until the -didLoad call.

I'll probably end up modifying this behavior for ASViewControllers though, so if you had to choose, feel free to focus on the video node first.

PR #933, opened 2015-12-11
Fixes out of order blocks running on main queue in ASDataController

garrettmoon:
ASDataController’s completedNodes is accessed by treating the main thread as a serial queue. The ​*problem*​ is, operations can cut in line by being added while on the main thread. I.e. it just calls the block instead of dispatch_async’ing to the main thread. This results in data inconsistency.

To address this, I've added a queue of operations which get run serially (not to be confused with a serial dispatch queue). Instead of running blocks directly on the main thread, if it’s called while not on the main thread, it dispatch_asyncs to the main thread and runs any blocks in the queue. If it ​*is*​ on the main thread, it runs any blocks already in the queue and then runs the block.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -16,6 +16,7 @@
 #import "ASMultidimensionalArrayUtils.h"
 #import "ASInternalHelpers.h"
 #import "ASLayout.h"
+#import "ASThread.h"
 
 //#define LOG(...) NSLog(__VA_ARGS__)
 #define LOG(...)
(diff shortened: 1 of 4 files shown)


garrettmoon:
One concern I have: should methods that are meant to be called on the main thread, also be wrapped in performBlockOnMainThread:?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -971,3 +984,51 @@ - (void)dealloc
 }
 
 @end
+
+@implementation ASMainQueueSerialQueue
```
appleguy: Since this code is completely generic in its application - although I am afraid to say I hope we never need to use it again, LOL - what do you think about moving it to ASThread or a similar helper / utility file?

Some solid magic in the runBlocks method...

garrettmoon: Sure, can you think of a better name while we're at it? I'm not in love with this one.

PR #935, opened 2015-12-12
fix overflow issues with ASEditableTextNode's textView

samhsiung:
Setting clipBounds = NO causes UITextView to overflow in iOS9 (works fine however in iOS8), so we fix by removing this config.

Addressing the deleted comment:
1. From my tests, selection handle doesn't clip anymore in iOS9+ and iOS8.4 simulator
2. If there is clipping of the handle in any case, we recently made the textView inset configurable which should address the problem by providing a way to give the handle room to breathe.

before:
simulator screen shot dec 11 2015 7 00 48 pm

after: (more than three lines, the extra lines are clipped)
simulator screen shot dec 11 2015 6 53 46 pm

This should also fix the placeholder text overflowing which has been reported by @steveram

### DIFF
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -124,7 +124,6 @@
       textView.opaque = NO;
     }
     textView.textContainerInset = self.textContainerInset;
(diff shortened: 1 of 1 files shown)


appleguy:
Thanks a lot for investigating this, @samhsiung!!

PR #936, opened 2015-12-13
Fix truncation calculation in ASTextNode by removing layoutManager.usesFontLeading = YES in renderer context

### LINKED ISSUE #930: (ASDK 1.9.3) ASTextNode truncates on first line instead of second line for single words that are long enough
For "hsiung@gmail.com" (see the last icon and caption in the screenshot below) it renders the text in two lines without truncation. However a longer email address such as "hsiung@mailinator.com" will result in truncation on the first line (see the first icon and caption in the screenshot below). In 1.9.2 ASTextNode was truncating correctly on the second line.

simulator screen shot dec 9 2015 11 09 16 am

Text node:
```
ASTextNode *textNode = NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
        paragraphStyle.lineBreakMode = NSLineBreakByWordWrapping;
        paragraphStyle.paragraphSpacing = paragraphStyle.lineSpacing;
        paragraphStyle.alignment = NSTextAlignmentCenter;
        UIFont *font = [UIFont fontWithName:@"HelveticaNeue-Bold" size:12.0];
        NSMutableDictionary *attributes = [[NSMutableDictionary alloc] initWithDictionary:@{ NSParagraphStyleAttributeName : paragraphStyle,
                                                                                             NSFontAttributeName : font,
                                                                                             NSForegroundColorAttributeName : [UIColor blackColor],
                                                                                             NSKernAttributeName : @(-0.2)}];
        [textNode setAttributedString:[[NSAttributedString alloc] initWithString:@"averylongemail@address.com" attributes:attributes]];
```
Layout code:
```
    CGSize exactSize = CGSizeMake(60, 30);
    ASStackLayoutSpec *verticalStack = [ASStackLayoutSpec verticalStackLayoutSpec];
    verticalStack.alignItems = ASStackLayoutAlignItemsStretch;
    verticalStack.children = @[textNode];
    verticalStack.layoutPosition = CGPointZero;
    verticalStack.sizeRange = ASRelativeSizeRangeMakeWithExactCGSize(exactSize);
    ASStaticLayoutSpec *staticSpec = [[ASStaticLayoutSpec alloc] init];
    staticSpec.children = @[verticalStack];
```

samhsiung:
After debugging #930 it was discovered that increasing the constrainedSize height slightly for ASTextKitRenderer's ASTextKitContext enabled it to truncate on the correct line rather than line N - 1.

After digging some more, found that `ASTextKitContext`'s `_layoutManager`'s `usesFontLeading` was set to NO. However we are using customized font leading in our project (though I'm not sure about `ASTextNode` in general),  so this resulted in a discrepancy between the calculated shrunk size of the text node and the layout+truncation calculation.

I suspect it should be okay to remove overriding of usesFontLeading since apple docs state that the default is YES. I also think it is better to customize via the `layoutManagerFactory` parameter in `ASTextKitContext`'s init method if `usesFontLeading` needs to be set to NO rather than overriding it for all cases.

Regardless, `removing usesFontLeading = NO` fixed #930 for me:
 
Before:
simulator screen shot dec 12 2015 8 16 48 pm

After:
simulator screen shot dec 12 2015 7 58 44 pm

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitContext.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitContext.mm
@@ -36,7 +36,7 @@
     // Create the TextKit component stack with our default configuration.
     _textStorage = (attributedString ? [[NSTextStorage alloc] initWithAttributedString:attributedString] : [[NSTextStorage alloc] init]);
     _layoutManager = layoutManagerFactory ? layoutManagerFactory() : [[NSLayoutManager alloc] init];
-    _layoutManager.usesFontLeading = NO;
+    _layoutManager.usesFontLeading = YES;
     [_textStorage addLayoutManager:_layoutManager];
     _textContainer = [[NSTextContainer alloc] initWithSize:constrainedSize];
     // We want the text laid out up to the very edges of the container.
--- a/AsyncDisplayKitTests/ASTextKitTests.mm
+++ b/AsyncDisplayKitTests/ASTextKitTests.mm
@@ -26,7 +26,7 @@ static UITextView *UITextViewWithAttributes(const ASTextKitAttributes &attribute
   textView.textContainer.lineFragmentPadding = 0.f;
   textView.textContainer.maximumNumberOfLines = attributes.maximumNumberOfLines;
   textView.textContainerInset = UIEdgeInsetsZero;
-  textView.layoutManager.usesFontLeading = NO;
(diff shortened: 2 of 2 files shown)


samhsiung:
Actually, I we're using customized line spacing at the paragraph level and not a customized font leading so I think it's some other issue.

PR #937, opened 2015-12-13
Make ASTextNode Align Its calculatedSize to the Pixel Grid

Adlai-Holler:
All hail the pixel grid.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -183,7 +183,12 @@ static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncation
   ASDisplayNodeRespectThreadAffinityOfNode(self, ^{
     [self setNeedsDisplay];
   });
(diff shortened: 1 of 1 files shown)


appleguy:
Yes.  However, we actually also need to do something like this for all ASLayouts.  Under-specified Stack layouts can crash due to setting NaN and/or Inf on CALayer, and they also routinely set really uneven floating point numbers.  Any idea on what layer we could inject this to properly apply that effect without messing up more manual calculations that folks do based on calculatedLayout?  I think ASLayout itself will just need to return the rounded values.  Wonder if the snapshot tests would pass with this.

cc @rcancro, @Adlai-Holler, @nguyenhuy 

This patch is fine to merge and I'm ready to take it, but since it would probably be superseded / replaced by the other change, I'm wondering if we can quickly find the right spot to drop that in.

appleguy:
Superseded by #945.  Thanks a lot for bringing this up @Adlai-Holler !

Issue #940, opened 2015-12-14
Adding any value for NSUnderlineStyleAttributeName causes NSParagraphStyle lineSpacing to be ignored


bsmith11:
Using `ASTextNode`, adding any value for `NSUnderlineStyleAttributeName` causes the `lineSpacing` value to be ignored in `NSParagraphStyleAttributeName` value objects. I have tested this using `UILabel` and it works correctly. Using ADK 1.9.3. Example:

```objc
NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
paragraphStyle.lineSpacing = 10.0f;
NSDictionary *attributes = @{NSForegroundColorAttributeName:[UIColor blackColor],
                             NSFontAttributeName:[UIFont systemFontOfSize:14.0f],
                             NSParagraphStyleAttributeName:paragraphStyle};
NSMutableAttributedString *mutableAttributedString = [[NSMutableAttributedString alloc] initWithString:@"Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book."
                                                                                            attributes:attributes];

[mutableAttributedString addAttribute:NSUnderlineStyleAttributeName 
                                value:@(NSUnderlineStyleSingle)
                                range:NSMakeRange(10, 10)];

self.textNode.attributedString = mutableAttributedString;
```

<img width="375" alt="screen shot 2015-12-14 at 12 46 12 pm" src="https://cloud.githubusercontent.com/assets/7103810/11788839/f6e77e2c-a260-11e5-832e-79962a884ec5.png">

<img width="376" alt="screen shot 2015-12-14 at 12 47 27 pm" src="https://cloud.githubusercontent.com/assets/7103810/11788845/fe63f6b2-a260-11e5-84da-18f5c5bf0b28.png">

samhsiung:
might be an issue with `ASCleanseAttributedStringOfCoreTextAttributes` in `ASTextKitCoreTextAdditions`  which scrubs unsupported CoreText attributes.

bsmith11:
[...]

PR #942, opened 2015-12-15
ASTextKit CoreText attributes cleanse bug

bsmith11:
Fixed ASTextKit CoreText attributes bug which was incorrectly reading an NSParagraphStyle as CTParagraphStyleRef

### DIFF
--- a/AsyncDisplayKitTests/ASTextKitCoreTextAdditionsTests.m
+++ b/AsyncDisplayKitTests/ASTextKitCoreTextAdditionsTests.m
@@ -12,6 +12,11 @@
 
 #import "ASTextKitCoreTextAdditions.h"
 
+BOOL floatsCloseEnough(CGFloat float1, CGFloat float2) {
+  CGFloat epsilon = 0.00001;
+  return (fabs(float1 - float2) < epsilon);
+}
+
 @interface ASTextKitCoreTextAdditionsTests : XCTestCase
 
 @end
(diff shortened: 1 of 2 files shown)


bsmith11:
#940

AsyncDisplayKitTests/ASTextKitCoreTextAdditionsTests.m
```diff
@@ -44,5 +49,21 @@ - (void)testNoAttributeCleansing
   XCTAssertTrue([testString isEqualToAttributedString:actualCleansedString], @"Expected the output string %@ to be the same as the input %@ if there are no core text attributes", actualCleansedString, testString);
 }
 
+- (void)testNSParagraphStyleNoCleansing
+{
+  NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
+  paragraphStyle.lineSpacing = 10.0;
+
+  //NSUnderlineStyleAttributeName flags the unsupported CT attribute check
```
appleguy: Space after // - also should go ahead and link to this issue or give slightly more description on the expected behavior, as it currently is unclear why we'd want an unsupported attribute to make it through

appleguy:
@bsmith11 thanks so much for debugging this and taking the time to implement & test a fix!  I am bummed this issue slowed down your development, though.

I have to admit that I am not familiar with this CoreText attribute cleansing code and its original or current purpose.  It seems like this basically circumvents that for a certain class of CT properties, which I certainly agree shouldn't be filtered out, although it's hard for me to say if this breaks a perhaps even rarer edge case originally intended to be addressed by the attribute cleansing.  I think this is better than it was before, so in it goes!

PR #952, opened 2015-12-18
[ASVideoNode] Basic Video node using AVPlayerLayer.

lappp9:
Here's a start at the video nodes with autoPlay and autoRepeat functionality.  The main property is the .asset property which can be set with any AVAsset object.  You can also add your own playButtonNode.  I also added two example apps, one called "Nic Cage TableView" and the other called "Videos".

I'm sure there's still performance stuff to do before it's really ready but I'd be interested to get some feedback on what's there.


appleguy:
Awesome, thanks @lappp9!  @Adlai-Holler, you should give us your feedback - still very early, but we want to start adding more APIs that become more opinionated on how things like the playhead / scrubber are specified and customized.  I'm thinking we will have a play-in-place mode that doesn't right away have support for a full scrubber control, but also offer a fullscreen mode backed by the existing UIKit media player.  The core value proposition would be efficient async fetchData / display range support for the thumbnail image as well as smoothly triggering autoplay upon the Visible .interfaceState (if some property like .autoplay = YES).

appleguy:
@Eke, you're using a video node as well!  Who am I forgetting?  I am sure there was at least 1 or 2 other great folks who also were planning to use this soon & I talked to recently...

AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -212,6 +212,11 @@ - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
   return stack;
 }
 
+- (void)layoutDidFinish
```
appleguy: Let's remove unless there are plans to add to it in the next couple weeks.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,30 @@
...
+
+typedef NS_ENUM(NSUInteger, ASVideoGravity) {
+  ASVideoGravityResizeAspect,
+  ASVideoGravityResizeAspectFill,
+  ASVideoGravityResize
+};
+
+@interface ASVideoNode : ASDisplayNode<_ASDisplayLayerDelegate>
+@property (atomic, strong, readwrite) AVAsset *asset;
+@property (nonatomic, assign, readwrite) BOOL shouldAutoPlay;
```
appleguy: API is looking great so far!  Let's lowercase play, and call it shouldAutoplay

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,30 @@
...
+@end
+
+@protocol ASVideoNodeDelegate <NSObject>
+@end
+
+@protocol ASVideoNodeDatasource <NSObject>
+@optional
+- (ASDisplayNode *)playButtonForVideoNode:(ASVideoNode *) videoNode;
+- (UIImage *)thumbnailForVideoNode:(ASVideoNode *) videoNode;
+- (NSURL *)thumbnailURLForVideoNode:(ASVideoNode *)videoNode;
```
appleguy: Cool, this is a pretty good setup for both URL and UIImage.  I wonder if these should just be properties rather than a dataSource protocol?

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,30 @@
...
+
+- (void)play;
+- (void)pause;
+
+@end
+
+@protocol ASVideoNodeDelegate <NSObject>
+@end
+
+@protocol ASVideoNodeDatasource <NSObject>
```
appleguy: If we do keep this protocol, DataSource capitalization I think

lappp9:
Side thought for later.  Would it make sense to add an enum property for what the tap action is supposed to trigger?  It could be playVideo, playAudio and goFullscreen.  Or something like that?

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,30 @@
...
+@optional
+- (ASDisplayNode *)playButtonForVideoNode:(ASVideoNode *) videoNode;
```
appleguy: No space after ).  Why does this return an ASDisplayNode rather than ASButtonNode?  Can we expect people to call into the playButton directly and set values on it instead of providing one from the dataSource, or is there a separate use case?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
+
+
+#import "ASVideoNode.h"
+
+@interface ASVideoNode () {
```
appleguy: { on next line.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  ASDN::RecursiveMutex _lock;
+  
+  __weak id<ASVideoNodeDatasource> _datasource;
```
appleguy: _dataSource

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  _playerNode = [[ASDisplayNode alloc] initWithLayerBlock:^CALayer *{
+    AVPlayerLayer *playerLayer = [[AVPlayerLayer alloc] init];
+    playerLayer.player = [[AVPlayer alloc] init];
+    return playerLayer;
+  }];
+  [self addSubnode:_playerNode];
+  
+  self.gravity = ASVideoGravityResizeAspect;
+  
+  [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(restartVideo:) name:AVPlayerItemDidPlayToEndTimeNotification object:nil];
```
appleguy: Let's call the selector -didPlayToEnd, and it can do whatever we want inside it - in this case restarting the video.

AsyncDisplayKit/ASVideoNode.h
lappp9: So I put these in here at first but then just added the property for the playButton and figured we could figure out the direction we wanted to go later.  Do you think we should just switch all these to properties?  Is that kind of the standard in the framework so far?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  }
+  
+  if ([[change objectForKey:@"new"] integerValue] == AVPlayerItemStatusFailed) {
+    
+  }
+}
+
+- (void)restartVideo:(NSNotification *)notification
+{
+  if ( [[[notification object] asset] isEqual:_asset]) {
```
appleguy: Let's use ASObjectIsEqual() - always prefer it.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+    }
+    if (_spinner) {
+      [self addSubnode:_spinner];
+    }
+  }
+}
+
+- (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary *)change context:(void *)context
+{
+  if ([[change objectForKey:@"new"] integerValue] == AVPlayerItemStatusReadyToPlay) {
```
appleguy: Are any of the enum values possibly 0?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  [self addSubnode:_playerNode];
+  
+  self.gravity = ASVideoGravityResizeAspect;
+  
+  [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(restartVideo:) name:AVPlayerItemDidPlayToEndTimeNotification object:nil];
+  
+  return self;
+}
+
+- (void)setInterfaceState:(ASInterfaceState)newState
```
appleguy: Let's add a // FIXME: Adopt interfaceStateDidChange API when it is available.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  [super setInterfaceState:newState];
+  
+  if (!(newState & ASInterfaceStateVisible)) {
+    [self pause];
+    [_spinner removeFromSupernode];
+  } else {
+    if (_shouldBePlaying) {
+      [self play];
+    }
+    if (_spinner) {
```
appleguy: Should we have to check here to add the spinner as a subnode?  Maybe we should call -startSpinnerIfNeeded or something, but just adding it seems weird.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+    
+    if (_autorepeat) {
+      [self play];
+    } else {
+      [self pause];
+    }
+  }
+}
+
+- (void)layoutDidFinish
```
appleguy: make this -layout.  You do have to call super, but then set the playerNode's frame.

I think this will have to be changed to -layoutSpecThatFits: in order for it to ideally interact with sizing / layout.  We may need to create a static layout spec or set a preferred size (more likely?) if we have access to a UIImage with the size of the placeholder.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+{
+  _playerNode.frame = self.bounds;
+}
+
+- (void)didLoad {
```
appleguy: { on next line.  Put this method right after init.

AsyncDisplayKit/ASVideoNode.mm
lappp9: Yeah, I'm pretty sure AVPlayerItemStatusUnknown is 0 and AVPlayerItemStatusReadyToPlay is 1

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  [super didLoad];
+  
+  UITapGestureRecognizer *tap = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(tapped)];
```
appleguy: Actually let's delete -didLoad because we should avoid creating a gesture recognizer if at all possible, which I think we can do by making sure this node is an ASControlNode, and adding a target / action directly to ourselves in init which calls -tapped.

We need to make sure the ASButtonNode for the play button can also trigger this, though.  Probably also want the play button to look selected even if the user taps somewhere else on the video.  So it might be best to set userInteractionEnabled = NO on the play button, and then even forward the -touchesBegan / ended methods?  hmm...maybe override hitTest: on this node to always hitTest: to the playButton, and then allow it to have user interaction, and set the target / action there would be best.

In any case, you should be able to kill the recognizer √

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+    [_currentItem addObserver:self forKeyPath:NSStringFromSelector(@selector(status)) options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew context:NULL];
+
+    if (((AVPlayerLayer *)_playerNode.layer).player) {
+      [((AVPlayerLayer *)_playerNode.layer).player replaceCurrentItemWithPlayerItem:_currentItem];
+    } else {
+      ((AVPlayerLayer *)_playerNode.layer).player = [[AVPlayer alloc] initWithPlayerItem:_currentItem];
+    }
+
+  }
+  if (_shouldAutoPlay) {
```
appleguy: {, next line :)

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+@end
+
+@implementation ASVideoNode
+
+- (instancetype)init {
+  if (!(self = [super init])) { return nil; }
+  
+  _playerNode = [[ASDisplayNode alloc] initWithLayerBlock:^CALayer *{
+    AVPlayerLayer *playerLayer = [[AVPlayerLayer alloc] init];
+    playerLayer.player = [[AVPlayer alloc] init];
```
appleguy: Since clearFetchedData clears this, maybe we should only create the player in -fetchData?  Is there a way that we can maintain the player without necessarily forcing the layer to be created, AKA not even trigger the layerBlock to run when -fetchData is called?  That way we can create it as late as possible while still preloading stuff (if we need the AVPlayer to kick off the network load, which I think we do when setting the asset)

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+{
+  ASDN::MutexLocker l(_lock);
+  
+  if (ASObjectIsEqual(((AVURLAsset *)asset).URL, ((AVURLAsset *)_asset).URL)) {
+    return;
+  }
+  
+  _asset = asset;
+  
+  if (self.interfaceState & ASInterfaceStateFetchData) {
```
appleguy: This is as much a reminder for me as anything, but add:

// FIXME: Adopt -setNeedsFetchData when it is available

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  }
+}
+
+- (AVAsset *)asset
+{
+  ASDN::MutexLocker l(_lock);
+  return _asset;
+}
+
+- (void)setGravity:(ASVideoGravity)gravity {
```
appleguy: {

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+  ASDN::MutexLocker l(_lock);
+  
+  [[((AVPlayerLayer *)_playerNode.layer) player] play];
+  _shouldBePlaying = YES;
+  _playButton.alpha = 0.0;
+  if ([self ready] && ![self.subnodes containsObject:_spinner]) {
+    _spinner = [[ASDisplayNode alloc] initWithViewBlock:^UIView *{
+      UIActivityIndicatorView *spinnnerView = [[UIActivityIndicatorView alloc] initWithFrame:_playButton.frame];
+      spinnnerView.color = [UIColor whiteColor];
+      [spinnnerView startAnimating];
```
appleguy: It's fine to wrap this as a UIActivityIndicatorView, but technically we don't want to always call startAnimating just when the view loads.  Make sure to check the usage of _spinner very carefully to ensure that it is impossible for it to be left animating when it is offscreen, as this is one of a few classes that will continue to spend CPU resources (low level but measurable) even when outside of a hierarchy.

Also it is not quite right to use initWithFrame: in the view block, as theoretically if we were out of a node hierarchy when this is called the addSubnode: will not trigger the view block to be called, and it may end up getting called before a layout pass has set the frame on _playButton.  You can instead just do _spinner.frame = in your -layout method.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,276 @@
...
+      return spinnnerView;
+    }];
+    
+    [self addSubnode:_spinner];
+  }
+}
+
+- (BOOL)ready;
+{
+  return [((AVPlayerLayer *)_playerNode.layer) player].currentItem.status != AVPlayerItemStatusReadyToPlay;
```
appleguy: probably worth creating a - (AVPlayer *)_player method so you can just call [self _player] in these places.

AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -0,0 +1,50 @@
...
+  
+  AVPlayerItem *item = [videoNode currentItem];
+  
+  videoNode.asset = [AVAsset assetWithURL:[NSURL URLWithString:@"secondURL"]];
+  AVPlayerItem *secondItem = [videoNode currentItem];
+  
+  XCTAssertNotEqualObjects(item, secondItem);
+}
+
+- (void)testVideoNodeDoesNotReplaceAVPlayerItemWhenSameURLIsSet {
```
appleguy: Thanks so much for adding tests in the first round!!  Really awesome!  { on the next lines :)

appleguy:
I did not review all the sample code, but have reviewed the core framework implementation.  This is a lot of work and reaffirms the fact that it will turn into a valuable component for many people in the community.  Thank you @lappp9 for your quick and tireless work on this and other changes recently!

Adlai-Holler:
Woot! This is an excellent first pass and I can't wait to put this component to good work. Since @appleguy already covered the implementation, and since I'm greedy, this will basically amount to feature requests :smile: . Playing video in iOS, the complexity can skyrocket really quickly, so I'm not necessarily pushing for all these immediately.

Muting/unmuting audio. This would mean adding a `muted` property to the node, and applying that to the latest player whenever either changes. If we add a delegate callback (or KVO but… eh…) for observing mutedness then users can implement their own mute/unmute toggle button. Giving them a default mute/unmute button sounds noble but would need a lot more consideration of value.

It would be valuable to detect when the video has reached the end time – here are a couple ways to pull it off:
- Make `currentItem` public, and client observes `AVPlayerItemDidPlayToEndTimeNotification`. Pro: very little code on our end, consistent with non-ASDK usage. Con: Need to inform client when player item changes, may be cumbersome for user to handle the common use case of "I don't care what item is playing, I want to know when this node finishes playing."
- Create a delegate method for it. Pro: Very little code on client's end, could even have a `shouldAutorepeat` option for the delegate to make the call, we are already observing the AVFoundation notification internally anyway. Con: none, really. 

I think the best approach is actually to do both of the above. I can't think of a reason to hide the player item. Clients may want to seek to a particular point, configure the audio mix, etc. There's so much functionality that we could safely offer users, that it might be worth informing the delegate when the current item changes so that they can configure it, or providing a `playerItemConfigurationBlock`.

Looking to the future, helping to handle buffering would be a great feature – perhaps by default showing/hiding the spinner during buffering and informing the delegate so they can handle it in other ways if desired. Observing a combination of `AVPlayerItemPlaybackStalledNotification`, `AVPlayer.rate`, `AVPlayerItem.playbackBufferEmpty`, and `AVPlayerItem.playbackLikelyToKeepUp`.

lappp9:
@Adlai-Holler, so I was wondering about the muting thing.  Would it make sense to add an enum property for what the tap action is supposed to trigger? It could be playVideo, playAudio and goFullscreen. Or something like that?

lappp9:
Or should we just have properties for each and just let people configure it in whatever combo they want?

AsyncDisplayKit/ASVideoNode.mm
lappp9: Will do!

AsyncDisplayKit/ASVideoNode.mm
lappp9: This is super pointless since if the spinner isn't nil then it's already been added as a subnode.

appleguy:
@lappp9 it looks like you need to pull ASDK master, and then on your branch, do "git merge master" - then sync the results back here.  Right now there are conflicts and we can't merge.

lappp9:
@appleguy ok, i'll do that now and hopefully finish up all these changes tomorrow!

AsyncDisplayKit/ASVideoNode.mm
lappp9: Hey @appleguy, would it be a good option to just set the hitTestSlop on the play button so that its touch area is the same size as the video node?  That's what I did for now, but if it's a bad idea just let me know!

Adlai-Holler:
@lappp9 I think using an enum to control the behavior is a little too much at this early stage. The more I think about it, the more I think that we should leave all tap handling up to the consumer including the play/pause behavior that's already wired up.

My rationale is, I expect practically all consumers to subclass `ASVideoNode` and add a bunch of other functionality to suit their needs. So giving them this one trivial piece (tap -> play/pause) may interfere with their intended tap handling, and it'd be very easy for them to replicate if the want it. Adding the enum you mentioned is a bit over-engineered, and it still won't alleviate the need for the consumer to subclass.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,34 @@
...
+- (void)play;
+- (void)pause;
+
+- (BOOL)isPlaying;
+
+@end
+
+@protocol ASVideoNodeDelegate <NSObject>
+@optional
+- (void)videoDidReachEnd:(ASVideoNode *)videoNode;
```
appleguy: Let's rename this to -videoPlaybackDidFinish:  — I know there is a slight ambiguity with whether playback was stopped or such, but I think it will fit in better with future delegate method names

appleguy:
@lappp9 very excited to merge this soon!  I had a chance to play with both test projects today.  A few bits of feedback:

- Make sure to use https URLs in both projects as they are triggering ATS.
- Rename the scrolling sample to something more descriptive of its purpose, rather than the actor in the sample video :)
- Consider merging both projects and using a simple UITabBarController or similar for the stationary + scrolling view.  This would help us test things like auto-pause when the view controller goes offscreen (visibility notifications are new to ASDK and so may not always work as expected right now - we should audit this behavior carefully and skeptically)
- Briefly refactor the test project code to minimize duplication from the Kittens app, removing anything not necessary and renaming things that still have Kittens in the name
- Different route, but one that is interesting: delete the test projects entirely and instead adopt ASVideoNode in several existing test apps.  Kittens app could have rows with full-width videos and no text, sized to fit the aspect ratio of the video correctly (find any kitten clip on the internet that is fairly short so it doesn't take a long time to load and seem slow - roughly correct resolution for an iPhone).  SocialAppLayout could have a video type, and you could modify the data source to add more stories (even if they just repeat over and over) to be able to do faster scrolling tests.  One of the best ones would be CustomCollectionView, as it is looking like it will be one of a handful of example apps that we will keep at the top level as a true reference app for people to look at (along with SocialAppLayout once its code is cleaned up, plus whatever new app is done for the guide).
- Not critical for the initial landing, but make a note to add a snapshot test (needs to use a local file URL / can't depend on network) hat checks the placeholder frame state, and also manually seeks forward to the last frame when in the playing state to check that frame.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,34 @@
...
+typedef NS_ENUM(NSUInteger, ASVideoGravity) {
+  ASVideoGravityResizeAspect,
+  ASVideoGravityResizeAspectFill,
+  ASVideoGravityResize
+};
+
+@protocol ASVideoNodeDelegate;
+
+@interface ASVideoNode : ASControlNode<_ASDisplayLayerDelegate>
+@property (atomic, strong, readwrite) AVAsset *asset;
```
appleguy: Note - although the atomic specifier is a good self-documentation hint to developers that these methods are thread safe, make sure not to use it if you are going to let the compiler synthesize either the getter or the setter for you.  It creates very slow locking that we should avoid.  So, best option is to implement all the getters / setters manually and grab the MutexLocker yourself, like ASDisplayNode.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,34 @@
...
+@property (atomic, strong, readonly) AVPlayer *player;
+@property (atomic, strong, readonly) AVPlayerItem *currentItem;
```
appleguy: I'm not that familiar with the different ways AVPlayerItems are used in practice, for looping videos etc, but I'm wondering if it is the right thing to have this called currentItem.  If it changes, maybe there should be a delegate method to signal that, and/or this should be an array of items?  It's important that we don't have to change the API surface later, only add to it.

Reading the documentation on AVPlayerItem yields much interesting information.  We should make very sure that the API we expose allows for asynchronously loading the key information.  For example if someone creates a node with an asset and then immediately calls for the currentItem, do we have a way to make sure we don't block waiting for that information to be parsed out of the video hosted on a remote server?  Even once the video is local, it can take surprisingly long (in terms of frame dropping times, on older devices) to parse information such as the duration from videos, so we should try to insulate users from that "forcibly" via the API design.  This may mean no property for the item at all, and only a method like currentItemWithCompletionBlock: or something (not a great name - think about it once you've learned more about the async fetching options from the docs).  Example snippet:

"Discussion
The value of each key in automaticallyLoadedAssetKeys will be automatically be loaded by the underlying AVAsset before the receiver achieves the status AVPlayerItemStatusReadyToPlay; i.e. when the item is ready to play, the value returned by invoking the asset property’s statusOfValueForKey:error: method will be one of the terminal status values, either AVKeyValueStatusLoaded, AVKeyValueStatusFailed, or AVKeyValueStatusCancelled.

IMPORTANT
The asset property keys @"playable" and @"compatibleWithSavedPhotosAlbum" are not eligible for automatic loading by AVPlayerItem. You must use the AVAsset method loadValuesAsynchronouslyForKeys:completionHandler: to load the values of those properties asynchronously."

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,34 @@
...
+
+typedef NS_ENUM(NSUInteger, ASVideoGravity) {
+  ASVideoGravityResizeAspect,
+  ASVideoGravityResizeAspectFill,
+  ASVideoGravityResize
+};
+
+@protocol ASVideoNodeDelegate;
+
+@interface ASVideoNode : ASControlNode<_ASDisplayLayerDelegate>
```
appleguy: Important to not expose _ASDisplayLayerDelegate in the header (also have a space before the <).  Can you put a second @interface ASVideoNode () <_ASDisplayLayerDelegate> in the .m file?  Is it absolutely necessary to have that delegate conformation at all?  E.g. does ASControlNode or ASDisplayNode not already have that?

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,34 @@
...
+@property (atomic, strong, readwrite) AVAsset *asset;
+@property (atomic, strong, readonly) AVPlayer *player;
+@property (atomic, strong, readonly) AVPlayerItem *currentItem;
+@property (nonatomic, assign, readwrite) BOOL shouldAutoplay;
+@property (atomic) ASVideoGravity gravity;
```
appleguy: We should probably just use the AV type for this (even though they are NSStrings, do they have any other type name for them?)  I suppose it is unlikely they'll introduce more, but still it would be best to minimize any new APIs unless they are easier for folks to use.  Would need to add a comment referencing the intended type / maybe just specify "// values from AVAsset.h gravity" or something, especially if it is just NSString.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,34 @@
...
+@property (atomic) BOOL autorepeat;
```
appleguy: later we should add support for autoreverse, too :)  I suppose the behavior would be to play once forward and once back, unless autorepeat is on too, in which case it would continue.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -0,0 +1,34 @@
...
+@property (atomic) ASButtonNode *playButton;
+
+@property (atomic, weak, readwrite) id<ASVideoNodeDelegate> delegate;
+
+- (void)play;
+- (void)pause;
+
+- (BOOL)isPlaying;
```
appleguy: Maybe we should have something like isPlayable or isReadyToPlay.  If we added that - I guess it could be later, as autoplay is the most key feature here - we'd need a delegate method too that notifies when it becomes ready to play.  There should also be some option to control whether the play button is visible before it's ready to play (it might pulsate the alpha with an autoreversing animation if the video is still loading), or if the button should only become visible once it's ready to play.

I'm suggesting many of these things because I think you probably already have a running list of such feature ideas from your own ideas, not because all are required right now - indeed wrapping this up with stable behavior and a minimized API that we are fairly confident won't change (only grow) is the best goal, because of the huge value of the programming guide.  It would be awesome if you could write a GitHub Issue with a loosely organized list of these kind of feature requests combined with the ones you've thought of, so that community members can join the discussion there and decide on what the "2.0" of this component looks like - and find who will build it!

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+  ASButtonNode *_playButton;
+  ASDisplayNode *_playerNode;
+  ASDisplayNode *_spinner;
+  ASVideoGravity _gravity;
+}
+
+@end
+
+@interface ASDisplayNode ()
+- (void)setInterfaceState:(ASInterfaceState)newState;
```
appleguy: Can now remove this entire forward declaration.  interfaceStateDidChange:forState: may still be in the +Beta header but is safe to use.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+
+- (instancetype)init
+{
+  if (!(self = [super init])) { return nil; }
+  
+  self.gravity = ASVideoGravityResizeAspect;
+  
+  [self addTarget:self action:@selector(tapped) forControlEvents:ASControlNodeEventTouchUpInside];
+  
+  [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(didPlayToEnd:) name:AVPlayerItemDidPlayToEndTimeNotification object:nil];
```
appleguy: Can you move this notification subscription to a later point when it is more likely to be required, even if that is -didLoad?  We may allocate thousands of nodes at once for measurement for table heights.

Also make sure we unsubscribe in -dealloc, but unsubscribe from the specific notification rather than all, as there is a pretty big efficiency boost from not having to scan all registered types of notifications for self as a possible observer.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+    }
+  }
+}
+
+- (void)layout
+{
+  [super layout];
+  _playerNode.frame = self.bounds;
+  
+  CGFloat horizontalDiff = (self.bounds.size.width - _playButton.bounds.size.width)/2;
```
appleguy: create a local variable for bounds

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+  CGFloat verticalDiff = (self.bounds.size.height - _playButton.bounds.size.height)/2;
+  _playButton.hitTestSlop = UIEdgeInsetsMake(-verticalDiff, -horizontalDiff, -verticalDiff, -horizontalDiff);
+  
+  _spinner.frame = _playButton.frame;
```
appleguy: Where is _playButton.frame set?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+    if (_player) {
+      [_player replaceCurrentItemWithPlayerItem:_currentItem];
+    } else {
+      _player = [[AVPlayer alloc] initWithPlayerItem:_currentItem];
+    }
+  }
+}
+
+// FIXME: Adopt interfaceStateDidChange API
+- (void)displayDidFinish
```
appleguy: I think this can be visibilityDidChange: - specifically should not be this method, because display may occur more than two screens away due to pre-rendering

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+{
+  [super displayDidFinish];
+  
+  ASDN::MutexLocker l(_lock);
+  
+  _playerNode = [[ASDisplayNode alloc] initWithLayerBlock:^CALayer *{
```
appleguy: Go ahead and create as well as add this in -didLoad instead.  Technically it would be fine to create in -init because the layer block won't be called anyway, but videoGravity may not be set to a custom value by then.  You should implement a -setVideoGravity, though, that does apply the videoGravity directly to the player layer if _playerNode.isNodeLoaded == YES.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+    return AVLayerVideoGravityResize;
+  }
+  if (_gravity == ASVideoGravityResizeAspectFill) {
+    return AVLayerVideoGravityResizeAspectFill;
+  }
+  
+  return AVLayerVideoGravityResizeAspect;
+}
+
+- (void)clearFetchedData
```
appleguy: Locate this next to -fetchData

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+  return _player;
+}
+
+- (void)setGravity:(ASVideoGravity)gravity
+{
+  ASDN::MutexLocker l(_lock);
+  
+  switch (gravity) {
+    case ASVideoGravityResize:
+      ((AVPlayerLayer *)_playerNode.layer).videoGravity = AVLayerVideoGravityResize;
```
appleguy: Local variable above switch for the player layer - critically though, the _playerNode may not be created yet, and even if it is, the layer may not be created yet.  We must not force the layer to be created on main, and don't want to force it to be created in general.

Storing the gravity in a instance variable, and just making it the NSString type that can be set directly on the layer would make sense.  It shortens this method and then you can just check if the node is loaded, and then if so, set it on the layer as well as the instance variable.  If it is not loaded, just rely on it being set in the initWithLayerBlock as seems to be done currently.

If someone reaches into the player layer and changes the value directly, we won't have it updated properly, so that's something to think about - probably a non-issue but a bit concerning.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -0,0 +1,352 @@
...
+{
+  ASDN::MutexLocker l(_lock);
+  return _shouldBePlaying;
+}
+
+#pragma mark - Lifecycle
+
+- (void)dealloc
+{
+  [[NSNotificationCenter defaultCenter] removeObserver:self];
```
appleguy: Since dealloc cost is quite high, let's remove the specific notification we subscribed to.  I only recently learned how much more expensive it is to do the remove-all, as it becomes more expensive the more types and number of notifications / objects are subscribed.

Issue #948, opened 2015-12-17
App Sometimes Hangs On Loading Node with ASTextNode (OSSpinLock on background color)


tettoffensive:
Seems to happen every few times this screen loads with my UIView that has some ASTextNodes. I am not explicitly setting the background color of any ASTextNode's just setting the background color of my node to clearColor. This node is directly added to a standard UIView via addSubnode.

Wondering if it has something to do with this? `#define _getFromLayer(layerProperty) __loaded ? _layer.layerProperty : self.pendingViewState.layerProperty`

If I pause the app in Xcode it stops here:
```
libsystem_platform.dylib`_OSSpinLockLockSlow$VARIANT$wfe + 96, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #0: 0x000000019972978c libsystem_platform.dylib`_OSSpinLockLockSlow$VARIANT$wfe + 96
    frame #1: 0x0000000188fbd164 QuartzCore`CA::Layer::getter(unsigned int, _CAValueType, void*) + 48
    frame #2: 0x0000000188fc04b8 QuartzCore`-[CALayer backgroundColor] + 44
  * frame #3: 0x0000000100dfc358 AsyncDisplayKit`-[ASDisplayNode(self=0x0000000156acf5e0, _cmd="backgroundColor") backgroundColor] + 504 at ASDisplayNode+UIViewBridge.mm:454
    frame #4: 0x0000000100e828d4 AsyncDisplayKit`-[ASTextNode drawParametersForAsyncLayer:](self=0x0000000156acf5e0, _cmd="drawParametersForAsyncLayer:", layer=0x0000000156b00d90) + 448 at ASTextNode.mm:385
    frame #5: 0x0000000100deed08 AsyncDisplayKit`-[ASDisplayNode(self=0x0000000156acf5e0, _cmd="drawParameters") drawParameters] + 124 at ASDisplayNode+AsyncDisplay.mm:75
[...]

PR #956, opened 2015-12-19
Remove CALayer and UIView implementations of "name" property.

appleguy:
This avoids CoreAnimation from capturing the name from a background thread deep within the internals
of AVPlayerLayer, which commits transactions off the main thread.  If the main thread is also accessing
properties that require the lock, in very specific app architectures relying on AVPlayerLayer, this
can cause a deadlock.

The "name" property is primarily used by ASDK's unit test infrastructure, so I'm leaving it in place for
that purpose - and it may still be used for debugging, if someone wants to set it.  It will remain
possible to get the node from a layer or view with asyncdisplaykit_node, and access name from there.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -685,7 +685,6 @@ typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
  * @param node The node to be added.
  */
 - (void)addSubnode:(ASDisplayNode *)node;
-- (NSString *)name;
 @end
 
 /** CALayer(AsyncDisplayKit) defines convenience method for adding sub-ASDisplayNode to a CALayer. */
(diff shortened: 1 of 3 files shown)


appleguy:
This should, with almost total certainty, resolve #948.  Shady Apple!

nguyenhuy:
Didn't know what seems to be a simple and harmless property can mess up
with CoreAnimation! Sorry for any inconvenience caused and thanks for
fixing it :)

On Sat, Dec 19, 2015, 3:34 AM appleguy <notifications@github.com> wrote:

> Merged #956 <https://github.com/facebook/AsyncDisplayKit/pull/956>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/facebook/AsyncDisplayKit/pull/956#event-496397703>.
>
-- 

Regards,
Huy Nguyen

Issue #958, opened 2015-12-19
[ASDisplayNode] Manually adding a subnode to a node's view doesn't auto-magically hook up the node hierarchy.


appleguy:
This relates to https://github.com/facebook/AsyncDisplayKit/issues/928

Basically, doing this should hook up the node hierarchy:

[node.view addSubnode:subnode];

OR

[node.view addSubview:subnode.view];

OR even

[node.layer addSublayer:subnode.layer];

appleguy:
This is nicely wrapped up.

The layer case will only work if you call -[CALayer addSubnode:], which is a provided method, and of course -addSubnode works on UIView too.  But it now also works on -[UIView addSubview:] and other insertion / removal methods.

Issue #928, opened 2015-12-09
[ASDisplayNode] Node display failing after alternating between view and node hierarchy in 1.9.3


1nput0utput:
I am facing an issue where I am animating an ASNetworkImageNode by removing it from ASCellNode to later add it to an UIViewController's view. Animating ASNetworkImageNode back to ASCellNode worked until 1.9.3, now when I add my image view back to ASCellNode, although the image view exists in ASCellNode view hierarchy, it doesn't show up. 

My -layoutSpecThatFits isn't being called on ASCellNode. I have tried -setNeedLayout upon my ASCellNode after adding my image view and that didn't help.

nguyenhuy:
Does adding your image to the cell node cause a resize? Please try to trigger an empty transaction on the table/collection view.

I'm happy to debug if you could put up a sample project. It's possible that ASCellNode's _layoutDelegate is gone by the time you add the image back ([code](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASCellNode.m#L59)), or perhaps there is a new bug introduced by #898.

appleguy:
@1nput0utput thanks for the report - I think I know what could have caused this, and can fix quickly, but need more information.

Could you include some code snippets or an example flow of steps to reproduce this, broken out by each method call / action involved to make this happen?

1nput0utput:
@nguyenhuy @appleguy sorry for not being quick to reply back. I'v quickly created a test project using the Kittens sample and I can see the exact same behaviour https://github.com/1nput0utput/ImageTest.git

Thanks for helping out!

appleguy:
@1nput0utput thanks a lot!  I'm looking at the sample project now, and should be able to fix this tonight or tomorrow - and roll a 1.9.4 before Monday.  Note your Podfile for the sample project appears to be specifying version 1.9, which you don't really want, so make sure your main project is getting the latest version!

appleguy:
@1nput0utput please try out the fix as soon as you have the opportunity.  It does fix your test project, although there is a flash that I see.  If the flash is your main problem now, let me know and I can definitely investigate fixing that specifically.  Overall though I would highly recommend you re-evaluate the architectural approach here and avoid re-hosting nodes between view and node hierarchies.

appleguy:
[...]

PR #961, opened 2015-12-20
[ASDisplayNode] Automatically find and hook up node hierarchies, even when -addSubview: is used directly.

appleguy:
Improve UIView & CALayer handling of -addSubnode:, and ensure node hierarchies are hooked up even when addSubview: is used directly.

@1nput0utput, could you please test this?  I have already done so with your app and am pretty sure it is the complete fix.  This one was a hell of a hard time to develop.  I tried all combinations of addSubview / addSubnode for both the ingoing and outgoing of the animation (including mismatched ones - going wild over here!)

I did quite a bit of debugging for the flash, but sadly and strangely have not been able to locate it yet.  The contents of the image node is NOT being cleared and the image is NOT re-decoding.  So I think the actual cause is likely that the view controllers are not swapped out in the same runloop or something like that.

Just so you know, the main reason this was so hard is that at the end of the animation you were adding the node back to the KittenNode while the hierarchy that owned the KittenNode (ASTableView etc) itself was not onscreen - its window was considered nil - and so this introduced some trickiness, but I think that is now handled.

I want to make sure the flicker is fixed, but I'm running low on time to debug that specifically.  Could you please take a look at a technique such as marking the other view .hidden or .alpha = 0.0, instead of removing it from the window completely, or at least verify that it is added and in the hierarchy behind the view that is being faded out before the transition starts?

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -996,13 +996,14 @@ static bool disableNotificationsForMovingBetweenParents(ASDisplayNode *from, ASD
   if (isMovingEquivalentParents) {
     [subnode __incrementVisibilityNotificationsDisabled];
   }
+  
   [subnode removeFromSupernode];
-
+  [oldSubnode removeFromSupernode];
+  
   if (!_subnodes)
     _subnodes = [[NSMutableArray alloc] init];
-
-  [oldSubnode removeFromSupernode];
   [_subnodes insertObject:subnode atIndex:subnodeIndex];
+  [subnode __setSupernode:self];
 
   // Don't bother inserting the view/layer if in a rasterized subtree, because there are no layers in the hierarchy and none of this could possibly work.
   if (!_flags.shouldRasterizeDescendants && [self __shouldLoadViewOrLayer]) {
@@ -1030,8 +1031,6 @@ static bool disableNotificationsForMovingBetweenParents(ASDisplayNode *from, ASD
   if (isMovingEquivalentParents) {
     [subnode __decrementVisibilityNotificationsDisabled];
   }
-
-  [subnode __setSupernode:self];
 }
 
 - (void)replaceSubnode:(ASDisplayNode *)oldSubnode withSubnode:(ASDisplayNode *)replacementSubnode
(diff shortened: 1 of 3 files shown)


appleguy:
This is intended to address https://github.com/facebook/AsyncDisplayKit/issues/958

1nput0utput:
@appleguy Thanks a lot and I am sorry that this bug put you through the torture :). I am going to test this one on my main project tonight.

As for the flickering, I think you might be right about view controllers swapping, I'll try a custom view controller transition, it can maybe fix this flicker. I was also thinking about overlaying the image view controller on tap with its on image view and then do the transition by either hiding or changing the alpha as you suggested. This way, maybe, I can avoid the whole mess I made in the image view controller.

appleguy:
@1nput0utput no worries - in the end this was actually a very important improvement to the framework's robustness.  As you can see if you look closely, I even had to update some of the expected values from a couple unit tests because of the new behavior!  It did take me half the day to get all the details right, but it is very satisfying to know that even the bizarre edge cases of something like a view-backed node contained within a layer-backed node (which actually asserts / is not supposed to happen, but could be forced at runtime by a crazy-ass user) should work correctly!!

As you are debugging the flickers, a key tip is to set a breakpoint in the -clearContents method of ASDisplayNode and also in -setNeedsDisplay on _ASAsyncLayer, and possibly the + drawing method (full selector is in ASDisplayNode+Subclasses.h) of the subclass in question, like ASImageNode.  If any of these trigger then you'll quickly be able to see what is triggering the node to redraw, and if it does redraw it will often flicker.

If you come up with a case where you see those triggering and you don't want them to, please let me know, because I can almost certainly fix it quickly.  The case in your test app right now is flicking WITHOUT any of those happening so I'm pretty sure it is the app, but it is VERY possible that as you are swapping nodes between different view and node hosting containers that the specific call order involved and other details could cause -clearContents to be triggered when you don't want it to.  In any case I'll help you get past that issue ASAP.

Do I have your email address?  Shoot me a note at asyncdisplaykit@gmail.com so I can ping you directly with other thoughts.

1nput0utput:
@appleguy my main app is working as expected without making any changes mentioned in #928. I am sending you an email as I type this comment.

PR #963, opened 2015-12-21
ASButtonNode content alignement properties

1nput0utput:
@appleguy This small change allows content alignment like in the UIButton. The reason I had initially implemented this was to align 4 ASButtonNodes with flexBasis set with ASRelativeDimension(type: .Percent, value: 0.25). I simply wanted the ability to align ASButtonNode's content to make my whole view look good. I will create another PR for other behaviours I have in my app.
 
screen shot 2015-12-20 at 9 36 10 pm

### DIFF
--- a/AsyncDisplayKit/ASButtonNode.h
+++ b/AsyncDisplayKit/ASButtonNode.h
@@ -31,6 +31,17 @@ typedef enum : NSUInteger {
  */
 @property (nonatomic, assign) BOOL laysOutHorizontally;
 
+/** Horizontally align content(text or image).
+ Defaults to ASAlignmentMiddle.
+ */
+@property (nonatomic, assign) ASHorizontalAlignment contentHorizontalAlignment;
+
+/** Vertically align content(text or image).
+ Defaults to ASAlignmentCenter.
+ */
+@property (nonatomic, assign) ASVerticalAlignment contentVerticalAlignment;
+
+
 - (NSAttributedString *)attributedTitleForState:(ASButtonState)state;
 - (void)setAttributedTitle:(NSAttributedString *)title forState:(ASButtonState)state;
 
--- a/AsyncDisplayKit/ASButtonNode.h
+++ b/AsyncDisplayKit/ASButtonNode.h
@@ -31,12 +31,12 @@ typedef enum : NSUInteger {
  */
 @property (nonatomic, assign) BOOL laysOutHorizontally;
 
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -195,8 +195,8 @@ - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
   ASStackLayoutSpec *stack = [[ASStackLayoutSpec alloc] init];
   stack.direction = self.laysOutHorizontally ? ASStackLayoutDirectionHorizontal : ASStackLayoutDirectionVertical;
   stack.spacing = self.contentSpacing;
-  stack.justifyContent = ASStackLayoutJustifyContentCenter;
-  stack.alignItems = ASStackLayoutAlignItemsCenter;
+  stack.horizontalAlignment = self.contentHorizontalAlignment ? self.contentHorizontalAlignment : ASAlignmentMiddle;
```
appleguy: @1nput0utput can you set _contentHorizontalAlignment, and _contentVerticalAlignment, to the default values in the -init method for the button?  (use the ivars, not the property for efficiency).  Then here, set the stack properties directly to the ivar values without checking them.  Nothing guarantees that the enum values won't be zero anyway, so I'm not sure it is ideal to use the ternary operator to check if they have been set to non-default values.

AsyncDisplayKit/ASButtonNode.h
```diff
@@ -31,6 +31,17 @@ typedef enum : NSUInteger {
  */
 @property (nonatomic, assign) BOOL laysOutHorizontally;
 
+/** Horizontally align content(text or image).
```
appleguy: Add space before (

appleguy:
Thanks for thinking about improvements to ASButtonNode!  @rcancro, @nguyenhuy, @Adlai-Holler - any ideas on a different way to do this with the current layout system, or does this seem like the best way?  It does minimize exposing different properties of the stack, and we can't really expose the stack layout spec directly anyway without holding a reference to it, which is not allowed with the current design.

AsyncDisplayKit/ASButtonNode.mm
1nput0utput: @appleguy I'll make the changes. It'll be much legible.

appleguy:
@1nput0utput thank you!  This is nice and clean, simple but useful!

PR #964, opened 2015-12-21
Review cleanups to ASPagerNode

levi:


aaronschubert0:
Hey @levi, this is awesome. Can't wait to use this. Just curious on why collection node's reload methods are substantially more expensive than UICollectionViews? Is this because there is a extra layer going through the node?

appleguy:
@aaronschubert0 it's because the nodes are destroyed and re-created.  Granted, nodes are cheap to allocate, and even in the case of a full reloadData, it doesn't force layout / height calculations (table) or size calculations (collection) to be done synchronously.  In addition to not forcing them synchronously, they are done concurrently on multiple cores with ASDK.  So as long as you don't call -reloadDataImmediately, it should actually be faster than UIKit to call -reloadData on an ASCollectionView if your cells have any amount of dynamic sizing (such as checking the size of text).

There is a /lot/ of documentation that is still needed for AsyncDisplayKit to help communicate these many details, but it's just been a tradeoff of either writing the code to make the framework that much more advanced, or stopping work to describe what has been done :-P

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -27,29 +27,34 @@ - (instancetype)init
...
   return self;
 }
 
-- (void)reloadData
+- (void)didLoad
 {
-  [self.view reloadData];
+  [super didLoad];
+  
+  self.view.asyncDataSource = self;
```
appleguy: Stylistically, but also for performance on the oldest devices, prefer creating a local variable for any properties accessed more than once (but especially more than three times).

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -27,29 +27,34 @@ - (instancetype)init
...
 }
 
-- (void)setTuningParameters:(ASRangeTuningParameters)tuningParameters forRangeType:(ASLayoutRangeType)rangeType
+#pragma mark - Helpers
+
+- (void)scrollToPageAtIndex:(NSInteger)index animated:(BOOL)animated
 {
-  [self.view setTuningParameters:tuningParameters forRangeType:rangeType];
+  NSIndexPath *indexPath = [NSIndexPath indexPathForItem:index inSection:0];
+  [self.view scrollToItemAtIndexPath:indexPath atScrollPosition:UICollectionViewScrollPositionLeft animated:animated];
```
appleguy: Hmm, interesting - I see the logic behind left, but I think most clients will use this with some spacing and intending to be centered onscreen...so maybe we should initially default to center and see how it goes.

appleguy:
Thanks @levi, some nice improvements here!  I think a next step is something similar to ASPendingState for collection / table node, but it is not as time-sensitive.

aaronschubert0:
Thanks for that explanation @appleguy clears up the myth that I believed that reloading ASCollectionView is a lot slower than the UIKit call. Yeah I definitely prefer advancing the framework :D

PR #966, opened 2015-12-22
[Examples] Measuring nested collection views to use during layout of parent.

KieranLafferty:
Demo project for nested collection node

### DIFF
--- a/examples/ASVerticalNestedCollectionNode/ASVerticalNestedCollectionNode.xcodeproj/project.pbxproj
+++ b/examples/ASVerticalNestedCollectionNode/ASVerticalNestedCollectionNode.xcodeproj/project.pbxproj
@@ -1,0 +1,615 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 46;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		77FD051F1C28FD8500645A65 /* main.m in Sources */ = {isa = PBXBuildFile; fileRef = 77FD051E1C28FD8500645A65 /* main.m */; };
+		77FD05221C28FD8500645A65 /* AppDelegate.m in Sources */ = {isa = PBXBuildFile; fileRef = 77FD05211C28FD8500645A65 /* AppDelegate.m */; };
+		77FD052A1C28FD8500645A65 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 77FD05291C28FD8500645A65 /* Assets.xcassets */; };
+		77FD052D1C28FD8500645A65 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 77FD052B1C28FD8500645A65 /* LaunchScreen.storyboard */; };
+		77FD05381C28FD8500645A65 /* ASVerticalNestedCollectionNodeTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 77FD05371C28FD8500645A65 /* ASVerticalNestedCollectionNodeTests.m */; };
+		77FD05431C28FD8600645A65 /* ASVerticalNestedCollectionNodeUITests.m in Sources */ = {isa = PBXBuildFile; fileRef = 77FD05421C28FD8600645A65 /* ASVerticalNestedCollectionNodeUITests.m */; };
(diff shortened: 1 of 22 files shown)


appleguy:
@KieranLafferty fantastic, thanks!  I'm triggering a rebuild, but am not sure if this one will resolve itself:

ERROR: Specified workspace doesn't exist: examples/ASVerticalNestedCollectionNode/Sample.xcworkspace

appleguy:
@KieranLafferty ok, this is a great example.  I have it running locally.  Let's not worry about committing it until there is more support for this use case and it works properly, of course, so no problem with the build not passing.

You labeled the problem correctly - creating the view there is a no-go.  However it's not immediately clear to me where / how you are using the size, and so therefore why it is actually necessary to create the view at that point.  Could you clarify that a bit just so I can be confident to implement a solution that addresses your use case as well as the one I have in mind for Pinterest?

KieranLafferty:
Hey sorry about that cherry picked it out of another project so looks like I missed a few things.

I'd liked the cell to be sized based on the content size of the nested collection node. i.e. the constrained size of the outer most collection node should be 

	- (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
	{
	    CGSize minSize = CGSizeMake(CGRectGetWidth(collectionView.bounds), 0);
	    CGSize maxSize = CGSizeMake(CGRectGetWidth(collectionView.bounds), CGFLOAT_MAX);
	    return ASSizeRangeMake(minSize, maxSize);
	}

Which should allow the cell (in this case ASEmbeddedCollectionCellNode) to be a fixed width and a flexible height (which my intention here is to have it be the natural size of the embedded collection node)

Hope that makes a bit more sense, let me know if I can expand on anything that still isn't clear, thanks again!

appleguy:
@KieranLafferty ok, we're getting somewhere.  I am actively working in this area and should have a solution for you in the next two weeks, hopefully sooner, but I don't want to over-promise either!

Can you email me at asyncdisplaykit@gmail.com so we can communicate more quickly / easily?

If you are able to do a really really basic sketch or color-block prototype of the look you're looking for (maybe that is what is in this sample project, if the crash were bypassed?), that would help.  The reason is that I need a degree of understanding around what point at which is right to trigger the internal collection view's /actual/ main thread layout pass.

The system I'm building will have an API that is basically like -measureEnoughCellsToFillEntireRect:, and then actually stop measuring after that point (or at least unblock the call), allowing the collection / table to come onscreen and be told about only the minimum number of index paths to satisfy that request and thus minimize the amount of main thread churning it does (even though the measurements will be cached, there is still nontrivial overhead especially to UICollectionView's layout attributes).  This will then pair excellently with .neverShowPlaceholders = YES.

My question is essentially: will the interior collection views be able to be scrolled onscreen rapidly with smooth scrolling, or is it in a pagination-style scenario?  If paginating, is ASPagerNode appropriate?  The API I'm working on will shine most with UINavigationController, UITabBarController, ASPagerNode, and also horizontally scrolling elements embedded in vertical scrolling ones.

An alternate take on this which I think is what you might be looking for: -measureAllCellsAndSetOrReturnContentSize.  Thanks for sharing as much info as you can so I can ensure both the API and implementation is suitable!

KieranLafferty:
Hey sorry for the delay in responding. I believe you are correct in that I am looking for -measureAllCellsAndSetOrReturnContentSize. Basically I want to size an ASCellNode  containing a collectionNode to the content size of the ASCollectionView using a layoutspec. The inner collection node would not be scrollable and would simply display it's contents in their entirety, since in this case the collectionview.bounds.size == contentSize.

Thanks!

appleguy:
We do now have a method that should solve this case, but it is still not the recommended approach because Collection views are inherently limited to main thread layout. If you use a stack, it would be quite a bit more efficient.

If you call waitUntilAllUpdatesAreCommitted, this will basically block on the completion of concurrent allocation of all the cell nodes (if you use nodeBlocks, which you should!!), and also concurrent layout measurement of everything - without forcing any reload as reloadDataImmediately would.  Then, the contentSize should be set.

There might be an issue where a runloop turn needs to occur before .contentSize is set.  If that's true, feel free to open an issue for that.  We're still looking at improvements here.

Issue #927, opened 2015-12-09
[ASDataController] Layout thread affinity for cells with nested view-based nodes


zintus:
Current implementation of ASDataController detects possibility of background layout by checking `isNodeLoaded` property on a cell node. In case such node, by the time of layout, have any subnode with   loaded view, background layout will fail.

What do you think about ensuring that whole cell node subtree have no loaded view/layer before doing background layout?

levi:
I was thinking about this issue too when running into a cell node trying to access view properties on a child collection node. The current work around is to access the properties in `didLoad`, which allows the cell to be loaded with background thread affinity.

Your suggestion is an interesting one that could be accomplished by inspecting a nodes sub layouts. At layout spec compilation time, the `calculateLayoutThatFits` method flattens the tree into a 1 level deep hierarchy of sublayouts. The data controller could then inspect these layouts to identify if the current cell has a synchronous child node somewhere in its hierarchy. Of course, this is a time complexity issues adding at least O(n) search time to the node look up for each node. Results could be pre-cached, but might be stale, especially since the flattened list of nodes are currently computed with `measureWithSizeRange:`.

zintus:
While layout flattening doing things structurally resembling our task (recurse into subnodes), i don't think it's good idea to mix layout and affinity calculations. Layout is cached differently and needed in different situations and it would clog layout.

Could we introduce `detachedFromMainThread` (not sure about naming here) implemented in spirit of `ASInterfaceState` and `ASHierarchyState` but propagating up over nodes tree.

appleguy:
I guess this is the opposite problem that ASHierarchyState solves.  We might want to introduce something comparable that up-propogates information / state of this nature as it changes, so that recursion is not necessary at access time.  However, I think it would be fine to add this check now, as it is very difficult for individual developers to understand and work around thread affinity assertions.  Adlai and I are meeting next week to discuss a next-generation threading model.

appleguy:
@zintus @levi - The community is planning an exciting long term road map for the project and getting organized around how to deliver these feature requests. We are moving to a new issue tracker with more sophisticated planning capabilities (check out www.realArtists.com and their new product: Ship!). 

If you are interested in helping contribute to this component or any other, don’t hesitate to send me an email at AsyncDisplayKit@gmail.com. If you would like to contribute for a few weeks, I can also add you to our Ship bug tracker so that you can see what everyone is working on and actively coordinate with us. 

[...]

PR #977, opened 2015-12-23
[ASDisplayNode] Implement recursivelyDetachedFromMainThread to indicate thread affinity state of subtrees.

zintus:
I've approached this problem as discussed in #927. 

New property used in data controller to ensure that faced hierarchy of nodes is available to measure on background thread. Effectively `recursivelyDetachedFromMainThread` is opposite of `isNodeLoaded` calculated over hierarchy, but it never becomes detached again after node unloaded, following comment on `view` and `layer` properties.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -172,6 +172,8 @@ typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
  */
 @property (atomic, readonly, assign, getter=isNodeLoaded) BOOL nodeLoaded;
 
+@property (atomic, readonly, assign, getter=isRecursivelyDetachedFromMainThread) BOOL recursivelyDetachedFromMainThread;
+
 /** 
  * @abstract Returns whether the node rely on a layer instead of a view.
  *
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -172,6 +172,11 @@ typedef NS_OPTIONS(NSUInteger, ASInterfaceState)
  */
 @property (atomic, readonly, assign, getter=isNodeLoaded) BOOL nodeLoaded;
 
+/**
+ * @abstract Reflects the fact of node and it's subnodes being thread agnostic.
+ *
+ * @return YES if node and it's subnodes hierarchy is safe to manipulate on any thread; NO otherwise.
+ */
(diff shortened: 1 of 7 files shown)


appleguy:
@zintus thanks for working on this!  It's an interesting improvement.  I will need to think a bit more about this because I am also working in this area, and might be going a slightly different direction.  However it is definitely true that right now it is possible to have loaded a subnode of a cell node, and then trigger an assertion on layout when the cell node tries to measure off the main thread (because the cell itself isn't loaded).  This improvement would fix that and enable a few other improvements.

In the not too distant future, @Adlai-Holler and I plan to completely rewrite the ASDK thread affinity system to actually remove the requirement to use only on the main thread after they are loaded.  That's a primary reason I'm wondering if this is the right direction...

zintus:
No problem, I'm completely ok with holding this feature until new threading
model arrives 🌞

appleguy:
@zintus thank goodness, the new threading model has been landed for a while now and has proven to be quite stable even with extensive usage!

Although I am not certain if this is 100% supported yet, in the very very near future it should be possible to have even loaded no it hierarchies be measured off the main thread by ASDataController.  In fact, we are using this capability in Pinterest - asynchronous measurement of loaded node hierarchies, with an animated transition being triggered at the moment that the measurement completes. This system has been constructed with thread safety front and center, so that even a call to -setNeedsLayout in the middle of the asynchronous measurement phase or the transition itself is handled consistently.

That support is relatively new, although also enjoying widespread use in a very stressful use case at Pinterest (multiple calls to -transition potentially before the prior finishes, due to various updates coming in to the underlying app / server data model).  The one part that I am not as certain of is, if the transition API is not use directly as is the case for the data controller, can we remove the support for main thread layout of loaded nodes?

@nguyenhuy wouldn't know the answer to this question, and although it Would take us at least a couple weeks to investigate, I think that will be the ultimate fix here...rather than supporting more easily creating dual-path code intending to support the previous, rather frustrating threading model.

Just as a quick note, it is still not thread safe to read properties that are bridged to the system objects off of the main thread, after the backing system objects are created. There is no plan to support this in the near future, although it might be possible with further sophistication in the pending state... The reason is that in my many years of using this framework, I have found that it is always quite easy to avoid relying on this capability.  Paper had a write-from-background threading model supported in its version of the framework, but I was always able to ensure that Reading from the background was avoided by providing other facilities (e.g. instead of self.bounds, look at constrainedSize).

appleguy:
Please feel free to open an Issue or email me at asyncdisplaykit@gmail.com, any time, even just with some comments about how you have used the framework or what it could do better - I would love to hear it!

PR #978, opened 2015-12-23
[ASMapNode] Greatly improve the internal logic of ASMapNode. Also fixes bug #971

### LINKED ISSUE #971: [ASMapNode] Snapshot size / zoom level may not always be invalidated at the right times.
I'm seeing that my `ASMapNode` instances all have a really zoomed-in look to them - the pin and text all look huge. I dug into it a little bit, and I think it's because each part of my setup of the node (`setRegion:`, `setAnnotations:`) is causing the node to take a snapshot. Since this takes a little while, by time the node is properly sized, `-resetSnapshotter` and `-takeSnapshot` are skipped over due to the line `if (!_snapshotter.isLoading) {`. I don't have the time to rework this, but I'm hoping someone familiar with the code can come up with a workaround or fix.

aaronschubert0:
This fixes #971 as raised by @tomizimobile. It also greatly improves the internal logic of ASMapNode. Most notably:
- You can now set _liveMap = YES on a background thread, and it will only be actioned when that node comes on screen.
- Added asserts to handle layer backing scenarios.
- Added a few additional comments to improve documentation.

Will add an example app and tests in a future diff (soon :D )
@appleguy

### DIFF
--- a/AsyncDisplayKit/ASMapNode.h
+++ b/AsyncDisplayKit/ASMapNode.h
@@ -12,7 +12,7 @@
 @interface ASMapNode : ASImageNode
 
 /**
- The current region of ASMapNode. This can be set at any time and ASMapNode will animate the change.
+ The current region of ASMapNode. This can be set at any time and ASMapNode will animate the change. This property may be set from a background thread before the node is loaded, and will automatically be applied to define the region of the static snapshot (if .liveMap = NO) or the internal MKMapView (otherwise).
  */
 @property (nonatomic, assign) MKCoordinateRegion region;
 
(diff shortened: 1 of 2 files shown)


appleguy:
@aaronschubert0 awesome, thanks for the ultra fast turnaround!  This component has already attracted a variety of users, and in January it will be put into the new Pinterest too :).

appleguy:
@aaronschubert0 I didn't realize until now all the options that MKMapSnapshotOptions provides - wow!  Later we should consider adding some more passthroughs.  I'm pretty sure we can treat that object as thread safe, so it may actually be ideal to just change the region property into the MKMapSnapshotOptions object (I had not realized this is how the snapshotter itself gets the region until converting some Pinterest code to use ASMapNode, and then noticed it would be an even more convenient conversion if it just took the same object).

appleguy:
https://github.com/facebook/AsyncDisplayKit/issues/988

aaronschubert0:
@appleguy Wow!! That's super awesome. Can't wait for that update! Good start to the New Year.

PR #985, opened 2015-12-24
[ASDisplayNode] Add delegate to ASDisplayNode, allowing ASViewController layout

### LINKED ISSUE #969: [ASViewController] Add a way to participate with ASLayoutSpec, analogous to -viewWillLayoutSubviews.
Hi,

I am very impressed how easy I build layout with ASLayout. And I want use it not only in cells, but in ViewController.

I found ASViewController, but as I understand, I need to create ASDisplayNode separately and init ASViewController with it.

I wonder, if you have any plans to tune ASViewController API to be able to specify layout in it. For instance:

``` swift
class LoginViewController: ASViewController {
    private let _loginNode = ...
    private let _passwordNode = ...
    private let _loginButtonNode = ...

    override func viewDidLoad() {
        super.viewDidLoad()
        node.addSubnode(_loginNode)
        node.addSubdode(_passwordNode)
        node.addSubnode(_loginButtonNode)
    }

    // I think this method is missing...
    override func layoutSpecThatFits(constrainedSize: ASSizeRange) -> ASLayoutSpec! {
        return ...
    }
}
```

levi:
Sketching out an initial direction of exposing layout and layout spec definition in an ASViewController. Introduced a `delegate` property on ASDisplayNode that allows external layout calculation, similar to the CALayer delegate design.

Note that this is currently not building, as we have a delegate property on a few display node subclasses which will need to inherit `ASDisplayNodeDelegate`. I wanted to put this up for feedback/discussion before I head too far down the road of bringing the other classes up to speed. Moreover, would like to include a few unit tests for this patch.

Fixes #969.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -273,6 +273,7 @@
 		92DD2FE81BF4D0A80074C9DD /* ASMapNode.h in Headers */ = {isa = PBXBuildFile; fileRef = 92DD2FE11BF4B97E0074C9DD /* ASMapNode.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		92DD2FE91BF4D4870074C9DD /* MapKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 92DD2FE51BF4D05E0074C9DD /* MapKit.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
 		92DD2FEA1BF4D49B0074C9DD /* MapKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 92DD2FE51BF4D05E0074C9DD /* MapKit.framework */; };
+		9B40BB671C4AEAE30074F2FE /* ASDisplayNodeLayoutTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 9B40BB661C4AEAE30074F2FE /* ASDisplayNodeLayoutTests.m */; };
 		9B92C8851BC2EB6E00EE46B2 /* ASCollectionDataController.mm in Sources */ = {isa = PBXBuildFile; fileRef = 251B8EF31BBB3D690087C538 /* ASCollectionDataController.mm */; };
 		9B92C8861BC2EB7600EE46B2 /* ASCollectionViewFlowLayoutInspector.m in Sources */ = {isa = PBXBuildFile; fileRef = 251B8EF51BBB3D690087C538 /* ASCollectionViewFlowLayoutInspector.m */; };
 		9C49C36F1B853957000B0DD5 /* ASStackLayoutable.h in Headers */ = {isa = PBXBuildFile; fileRef = 9C49C36E1B853957000B0DD5 /* ASStackLayoutable.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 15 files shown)


appleguy:
@levi Interesting!  We can't use the name delegate, since many app subclasses use that name already, and in comparison this will be rarely used.  I almost added this recently when I was adding interfaceStateDidChange:forState: and visibilityDidChange: because it would be convenient in certain cases to be able to monitor these calls from another class (indeed, specifically ASViewController came up as a class that could use it.)

appleguy:
We can't use the name delegate, because many apps have this property in their subclasses.

/Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKit/ASEditableTextNode.mm:208:13: error: type of property 'delegate' ('id<ASEditableTextNodeDelegate,ASDisplayNodeDelegate>') does not match type of instance variable '_delegate' ('__weak id<ASEditableTextNodeDelegate>')

This is an interesting idea, and indeed one I almost added a couple days ago when desiring to monitor the new interfaceStateDidChange:fromState: and visibilityDidChange: notifications from ASViewController's base class.  I haven't reviewed fully yet and am not yet sure what the name should be, but would be curious to hear ideas!  We should probably think ahead to as many plausible usages of the delegate as possible to see if there is a name that can encompass them which adds at least one more word alongside delegate in the property name.

levi:
interfaceDelegate, layoutDelegate, layoutContext, displayHandler, to name a few suggestions

appleguy:
@levi one concern - if this delegate is public, which it probably should be, then ASViewController can't necessarily take it over.  If it did, it would fail if a user were to set the delegate of the node to their object later...

levi:
Can't the same argument be made about CALayer#delegate?

appleguy:
@levi Definitely interested to look to Apple APIs for comparison / precedent.  CALayer's delegate is set to UIView when attached to a view, but this is prominently documented (in UIView) - and more importantly, the CALayer delegate has essentially no functionality that can't be handled in UIView itself (https://developer.apple.com/library/ios/documentation/QuartzCore/Reference/CALayerDelegate_protocol/index.html#//apple_ref/doc/uid/TP40012871)  Also, UIView and CALayer are very similar objects conceptually, so subclassing one and doing the operations there is not as strange or unexpected as requiring the implementation be in a different type of object (ASViewController vs ASDisplayNode).

This consideration simple underscores the expected value of this delegate - not only the initial set of features, but in the future it could become quite an essential construct!  The solution is probably just an ASDelegateProxy owned by ASViewController, in addition to a name besides "delegate".

levi:
Back from vacation. Totally forgot about this PR. Will look at it over the weekend to see how to bring it up to speed.

levi:
OK, I just pushed some updates that bring this patch up to speed with the feedback and also provides unit test coverage for the new functionally. I'm having some issues running the entire test suite when it comes to the snapshot tests. They seem to be looking for an image asset to compare to, but unable to find one. I haven't looked too closely into the issue, but hoping this is something others have encountered (cc/ @appleguy, @nguyenhuy).

nguyenhuy:
Assuming you are having problems with existing tests, definitely check for the architecture that you are running the tests on. Reference images are available for 64bit only.

Great to see a PR that includes tests! Definitely way to go haha :dancer: :dancer: :dancer:

levi:
@nguyenhuy, hrm it looks like I'm running them in 64bit locally. Are you able to pull the patch down on your end to confirm?

nguyenhuy:
Are you running on a @3x devices (iPhone 6S and 6S+)? We don't have @3x snapshots yet. Should be easy to add though.

nguyenhuy:
For the record, I pulled your patch and all tests passed on a @2x simulator.

rcancro:
I just took a quick glance at this, but wondered since ASViewController will probably be used most often as a base class for a custom VC, does it make sense to add this functionality more like the way UIViewController does? i.e., with methods like:

`- (ASLayoutSpec *)layoutSpecForSubnodes;`
`- (ASLayout *)layoutForSubnodes;`

that a subclass can implement?

levi:
Thanks for the feedback, @rcancro. Are you specifically asking for a naming change? As far as the existing functionality stands, subclasses of ASViewController will be able to override these methods to be used in the way that you're describing.

appleguy:
Continuing to discuss this offline, I'm going to close this for now but we can certainly find it to reopen or pull the code as a base for a further change!

PR #1003, opened 2015-12-27
Fix an issue, ASNetworkImageNode can't load image from bundle via NSURL.

PonyCui:
Issue report and PR.

If only provide a file URL to ASNetworkImageNode.URL, it will not load the bundle image.
But if you deliver an UIImage to ASNetworkImgaeNode, that's not good.
The problem is, imageNode shouldCacheImage default values is true, but [UIImage imageNamed:_URL.path] requires an pure file name without any other characters.
See ASNetworkImageNode.mm:194, that's the issue code. If you use ASNetworkImageNode.mm:195 code, the issue solved.
This issue can be found at [https://github.com/PonyCui/AsyncDisplayKit/tree/ASNetworkImageNode-BundleImageIssue](https://github.com/PonyCui/AsyncDisplayKit/tree/ASNetworkImageNode-BundleImageIssue) branch, examples/ASNetworkImageNode-BundleImageIssue directory.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -191,7 +191,7 @@
           _imageLoaded = YES;
 
           if (self.shouldCacheImage) {
-            self.image = [UIImage imageNamed:_URL.path];
+            self.image = [UIImage imageNamed:[_URL.path stringByReplacingOccurrencesOfString:@"/" withString:@""]];
           } else {
             self.image = [UIImage imageWithContentsOfFile:_URL.path];
           }


appleguy:
Thanks so much for investigating this!  I haven't used this component much with file URLs, but it is a fine idea.

I will review your patch in the next few days.

appleguy:
@PonyCui Thanks to your test case, I was able to confirm that the fix I believe is more correct still works for your case.  Feel free to check out the code change that I've made.  Most importantly, please test it with any other code you have (such as any main project you're using where you first discovered this) to ensure it works as you expect.

File a new GitHub issue for any other problems or questions you have, and feel free to email me at asyncdisplaykit@gmail.com any time :).

https://github.com/facebook/AsyncDisplayKit/pull/1006

Issue #991, opened 2015-12-25
[ASDisplayNode] When using UIViewController directly, ensure contents is not cleared during nav push.


tuyuan2012:
When a controller pushes to another controller, the ASNetworkImageNode which is added to a view calls 'willMoveToWindow' method.

The imageNode will call __exitHierarchy so that the its image is clear. When the controller pops out. The imageNode will call enterHierarchy so that its image is display. But it will flash in case of re-display..

The git url of this case is https://github.com/tuyuan2012/LearnASDNode. You can feel the description in the case.

appleguy:
@tuyuan2012 thanks for the test project!  I'm generally aware of this issue, but it's hard to prevent in all cases that people handle nodes — still it is a valid concern and I will double down my efforts to try to make this less likely to happen for users not aware of being careful not to trigger this.  Using something like ASViewController, ASTableView, ASCollectionView should avoid this - but it should really work intuitively / automatically in as many cases as are technically possible.

PR #1005, opened 2015-12-27
[ASDisplayNode] Preserve contents after non-range-managed nodes are removed from superviews or windows.

appleguy:
This behavior changed in 1.9.3 and introduced flickering in some cases.  Preserving the contents is closer to UIKit behavior.

Thank you @tuyuan2012 for reporting this in https://github.com/facebook/AsyncDisplayKit/issues/991, and providing a test project that made it easier to fix in the most correct way!  @Adlai-Holler, heads up regarding this behavior - I think you may have been expecting it to clear the contents here, so make sure your app calls -clearFetchedData and/or -clearContents manually if so.  I am thinking of other ways to refine this behavior in the future, such as suspending updates to interfaceState in certain situations and then un-suspending them to apply a pending state.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1775,18 +1775,22 @@ static BOOL ShouldUseNewRenderingRange = NO;
     if (newState & ASInterfaceStateFetchData) {
       [self fetchData];
     } else {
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
Ah understood, thanks for the heads up man!

PR #1008, opened 2015-12-28
Bridge UITableViewCell pointInside to ASNodeCell

yury:
Allow ASNodeCell to specify pointInside of UITableViewCell.
This is very usefull, if ASNodeCell is presented as bubble aligned to
left or right (like Messages.app) and we need to be able to select row
only if user taps on bubble.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -75,6 +75,10 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   _node.highlighted = highlighted;
 }
 
+- (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event {
+    return [_node pointInside:point withEvent:event];
+}
+
 @end
 
 #pragma mark -
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -75,8 +75,9 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   _node.highlighted = highlighted;
 }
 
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASTableView.mm
```diff
@@ -75,6 +75,10 @@ - (void)setHighlighted:(BOOL)highlighted animated:(BOOL)animated
   _node.highlighted = highlighted;
 }
 
+- (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event {
```
appleguy: Great idea!  Should we bridge hitTest too?  Will that work correctly with the logic already in ASDisplayNode that is used for hitTesting, if we bridge it directly to the node without a special implementation? (it should)

Bump the { on the next line to match the style, and I'm happy to merge this.  Optionally, add a comment to the header indicating that this is supported.

yury:
@appleguy  Please take a look at [sample project](https://github.com/yury/BubbleCells). Works fine for me.

`hitTest` seems not in ASCellNode

appleguy:
@yury isn't hitTest implemented in ASDisplayNode, just like pointInside?  Thanks for the brace fix!

appleguy:
I'll go ahead and merge this now so it makes the 1.9.4 release (launch for that release is looking like tomorrow evening - I don't want to rush my testing / writing the release notes, and it is getting late tonight for what I have to do in the morning).

If you do have the time to try it out, it would be great to see the hitTest version added unless you predict a problem that could cause.  It should be another diff almost identical to this one, very short one-line implementation.

appleguy:
@yury I can't believe I didn't mention / notice this before, but - even more important than the hitTest: version, which is very helpful to be able to use something like hitTestSlop or other uses for hitTest: going to a specific subview, this same pointInside implementation should be added to the equivalent place in ASCollectionView!

PR #1009, opened 2015-12-28
[ASDisplayNode] Allow UIKit / CA Bridged Properties to Always Be Writeable On Background Threads

Adlai-Holler:
@appleguy This is a first-pass implementation, I'm not satisfied with it for a number of reasons, and I don't expect to merge it. But this will get the ball rolling and hopefully expose important questions. 

The major problems I have with this branch are:
- If the user directly sets properties on the underlying view, we won't find out about it. If they read the corresponding node property, we'll return the old value. This is a tough problem, because we HAVE to buffer property values outside the view to make them readable from any thread. I can't see any solution to this except swizzling the UIKit setters :scream:.
- Order-of-operations is not preserved. If you're in the middle of buffering main-queue commands on thread A and you set a property to X, and then on thread B (while not in a transaction) you set it to Y, and then A flushes the main queue command buffer, the property will get set to X when it should be Y. What's worse, the pending-state value and the view-value will be out-of-sync. This one shouldn't be hard to solve – we should always store the pending value in the pending state, and then always flush THAT rather than just enqueuing individual property sets.
- Related to above, hard to debug since we don't explicitly track the pending state of a node (in fact, such a state barely exists).

For the next pass here's the plan:
- Construct a "weak set" using a weak-to-strong-objects NSMapTable that goes `ASDisplayNode : [NSNull null]`. 
- Create a singleton that has a weak set of nodes that have pending state. When you set a bridged node property off-main on a loaded node, if the pending state was previously empty, we add this node into that set. 
- For setting 
- For the flush, the singleton just calls `applyToView:` or `applyToLayer:` on all the dirty nodes' pending states.
- For transactionalization, I think all we need is to ensure that there are no open transactions when the flush happens, right?
 - Let's say we have four background threads A, B, C, and D:
  - Thread A opens a transaction and starts setting some properties.
  - Thread B opens a transaction and starts setting some properties.
  - Thread C sets a property outside any transaction.
  - Thread A's transaction ends.
  - Thread B's transaction ends.
  - We notice there are no transactions and no flush enqueued. Enqueue a flush on the main thread.
  - Thread D opens a transaction.
  - Main thread: it's time to flush. Lock the singleton (@appleguy dangerous?). But oh! There's an open transaction. Reject flush.
  - Thread D closes its transaction.
  - We notice there are no transactions and no flush enqueued. Enqueue a flush on the main thread.
  - Main thread: it's time to flush. Lock the singleton, confirm no open transactions, let's do this. Apply all pending states to nodes in weak set, empty the set, unlock the singleton.
 - Would it be valuable to have an option to tell the singleton "flush everything NOW" on the main thread? 
I think this would be doable if we accept that the flush may get rejected. Maybe something like `flushChangesWithCompletion:^{}`, so we can wait for any open transactions to end?

Maybe I'm making this way more complicated than it needs to be.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -451,6 +451,10 @@
 		B350625F1B0111800018CF92 /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 058D09AF195D04C000B7D73C /* Foundation.framework */; };
 		C78F7E2A1BF7808300CDEAFC /* ASTableNode.m in Sources */ = {isa = PBXBuildFile; fileRef = B0F880591BEAEC7500D17647 /* ASTableNode.m */; };
 		C78F7E2B1BF7809800CDEAFC /* ASTableNode.h in Headers */ = {isa = PBXBuildFile; fileRef = B0F880581BEAEC7500D17647 /* ASTableNode.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		CC4CA2D91C2F73AE003D381F /* ASMainQueueTransaction.h in Headers */ = {isa = PBXBuildFile; fileRef = CC4CA2D71C2F73AE003D381F /* ASMainQueueTransaction.h */; };
+		CC4CA2DA1C2F73AE003D381F /* ASMainQueueTransaction.h in Headers */ = {isa = PBXBuildFile; fileRef = CC4CA2D71C2F73AE003D381F /* ASMainQueueTransaction.h */; };
+		CC4CA2DB1C2F73AE003D381F /* ASMainQueueTransaction.m in Sources */ = {isa = PBXBuildFile; fileRef = CC4CA2D81C2F73AE003D381F /* ASMainQueueTransaction.m */; };
+		CC4CA2DC1C2F73AE003D381F /* ASMainQueueTransaction.m in Sources */ = {isa = PBXBuildFile; fileRef = CC4CA2D81C2F73AE003D381F /* ASMainQueueTransaction.m */; };
 		CC7FD9DE1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.h in Headers */ = {isa = PBXBuildFile; fileRef = CC7FD9DC1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		CC7FD9DF1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.m in Sources */ = {isa = PBXBuildFile; fileRef = CC7FD9DD1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.m */; };
 		CC7FD9E11BB5F750005CCB2B /* ASPhotosFrameworkImageRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CC7FD9E01BB5F750005CCB2B /* ASPhotosFrameworkImageRequestTests.m */; };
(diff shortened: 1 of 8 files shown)


appleguy:
Woah, thanks @Adlai-Holler!!  Even just getting this started is **huge**.  I have a handful of comparably risky / ambitious changes planned for right after 1.9.4 is cut (any day now, for reals :) ), and will keep running Pinterest internal against the latest to get us fairly substantial test coverage on these things.  @lappp9 has the programming guide / 2.0 documentation underway, and we plan to add more tests soon too.

We should catch up soon about one of the most important use cases in the framework: ASDataController / ASRangeController and its possible successors / improvements.  Most relevant to this, I want to remove the current checks for isNodeLoaded doing main thread measurement, and I also want things like reloadSection to first process the new state and then apply the updates rather than deleting the existing one initially and then adding the replacement when it's done.

Somewhat more tangentially, there's also my upcoming change to support a "cellNodeBlock" so that the block can capture call-time data state and yet be used to allocate the cell node on a background thread, along with the allocation range / measurement range / virtualization stuff we discussed (in service of the need to measure only slightly more than is required to populate a given rect, allowing more liberal use of synchronously preparing those areas).

But lastly and also relevant to the idea of a transactional system we might expand here, there is the increasingly-essential (for my use cases at least) concept of transparent "runloop amortization", where the materialization of views and layers is spread out over multiple runloop turns to allow the system opportunities to service display link timers (physics-based animations) or gestures or network callbacks.  This is tricky because it will also involve applying the properties to those views and we can't allow the sub-hierarchy onscreen until it is fully materialized, even if a later call does force it to be.  It will be most useful when the ASRangeController has a cell node entering the display range, or a view controller wants to pre-load its contents, but once the cell is visible or the controller is being presented it is too late to use this technique (unlike synchronous concurrency, which blocks main on a concurrent background operation).  So we will need the ability to force-finish in one runloop an operation that is in the process of being amortized across multiple loops, and that's where waiting on a transaction type setup (calling "run" or "finish" on the transaction, perhaps, if the contents of that transaction is responsible for poking the view / layer and applying pending state).

Rambling complete, I'm now leaving the Chipotle in Central California (Grapevine, CA) to continue the drive home to SoCal :).  Hope to see you again soon - I'll be online and working for the next week if you want to talk by FaceTime Audio about this.  Thanks for the delightful surprise and ideas to think about on the drive!

Adlai-Holler:
Hey @appleguy just wanted to let you know I haven't forgotten or bailed on this! I think about it on the daily but I haven't yet set aside the time to bang it out. I'll execute on it before Friday.

Most of my musing has been about run loop amortization. What do your use cases look like? Are you thinking of We could make a generalized controller that amortizes an array of main-queue-only blocks by, say, running max 5 blocks per turn. The downside here is that the controller has no idea how much work each block contains, so 5 blocks could mean 1000 units of work or 10. Would it be worthwhile to let the client provide an optional `cost` param? Probably not. It's a small downside though, since it's a private class so we'll be able to tune performance for our use cases anyway.

Or we could make a specialized controller, that uses knowledge about the work itself to decide how to divide the work up. I don't really like this idea but I want to throw it out. This would basically merge the run-loop amortization code with the property buffering code, and decide how to break up the work based on "which nodes' properties to push per turn of the run loop."

As for transactionalizing and flushing this stuff if we run out of time, am I understanding it right that the transaction would contain `[node flushPendingState]` i.e we may actually apply changes that were made outside of our transaction "block"? It seems like we have to accept this, because the alternative is storing separate pending states for separate transactions :scream:.

Want to talk data controller & range controller this week? I could do Wednesday or Friday morning/afternoon.

Adlai-Holler:
The party has moved to #1048

PR #1016, opened 2015-12-30
Ensure that the uncommon __unloadNode codepath does not unintentionally trigger node removal.

appleguy:


appleguy:
This ensures that shouldRasterizeDescendants can be enabled and disabled at app runtime even after a full sub-hierarchy of nodes has been loaded.  This wasn't supported by ASDK for years, until early October 2015, and was recently broken by one of my changes about a week ago in master..

Issue #1011, opened 2015-12-28
[ASCellNode] interface state sometimes doesn't update


knopp:
This happens when

`[UICollectionView scrollToItemAtIndexPath:atScrollPosition:animated:NO]` is called from within the `[ASCollectionView reloadDataWithCompletion:]` block

i.e.

```objectivec
[collectionView reloadDataWithCompletion:^{
    [collectionView scrollToItemAtIndexPath:activeIndexPath
                           atScrollPosition:UICollectionViewScrollPositionCenteredVertically animated:NO];
}];
```

the problem seems to be caused by `[ASRangeController _updateVisibleNodeIndexPaths]`
```objectivec
// Notify to remove indexpaths that are leftover that are not visible or included in the _layoutController calculated paths
NSMutableSet *removedIndexPaths = _rangeIsValid ? [_rangeTypeIndexPaths[rangeKey] mutableCopy] : [NSMutableSet set];
```
in this instance, `_rangeIsValid` is set to false, because items are being still added, and thus `removedIndexPaths` will always be set to empty set; `addedIndexPaths` works properly because it doesn't check for `_rangeIsValid`;

Why this inconsistency? Is the `_rangeIsValid`? check really necessary when determining `removedIndexPaths`? When I remove it, i get correctly reported interface state
```objectivec
// Notify to remove indexpaths that are leftover that are not visible or included in the _layoutController calculated paths
NSMutableSet *removedIndexPaths = [_rangeTypeIndexPaths[rangeKey] mutableCopy];
```

appleguy:
@knopp Good bug - thank you for reporting this.

[...]

PR #1020, opened 2015-12-31
Ensure that ASRangeController immediately removes any deleted nodes from its range state.

appleguy:
This is a fix for #1011.  @knopp, give it a try when you have a chance!  I think I'll include this in 1.9.4.


knopp:
I just tried this, works for me! Thanks!

PR #1022, opened 2015-12-31
Prevent calling unsupported UICollectionViewDelegate methods (supplementary view appearance)

appleguy:
cc @levi, @stowy maybe we can fix this at some point soon, but for now, away they go - they're not safe to use unless reconciled with the async layout delay.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.m
+++ b/AsyncDisplayKit/ASCellNode.m
@@ -63,13 +63,6 @@
   return self;
 }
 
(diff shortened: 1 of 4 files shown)


levi:
The current supplementary node support doesn't interface with the range controller at the level of a regular collection view item. We should use these methods to control the interface state of supplementary nodes.

stowy:
thanks for the clarification and the update @appleguy

PR #1025, opened 2016-01-01
[ASButtonNode] Add support for backgroundImage.  Extend & simplify API with separate title and color.

1nput0utput:
Align the ASButtonNode API with that of UIButton. These new methods will ease the pain of creating NSAttributedString when you only really want is set a title or a colour on ASButtonNode.

### DIFF
--- a/AsyncDisplayKit/ASButtonNode.h
+++ b/AsyncDisplayKit/ASButtonNode.h
@@ -47,7 +47,7 @@
 - (NSAttributedString * _Nullable)attributedTitleForState:(ASControlState)state;
 
 /**
- *  Sets the styled title to use for the specified state.
+ *  Sets the styled title to use for the specified state. This will reset styled title previously set with -setTitle:withFont:withColor:forState.
  *
  *  @param title The styled text string to use for the title.
  *  @param state The state that uses the specified title. The possible values are described in ASControlState.
@@ -55,7 +55,7 @@
 - (void)setAttributedTitle:(nullable NSAttributedString *)title forState:(ASControlState)state;
 
 /**
- *  Sets the title to use for the specified state.
+ *  Sets the title to use for the specified state. This will reset styled title previously set with -setAttributedTitle:forState.
  *
  *  @param title The styled text string to use for the title.
  *  @param font The font to use for the title.
--- a/AsyncDisplayKit/ASButtonNode.h
+++ b/AsyncDisplayKit/ASButtonNode.h
@@ -11,8 +11,9 @@
 
 @interface ASButtonNode : ASControlNode
 
-@property (nonatomic, readonly) ASTextNode *titleNode;
-@property (nonatomic, readonly) ASImageNode *imageNode;
+@property (nonatomic, readonly) ASTextNode  * _Nonnull titleNode;
+@property (nonatomic, readonly) ASImageNode * _Nonnull imageNode;
+@property (nonatomic, readonly) ASImageNode * _Nonnull backgroundImageNode;
 
 /**
  Spacing between image and title. Defaults to 8.0.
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -126,6 +140,25 @@ - (void)setLaysOutHorizontally:(BOOL)laysOutHorizontally
   [self setNeedsLayout];
 }
 
+
+- (void)setTitle:(NSString *)title forState:(ASButtonState)state
+{
+  UIColor *currentTitleColor = [self titleColorForState:state];
+  NSDictionary *attributes = @{
+                               NSFontAttributeName: [UIFont systemFontOfSize:[UIFont buttonFontSize]],
```
appleguy: I like the idea of the simplified setTitle:, but I'm afraid that very few apps will want to just use the off-the-shelf buttonFontSize.  I don't think the API as it stands adds enough convenience with sufficient expressive power, at least to justify the number of added methods.

I wonder if a setTitle:withFont:color:forState: would be a solution.  Not sure those are the right parameter names / grammar for the method, but it seems closer to optimal?

1nput0utput: @appleguy I'd replace the setTitle with setTitle:withFont:withColor:forState: 

Would you keep the getters for each parameters (titleForState:, titleColorForState, andtitleFontForState:)? The attributedTitleForState: has those values but it's not so elegant from the user's perspective.

1nput0utput:
@appleguy finally got some time to make these small changes. I've removed all the convenient methods for title, only kept one, ```setTitle:withFont:withColor:forState```.

@wy19901227 here's the background image for ASButtonNode.

appleguy:
@1nput0utput thanks so much for your work on this!

AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -41,10 +47,12 @@ - (instancetype)init
 
     _titleNode = [[ASTextNode alloc] init];
     _imageNode = [[ASImageNode alloc] init];
+    _backgroundImageNode = [[ASImageNode alloc] init];
```
appleguy: @1nput0utput  Just realized this - but indeed we should be able to set .isLayerBacked = YES on all of these!  Because the ASButtonNode itself is an ASControlNode, it will do all of the touch handling that we need in this class.  Doing this would significantly reduce the overhead of this simpler design of creating all three of these nodes up front even if a given use case only requires one of them.

appleguy:
This looks very clean.  I think it can land as-is, but I haven't been able to fully complete a review yet.  I should be able to do so within the next few days.  If you can drop in isLayerBacked = YES, great - but that is not at all a blocker and I can do it later pretty easily.

@nguyenhuy / @rcancro / @garrettmoon if any of you get a chance to review this code, go for it, and I'm happy to merge if you give it the go-ahead.

AsyncDisplayKit/ASButtonNode.h
```diff
@@ -36,11 +37,66 @@
...
 
-- (NSAttributedString *)attributedTitleForState:(ASControlState)state;
-- (void)setAttributedTitle:(NSAttributedString *)title forState:(ASControlState)state;
+/**
+ *  Sets the styled title to use for the specified state.
+ *
+ *  @param title The styled text string to use for the title.
+ *  @param state The state that uses the specified title. The possible values are described in ASControlState.
+ */
+- (void)setAttributedTitle:(nullable NSAttributedString *)title forState:(ASControlState)state;
```
appleguy: We need to clearly describe on this as well as the title + color APIs that the two are mutually exclusive / overwrite each other, right?

1nput0utput:
@appleguy Thanks for the feedback! As you mentioned the subnodes should be layer backed, I've also considered checking that no one sets .isLayerBacked = YES on an ASButtonNode itself, as consequence it will not receive any touch events. I wonder though, should this check be in ASControlNode??

AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -41,33 +47,50 @@ - (instancetype)init
...
+    [self addSubnode:_backgroundImageNode];
     [self addSubnode:_titleNode];
     [self addSubnode:_imageNode];
   }
   return self;
 }
 
+- (void)setLayerBacked:(BOOL)layerBacked
+{
+  ASDisplayNodeAssert(!layerBacked, @"ASButtonNode must not be layer backed!");
```
appleguy: Interesting point with this one.  I'd actually suggest removing this — the reason is that it's not unreasonable to imagine a button that doesn't need a tap state and has a higher level node do a hitTest: to check if it falls within the button.  Buttons are often at the leaves of the hierarchy and so if all of them are forced to be view backed, it would potentially limit more optimized apps from layer-backing big chunks of the hierarchy.

The sad thing is that you're almost certainly right that as many or more developers will hit this unexpectedly and wonder why their button doesn't work, as do this intentionally.  That's mostly a documentation problem, though, and probably best addressed some other way.

AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -258,12 +362,18 @@ - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
...
+                                                        background:self.backgroundImageNode];
+  } else {
+      return stack;
+  }
 }
 
 - (void)layout
 {
   [super layout];
+  self.backgroundImageNode.hidden = self.backgroundImageNode.image == nil;
```
appleguy: In an ideal world, we'd not allocate and add all of the subnodes until we know they're needed - but there are some complexities with doing that if e.g. the title were set in the background after initial loading of the views / layers.  This is probably fine for now especially with layer-backing now on (less in-app CPU time to create 3 layers than 2 views - by a mile), but the layers really add up in terms of CATransaction cost, and the much-harder-to-measure render server per frame recursion cost to set up the OpenGL compositing of the frame.  Even if hidden, they add measurable cost to the hierarchy scanning.

appleguy:
@1nput0utput Sorry it took a while to review, but I'm really happy to have this landing now!  Very much appreciate your help over the past few months.

PR #1026, opened 2016-01-02
Fix ASControlNode mutation crash

1nput0utput:
Mutation crash in ```- (void)sendActionsForControlEvents:(ASControlNodeEvent)controlEvents withEvent:(UIEvent *)event```

The crashed is due to removing a target within the selector body.
```
   func addTarget() {
        self.addTarget(self, action: "tapMe", forControlEvents:  .TouchUpInside)
    }

   func tapMe() {
        self.removeTarget(self, action: "tapMe", forControlEvents:  .TouchUpInside)
    }
```

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -1547,7 +1547,6 @@
 				058D09B9195D04C000B7D73C /* Frameworks */,
 				058D09BA195D04C000B7D73C /* Resources */,
 				3B9D88CDF51B429C8409E4B6 /* Copy Pods Resources */,
-				2AC85249A2221047409FABBB /* Embed Pods Frameworks */,
 			);
 			buildRules = (
 			);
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1646,6 +1647,21 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		2AC85249A2221047409FABBB /* Embed Pods Frameworks */ = {
```
appleguy: Can you remove this change?  I think it may be added by a newer version of Cocoapods, but we can't move off of 0.37.2 yet due to compatibility issues.  Looking forward to those guys finally getting the 1.0 out.

AsyncDisplayKit/ASControlNode.m
```diff
@@ -339,7 +339,7 @@ - (void)sendActionsForControlEvents:(ASControlNodeEvent)controlEvents withEvent:
   _ASEnumerateControlEventsIncludedInMaskWithBlock(controlEvents, ^
     (ASControlNodeEvent controlEvent)
     {
-      NSMapTable *eventDispatchTable = [_controlEventDispatchTable objectForKey:_ASControlNodeEventKeyForControlEvent(controlEvent)];
+      NSMapTable *eventDispatchTable = [[_controlEventDispatchTable objectForKey:_ASControlNodeEventKeyForControlEvent(controlEvent)] copy];
```
appleguy: Could you add a comment above this, in the style of the file explaining the rationale?  It's not immediately obvious that the action perform would trigger a call into remove, but of course makes sense with a few words of description.

I'm more worried about any performance implication of this, though; copy is probably faster than adding items for the first time, but I've seen the latter prominently in profiling the display system's usage of NSMapTable.

It's just a shame that probably ~10,000+ events will be delivered for 1 that causes an action removal, since most apps won't do this.  It could make sense to set a boolean flag before iterating, and if that flag is encountered at the mutation point, perform the copy and overwrite the local variable with the copy (as well as clear the flag).  I'll land this PR with the project file revert and comment though.

appleguy:
@1nput0utput Thank you for upstreaming the fix to this after encountering it!

AsyncDisplayKit/ASControlNode.m
1nput0utput: @appleguy I tried to see if a boolean flag would help, unfortunately the iteration has already began   , checking the flag doesn't help. I did not yet see a clear solution.

appleguy:
Thanks @1nput0utput !

Issue #809, opened 2015-11-02
tvOS support


zdenektopic:
Hi,

I would like to ask if it is possible to use ASDK with tvOS or if the support is planned/in the roadmap?

appleguy:
@zdenektopic great question and I'd be really curious to know the answer.  I have not looked deeply into tvOS.

Does it provide access to Core Animation / CALayer in addition to UIView?  Is the UIView API the same as the iOS UIKit one?  If both of those are true, it's very likely that it can be made to work with little effort.

clooth:
@appleguy It actually does give you support to those. I've already created a random project for tvOS with Core Animation and custom layer styling and such.

zdenektopic:
I can confirm that.

clooth:
With a very quick trial I managed to create a tvOS target and it *did* compile after I removed some things for the tvOS target:

- exclusiveTouch is not available on tvOS
- Photos.framework and AssetLibrary.framework are not available on tvOS

It did build after removing references to those in files like ASMultiplexImageNode(and Request) and some others as well as the target's linked frameworks.

appleguy:
Thanks a lot @clooth & @zdenektopic :).  I just created my first tvOS app today and am very very excited to get ASDK working on it.  Unfortunately I'm bogged down with a bunch of other stuff that I absolutely have to prioritize (even though it would be so much more satisfying to hack on this for the weekend, seriously hard to resist lol!!)

Please do let me know if you end up creating a sample app with it.  I definitely anticipated the Photos and AssetLibrary issues (@Adlai-Holler knows that code well and may be able to help us gate the frameworks).  exclusiveTouch is a funny thing to specifically remove...I could imagine a future TV having a bigger trackpad with multitouch...but fair enough, should also be easy to work around.

appleguy:
Pulling this forward to 2.1 as it is probably not horribly difficult and would be really awesome, so I'd like to see it happen in the next 2-3 months!

aaronschubert0:
tvOS 9.2 supports MapKit so this has got significantly cleaner! Now I don't need to gate out ASMapNode :)

appleguy:
@zdenektopic @clooth @aaronschubert0 - The community is planning an exciting long term road map for the project and getting organized around how to deliver these feature requests. We are moving to a new issue tracker with more sophisticated planning capabilities (check out www.realArtists.com and their new product: Ship!). 

[...]

PR #1036, opened 2016-01-05
[ASMapNode] now supports MKMapSnapshotOptions 

### LINKED ISSUE #988: [ASMapNode] Add support for setting MKMapSnapshotOptions objects.
I didn't realize until now all the options that MKMapSnapshotOptions provides - wow!

I'm pretty sure we can treat this object as thread safe, so it may actually be ideal to just change the region property into the MKMapSnapshotOptions object (I had not realized this is how the snapshotter itself gets the region until converting some Pinterest code to use ASMapNode, and then noticed it would be an even more convenient conversion if it just took the same object).

aaronschubert0:
@appleguy this will resolve #988 

Breaking Change:
`region` property has been superseded by `options` which is a MKMapSnapshotOptions object. This allows for more flexibility and customisation. All snapshot options will be applied to the MKMapView when it is activated.

Currently working on an example app and tests to better showcase ASMapNode!

### DIFF
--- a/AsyncDisplayKit/ASMapNode.h
+++ b/AsyncDisplayKit/ASMapNode.h
@@ -12,9 +12,9 @@
 @interface ASMapNode : ASImageNode
 
 /**
- The current region of ASMapNode. This can be set at any time and ASMapNode will animate the change. This property may be set from a background thread before the node is loaded, and will automatically be applied to define the region of the static snapshot (if .liveMap = NO) or the internal MKMapView (otherwise).
+ The current options of ASMapNode. This can be set at any time and ASMapNode will animate the change. This property may be set from a background thread before the node is loaded, and will automatically be applied to define the behavior of the static snapshot (if .liveMap = NO) or the internal MKMapView (otherwise).
  */
-@property (nonatomic, assign) MKCoordinateRegion region;
+@property (nonatomic, readwrite) MKMapSnapshotOptions *options;
 
 /**
  This is the MKMapView that is the live map part of ASMapNode. This will be nil if .liveMap = NO. Note, MKMapView is *not* thread-safe.
--- a/AsyncDisplayKit/ASMapNode.h
+++ b/AsyncDisplayKit/ASMapNode.h
@@ -12,9 +12,9 @@
 @interface ASMapNode : ASImageNode
 
 /**
- The current options of ASMapNode. This can be set at any time and ASMapNode will animate the change. This property may be set from a background thread before the node is loaded, and will automatically be applied to define the behavior of the static snapshot (if .liveMap = NO) or the internal MKMapView (otherwise).
+ The current options of ASMapNode. This can be set at any time and ASMapNode will animate the change.<br><br>This property may be set from a background thread before the node is loaded, and will automatically be applied to define the behavior of the static snapshot (if .liveMap = NO) or the internal MKMapView (otherwise).<br><br> Changes to the region and camera options will only be animated when when the liveMap mode is enabled, otherwise these options will be applied statically to the new snapshot. <br><br> The options object is used to specify properties even when the liveMap mode is enabled, allowing seamless transitions between the snapshot and liveMap (as well as back to the snapshot).
  */
-@property (nonatomic, readwrite) MKMapSnapshotOptions *options;
+@property (nonatomic, strong) MKMapSnapshotOptions *options;
 
 /**
  This is the MKMapView that is the live map part of ASMapNode. This will be nil if .liveMap = NO. Note, MKMapView is *not* thread-safe.
@@ -38,7 +38,7 @@
 @property (nonatomic, weak) id <MKMapViewDelegate> mapDelegate;
 
 /**
- * @discussion This method set the annotations of the static map view and also to the live map view. Passing an empty array clears the map of any annotations.
+ * @discussion This method sets the annotations of the static map view and also to the live map view. Passing an empty array clears the map of any annotations.
  * @param annotations An array of objects that conform to the MKAnnotation protocol
  */
 - (void)setAnnotations:(NSArray *)annotations;
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -39,11 +39,6 @@
   _needsMapReloadOnBoundsChange = YES;
   _liveMap = NO;
   _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
-  
-  _options = [[MKMapSnapshotOptions alloc] init];
-  //Default world-scale view
-  _options.region = MKCoordinateRegionForMapRect(MKMapRectWorld);
-  
   return self;
 }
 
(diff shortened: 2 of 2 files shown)


AsyncDisplayKit/ASMapNode.h
```diff
@@ -12,9 +12,9 @@
 @interface ASMapNode : ASImageNode
 
 /**
- The current region of ASMapNode. This can be set at any time and ASMapNode will animate the change. This property may be set from a background thread before the node is loaded, and will automatically be applied to define the region of the static snapshot (if .liveMap = NO) or the internal MKMapView (otherwise).
+ The current options of ASMapNode. This can be set at any time and ASMapNode will animate the change. This property may be set from a background thread before the node is loaded, and will automatically be applied to define the behavior of the static snapshot (if .liveMap = NO) or the internal MKMapView (otherwise).
  */
-@property (nonatomic, assign) MKCoordinateRegion region;
+@property (nonatomic, readwrite) MKMapSnapshotOptions *options;
```
appleguy: this should just say "strong" instead of readwrite (readwrite is the default anyway, and strong is actually the analogous to assign - it is the specifier of the memory behavior.  e.g. you could have an assign, readwrite)

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -41,11 +40,9 @@ - (instancetype)init
   _liveMap = NO;
   _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
   
-  //Default world-scale view
-  _region = MKCoordinateRegionForMapRect(MKMapRectWorld);
-  
   _options = [[MKMapSnapshotOptions alloc] init];
-  _options.region = _region;
+  //Default world-scale view
+  _options.region = MKCoordinateRegionForMapRect(MKMapRectWorld);
```
appleguy: Don't you need to allocate an Options first?  I think this will just fall on "nil" ears :).

The better route is probably to go to the point of use for the _options, and if it is nil there, create one with some default values.  I'd much rather we avoid creating objects in -init in framework classes if they're avoidable via some lazy strategy such as this.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -190,6 +186,15 @@ - (void)resetSnapshotter
   _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
 }
 
+- (void)applySnapshotOptions
+{
+  [_mapView setCamera:_options.camera animated:YES];
```
appleguy: Cool!  Pretty awesome that we can use the Options thing to store these kind of parameters even before the view is loaded, even if the only use of the map for that user is as a live map.

appleguy:
@aaronschubert0 This is really great, thanks for continuing with followups here!  I'm excited to hear that you are doing a sample app (definitely beyond my expectations).  If you happen to be interested in pursuing another ASDK project, let me know and I'll come up with some ideas for you to consider :).  You might also want to work with @lappp9 who is leading our documentation effort, including a large new sample app that will integrate ASMapNode and other components.

aaronschubert0:
@appleguy Wow! I'm definitely interesting in pursuing another ASDK project. I have just returned from the Apple TV Tech Talk so I'm feeling inspired to provide basic tvOS support for ASDK next (#809). I'd love to hear the ideas that you come up with just now so I can start having a think about them! :)

In terms of working with @lappp9 I'm more than happy to contribute to those efforts concurrently. Would it be better for the project if I spent the time I would have allocated to a sample app, to @lappp9 efforts instead?

In terms of this PR, I will post some follow up diffs shortly!

AsyncDisplayKit/ASMapNode.h
aaronschubert0: Ha! I was going against my gut here, opted for `strong` first but then seen other parts of the framework use `readwrite` so went for that. No problem, will fix that.

In terms of animating the properties that can be animated, will be. This animation obviously only occurs for the liveMap so will make these comments clearer, also highlighting which properties will animate.

AsyncDisplayKit/ASMapNode.mm
aaronschubert0: I totally agree. I'll move this to `-setUpSnapshotter` so that it is only allocated when needed. For the bigger picture of avoid creating objects in -init, I think that is a very worthwhile strategy to undertake, since often we init a node before we use it, so this will help :)

appleguy:
@aaronschubert0 Could you send me an email to asyncdisplaykit@gmail.com?  Would love to be in touch with you about your work on the project, to see how I can help, and organize the best intersection between your interests and priorities!

aaronschubert0:
@appleguy Just sent you an email :D

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -190,6 +183,23 @@ - (void)resetSnapshotter
...
+{
+  _options = [[MKMapSnapshotOptions alloc] init];
+  //Default world-scale view
+  _options.region = MKCoordinateRegionForMapRect(MKMapRectWorld);
+  _options.size = self.calculatedSize;
+}
+
+- (void)applySnapshotOptions
+{
+  [_mapView setCamera:_options.camera animated:YES];
```
appleguy: This should check if _options is nil, and either early return / not make any calls, or set default values.  Messaging nil for structure return values is actually not safe in Obj-C; it may be that your call points to this are gated, but we should do this anyway to protect against future calls being added.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -190,6 +183,23 @@ - (void)resetSnapshotter
   _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
 }
 
+- (void)createInitialOptions
+{
+  _options = [[MKMapSnapshotOptions alloc] init];
+  //Default world-scale view
```
appleguy: Ah, I guess this might prevent _options from being nil below, but...how about putting the nil check in applySnapshotOptions, and if _options is nil, then perform these three lines?  In that case you can actually delete -createInitialOptions as a method, which I don't think is really needed.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -180,7 +171,9 @@ - (void)takeSnapshot
 - (void)setUpSnapshotter
 {
   ASDisplayNodeAssert(!CGSizeEqualToSize(CGSizeZero, self.calculatedSize), @"self.calculatedSize can not be zero. Make sure that you are setting a preferredFrameSize or wrapping ASMapNode in a ASRatioLayoutSpec or similar.");
-  _options.size = self.calculatedSize;
+  if (!_options) {
+    [self createInitialOptions];
+  }
   _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:_options];
```
appleguy: Normally I prefer accessing instance variables and so as you have this written is probably more efficient, but because this won't be called frequently at all, maybe we should do self.options here and delete the nil check.  Then in the options accessor method, we could make sure that a nil options has the default options created for it.

appleguy:
Great improvement, thanks @aaronschubert0!  I'm going to get this in now to make the upcoming release so the API is updated.  Some minor refactoring requests here but not urgent, could also file a task so someone else comes back to revisit these comments later.

aaronschubert0:
Hey @appleguy ! Thanks for having a look at this and providing a thorough review as always. I'm happy to have a look at these changes and post a follow up diff on Monday. I like to finish what I started on this.

PR #1042, opened 2016-01-06
Make ASControlState imitate UIControlState

binl:
- selected state isn't implicitly controlled by ASControlNode anymore
- highlighted state can be overriden by users
- Decouple ASButtonNode's state from control events

@appleguy @connor

### DIFF
--- a/AsyncDisplayKit/ASButtonNode.mm
+++ b/AsyncDisplayKit/ASButtonNode.mm
@@ -53,23 +53,23 @@
 
 - (void)setEnabled:(BOOL)enabled
 {
-    [super setEnabled:enabled];
-    [self updateImage];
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -47,16 +47,29 @@ - (instancetype)init
...
 }
 
-- (void)controlEventUpdated:(ASControlNode *)node
+- (void)setEnabled:(BOOL)enabled
 {
-  [self updateImage];
-  [self updateTitle];
+    [super setEnabled:enabled];
+    [self updateImage];
+    [self updateTitle];
```
binl: oops.. I will need to fix spacing!

PR #1048, opened 2016-01-08
[ASDisplayNode] Eliminate Thread Affinity for Writing in Background - UIKit / CA Bridged Properties

Adlai-Holler:
Yo @appleguy here's something pretty neat. Let me know what you think!

- Optimize `_ASPendingState` by removing unneeded sets, and by loading all flags at once before doing batch comparisons.
- Add `ASWeakSet` that is built on top of `NSMapTable`. This does not support ordering, so if it's valuable to be able to flush nodes in the order that they were dirtied, we can change the approach to be built on top of `NSMutableOrderedSet` but it'll be a little less pleasant.
- Add `ASPendingStateController` singleton that keeps a weak set of dirty nodes and flushes them at the end of each turn of the run loop. You can also flush it manually at any time from the main queue (obviously if you want to flush from the background, just use `dispatch`.) It does not yet support dividing the work up into chunks. It'd be easy enough to only handle max N nodes per pass but I wanted to wait for your comments before diving into that optimization.
- Modify `ASDisplayNode` to always have a pending state. It no longer discards the pending state after node is loaded; instead it just clears the pending state's flags.
- Modify display node's `UIViewBridging` category to:
  - Always read properties from `ASPendingState`.
  - Always write properties to `ASPendingState`.
  - If properties are modified on main and node is loaded, properties are immediately propagated to view/layer.
  - If node is not dirty, but is loaded, and a bridged property is set off-main, we register the node with the pending state controller. Note that we register it before updating the pending state, but this is safe because we hold the node's property lock during this time and pending states can't be applied until lock is relinquished. 

I'm using `performSelectorOnMainThread:withObject:waitUntilDone:inModes` to schedule the run loop flush event, and it works well even when the change originated from an async dispatch onto a global concurrent queue. The docs are a little confusing here though – it mentions at one point that it dispatches onto "the run loop of the main thread" and then later that it registers with "the runloop of its current context". I can definitely swap it out with an `NSTimer` of duration 0 to act as the run loop observer but it's a bit less direct. Thoughts?

TODO:
- [x] Add more tests for the pending state controller.
- [x] Talk about back-propagation i.e. if user sets property on view/layer, what do we do? Only API-safe way to know about it is to swizzle UIView/CALayer properties which is a non-starter.
- [x] Talk about run loop amortization as an optimization.
- [x] Figure out why bridging `backgroundColor` won't work. I've tried migrating ASPendingState over to `UIColor *backgroundColor` but no dice. It's probably something stupid.


knopp:
@appleguy, @Adlai-Holler, assuming this gets finished, will it be eventually possible to relayout on background thread?

Adlai-Holler:
@appleguy Yo! This thing is so insanely close, but I'm having trouble taking it to the hoop.

ASCellNode overrides `setNeedsLayout`, and it's currently tied to the node's thread affinity because it's involved with the layout system. Whenever I fiddle with it to get it to be safe to call `[ASCellNode setNeedsLayout]` from the background, I break the table view tests.

So the state is, all tests are passing (including the new `ASBridgedPropertiesTests`) except the ones to do with cell nodes. Can we get a cell node expert in here to help me out? @nguyenhuy @levi

Adlai-Holler:
@appleguy Progress report: I've added a demo project to test the thread affinitylessness and resolved more issues. The demo app works well but I've got two issues remaining:

#### Layout Transitions

Running layout transition animations from the background doesn't work. The current issue is the implicit hierarchy management. After we make that thread-safe we'll need to (sometimes) asynchronously jump to the main thread to run the actual animation. That should be easy, although defining (and capturing) the "correct" initial/final animation could take a little thinking.

#### `setFrame:`

Bridging setFrame is troublesome. The current implementation works and maintains the old behavior very well, but there's duplicated code between ASDisplayNode and ASPendingState, and without thread affinity, there are order of operation edge cases – if I set frame then set bounds on a synchronous wrapper node, we should apply the update in the same order.

There are two modes for setting the frame on a node – synchronous wrapper nodes like ASTableNode need to receive setFrame: intact. For all other nodes, we prefer setting bounds and position so that the node's transform doesn't invalidate the frame. The problem is, we can't decompose a frame into bounds/position without reading the current bounds & anchorPoint, which we can't do in the background.

Currently the solution is to wait to decompose the new frame until `pendingViewState applyToView:`, which is working well but results in some extra code for ASPendingState to encapsulate the same logic.

#### Next Up

The only tests that are failing for me are the table view layout count tests. They all expect X layouts on the main thread but they're observing (X + 1) layouts. I'll investigate this tomorrow.

Adlai-Holler:
@appleguy Woot! This thing is ready for review! For QA the new BackgroundPropertySetting app and all tests are working great. 

setFrame ended up being pretty messy, and unfortunately we still can't layout loaded nodes off-main because of implicit hierarchy management. I'm scared to tackle modifying the hierarchy of loaded nodes off-main so I reckon we'll do that later.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -277,6 +277,7 @@ - (void)_initializeInstance
   _contentsScaleForDisplay = ASScreenScale();
   _displaySentinel = [[ASSentinel alloc] init];
   _preferredFrameSize = CGSizeZero;
+  _pendingViewState = [_ASPendingState new];
```
appleguy: Is this necessary? I have definitely seen overhead from the pending state object in traces, and there are nodes that don't end up using it at all.

Assuming it is necessary though, we should use the standard alloc / init simply because the new method is not used almost anywhere else in the framework.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -981,6 +981,28 @@ - (void)setContentsScaleForDisplay:(CGFloat)contentsScaleForDisplay
   _contentsScaleForDisplay = contentsScaleForDisplay;
 }
 
+- (void)applyPendingViewState
+{
+  ASDisplayNodeAssertMainThread();
+  ASDN::MutexLocker l(_propertyLock);
+
+  // FIXME: Ideally we'd call this as soon as the node receives -setNeedsLayout
```
appleguy: cc @levi, we should indeed fix this soon. I think we could probably separate out the addition of subnodes, after the node is loaded,. There is a great deal of value in ensuring the measurement pass occurs off the main thread, of course.

appleguy:
@Adlai-Holler I have not finished reviewing this yet, but will get to it tomorrow. I did apply it to the Pinterest app, and surprisingly I'm seeing a glitch in an existing animation that had not existed before -- somehow a behavior change is been generated, even though presumably the original behavior would be gated by any main thread assertions that previously existed.

I am still optimistic about landing this is in the next few days, but let's disable the desirable but problematic test and understand because of any behavior changes.  It is extremely close.

appleguy:
@Adlai-Holler is your app been broadly tested running on this? Give it a run through and see if you notice anything similar. We should also try a variety of the example apps.

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: OK cool, I'll load it lazily. Since we may use it often during the lifetime of a given node, do you think it's worth making the accessor a C function e.g. `ASDisplayNodeGetPendingState(ASDisplayNode *)`? Or should I use a normal Objective-C lazy accessor?

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: Totally agree! I only hesitate to do it in this PR because we're already 2500 lines from shore.

Adlai-Holler:
@appleguy Sucks to hear about the animation issue you're seeing – I haven't broadly tested Tripstr with this yet but I'll do so early tomorrow morning and post an update. In the meantime any info about the glitch you're seeing would be great!

Adlai-Holler:
@appleguy I've found a regression in Tripstr on the affinity branch. Investigating…

Adlai-Holler:
@appleguy OK we are freshly rebased, much less thread-affine, and a gargantuan bug in `setNeedsLayout` has been resolved. Issues in Tripstr and Kittens are all resolved. Bring on the Pinterest man!

AsyncDisplayKit/ASDisplayNode.mm
appleguy: Yes, I have realized progressively more how much of a difference this makes. A follow-up is fine, but it would be a nice optimization to use a C accessor.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1036,6 +1037,21 @@ - (void)__setNeedsLayout
   }
 }
 
+// If not rasterized (and therefore we certainly have a view or layer),
```
appleguy: This comment still seems relevant, but I believe it was originally in the bridge method… The one without two underscores.  It does not seem to be referencing the actions taken in the method here, even though it mentions schedule display that is intended to be from the position of -setNeedsDisplay calling this method and having some internal knowledge about the fact that it does that.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1858,25 +1874,25 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
...
 {
-  ASDisplayNodeAssertThreadAffinity(self);
+  ASDN::MutexLocker l(_propertyLock);
   return _layout;
 }
 
 - (CGSize)calculatedSize
 {
-  ASDisplayNodeAssertThreadAffinity(self);
+  ASDN::MutexLocker l(_propertyLock);
```
appleguy: Add a property lock call to the method below this, -constrainedSizeForCalculatedLayout

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -123,11 +123,8 @@ - (void)setImage:(UIImage *)image
   if (!ASObjectIsEqual(_image, image)) {
     _image = image;
 
-    ASDN::MutexUnlocker u(_imageLock);
```
appleguy: This is so awesome! Presumably we want to keep this unlocked call, however, or we will be holding the lock for the next two methods which is not required?

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -37,29 +40,46 @@
 #define __loaded (_layer != nil)
 
 #if DISPLAYNODE_USE_LOCKS
-#define _bridge_prologue ASDisplayNodeAssertThreadAffinity(self); ASDN::MutexLocker l(_propertyLock)
+#define _bridge_prologue_read ASDN::MutexLocker l(_propertyLock); ASDisplayNodeAssertThreadAffinity(self)
```
appleguy: read / write is a great separation to have here. Looking good, love this!

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -37,29 +40,46 @@
...
+#define _bridge_prologue_read ASDisplayNodeAssertThreadAffinity(self)
+#define _bridge_prologue_write
 #endif
 
+/// Returns YES if the property set should be applied to view/layer immediately.
+/// Side Effect: Registers the node with the shared ASPendingStateController if
+/// the property cannot be immediately applied and the node does not already have pending changes.
+/// This function must be called with the node's lock already held (after _bridge_prologue_write).
+ASDISPLAYNODE_INLINE BOOL ASDisplayNodeShouldApplyBridgedWriteToView(ASDisplayNode *node) {
+  if (ASDisplayNodeThreadIsMain()) {
```
appleguy: It might be worth adding an assertion here that checks that the lock is held. It is definitely a good idea to avoid creating another Mutex locker here, even though they are recursive, because even with the lock held this stack object would be unfortunate overhead

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -148,55 +168,55 @@ - (BOOL)canPerformAction:(SEL)action withSender:(id)sender
...
-  _bridge_prologue;
+  _bridge_prologue_read;
   return _getFromLayer(cornerRadius);
 }
 
 - (void)setCornerRadius:(CGFloat)newCornerRadius
 {
-  _bridge_prologue;
+  _bridge_prologue_write;
   _setToLayer(cornerRadius, newCornerRadius);
```
appleguy: My first thought upon seeing the read and write terminology was - what about get and set?  I think read and write is probably better, as they are more prominent and visibly different as they probably should be.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -214,9 +234,11 @@ - (CGRect)frame
 
 - (void)setFrame:(CGRect)rect
 {
-  _bridge_prologue;
-
-  if (_flags.synchronous && !_flags.layerBacked) {
+  _bridge_prologue_write;
+  BOOL setFrameDirectly = _flags.synchronous && !_flags.layerBacked;
+  BOOL isMainThread = ASDisplayNodeThreadIsMain();
+  BOOL nodeLoaded = self.nodeLoaded;
```
appleguy: This method is tremendously performance sensitive. Can you access the instance variable for nodeLoaded directly?

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -224,47 +246,31 @@ - (void)setFrame:(CGRect)rect
...
 #endif
+    _view.frame = rect;
+  } else if (!nodeLoaded || (isMainThread && !setFrameDirectly)) {
+    /**
+     * Sets a new frame to this node by changing its bounds and position. This method can be safely called even if
+     * the transform is a non-identity transform, because bounds and position can be set instead of frame.
+     * This is NOT called for synchronous nodes (wrapping regular views), which may rely on a [UIView setFrame:] call.
+     * A notable example of the latter is UITableView, which won't resize its internal container if only layer bounds are set.
+     */
+    CGRect bounds;
```
appleguy: I know it is not required, but it would be nice to in initialize these to zero values.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -224,47 +246,31 @@ - (void)setFrame:(CGRect)rect
     // Checking if the transform is identity is expensive, so disable when unnecessary. We have assertions on in Release, so DEBUG is the only way I know of.
     ASDisplayNodeAssert(CATransform3DIsIdentity(self.transform), @"-[ASDisplayNode setFrame:] - self.transform must be identity in order to set the frame property.  (From Apple's UIView documentation: If the transform property is not the identity transform, the value of this property is undefined and therefore should be ignored.)");
 #endif
+    _view.frame = rect;
+  } else if (!nodeLoaded || (isMainThread && !setFrameDirectly)) {
```
appleguy: The logic of this conditional is too complex for my liking, Particularly for critical code, And for having three different clauses. It is not immediately obvious if there is a situation where none of the clauses would run, Although I trusted that they do cover of the possibility space because of who wrote the code :)

Can you think of a way to modify it, perhaps with a higher-level check that forks based on any one of the three BOOLs?  Then you would have to large conditionals, and within each of those, at most two others (but presumably 2 and 1)

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -224,47 +246,31 @@ - (void)setFrame:(CGRect)rect
...
- * A notable example of the latter is UITableView, which won't resize its internal container if only layer bounds are set.
- */
-- (void)__setSafeFrame:(CGRect)rect
-{
-  ASDisplayNodeAssertThreadAffinity(self);
-  ASDN::MutexLocker l(_propertyLock);
-  
-  BOOL useLayer = (_layer && ASDisplayNodeThreadIsMain());
-  
-  CGPoint origin      = (useLayer ? _layer.bounds.origin : self.bounds.origin);
```
appleguy: I'm pretty sure it was a profiled optimization to use the layer accessor when possible, and we should preserve this. Going through the view simply adds more method calls, because it must go to the layer anyway.

In general, the modifications to this frame code are my biggest concern with the patch so far. It seems very plausible that they are correct, but I think it adds risk to an already large diff that is not strictly necessary. If there is a way to start with the original code and make a minimum of modification so it is clear what has changed, that would be my preference. This won't block us from landing though, just have some nightmares about this code and don't feel like Messing around with it too much right now :)

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -224,47 +246,31 @@ - (void)setFrame:(CGRect)rect
...
-  
-  if (useLayer) {
-    _layer.bounds = bounds;
-    _layer.position = position;
-  } else {
     self.bounds = bounds;
     self.position = position;
+  } else if (nodeLoaded && !isMainThread) {
+    if (!_pendingViewState.hasChanges) {
+      [ASPendingStateController.sharedInstance registerNode:self];
```
appleguy: Is there a reasonable, and efficient, way to make use of the macros / _set* methods instead of handling the pending state controller directly?

If doing this directly means there are fewer method calls, we should keep it this way.

As a point of framework-specific style, avoid using dot syntax with class methods

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -282,212 +288,222 @@ - (void)setNeedsDisplay
       [rasterizedContainerNode setNeedsDisplay];
     });
   } else {
-    // If not rasterized (and therefore we certainly have a view or layer),
-    // Send the message to the view/layer first, as scheduleNodeForDisplay may call -displayIfNeeded.
```
appleguy: Bring this comment block back here, and remove it from the place that it was added to.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1036,6 +1037,21 @@ - (void)__setNeedsLayout
   }
 }
 
+// If not rasterized (and therefore we certainly have a view or layer),
+// Send the message to the view/layer first, as scheduleNodeForDisplay may call -displayIfNeeded.
+// Wrapped / synchronous nodes created with initWithView/LayerBlock: do not need scheduleNodeForDisplay,
```
appleguy: Okay, let's actually split this common block in half. For this one, let's use this:

// Wrapped / synchronous nodes created with initWithView/LayerBlock: do not need scheduleNodeForDisplay,
// as they don't need to display in the working range at all - since the moment they are visible, any call to
// setNeedsDisplay (triggered by the caller of __setNeedsDisplay) will sync. display at transaction commit on main.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1036,6 +1037,21 @@ - (void)__setNeedsLayout
...
+// as they don't need to display in the working range at all - since at all times onscreen, one
+// -setNeedsDisplay to the CALayer will result in a synchronous display in the next frame.
+- (void)__setNeedsDisplay
+{
+  BOOL nowDisplay = ASInterfaceStateIncludesDisplay(_interfaceState);
+  // FIXME: This should not need to recursively display, so create a non-recursive variant.
```
appleguy: This is an important follow-up. I know this has existed for a couple months now, but could you add it to a list of follow-ups? I am probably the best person to deal with this, but it would be helpful to have the direct reminder as we talk about next steps following this.  a github issue would just get lost :)

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -282,212 +288,222 @@ - (void)setNeedsDisplay
...
-    _messageToViewOrLayer(setNeedsDisplay);
-
-    BOOL nowDisplay = ASInterfaceStateIncludesDisplay(_interfaceState);
-    // FIXME: This should not need to recursively display, so create a non-recursive variant.
-    // The semantics of setNeedsDisplay (as defined by CALayer behavior) are not recursive.
-    if (_layer && !_flags.synchronous && nowDisplay && [self __implementsDisplay]) {
-      [ASDisplayNode scheduleNodeForRecursiveDisplay:self];
+    if (self.nodeLoaded) {
+      if (ASDisplayNodeThreadIsMain()) {
+        _messageToViewOrLayer(setNeedsDisplay);
```
appleguy: @Adlai-Holler this condition actually does not seem correct to me. I believe we must call __setNeedsDisplay after messaging to view or layer.

Is this a scenario where we should be using ASDisplayNodeShouldApplyBridgedWriteToView?

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -282,212 +288,222 @@ - (void)setNeedsDisplay
...
 
 - (void)setNeedsLayout
 {
-  _bridge_prologue;
-  [self __setNeedsLayout];
-  _messageToViewOrLayer(setNeedsLayout);
+  _bridge_prologue_write;
+  if (self.nodeLoaded) {
+    if (ASDisplayNodeThreadIsMain()) {
+       [self __setNeedsLayout];
```
appleguy: // Quite the opposite of setNeedsDisplay, we must call __setNeedsLayout before messaging
// the view or layer to ensure that measurement and implicitly added subnodes have been handled.

This should make it just a bit clearer that the ordering is intentionally different than the case just above.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -282,212 +288,222 @@ - (void)setNeedsDisplay
...
+        _messageToViewOrLayer(setNeedsLayout);
+    } else {
+      if (!_pendingViewState.hasChanges) {
+        [ASPendingStateController.sharedInstance registerNode:self];
+      }
+      // NOTE: We will call [self __setNeedsLayout] just before we apply
```
appleguy: Why should this be done before the application of the pending state? It seems like during the measurement pass, some of the values in the pending state may be useful. I suppose the reading path should still provide access to those if the pending state exists and has not been cleared?

AsyncDisplayKit/Private/ASDisplayNodeInternal.h
```diff
@@ -25,7 +25,6 @@
 @class _ASDisplayLayer;
 
 BOOL ASDisplayNodeSubclassOverridesSelector(Class subclass, SEL selector);
-void ASDisplayNodeRespectThreadAffinityOfNode(ASDisplayNode *node, void (^block)());
```
appleguy: Woooo-hooooo!!!

AsyncDisplayKit/Private/ASInternalHelpers.h
```diff
@@ -49,6 +49,13 @@ ASDISPLAYNODE_INLINE void ASPerformBlockWithoutAnimation(BOOL withoutAnimation,
   }
 }
 
+ASDISPLAYNODE_INLINE void ASBoundsAndPositionForFrame(CGRect rect, CGPoint origin, CGPoint anchorPoint, CGRect *bounds, CGPoint *position)
+{
+  *bounds      = (CGRect){ origin, rect.size };
+  *position    = CGPointMake(rect.origin.x + rect.size.width * anchorPoint.x,
```
appleguy: Spacing could be improved -- less before the equals, one additional space on the second line of arguments to align the start of them

AsyncDisplayKit/Private/ASPendingStateController.h
```diff
@@ -0,0 +1,47 @@
...
+ have been set on background threads.
+ 
+ This controller will enqueue run-loop events to flush changes
+ but if you need them flushed now you can call `flush` from the main thread.
+ */
+@interface ASPendingStateController : NSObject
+
++ (ASPendingStateController *)sharedInstance;
+
+@property (nonatomic, readonly) BOOL hasChanges;
```
appleguy: flushScheduled might be more specific, but it doesn't read as nicely...

AsyncDisplayKit/Private/ASPendingStateController.mm
```diff
@@ -0,0 +1,110 @@
...
+  }
+  return self;
+}
+
++ (ASPendingStateController *)sharedInstance
+{
+  static dispatch_once_t onceToken;
+  static ASPendingStateController *controller;
+  dispatch_once(&onceToken, ^{
+    controller = [ASPendingStateController new];
```
appleguy: alloc / init;  *controller = nil

AsyncDisplayKit/Private/ASPendingStateController.mm
```diff
@@ -0,0 +1,110 @@
...
+  });
+  return controller;
+}
+
+#pragma mark External API
+
+- (void)flush
+{
+  ASDisplayNodeAssertMainThread();
```
appleguy: Should this method exist?  Maybe just put the assert in -flushNow and rename that to flush.

AsyncDisplayKit/Private/ASWeakSet.m
```diff
@@ -0,0 +1,76 @@
...
+}
+
+- (BOOL)containsObject:(id)object
+{
+  return [_mapTable objectForKey:object] != nil;
+}
+
+- (BOOL)isEmpty
+{
+  for (__unused id object in _mapTable) {
```
appleguy: lol!!! nice one

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -560,95 +558,104 @@ - (void)setAccessibilityIdentifier:(NSString *)newAccessibilityIdentifier
...
     layer.asyncdisplaykit_asyncTransactionContainer = asyncTransactionContainer;
 
-  if (_flags.setOpaque)
+  if (flags.setOpaque)
     ASDisplayNodeAssert(layer.opaque == opaque, @"Didn't set opaque as desired");
-  
-  if (_flags.setFrame)
-    ASDisplayNodeAssert(NO, @"Frame property should only be used for synchronously wrapped nodes.  See setFrame: in ASDisplayNode+UIViewBridge");
+
+  if (flags.setFrame) {
```
appleguy: I'm not totally sure it makes sense to give frame precedence, especially if both bounds and position are set. We may want to trigger an assertion if various overlapping properties are set in the same pending state. This is unfortunate, however, because they are reconciled by CA...

I'm not sure what to recommend here, but it speaks to my concern about the other Area of this patch that modified frame handling. It is by far one of the most tricky parts of the bridge, And is most definitely not as perfect as it could be.

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -660,144 +667,175 @@ - (void)applyToView:(UIView *)view
 
   CALayer *layer = view.layer;
 
-  if (_flags.setAnchorPoint)
+  ASPendingStateFlags flags = _flags;
+  if (flags.needsDisplay
+      || (flags.setOpaque && opaque != view.opaque)
+      || (flags.setBackgroundColor && !CGColorEqualToColor(backgroundColor, layer.backgroundColor))) {
```
appleguy: It seems like code of this nature should not appear in two places. Maybe we can use a #define  macro, or an in-line function?

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -660,144 +667,175 @@ - (void)applyToView:(UIView *)view
...
 
-  if (_flags.setShouldGroupAccessibilityChildren)
+  if (flags.setShouldGroupAccessibilityChildren)
     view.shouldGroupAccessibilityChildren = shouldGroupAccessibilityChildren;
 
-  if (_flags.setAccessibilityIdentifier)
+  if (flags.setAccessibilityIdentifier)
     view.accessibilityIdentifier = accessibilityIdentifier;
+
+  // FIXME: How should we reconcile order-of-operations between setting frame, bounds, position?
```
appleguy: Layer-backing is very much a first-class feature of the framework. Why is the behavior of this method different than the layer one? Can we consolidate and share the code, and thereby ensure that any subtleties of the behavior that we do your support are reliably consistent across view and layer back nodes?

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -890,134 +878,120 @@ + (_ASPendingState *)pendingViewStateFromView:(UIView *)view
...
+}
+
+- (BOOL)hasSetNeedsDisplay
+{
+  return _flags.needsDisplay;
+}
+
+- (BOOL)hasChanges
+{
+  ASPendingStateFlags flags = _flags;
```
appleguy: I wonder if we can cast this to an unsigned long long and test it against zero?

examples/BackgroundPropertySetting/Sample/DemoCellNode.swift
```diff
@@ -0,0 +1,87 @@
...
+	var state = State.Right
+
+	override init() {
+		super.init()
+		usesImplicitHierarchyManagement = true
+	}
+
+	override func layoutSpecThatFits(constrainedSize: ASSizeRange) -> ASLayoutSpec {
+		let specA = ASRatioLayoutSpec(ratio: 1, child: childA)
+		specA.flexBasis = ASRelativeDimensionMakeWithPoints(1)
```
appleguy: Should this be with points?  Not percent / fraction?

examples/BackgroundPropertySetting/Sample/ViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+			}
+		}
+	}
+
+	@objc private func didTapLayoutButton() {
+		let currentlyVisibleNodes = collectionNode.view.visibleNodes()
+		let queue = dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0)
+		dispatch_async(queue) {
+			for case let node as DemoCellNode in currentlyVisibleNodes {
+				node.state.advance()
```
appleguy: This app is really fun!  Thanks for the great new example!

appleguy:
Merging this is an epic milestone for the framework. Has literally been years in the making, And it is likely in the top three gifts we have ever accepted from and open source contributor.  Thank you, @Adlai-Holler, for being such a great friend of AsyncDisplayKit!

@knopp yes, we are already doing background layout in Pinterest.   Note however that this is not quite supported out-of-the-box yet, and it will take a few refinement changes next week to enable it. It should be well under 100 lines to do some so.

Be sure to look at usesImplicitHierarchyManagement, as well as the new transition APIs, although they will remain in the +Beta.h header for a while.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
Adlai-Holler: @appleguy Hmm this is pretty interesting.
- The `__loaded` macro in this file evaluates to `_layer != nil`. As far as I can tell, this is incorrect because `_layer` will never bet set on view-backed nodes. (`__loadNode` -> `[self layer]` -> `return self.view.layer` short-circuit.)
- `nodeLoaded` isn't backed by an instance variable – it's computed as a combination of `_view` `_layer` and `_flags.layerBacked`.

How should I proceed? Is there some reason the `__loaded` macro behaves this way? It's used all over this file so we need it to be correct. The lightest-weight solution here is to replace the short-circuit I mentioned above in `-layer` with `_layer = self.view.layer; return _layer` so that after `__loadNode` we guarantee that `_layer` is non-nil.

For now what I'll do is apply the same logic that that method uses here.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
Adlai-Holler: Eh not really. Currently the only reusable presence of `registerNode:` is in `ASDisplayNodeShouldApplyBridgedWriteToView` which is a very fast package of [check node, check thread, dirty if needed, return if should set to view]. In setFrame: we need to use some of those intermediate results. We could call it here, and it wouldn't result in any more method calls but it would result in another ASDisplayNodeThreadIsMain and another node-loadedness check. Doesn't seem worth it to me, let me know if you disagree.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
Adlai-Holler: Hmm unfortunately, if we read from the pending state in those cases we'll miss any changes the user made to the underlying view/layer. It would be safe to call `__setNeedsLayout` just after applying the pending state, though, right?

AsyncDisplayKit/Private/ASDisplayNodeInternal.h
Adlai-Holler: Hell yeah!

AsyncDisplayKit/Private/ASPendingStateController.mm
Adlai-Holler: IMO the separation of external API from an internal operation is valuable here but you're the boss.

AsyncDisplayKit/Private/ASWeakSet.m
Adlai-Holler: :sunglasses:

PR #1050, opened 2016-01-09
[Example App] New sample that uses a network-loading collection view on iPhone and iPad.

stowy:
As discussed with @appleguy, creating a pull request for review for this sample project we have been collaborating on.

This includes a new layout spec 'ASRelativePositionLayoutSpec' used for positioning views, much like the center layout spec, but with more options - center, left/top or right/bottom align - on both axis.

The sample project includes a fairly complex node layout, as well as infinite scrolling. 

Cat deals FTW! :-)


appleguy:
@stowy love it, thanks Sam!  It's been a blast working with you.

@lappp9 you might be interested in checking this out.  We could consider using elements from it for the main sample project, or at least keep it in mind as part of the portfolio of sample projects if we want to go in a different direction.  Whatever we do should definitely include orthogonal nested scrollers, like an ASPagerNode containing ASTableNodes, or vertical ASCollectionNode containing a horizontal one in some of its cells (like the photo story side-scroll units in Facebook)

stowy:
Thanks @appleguy been great working with you too, I hope it can continue. I would be interested in adding the ASPagerNode and also the ascollectionnode because I'd like to know how to use these in my main project. Is ASPagerNode already released?

Also, I noticed that there was a build failure due to the relative position layout spec I created as its not included in the snapshot tests. Is this easily resolved? Or do I need to remove the new spec and rewrite that part of the layout?

appleguy:
@stowy no rush, but if you could fix the merge conflict, that might even surface what was causing the build issue (since this is mostly a new project, anything that's conflicting with a change in the rest of the framework may be the thing that needs to be removed to fix the build from failing on the new layout spec).  If you could either put up the layout spec in a new PR, or **probably better** just send the implementation files to me so I could do it for you (if my assumption is correct that you're time-pressured right now), that would be awesome!

stowy:
@appleguy just pushed a squashed commit with the code fixed up - there were some references to the layout spec that i deleted still hanging around. Since it's now only new files then the tests should all pass.  Will create a separate PR with the layout spec, once one or the other are integrated i can revise the sample to use the layout spec again.

appleguy:
@stowy awesome, thanks for pushing this through!

PR #1054, opened 2016-01-11
[ASRangeController] properly update _rangeTypeIndexPaths when adding/removing index paths and sections

### LINKED ISSUE #1028: [ASRangeController] Old range controller (current stable 1.9.4 release) issue when deleting items.
In ADK 1.9.4. It will throw exception when trying to delete items. It works fine in 1.9.3
```
2016-01-02 22:47:38.186 Sample[51411:3328724] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[NSArray objectsAtIndexes:]: index 5 in index set beyond bounds [0 .. 4]'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010dc4de65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x000000010d6c4deb objc_exception_throw + 48
	2   CoreFoundation                      0x000000010db8fad8 -[NSArray objectsAtIndexes:] + 952
	3   Sample                              0x000000010cb75c40 __ASFindElementsInMultidimensionalArrayAtIndexPaths_block_invoke + 128
	4   Sample                              0x000000010cb74c7d _ZL52ASRecursivelyUpdateMultidimensionalArrayAtIndexPathsP14NSMutableArrayPK7NSArrayRmP11NSIndexPathmU13block_pointerFvS0_P10NSIndexSetmE + 1533
	5   Sample                              0x000000010cb748ca _ZL52ASRecursivelyUpdateMultidimensionalArrayAtIndexPathsP14NSMutableArrayPK7NSArrayRmP11NSIndexPathmU13block_pointerFvS0_P10NSIndexSetmE + 586
	6   Sample                              0x000000010cb75803 ASFindElementsInMultidimensionalArrayAtIndexPaths + 675
	7   Sample                              0x000000010cb142f1 -[ASDataController nodesAtIndexPaths:] + 577
	8   Sample                              0x000000010caf9070 -[ASCollectionView rangeController:nodesAtIndexPaths:] + 96
	9   Sample                              0x000000010cb8805b -[ASRangeController _updateVisibleNodeIndexPaths] + 1547
	10  Foundation                          0x000000010d2acb89 __NSFireDelayedPerform + 402
	11  CoreFoundation                      0x000000010dbadc84 __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__ + 20
	12  CoreFoundation                      0x000000010dbad831 __CFRunLoopDoTimer + 1089
	13  CoreFoundation                      0x000000010db6f241 __CFRunLoopRun + 1937
	14  CoreFoundation                      0x000000010db6e828 CFRunLoopRunSpecific + 488
	15  GraphicsServices                    0x000000011217dad2 GSEventRunModal + 161
	16  UIKit                               0x000000010e6c1610 UIApplicationMain + 171
	17  Sample                              0x000000010cac50af main + 111
	18  libdyld.dylib                       0x000000010fd4492d start + 1
)
```

Seems relative to logic change in line https://github.com/facebook/AsyncDisplayKit/blob/1.9.4/AsyncDisplayKit/Details/ASRangeController.mm#L134

knopp:
This fixes #1028

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeController.mm
+++ b/AsyncDisplayKit/Details/ASRangeController.mm
@@ -235,11 +235,31 @@
     // otherwise _updateVisibleNodeIndexPaths may try to retrieve nodes from dataSource that aren't there anymore
     for (NSInteger i = 0; i < ASLayoutRangeTypeCount; i++) {
       id rangeKey = @((ASLayoutRangeType)i);
+
       NSMutableSet *rangePaths = [_rangeTypeIndexPaths[rangeKey] mutableCopy];
-      for (NSIndexPath *path in indexPaths) {
-        [rangePaths removeObject:path];
+      for (NSIndexPath *pathToRemove in indexPaths) {
+        [rangePaths removeObject:pathToRemove];
       }
-      _rangeTypeIndexPaths[rangeKey] = rangePaths;
+        
+      // After removing paths we are left with gaps; We need to shift all remaining index paths
+      // so that there are no gaps left
+      NSMutableSet *shiftedRangePaths = [NSMutableSet new];
+      for (NSIndexPath *path in rangePaths) {
+        int shift = 0;
+        for (NSIndexPath *pathToRemove in indexPaths) {
+          if (pathToRemove.section == path.section && pathToRemove.item < path.item) {
+            // removed path was before this path, we need to shift it back
+            ++shift;
+          }
+        }
+        NSIndexPath *shifted = path;
+        if (shift > 0) {
+          shifted = [NSIndexPath indexPathForItem:path.item - shift inSection:path.section];
+        }
+        [shiftedRangePaths addObject:shifted];
+      }
+
+      _rangeTypeIndexPaths[rangeKey] = shiftedRangePaths;
     }
     
     [_delegate rangeController:self didDeleteNodes:nodes atIndexPaths:indexPaths withAnimationOptions:animationOptions];
@@ -270,7 +290,26 @@
           [rangePaths removeObject:path];
         }
       }
-      _rangeTypeIndexPaths[rangeKey] = rangePaths;
+        
+      // After removing paths we are left with gaps; We need to shift all remaining index paths
+      // so that there are no gaps left
+      NSMutableSet *shiftedRangePaths = [NSMutableSet new];
+      for (NSIndexPath *path in rangePaths) {
+        __block int shift = 0;
+        [indexSet enumerateIndexesUsingBlock:^(NSUInteger indexToRemove, BOOL *stop) {
+          if (indexToRemove < path.section) {
+            ++shift;
+          }
+        }];
+
+        NSIndexPath *shifted = path;
+        if (shift > 0) {
+          shifted = [NSIndexPath indexPathForItem:path.item inSection:path.section - shift];
+        }
+        [shiftedRangePaths addObject:shifted];
+      }
+        
+      _rangeTypeIndexPaths[rangeKey] = shiftedRangePaths;
     }
     
     [_delegate rangeController:self didDeleteSectionsAtIndexSet:indexSet withAnimationOptions:animationOptions];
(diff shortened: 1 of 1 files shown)


appleguy:
@knopp thanks a lot, I can't tell you how awesome it is to have your help here - especially in the production implementation, as I'm hoping to get the Beta working well enough in the next several weeks, but still it will be a riskier direction for any production apps than fixing any holes in the Stable version.

appleguy:
This issue is avoided by the recompute-everything architecture of the Beta controller, right?

We are starting work soon on asynchronously applying the Display and FetchData ranges in the Beta version, on top of really substantial performance improvements I landed recently (I'd made a couple pretty stupid mistakes that just hadn't been tested through yet).  You might try out Beta in your code and let me know if there are any immediate problems you detect.

knopp:
@appleguy, in a way this is also an issue with `ASRangeControllerBeta`, because you store `_allPreviousIndexPaths` between `_updateVisibleNodeIndexPaths` invocations. When adding / removing nodes, the index paths in `_allPreviousIndexPaths` become stale and may point to items no longer existing data controller.

appleguy:
@knopp aha, ok.  Yes I actually ran into that but fixed it by handling nil nodes and skipping over them in the Beta controller, so I'm pretty sure that is a stable / correct behavior for the newer one.  Most important thing now is getting a new release out to fix this.

knopp:
@appleguy, which commit are you referring to? Also, are you sure that merely handling nil nodes is correct? When adding / removing nodes, the index paths in `_allPreviousIndexPaths` may simply point to wrong nodes, but still be non null. So even if it doesn't crash, it won't behave correctly. 

I think there are two possible solutions - either do what I did in the branch - update `_allPreviousIndexPaths` every time the data controller adds/removes nodes/sections, or simply discard `_allPreviousIndexPaths` when adding/removing and next time `_updateVisibleNodeIndexPaths` is called, iterate over all nodes from data source.

appleguy:
@knopp are you sure it won't behave correctly in the *Beta version?  Very very interested in your thoughts there.  For each index path, we independently check if it currently lies within each range.  I think that should be sufficient, but I have not yet deeply considered this _editing vs _completed nodes issue.  _editingExternalNodes was added after my original refactor of ASDataController and I'm actually pretty concerned about it, and will evaluate if it is possible to re-simplify some of that stuff.

For this diff, do you have any test apps / test cases that you used?  I want to get a fix for this issue out as quickly as possible, and I would believe that this approach addresses conditions that #1060 does not, but I am worried about pushing this out with minimal testing.  Is there a scenario you can describe that does in fact break #1060 (or more importantly, *Beta, because it will become Stable within a couple weeks), but is addressed by this change?

knopp:
@appleguy, here is project that is broken with current `ASRangeController` but works with this pull request: https://github.com/tomizimobile/ASDKSearchIssue

--

Regarding `ASRangeControllerBeta`: *Each index path* (`allIndexPaths`) that you check belongs to union of fetch, display, visible and `_allPreviousIndexPaths`. Now consider following scenario:

We have node`[0,0]` that is visible, and appropriate indexPath `[0,0`] is in `_allPreviousIndexPaths`; Let's call it original node;

Now this happens between invocations of  `_updateVisibleNodeIndexPaths`

1. New node is added to `[0,0]`; This causes original node index path `[0,0]` to become `[0,1]`.(`_allPreviousIndexPaths` doesn't change in any way though and still contains `[0,0]`)
2. Collection view is instantly scrolled in a way where section 0 is no longer in fetch, display or visible;

The original node is still marked as visible, even though it no longer is. That is because the union of fetch, display, visible and `_allPreviousIndexPaths` contains `[0,0]`. But original node is no longer `[0,0]`, it is now `[0,1]` (because another node was added to `[0,0]`. So the original node is missed while iterating `allIndexPaths` and not updated.

knopp:
It's probably not worth committing this. ASRangeControllerStable has other issues as well, so let's focus on ASRangeControllerBeta instead.

appleguy:
@knopp ok, I'll take your advice.  I tend to agree insofar as there is risk in changing code that is this fragile.

PR #1058, opened 2016-01-12
Add New didEndDisplayingNode:forItemAtIndexPath: delegate methods to provide removed node

Adlai-Holler:
If I delete a node from the data source while it's on-screen, I'd like the table view delegate to know which node will end displaying. Using `nodeForRowAtIndexPath:` is incorrect because the node will already have been removed from the table view at this point, and I'll get some other node instead.

I've tested this in my project and it's working well, both for scrolling nodes offscreen and deleting/replacing nodes.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -376,6 +376,16 @@ NS_ASSUME_NONNULL_BEGIN
 @optional
 
 - (void)collectionView:(ASCollectionView *)collectionView willDisplayNodeForItemAtIndexPath:(NSIndexPath *)indexPath;
+
+/**
+ * Informs the delegate that the collection view did remove the provided node from the view hierarchy.
+ * This may be caused by the node scrolling out of view, or by deleting the item
+ * or its containing section with @c deleteItemsAtIndexPaths: or @c deleteSections: .
+ * 
+ * @param collectionView The sender.
+ * @param node The node which was removed from the view hierarchy.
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASCollectionView.h
```diff
@@ -406,6 +406,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (BOOL)shouldBatchFetchForCollectionView:(ASCollectionView *)collectionView;
 
+/**
+ * Use `collectionView:didEndDisplayingNode:forItemAtIndexPath:` instead.
```
appleguy: Is this comment finished?

Can you add didEndDisplayingCell: as an asserted not-to-implement method in the delegate proxy if it isn't already?  I'm pretty sure it is already there, just want to be sure.

Adlai-Holler: I had considered that comment finished, but I take your question as asking me to beef it up, so I did that. Seems a little out of character to me so maybe there's a miscommunication here?

I'm not sure I follow about the assertion. There's no assertion around `didEndDisplayingCell:` currently but `ASCollectionViewProxy` does intercept that method. We only assert against implementing dead-end methods, but `didEndDisplayingCell:` wasn't and isn't a dead-end method. We still use `didEndDisplayingCell:` to drive our new `didEndDisplayingNode:forItemAtIndexPath:` method just like the old one so I don't think we need to assert against it.

appleguy:
@Adlai-Holler thanks for supporting the old stuff at runtime!

PR #1059, opened 2016-01-13
[WIP][tvOS] ASDK tvOS support

aaronschubert0:
This is currently WIP but I wanted to issue a PR to start a discussion about how to best support tvOS. My initial findings are that it is relatively easy to make ASDK build on tvOS and further more expose the relevant methods (`UIFocusEnvironment` Protocol) that are needed to lay the foundation of tvOS support, as witnessed by this PR.

The **main** work is making all the nodes in ASDK respond to the UIFocus Engine. I am in favour of having a default behaviour implementation on ASDisplayNode that is then inherited by all other nodes. This will ensure that all nodes have a basic behaviour, which can then be overridden to provide a better experience.

Once the default behaviour approach is decided, the second decision to be made is who's responsibility it is to provide great UIFocus behaviour? Would it be the author of that certain node or would someone provide a coherent experience across all the nodes? I'm wanting to implement a tvOS app at work soon using ASDK, so I'm happy to provide a heavy lift to support these efforts, especially for ASImageNode which has the most work to do to match the focus behaviour of UIImageView. Having said that, it could be faster, and/or favoured by the community to implement their own behaviours. 

Another aspect is that the design of the tvOS UI is counter-culture to what has been recommended for ASDK, mainly when items are in focus. tvOS makes heavy use of rounded corners and drop shadows, which ASDK has traditionally (for performance reasons) been strongly against.

A number of nodes are currently affected by a lack of certain properties/frameworks in tvOS, I need to do more research to fully determine these aspects but here is the initial list:
- `ASEditableTextNode`. UITextView.editable isn't currently available on tvOS, therefore undermining the purpose of this node.
-  `ASMapNode`. MapKit isn't available on tvOS 9.0 but *is* available on tvOS 9.2, so this isn't a big deal currently.
- `ASMultiplexImageNode`. The Photos and AssetsLibrary frameworks aren't available on tvOS so ASMultiplexImageNode is currently relegated to only displaying URL's from the web (which is probably the biggest use case, but still)
- `ASPagerNode`. UICollectionView.pagingEnabled & .scrollsToTop properties aren't available on tvOS, I suspect this is due a new mechanism for paging tvOS, so maybe this node can be adapted for this. cc @levi 
- `ASTableViewProtocols`. UITableViewRowAction is not available on tvOS. I have currently only gated the method that highlights this (for Swift support) but further investigation is needed to see if UITableViews are editable at all.

There are still a number of things to test and to-do to fully lay the ground work. I have only tested this with CocoaPods (which works). There are the things still to test/to-do, which I will update, as I come across additional items:
* [ ] Provide a static tvOS framework build
* [ ] Thoroughly test each node's behaviour 
* [ ] Provide a tvOS example app, similar to UIKitCatalog to showcase all of ASDK's nodes in a tvOS environment

@appleguy Could you have a look this, and also as you know the community better, could you tag as many people from the community who could best input?

### DIFF
--- a/AsyncDisplayKit.podspec
+++ b/AsyncDisplayKit.podspec
@@ -47,4 +47,5 @@ Pod::Spec.new do |spec|
   }
 
   spec.ios.deployment_target = '7.0'
+  spec.tvos.deployment_target = '9.0'
 end
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -641,7 +641,9 @@ NS_ASSUME_NONNULL_END
 @property (atomic, assign)           UIViewContentMode contentMode;         // default=UIViewContentModeScaleToFill
 
 @property (atomic, assign, getter=isUserInteractionEnabled) BOOL userInteractionEnabled; // default=YES (NO for layer-backed nodes)
+#if TARGET_OS_IOS
 @property (atomic, assign, getter=isExclusiveTouch) BOOL exclusiveTouch;    // default=NO
+#endif
 @property (atomic, assign, nullable) CGColorRef shadowColor;                // default=opaque rgb black
 @property (atomic, assign)           CGFloat shadowOpacity;                 // default=0.0
 @property (atomic, assign)           CGSize shadowOffset;                   // default=(0, -3)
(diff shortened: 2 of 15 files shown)


appleguy:
@aaronschubert0 I am /very/ excited about this - it is something I have been wanting to work on every weekend for like 2 months hahah.  I was pretty sad to realize recently that I can't justify spending time on it in the near future, nor would I have expected anyone else to :-P.  Will review by the end of the weekend, feel free to leave more comments or continue the iteration!

Issue #1055, opened 2016-01-11
Crash on deleteRowsAtIndexPaths:withRowAnimation:


kanvuduc:
I got exception when trying to delete a row. It sounds similar to #484.

I use latest release, ADK 1.9.4.

*CODE*
```Objective-C
[_tableView beginUpdates];
[_lstItems removeObjectAtIndex:index.row];
[_tableView deleteRowsAtIndexPaths:[NSArray arrayWithObject:index]
                          withRowAnimation:UITableViewRowAnimationLeft];
[_tableView endUpdates];
```

*CONSOLE LOG*
```
2016-01-11 14:36:25.811 Handhome[2945:72549] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[NSArray objectsAtIndexes:]: index 8 in index set beyond bounds [0 .. 7]'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001041d1e65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x0000000103948deb objc_exception_throw + 48
	2   CoreFoundation                      0x0000000104113ad8 -[NSArray objectsAtIndexes:] + 952
	3   Handhome                            0x0000000101e22680 __ASFindElementsInMultidimensionalArrayAtIndexPaths_block_invoke + 128
	4   Handhome                            0x0000000101e216bd _ZL52ASRecursivelyUpdateMultidimensionalArrayAtIndexPathsP14NSMutableArrayPK7NSArrayRmP11NSIndexPathmU13block_pointerFvS0_P10NSIndexSetmE + 1533
	5   Handhome                            0x0000000101e2130a _ZL52ASRecursivelyUpdateMultidimensionalArrayAtIndexPathsP14NSMutableArrayPK7NSArrayRmP11NSIndexPathmU13block_pointerFvS0_P10NSIndexSetmE + 586
	6   Handhome                            0x0000000101e22243 ASFindElementsInMultidimensionalArrayAtIndexPaths + 675
	7   Handhome                            0x0000000101dc0d31 -[ASDataController nodesAtIndexPaths:] + 577
	8   Handhome                            0x0000000101e57b30 -[ASTableView rangeController:nodesAtIndexPaths:] + 96
	9   Handhome                            0x0000000101e34a9b -[ASRangeController _updateVisibleNodeIndexPaths] + 1547
	10  Foundation                          0x0000000103531b89 __NSFireDelayedPerform + 402
[...]

Issue #1047, opened 2016-01-08
It crashes when a ASTableView deletes some cell in a section in the version 1.9.4


tuyuan2012:
I use the pod version which is 1.9.4. It crashes when a ASTableView deletes some cell in a section. The url of a test case is https://github.com/tuyuan2012/TestDeleteRow.git. In the older version there is not such problem. 
Well, our app will be uploaded to apple store recently. So I am a little worrying. Then, how can pop the older version. Thank you @appleguy very much....
The follow image shows the crash log of mine project not the test case. 
![08642c85-d52a-42a8-9c5d-fd3028a2caf9](https://cloud.githubusercontent.com/assets/3121889/12193562/782eda30-b624-11e5-95dd-6509d51b1be5.png)

knopp:
@tuyuan2012, have you tried these two patches? https://github.com/knopp/AsyncDisplayKit/commit/3c162ec03dcef00fe5170f49b4202caf9f66ffd2 https://github.com/knopp/AsyncDisplayKit/commit/c0c98c0cefaa99fa2f56892818567f0e7eb283c8

tuyuan2012:
Can I pod the new code of knopp@c0c98c0 @knopp ? Thank you....

tuyuan2012:
Thank you @knopp...With the knopp@3c162ec,  there is not such problem. How can I  to pod the new code of the ASRangeController.

tuyuan2012:
Thank you @knopp...With the knopp@3c162ec, there is not such problem. How can I pod the new code of the ASRangeController.

tuyuan2012:
@appleguy @knopp Thank you very much. I pod the sources with the latest commit, there is no such problem.

knopp:
Don't close this issue. This is only fixed in my fork.

tuyuan2012:
I am so sorry @knopp, I have reopened it.

vitalybaev:
The same issue #1055

appleguy:
@tuyuan2012 sorry, I really failed you guys here.  I've been so intensely absorbed in the Pinterest rewrite...the positive news for all of you is that we now have several great engineers at the company working on major continued improvements to ASDK, on top of what is being done by the community.  AND, http://raywenderlich.com is doing a full 2.0 tutorial, including for ASLayoutSpec!

[...]

PR #1060, opened 2016-01-13
Resolve Internal Inconsistency During Updates When Using Stable Range Controller

Adlai-Holler:
This addresses #1028 #1055 #1047 .

The issue seems to be caused by one of the changes in ffcddf3 where we stopped checking if the range was valid before removing nodes from the working range. I just reverted that change exactly as-was and the issue is gone for me.

@appleguy I'm not sure the implications of this change. I suspect it means that under certain conditions, nodes won't be removed from the working range. We may now be throwing the baby out with the bathwater.

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeController.mm
+++ b/AsyncDisplayKit/Details/ASRangeController.mm
@@ -156,8 +156,7 @@
       NSSet *indexPaths = [_layoutController indexPathsForScrolling:_scrollDirection rangeType:rangeType];
 
       // Notify to remove indexpaths that are leftover that are not visible or included in the _layoutController calculated paths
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
Ah I just saw #1054 . This may work as a hotfix.

appleguy:
@Adlai-Holler thank you for debugging my regression.  I also appreciate you making a minimized change, as this code has actually already been tested (although granted, not with every more recent change).  I'll push out a new dot release shortly, probably later tonight.

knopp:
@appleguy, with this fix will break https://github.com/facebook/AsyncDisplayKit/issues/1011 again :-/

PR #1064, opened 2016-01-14
LayoutFacilitator

binl:
Adding a LayoutFacilitator delegate for ASCollectionViews, the purpose is so that we could use this to provide information that the layout might need about the ASCV.

Details up for debate:
- I created this new delegate instead of adding the hooks to the inspector because I think the two has very different responsibilities.
- The hooks are invoked when the rangeController's delegate methods are called, because that's when the cellNodes are fully measured and ready to be used by any layout.

Thoughts? @appleguy @levi


levi:
Interesting addition, @binl. Is this the new asynchronous UICVL handling that Scott was talking about?

binl:
@levi  not yet async, but trying to make CVLayouts smarter and so more performant. what do you think is that best place to add this kind of hook so that ASCV can inform "facilitator" about potential cell editing? (insert, delete, update, reload, etc.)

levi:
Bin, I'd recommend adding these callbacks to the layoutDelegate object. Right now the delegate implements an "inspector" object for providing details into how the layout behaves. Providing a custom implementation is required for custom cv layouts and you're more than likely going to want to extend this inspector to interact with cell editing hooks.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -43,6 +44,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (instancetype)initWithCollectionViewLayout:(UICollectionViewLayout *)layout;
 - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout;
+- (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout layoutFacilitator:(id<ASCollectionViewLayoutFacilitatorProtocol>)layoutFacilitator;
```
appleguy: @binl can you add a similar initializer to ASCollectionNode?  I think those leave out the frame, but we should have one with the layout and layoutFacilitator.  Going forward we should always create Collection / Table Nodes rather than Views, as the direct allocation of the Views will be deprecated in 2.0.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -913,7 +924,7 @@ - (void)rangeController:(ASRangeController *)rangeController didInsertSectionsAt
 - (void)rangeController:(ASRangeController *)rangeController didDeleteSectionsAtIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
 {
   ASDisplayNodeAssertMainThread();
-  
+  [_layoutFacilitator collectionViewEditingSectionsAtIndexSet:indexSet];
```
appleguy: Do we need any calls like this for either full reloads, or move operations?

AsyncDisplayKit/ASCollectionViewLayoutFacilitatorProtocol.h
```diff
@@ -0,0 +1,19 @@
+//
+//  ASCollectionViewLayoutFacilitatorProtocol.h
+//  Pods
+//
+//  Created by Bin Liu on 1/13/16.
```
appleguy: Can you use the standard comment at the top of files?  It's supposed to call out the copyright unfortunately :(

AsyncDisplayKit/ASCollectionView.h
binl: done

AsyncDisplayKit/ASCollectionView.mm
binl: As far as I can tell, full reloads and move operations are both done through deleting and inserting nodes, and these delegates will be called.

appleguy: @binl importantly, this has changed with @nguyenhuy's latest changes to reloading.  Although it is true that moves currently work that way, this is something that needs to be changed soon, as the current mechanism causes both incorrect / unpredictable animations and is less efficient than moving the actual node (and performing a relayout if it needs, but not actually destroying and recreating the nodes and thus forcing them to re-render).

appleguy:
@binl I'm going to merge this now so that you can continue at full speed, but it's important to me that we move this initializer into a +Beta.h file so that it can't be used without an engineer being clear that it's a new API that hasn't been widely tested / broadly discussed with tech leads of different apps with their many differences in use cases.

In particular we should work together to solidify the relationship between the layout inspector and the layout facilitator, and although I think it's likely the API will remain unchanged for the final version (just as you have it now), it's important to put it in +Beta because there could be some bugs / inconsistencies that haven't been sorted out yet (like supporting @nguyenhuy's new reloading setup).

I need to release a new version of ASDK in the next ~24hrs, so if you won't have a chance to create the +Beta.h file, just let me know and I can make the tweak before that release goes out (it's needed for a couple other critical fixes).

Issue #1063, opened 2016-01-14
[ASEditableTextNode] Add support for scrollEnabled property while preserving workaround for longstanding Apple bug.


lkzhao:
I wonder if this is a design decision made previously. But why isn't the ASEditableTextNode support scrolling.
If we want to be able to scroll the textnode when text are long, what is the best way to do it with ASDK?

appleguy:
@lkzhao I got some answers on this from the original author of the code.

"Hey Scott — following up on the radar. Apple closed it as behaving correctly. I clarified the bug in the radar — the bug is really that "When `scrollEnabled` is NO, UITextView no longer updates its `contentSize`."  This makes it impossible to tell how large the text is / how big to make the text view if you want to embed it inside of another custom UIScrollView."

So basically in Paper we had to leave scrollEnabled as YES so the sizing calculation was done, /but/, we actually did not want it to scroll by itself, because we needed the text to scroll inside a shared scroll view with a bunch of other elements alongside the text.

@lkzhao can you make a change so that _ASDisabledPanUITextView is instead named ASPanningOverriddenUITextView, and then add an instance variable for "_shouldBlockPanGesture"?  The ASPanningOverriddenUITextView should also override setScrollEnabled:.  It should always call [super setScrollEnabled:YES], but then just set the value appropriately for the _shouldBlockPanGesture.  Then the instance variable can be used in the UIGestureRecognizer method.

I know this feels hacky, because it is, but I don't want this change to break Instant Articles that probably still depends on this behavior - the Apple behavior is really not correct in my opinion.

Lastly, please expose a scrollEnabled property on the ASEditableTextNode in the header.  Ideally you can implement this in such a way that it does not trigger the view to be created, because it is extremely expensive to create a UITextView, and ASDisplayNodes are sometimes created and destroyed without their view ever being loaded.  The value for scrollEnabled: can be applied in -didLoad.

appleguy:
This is fixed.

PR #1069, opened 2016-01-15
[ASEditableTextNode] Support scrolling while maintaining workaround for Apple bug when disabling it.

lkzhao:


AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -207,11 +187,17 @@ - (void)setLayerBacked:(BOOL)layerBacked
 #pragma mark - Configuration
 @synthesize delegate = _delegate;
 
+- (void)setScrollEnabled:(BOOL)scrollEnabled{
```
appleguy: open bracket on next line - match the other methods.  Not a blocker, just FYI :)

appleguy:
@lkzhao all the details are in #1063 - thank you for raising this bit of ASDK history I'd long forgotten, and bringing this improvement to others!

AsyncDisplayKit/ASEditableTextNode.h
```diff
@@ -24,6 +24,11 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma mark - Configuration
 
 /**
+ @abstract Enable scrolling on the textView
+ */
+@property (nonatomic) BOOL scrollEnabled;
```
appleguy: Specify the @default value for this

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -16,16 +16,31 @@
 #import "ASTextNodeWordKerner.h"
 #import "ASThread.h"
 
-//! @abstract This subclass exists solely to ensure the text view's panGestureRecognizer never begins, because it's sporadically enabled by UITextView. It will be removed pending rdar://14729288.
-@interface _ASDisabledPanUITextView : UITextView
+//! @abstract This subclass forces the parent UITextView's scrollEnabled property to always be true. Instead, it disables the panGestureRecognizer when scrollEnabled is set to false. This ensures that the contentSize is caculated correctly.
+//! See issue: https://github.com/facebook/AsyncDisplayKit/issues/1063
```
appleguy: Break out the abstract into a few lines.  Pretty good as-written!  I would add a clarification / maybe replace some of your words with:

"As originally reported in rdar://14729288, when scrollEnabled = NO, UITextView does not calculate its contentSize.  This makes it difficult for a client to embed a UITextView inside a different scroll view with other content (setting scrollEnabled = NO on the UITextView itself, because the containing scroll view will handle the gesture)...because accessing contentSize is typically necessary to perform layout.  Apple later closed the issue as expected behavior.  This works around the issue by ensuring that contentSize is always calculated, while still providing control over the UITextView's scrolling."

appleguy:
Go ahead and update the comments in another diff.  Maybe you can also rename the method I mentioned in this one - https://github.com/facebook/AsyncDisplayKit/pull/1078

Issue #1077, opened 2016-01-19
[ASTableView] Rarely, ASCellNode is provided with a default cell height as a constrained size.


aaronschubert0:
This is a peculiar bug that I've only came across while running on an iPad (on iOS 7 but haven't tested yet if it's the OS or Device). The problem is that sometimes my nodes in the ASTableView receive the wrong constrainedSize. By "wrong" I mean that the nodes by default should receive the full width of the ASTableView (768pts in this example, iPad 3 Portrait) but occasionally some nodes receive a constrainedSize of 320,44pts. 
Now this size looks very peculiar because I'm almost certain that this is the default value for a UITableViewCell's contentView, which ASTableView actually uses to default the constrainedSize when there is a mismatch.:
```
- (void)willLayoutSubviewsOfTableViewCell:(_ASTableViewCell *)tableViewCell
{
  CGFloat contentViewWidth = tableViewCell.contentView.bounds.size.width;
  ASCellNode *node = tableViewCell.node;
  ASSizeRange constrainedSize = node.constrainedSizeForCalculatedLayout;
  
  // Table view cells should always fill its content view width.
  // Normally the content view width equals to the constrained size width (which equals to the table view width).
  // If there is a mismatch between these values, for example after the table view entered or left editing mode,
  // content view width is preferred and used to re-measure the cell node.
  if (contentViewWidth != constrainedSize.max.width) {
    constrainedSize.min.width = contentViewWidth;
    constrainedSize.max.width = contentViewWidth;

    // Re-measurement is done on main to ensure thread affinity. In the worst case, this is as fast as UIKit's implementation.
    //
    // Unloaded nodes *could* be re-measured off the main thread, but only with the assumption that content view width
    // is the same for all cells (because there is no easy way to get that individual value before the node being assigned to a _ASTableViewCell).
    // Also, in many cases, some nodes may not need to be re-measured at all, such as when user enters and then immediately leaves editing mode.
    // To avoid premature optimization and making such assumption, as well as to keep ASTableView simple, re-measurement is strictly done on main.
    [self beginUpdates];
    CGSize calculatedSize = [[node measureWithSizeRange:constrainedSize] size];
    node.frame = CGRectMake(0, 0, calculatedSize.width, calculatedSize.height);
    [self endUpdates];
  }
}
```
[...]

PR #1078, opened 2016-01-19
[ASTableView] Ensure -[super layoutSubviews] is called before messaging delegate.

aaronschubert0:
Nice and easy fix.
Details: #1077 

@appleguy

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -47,8 +47,8 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 
 - (void)layoutSubviews
 {
-  [_delegate willLayoutSubviewsOfTableViewCell:self];
   [super layoutSubviews];
+  [_delegate willLayoutSubviewsOfTableViewCell:self];
 }
 
 - (void)didTransitionToState:(UITableViewCellStateMask)state


appleguy:
Hmm, well, now the name of the method is incorrect - it should be didLayout :).  This is probably the right fix for now anyway, but a bit of a bummer.  Maybe we should deprecate the method or at least add a comment to the header...

AsyncDisplayKit/ASTableView.mm
```diff
@@ -47,8 +47,8 @@ @implementation _ASTableViewCell
 
 - (void)layoutSubviews
 {
-  [_delegate willLayoutSubviewsOfTableViewCell:self];
   [super layoutSubviews];
+  [_delegate willLayoutSubviewsOfTableViewCell:self];
```
appleguy: Oh, excellent, I just realized this is in our Cell subclass and is entirely internal.  Let's rename it to -didLayoutSubviewsOfTableViewCell:, and we're good to land!  Really thrilled you investigated this issue right on the heels of writing a detailed report - thank you, @aaronschubert0!

appleguy:
I'm going to take this fix now, and in any future diff, feel free to rename :)

appleguy:
Interesting that ASCollectionView doesn't have this method overridden...

AsyncDisplayKit/ASTableView.mm
aaronschubert0: @appleguy Sweet! Will rename it. Makes more sense now that this was an intended change previously but I think this is the more coherent approach.

Issue #1076, opened 2016-01-19
Text with headIndents is not rendered correctly


nvh:
When specifying a `headIndent` and a `firstLineHeadIndent`, text is rendered incorrectly:

![simulator screen shot 19 jan 2016 10 56 34](https://cloud.githubusercontent.com/assets/248861/12417494/c15556f8-bea8-11e5-9608-c9285850e17f.png)

Initially the text is clipped on the right side of the node, but when rerendering after scrolling (the bottom half of the screenshot) the text is rendered inside the box on the left and right, but falls of the bottom.

A test project where this behaviour is displayed can be found [here](https://github.com/nvh/ASTextNodeTest)

aaronschubert0:
Hi @nvh thanks for reporting this issue and providing an example project! Makes it a lot easier to debug.  The problem property is actually just `headIndent`, `firstLineHeadIndent` seems to be rendering properly. I'll put up a PR shortly that address this specific problem but a deeper inspection/fix may be needed.

appleguy:
I believe this is fixed in master - @nvh, could you confirm that in your original use case it is handled?  I'm keeping this open because I want to better understand the issue and ensure this is covered in the expanded test cases we're starting to develop for the text stack.

aaronschubert0:
Hey @appleguy, talking about expanded test cases, could you link me to some WIP (if there is any) I was planning to write some for nsattribrutedstring & ASTextNode soon. Just so we have complete coverage going forward and can better understand any issues (and prevent those!)

appleguy:
This test project was essential to my detailed understanding of this code.  I'm very thankful for the excellent reports you have created over time, @nvh.  You have made a material difference in the project as a whole - really!

nvh:
Thank you for looking into this @appleguy, I'm happy to help!

PR #1079, opened 2016-01-19
[ASTextKitRenderer] Potential fix headIndent attribute not being accounted for by the shadower.

aaronschubert0:
Hey @appleguy This is to fix the bug specified in #1076. I'm not entirely sure this is the correct place in the stack to fix this issue. The issue stems from the fact that the `headIndent` property of NSAttributedString actually alters the origin of the boundRect rather than just increase it's size. This bug fix address this by incorporating the origin offset into the final size calculation.

The `tailIndent` property seems to break things even more, that can be fix in a future diff (haven't got to the bottom of that)

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -111,7 +111,7 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
   // to make sure our width calculations aren't being offset by glyphs going beyond the constrained rect.
   boundingRect = CGRectIntersection(boundingRect, {.size = constrainedRect.size});
 
-  _calculatedSize = [_shadower outsetSizeWithInsetSize:boundingRect.size];
+  _calculatedSize = [_shadower outsetSizeWithInsetSize:CGSizeMake(boundingRect.size.width + boundingRect.origin.x, boundingRect.size.height + boundingRect.origin.y)];
 }
 
 - (CGSize)size


appleguy:
@aaronschubert0 this is fascinating.  Thank you for researching this.  It is an extremely important issue, yet there are very few folks prepared to debug the text system.

What kind of test case(s) do you have that helped you diagnose the issue?  My top priority is getting the fix in, especially since it looks like I'm going to do a 1.9.5.1 shortly (unfortunately - see #1083).  However, if you have appropriate test cases, I want to at least file an issue to ensure that we add a snapshot test to verify this behavior in the face of future text changes.

Could you explain why the new calculation is correct?  Why exactly would we add the origin value to the size, in order to generate a size?

appleguy:
Going to take this now and start testing it in Pinterest, but would still like to continue discussing it - feel free to send me an email if you leave a comment here (it would be good to have the discussion here though so that it is documented).

aaronschubert0:
Hey @appleguy awesome. Yeah it's definitely a crucial area to invest time in. The test case that I used was the example project linked in #1076 and the debugger. So I definitely need to write some actual tests. Basically it seems that most properties of NSAttributedString are interpreted properly by our text size calculations since they change the size of the text container to achieve their 'effect'. In contrary headIndent (and I assume tailIndent) achieves it's effect by altering the origin of the whole text container, rather than altering the size, so .headIndent = 10 will result in the size to remain the same but the origin will now be (10, 0) rather than (0, 0). Since the positioning of our nodes are usually done by the parent (i.e in a layoutSpec method) it is appropriate to just alter the resulting size of textNode since that will result in the desired affect (that was my thinking behind the change anyway)

PR #1082, opened 2016-01-19
Remove SpinLock to ensure it doesn't get used in new code.

Adlai-Holler:
Looks like not only are spin locks dumb, they're also dangerous!

Recently big time Apple engineers have said spin locks are illegal on iOS. So for the sake of cleanliness, I think we should remove the corresponding class.

Source: http://mjtsai.com/blog/2015/12/16/osspinlock-is-unsafe/


appleguy:
Sure, good idea :).  I think we used to use it in one place, but it's been a while!

Issue #1086, opened 2016-01-20
[GitHub] Add CocoaPods download badges.


appleguy:
ASDK has over 115,000 downloads and is used in almost 1,700 apps, including 515 unique apps that are actively developing with it enough to have used pod install with it this week.  Pretty awesome!

http://cocoapods.org/pods/AsyncDisplayKit

Let's add a badge to showcase this.  This is an old task and may not explain how to add the badges correctly, but we should check it out.  @aaronschubert0 have you ever looked at this?

https://github.com/CocoaPods/CocoaPods/issues/1092

aaronschubert0:
@appleguy  Never done it before but seemed really straightforward!

PR #1087, opened 2016-01-20
Update ReadMe to include Cocoapods downloads badge

aaronschubert0:
As specified in #1086. Hey @appleguy Turns out these links have been there all along but were just hidden since they were using http (I think). The img.shields.io is a bit unstable at times but seems to be the best solution so far. Will update in future if better solutions come up!


aaronschubert0:
@appleguy There doesn't seem to be a way to automatically pull the download numbers from Cocoapods and create a badge at the moment. So this adds a manual downloads badge. But it can easily be changed by altering the text in the URL. Maybe we can write a small function and submit a PR to the badges project (https://github.com/badges/shields) so that we can handle this automatically in the future.

appleguy:
@aaronschubert0 thank you for looking into this random auxiliary thing!  Really glad to see this!

PR #1089, opened 2016-01-20
Make ASTextNode Ignore Setting Text to Nil When Text Is Already Empty

Adlai-Holler:
This is a small fix where we coalesce the nil before checking equality so that the text node will correctly ignore setting the text to nil when it is already at "". 

A bigger change would be to make ASTextNode's nullability consistent. Currently it defaults to nil at initialization, and then if you set it to nil any time after that it gets set to "". I don't know much about the our TextKit setup so I didn't want to pull that thread in this PR.


appleguy:
@Adlai-Holler ha, good find / example.  I have a task on me to make setAttributedString: more efficient because right now it is pretty rough, as is init for ASTextNode, but both of these issues will be mitigated by the change to support creating ASCellNodes in the background...so I'm probably going to defer optimizing these until at least after that point as it is the bigger win.

Issue #1080, opened 2016-01-19
[Performance] GCD can spawn too many threads, reducing efficiency.


knopp:
I'm still not quite happy with some performance aspects, so I have put together small project that tests AsyncDisplayKit overhead here

https://github.com/knopp/AsyncDisplayKitKiller (requires some patience upon start, it creates 5000 cell nodes)

Main problem with this example is, that during somewhat faster scrolling, after a while the app gives up and starts dropping a lot of frames, which looks very bad (even on iPad Air 2). I could live with the scrolling not running 60fps in this case, if the performance was at least somewhat consistent, but unfortunately it is not.

One thing I noticed is that there are over 60 rendering threads. Even though most of them are waiting on `__ASDisplayLayerConcurrentDisplaySemaphore`, it still seems that scheduling the call simply using `dispatch_async` is rather wasteful. It would probably be better if a queue was maintained internally (similar to how `NSOperationQueue` works) so that AsyncDisplayKit could limit number of spawned threads. 

This would also allow to have less overhead when cancelling rendering, the operation could simply be removed from internal queue (whereas now the block needs to be scheduled and executed anyway, just to compare the displaySentinel value).

There are probably other things too. Any ideas on how to make the performance more consistent are very welcome.

knopp:
So, it looks like this might be a red herring. While there is certainly room for improvement, it seems that in my actual project the main cause for frame drops is setting the rendered image layer content. Which doesn't make me very happy since there doesn't seem to be much else to do.

knopp:
So, this is not as much of a red herring as I though. The issue with setting `_layer.contents` was mostly caused by display range of 0, which means all items that were rendered were already onscreen. Setting layer content on onscreen element is obviously more expensive and it blocks until the image is ready to be displayed. Relaxing the display range significantly alleviated this.

But the main issue of dozens of threads killing the CPU persist - it's especially obvious on slower devices (iPad 3). I did a very ugly hack to get _ASAsyncTransaction to process the rendering sequentially and the performance on iPad 3 was much smoother.

appleguy:
This is a great piece of performance research.  I wish we were working together 1-1 on a day to day basis, @knopp!

[...]

PR #1090, opened 2016-01-20
[_ASAsyncTransaction] limit number of spawned threads

knopp:
Relevant for #1080 

GCD can get *very* impatient. I've seen dozens of spawned threads during scrolling on iPad 3 when the rendering doesn't keep up. The semaphore in `ASDisplayNode+AsyncDisplay` doesn't help much when this happens and the main thread gets crushed.

The patch introduces a very lightweight queue/group that reliably limits number of drawing threads and reduces GCD calls (which can get expensive when there's lot queued). In future the queue could also be used for rendering prioritization if desired (i.e. prioritize rendering labels to images, etc).

### DIFF
--- a/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
+++ b/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
@@ -9,6 +9,9 @@
 #import "_ASAsyncTransaction.h"
 #import "_ASAsyncTransactionGroup.h"
 #import "ASAssert.h"
+#import "ASThread.h"
+#import <list>
+#import <map>
 
 @interface ASDisplayNodeAsyncTransactionOperation : NSObject
 - (id)initWithOperationCompletionBlock:(asyncdisplaykit_async_transaction_operation_completion_block_t)operationCompletionBlock;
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -47,9 +50,220 @@ - (NSString *)description
...
+  DispatchEntry &entry = q._entries[queue];
+  
+  Operation operation;
+  operation._block = block;
+  operation._group = this;
+  entry._operations.push_back(operation);
+  
+  ++_pendingOperations; // enter group
+  
+  NSUInteger maxThreads = [NSProcessInfo processInfo].activeProcessorCount;
```
appleguy: Is there a reason to change this from the previously determined concurrent count value?  I think this should be at least *2, as render operations may block on loading an image from disk cache or other situations.

appleguy:
@knopp very impressive.  This is one of the most technical patches I've ever received from the community, and is definitely something I'm very happy to have - an area I've wanted to work on forever, but really had to hope of getting to soon, thus my enthusiasm!

I'd love to merge this soon, but have to ask: how did you go about testing this, both when you were developing it and before finalizing / putting it up?  Would you be willing to write a few basic tests?

appleguy:
I will take this long before 2.0, but would first like to get 1.9.6 out in the next day or so - this will probably go into 1.9.7.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
knopp: Not sure what you mean by previously determined concurrent count value. If you mean `__ASDisplayLayerMaxConcurrentDisplayCount` 8 seems really overkill to me, especially on slow devices (like iPad 3) main thread doesn't get enough time. Also being it that much over number of cores it introduces lot of scheduling overhead.

So `activeProcesorCount` (but at least 2) seems like a good value to me, but maybe it doesn't work for your use case?

knopp:
@appleguy, thanks for looking into this. It's been tested in with my internal project and AsyncDisplayKit examples. Also all tests pass.

Regarding the tests, the coverage is actually pretty good already, as lot of tests involve the transaction methods at least indirectly, with there also being test specifically for `_ASAsyncTransaction`. But I'll look into which codepaths are not currently being tested and will hopefully think of something.

Also you could test this with pinterest and let me know if everything work as expected :)

appleguy:
Now that 1.9.6 is out, excited to start testing this at Pinterest!  Thanks for your intense and competent focus on the framework in the last few weeks, @knopp!

PR #1091, opened 2016-01-21
[ASTextNode] Support iteratively shrinking font size in order to ensure no single words are wrapped.

lkzhao:
User can specify a minimumScaleFactor on ASTextNode. 
Then the ASTextRenderer will shrink the font size if there are words that doesn't fit in the container.

If the word is too long, i.e. even with the minimumScaleFactor applied, it sill doesn't fit in the container. Then it will fall back to the default font size.


appleguy:
This is the start of great things to come!  Will merge after tests run.  Thank you @lkzhao, very much.

PR #1093, opened 2016-01-21
[ASDataController] Optimize reloadData to avoid performing deletions or insertions.

nguyenhuy:
and notify its delegate once instead of a series of deletes and inserts. UITableView and UICollectionView are notified via `reloadData` and so flashes are minimized (if not eliminated).

I'm not sure if I also need to change anything in ASCollectionDataController, especially `prepareForReloadData` and `willReloadData`? Tests are passing though (excepts snapshot tests, which is a different story I guess).

Will consider applying the same fix to `reloadRows` and `reloadSections` if this diff gets through.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -43,6 +43,7 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
   BOOL _delegateDidDeleteNodes;
   BOOL _delegateDidInsertSections;
   BOOL _delegateDidDeleteSections;
+  BOOL _delegateDidReloadData;
 }
 
 @property (atomic, assign) NSUInteger batchUpdateCounter;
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -240,6 +258,14 @@ - (void)deleteNodesOfKind:(NSString *)kind atIndexPaths:(NSArray *)indexPaths co
   }];
 }
 
+- (void)deleteAllNodesOfKind:(NSString *)kind
+{
+  [_editingNodes[kind] removeAllObjects];
```
appleguy: Do we also need to clear _externalCompletedNodes here, @nguyenhuy ?

nguyenhuy: No, I think. _externalCompletedNodes is only used during batch updates and should be managed entirely within endUpdatesAnimated:completion.

Issue #1046, opened 2016-01-08
[ASTextNode] ASTextKitRenderer not always invalidated when needed.


timominous:
After updating to 1.9.4, my single character text node is being rendered incomplete.

Before:
![before](https://cloud.githubusercontent.com/assets/733151/12192113/ed2cb612-b616-11e5-96e1-3abaaffaf3ca.PNG)
After:
![after](https://cloud.githubusercontent.com/assets/733151/12192116/f885038e-b616-11e5-9126-e6e799fb0468.PNG)

We can see from the dark gray area that the size is calculated correctly but the text isn't rendered properly.

I am experiencing this when the supernode (custom `ASCellNode` subclass) was just initialized by `-tableView:nodeForRowAtIndexPath:` and is displayed immediately. After scrolling up and down for a few times, the rendering of the text will appear completed.

I'm really new to AsyncDisplayKit so I'm not sure if this has something to do with my code or the framework's code.

bsmith11:
#1031 

Getting the same behavior. Seems to only happen when `alignment = NSTextAlignmentCenter`.

AttilaTheFun:
I'm having this issue as well with a text node inside a horizontal stack layout spec.

nvh:
I'm having this issue as well, and have narrowed it down to the invalidation of the renderer in the text node, specifically in the line that is highlighted here:
https://github.com/facebook/AsyncDisplayKit/commit/1872ac07aa7f4c03678d6b00aa6f183514599acd#diff-221147908d17673747bdcbd88905332cR312
When YES is returned instead of NO here (invalidating the renderer for this case) The text is rendered as expected.

appleguy:
@nvh thanks for looking at this - what version of ASDK are you running on?  It looks like a commit from mid-December, and I think this issue was fixed shortly after that...it should be in 1.9.3-1.9.5 I think.  I believe the initial regression of this was in 1.9.2.

@timominous what version of ASDK are /you/ running on?  :)

appleguy:
@nguyenhuy, @samhsiung if you have a chance to take a look in the next day or two — please do!  The 6 month plan review is tomorrow, so I'm pretty jammed until late Wednesday or around the end of the week.  But this could be a bug important enough to do a hotfix for, if somehow we've missed detecting it previously.

nguyenhuy:
Will take a look as soon as I can :)

nvh:
[...]

Issue #1031, opened 2016-01-04
[Layout Specs] Best way to center ASTextNode with attributed string?


yxztj:
Hi,

Repro steps:
1. Create a ASTableView, implement dataSource
2. Create CellNode, which containing a subnode, which contains a ASTextNode with attributedString in turn.
3. When the cells are first rendered on screen, the text is centered with an unexpected offset(The string should be "TEST TEST TEST", but now the latter part is outside text node so you can only see "TEST T"). Looks like the attributes "thinks" the node has a bigger size, so the text is centered in a bigger frame rather than its actual frame. 
After scrolling down and up again(re-render the cell), the issue goes away.

Attached gif:
![ezgif com-video-to-gif 1](https://cloud.githubusercontent.com/assets/3334458/12083416/124eabd0-b2de-11e5-82fd-7dddc106fdb1.gif)

I've created a repo to demonstrate the issue. Please let me know if I'm missing anything.
https://github.com/yxztj/ASTextNodeTest

Thanks!

appleguy:
@yxztj this is a good test case.  It might be that you need to set .flexGrow or some other ASLayoutable / ASStackLayoutable property.

Have you tried using ASCenterLayoutSpec?  Let me know what that does for you.  I will debug this by the end of the weekend - hopefully tomorrow, though.

yxztj:
@appleguy The frame of the text node is correct. The problem is that text is not rendered correctly in the text node, instead it has an unexpected horizontal offset.(Looks like the text inside the node is centered in a larger frame)
We are using text attributes to align the text at center, so if there are multiple lines they are all centered.
Expected:
┌─────────┐
│  TEXT TEXT TEXT│
│            TEXT           │
└─────────┘
Actual
┌─────────┐
│                    TEXT T│
│                              T│
└─────────┘

bsmith11:
I'm running into the same problem and it seems to have only started after I updated to ADK 1.9.4.

I don't think it's related to layout specs since I'm still using the old style of manually setting frames. It seems to only happen if `paragraphStyle.alignment = NSTextAlignmentCenter`. Below are screenshots showing `NSTextAlignmentCenter` and the default value `NSTextAlignmentLeft`. The frame of the `ASTextNode` has a white background so we can see that it's sizing itself correctly.

`NSTextAlignmentLeft`
[...]

PR #1100, opened 2016-01-24
[ASTextNode] Optimize handling of constrained size to almost never recreate NSLayoutManager

appleguy:
This also fixes three fairly subtle but serious bugs: #1076, #1046 and #1031.

cc @aaronschubert0 - note I ended up tracing a logic flaw to the most recent patch in this area (#1079), which altered the inset handling with the shadower.  I'm still glad you put up this change because it did very clearly appear to improve the behavior, but worsened other situations, like the centering case.  

@nvh - you'd be the most valuable tester, although I confirmed both types of clipping miscalculations very strategically recreated in your awesome sample app appear fully corrected.

@yxztj - I actually was using your sample app for the majority of the time it took me to carefully understand and adjust this code.

@bsmith11, @timominous, @AttilaTheFun, @nguyenhuy, @1nput0utput, @stowy, @lappp9, @RCacheaux, @samhsiung, @binl, @rcancro - Please join us in testing!

I'm excited that I feel confident working in this part of the codebase; it's probably one of the few scary corners remaining to be confronted.  One of the primary needs now is a big expansion in automated performance and logic testing...

The overall code quality and engineering style in the framework, to me, feels awesome. I'm glad to be working on continuing to do industry-first things as a community together!

### DIFF
--- a/AsyncDisplayKit/ASCollectionNode+Beta.h
+++ b/AsyncDisplayKit/ASCollectionNode+Beta.h
@@ -6,6 +6,7 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
+#import "ASCollectionNode.h"
 @protocol ASCollectionViewLayoutFacilitatorProtocol;
 
 NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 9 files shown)


aaronschubert0:
Awesome work @appleguy thanks for putting in the effort to make this area of ASDK so solid!

nvh:
@appleguy Thanks for looking into this! Most of the bugs, especially those with the centering are fixed in my original use cases. With the firstHeadlineIndent there still is a problem with clipping text on a single line. I've updated my [test project](https://github.com/nvh/ASTextNodeTest) to display this behaviour.

PR #1101, opened 2016-01-24
[ASRangeControllerBeta] Utilize NSMutableOrderedSet to ensure visible items are prioritized.

appleguy:
This also adopts Objective-C generics for the various collections in this class.

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeControllerBeta.mm
+++ b/AsyncDisplayKit/Details/ASRangeControllerBeta.mm
@@ -21,6 +21,7 @@
 {
   BOOL _rangeIsValid;
   BOOL _queuedRangeUpdate;
+  BOOL _layoutControllerImplementsSetVisibleIndexPaths;
   ASScrollDirection _scrollDirection;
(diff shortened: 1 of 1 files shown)


appleguy:
cc @knopp, @nguyenhuy, @tomizmobile - please test it out!  BTW I did detailed performance checking on iPhone 4 and the performance difference is absolutely negligible compared to the other operations in this method.  I can think of a couple different optimizations but they're not worth pursuing at this time as there would be an increase in code complexity.

wendylu:
lgtm

knopp:
@appleguy, seems to work fine for me. Btw. Would something like this fly? https://github.com/facebook/AsyncDisplayKit/pull/1102

appleguy:
Here's the before and after performance comparison - must only use the percentages as the amount of work done in terms of objects is not exactly equal.
screen shot 2016-01-24 at 6 16 36 pm

knopp:
@appleguy, in my case it is about twice as slow, but nothing to write home about. 
screen shot 2016-01-25 at 1 56 05 pm

PR #1102, opened 2016-01-24
[ASRangeController] don't get index paths twice if tuning parameters …

knopp:
…are same

In my case I have same tuning parameters for fetch and display. Getting the index paths is quite expensive with large datasets, this reduces the time in half (in some cases).


appleguy:
@knopp very smart idea.  By default they aren't set this way, and I think in many apps it is more optimal for them to not be equal, but it is absolutely worth exploiting this if they are equal.

PR #1104, opened 2016-01-25
[ASLayout, ASTextNode] Create -calculatedLayoutDidChange and use it in text node.

appleguy:
This allows the change in size for the NSTextContainer to occur off the main thread, whenever that size change is necessary.  Then the text relayout can occur earlier, during the process of computing ASLayoutSpecs.


nguyenhuy:
:+1:

PR #1105, opened 2016-01-25
[ASRangeControllerBeta] Scan all nodes after edit operations.

appleguy:
@knopp, please do take a look!  Is there any more efficient way you can think to do this?  I'm particularly bummed about creating all the index paths - I think we could make an alternate codepath that iterates over the nodes themselves, keeping a local integer of section and item number, doing something like the multidimensional helper function does internally by building up the index path by index - but not having to create a full array of them.  This might not help efficiency much, though it would also eliminate accessing the nodes out of the 2D array.

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeControllerBeta.mm
+++ b/AsyncDisplayKit/Details/ASRangeControllerBeta.mm
@@ -134,6 +134,10 @@
   [allIndexPaths unionSet:_allPreviousIndexPaths];
   _allPreviousIndexPaths = allCurrentIndexPaths;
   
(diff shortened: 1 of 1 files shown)


knopp:
I hear you, but in the end, you still need to check
fetchDataIndexPaths, displayIndexPaths and visibleIndexPaths, and you need
NSIndexPaths for that, don’t you? And you need to do it for every single
item. But with separate code path you could maybe at least
avoid NSMutableOrderedSet overhead, which miight not be small if you have
thousands of index paths.

-Matej

PR #1106, opened 2016-01-25
[ASVideoNode] Performance enhancements

lappp9:
•  Asynchronously loads first frame as a placeholder when video doesn't need to be playing
•  Removed extra pauses that caused some slow down.


AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -58,8 +65,6 @@ - (void)interfaceStateDidChange:(ASInterfaceState)newState fromState:(ASInterfac
       if (_shouldBePlaying) {
         [self pause];
         _shouldBePlaying = YES;
```
appleguy: if _shouldBePlaying is already set, we should not need to set it?! :) @lappp9

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -286,6 +312,20 @@ - (void)play
     }];
   }
   
+  if (!_playerNode) {
```
appleguy: We need to factor out this code to reduce duplication / chance of it changing separately.  Maybe define a method -playerNode that creates and returns the playerNode if not created.  Do not add it as a subnode in that method, but instead have a separate add or insert call wherever that operation is needed.

AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -127,15 +127,28 @@ - (void)testPlayerIsCreatedInFetchData
...
   _videoNode.asset = _firstAsset;
 
+  [_videoNode setInterfaceState:ASInterfaceStateNone];
   [_videoNode didLoad];
   
-  XCTAssert([_videoNode.subnodes containsObject:_videoNode.playerNode]);
+  XCTAssert(![_videoNode.subnodes containsObject:_videoNode.playerNode]);
 }
 
+- (void)testPlayerLayerNodeIsNotAddedIfVisibleButShouldNotBePlaying
```
appleguy: Thanks for continuing to improve tests!!

AsyncDisplayKit/ASVideoNode.mm
lappp9: Ah yeah, @appleguy  this is a weird thing i need to change but didn't want to until i had time to put a test or two around the logic.  I reset it to _shouldBePlaying because a call to -pause sets shouldBePlaying to NO.  Probably a side effect i need to get rid of, but I was originally trying to cut down on duplication of setting that property, lol

PR #1107, opened 2016-01-25
Bring reloadData optimizations back

nguyenhuy:
The ASCollectionView data inconsistency crash introduced by #1093 occurs because I didn't apply the same workaround added in #672. With the root cause discovered, I'd like to bring that PR back. When time permits, I will write a unit test to make sure this kind of crash won't happen again.


appleguy:
Haha, I accidentally announced this just slightly too early!  I'll copy-paste this in 1.9.7 :-P

* **ASTableNode** and **ASCollectionNode** -reloadData methods are faster, substantially exceeding performance of UIKit by performing concurrent cell layout in chunks.  This release accelerates the teardown of ASDataController objects.  Just as with UIKit however, It's still always recommended to use the many edit commands available instead of reloadData (when possible).

Issue #1033, opened 2016-01-04
ASTextNode truncationAttributedString not showing


tomizimobile:
I just realized that neither the `truncationAttributedString` or `additionalTruncationMessage` are showing for me anymore. It was working in 1.9.3, but stopped working in 1.9.4, and is still not working in master. I haven't tried to find an exact commit yet that breaks it though. This is easy to verify by modifying the "Kittens" example project, "KittenNode.mm" to set `_textNode.maximumNumberOfLines = 5` in initialization

appleguy:
@tomizimobile thanks for the report.  There were definitely changes in this code.  It is possible that in certain initialization orders, the ASTextKitTailTruncater doesn't work correctly.  See my commit hash 99fbc97 if you want to experiment with reverting it to test, or look at the changelog for ASTextKitRenderer and ASTextNode.  Most of the changes there since 1.9.3 should be possible to revert without affecting any other dependencies across AsyncDisplayKit.

I will most likely be able to investigate this by the end of the weekend, and might have time tomorrow (otherwise it's Friday night).  Let me know if you happen to learn anything else about it, but you should save your time if work is demanding right now!

appleguy:
There are unit test cases for truncation.  We should inspect those to ensure they are functioning as desired - I'm surprised this would get through.  It may be precisely that the initialization order used with the tests does not trigger this bug.

tomizimobile:
@appleguy Thanks for the lead! I took a quick dig into it and I found that things work as expected if `_truncater` is set during initialization. I'm not too familiar with the code, so I don't know if this has any side effects, but I'd guess it's basically negating any optimizations that were attempted. I'll have to leave reworking that to someone else, but at least for now I've got an interim fix.

appleguy:
@tomizimobile could you please test this on master and let me know if it's fixed?  I think it is resolved but I don't think I have a test case for this.  Might be worth checking the implementation of the unit tests and adding / modifying one to include whatever code you're using to test this, because there are snapshot tests for the truncater already which are passing...but also weren't failing before!

tomizimobile:
@appleguy I just tested on master, and it's only kind-of working. The text isn't truncated and the truncation doesn't show't, but the area where the truncation token would be responds to touches and highlights. My patch to fix this is simple, but I don't know what side effects it has: [patch.diff.txt](https://github.com/facebook/AsyncDisplayKit/files/103340/patch.diff.txt)

appleguy:
[...]

PR #1110, opened 2016-01-25
[ASTextNode] Prompt creation of truncater object when the context object is accessed.

### LINKED ISSUE #865: [ASTextNode] Usage of .additionalTruncationMessage is not clear / documented.
I create textNode
```Objective-c


self.descriptionTextNode                  = [[ASTextNode alloc] init];
self.descriptionTextNode.maximumLineCount = 3;
self.descriptionTextNode.truncationMode = NSLineBreakByTruncatingTail;
self.descriptionTextNode.attributedString = [[NSAttributedString alloc] initWithString:@"More text Emitters walk from futures like ordinary planets.The twisted mermaid technically pulls the c-beam.Sonic showers resist with advice at the conscious center!Rudely, indeed, interstellar courage!I fight this adventure, it's called evil collision course.string"
                                                                                    attributes:[self p_descriptionAttributes]];
self.descriptionTextNode.additionalTruncationMessage = [[NSAttributedString alloc] initWithString:@"TruncationText dsfdfs a fad sd asdfasfasf "
                                                                                              attributes:[self p_descriptionAttributes]];
        [self addSubnode:self.descriptionTextNode];
```

And what i must do for see `additionalTruncationMessage` ?

yury:
See [comment](https://github.com/facebook/AsyncDisplayKit/commit/99fbc97bdaecd2f99009b9734ba0f8f2b665632b#commitcomment-15649130)

Fixes #865.

attributedString on ASTextNode should be set after all truncationStrings...

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -88,6 +88,8 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
                                                    exclusionPaths:attributes.exclusionPaths
                                                   constrainedSize:shadowConstrainedSize
                                              layoutManagerFactory:attributes.layoutManagerFactory];
+
+    [self truncater];
   }
   return _context;
 }


appleguy:
@yury what exactly is the dependency here?  Why doesn't the truncater get accessed, e.g. [self truncater] get called, at a later point as soon as it is required?  Thanks a lot for looking at this!

yury:
@appleguy as my debugging shows, `[self truncater]` is called only to get visibleRange to detect links. And it is to late at that moment.

Truncated should be initialized ahead of size calculation so it can actually truncate string in it's initializer.
@tomizimobile patch here https://github.com/facebook/AsyncDisplayKit/issues/1033 also init truncater ahead.

yury:
Actual message text

```
hello
this is
very long
message here
and here
```

with `maximumNumberOfLines = 4`, `truncationAttributedString = " ..."` and `additionalTruncationMessage = "read more"`

Before fix
screen shot 2016-01-26 at 01 13 51

after 
screen shot 2016-01-26 at 01 14 37

appleguy:
@yury thank you for those additional details.  It is still mysterious to me why this object has to be created ahead of time and I wonder if there is a later point than init where it can be created, as init may occur on the main thread, but none of our sizing calculations occur there.  If you're able to share a test case, I can verify, but it's possible the perf impact of this is negligible with the current factoring of the code (it used to be measurable as suboptimal).

appleguy:
Merging this for correctness now, optimization later with a test case.  Please email me at asyncdisplaykit@gmail.com with any other thoughts!  cc @tomizimobile

tomizimobile:
@appleguy I've tried out master just now, and while the truncation string now appears, it seems that you can't actually interact with it. Using the workaround I talked about in https://github.com/facebook/AsyncDisplayKit/issues/1033 still seems to work for me. 
@yury Are you seeing this same behavior?

yury:
@tomizimobile will check tomorrow.

yury:
@appleguy, @tomizimobile confirming that textNodeTappedTruncationToken doesn't called.

And confirming, that @tomizimobile workaround fixes this.

@tomizimobile can you make PR?

PR #1111, opened 2016-01-25
[ASTextNode] Fix out-of-bounds error when truncation string is longer than the original string.

yury:
Actually it is possible that truncated string is longer then original
string.

For example, original string:

```
hello
this is
very long
message here
and there
```

with maximumNumberOfLines=4, truncationAttributedString = ' ...' and
additionalTruncationMessage = 'read more'

will give

```
hello
this is
very long
message here ... read more
```

So `[attributedString attribute:attributeName atIndex:characterIndex
longestEffectiveRange:&range inRange:visibleRange]` will crash.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -479,6 +479,7 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   ASTextKitRenderer *renderer = [self _renderer];
   NSRange visibleRange = renderer.visibleRanges[0];
   NSAttributedString *attributedString = _attributedString;
+  NSRange clampedRange = NSIntersectionRange(visibleRange, NSMakeRange(0, attributedString.length));
 
   // Check in a 9-point region around the actual touch point so we make sure
   // we get the best attribute for the touch.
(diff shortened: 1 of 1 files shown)


yury:
Also related to #1110

yury:
Also, there are two unported commits from ComponentKit [1](https://github.com/facebook/componentkit/commit/d58d8635bbf8dd183a8845150e7942019c2dbac9#diff-0ad1a5fb62b0e43439ca5309c769bab3) and [2](https://github.com/facebook/componentkit/commit/3e5f6a7b0f840826f0bdeca4a793564a1c65b163#diff-0ad1a5fb62b0e43439ca5309c769bab3). I can port them with new PR

appleguy:
@yury thanks a lot for porting the other CK commits!!  @ocrickard anything else we should be aware of?  I wonder if there's a way to set up something similar to a Herald rule to notify me when text components change :)

appleguy:
@yury could you send me an example test case for this so I can integrate it into the unit tests?  If you're able to add a snapshot / unit test for this directly, that would be phenomenal too :).  asyncdisplaykit@gmail.com

yury:
@appleguy ported to commits from component kit #1119.

PR #1115, opened 2016-01-26
Fix reloadData regression

nguyenhuy:
Need to grab `_externalCompletedNodes` while reloading new nodes, otherwise there will be data inconsistency if external objects (e.g backing views) try to access completed nodes before `ASDataController` finishes the transaction.

Also update `ASTableViewTests` to reliably fail without this new patch.


nguyenhuy:
The base diff is about to be reverted (#1120). Closing this PR.

PR #1117, opened 2016-01-26
[_ASAsyncTransaction] implement operation priority

knopp:
Something to consider.

Two commits here. One somewhat unrelated tweaks number of threads. Just `activeProcessorCount` does indeed seem to be a bit conservative. New code does `activeProcessorCount * 2` and decreases the result by 1 when tracking.

Second implements prioritization of transaction operations. Priority is a number, operations with higher priority are more likely to get executed sooner. Some details:

* one (first) processing threads executes all operation in queue order; This is to ensure that even lower priority operations will not be completely ignored; Subsequent threads execute higher priority operations first
* scheduling and retrieving scheduled operations is done in constant time (in practice; corner cases where no two operations have same priority, would be log(n), but that's extremely unlikely to happen)
* Node drawing priority can be specified by the `drawingPriority` property. The assumption is that most nodes would stick to default priority, so instead of field a flag and associated object is used for nodes with custom priority.

This all seems to help quite a lot in our case, where we strongly prefer labels to be rendered during scrolling to thumbnails, which can now be easily achieved by
```
imageNode.drawingPriority = ASDefaultDrawingPriority - 1;
```
It doesn't make much difference on iPad Air 2, but does on iPad 3. :)

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -26,6 +26,8 @@
 #import "ASLayoutSpec.h"
 #import "ASCellNode.h"
 
+NSUInteger const ASDefaultDrawingPriority = ASDefaultTransactionPriority;
+
 @interface ASDisplayNode () <UIGestureRecognizerDelegate>
 
 /**
@@ -2234,6 +2236,31 @@ static void _recursivelySetDisplaySuspended(ASDisplayNode *node, CALayer *layer,
       [_supernode subnodeDisplayWillStart:self];
     }
   }
+}
+
+static const char *ASDisplayNodeDrawingPriorityKey = "ASDrawingPriority";
+
+- (void)setDrawingPriority:(NSUInteger)drawingPriority
+{
+  ASDisplayNodeAssertThreadAffinity(self);
+  ASDN::MutexLocker l(_propertyLock);
+  if (drawingPriority == ASDefaultDrawingPriority) {
+    _flags.hasCustomDrawingPriority = NO;
+    objc_setAssociatedObject(self, ASDisplayNodeDrawingPriorityKey, nil, OBJC_ASSOCIATION_ASSIGN);
+  } else {
+    _flags.hasCustomDrawingPriority = YES;
+    objc_setAssociatedObject(self, ASDisplayNodeDrawingPriorityKey, [NSNumber numberWithUnsignedInteger:drawingPriority], OBJC_ASSOCIATION_RETAIN);
+  }
+}
+
+-(NSUInteger)drawingPriority
+{
+  ASDisplayNodeAssertThreadAffinity(self);
+  ASDN::MutexLocker l(_propertyLock);
+  if (!_flags.hasCustomDrawingPriority)
+    return ASDefaultDrawingPriority;
+  else
+    return [objc_getAssociatedObject(self, ASDisplayNodeDrawingPriorityKey) unsignedIntegerValue];
 }
 
 - (BOOL)isInHierarchy
--- a/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
+++ b/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
@@ -29,6 +29,8 @@ typedef NS_ENUM(NSUInteger, ASAsyncTransactionState) {
   ASAsyncTransactionStateComplete
 };
 
+extern NSUInteger const ASDefaultTransactionPriority;
+
 /**
  @summary ASAsyncTransaction provides lightweight transaction semantics for asynchronous operations.
 
@@ -95,6 +97,25 @@ typedef NS_ENUM(NSUInteger, ASAsyncTransactionState) {
                    completion:(asyncdisplaykit_async_transaction_operation_completion_block_t)completion;
 
 /**
+ @summary Adds a synchronous operation to the transaction.  The execution block will be executed immediately.
+ 
+ @desc The block will be executed on the specified queue and is expected to complete synchronously.  The async
+ transaction will wait for all operations to execute on their appropriate queues, so the blocks may still be executing
+ async if they are running on a concurrent queue, even though the work for this block is synchronous.
+ 
+ @param block The execution block that will be executed on a background queue.  This is where the expensive work goes.
+ @param priority Execution priority; Tasks with higher priority will be executed sooner
+ @param queue The dispatch queue on which to execute the block.
+ @param completion The completion block that will be executed with the output of the execution block when all of the
+ operations in the transaction are completed. Executed and released on callbackQueue.
+ */
+- (void)addOperationWithBlock:(asyncdisplaykit_async_transaction_operation_block_t)block
+                     priority:(NSUInteger)priority
+                        queue:(dispatch_queue_t)queue
+                   completion:(asyncdisplaykit_async_transaction_operation_completion_block_t)completion;
+
+
+/**
  @summary Adds an async operation to the transaction.  The execution block will be executed immediately.
 
  @desc The block will be executed on the specified queue and is expected to complete asynchronously.  The block will be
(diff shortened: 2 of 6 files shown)


AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
```diff
@@ -95,6 +97,25 @@ typedef NS_ENUM(NSUInteger, ASAsyncTransactionState) {
                    completion:(asyncdisplaykit_async_transaction_operation_completion_block_t)completion;
 
 /**
+ @summary Adds a synchronous operation to the transaction.  The execution block will be executed immediately.
+ 
+ @desc The block will be executed on the specified queue and is expected to complete synchronously.  The async
+ transaction will wait for all operations to execute on their appropriate queues, so the blocks may still be executing
+ async if they are running on a concurrent queue, even though the work for this block is synchronous.
```
appleguy: Having a bit of trouble understanding this comment - well written overall, but by "synchronous" do you mean something like "once the block is started, it will continue to run on that display thread until it is completed, even if higher priority operations are scheduled?"

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
```diff
@@ -95,6 +97,25 @@ typedef NS_ENUM(NSUInteger, ASAsyncTransactionState) {
...
+ 
+ @param block The execution block that will be executed on a background queue.  This is where the expensive work goes.
+ @param priority Execution priority; Tasks with higher priority will be executed sooner
+ @param queue The dispatch queue on which to execute the block.
+ @param completion The completion block that will be executed with the output of the execution block when all of the
+ operations in the transaction are completed. Executed and released on callbackQueue.
```
appleguy: name of parameter is now "queue"

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
```diff
@@ -112,6 +133,27 @@ typedef NS_ENUM(NSUInteger, ASAsyncTransactionState) {
...
+ 
+ WARNING: Consumers MUST call the completeOperationBlock passed into the work block, or objects will be leaked!
+ 
+ @param block The execution block that will be executed on a background queue.  This is where the expensive work goes.
+ @param priority Execution priority; Tasks with higher priority will be executed sooner
+ @param queue The dispatch queue on which to execute the block.
+ @param completion The completion block that will be executed with the output of the execution block when all of the
+ operations in the transaction are completed. Executed and released on callbackQueue.
+ */
+- (void)addAsyncOperationWithBlock:(asyncdisplaykit_async_transaction_async_operation_block_t)block
```
appleguy: ok, I hadn't realized there would be an async variant, still reading further to make sure I understand.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -13,6 +13,8 @@
 #import <list>
 #import <map>
 
+NSUInteger const ASDefaultTransactionPriority = 128;
```
appleguy: Maybe we should type this as an NSInteger rather than NSUInteger, especially because we don't really want to expose / have peolpe depend on the base value of 128.

Also then, for debugging purposes, we might as well have the base value be 0 perhaps?  or maybe 10, 100, some other human-percieved-round value.

Once this is in and tested, I am definitely interested in boosting the priority of ASTextNode drawing above images.  The feature you included to always have one thread running in queue order is a very good / important idea.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -159,32 +207,29 @@ - (NSString *)description
...
-  NSUInteger maxThreads = [NSProcessInfo processInfo].activeProcessorCount;
-  if (maxThreads < 2) { // it is reasonable to have at least two working threads, also
-    maxThreads = 2;     // [_ASDisplayLayerTests testTransaction] requires at least two threads
-  }
+  NSUInteger maxThreads = [NSProcessInfo processInfo].activeProcessorCount * 2;
 
-#if 0
-  // Bit questionable - we can give main thread more CPU time during tracking;
-  if (maxThreads > 1 && [[NSRunLoop mainRunLoop].currentMode isEqualToString:UITrackingRunLoopMode])
+  // Bit questionable maybe - we can give main thread more CPU time during tracking;
```
appleguy: I love this idea.  I think it is the right thing to do, 100%.  We do NOT want to excessively hurt concurrency / overall CPU utilization even in this case, but lightening the load a touch will allow the scheduler to more consistently succeed at getting the main thread as much run time as it needs (obviously the thread is set to a much higher priority anyway, but your work has certainly confirmed the scheduler is imperfect when there are several threads at full load and main starts becoming active when it previously wasn't - or periodically needs to run every frame but confuses the scheduler because the overall load index looks a lot less compared to the long-lived background threads)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2236,6 +2238,31 @@ - (void)setDisplaySuspended:(BOOL)flag
...
+
+- (void)setDrawingPriority:(NSUInteger)drawingPriority
+{
+  ASDisplayNodeAssertThreadAffinity(self);
+  ASDN::MutexLocker l(_propertyLock);
+  if (drawingPriority == ASDefaultDrawingPriority) {
+    _flags.hasCustomDrawingPriority = NO;
+    objc_setAssociatedObject(self, ASDisplayNodeDrawingPriorityKey, nil, OBJC_ASSOCIATION_ASSIGN);
+  } else {
+    _flags.hasCustomDrawingPriority = YES;
```
appleguy: I think you did a great job of implementing this efficiently.  However, could you explain why you chose to prefer associated objects?  I'm curious if it is possible to just have a property for the number, which would help when we start to use this more heavily, e.g. to always prioritize text rendering.

appleguy:
@knopp Really fantastic work here.  I shared it with all of the iOS developers at Pinterest, and we are extremely excited to have your last two changes in.  We will be testing them with the full force of the company, as all of our internal iOS testers will be on this by the end of the week :)

I'm merging it now for that reason - to get test coverage both in our organization and from the other great developers who follow ASDK master - and because your first implementation here is /very/ clean, very efficiently implemented, demonstrating many best practices for developing in highly perf- and reliability-critical framework code.

Feel free to answer my questions here, or in email - maybe send me a note when you have a chance to reply.  I don't think a followup diff is required, but if you have any ideas for improvements, I'll certainly be thinking about what we can do next and would be happy to discuss!  Your changes are most welcome, but also don't hesitate to surface priorities / problem points that my team or the much larger community can help with as well.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
knopp: @appleguy, I just copied the entire comment from original method when adding variant with priority. There are two ways to schedule transaction operation, but the async one (which requires scheduled block to call completion block) is not used anywhere as far as I can tell.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
knopp: @appleguy, again, the async variant was already there :) I didn't introduce it, just added method variants with priority.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
knopp: Yeah, NSUInteger didn't feel right from the beginning. I'll change it to NSInteger with default to zero. Makes much more sense.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
knopp: Yeah. This way you get 3 threads on double core CPU while tracking, 4 threads while still. It can always be tweaked, but this seems like a decent compromise in my case. Still, if you play with it and find something that works better for pinterest I'm all ears

AsyncDisplayKit/ASDisplayNode.mm
knopp: @appleguy, I did it under assumption that most nodes would have default priority, only few others (i.e. images) would have different values (i.e. instead of labels having increased priority I decreased priority for just the image). 

Of course if you start setting priorities for more nodes, at some point ivar will become better choice than associated object. I'm not sure what the constraints are, if it is considered cheap to add 4 or 8 bytes to every node then let's do it :)

knopp:
@appleguy, thanks for the feedback. I changed priority to NSInteger here https://github.com/facebook/AsyncDisplayKit/pull/1122

I think I answered most of your questions here, if I missed something let me know

PR #1122, opened 2016-01-26
[_ASAsyncTransaction] make priority a NSInteger with default to 0

knopp:
For some reason it looks much better than NSUInteger and 128 :)

Signed-off-by: Matej Knopp <matej.knopp@gmail.com>


appleguy:
very nice!

Issue #1116, opened 2016-01-26
Extending ASMultiplexImageNode in Swift fails in master


nvh:
Since [this commit](https://github.com/facebook/AsyncDisplayKit/commit/fe572c23fb0e6b9dfac3c7edca9b92066516ce18) I get this error when trying to extend ASMultiplexImageNode in Swift:

```
Ld /Users/nvh/Library/Developer/Xcode/DerivedData/AsyncDisplayKit-Bugs-aadskzqczecgdlfjliubdhezulqc/Build/Products/Debug-iphonesimulator/AsyncDisplayKit-Bugs.app/AsyncDisplayKit-Bugs normal x86_64
    cd /Users/nvh/ios/AsyncDisplayKit-Bugs
    export IPHONEOS_DEPLOYMENT_TARGET=9.2
    export PATH="/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/usr/bin:/Applications/Xcode.app/Contents/Developer/usr/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
[...]

PR #1124, opened 2016-01-27
Added Prefix Headers to framework target

nvh:
This fixes the problems described in #1116 

I guess because the prefix headers weren't used when compiling the framework target, the preprocessor macro's weren't available and the code in the `ASMultiPlexImageNode` implementation was removed by the preprocessor.


appleguy:
Cool, thanks for figuring that out!

Issue #967, opened 2015-12-22
[ASTextNode] Add `linkTextAttributes` for customizing link attributes


samhsiung:
This would model `UITextView` interface which has a `linkTextAttributes` property for customizing link text attributes. `NSForegroundColorAttributeName` is otherwise ignored if `NSLinkAttributeName` is set.

http://stackoverflow.com/questions/25457131/setting-nslinkattributename-font-color

appleguy:
@samhsiung do you happen to know what is causing these attributes to be ignored in the ASDK implementation anyway?  I haven't looked at it in detail yet, but I'm wondering if it is necessary for them to be skipped and if we absolutely need new API or if we can just allow them to be respected while still handling the entire multi-styled substring as a link if it is so attributed.

samhsiung:
According to this writeup: https://www.cocoanetics.com/2015/03/customizing-uilabel-hyperlinks/  

>NSLayoutManager has a internal color hard-coded for glyph ranges belonging to a hyperlink. It ignores the NSForegroundColorAttributeName, we believe wrongfully so. All drawing functions are called with the CGContext already configured by layout manager. This includes setting the fill color for the glyphs.

They were able to work around this behaviour by subclassing `NSLayoutManager` and overriding `showCGGlyphs...`
```
- (void)showCGGlyphs:(const CGGlyph *)glyphs 
           positions:(const CGPoint *)positions 
               count:(NSUInteger)glyphCount
                font:(UIFont *)font 
              matrix:(CGAffineTransform)textMatrix 
          attributes:(NSDictionary *)attributes
           inContext:(CGContextRef)graphicsContext
{
   UIColor *foregroundColor = attributes[NSForegroundColorAttributeName];
 
   if (foregroundColor)
   {
      CGContextSetFillColorWithColor(graphicsContext, foregroundColor.CGColor);
   }
 
   [super showCGGlyphs:glyphs 
             positions:positions 
                 count:glyphCount 
                  font:font 
                matrix:textMatrix 
            attributes:attributes 
             inContext:graphicsContext];
}
```

We could do something similar if we wanted to support using `NSForegroundColorAttributeName` for ranges with `NSLinkAttributeName` set without adding a new API.

In CSS, if a color is specified for a div, it won't apply to the links underneath it. There has to be a specific rule targeting links in order to style them. Perhaps Apple wanted to mirror the way link styling is handled in CSS.

bsmith11:
[...]

PR #1127, opened 2016-01-27
[ASTextKit] Use NSForegroundColorAttributeName for links in ASTextKitRenderer

samhsiung:
Addresses issue #967 by subclassing `NSLayoutManager` to force TextKit to use `NSForegroundColorAttributeName` in all cases including when rendering text with link attributes. Also adds tests for this by comparing against a `UITextView` with `linkTextAttributes` set to the same color.


appleguy:
yay, thank you @samhsiung !  Always high quality.

PR #1130, opened 2016-01-28
Add `setNeedsDataFetch` to trigger in-range `fetchData` calls

### LINKED ISSUE #844: [ASRangeController] Support "setNeedsDataFetch" or similar to enable efficiently reloading images.
Still loving this framework! Sorry I haven't been contributing lately, I'm drowning in work after our first release. I have been lurking though!

So I'm using `ASCellNode` in kind of a weird way. I've got really complex content, so I create the entire collection of nodes upfront, and then keep them up-to-date with my model by listening for `NSManagedObjectContextObjectsDidChangeNotification` on a background context. If I added or removed nodes, I inform the table view.

Trouble is, when I update `ASMultiplexImageNode.imageIdentifiers`, nothing happens until someone next calls `fetchData`. This is normally the range controller. But if the cell node is already in the loading range, then `fetchData` won't get called again. If I call `fetchData` every time I set `imageIdentifiers`, though, I may end up loading images for cells that are far out of the working range.

Any advice?

levi:
Github being down is doing some funky things with this PR. Not all the commits are showing up on the website as I'm writing this. If you're only seeing one commit, this problem is still persistent. There should be about 6 commits in this PR total.

Fixes #844

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -442,7 +442,6 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @see displaySuspended and setNeedsDisplay
  */
-
 - (void)recursivelyClearContents;
 
 /**
@@ -464,6 +463,13 @@ NS_ASSUME_NONNULL_BEGIN
  * @see [ASDisplayNode(Subclassing) fetchData] and [ASDisplayNode(Subclassing) clearFetchedData]
  */
 - (void)recursivelyFetchData;
+
+/**
+ * @abstract Marks the node as needing to call fetchData
+ * @discussion If the node is outside of the preload range, it is queued to call fetchData the next time it enters the range.
+ * Otherwise, fetchData is called immediately if the node is currently within the preload range.
+ */
+- (void)setNeedsDataFetch;
 
 /**
  * @abstract Toggle displaying a placeholder over the node that covers content until the node and all subnodes are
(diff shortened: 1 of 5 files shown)


AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -466,6 +465,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)recursivelyFetchData;
 
 /**
+ * @abstract Marks the node as needing to call fetchData
+ * @discussion If the node is outside of the preload range, it is queued to call fetchData the next time it enters the range.
+ * Otherwise, fetchData is called immediately if the node is currently within the preload range.
```
levi: This is out of date. Will do an update.

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -466,6 +465,12 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)recursivelyFetchData;
 
 /**
+ * @abstract Triggers a call to fetchData for nodes in the preload range
```
appleguy: Let's be more specific here, and say that is a node has an interface state that specifies it is in the fetch data state. Then sometime after this call, the fetch data method will be called.

Note that the terminology has changed a bit, And I've dropped a turn preload in favor of fetch data. This is because preload is too generic of the term and can be misunderstood as preparing display as well.

appleguy:
@Adlai-Holler

PR #1132, opened 2016-01-28
Switch to instance methods of draw and display

garrettmoon:
This patch switches to instance methods of draw and display for
ASTextNode and ASImageNode to attempt to increase their performance.

It also fixes some thread safety issues in ASImageNode which appear
to have been regressions (though probably not hit very often).

And it sets up work for allowing modification of CGContexts before
and after a node's contents are drawn.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -6,6 +6,8 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
+#import "_ASDisplayLayer.h"
+
 @interface ASDisplayNode (Beta)
 
 + (BOOL)shouldUseNewRenderingRange;
@@ -19,5 +21,26 @@
  * @see Full documentation in ASDisplayNode+FrameworkPrivate.h
  */
 - (void)recursivelyEnsureDisplaySynchronously:(BOOL)synchronously;
+
+- (void)drawRect:(CGRect)bounds withParameters:(id <NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelledBlock isRasterizing:(BOOL)isRasterizing;
+
+- (UIImage *)displayWithParameters:(id <NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelled;
+
+/**
+ * @abstract allow modification of a context before the node's content is drawn
+ *
+ * @discussion Set the block to be called after the context has been created and before the node's content is drawn.
+ * You can override this to modify the context before the content is drawn. You are responsible for saving and
+ * restoring context if necessary. Restoring can be done in contextDidDisplayNodeContent
+ * This block can be called from *any* thread and it is unsafe to access any UIKit main thread properties from it.
+ */
+@property (nonatomic, strong) ASDisplayNodeContextModifier willDisplayNodeContentBlock;
+
+/**
+ * @abstract allow modification of a context after the node's content is drawn
+ *
+ * @discussion
+ */
+@property (nonatomic, strong) ASDisplayNodeContextModifier didDisplayNodeContentBlock;
 
 @end
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -6,8 +6,6 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-#import "_ASDisplayLayer.h"
-
 @interface ASDisplayNode (Beta)
 
 + (BOOL)shouldUseNewRenderingRange;
@@ -22,10 +20,6 @@
  */
 - (void)recursivelyEnsureDisplaySynchronously:(BOOL)synchronously;
 
-- (void)drawRect:(CGRect)bounds withParameters:(id <NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelledBlock isRasterizing:(BOOL)isRasterizing;
-
-- (UIImage *)displayWithParameters:(id <NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelled;
-
 /**
  * @abstract allow modification of a context before the node's content is drawn
  *
@@ -34,13 +28,13 @@
  * restoring context if necessary. Restoring can be done in contextDidDisplayNodeContent
  * This block can be called from *any* thread and it is unsafe to access any UIKit main thread properties from it.
  */
-@property (nonatomic, strong) ASDisplayNodeContextModifier willDisplayNodeContentBlock;
+@property (nonatomic, strong) ASDisplayNodeContextModifier willDisplayNodeContentWithRenderingContext;
 
 /**
  * @abstract allow modification of a context after the node's content is drawn
  *
  * @discussion
  */
-@property (nonatomic, strong) ASDisplayNodeContextModifier didDisplayNodeContentBlock;
+@property (nonatomic, strong) ASDisplayNodeContextModifier didDisplayNodeContentWithRenderingContext;
 
 @end
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -145,8 +145,6 @@
 
 - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer;
 {
-  BOOL hasValidCropBounds = _cropEnabled && !CGRectIsNull(_cropDisplayBounds) && !CGRectIsEmpty(_cropDisplayBounds);
-
   return [[_ASImageNodeDrawParameters alloc] initWithBounds:self.bounds
                                                      opaque:self.opaque
                                               contentsScale:self.contentsScaleForDisplay
@@ -177,7 +175,6 @@
   
   CGRect bounds = (hasValidCropBounds ? cropDisplayBounds : parameters.bounds);
   BOOL isOpaque = parameters.opaque;
-  CGFloat contentsScaleForDisplay = parameters.contentsScale;
   UIColor *backgroundColor = parameters.backgroundColor;
   UIViewContentMode contentMode = parameters.contentMode;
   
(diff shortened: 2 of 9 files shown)


AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -20,4 +22,25 @@
  */
 - (void)recursivelyEnsureDisplaySynchronously:(BOOL)synchronously;
 
+- (void)drawRect:(CGRect)bounds withParameters:(id <NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelledBlock isRasterizing:(BOOL)isRasterizing;
```
appleguy: These method signatures look great, thank you for adding them! We might want to put the stuff in framework private for now, although I suppose it is reasonable to put it on a track to eventually make public. Mainly thinking that we won't really recommended for use, Because it should be quite rare to have a custom drawing node that occurs as frequently as image and text nodes that have demonstrated a performance hotspot in creating the display parameters on the main thread.

AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -20,4 +22,25 @@
...
+
+/**
+ * @abstract allow modification of a context before the node's content is drawn
+ *
+ * @discussion Set the block to be called after the context has been created and before the node's content is drawn.
+ * You can override this to modify the context before the content is drawn. You are responsible for saving and
+ * restoring context if necessary. Restoring can be done in contextDidDisplayNodeContent
+ * This block can be called from *any* thread and it is unsafe to access any UIKit main thread properties from it.
+ */
+@property (nonatomic, strong) ASDisplayNodeContextModifier willDisplayNodeContentBlock;
```
appleguy: I actually don't have a better name suggestion, but wonder if there is something we can do to improve this one. In particular, the name does not particularly capture that this Block is called from inside the drawing operation itself.

AsyncDisplayKit/ASDisplayNode+Beta.h
garrettmoon: Cool, I'll move

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -22,42 +23,34 @@
...
 @property (nonatomic, assign) CGFloat contentsScale;
 @property (nonatomic, retain) UIColor *backgroundColor;
 @property (nonatomic, assign) UIViewContentMode contentMode;
-@property (nonatomic, assign) CGRect cropRect;
-@property (nonatomic, copy) asimagenode_modification_block_t imageModificationBlock;
 
 @end
 
 // TODO: eliminate explicit parameters with a set of keys copied from the node
 @implementation _ASImageNodeDrawParameters
```
appleguy: If we end up keeping a drop parameters concept longer term, I wonder if we could use a struct for this and pass by reference as a void pointer or something? Definitely not a fully formed thought, just curious if we could further cut down on object cost for something called thousands of times in periods of a few seconds.  I will start by profiling this and may prove that it is unnecessary.

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -153,88 +147,104 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer;
...
+                                            backgroundColor:self.backgroundColor
+                                                contentMode:self.contentMode];
 }
 
-+ (UIImage *)displayWithParameters:(_ASImageNodeDrawParameters *)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelled
+- (UIImage *)displayWithParameters:(_ASImageNodeDrawParameters *)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelled
 {
-  UIImage *image = parameters.image;
-
+  ASDN::MutexLocker l(_imageLock);
```
appleguy: If at all possible, we should minimize the critical section to a compact set of lines that accesses all of the needed information to put it into local variables from the instance variables, then unlocks and allows the rest of the method to proceed

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -271,8 +285,9 @@ - (void)displayDidFinish
 {
   [super displayDidFinish];
 
+  ASDN::MutexLocker l(_imageLock);
   // If we've got a block to perform after displaying, do it.
-  if (self.image && _displayCompletionBlock) {
+  if (_image && _displayCompletionBlock) {
 
     // FIXME: _displayCompletionBlock is not protected by lock
     _displayCompletionBlock(NO);
```
appleguy: It is likely essential that we release the lock before calling this block. We may want to access the block pointer and store it in a local variable before calling it, which could be a plausible threading issue if the display completion block is changed in the case that a display is canceled or restarted while one is in progress.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -48,30 +42,16 @@ - (instancetype)initWithRenderer:(ASTextKitRenderer *)renderer
...
   if (self = [super init]) {
-    _renderer = renderer;
-    _textOrigin = textOrigin;
+    _bounds = bounds;
     _backgroundColor = backgroundColor;
   }
   return self;
 }
 
-- (void)dealloc
```
appleguy: Unless I'm missing something, please re add this. I recently audited the necessity of this code, and it is currently very important, although someday I would like to be able to make sure that system provided nodes--where we can ensure their thread safety on dealloc--can be de-allocated off the main thread. This is what the Paper app is able to do with its nodes. The reason this is not the case in the open source framework is that we cannot guarantee that a node will not have an instance variable for something like a gesture recognizer which must not be released off the main thread.

appleguy:
I am tremendously excited about this patch! It will definitely make a noticeable difference in the number of frame drops in our app. I think it can be landed in its current state, but it would be important to do a few follow-ups in it the next week or so.

AsyncDisplayKit/ASImageNode.mm
garrettmoon: I'm not opposed to do manual memory management for good perf :)

AsyncDisplayKit/ASImageNode.mm
garrettmoon: I *think* I've done that. the other parameters that are accessed after this are all now protected by _imageLock. I also have the unlock after.

garrettmoon: I *think* I've done that. the other parameters that are accessed after this are all now protected by _imageLock. I also have the unlock after accessing them?

AsyncDisplayKit/ASTextNode.mm
garrettmoon: Looks like this may have been needed on the TextNode itself instead / or as well. I'll move it there.

appleguy:
Sometimes I hate nullability lol:
ASDisplayNode.h:43:46: error: pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -291,6 +312,7 @@ - (void)setNeedsDisplayWithCompletion:(void (^ _Nullable)(BOOL canceled))display
 
   // Stash the block and call-site queue. We'll invoke it in -displayDidFinish.
   // FIXME: _displayCompletionBlock not protected by lock
```
appleguy: You can remove this fix me comment now, correct? Very nice!

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -420,13 +406,17 @@ - (NSArray *)exclusionPaths
...
   CGContextRef context = UIGraphicsGetCurrentContext();
   ASDisplayNodeAssert(context, @"This is no good without a context.");
-
+  
   CGContextSaveGState(context);
-
+  
+  ASTextKitRenderer *renderer = [self _rendererWithBounds:parameters.bounds];
+  UIEdgeInsets shadowPadding = [self shadowPaddingWithRenderer:renderer];
+  CGPoint textOrigin = CGPointMake(parameters.bounds.origin.x - shadowPadding.left, parameters.bounds.origin.y - shadowPadding.top);
```
appleguy: Mainly for readability, create a cgpoint bounds origin local variable?

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -435,28 +425,20 @@ + (void)drawRect:(CGRect)bounds withParameters:(ASTextNodeDrawParameters *)param
       UIRectFillUsingBlendMode(CGContextGetClipBoundingBox(context), kCGBlendModeCopy);
     }
   }
-
+  
   // Draw shadow
-  [[parameters.renderer shadower] setShadowInContext:context];
```
appleguy: Although the object creation was the main cost, it's awesome to see this diff remove 10 or more method calls to the parameters object as well!

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -1016,9 +998,15 @@ - (void)setShadowRadius:(CGFloat)shadowRadius
   }
 }
 
+//only safe to call on main thread
```
appleguy: Could you add just a couple words to explain what about this method is not thread safe?  Shouldn't require a multi line explanation, but I am left wondering if it is the shadow padding method or the renderer call.

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -224,25 +225,30 @@ - (asyncdisplaykit_async_transaction_operation_block_t)_displayBlockWithAsynchro
...
         return nil;
       }
-
+      
       ASDN_DELAY_FOR_DISPLAY();
-
-      UIImage *result = [[self class] displayWithParameters:drawParameters isCancelled:isCancelledBlock];
+      
+      UIImage *result = nil;
+      if (_flags.implementsInstanceImageDisplay) {
```
appleguy: Should we call the pre-and post context blocks here? If not, maybe add a quick comment explaining why not?

PR #1135, opened 2016-01-28
Observe decelerating scroll velocity for accurate direction reporting

levi:
We were only using the pan gesture's velocity method, which only returns a valid result when the gesture is being tracked. This new implementation identifies the general velocity of the deceleration after touch up and reports the value while the touch is not tracked in the collection/table view's scroll view.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -88,6 +88,8 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   NSMutableSet *_registeredSupplementaryKinds;
   
+  CGPoint _deceleratingVelocity;
+  
   /**
    * If YES, the `UICollectionView` will reload its data on next layout pass so we should not forward any updates to it.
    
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -629,6 +636,11 @@ - (void)layoutSubviews
 
 - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoint)velocity targetContentOffset:(inout CGPoint *)targetContentOffset
 {
+  _deceleratingVelocity = CGPointMake(
+    scrollView.contentOffset.x - targetContentOffset->x,
```
appleguy: we can probably trust the pointer is not nil, but should we check before dereferencing it?

AsyncDisplayKit/ASTableView.mm
```diff
@@ -563,14 +565,23 @@ - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger
 
 - (ASScrollDirection)scrollDirection
 {
-  CGPoint scrollVelocity = [self.panGestureRecognizer velocityInView:self.superview];
+  CGPoint scrollVelocity;
```
appleguy: Can't wait to move this logic into the node classes, and then have a shared superclass!

PR #1137, opened 2016-01-28
[ASVideoNode] Bug fix and default play button

lappp9:
•  Fixes format problem for some preview frames
•  Adds default play button when one isn't provied.

Discussion: Should probably add bool for disabling default play button?  It's 2am though so I'll add it laters.

### DIFF
--- a/AsyncDisplayKit/Private/ASDefaultPlayButton.m
+++ b/AsyncDisplayKit/Private/ASDefaultPlayButton.m
@@ -1,0 +1,72 @@
+//
+//  ASDefaultPlayButton.m
+//  AsyncDisplayKit
+//
+//  Created by Luke Parham on 1/27/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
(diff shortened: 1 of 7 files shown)


AsyncDisplayKit/Private/ASDefaultPlayButton.m
```diff
@@ -0,0 +1,71 @@
...
+  if (bounds.size.width > bounds.size.height) {
+    //use the height
+    widthHeight = bounds.size.height/2;
+    originX = (bounds.size.width - widthHeight)/2;
+    buttonBounds = CGRectMake(originX, (bounds.size.height - widthHeight)/2, widthHeight, widthHeight);
+  }
+  
+  if (!isRasterizing) {
+    [[UIColor clearColor] set];
+    UIRectFill(bounds);
```
appleguy: This has no effect because clear color is completely transparent, correct? What is the intent here, maybe there is a better way?

PR #1139, opened 2016-01-28
[ASTableView] Respect section inserts and deletions in content offset adjustment process

zintus:
Implementation looks quite straightforward, did i miss any intricate detail?


appleguy:
@nguyenhuy, @knopp, @levi do you have time to help me review this? Thanks for the change! Very glad to have some people helping out in this code in such a material way, @zintus <3 :)

appleguy:
@zintus thank you for this change!  Sorry it has taken this long to review it. This looks very good to me, but I am going to defer to the people who actively use this feature in their applications to make sure they have an opportunity to test it as well as do a quick review.

@eanagel, @Adlai-Holler — I think you both heavily rely on this content adjustment feature.  Thoughts?

Adlai-Holler:
I haven't used the content offset adjustment in a while so nothing useful from me. 😬

AsyncDisplayKit/ASTableView.mm
```diff
@@ -903,6 +915,21 @@ - (void)rangeController:(ASRangeController *)rangeController didDeleteSectionsAt
...
   }
 
+  if (_contentOffsetAdjustmentTopVisibleRow != nil) {
+    __block NSInteger adjustedSection = _contentOffsetAdjustmentTopVisibleRow.section;
+    __block NSInteger adjustedRow = _contentOffsetAdjustmentTopVisibleRow.row;
+    [indexSet enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
+        if (adjustedSection > idx) {
+          adjustedSection--;
+        } else if (adjustedSection == idx) {
+          adjustedRow = 0;
```
eanagel: Things can get tricky here if there are sections with no rows. To do this really correctly we would need to detect zero-row sections and skip them (because in this case adjustedRow = 0 would be an invalid index path.) This could be done outside the enumerate block or perhaps in the else where we actually set adjustedRow to 0. I'm not sure offhand what the best approach is to get the valid (and safe) row count at this point in the code base. Unfortunately, zero length sections are a fairly common way to optionally hide content, so this could happen.

zintus: From my understanding it's fine to have non-existent index in `_contentOffsetAdjustmentTopVisibleRow`, because it's never used to read nodes, it's used only to know the fact of being above or below first row.

appleguy:
@zintus I would love to get this in now, but I'm just a bit nervous not having anything that tests this. It's okay if you don't have time @zintus, but one great thing to have would be a Single unit test that fails before this change and passes afterward. Another approach might be upgrading one of the existing example applications to use this feature

AsyncDisplayKit/ASTableView.mm
eanagel: Hmmm in looking at the code I think you're right about that. I don't see you updating the actual _contentOffsetAdjustment, though to account for the height of the added or removed sections if they are before the topVisibleRow. Is that happening in a subtle way that I've missed?

levi:
This is something we should put into a sample project for testing. @zintus are you available to help us get this into a mergeable state?

appleguy:
I think this is very very close, and I almost just closed my eyes and merged it :). The code looks correct, but even the simplest example would really help improve confidence.

If you want to create a new example app, just do Cmd-D in the finder on any app in examples/ -- the Xcode files and scheme are all called Sample for this reason.  If you can see a way to add it to any existing example as a new feature rather than creating a new one, that would be even better.

Hannah, could you add a documentation page we can point folks to with instructions on how to make an example app?  Later we might create a specific base project that limits the files that would otherwise need to be deleted.

hannahmbanana:
@appleguy: I've seen several contentOffset open issues that we should follow up / close. #1411

AsyncDisplayKit/ASTableView.mm
appleguy: cc @zintus - I'd love to get this landed.  Do you have any thoughts on @eanagel's question here?

@maicki do you think we should land this without further testing?  Since it looks like we may not be able to get a test project, and this feature is not used by that many people, it's probably better to take the fix and rely on @zintus' own testing.

hannahmbanana:
Just want to note that @alibasta checked this against his issue (#1411) and it resolved it.

hannahmbanana:
@Adlai-Holler @maicki: This is just needs a line by line read by 1 of you to make sure there are no off-by one indexing issues or anything else.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -888,6 +888,18 @@ - (void)rangeController:(ASRangeController *)rangeController didInsertSectionsAt
     return; // if the asyncDataSource has become invalid while we are processing, ignore this request to avoid crashes
   }
 
+  if (_contentOffsetAdjustmentTopVisibleRow != nil) {
+    __block NSInteger adjustedSection = _contentOffsetAdjustmentTopVisibleRow.section;
+    [indexSet enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
+        if (adjustedSection <= idx) {
+          adjustedSection++;
```
Adlai-Holler: @zintus Should this be `if (idx <= adjustedSection)`? Aren't we interested in sections that were inserted above the top visible row? Also as a bonus, would you mind adding `else { *stop = YES; }`?

Adlai-Holler:
@zintus Could you give an update on this? Would love to get it landed!

PR #1140, opened 2016-01-28
Added ASTextNodeWordKerner support to ASTextNode

rcancro:
* Added a member to `ASTextKitAttributes` that is an optional delegate to the struct's NSLayoutManager.
* Changed ASTextNode to set this delegate to an instance of ASTextNodeWordKerner.
* Updated init method of `ASTextKitContext` to take an optional NSLayoutManager delegate
* Added the files in TextKit folder to the public headers (so we can include ASTextNodeTypes.h)

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -21,6 +21,7 @@
 #import "ASTextKitRenderer.h"
 #import "ASTextKitRenderer+Positioning.h"
 #import "ASTextKitShadower.h"
+#import "ASTextNodeWordKerner.h"
 
 #import "ASInternalHelpers.h"
 #import "ASEqualityHelpers.h"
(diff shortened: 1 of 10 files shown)


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -143,6 +148,8 @@ - (instancetype)init
     // on the special placeholder behavior of ASTextNode.
     _placeholderColor = ASDisplayNodeDefaultPlaceholderColor();
     _placeholderInsets = UIEdgeInsetsMake(1.0, 0.0, 1.0, 0.0);
+    
+    _wordKerner = [[ASTextNodeWordKerner alloc] init];
```
appleguy: Can you make this lazy initialize? The init method of text node is already in great need of optimization… of course you should not worry about that right now, but check out ASTextKitRenderer's mechanisms to use getter methods for lazy initialization.  This has the additional benefit of making any dependencies more clear.

PR #1141, opened 2016-01-28
[ASBatchFetching] Support head fetching

tomizimobile:
I created this to support a chat-like tableView or collectionView where the content stays focused at the bottom, and old messages are fetched as you scroll up. This changes the existing delegate methods related to batch fetching to include a parameter `atTail` so that you can easily choose to fetch at the head, tail, or both. 

It doesn't look like there's a way in ASCollectionView or ASTableView to force it to load contents initially at the bottom, which is not required to make this work but it certainly goes hand-in-hand. I don't really love the way I made that work in my app so I haven't tried to create that functionality just yet. If I get time to (or someone else wants to), I could see where tying this directly to some property on ASCollectionView and ASTableView that dictates if it's a bottom-loaded scrollView could get rid of the extra property `atTail`.

This also fixes a couple incorrectly-written tests, -testHorizontalScrollToLessThanLeading and -testHorizontalScrollingPastContentSize.


samhsiung:
Hi @tomizimobile , thanks for bringing this up. I think another approach we could consider is making ASTableView and ASCollectionViews transform aware for batch fetching logic when using the scrollDirection. This wouldn't require a change to delegates. 

Instead, all you have to do is invert the transform of the scroll view and its direct subviews.

You can check out SLKTextViewController for a sample implementation. What do you think about this approach?

samhsiung:
Put up a PR for what I described here: https://github.com/facebook/AsyncDisplayKit/pull/1143

appleguy:
@tomizimobile, interested in your thoughts on the transform approach. I think it would reduce the total amount of code we have to maintain while providing equally good performance. However, it is very possible that there is some other consequence of the approach that I haven't considered that limits certain use cases?

If we are in agreement on this approach, @samhsiung would you be able to implement a BOOL option on the collection and table -- when enabled, applies the transform to the table and each cell?

Of course this is pretty easy for anyone to implement themselves, but because the approach is far from obvious to many people (including myself when I first heard of it years ago), I think it is justified to have a convenience API.  Maybe scrollsFromBottom or something.  A similar approach could support horizontal views that need to load leftwards, So maybe there is a more general name?

tomizimobile:
@samhsiung Great idea! I haven't come across that approach before - definitely not obvious but smart. @appleguy I'd be comfortable going with this approach rather than mine, and would definitely like some property such as scrollFromEnd or something like that. 

There are a couple use cases we'd lose by going with the transform approach:
- If you have a normal-scrolling table/collectionView that you wanted to fetch from the head. I don't think this is common though, and a refresh control seems to be the more standard way to handle that. 
- If you wanted to support fetching from both the head and tail. This is another case that I'd believe is pretty uncommon. 

I think the transform approach is going to support the much more common scenario of having a table/collectionView that scrolls from the bottom.

robb:
> If you wanted to support fetching from both the head and tail. This is another case that I'd believe is pretty uncommon.

I'm currently wondering how to implement head fetching precisely for this use case, a long timeline that users start in the middle of and scroll in either direction. While I agree that it might not be very common, it would be unfortunate to have to go against the grain of ASKD to make it happen.

keleftheriou:

Adlai-Holler:
@tomizimobile @samhsiung Hey folks! ASTableView and ASCollectionView are now transform-aware and so we should cover the most common case (Slack-style messaging). Since the codebase has diverged so significantly since this was last updated, I'm closing this for now. Thanks for the insight and implementation work on this!

Issue #1126, opened 2016-01-27
[ASTableView] Make ASTableViews and ASCollectionViews transform aware when tail fetching


samhsiung:
Currently ASTableView only supports batch fetching when `scrollDirection =  ASScrollDirectionUp` due to only handling tail fetching. However when the scroll view is vertically inverted via transform we actually want to watch for `ASScrollDirectionDown` instead of `ASScrollDirectionUp` for tail fetching.

This will add support for libraries such as SLKTextViewController which invert the scroll view so that messages appear in ascending order based on time.

samhsiung:
https://github.com/facebook/AsyncDisplayKit/pull/1143

appleguy:
@samhsiung thanks for finding this, and the fix!

PR #1143, opened 2016-01-29
[ASTableView] make scrollDirection transform aware in ASCollectionView and ASTableView

samhsiung:
Addresses Issue #1126 by adding support for batch fetching for ASTableView and ASCollectionViews with inverted transforms by having the scrollDirection take the current transform into account.


appleguy:
Nicely done, thanks @samhsiung!

PR #1147, opened 2016-01-29
[ASVideoNode] Video muting, tap delegate override, general cleanup

lappp9:
Removed unnecessary clear color and improved comment


lappp9:
Hey @appleguy, I added muting the video and a delegate method to override tap behavior.

appleguy:
Nice refinements, thanks a lot for sending these great incremental patches as they are available @lappp9 !

PR #1148, opened 2016-01-29
Fix text truncation

tomizimobile:
It seems that text truncation doesn't work if the truncater is initialized after the context is. I previously mentioned a fix in https://github.com/facebook/AsyncDisplayKit/issues/1033 which initialized the truncater in -init, but I think it can be deferred to -_calculateSize.

More background: https://github.com/facebook/AsyncDisplayKit/pull/1110


appleguy:
I'm worried that the unit tests were not detecting this

PR #1152, opened 2016-01-29
Allow images returned by placeholderImage to be stretchable

garrettmoon:


appleguy:
Very happy about this!

PR #1155, opened 2016-01-31
Reduce early creation of views and layers in collection view

nguyenhuy:
by avoiding pre-rendering when it is not visible or not scrolling.

Two things to consider:
- [x] Make render/display range a bit bigger than visible range, so it feels better when user starts scrolling. The difficult thing is to figure out a good value here.
- [x] If user comes back to the collection view after leaving it for a while (magic words: open close-up and go back to feed), scroll direction is none (again), causing the render range to shrink. This *may* cause some hard earned views to be released. I need to spend more time investigating this theory though.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -91,6 +91,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) id<ASCollectionViewLayoutInspecting> layoutInspector;
 
 /**
+ * The ASInterfaceState of this collection view.
+ */
+@property (nonatomic, readonly) ASInterfaceState interfaceState;
+
+/**
  *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. This method must be called from the main thread. 
  *  The asyncDataSource must be updated to reflect the changes before the update block completes.
  *
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -91,11 +91,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) id<ASCollectionViewLayoutInspecting> layoutInspector;
 
 /**
- * The ASInterfaceState of this collection view.
- */
-@property (nonatomic, readonly) ASInterfaceState interfaceState;
-
-/**
  *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. This method must be called from the main thread. 
  *  The asyncDataSource must be updated to reflect the changes before the update block completes.
  *
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -242,19 +242,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   return _flowLayoutInspector;
 }
 
-- (ASInterfaceState)interfaceState
-{
-  ASCollectionNode *collectionNode = self.collectionNode;
-  if (collectionNode) {
-    return collectionNode.interfaceState;
-  } else {
-    // Until we can always create an associated ASCollectionNode without a retain cycle,
-    // we might be on our own to try to guess if we're visible.  The node normally
-    // handles this even if it is the root / directly added to the view hierarchy.
-    return (self.window != nil ? ASInterfaceStateVisible : ASInterfaceStateNone);
-  }
-}
-
 #pragma mark -
 #pragma mark Overrides.
 
@@ -841,7 +828,15 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 
 - (ASInterfaceState)interfaceStateForRangeController:(ASRangeController *)rangeController
 {
-  return self.interfaceState;
+  ASCollectionNode *collectionNode = self.collectionNode;
+  if (collectionNode) {
+    return collectionNode.interfaceState;
+  } else {
+    // Until we can always create an associated ASCollectionNode without a retain cycle,
+    // we might be on our own to try to guess if we're visible.  The node normally
+    // handles this even if it is the root / directly added to the view hierarchy.
+    return (self.window != nil ? ASInterfaceStateVisible : ASInterfaceStateNone);
+  }
 }
 
 - (NSArray *)rangeController:(ASRangeController *)rangeController nodesAtIndexPaths:(NSArray *)indexPaths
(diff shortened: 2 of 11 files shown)


AsyncDisplayKit/ASCollectionView.h
```diff
@@ -91,6 +91,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) id<ASCollectionViewLayoutInspecting> layoutInspector;
 
 /**
+ * The ASInterfaceState of this collection view.
+ */
+@property (nonatomic, readonly) ASInterfaceState interfaceState;
```
appleguy: Is there any need to expose this is public API? I am asking because in 2.0, we are going to eliminate the ability to directly initialize this class, and the official interface will be the node subclasses, which already have this property defined. So it may make more sense for now to have just the range controller delegate method provide access to this information.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -535,16 +547,16 @@ - (ASScrollDirection)_scrollDirectionForVelocity:(CGPoint)scrollVelocity
   ASScrollDirection scrollableDirections = [self scrollableDirections];
   
   if (ASScrollDirectionContainsHorizontalDirection(scrollableDirections)) { // Can scroll horizontally.
-    if (scrollVelocity.x >= 0) {
```
appleguy: Great change, clearly this is an important detail!

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -829,15 +841,7 @@ - (CGSize)viewportSizeForRangeController:(ASRangeController *)rangeController
...
-  ASCollectionNode *collectionNode = self.collectionNode;
-  if (collectionNode) {
-    return self.collectionNode.interfaceState;
-  } else {
-    // Until we can always create an associated ASCollectionNode without a retain cycle,
-    // we might be on our own to try to guess if we're visible.  The node normally
-    // handles this even if it is the root / directly added to the view hierarchy.
-    return (self.window != nil ? ASInterfaceStateVisible : ASInterfaceStateNone);
-  }
+  return self.interfaceState;
```
appleguy: We could simply leave this method alone, Or we could keep the newly defined interface state method and simply not expose that API. We could also put that API in the +Beta header

AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
```diff
@@ -164,4 +165,19 @@ - (CGRect)rangeBoundsWithScrollDirection:(ASScrollDirection)scrollDirection
   return CGRectExpandToRangeWithScrollableDirections(rect, tuningParameters, _scrollableDirections, scrollDirection);
 }
 
+- (ASRangeTuningParameters)tuningParametersForRangeType:(ASLayoutRangeType)rangeType
+{
+    BOOL isVisible = ASInterfaceStateIncludesVisible(_collectionView.interfaceState);
```
appleguy: Okay, now I see where this is used :-) maybe we could put this method definition in an internal header.

AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
```diff
@@ -164,4 +165,19 @@ - (CGRect)rangeBoundsWithScrollDirection:(ASScrollDirection)scrollDirection
...
+    BOOL isScrolling = (_collectionView.scrollDirection != ASScrollDirectionNone);
```
appleguy: How should we think about the distinction between a view that has appeared and never been scrolled before, versus a view that has been scrolled and come to a stop?

In the latter case, we probably don't want to restrict the Display range once it stops and destroy objects that were prepared in case the user scrolls again, because it is likely in that case. I wonder if there is room for two different variables here, Or an additional bit in the bit field

AsyncDisplayKit/Details/ASCollectionViewLayoutController.mm
```diff
@@ -164,4 +165,19 @@ - (CGRect)rangeBoundsWithScrollDirection:(ASScrollDirection)scrollDirection
...
+    // When the collecion view is not visible or not scrolling, make display range only as big as visible range.
+    // This reduce early creation of views and layers.
+    if (!isVisible || !isScrolling) {
+        if (rangeType == ASLayoutRangeTypeDisplay) {
+            return [super tuningParametersForRangeType:ASLayoutRangeTypeVisible];
```
appleguy: Technically, this behavior should be implemented in the range controller. It is fairly expensive to request the index paths that lie within a range, so if we are going to use an identical range, we should simply use the same NSSet from one request. The biggest question is how we could expand this range slightly with a minimum of cost. It probably makes sense to keep the independent calculation of the display range for now, whenever it is different than the visible range, even though that would be less efficient than simply reusing the visible range -- there is probably no generalizable way to extend the range "a bit" without running the full range index path calculation, at least for collection view custom layouts.  In the state where the view has never been scrolled before, defined the display range to be something like one quarter of the screen in each direction beyond visible, leading and trailing (of course scroll direction is unknown at this point, and in many cases the view may only be scrollable in one direction, so that will automatically ensure the other direction isn't adding cost).

A very important case to consider is the pager node. When we first open the page in the pager node, we ideally you would want that one page to load very quickly, while around the same time calling fetch data for the pages to the left and right, without including them in the display range. We need to create a new notification / delegate call of some kind to trigger when most of the loading for the visible area is done, and only at that time, extend the display range to include the left and right pages while it also extending the data range to include two pages to the left and right. In this state, the display range for the 1-left and 1-right pages should be identical to the visible area of those pages, allowing us to reduce calls to the collection layout to a maximum of 2.  We should also constrain the fetch data range to not go as far as it normally does for an item which isn't visible at all yet, e.g. right now for those left and right pages we may be fetching 2-3 screens vertically off the bottom even when the user may never begin viewing that page. It is only once those pages have become visible that the display range should include one quarter of the screen in each the leading and trailing direction, and once the user initiates the very first scroll on that view, that we should immediately expand them to the full / normal ranges.

This idea of scaling the ranges is fairly new, and how we handle ranges today does not have a great mechanism to do it. The right implementation may be to apply a percentage scale factor of some kind to the ranges, or to name these different kinds of states (maybe a bitfield encompassing states such as, offscreen by one, or on the screen but not yet scrolled in the direction of the range) to allow some degree of tuning in control over this behavior for different types of content that may have different requirements in terms of and resource consumption.  With a small number of new abstractions in this area, we should be able to automate certain optimization such as checking if a range is the same and only requesting from the layout controller that set of index paths once and using that set for any appropriate calculations in the range controller.

nguyenhuy:
@appleguy Updated the diff to introduce `minimumTuningParameters`. Range controller now selects whether the normal or minimum ranges should be used. The minimum ones are used when the view is not visible, not scrolling or normal ranges were not used before. We switch to normal ones once user starts scrolling, which I think is a good signal. 

Working on a layout delegate that will trigger a switch to normal ranges once cells in minimum ones are taken cared of.

appleguy:
@nguyenhuy Love this diff!  Merging now and let's go from here.

Issue #997, opened 2015-12-26
[Coveralls / Slather] Test coverage statistics not working with Xcode 7.


appleguy:
Originally added in this commit, Coveralls has proved to be a very useful tool:

https://github.com/facebook/AsyncDisplayKit/pull/434

I've spent a couple hours so far looking into how this stuff is implemented, and checked all appropriate project settings, but unfortunately am not having any luck restoring it.  The code coverage stuff is crashing the tests at the end, too.  It looks like Apple is aggressively deprecating gcov and trying to replace it with llvm-cov, but I don't see evidence that Coveralls supports that — perhaps we need to switch to another tool?

@dasmer, @modocache, @Daniel1of1 - do any of you guys have experience with Xcode 7 gcov support, or have time to investigate this?  I'm glad we are on Xcode 7 on Travis now, but am regretting it a bit because the coverage stats are critical for the contributors who are just now ramping up on writing a new wave of tests for AsyncDisplayKit.

appleguy:
This may need to be reverted, because it disables instrument program flow for the test target and test host (although leaves it on for the frameworks), but that is the only way I could stop the error spew.  Code coverage does work in the Xcode 7 UI after running tests.

https://github.com/facebook/AsyncDisplayKit/pull/998

The error spew is the same as this: https://github.com/markohlebar/Peckham/issues/69 - they do have a PR to "fix" it but I don't think they use Coveralls, which is really the only thing not working at this point.  https://github.com/markohlebar/Peckham/pull/70

modocache:
Sorry, I haven't had the time to look into coverage on Xcode 7 yet. We've reverted slather integration on ComponentKit, but not due to Xcode issues: https://github.com/facebook/componentkit/pull/433

Would be very interested in hearing how you proceed here.

appleguy:
@modocache - The community is planning an exciting long term road map for the project and getting organized around how to deliver these feature requests. We are moving to a new issue tracker with more sophisticated planning capabilities (check out www.realArtists.com and their new product: Ship!). 

If you are interested in helping contribute to this component or any other, don’t hesitate to send me an email at AsyncDisplayKit@gmail.com. If you would like to contribute for a few weeks, I can also add you to our Ship bug tracker so that you can see what everyone is working on and actively coordinate with us. 

As always, keep filing issues and submitting pull requests here on Github and we will only move things to the new tracker if they require long term coordination.

PR #1157, opened 2016-02-01
Disable "Instrument Program Flow" for Release Build of iOS Framework

nesium:
Hi. I've disabled the "Instrument Program Flow" setting because that would spam the console when testing a project which links against AsyncDisplayKit. This had already been set for the internal test targets in #998. But this problem occurs in external projects as well.

This appears to happen due to a bug in Xcode 7, as mentioned [here](https://forums.developer.apple.com/thread/9765), [here](https://github.com/specta/specta/issues/167) and [here](http://stackoverflow.com/questions/22519530/dozens-of-profilinginvalid-arc-tag-when-running-code-coverage-in-xcode-5).

The change might break the code coverage testing but that appears to already be the case: #997. But that was the reason why I've changed it for the Release build only.


RuiAAPeres:
I can confirm:

>  (...)  spam the console when testing a project which links against AsyncDisplayKit

appleguy:
@nesium thank you, this drove me crazy and I literally gave up after trying so many things!

PR #1158, opened 2016-02-01
Resolve Build Error When use_frameworks! Option Set in CocoaPods

Adlai-Holler:
Boom! #1153 resolved.

PR #1160, opened 2016-02-01
Update Swift Example

Adlai-Holler:
- Tweak `Kittens` to use `dispatch_after` rather than sleeping on a background thread.
- Update Swift example to `use_frameworks!` instead of bridging from Objective-C. This is the preferred method.
- Update Swift example project to latest recommended project settings.
- Cherry-pick 819a09d from #1158 so that it builds.
- Rewrite the Swift example's view controller to use `ASViewController` and support batch-fetching and all-around get schwifty.


appleguy:
Awesome!!  Thank you!

Issue #1145, opened 2016-01-29
[ASControlNode] init method can be optimized.


appleguy:
This is a relatively small optimization, but will help, because nearly everything is an ASControlNode - such as ASTextNode.
Right now we set self.userInteractionEnabled = NO in ASControlNode's init. We should delay doing this until -didLoad, and after calling super, setting this property will be faster.
We need a way to track if addTarget: has occurred, because if it has, then we should leave this untouched. In cases where addTarget is used, it will almost always be called before -didLoad.
We should introduce another BOOL which is _hasAtLeastOneTarget (find a better name). Then _hasAtLeastOneTarget can be used to bypass the touchesBegan, etc methods that _enabled is currently not protecting very well (since it needs to be set to YES in -init, and is a separate API level control), and then _hasAtLeastOneTarget can also be used in -didLoad to set userInteractionEnabled.
In the trace below there's also evidence that a similar change would benefit -[ASTextNode init], deferring its various default property sets to get applied in -didLoad if they are needed at all.
Running Time	Self (ms)	Symbol Name
58.0ms 100.0%	0.0	-[ASTextNode init]
13.0ms 22.4%	0.0	-[ASControlNode init]
8.0ms 13.7%	0.0	-[ASDisplayNode(UIViewBridge) setUserInteractionEnabled:]
5.0ms 8.6%	0.0	-[ASDisplayNode pendingViewState]
3.0ms 5.1%	0.0	_objc_rootAlloc
2.0ms 3.4%	0.0	-[_ASPendingState init]
2.0ms 3.4%	0.0	ASDN::Locker<ASDN::Mutex>::~Locker()
1.0ms 1.7%	1.0	objc_object::sidetable_release(bool)
4.0ms 6.8%	0.0	-[ASDisplayNode init]
1.0ms 1.7%	1.0	DYLD-STUB$$objc_msgSend
12.0ms 20.6%	1.0	-[ASDisplayNode(UIViewBridge) setBackgroundColor:]
8.0ms 13.7%	0.0	-[ASDisplayNode(UIViewBridge) setOpaque:]
4.0ms 6.8%	2.0	-[ASDisplayNode(UIViewBridge) shadowColor]
3.0ms 5.1%	1.0	-[ASDisplayNode(UIViewBridge) shadowOffset]
2.0ms 3.4%	0.0	-[ASDisplayNode(UIViewBridge) shadowOpacity]
[...]

PR #1162, opened 2016-02-02
[ASControlNode] optimize init method

maicki:
Should Address Issue #1145 by introducing a new instance variable _targetAdded that delays setting of self.userInteractionEnabled to viewDidLoad.

### DIFF
--- a/AsyncDisplayKit/ASControlNode+Subclasses.h
+++ b/AsyncDisplayKit/ASControlNode+Subclasses.h
@@ -20,6 +20,12 @@ NS_ASSUME_NONNULL_BEGIN
 @interface ASControlNode (Subclassing)
 
 /**
+ @abstract Indicates whether or not at least one target was added to the receiver
+ @discussion YES if the receiver has at least one target; NO otherwise.
+ */
+@property (nonatomic, readonly, assign, getter=isTargetAdded) BOOL targetAdded;
+
+/**
  @abstract Sends action messages for the given control events.
  @param controlEvents A bitmask whose set flags specify the control events for which action messages are sent. See "Control Events" in ASControlNode.h for bitmask constants.
  @param touchEvent An event object encapsulating the information specific to the user event.
--- a/AsyncDisplayKit/ASControlNode+Subclasses.h
+++ b/AsyncDisplayKit/ASControlNode+Subclasses.h
@@ -20,12 +20,6 @@ NS_ASSUME_NONNULL_BEGIN
 @interface ASControlNode (Subclassing)
 
 /**
- @abstract Indicates whether or not at least one target was added to the receiver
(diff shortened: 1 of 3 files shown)


appleguy:
Nice tweak, thanks @maicki !  Sometimes surprising what the profiler will find...

AsyncDisplayKit/ASControlNode+Subclasses.h
```diff
@@ -20,6 +20,12 @@ NS_ASSUME_NONNULL_BEGIN
 @interface ASControlNode (Subclassing)
 
 /**
+ @abstract Indicates whether or not at least one target was added to the receiver
+ @discussion YES if the receiver has at least one target; NO otherwise.
+ */
+@property (nonatomic, readonly, assign, getter=isTargetAdded) BOOL targetAdded;
```
appleguy: @maicki is this needed?  We should probably declare it internal to the .m file, or just use the instance variable, unless there is a clear usecase for another class (incl. subclass) to use it.  I think we should expect the base class to internally manage this as we should expect any usage or subclass to call addTarget(...).

maicki: @appleguy Yep makes sense. The initial thoughts was that e.g. ASTextNode would be able to check if a target was added, but you are right there is no use case at the moment. I removed the property and moved to use the instance variable only.

AsyncDisplayKit/ASControlNode.m
```diff
@@ -254,6 +269,7 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
       [targetActions addObject:NSStringFromSelector(action)];
     });
 
+  _targetAdded = YES;
```
appleguy: We would need to clear this in the remove target method, right? In that case, does it make sense to instead remove the state and inspect _controlEventDispatchTable, potentially in an accessory method like -(BOOL)hasTarget ?

maicki: That was my first idea to create an accessory method and basically return _controlEventDispatchTable.count. You think there will be any performance hits due to checking of the count property? That was the main reason I decided against it.

appleguy: @maicki that is a reasonable concern, and there definitely would be some amount of overhead. However, in the code path we care about most, this variable will be nil and so it will be extremely fast.  I think the value of limiting total state and eliminating the possibility of it getting out of sync, for example with the remove target case, is high enough to suggest we should go that route that you initially considered!

appleguy:
Thanks a lot, @maicki!  Many details to consider with this one.  But, I'm very proud of the fact that this framework represents the combined work of many detail-oriented engineers, so your contributions are a big part of that!

PR #1163, opened 2016-02-02
Range controller automatically extends to full mode

nguyenhuy:
Follow up on #1155: When range controller uses minimum mode, it observes to rendering engine and extends to full mode if conditions meet.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -29,6 +29,7 @@
 #import "ASCellNode.h"
 
 NSInteger const ASDefaultDrawingPriority = ASDefaultTransactionPriority;
+NSString * const ASRenderingEngineDidDisplayScheduledNodesNotification = @"ASRenderingEngineDidDisplayScheduledNodes";
 
 @interface _ASDisplayNodePosition : NSObject
 
@@ -279,6 +280,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
       for (ASDisplayNode *node in displayingNodes) {
         [node __recursivelyTriggerDisplayAndBlock:NO];
       }
+      [[NSNotificationCenter defaultCenter] postNotificationName:ASRenderingEngineDidDisplayScheduledNodesNotification
+                                                          object:nil];
     });
   }
 }
--- a/AsyncDisplayKit/Details/ASAbstractLayoutController.mm
+++ b/AsyncDisplayKit/Details/ASAbstractLayoutController.mm
@@ -11,8 +11,7 @@
 #include <vector>
 
 @interface ASAbstractLayoutController () {
-  std::vector<ASRangeTuningParameters> _tuningParameters;
-  std::vector<ASRangeTuningParameters> _minimumTuningParameters;
+  std::vector<std::vector<ASRangeTuningParameters>> _tuningParameters;
   CGSize _viewportSize;
 }
 @end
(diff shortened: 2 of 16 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -279,6 +280,8 @@ + (void)scheduleNodeForRecursiveDisplay:(ASDisplayNode *)node
       for (ASDisplayNode *node in displayingNodes) {
         [node __recursivelyTriggerDisplayAndBlock:NO];
       }
+      [[NSNotificationCenter defaultCenter] postNotificationName:ASRenderingEngineDidDisplayScheduledNodesNotification
```
appleguy: Do we need away to check if a specific node is in the set that displayed? Maybe we can actually pass the displaying nodes object as the object, and any listeners can check for contains object?

AsyncDisplayKit/Details/ASAbstractLayoutController.mm
```diff
@@ -25,32 +24,32 @@ - (instancetype)init
     return nil;
   }
   
-  _tuningParameters = std::vector<ASRangeTuningParameters>(ASLayoutRangeTypeCount);
-  _tuningParameters[ASLayoutRangeTypeVisible] = {
+  _tuningParameters = std::vector<std::vector<ASRangeTuningParameters>> (ASLayoutRangeModeCount, std::vector<ASRangeTuningParameters> (ASLayoutRangeTypeCount));
+  
+  _tuningParameters[ASLayoutRangeModeMinimum][ASLayoutRangeTypeVisible] = {
```
appleguy: Maybe it is time to remove the visible type from range parameters? This used to exist because the old range controller implementation had individual range handlers and it helped propagate the pattern, but now we can quite easily assuming that they are zero and make the appropriate call in the range controller. If you see any value to having them be explicit, feel free to keep them.

AsyncDisplayKit/Details/ASAbstractLayoutController.mm
```diff
@@ -25,32 +24,32 @@ - (instancetype)init
...
     .leadingBufferScreenfuls = 0,
     .trailingBufferScreenfuls = 0
   };
-  _tuningParameters[ASLayoutRangeTypeDisplay] = {
-    .leadingBufferScreenfuls = 1.5,
-    .trailingBufferScreenfuls = 0.75
+  _tuningParameters[ASLayoutRangeModeMinimum][ASLayoutRangeTypeDisplay] = {
```
appleguy: I know that pager node has custom values for ranges, and we should probably have it set up appropriate minimum ranges as well.

AsyncDisplayKit/Details/ASLayoutRangeType.h
```diff
@@ -8,6 +8,21 @@
 
 #import <Foundation/Foundation.h>
 
+/// Each mode has a complete set of tuning parameters for range types.
```
appleguy: // instead of /// ?

AsyncDisplayKit/Details/ASLayoutRangeType.h
```diff
@@ -8,6 +8,21 @@
...
+/// Depends on some conditions (including interface state and direction of the scroll view, state of rendering engine, etc),
```
appleguy: Depending

AsyncDisplayKit/Details/ASLayoutRangeType.h
```diff
@@ -8,6 +8,21 @@
...
+/// a range controller can choose which mode it should use at a given time.
+typedef NS_ENUM(NSUInteger, ASLayoutRangeMode) {
+  /// Minimum mode is used when a range controller should limit the amount of work it performs.
+  /// Thus, less views/layers are created and less data is fetched.
```
appleguy: fewer views / layers [...], saving system resources

[...] when conditio

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
```diff
@@ -38,16 +40,49 @@ - (instancetype)init
...
   _rangeIsValid = YES;
+  _currentRangeMode = ASLayoutRangeModeCount;
   
   return self;
 }
 
+- (void)dealloc
+{
+  if (_didRegisterForNotifications) {
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
```
appleguy: I have always preferred this pattern as well, rather than specifying the notification to remove. But because our app uses notifications so heavily, I have heard that this is meaningfully less efficient then removing this specific notification. Could you change it to two calls to this to specify the notification that we subscribe to?

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
```diff
@@ -38,16 +40,49 @@ - (instancetype)init
...
+}
+
 #pragma mark - Core visible node range managment API
 
++ (ASLayoutRangeMode)rangeModeForInterfaceState:(ASInterfaceState)interfaceState
+                                scrollDirection:(ASScrollDirection)scrollDirection
+                               currentRangeMode:(ASLayoutRangeMode)currentRangeMode
+{
+  // If we used full mode, don't switch to minimum mode. That will destroy all the hard work done before.
+  if (currentRangeMode == ASLayoutRangeModeFull) {
```
appleguy: At some point, it is critical that we do release the very expensive backing stores created for the display range in particular. The fetch data range probably make sense to keep her around for a while longer, although the minimum fetch data range is still pretty generous in size. An example here is switching from feed to library. It probably makes sense to release most of the Display range in such a situation. Similarly, if a user scrolls through a pin and then swipes horizontally to another one, we should reclaim most of the display range from the previous one. I think the same is true for zooming into content, Like opening a related pin. I think the uniting theme is that if we are no longer visible, We should fall back to the minimum range?

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
```diff
@@ -106,21 +141,25 @@ - (void)_updateVisibleNodeIndexPaths
...
+                                                        rangeMode:rangeMode
+                                                        rangeType:ASLayoutRangeTypeFetchData];
 
-  ASRangeTuningParameters parametersDisplay = [_layoutController tuningParametersForRangeType:ASLayoutRangeTypeDisplay isFullRange:shouldUseFullRange];
-  ASRangeTuningParameters parametersFetchData = [_layoutController tuningParametersForRangeType:ASLayoutRangeTypeFetchData isFullRange:shouldUseFullRange];
+  ASRangeTuningParameters parametersDisplay = [_layoutController tuningParametersForRangeMode:rangeMode
+                                                                                    rangeType:ASLayoutRangeTypeDisplay];
+  ASRangeTuningParameters parametersFetchData = [_layoutController tuningParametersForRangeMode:rangeMode
+                                                                                      rangeType:ASLayoutRangeTypeFetchData];
   if (parametersDisplay.leadingBufferScreenfuls == parametersFetchData.leadingBufferScreenfuls &&
```
appleguy: I wonder if there is an easy way to generalize this? We should check if either range is equal to the visible range, As a developer could very reasonably choose to set their minimum display and a fetch data ranges to zero especially if they're Collection view layout is expensive and they need to minimize calls to get the index paths that lie within a given range

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
```diff
@@ -225,6 +266,33 @@ - (void)_updateVisibleNodeIndexPaths
 #endif
 }
 
+#pragma mark - Notification observers
+
+- (void)registerForNotificationsIfNeeded
+{
+  if (!_didRegisterForNotifications) {
+    BOOL selfInterfaceState = [_dataSource interfaceStateForRangeController:self];
```
appleguy: Perhaps we should pass this in as an argument. Not a big deal, but it would eliminate several method calls.

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
```diff
@@ -225,6 +266,33 @@ - (void)_updateVisibleNodeIndexPaths
...
+                                                   name:ASRenderingEngineDidDisplayScheduledNodesNotification
+                                                 object:nil];
+      _didRegisterForNotifications = YES;
+    }
+  }
+}
+
+- (void)scheduledNodesDidDisplay
+{
+  [[NSNotificationCenter defaultCenter] removeObserver:self];
```
appleguy: Since this is unreleased and just in master, I am not worried at the moment, but have you audited this for any possible race conditions? Is it possible that another range controller could be operating at the same time, Such as two scroll views next to one another on iPad, or that a prior display operation could trigger this method to be called and cause us to and observe right after we complete an editing transaction that kicks off more node loading?

PR #1171, opened 2016-02-03
[ASCollectionView / ASTableView] Optimize reloadData and reloadSection: methods.

lkzhao:
Calls reloadData & reloadSection on UITableView/UICollectionView instead of deleting/inserting one by one.  
Similar to #1093.

Please review this but do not merge yet. still seeing that bug where collectionView will cause inconsistent crash when inserting rows unless [collectionView numberOfSections] is called before inserting.
@nguyenhuy 
@appleguy


appleguy:
@lkzhao this is great, very glad to see this! Are there any areas in particular that I should focus on for review?  @nguyenhuy 's original disk to attempt something similar was reverted a couple times, but I can't remember if we ever really landed it after the second time… this this is a attempting to do something similar with a new approach, correct? Just want to make sure I understand where the prior one left off, it will not re-land again?

lkzhao:
@appleguy Yes, both are optimizing reload calls. This patch includes the other one that Huy made and fixed the crashes we had earlier. I actually took some code directly from his fork. We can land this instead. 


Few notes about what I did here:

1. Removed `_deleteSection`, `_insertSection` etc. These don't do much and made it very confusing to read the code.

2. **(Please review this part)** Every node editing methods previously all have a completion callback scheduled on the main thread and update the _completedNodes on the main thread. This made it hard and cause crashes when doing batch operations like ReloadSection/ReloadData etc. 
Now these methods only change _editingNodes and do nothing with the _completedNodes. After we are done with the operations, we call `commitChangesToNodesOfKind ` and provide a callback there. 

3. Optimized reloadSection/reloadData/reloadItems/deleteSection/insertSections on TableView and CollectionView. Previously, all of these actions result in deleteItems/insertItems calls to TableView and CollectionView. Now it makes one specific call. (for example, reloadSection calls tableview `reloadSections: withRowAnimation` instead of making `deleteRowsAtIndexPaths: and insertRowsAtIndexPaths:`)

appleguy:
Thank you Luke, this summary was extremely helpful for me.  I will review in detail soon.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -962,6 +1002,23 @@ - (void)rangeController:(ASRangeController *)rangeController didDeleteSectionsAt
...
 }
 
+- (void)rangeControllerDidReloadData:(ASRangeController *)rangeController{
+  ASDisplayNodeAssertMainThread();
+  if (!self.asyncDataSource || _superIsPendingDataLoad) {
+    return; // if the asyncDataSource has become invalid while we are processing, ignore this request to avoid crashes
+  }
+
+  if (_performingBatchUpdates) {
+    [_batchUpdateBlocks addObject:^{
```
nguyenhuy: Do we need to notify the `_layoutFacilitator` here? cc @binl

AsyncDisplayKit/ASTableView.mm
```diff
@@ -303,7 +303,7 @@ - (void)reloadDataWithCompletion:(void (^)())completion
   ASPerformBlockOnMainThread(^{
     [super reloadData];
```
nguyenhuy: Not sure if possible, but I really want to eliminate this call. We will do this afterward anyway.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -314,7 +314,7 @@ - (void)reloadData
 - (void)reloadDataImmediately
 {
   ASDisplayNodeAssertMainThread();
-  [_dataController reloadDataImmediatelyWithAnimationOptions:UITableViewRowAnimationNone];
+  [_dataController reloadDataImmediately];
   [super reloadData];
```
nguyenhuy: Same as above

AsyncDisplayKit/ASTableView.mm
```diff
@@ -831,6 +831,25 @@ - (void)rangeController:(ASRangeController *)rangeController didDeleteNodes:(NSA
...
+  if (!self.asyncDataSource) {
+    return; // if the asyncDataSource has become invalid while we are processing, ignore this request to avoid crashes
+  }
+
+  BOOL preventAnimation = animationOptions == UITableViewRowAnimationNone;
+  ASPerformBlockWithoutAnimation(preventAnimation, ^{
+    [super reloadRowsAtIndexPaths:indexPaths withRowAnimation:(UITableViewRowAnimation)animationOptions];
+  });
+
+  if (_automaticallyAdjustsContentOffset) {
```
nguyenhuy: Not sure what is the policy to adjust content offset? It is done in after reloading nodes, but not sections or everything/data.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -58,11 +58,11 @@ - (void)willReloadData
   [_pendingNodes enumerateKeysAndObjectsUsingBlock:^(NSString *kind, NSMutableArray *nodes, BOOL *stop) {
     // Remove everything that existed before the reload, now that we're ready to insert replacements
     NSArray *indexPaths = [self indexPathsForEditingNodesOfKind:kind];
-    [self deleteNodesOfKind:kind atIndexPaths:indexPaths completion:nil];
+    [self deleteNodesOfKind:kind atIndexPaths:indexPaths];
     
     NSArray *editingNodes = [self editingNodesOfKind:kind];
-    NSMutableIndexSet *indexSet = [[NSMutableIndexSet alloc] initWithIndexesInRange:NSMakeRange(0, editingNodes.count)];
-    [self deleteSectionsOfKind:kind atIndexSet:indexSet completion:nil];
+    NSMutableIndexSet *indexSet = [[NSMutableIndexSet alloc]initWithIndexesInRange:NSMakeRange(0, editingNodes.count)];
```
nguyenhuy: nit: white space before `initWithIndexesInRange`

AsyncDisplayKit/Details/ASDataController.h
```diff
@@ -85,15 +85,30 @@ FOUNDATION_EXPORT NSString * const ASDataControllerRowNodeKind;
...
 
 /**
  Called for deletion of sections.
  */
 - (void)dataController:(ASDataController *)dataController didDeleteSectionsAtIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions;
 
+/**
+ Called for reload of sections.
+ */
+- (void)dataController:(ASDataController *)dataController didReloadSectionsAtIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions;
```
nguyenhuy: Maybe just `didReloadSections:(NSIndexSet *)sections`?

AsyncDisplayKit/Details/ASDataController.h
```diff
@@ -85,15 +85,30 @@ FOUNDATION_EXPORT NSString * const ASDataControllerRowNodeKind;
...
 /**
+ Called for reload of elements.
+ */
+- (void)dataController:(ASDataController *)dataController didReloadNodes:(NSArray<ASCellNode *> *)nodes atIndexPaths:(NSArray<NSIndexPath *> *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions;
+
+/**
  Called for insertion of sections.
  */
-- (void)dataController:(ASDataController *)dataController didInsertSections:(NSArray<NSArray<ASCellNode *> *> *)sections atIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions;
+- (void)dataController:(ASDataController *)dataController didInsertSectionsAtIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions;
```
nguyenhuy: Same as below, `didInsertSections:(NSIndexSet)sections`?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -118,7 +127,7 @@ - (void)batchLayoutNodes:(NSArray *)nodes ofKind:(NSString *)kind atIndexPaths:(
     NSArray *batchedIndexPaths = [indexPaths subarrayWithRange:batchedRange];
     NSArray *batchedNodes = [nodes subarrayWithRange:batchedRange];
     
-    [self _layoutNodes:batchedNodes ofKind:kind atIndexPaths:batchedIndexPaths completion:completionBlock];
+    [self _layoutNodes:batchedNodes ofKind:kind atIndexPaths:batchedIndexPaths ];
```
nguyenhuy: nit: no whitespace after `atIndexPaths:batchedIndexPaths`

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
```diff
@@ -275,15 +283,22 @@ - (void)dataController:(ASDataController *)dataController didDeleteNodes:(NSArra
   });
 }
 
-- (void)dataController:(ASDataController *)dataController didInsertSections:(NSArray *)sections atIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
+- (void)dataController:(ASDataController *)dataController didInsertSectionsAtIndexSet:(NSIndexSet *)indexSet withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
 {
-  ASDisplayNodeAssert(sections.count == indexSet.count, @"Invalid sections");
```
nguyenhuy: Shouldn't we keep this assert?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -391,30 +317,26 @@ - (void)_reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animat
...
-          [sections addObject:[[NSMutableArray alloc] init]];
-        }
-        
-        [self _insertSections:sections atIndexSet:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, sectionCount)] withAnimationOptions:animationOptions];
-        
-        [self _batchLayoutNodes:updatedNodes atIndexPaths:updatedIndexPaths withAnimationOptions:animationOptions];
-        
-        if (completion) {
-          dispatch_async(dispatch_get_main_queue(), completion);
+
```
nguyenhuy: Should probably call `[self willReloadData];` here.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -391,30 +317,26 @@ - (void)_reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animat
...
+        for (NSUInteger i = 0; i < sectionCount; i++) {
+          [editingNodes addObject:[NSMutableArray array]];
         }
+
+        ASInsertElementsIntoMultidimensionalArrayAtIndexPaths(editingNodes, updatedIndexPaths, updatedNodes);
+        _editingNodes[ASDataControllerRowNodeKind] = editingNodes;
+
+        [self batchLayoutNodes:updatedNodes ofKind:ASDataControllerRowNodeKind atIndexPaths:updatedIndexPaths];
+
+        [self commitChangesToNodesOfKind:ASDataControllerRowNodeKind withCompletion:^{
```
nguyenhuy: Small optimization here: can pass `NIL` here if `_delegateDidReloadData` is `NO` and `completion` is `nil`. That way we don't schedule a block on main thread that essentially does nothing.

nguyenhuy:
I have 2 main concerns/feedbacks about this PR:

1. We should clearly document a few things for *each* of the methods in `ASDataController+Subclasses.h`:
  - It should (and is assuming to) be called on `_editingTransactionQueue`.
  - It blocks the calling (background) thread.
  - It only modifies `_editingNodes`.
  - Its caller is expected to commit the changes to `_completedNodes` on main afterward.

2. I think adding `commitChangesToNodesOfKind` is a really good idea that helps to simplify the controller. However, deep copying a 2D array is expensive and should be avoided. That's why in the current implementation, we don't deep copy after deletes, but instead directly remove items. Can we do the same thing here? 

Other than that, I reckon this new implementation should be faster than the one on master because we update `_completedNodes` and notify delegate once after all operations finished :+1:

AsyncDisplayKit/ASTableView.mm
appleguy: @nguyenhuy @lkzhao I think we should honor it wherever it can be reasonably accommodated, especially inserting and deleting elements, as well as moving or reloading individual items and sections as that may cause them to calculate a different height. I think the main challenge with doing it on a full table reload at the moment is that we break it into a delete and insert, At least historically, and because of the asynchronous delay of the insert it was not always feasible to guarantee.

nguyenhuy: Now it is feasible thanks to this awesome diff. Let's do it then!

lkzhao:
@nguyenhuy Thanks for the review & feedbacks!! There are some code that I copied/converted and I am not entirely sure what they do. Let me talk to you tomorrow about some of the details int your review.

AsyncDisplayKit/Details/ASDataController.h
lkzhao: Tableview reloadSection actually supports animation. I think we should keep the animationOption here.

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
lkzhao: I eleminated the section parameter on this call. therefore no need to check length equality anymore.

AsyncDisplayKit/Details/ASDataController.mm
lkzhao: Nice catch

AsyncDisplayKit/Details/ASRangeControllerBeta.mm
nguyenhuy: oh, right. I missed that :)

AsyncDisplayKit/Details/ASDataController.h
nguyenhuy: I meant the naming of that particular sections param.

lkzhao: The ASRangeController and ASDataController all uses `didXXXXSectionsAtIndexSet` pattern. I think we should keep it consistent.

for example:
```
- (void)rangeController:(ASRangeController *)rangeController didDeleteSectionsAtIndexSet:(NSIndexSet *)indexSet withAnimationOptions:
```

AsyncDisplayKit/ASCollectionView.mm
binl: yes, we will need to do
```
[_layoutFacilitator collectionViewWillEditCellsAtIndexPaths:@[[NSIndexPath indexPathForItem:0 section:0]] batched:YES/NO];
```

appleguy:
Very excited to land this!  Thank you @lkzhao

PR #1173, opened 2016-02-03
Move ASCellNode allocation off the main thread

rahul-malik:
Introducing a block-based API for ASCellNode creation in ASDataController. Move allocations and loaded node layouts to occur during batch layout phase. Existing nodeForItemAtIndex path methods will now have their value's wrapped in a block that will be resolved later in ASDataController during layout.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -347,8 +347,9 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @param indexPath The index path of the requested node.
  *
- * @returns a node for display at this indexpath. This will be called on the main thread and should not implement reuse (it will be called once per row).  Unlike UICollectionView's version, this method
- * is not called when the row is about to display.
+ * @returns a node for display at this indexpath. This will be called on the main thread and should
+ *   not implement reuse (it will be called once per row).  Unlike UICollectionView's version,
+ *   this method is not called when the row is about to display.
  */
 - (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtIndexPath:(NSIndexPath *)indexPath;
 
@@ -361,10 +362,11 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @param indexPath The index path of the requested node.
  *
- * @returns a block that creates the node for display at this indexpath.  Must be thread-safe (can be called on the main thread or a background
- *          queue) and should not implement reuse (it will be called once per row).
+ * @returns a block that creates the node for display at this indexpath.  
+ *   Must be thread-safe (can be called on the main thread or a background
+ *   queue) and should not implement reuse (it will be called once per row).
  */
-- (ASDataControllerCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockAtIndexPath:(NSIndexPath *)indexPath;
+- (ASCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockForItemAtIndexPath:(NSIndexPath *)indexPath;
 
 /**
  * Asks the collection view to provide a supplementary node to display in the collection view.
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -347,13 +347,24 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @param indexPath The index path of the requested node.
  *
- * @returns a node for display at this indexpath.  Must be thread-safe (can be called on the main thread or a background
- * queue) and should not implement reuse (it will be called once per row).  Unlike UICollectionView's version, this method
+ * @returns a node for display at this indexpath. This will be called on the main thread and should not implement reuse (it will be called once per row).  Unlike UICollectionView's version, this method
  * is not called when the row is about to display.
  */
 - (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtIndexPath:(NSIndexPath *)indexPath;
 
+
 @optional
+
+/**
+ *
+ * @param collectionView The sender.
+ *
+ * @param indexPath The index path of the requested node.
+ *
+ * @returns a block that creates the node for display at this indexpath.  Must be thread-safe (can be called on the main thread or a background
+ *          queue) and should not implement reuse (it will be called once per row).
+ */
+- (ASDataControllerCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockAtIndexPath:(NSIndexPath *)indexPath;
 
 /**
  * Asks the collection view to provide a supplementary node to display in the collection view.
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -77,6 +77,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   BOOL _asyncDelegateImplementsInsetSection;
   BOOL _collectionViewLayoutImplementsInsetSection;
   BOOL _asyncDataSourceImplementsConstrainedSizeForNode;
+  BOOL _asyncDataSourceImplementsNodeBlockForItemAtIndexPath;
   BOOL _queuedNodeSizeUpdate;
   BOOL _isDeallocating;
   
@@ -301,10 +302,12 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     _asyncDataSource = nil;
     _proxyDataSource = _isDeallocating ? nil : [[ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
     _asyncDataSourceImplementsConstrainedSizeForNode = NO;
+    _asyncDataSourceImplementsNodeBlockForItemAtIndexPath = NO;
   } else {
     _asyncDataSource = asyncDataSource;
     _proxyDataSource = [[ASCollectionViewProxy alloc] initWithTarget:_asyncDataSource interceptor:self];
-    _asyncDataSourceImplementsConstrainedSizeForNode = ([_asyncDataSource respondsToSelector:@selector(collectionView:constrainedSizeForNodeAtIndexPath:)] ? 1 : 0);
+    _asyncDataSourceImplementsConstrainedSizeForNode = [_asyncDataSource respondsToSelector:@selector(collectionView:constrainedSizeForNodeAtIndexPath:)];
+    _asyncDataSourceImplementsNodeBlockForItemAtIndexPath = [_asyncDataSource respondsToSelector:@selector(collectionView:nodeBlockForItemAtIndexPath:)];
   }
   
   super.dataSource = (id<UICollectionViewDataSource>)_proxyDataSource;
@@ -691,27 +694,32 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 }
 
 
-- (ASDataControllerCellNodeBlock)dataController:(ASDataController *)dataController nodeBlockAtIndexPath:(NSIndexPath *)indexPath
+- (ASCellNodeBlock)dataController:(ASDataController *)dataController nodeBlockAtIndexPath:(NSIndexPath *)indexPath
 {
-  if (![_asyncDataSource respondsToSelector:@selector(collectionView:nodeBlockAtIndexPath:)]) {
+  if (!_asyncDataSourceImplementsNodeBlockForItemAtIndexPath) {
     ASCellNode *node = [_asyncDataSource collectionView:self nodeForItemAtIndexPath:indexPath];
+    ASDisplayNodeAssert([node isKindOfClass:ASCellNode.class], @"invalid node class, expected ASCellNode");
+    __weak __typeof__(self) weakSelf = self;
     return ^{
+      __typeof__(self) strongSelf = weakSelf;
       [node enterHierarchyState:ASHierarchyStateRangeManaged];
-      ASDisplayNodeAssert([node isKindOfClass:ASCellNode.class], @"invalid node class, expected ASCellNode");
       if (node.layoutDelegate == nil) {
-        node.layoutDelegate = self;
+        node.layoutDelegate = strongSelf;
       }
       return node;
     };
   }
 
-  ASDataControllerCellNodeBlock block = [_asyncDataSource collectionView:self nodeBlockAtIndexPath:indexPath];
-  ASDisplayNodeAssertNotNil(block, @"Invalid block, expected nonnull ASDataControllerCellNodeBlock");
+  ASCellNodeBlock block = [_asyncDataSource collectionView:self nodeBlockForItemAtIndexPath:indexPath];
+  ASDisplayNodeAssertNotNil(block, @"Invalid block, expected nonnull ASCellNodeBlock");
+  __weak __typeof__(self) weakSelf = self;
   return ^{
+    __typeof__(self) strongSelf = weakSelf;
+
     ASCellNode *node = block();
     [node enterHierarchyState:ASHierarchyStateRangeManaged];
     if (node.layoutDelegate == nil) {
-      node.layoutDelegate = self;
+      node.layoutDelegate = strongSelf;
     }
     return node;
   };
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -690,6 +690,33 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   return node;
 }
 
+
+- (ASDataControllerCellNodeBlock)dataController:(ASDataController *)dataController nodeBlockAtIndexPath:(NSIndexPath *)indexPath
+{
+  if (![_asyncDataSource respondsToSelector:@selector(collectionView:nodeBlockAtIndexPath:)]) {
+    ASCellNode *node = [_asyncDataSource collectionView:self nodeForItemAtIndexPath:indexPath];
+    return ^{
+      [node enterHierarchyState:ASHierarchyStateRangeManaged];
+      ASDisplayNodeAssert([node isKindOfClass:ASCellNode.class], @"invalid node class, expected ASCellNode");
+      if (node.layoutDelegate == nil) {
+        node.layoutDelegate = self;
+      }
+      return node;
+    };
+  }
+
+  ASDataControllerCellNodeBlock block = [_asyncDataSource collectionView:self nodeBlockAtIndexPath:indexPath];
+  ASDisplayNodeAssertNotNil(block, @"Invalid block, expected nonnull ASDataControllerCellNodeBlock");
+  return ^{
+    ASCellNode *node = block();
+    [node enterHierarchyState:ASHierarchyStateRangeManaged];
+    if (node.layoutDelegate == nil) {
+      node.layoutDelegate = self;
+    }
+    return node;
+  };
+}
+
 - (ASSizeRange)dataController:(ASDataController *)dataController constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
 {
   ASSizeRange constrainedSize = kInvalidSizeRange;
--- a/AsyncDisplayKit/ASPagerNode.m
+++ b/AsyncDisplayKit/ASPagerNode.m
@@ -16,6 +16,7 @@
   UICollectionViewFlowLayout *_flowLayout;
   ASPagerNodeProxy *_proxy;
   __weak id <ASPagerNodeDataSource> _pagerDataSource;
+  BOOL _pagerDataSourceImplementsNodeBlockAtIndex;
 }
 
 @end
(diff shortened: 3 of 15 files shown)


AsyncDisplayKit/ASCollectionView.h
```diff
@@ -319,15 +319,26 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @param indexPath The index path of the requested node.
  *
- * @returns a node for display at this indexpath.  Must be thread-safe (can be called on the main thread or a background
- * queue) and should not implement reuse (it will be called once per row).  Unlike UICollectionView's version, this method
+ * @returns a node for display at this indexpath. This will be called on the main thread and should not implement reuse (it will be called once per row).  Unlike UICollectionView's version, this method
```
appleguy: This should still be on three lines - it is causing horizontal scrolling on Github :)  Though we don't have a fixed max line length, I think this is longer than some of the next-longest in this file.

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -319,15 +319,26 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
+ *
+ * @param collectionView The sender.
+ *
+ * @param indexPath The index path of the requested node.
+ *
+ * @returns a block that creates the node for display at this indexpath.  Must be thread-safe (can be called on the main thread or a background
+ *          queue) and should not implement reuse (it will be called once per row).
+ */
+- (ASDataControllerCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockAtIndexPath:(NSIndexPath *)indexPath;
```
appleguy: How can we best handle the fact that nodeForItemAtIndexPath: should not be required if nodeBlockAtIndexPath is implemented?

Probably should just go to runtime assertions.

How intentional was it to leave out the word "item" in the name?  I'm supportive of this to unify the collection / table APIs, but bummed that it doesn't match the non-block method.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -691,6 +691,33 @@ - (ASCellNode *)dataController:(ASDataController *)dataController nodeAtIndexPat
...
 
+
+- (ASDataControllerCellNodeBlock)dataController:(ASDataController *)dataController nodeBlockAtIndexPath:(NSIndexPath *)indexPath
+{
+  if (![_asyncDataSource respondsToSelector:@selector(collectionView:nodeBlockAtIndexPath:)]) {
+    ASCellNode *node = [_asyncDataSource collectionView:self nodeForItemAtIndexPath:indexPath];
+    return ^{
+      [node enterHierarchyState:ASHierarchyStateRangeManaged];
+      ASDisplayNodeAssert([node isKindOfClass:ASCellNode.class], @"invalid node class, expected ASCellNode");
+      if (node.layoutDelegate == nil) {
```
appleguy: I assume this check is pre-existing.  I wonder why we would need to check if it is nil rather than unconditionally setting it...maybe Huy knows :)

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -84,6 +84,17 @@ - (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtI
   return pageNode;
 }
 
+- (ASDataControllerCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockAtIndexPath:(NSIndexPath *)indexPath {
```
appleguy: { on new line.

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -84,6 +84,17 @@ - (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtI
...
+  ASDisplayNodeAssert(_pagerDataSource != nil, @"ASPagerNode must have a data source to load nodes to display");
+  if (![_pagerDataSource respondsToSelector:@selector(pagerNode:nodeBlockAtIndex:)]) {
+    ASCellNode *node = [_pagerDataSource pagerNode:self nodeAtIndex:indexPath.item];
+    return ^{ return node; };
+  }
+  ASDataControllerCellNodeBlock block = [_pagerDataSource pagerNode:self nodeBlockAtIndex:indexPath.item];
+  ASDisplayNodeAssertNotNil(block, @"Invalid node block. Block should be non-nil.");
```
appleguy: Presumably this is asserted at the level of the caller of this method, so let's eliminate two lines in just directly return the data source call.

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -84,6 +84,17 @@ - (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtI
   return pageNode;
 }
 
+- (ASDataControllerCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockAtIndexPath:(NSIndexPath *)indexPath {
+  ASDisplayNodeAssert(_pagerDataSource != nil, @"ASPagerNode must have a data source to load nodes to display");
+  if (![_pagerDataSource respondsToSelector:@selector(pagerNode:nodeBlockAtIndex:)]) {
```
appleguy: Although it feels like premature optimization at this stage, this method will be called hundreds or even thousands of times in a single run on the main thread for some applications, So we need to add a BOOL to cash weather the data source responds to the block selector.  Less problematic for the pager in some cases, but may as well do it for here, table, and collection.

Absolutely fine to do this in a follow-up diff, or even file a task on me to distribute it to the community member

AsyncDisplayKit/ASTableView.h
```diff
@@ -294,8 +294,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @param indexPath The index path of the requested node.
  *
- * @returns a node for display at this indexpath.  Must be thread-safe (can be called on the main thread or a background
- * queue) and should not implement reuse (it will be called once per row).  Unlike UITableView's version, this method
+ * @returns a node for display at this indexpath. This will be called on the main thread and should not implement reuse (it will be called once per row). Unlike UITableView's version, this method
```
appleguy: Keep two lines, probably

AsyncDisplayKit/ASTableView.h
```diff
@@ -303,6 +302,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * @param tableView The sender.
+ *
+ * @param indexPath The index path of the requested node.
+ *
+ * @returns a block that creates the node for display at this indexpath.  Must be thread-safe (can be called on the main thread or a background
+ * queue) and should not implement reuse (it will be called once per row).
+ */
+
+- (ASDataControllerCellNodeBlock)tableView:(ASTableView *)tableView nodeBlockForRowAtIndexPath:(NSIndexPath *)indexPath;
```
appleguy: Can we rename the block type to be a much simpler "ASCellNodeBlock", or do you have any thoughts on why it should be more specific? Sometimes the typeahead requires specifying the correct return type, and because data controller is internal implementation detail anyway, we should probably not introduced the terminology in the more public facing API

AsyncDisplayKit/ASTableView.mm
```diff
@@ -876,6 +876,33 @@ - (ASCellNode *)dataController:(ASDataController *)dataController nodeAtIndexPat
...
+      if (node.layoutDelegate == nil) {
+        node.layoutDelegate = self;
+      }
+      return node;
+    };
+  }
+
+  ASDataControllerCellNodeBlock block = [_asyncDataSource tableView:self nodeBlockForRowAtIndexPath:indexPath];
+  __weak __typeof__(self) weakSelf = self;
+  ASDataControllerCellNodeBlock configuredNodeBlock = ^{
```
appleguy: This method is not completely symmetrical with the Collection view version. Unless it is important they're different, make them the same, So that a future re-factor that enables codesharing does not spend time considering the significance of the differences

AsyncDisplayKit/Details/ASCollectionDataController.h
```diff
@@ -30,6 +30,10 @@
 
 - (NSUInteger)dataController:(ASCollectionDataController *)dataController supplementaryNodesOfKind:(NSString *)kind inSection:(NSUInteger)section;
 
+@optional
+
+- (ASDataControllerCellNodeBlock)dataController:(ASCollectionDataController *)dataController supplementaryNodeBlockOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath;
```
appleguy: Wow, you really went all the way here! I had implicitly assumed we would leave supplementary node creation synchronous, but this is great to have. Definitely a motivator to do that refactor to combine in the whale code for supplementary and regular cell nodes, while also adding support for headers and footers… upshot is, if blocks for supplementary nodes in particular are adding complexity, feel free to nuke them from this current effort.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -176,7 +176,14 @@ - (void)_populateSupplementaryNodesOfKind:(NSString *)kind withMutableNodes:(NSM
     for (NSUInteger j = 0; j < rowCount; j++) {
       NSIndexPath *indexPath = [sectionIndexPath indexPathByAddingIndex:j];
       [indexPaths addObject:indexPath];
-      [nodes addObject:[self.collectionDataSource dataController:self supplementaryNodeOfKind:kind atIndexPath:indexPath]];
+      ASDataControllerCellNodeBlock supplementaryCellBlock;
+      if ([self.collectionDataSource respondsToSelector:@selector(dataController:supplementaryNodeBlockOfKind:atIndexPath:)]) {
```
appleguy: Same deal with caching responds to selector

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -94,6 +97,16 @@ - (void)setDelegate:(id<ASDataControllerDelegate>)delegate
   _delegateDidDeleteSections  = [_delegate respondsToSelector:@selector(dataController:didDeleteSectionsAtIndexSet:withAnimationOptions:)];
 }
 
+- (void)setDataSource:(id<ASDataControllerSource>)dataSource {
+  if (_dataSource == dataSource) {
+    return;
+  }
+
+  _dataSource = dataSource;
+  // This probably won't be sufficient to tell if we should call the node block
```
appleguy: Why would it not be sufficient? It seems to me that if we implement the node block version, we would not ever care to call the other version

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -117,12 +130,11 @@ - (void)batchLayoutNodes:(NSArray *)nodes ofKind:(NSString *)kind atIndexPaths:(
     NSRange batchedRange = NSMakeRange(i, MIN(indexPaths.count - i, blockSize));
     NSArray *batchedIndexPaths = [indexPaths subarrayWithRange:batchedRange];
     NSArray *batchedNodes = [nodes subarrayWithRange:batchedRange];
-    
     [self _layoutNodes:batchedNodes ofKind:kind atIndexPaths:batchedIndexPaths completion:completionBlock];
   }
 }
 
-- (void)layoutLoadedNodes:(NSArray *)nodes ofKind:(NSString *)kind atIndexPaths:(NSArray *)indexPaths {
+- (void)layoutLoadedNodes:(NSArray<ASCellNode *> *)nodes ofKind:(NSString *)kind atIndexPaths:(NSArray<NSIndexPath *> *)indexPaths {
```
appleguy: Thank you for the type annotations, I do appreciate that

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -159,22 +171,50 @@ - (void)_layoutNodes:(NSArray *)nodes ofKind:(NSString *)kind atIndexPaths:(NSAr
...
     return;
   }
-  
+
+  NSUInteger nodeCount = nodes.count;
+  NSMutableArray<ASCellNode *> *allocatedNodes = [NSMutableArray arrayWithCapacity:nodeCount];
   dispatch_group_t layoutGroup = dispatch_group_create();
-  ASSizeRange *nodeBoundSizes = (ASSizeRange *)malloc(sizeof(ASSizeRange) * nodes.count);
+  ASSizeRange *nodeBoundSizes = (ASSizeRange *)malloc(sizeof(ASSizeRange) * nodeCount);
+  BOOL isMainThread = [NSThread isMainThread];
```
appleguy: I need to go around the framework to enforce adoption, as it is inconsistent at the moment, but probably worth using ASDisplayNodeThreadIsMain() here.  We may want to rename that too, maybe ASOnMainThread() or something.

AsyncDisplayKitTests/ASTableViewTests.m
```diff
@@ -72,6 +72,11 @@ - (ASCellNode *)tableView:(ASTableView *)tableView nodeForRowAtIndexPath:(NSInde
   return nil;
 }
 
+- (ASDataControllerCellNodeBlock)tableView:(ASTableView *)tableView nodeBlockForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+  return nil;
+}
```
appleguy: Simple, but beautiful.  Thanks for starting test-first, and even these simple ones achieve two important goals - 1) testing common cases, of actual relevance ; and 2) provide an environment / hint to others that we can expand testing here, and there's a spot just for you if you're adding node block tests :)

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -409,9 +451,22 @@ - (void)_reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animat
         }
         
         [self _insertSections:sections atIndexSet:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, sectionCount)] withAnimationOptions:animationOptions];
-        
+
         [self _batchLayoutNodes:updatedNodes atIndexPaths:updatedIndexPaths withAnimationOptions:animationOptions];
-        
+
+//        if (_dataSourceImplementsAsyncNodeAtIndexPath) {
```
appleguy: Send me an iMessage when you have another version up for me to review. I would love to help get this in tomorrow, my review is complete and everything looks good to me so far, but I am going to hold off until these comments are corrected and it is more clear what the intended code path is in this critical area. @rahul-malik

AsyncDisplayKit/ASCollectionView.h
rahul-malik: Not intentional at all, updating this now

AsyncDisplayKit/ASCollectionView.h
rahul-malik: thanks, fixing

AsyncDisplayKit/ASCollectionView.mm
rahul-malik: @nguyenhuy - Any thoughts here?

AsyncDisplayKit/ASPagerNode.m
rahul-malik: I actually don't believe this is premature since this value doesn't change until another dataSource is assigned. It would be nice to establish a pattern to abstraction for caching these calls

AsyncDisplayKit/ASPagerNode.m
rahul-malik: Fixed!

AsyncDisplayKit/ASTableView.h
rahul-malik: Yeah I think that would be simpler, changing now.

AsyncDisplayKit/ASCollectionView.mm
rahul-malik: @nguyenhuy just confirmed that this was to avoid overriding the layoutDelegate if the node creator wants to handle it.

AsyncDisplayKitTests/ASTableViewTests.m
rahul-malik: Will add tests in a follow-up PR. Manually ran through use-cases to ensure that there are no obvious issues but will have to enforce this with tests in a follow up PR

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -301,10 +302,13 @@ - (void)setAsyncDataSource:(id<ASCollectionViewDataSource>)asyncDataSource
     _asyncDataSource = nil;
     _proxyDataSource = _isDeallocating ? nil : [[ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
     _asyncDataSourceImplementsConstrainedSizeForNode = NO;
+    _asyncDataSourceImplementsNodeBlockForItemAtIndexPath = NO;
   } else {
     _asyncDataSource = asyncDataSource;
     _proxyDataSource = [[ASCollectionViewProxy alloc] initWithTarget:_asyncDataSource interceptor:self];
     _asyncDataSourceImplementsConstrainedSizeForNode = ([_asyncDataSource respondsToSelector:@selector(collectionView:constrainedSizeForNodeAtIndexPath:)] ? 1 : 0);
+
```
appleguy: Newline not needed, or add one before the two respondsToSelector calls.  Could also remove the odd / ridiculous ? 1 : 0 on the previous line ha

AsyncDisplayKit/ASTableView.mm
```diff
@@ -876,6 +876,35 @@ - (ASCellNode *)dataController:(ASDataController *)dataController nodeAtIndexPat
   return node;
 }
 
+- (ASCellNodeBlock)dataController:(ASDataController *)dataController nodeBlockAtIndexPath:(NSIndexPath *)indexPath {
+  if (![_asyncDataSource respondsToSelector:@selector(tableView:nodeBlockForRowAtIndexPath:)]) {
+    ASCellNode *node = [_asyncDataSource tableView:self nodeForRowAtIndexPath:indexPath];
+    ASDisplayNodeAssert([node isKindOfClass:ASCellNode.class], @"invalid node class, expected ASCellNode");
+    __weak __typeof__(self) weakSelf = self;
```
appleguy: I can't immediately think of how to deduplicate this from below, but we have 4 copies of largely-the-same code here...I wonder if we can share this as "set up cell node for use by range controller with layoutDelegate" in a different file, used by both table and collection?  Not needed for this diff, just a highly desirable to keep the file size down in the future.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -176,7 +178,14 @@ - (void)_populateSupplementaryNodesOfKind:(NSString *)kind withMutableNodes:(NSM
     for (NSUInteger j = 0; j < rowCount; j++) {
       NSIndexPath *indexPath = [sectionIndexPath indexPathByAddingIndex:j];
       [indexPaths addObject:indexPath];
-      [nodes addObject:[self.collectionDataSource dataController:self supplementaryNodeOfKind:kind atIndexPath:indexPath]];
+      ASCellNodeBlock supplementaryCellBlock;
+      if (_dataSourceImplementsSupplementaryNodeBlockOfKindAtIndexPath) {
+        supplementaryCellBlock = [self.collectionDataSource dataController:self supplementaryNodeBlockOfKind:kind atIndexPath:indexPath];
```
appleguy: Hmm, I wonder if we need to set the same hierarchy state and layout delegate thing here?  I bet we are not actually pre-displaying or maybe not even calling fetchData on supplementary nodes...this goes back to my concern about how they are integrated, and the need to support headers and footers in tables.  cc @levi, @rahul-malik

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -236,4 +251,9 @@ - (NSArray *)supplementaryKinds
   return (id<ASCollectionDataControllerSource>)self.dataSource;
 }
 
+- (void)setDataSource:(id<ASDataControllerSource>)dataSource {
```
appleguy: { on new line

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -159,22 +160,49 @@ - (void)_layoutNodes:(NSArray *)nodes ofKind:(NSString *)kind atIndexPaths:(NSAr
...
-    for (NSUInteger k = j; k < j + batchCount; k++) {
-      ASCellNode *node = nodes[k];
-      if (!node.isNodeLoaded) {
-        nodeBoundSizes[k] = [self constrainedSizeForNodeOfKind:kind atIndexPath:indexPaths[k]];
+
+
+    if (ASDisplayNodeThreadIsMain()) {
+      dispatch_semaphore_t sema = dispatch_semaphore_create(0);
+      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        for (NSUInteger k = j; k < j + batchCount; k++) {
```
appleguy: Should this code be defined in the block, provided directly to dispatch a sink here, while we just call it in line in the else condition below? I guess the dispatch semaphore signaling is the main difference, maybe it is not worth trying to figure out how to de-duplicate that but we could use a Boolean variable to only call it in the case that is needed, although it would be a bit weird for the sema variable to be NULL in that case it would probably be better than the duplication.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -577,9 +600,6 @@ - (void)insertSections:(NSIndexSet *)sections withAnimationOptions:(ASDataContro
       NSMutableArray *updatedNodes = [NSMutableArray array];
       NSMutableArray *updatedIndexPaths = [NSMutableArray array];
       [self _populateFromDataSourceWithSectionIndexSet:sections mutableNodes:updatedNodes mutableIndexPaths:updatedIndexPaths];
-      
-      // Measure nodes whose views are loaded before we leave the main thread
-      [self layoutLoadedNodes:updatedNodes ofKind:ASDataControllerRowNodeKind atIndexPaths:updatedIndexPaths];
```
appleguy: Glad you found a cleaner way to integrate this at the layout dispatch level.  This has always bothered me.  Fortunately, in a couple months the work on the new threading model should allow the distinction to be eliminated for loaded nodes.

AsyncDisplayKitTests/ASCollectionViewFlowLayoutInspectorTests.m
```diff
@@ -25,6 +25,10 @@ - (ASCellNode *)collectionView:(ASCollectionView *)collectionView nodeForItemAtI
   return [[ASCellNode alloc] init];
 }
 
+- (ASCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockAtIndexPath:(NSIndexPath *)indexPath {
```
appleguy: { on new line

PR #1178, opened 2016-02-04
More on range mode and range controller

nguyenhuy:
Follow up on #1163

One problem/concern left: Pages are not notified when they fall outside visible range of the hosting pager node. So they don't update their range mode and don't shrink back to minimum one.

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeControllerBeta.mm
+++ b/AsyncDisplayKit/Details/ASRangeControllerBeta.mm
@@ -27,6 +27,7 @@
   NSSet<NSIndexPath *> *_allPreviousIndexPaths;
   ASLayoutRangeMode _currentRangeMode;
   BOOL _didRegisterForNotifications;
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/Details/ASRangeControllerBeta.mm
```diff
@@ -268,16 +278,15 @@ - (void)_updateVisibleNodeIndexPaths
 
 #pragma mark - Notification observers
 
-- (void)registerForNotificationsIfNeeded
+- (void)registerForNotificationsIfNeededForInterfaceState:(ASInterfaceState)interfaceState
```
levi: `registerForNotificationsForInterfaceStateIfNeeded`

nguyenhuy: Renamed. Thanks!

PR #1183, opened 2016-02-05
[ASLayoutSpec] Add API to animate layout spec transitions, with implicit subnode management.

### LINKED ISSUE #803: [ASLayoutSpec] New Feature: Animate between specs, including Node addition & removal
Animating between ASLayoutSpecs should not be hard to implement.  What would be more difficult, and far more powerful to become a core mechanic used by most ASDK clients, would be supporting the ability for an ASLayoutSpec transition to add and remove nodes as well.

All of the necessary information is in the ASLayoutSpec.  If a node is present in the spec hierarchy that wasn't previously, it needs to be added (and by default, fade in).  If one was present in the old spec but no longer, it needs to be removed (and fade out).  Because ASLayoutSpecs are only designed to handle immediate children of the current node, this makes it clear where in the hierarchy the subnodes should be added.

Naturally an important consideration is to figure out how to execute the transition recursively, but without undue performance overhead.

Community thought leaders who might want to think about this and weigh in on API naming details or implementation approaches: @nguyenhuy, @rcancro, @garrettmoon, @eanagel, @levi

levi:
Resolves #803.


AsyncDisplayKit/ASContextTransitioning.h
```diff
@@ -0,0 +1,33 @@
...
+
+#import <AsyncDisplayKit/AsyncDisplayKit.h>
+
+@protocol ASContextTransitioning <NSObject>
+
+/**
+ @abstract The frame for the given node before the transition began.
+ @discussion Returns CGRectNull if the node was not in the hierarchy before the transition.
+ */
+- (CGRect)initialFrameForNode:(ASDisplayNode *)node;
```
appleguy: Frame might be worth including in the API as a convenience using many cases, but for this to be fully generalized, we would have to support setting balance and position along with transform since Frame is not defined with a non identity transform

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -571,6 +571,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 @end
 
+@interface ASDisplayNode (Transitioning)
+
+/**
+ @abstract Invalidates the current layout and begins a relayout of the node to the new layout returned in `calculateLayoutThatFits:`.
+
+ @discussion Animation is optional, but will still proceed through the `transitionLayout` methods with `isAnimated == NO`.
+ */
+- (void)transitionLayoutWithAnimation:(BOOL)animated;
```
appleguy: This looks great. How should we allow the client to control the duration? Should we use just a duration argument, and use 0 for non animated?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -988,6 +988,119 @@ - (CGRect)convertRect:(CGRect)rect toNode:(ASDisplayNode *)node
   return CGRectApplyAffineTransform(rect, flattenedTransform);
 }
 
+#pragma mark - Layout Transition
+
+- (void)transitionLayoutWithAnimation:(BOOL)animated
+{
+  [self transitionLayoutToFit:_constrainedSize animated:animated];
```
appleguy: It doesn't need to be supported in the first version, but one of the key use cases to support his transitioning between different constrained sizes. Let's consider how we can integrate this into the API, Even if it does not work initially.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -988,6 +988,119 @@ - (CGRect)convertRect:(CGRect)rect toNode:(ASDisplayNode *)node
...
+- (void)__transitionLayoutWithAnimation:(BOOL)animated
+{
+  _transitionContext = [[_ASTransitionContext alloc] initWithAnimation:animated delegate:self];
+
+  [self willTransitionLayout:_transitionContext];
+  if ([_insertedSubnodes count]) {
+    for (_ASDisplayNodePosition *position in _insertedSubnodes) {
+      [self _implicitlyInsertSubnode:position.node atIndex:position.index];
+    }
+    _insertedSubnodes = @[];
```
appleguy: It's a minimal distinction, but let's just use nil for this sort of thing unless there is a functional reason that the empty array is convenient

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -988,6 +988,119 @@ - (CGRect)convertRect:(CGRect)rect toNode:(ASDisplayNode *)node
...
+  [context completeTransition:YES];
+}
+
+- (void)didCompleteTransitionLayout:(id<ASContextTransitioning>)context
+{
+  if ([_deletedSubnodes count]) {
+    for (_ASDisplayNodePosition *position in _deletedSubnodes) {
+      [self _implicitlyRemoveSubnode:position.node atIndex:position.index];
+    }
+    _deletedSubnodes = @[];
```
appleguy: same re: nil

appleguy:
@levi I think one of your new files needs to be added to the dynamic framework target.

Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$__ASTransitionContext", referenced from:
      objc-class-ref in ASDisplayNode.o
ld: symbol(s) not found for architecture x86_64

AsyncDisplayKit/ASContextTransitioning.h
```diff
@@ -0,0 +1,36 @@
...
+@protocol ASContextTransitioning <NSObject>
+
+/**
+ @abstreact Defines if the given transition is animated
+ */
+- (BOOL)isAnimated;
+
+/**
+ @abstract The frame for the given node before the transition began.
+ @discussion Returns CGRectNull if the node was not in the hierarchy before the transition.
```
appleguy: Could you include in this file a description of why this is a protocol - should the client ever implement objects that conform to it?  Initially it seemed like they would, maybe a controller object managing the transition at a data level that provided frames for transitions near the cellNodeFor...: method.  But this documentation made me think the user would be calling it on a framework-provided object only, unless we are asking clients here to ensure they do return CGRectNull when appropriate?

AsyncDisplayKit/ASContextTransitioning.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+- (CGRect)initialFrameForNode:(ASDisplayNode *)node;
+
+/**
+ @abstract The frame for the given node when the transition completes.
+ @discussion Returns CGRectNull if the node is no longer in the hierarchy after the transition.
+ */
+- (CGRect)finalFrameForNode:(ASDisplayNode *)node;
```
appleguy: These selector names are excellent.  Very in-line with the framework and awesomely short/clear compared to UIKit selectors.  So is the language of the documentation file-wide, apart from the slight ambiguity regarding CGRectNull

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -155,6 +156,19 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)invalidateCalculatedLayout;
 
 
+/** @name Layout Transitioning */
+
+/**
+ @discussion A place to perform your animation. New nodes have been inserted here. You can also use this time to re-order the hierarchy.
```
appleguy: Great place to have developers encode any specific ordering requirements, if any are needed - this combined with being able to use addSubnode in init (particularly for nodes that don't need transitions) means that we should feel confident without additional handling of the hierarchy "z-order" management.

AsyncDisplayKit/ASDisplayNode.h
appleguy: What are your latest thoughts here?  Do we need one that provides a sizeRange for a new constraint?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -60,6 +60,11 @@ - (instancetype)initWithNode:(ASDisplayNode *)node atIndex:(NSUInteger)index
   return self;
 }
 
+- (NSString *)description
+{
+  return [NSString stringWithFormat:@"<%@ %p node = %@, index = %ld>", self.class, self, _node, (long)_index];
```
appleguy: Let's go ahead and add the %p of the view and layer.  Their descriptions shouldn't be included since at least the view also includes the node's full description (IIRC)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -650,15 +655,23 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
-        _insertedSubnodes = [self _filterNodesInLayouts:newLayout.sublayouts withIndexes:insertions];
-        _deletedSubnodes = [self _filterNodesInLayouts:_layout.sublayouts withIndexes:deletions];
+        _insertedSubnodes = [self _nodesInLayout:newLayout atIndexes:insertions];
+        _deletedSubnodes = [self _nodesInLayout:_layout atIndexes:deletions filterNodes:_insertedSubnodes];
       } else {
-        NSIndexSet *indexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, [newLayout.sublayouts count])];
-        _insertedSubnodes = [self _filterNodesInLayouts:newLayout.sublayouts withIndexes:indexes];
-        _deletedSubnodes = @[];
+        NSIndexSet *indexes = [NSIndexSet indexSetWithIndexesInRange:NSMakeRange(0, [newLayout.immediateSublayouts count])];
+        _insertedSubnodes = [self _nodesInLayout:newLayout atIndexes:indexes];
```
appleguy: Does this method _nodesInLayout only return nodes that aren't currently subnodes?  Can you think of an intuitive naming tweak to communicate that, maybe the variable name actually?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -991,6 +1026,90 @@ - (CGRect)convertRect:(CGRect)rect toNode:(ASDisplayNode *)node
...
+
+- (void)transitionLayoutWithAnimation:(BOOL)animated
+{
+  [self transitionLayoutThatFits:_constrainedSize animated:animated];
+}
+
+- (void)transitionLayoutThatFits:(ASSizeRange)constrainedSize animated:(BOOL)animated
+{
+  [self invalidateCalculatedLayout];
+  _deferImmediateHierarchyManagement = YES;
```
appleguy: I am not certain, but does it make sense to factor the hierarchy management into a second method call, and call it only where we do not flip this flag on?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -991,6 +1026,90 @@ - (CGRect)convertRect:(CGRect)rect toNode:(ASDisplayNode *)node
...
+    for (_ASDisplayNodePosition *position in _insertedSubnodes) {
+      [self insertSubnode:position.node atIndex:position.index];
+    }
+    _insertedSubnodes = nil;
+  }
+}
+
+- (void)__implicitlyRemoveSubnodes
+{
+  if ([_deletedSubnodes count]) {
```
appleguy: For these two methods, in their final form is it possible to remove the if clause?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2060,78 +2179,53 @@ - (void)layout
...
-  } else {
-    [_managedSubnodes insertObject:node atIndex:idx];
+  for (ASLayout *subnodeLayout in _layout.sublayouts) {
+    ((ASDisplayNode *)subnodeLayout.layoutableObject).frame = [self _adjustedFrameForLayout:subnodeLayout];
   }
-  [self addSubnode:node];
 }
 
-- (void)_implicitlyRemoveSubnode:(ASDisplayNode *)node atIndex:(NSUInteger)idx
+- (CGRect)_adjustedFrameForLayout:(ASLayout *)layout
```
appleguy: Thanks for creating this method!  Should it be a C function for performance, if changing to ASDisplayNodeCAssert would remove the last dependency on self? (I may have missed one).  If it is only called in one place, then we could also inline it without code bloat.

AsyncDisplayKit/Layout/ASLayout.h
```diff
@@ -21,20 +21,39 @@ extern BOOL CGPointIsNull(CGPoint point);
...
+
+/**
  * Array of ASLayouts. Each must have a valid non-null position.
  */
 @property (nonatomic, readonly) NSArray<ASLayout *> *sublayouts;
 
 /**
+ * A list of sublayouts that were not already flattened.
+ */
+@property (nonatomic, readonly) NSArray<ASLayout *> *immediateSublayouts;
```
appleguy: Does this need to be retained for the rest of the node lifecycle to support transitions?  If so, main concern is that it not lose its final reference when deallocating the node on the main thread, since we must dealloc all nodes there in order to ensure developers' code is safe (as the node may have a gesture recognizer or UIView type instance variable)

AsyncDisplayKit/_ASTransitionContext.h
```diff
@@ -0,0 +1,30 @@
...
+
+@protocol _ASTransitionContextDelegate <NSObject>
+
+- (void)transitionContext:(_ASTransitionContext *)context didComplete:(BOOL)didComplete;
+- (CGRect)transitionContext:(_ASTransitionContext *)context initialFrameForNode:(ASDisplayNode *)node;
+- (CGRect)transitionContext:(_ASTransitionContext *)context finalFrameForNode:(ASDisplayNode *)node;
+
+@end
+
+@interface _ASTransitionContext : NSObject <ASContextTransitioning>
```
appleguy: This file seems like it provides an interesting abstraction, but the only use case I could see, it seemed like we might just be able to call self directly to replace it. However, it feels like you have a plan for this that is not fully built out yet… Is that support in ASViewController?

aaronschubert0:
@levi Let me know if you want me to try some of this code out, I'm currently trying out some state behaviour, so this would fit nicely along side being able to re-render a node once the state changes.

jellenbogen:
This is awesome, can not wait to try this. It is very similar to what i'm already doing manually, I'll definitely be testing this out this weekend!

PR #1186, opened 2016-02-05
Enter hierarchy state on the  main thread for nodes that are loaded via legacy node datasource methods 

rahul-malik:
Updating flow for setting hierarchy state for nodeForItemAtIndexPath/nodeForRowAtIndexPath in ASTableView / ASCollectionView.
Add assertion in block based API's that the node that is created is not already
loaded.


rahul-malik:
Abandoning this in favor of the updates introduced to allow hierarchy state to be introduced off the main thread.

PR #1189, opened 2016-02-05
Revert "Merge pull request #1162 from maicki/controlnode-init-optimization"

connor:
This reverts commit 35a4b268d64e922b3928f247614b664a9413ad3b, reversing
changes made to 4201df5f31077ea31bf014fc01828c065bd31c1c.

After #1162 was landed, setting `userInteractionEnabled = YES` on a node before adding it as a subnode and then attaching gesture recognizers directly to the node's view no longer worked because the `addTarget:action:forControlEvents` never got fired.

One workaround is to set `userInteractionEnabled = YES` _after_ adding the node as a subnode. However, that seems buggy to me so I think we should revert the original diff.

### DIFF
--- a/AsyncDisplayKit/ASControlNode.m
+++ b/AsyncDisplayKit/ASControlNode.m
@@ -7,7 +7,6 @@
  */
 
 #import "ASControlNode.h"
-#import "ASDisplayNode+Subclasses.h"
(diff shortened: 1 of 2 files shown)


appleguy:
@maicki I would have never expected this!  Sorry for a much greater difficulty with this task than I would have expected :(

PR #1191, opened 2016-02-05
ASMultiplexImageNode actually always calls with the main queue. There…

garrettmoon:
… shouldn't be a reason to *not* provide a queue.


appleguy:
I may not get the full context of the diff, but should we land this change? It seems like it might result in an issue for clients who currently are not providing the queue, assuming this is public API.

Issue #330, opened 2015-02-24
[ASControlNode] Add locking for event dispatch table.


cfcommando:
No property lock or anything, just straight-up usage

appleguy:
@cfcommando sorry this is taking so long; fortunately, in nearly all use cases, the object will have its targets and actions added during the initialization process (always single-threaded). It is incredibly rare to see concurrent access to target / action information in a real application, as they do not frequently change and only the main thread triggers the events.

However, this is not a good reason to leave out locking in this class, especially because it is a core component of the framework. We will make sure it is added soon

@maicki since you are familiar with this code from recent work, could you add a MutexLocker similar to how they are used in classes like the text, image, and display nodes?

appleguy:
@cfcommando this work was completed, Let me know if you spot any other deficiencies. If you're still actively using framework, it would be great to see you on our slack channel! Send me an email if you want to invite.

PR #1199, opened 2016-02-08
[ASControlNode] Add locking for event dispatch table.

maicki:
Adds locking to ASControlNode class. Addresses issue #330.


appleguy:
Thanks @maicki !  cc @garrettmoon :) ++threadSafety

garrettmoon:
Yeesh, I'm starting to wonder if we should do an audit of thread safety. @appleguy do you know of a good way to prioritize something like that? I know ASMultiplexNode almost completely ignores it :/

PR #1200, opened 2016-02-08
[ASVideoNode] Rename lock instance variable

maicki:
To be consistent with ASControlNode and ASImageNode lock variable naming, maybe it would make sense to rename the lock variable name within ASVideoNodeLock from a generic to a more specific one.

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -11,7 +11,7 @@
 
 @interface ASVideoNode ()
 {
-  ASDN::RecursiveMutex _lock;
+  ASDN::RecursiveMutex _videoLock;
   
   __weak id<ASVideoNodeDelegate> _delegate;
 
@@ -209,7 +209,7 @@
   }
 
   {
-    ASDN::MutexLocker l(_lock);
+    ASDN::MutexLocker l(_videoLock);
     _currentItem = [[AVPlayerItem alloc] initWithAsset:_asset];
     [_currentItem addObserver:self forKeyPath:NSStringFromSelector(@selector(status)) options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew context:NULL];
 
@@ -227,7 +227,7 @@
   [super clearFetchedData];
   
   {
-    ASDN::MutexLocker l(_lock);
+    ASDN::MutexLocker l(_videoLock);
     ((AVPlayerLayer *)_playerNode.layer).player = nil;
     _player = nil;
   }
@@ -235,7 +235,7 @@
 
 - (void)visibilityDidChange:(BOOL)isVisible
 {
-  ASDN::MutexLocker l(_lock);
+  ASDN::MutexLocker l(_videoLock);
   
   if (_shouldAutoplay && _playerNode.isNodeLoaded) {
     [self play];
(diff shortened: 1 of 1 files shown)


appleguy:
@maicki cool, I like this

appleguy:
@maicki I am retrying both of your builds. Could you check into the cause of the failures if they fail again?

appleguy:
hmm, this is concerning... @knopp any possibility this is from your changes?  Although they have been stable in my use cases, I would like to push out a new release in less than a week to fix some important issues, including a swift build error. So we should make sure to do any final vetting on the new transaction system.
 
/Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKitTests/ASDisplayLayerTests.m:515: ((ASDisplayNodeRunRunLoopUntilBlockIsTrue(^BOOL{ return (layer2Delegate.displayCount == 1); })) is true) failed:
512   dispatch_resume([_ASDisplayLayer displayQueue]);
513   XCTAssertTrue(ASDisplayNodeRunRunLoopUntilBlockIsTrue(^BOOL{
514     return (layer2Delegate.displayCount == 1);
515   }));
      ~~~~
516 
517   // check layer1 has not had async display run
━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    ✗ -[_ASDisplayLayerTests testTransaction] (30004 ms) (0)
    236 passed, 1 failed, 0 errored, 237 total (49162 ms)
Failures:
  0) -[_ASDisplayLayerTests testTransaction] (AsyncDisplayKitTests.xctest)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
/Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKitTests/ASDisplayLayerTests.m:515: ((ASDisplayNodeRunRunLoopUntilBlockIsTrue(^BOOL{ return (layer2Delegate.displayCount == 1); })) is true) failed:
512   dispatch_resume([_ASDisplayLayer displayQueue]);
513   XCTAssertTrue(ASDisplayNodeRunRunLoopUntilBlockIsTrue(^BOOL{
514     return (layer2Delegate.displayCount == 1);
515   }));
      ~~~~
516 
517   // check layer1 has not had async display run

appleguy:
@maicki awesome getting pull request #1200!

a little bit concerning that the tests passed this time, clearly there is some kind of timing issue…

knopp:
@appleguy, it's very hard to say, given that the problem doesn't seem to be reproducible. I've tested the new transaction code a lot and never seen and draw calls missing. 

Anyway, I'll leave the test running locally with only once CPU core enabled (like on travis) over and over again, we'll see if it can be reproduced.

knopp:
@appleguy, I have some updates here.

The problem can be reproduced locally. You need to disable all CPU cores in instruments (only leave one active) and then run the `testTransaction` test for couple of tens of thousands times.

Eventually, it will get stuck. But the reason why it gets stuck is that GCD will refuse to overcommit the display queue. So the `dispatch_semaphore_wait` is blocking the display queue, and all subsequent calls are stuck, because the task that unblocks the semaphore is waiting in queue too.

You can easily test this. When the test gets stuck, you can try scheduling a block on display queue inside the `ASDisplayNodeRunRunLoopUntilBlockIsTrue` loop - it will never be executed. 

So while the test fails occasionally, I think this is pretty harmless, because normally (outside of tests) one display queue task will not block waiting for another display queue block. 

I'm not sure how to fix the test though. GCD doesn't make any guarantees about overcommit, so you can't expect to have more threads running than CPU cores (even though it usually it is the case), which is required for this test to pass.

PR #1207, opened 2016-02-09
Insert/delete should take precedence over Update/Reload

binl:
@levi @appleguy @nguyenhuy 
When user performs a batchUpdate on a collectionView and combines insertions and reloads:
e.g.
```
insert a row at index 2
update a row at indexes 0 and 3 (after insertion it will be 4)
```
At this point, the data sources for these rows should've updated, meaning:
```
index 0: content is updated
index 2: content is newly inserted
index 3: old content that used to be at index 2
index 4: updated content that used to be at index 3
```

With our current implementation, when you try to update row at old index 3:
You either pass in the index 3:
```
changeSet will trigger reload, which will ask our ASCV dataSource for a new node at index **3** 
and that Node will have **old content that used to be at index 2**, 
but will be put in the correct position because insertion happens afterwards
```
Or you pass in the index 4:
```
changeSet will trigger reload, which will ask our ASCV dataSource for a new node at index 4 and
that Node will have the correct content, but since insertion hasn't happen yet, 
this new node replaces the node at index 4 which has nothing to do with this whole operation
```

Therefore, Update/Reloads should happen after insertion/deletions.

### DIFF
--- a/AsyncDisplayKit/Details/ASChangeSetDataController.m
+++ b/AsyncDisplayKit/Details/ASChangeSetDataController.m
@@ -51,14 +51,6 @@
     [_changeSet markCompleted];
     
     [super beginUpdates];
-  
-    for (_ASHierarchySectionChange *change in [_changeSet sectionChangesOfType:_ASHierarchyChangeTypeReload]) {
-      [super reloadSections:change.indexSet withAnimationOptions:change.animationOptions];
-    }
-    
-    for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeReload]) {
-      [super reloadRowsAtIndexPaths:change.indexPaths withAnimationOptions:change.animationOptions];
-    }
     
     for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeDelete]) {
       [super deleteRowsAtIndexPaths:change.indexPaths withAnimationOptions:change.animationOptions];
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
Thank you for bringing this up! This is hugely important. I like this change, but I believe we need to map the reloaded indexes to their after-the-update values. The indexes the user provides us for reloads are before-the-update indexes, so as soon as we process any inserts/deletes, those indexes are invalid.

_ASHierarchyChangeSet already has `newSectionForOldSection` so what we need to do is write a corresponding `newIndexPathForOldIndexPath` method. Then use those methods to map all the reloaded indexes/paths to their new values, and then call `[super reload…]` using those new indexes.

`newIndexPathForOldIndexPath` is a medium-difficulty task, if we want it to be efficient.

The first part is easy – use newSectionForOldSection to see what new section the item corresponds to. This should never return `NSNotFound` because we filter out reloads that target inserted/reloaded sections.

The harder part is, we need to compute a pair of NSIndexSets for each section that was modified, and just like in `newSectionForOldSection`, we'll use two calls to `countOfIndexesInRange` to compute the new item index, reading the deletes from the old section and inserts from the new one.

Here's an example implementation:

```objectivec
- (nullable NSIndexPath *)newIndexPathForOldIndexPath:(NSIndexPath *)oldIndexPath {
    
    // Is there a new section that corresponds to this one?
    const NSInteger oldSection = [oldIndexPath indexAtPosition:0];
    const NSInteger newSection = [self newSectionForOldSection:oldSection];
    if (newSection == NSNotFound) {
        return nil;
    }
    
    // Start with old, remove deletes/moves-out beneath, then add moves-in/inserts beneath
    const NSInteger oldIndex = [oldIndexPath indexAtPosition:1];
    const NSInteger indexAfterDeletes = oldIndex - [[self indexesRemovedFromSection:oldSection] countOfIndexesInRange:NSMakeRange(0, oldIndex)];

    __block NSInteger newIndex = indexAfterDeletes;
    NSIndexSet *inserts = [self indexesInsertedInSection:newSection];
    [inserts enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL * _Nonnull stop) {
        if (idx <= newIndex) {
            newIndex += 1;
        } else {
            *stop = YES;
        }
    }];

    return [NSIndexPath indexPathForItem:newIndex inSection:newSection];
}
```

Also, we need to fix the implementation of `newSectionForOldSection` because as I look at it, I see it uses `[_insertedSections countOfIndexesInRange:]` to do inserts, which isn't safe because of the way a prior insert can enable a later insert. That newSectionForOldSection should compute the new section using the same method I showed above, with `enumerateIndexesUsingBlock`. 

Are you up for handling this? I can absolutely answer any questions or trouble you encounter but I'm a bit bandwidth-limited right now. Also @appleguy should take a looky-loo.

binl:
@Adlai-Holler  Yes! I have came to the exact same conclusion after looking at UIKit's behavior and our code.

appleguy:
@binl after more detailed discussions, it works like a new diff will be coming to address this issue at a deeper level.

PR #1209, opened 2016-02-09
Add pin remote image downloader

garrettmoon:


appleguy:
This is a hugely important change for the framework… Thank you so much for working on it  @garrettmoon! It will meaningfully improve in the-application experiences as well as the developer experiences for any of the 2000 and clients who update the framework, and no doubt for thousands more to come!

PR #1211, opened 2016-02-10
[Table, Collection, Pager] Update data source protocols to not @require non-block node method.

rahul-malik:
This is mostly a follow-up to the original addition of the block-based asynchronous node creation api that was introduced recently. This change makes synchronous node creation api methods optional but enforces that one of the two approaches is implemented when the dataSource property is set.


appleguy:
Very awesome, thanks @rahul-malik!

PR #1212, opened 2016-02-10
Fix allocatedNodes Subarray to match indexPaths subarray

binl:
@appleguy @rahul-malik @nguyenhuy


appleguy:
@binl this is a great find, when was this introduced?  Temped to merge now from mobile, but will wait for build.

binl:
Thanks! This will help fix a header layout issue! Not sure, it was probably in the massive async allocation work.

rahul-malik:
Nice find! 

Yep this was part of my change. I thought it was safe at the time since the layout method it calls iterated through the indexPaths array which I thought accessed the allocated nodes array in the appropriate indexes. I was mainly trying to avoid creation of another sub array but obviously the bug fix is more important.

PR #1219, opened 2016-02-11
[ASDataController] Adjust index paths for reloads contained in the same changeset as deletes / inserts.

binl:
I'm basically updating the indexPaths of all the reloads based on any deletion/insertions before it.

Here's the verified UIKit behavior, for reference:
**delete/reload** indexPaths that you pass in should all be their** current** indexPaths
**insert** indexPaths that you pass in should all be their **future** indexPaths after deletions

@appleguy @rahul-malik @Adlai-Holler

### DIFF
--- a/AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
+++ b/AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
@@ -198,8 +198,40 @@
 
     // Ignore item inserts in reloaded(new)/inserted sections.
     [_ASHierarchyItemChange sortAndCoalesceChanges:_insertItemChanges ignoringChangesInSections:insertedOrReloaded];
+    
+
+    // reload itemsChanges need to be adjusted so that we access the correct indexPaths in the datasource
+    NSDictionary *insertedIndexPathsMap = [_ASHierarchyItemChange sectionToIndexSetMapFromChanges:_insertItemChanges ofType:_ASHierarchyChangeTypeInsert];
+    NSDictionary *deletedIndexPathsMap = [_ASHierarchyItemChange sectionToIndexSetMapFromChanges:_deleteItemChanges ofType:_ASHierarchyChangeTypeDelete];
+
+    for (_ASHierarchyItemChange *change in _reloadItemChanges) {
+      NSAssert(change.changeType == _ASHierarchyChangeTypeReload, @"It must be a reload change to be in here");
+      NSMutableArray *newIndexPaths = [NSMutableArray array];
+      
+      // Every indexPaths in the change need to update its section and/or row
+      // depending on all the deletions and insertions
+      for (NSIndexPath *indexPath in change.indexPaths) {
+        NSUInteger section = indexPath.section;
+        NSUInteger row = indexPath.row;
+        
+        section -= [_deletedSections countOfIndexesInRange:NSMakeRange(0, section)];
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
```diff
@@ -198,9 +198,41 @@ - (void)_sortAndCoalesceChangeArrays
...
+        NSUInteger section = indexPath.section;
+        NSUInteger row = indexPath.row;
+        
+        section -= [_deletedSections countOfIndexesInRange:NSMakeRange(0, section)];
+        section += [_insertedSections countOfIndexesInRange:NSMakeRange(0, section)];
+        
+        NSIndexSet *indicesInsertedInSection = insertedIndexPathsMap[@(section).stringValue];
+        row += [indicesInsertedInSection countOfIndexesInRange:NSMakeRange(0, row)];
+        NSIndexSet *indicesDeletedInSection = deletedIndexPathsMap[@(section).stringValue];
+        row += [indicesDeletedInSection countOfIndexesInRange:NSMakeRange(0, row)];
```
Adlai-Holler: Hmm I'm pretty sure the logic here is incorrect at both the section and row levels.

`countOfIndexesInRange` won't handle cascaded inserts correctly. For instance, say you reload item 0, and insert item at 0. We expect the reloaded index to map to 1. But `countOfIndexesInRange:(0, 0) --> 0` so we'll end up having it map to 1.

Deletes should be subtracted before inserts are added. This is because the deleted indexes are "old indexes" whereas inserted indexes are new indexes. 

Typo: Line 221 should be `-=`

See [this comment](https://github.com/facebook/AsyncDisplayKit/pull/1207#issuecomment-182172720) for a thoroughly-tested example.

AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
```diff
@@ -301,6 +333,28 @@ - (instancetype)initWithChangeType:(_ASHierarchyChangeType)changeType indexPaths
...
+// e.g. changes: (0 - 0), (0 - 1), (2 - 5)
+//  will become: {@0 : [0, 1], @2 : [5]}
++ (NSDictionary *)sectionToIndexSetMapFromChanges:(NSArray *)changes ofType:(_ASHierarchyChangeType)changeType
+{
+  NSMutableDictionary *sectionToIndexSetMap = [NSMutableDictionary dictionary];
+  for (_ASHierarchyItemChange *change in changes) {
+    NSAssert(change.changeType == changeType, @"The map we created must all be of the same changeType as of now");
+    for (NSIndexPath *indexPath in change.indexPaths) {
+      NSString *sectionKey = @(indexPath.section).stringValue;
+      NSMutableIndexSet *indexSet = sectionToIndexSetMap[sectionKey];
```
Adlai-Holler: Why parse the section number into a string? Why not use the NSNumber as the key?

AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
binl: @Adlai-Holler absolutely! I realized these issues as I did more testing. addressing them now. I also found other issues with reloads as we just delete+insert them, and after the nodes are completed, we call the super methods on the UICollectionView. This means that the delete indexPaths should be current and insert indexPaths should be future, which will be different in many cases, but we are passing through the same set.

I will have a PR soon.

PR #1221, opened 2016-02-12
ASPagerNode uses its view's size for node constrained sizes

nguyenhuy:


levi:
What was happening on account of using the bridging property?

PR #1223, opened 2016-02-12
[ASTextNode] Add support for adjusting font size to make text fit within constrained size without truncating.

rcancro:
First attempt to scale a font to have it fit as much as possible within its maximum line count.

The change added a currentScaleFactor member to the renderer and text node, since scaling the renderer's attributed string would only last the lifetime of the renderer (which isn't very long).

Here is an example of it doing its thing in the Kitten app:
screen shot 2016-02-12 at 10 37 07


AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -0,0 +1,185 @@
...
+}
+
+- (NSUInteger)lineCountForString:(NSAttributedString *)attributedString
+{
+  NSUInteger lineCount = 0;
+  
+  static std::mutex __static_mutex;
+  std::lock_guard<std::mutex> l(__static_mutex);
+  
+  NSTextStorage *textStorage = [[NSTextStorage alloc] initWithAttributedString:attributedString];
```
rcancro: Using the current storage, manager and container didn't seem to work, so I made new ones. Not sure if I need the lock here since we are already in the _context's lock.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -176,11 +183,32 @@ - (void)drawInContext:(CGContextRef)context bounds:(CGRect)bounds;
   LOG(@"%@, shadowInsetBounds = %@",self, NSStringFromCGRect(shadowInsetBounds));
   
   [[self context] performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
+    
+    NSTextStorage *scaledTextStorage = nil;
+    BOOL isScaled = (self.currentScaleFactor > 0 && self.currentScaleFactor < 1.0);
+
+    if (isScaled) {
```
rcancro: This seemed questionable. Is it kosher to do things like this?

rahul-malik:
This is awesome!

nguyenhuy:
:+1: :100:

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
appleguy: @rcancro my main objection to this is accessing the property twice :).  What specifically were you worried about, though? I think it is a fine practice to create the local variable isScaled, but you might instead create "scaleFactor" (for property access) and put the checks in the if () directly, with a brief comment about what the scenario is designed to detect.

appleguy:
Some serious work here!!  Thanks @rcancro!  This will definitely be called out in the next release notes, At least as soon as we feel it is tested and stable.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
rcancro: Oops, I highlighted the wrong section. The part I wasn't sure about was swapping out the layout manager's text storage right before render.

Issue #1169, opened 2016-02-03
[ASPagerNode] Add support for rotation.


dermotos:
I've swapped out a `UIPageViewController` and replaced it with an `ASPagerNode`. Works great, but when rotating/transitioning to a new size, the pager node doesn't remain on the current page, and instead scrolls off to the left/right and leaves the ASPagerNode halfway between pages.

Also receive the following error:
```
2016-02-03 14:53:34.323 AFR-dev[1227:400873] the behavior of the UICollectionViewFlowLayout is not defined because:
2016-02-03 14:53:34.323 AFR-dev[1227:400873] the item height must be less than the height of the UICollectionView minus the section insets top and bottom values, minus the content insets top and bottom values.
2016-02-03 14:53:34.326 AFR-dev[1227:400873] The relevant UICollectionViewFlowLayout instance is <UICollectionViewFlowLayout: 0x141096330>, and it is attached to <ASCollectionView: 0x1400e2e00; baseClass = UICollectionView; frame = (0 0; 1024 768); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x14105a090>; animations = { position=<CABasicAnimation: 0x141080490>; bounds=<CABasicAnimation: 0x14162a6f0>; }; layer = <<ASPagerNode: 0x13ff97d00>, layer = <_ASDisplayLayer: 0x13ffa3210>>; contentOffset: {0, 0}; contentSize: {22272, 1024}> collection view layout: <UICollectionViewFlowLayout: 0x141096330>.
2016-02-03 14:53:34.326 AFR-dev[1227:400873] Make a symbolic breakpoint at UICollectionViewFlowLayoutBreakForInvalidSizes to catch this in the debugger.
```

levi:
Thanks for catching this. The current iteration of ASPagerNode wasn't built to support orientation changes, albeit it should. I will add it to the list of improvements to make. In the meantime, feel free to submit a patch that keeps the current page in view on orientation changes.

aaronschubert0:
@dermotos @levi I've came across this as well. Currently, I store the current page and then in `didRotateFromInterfaceOrientation` I call scrollToPage on ASPagerNode without animation, this seems to work relatively well.

appleguy:
@levi we will need to figure out how to make this work for the notifications tab, with your work on iPad rotation. Could you prioritize that in this coming week, and update this task with your findings?

[...]

PR #1224, opened 2016-02-12
[ASPagerNode] Add automatic rotation support

levi:
Adds a custom UICollectionViewLayout that identifies the centered item upon rotation and updates the contentOffset when the bounds change has completed. The final contentOffset will be centered horizontally.

### DIFF
--- a/AsyncDisplayKit/ASPagerFlowLayout.m
+++ b/AsyncDisplayKit/ASPagerFlowLayout.m
@@ -1,0 +1,63 @@
+//
+//  ASPagerFlowLayout.m
+//  AsyncDisplayKit
+//
+//  Created by Levi McCallum on 2/12/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import "ASPagerFlowLayout.h"
+
+@interface ASPagerFlowLayout ()
+
+@property (strong, nonatomic) NSIndexPath *currentIndexPath;
+
+@end
+
+@implementation ASPagerFlowLayout
+
+- (void)invalidateLayout
+{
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/ASPagerFlowLayout.m
```diff
@@ -0,0 +1,63 @@
...
+
+@property (strong, nonatomic) NSIndexPath *currentIndexPath;
+
+@end
+
+@implementation ASPagerFlowLayout
+
+- (void)invalidateLayout
+{
+  self.currentIndexPath = [self _indexPathForVisiblyCenteredItem];
```
appleguy: Because it looks like we only set this upon invalidation of the layout, we should probably call it previous page indexpath, or something else. Having trouble thinking of a good name, but I am worried that it might later be used as though it is up-to-date value at all times.

Maybe I don't totally understand all the conditions under which invalidateLayou is called, though.

levi: It's basically called whenever there's a bounds change. Technically the `currentIndexPath` should be the current index path even after the invalidation, as we're keeping state here of what item was centered before the bounds change and ensuring that it remains centered after the layout rebuilds itself.

raphaeloliveira:
Hi guys, I'm coming from the issue https://github.com/facebook/AsyncDisplayKit/issues/1169 and even though the pager node stays on the same page when rotating I still can see parts of the other screen. 

This is the pager node sample project running on an iPhone 7:

rotation_issue

Do you guys have any tips on how to solve this?

PR #1225, opened 2016-02-12
[ASNetworkImageNode] Add fetch data delegate method; cache respondsToSelector results.

garrettmoon:


AsyncDisplayKit/ASNetworkImageNode.h
```diff
@@ -96,6 +96,15 @@ NS_ASSUME_NONNULL_BEGIN
 @optional
 
 /**
+ * Notification that the image node started to load
+ *
+ * @param imageNode The sender.
+ *
+ * @discussion Called on a background queue.
+ */
+- (void)imageNodeDidStartLoading:(ASNetworkImageNode *)imageNode;
```
appleguy: Subtle difference but maybe call this imageNodeDidStartFetchingData -- loading is fairly clear but might be viewed as being related to display, and FetchData is the interfaceState terminology to acknowledge local database / disk cache loads.  Please consider alternative names too as I don't love that one!

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -35,10 +35,15 @@ @interface ASNetworkImageNode ()
 
   BOOL _imageLoaded;
   
+  BOOL _delegateSupportsDidStartLoading;
```
appleguy: yay!

AsyncDisplayKit/ASNetworkImageNode.h
garrettmoon: I actually like imageNodeDidStartFetchingData :)

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -377,7 +396,8 @@ - (void)_lazilyLoadImageIfNecessary
 - (void)asyncdisplaykit_asyncTransactionContainerStateDidChange
 {
   if (self.asyncdisplaykit_asyncTransactionContainerState == ASAsyncTransactionContainerStateNoTransactions) {
-    if (self.layer.contents != nil && [self.delegate respondsToSelector:@selector(imageNodeDidFinishDecoding:)]) {
+    ASDN::MutexLocker l(_lock);
+    if (self.layer.contents != nil && _delegateSupportsImageNodeDidFinishDecoding) {
       [self.delegate imageNodeDidFinishDecoding:self];
```
appleguy: Should we avoid holding the lock during the delegate calls?  I'm not totally sure, this is probably OK since they are recursive locks, just wondering if any use case might have some other thread interdependency.

garrettmoon: I was thinking about that too, but to be truly thread safe, we need to, otherwise the delegate *could* change between when we get the value in the lock and when we call the delegate method. It probably would never happen, but…

PR #1228, opened 2016-02-16
remove second [self addSubnode:self.soldOutOverlay]

keywind:
addSubnode self.soldOutOverlay twice


appleguy:
@keywind haha, nice!  thanks

PR #1229, opened 2016-02-16
Remove wordkerner in ASTextNode to fix jumbled text bug

rcancro:
Before:
screen shot 2016-02-16 at 09 31 52

After:
screen shot 2016-02-16 at 09 31 17

Here is a simple app to recreate the bug:
[JumbledText.zip](https://github.com/facebook/AsyncDisplayKit/files/132985/JumbledText.zip)

Eventually I want to look into the issue, but until then this will stop the text error.

Word Kerner is also used in ASEditableTextNode. Should we take it out there as well?


levi:
Fixing the bug by deleting stuff. My hero!

PR #1233, opened 2016-02-16
[ASRangeController] Ensure Cells That Are in Window Are Always in FetchData, Display Ranges

Adlai-Holler:
@appleguy I decided to run this alongside the existing cables for `_ASTableViewCellDelegate`. 

It works great! With this change, I was able to ensure the first cell in a presented table view was rendered before it came on screen! :zap:


Adlai-Holler:
The use case this resolved for me was already fixed before the PR. I had only tested my use case against prior commit 35a4b268d64e922b3928f247614b664a9413ad3b.

PR #1235, opened 2016-02-17
[ASNetworkImageNode] Provide Synchronous Loading Path on Cache Hits

Adlai-Holler:
A good old fashioned main-thread short-circuit for our image loading. @appleguy


appleguy:
@Adlai-Holler thanks, this looks like a solid improvement!

PR #1236, opened 2016-02-17
[ASMultiplexImageNode] Check loadedImageIdentifer before sending it to delegate

ngleanh:
As the delegate `- (nullable NSURL *)multiplexImageNode:(ASMultiplexImageNode *)imageNode URLForImageIdentifier:(ASImageIdentifier)imageIdentifier;` does not require `imageIdentifier` to be nullable, we need to check `loadedImageIdentifier` before sending it to the delegate.

I'm not pretty sure why this variable is null, but when using this class on Swift, it causes crashing from my end.


appleguy:
@garrettmoon, @Adlai-Holler -- any ideas on the scenario that would cause this? If it is of no concern that this is possible, I'm happy to merge this.

garrettmoon:
LGTM, loadedImageIdentifier can indeed be nil at this point if an image isn't loaded before cleareFetchedData is called.

PR #1246, opened 2016-02-19
[ASCellNode] Support monitoring scrollViewDidScroll events / visible amount of self

Xinchi:
Sending out the PR early for high-level design issues.  I am sure there are details that are missing from this PR, or major issues.


AsyncDisplayKit/ASCellNode.h
```diff
@@ -16,6 +16,7 @@ typedef NSUInteger ASCellNodeAnimation;
 
 @protocol ASCellNodeLayoutDelegate <NSObject>
 
+- (void)scrollViewDidScroll:(UIScrollView *)scrollView cellFrameInScrollView:(CGRect)cellFrame;
```
appleguy: Because this is a delegate, we need to send the cell too.

Let's do visibleNode:(ASCellNode *)node scrollViewDidScroll:(UIScrollView *)scrollView withCellFrameInScrollView:(CGRect)cellFrame

AsyncDisplayKit/ASCellNode.h
```diff
@@ -24,6 +25,8 @@ typedef NSUInteger ASCellNodeAnimation;
  * @param sizeChanged `YES` if the node's `calculatedSize` changed during the relayout, `NO` otherwise.
  */
 - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged;
+
+
```
appleguy: Only one return

AsyncDisplayKit/ASCellNode.h
```diff
@@ -75,6 +78,7 @@ typedef NSUInteger ASCellNodeAnimation;
  */
 @property (nonatomic, weak) id<ASCellNodeLayoutDelegate> layoutDelegate;
 
+@property (nonatomic, assign) BOOL shouldObserveVisibility;
```
appleguy: I'm not sure we should provide this API. It is a bit confusing to clients. This is probably fine for now, Not a reason to block the diff, but it would be important to change this before the next production release of the framework (which could be as early as this weekend) if it will be changed.

AsyncDisplayKit/ASCellNode.h
```diff
@@ -107,6 +111,7 @@ typedef NSUInteger ASCellNodeAnimation;
  */
 - (instancetype)initWithViewControllerBlock:(ASDisplayNodeViewControllerBlock)viewControllerBlock didLoadBlock:(ASDisplayNodeDidLoadBlock)didLoadBlock;
 
+- (void)updateScrollSituationWithScrollVIew:(UIScrollView *)scrollView;
```
appleguy: Let's call this visibleNodeScrollViewDidScroll:(UIScrollView *)scrollView withCellFrameInScrollView:(CGRect)cellFrame

AsyncDisplayKit/ASCellNode.m
```diff
@@ -134,6 +134,12 @@ - (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event
   [(_ASDisplayView *)self.view __forwardTouchesCancelled:touches withEvent:event];
 }
 
+- (void)updateScrollSituationWithScrollVIew:(UIScrollView *)scrollView
+{
+    // TODO(Max): Fix the cellFrame here
+    [self.layoutDelegate scrollViewDidScroll:scrollView cellFrameInScrollView:CGRectZero];
```
appleguy: We should implement this in hey way that works even if the developer overrides this method and forgets to call the super method. You can do this by using a different method name, such as one that starts with _, as the core internal implementation that is responsible for calling both the method that might be subclassed as well as the delegate.

I think we should replace the property value added to control whether we track visibility, And instead look at response to selector were both the delegate and the cell subclass

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -212,6 +212,7 @@ - (instancetype)_initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionV
   
   _registeredSupplementaryKinds = [NSMutableSet set];
   
+  _cellsForVisibilityUpdates = [[NSMutableArray alloc] init];
```
appleguy: Let's make this a mutable set.  It will not be quite as efficient, but It will address the possibility of calls being out of sync, and prevent anything from being added twice. This is mainly a risk on IOS seven where there is not a true implementation for the will display cell call, and we call it manually.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -650,6 +653,17 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
   }
 }
 
+- (void)scrollViewDidScroll:(UIScrollView *)scrollView{
```
appleguy: { on new line

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -650,6 +653,17 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
...
+    for (ASCellNode *node in _cellsForVisibilityUpdates) {
+        if (node.shouldObserveVisibility) {
+            [node updateScrollSituationWithScrollVIew:scrollView];
+        }
+    }
+    if ([_asyncDelegate respondsToSelector:@selector(scrollViewDidScroll:)]) {
```
appleguy: Can you add a Boolean instance variable to cache the result of this? That would need to be set in the setAsyncDelegate method.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -650,6 +653,17 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
   }
 }
 
+- (void)scrollViewDidScroll:(UIScrollView *)scrollView{
+    for (ASCellNode *node in _cellsForVisibilityUpdates) {
+        if (node.shouldObserveVisibility) {
```
appleguy: We could actually keep this property, but I would not expose it as public API. Instead, we should internally set the value of this property based on whether either the cell subclass or its layout delegate implement the method. In order to be able to collect here, without it being in the public header, you could add a @interface ASCellNode (Internal) at the top of this file… or look for another file where it Would be appropriate to put this kind of internal method

AsyncDisplayKit/ASCellNode.h
```diff
@@ -24,6 +24,16 @@ typedef NSUInteger ASCellNodeAnimation;
  * @param sizeChanged `YES` if the node's `calculatedSize` changed during the relayout, `NO` otherwise.
  */
 - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged;
+
+@optional
+/**
+ * Notifies the delegate that the specified cell node has scrolled
+ *
+ * @param node: A node informing the delegate about the scroll
+ * @param cellFrame: The frame of the cell that has just scrolled
```
appleguy: add to the cellFrame - "...scrolled, in the coordinate system of the scroll view (this is different than the ASCellNode's frame).  Use this value for efficiency instead of convertRect: from the ASCellNode to the UIScrollView."

AsyncDisplayKit/ASCellNode.h
```diff
@@ -107,6 +119,8 @@ typedef NSUInteger ASCellNodeAnimation;
  */
 - (instancetype)initWithViewControllerBlock:(ASDisplayNodeViewControllerBlock)viewControllerBlock didLoadBlock:(ASDisplayNodeDidLoadBlock)didLoadBlock;
 
+- (void)_visibleNodeDidScroll:(UIScrollView *)scrollView withCellFrame:(CGRect)cellFrame;
```
appleguy: This should not be in the public header - move it to the +Internal

AsyncDisplayKit/ASCellNode.m
```diff
@@ -9,6 +9,7 @@
 #import "ASCellNode.h"
 
 #import "ASInternalHelpers.h"
+#import "ASCollectionView.h"
```
appleguy: This should not be required?

AsyncDisplayKit/ASCellNode.m
```diff
@@ -36,7 +38,9 @@ - (instancetype)init
   // Use UITableViewCell defaults
   _selectionStyle = UITableViewCellSelectionStyleDefault;
   self.clipsToBounds = YES;
-
+  if ([self.layoutDelegate respondsToSelector:@selector(visibleNodeDidScroll:inScrollView:withCellFrame:)]) {
```
appleguy: Layout delegate will not be set here.  This point, we should only check [self respondsToSelector:(the selector that subclasses of cell node can override)]

AsyncDisplayKit/ASCellNode.m
```diff
@@ -57,6 +61,9 @@ - (instancetype)initWithViewControllerBlock:(ASDisplayNodeViewControllerBlock)vi
     [self addSubnode:_viewControllerNode];
   }
   
+  if ([self.layoutDelegate respondsToSelector:@selector(visibleNodeDidScroll:inScrollView:withCellFrame:)]) {
```
appleguy: This can be removed entirely, the other init method will handle it.

AsyncDisplayKit/ASCellNode.m
```diff
@@ -134,6 +141,13 @@ - (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event
   [(_ASDisplayView *)self.view __forwardTouchesCancelled:touches withEvent:event];
 }
```
appleguy: Somewhere this file, we need to implement a set method for the layout delegate, so that when a layout delegate is provided, we can check if it responds to the scroll selector.

I am wondering though, what use case do you have for the layout delegate getting this call? Maybe we can avoid this completely for now, and require that developers implement a subclass of the cell to listen to the method, and forward it to a delegate of its own creation and if they will

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -537,6 +541,7 @@ - (void)collectionView:(UICollectionView *)collectionView willDisplayCell:(UICol
   if (cellNode.neverShowPlaceholders) {
     [cellNode recursivelyEnsureDisplaySynchronously:YES];
   }
+  [_cellsForVisibilityUpdates addObject:cell];
```
appleguy: This is where we should check if [cell node] respondsToSelector: / needs the visibility calls.

You can change the type of the argument to willDisplayCell:(_ASCollectionViewCell *)cell  - so that you can call the -node method on it.

You should still add the cell itself to the set, but check the node for whether it needs the notifications.

AsyncDisplayKit/ASCellNode.m
```diff
@@ -134,6 +133,11 @@ - (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event
...
 
+- (void)_visibleNodeDidScroll:(UIScrollView *)scrollView withCellFrame:(CGRect)cellFrame
```
appleguy: rename all instances of this to not have _.  now that we support the subclass override thing, we should make this method selector public and no _.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -537,6 +542,9 @@ - (void)collectionView:(UICollectionView *)collectionView willDisplayCell:(UICol
   if (cellNode.neverShowPlaceholders) {
     [cellNode recursivelyEnsureDisplaySynchronously:YES];
   }
+  if (ASSubclassOverridesSelector([ASCellNode class], [cellNode class], @selector(_visibleNodeDidScroll:withCellFrame:))) {
```
appleguy: perfect, this is awesome

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -548,6 +556,10 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
     ASDisplayNodeAssertNotNil(node, @"Expected node associated with removed cell not to be nil.");
     [_asyncDelegate collectionView:self didEndDisplayingNode:node forItemAtIndexPath:indexPath];
   }
+  if ([_cellsForVisibilityUpdates containsObject:cell]) {
+    [_cellsForVisibilityUpdates removeObject:cell];
```
appleguy: just do removeObject without checking, no if() required

appleguy:
This is a very exciting capability to have! Thank you, @Xinchi !  cc @Adlai-Holler

Adlai-Holler:
Awesome, thank you @Xinchi! Gonna start using this immediately.

PR #1250, opened 2016-02-19
fix scheduling issue that causes collectionView to not animate

lkzhao:


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -461,8 +463,10 @@ - (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
     }];
     [_pendingEditCommandBlocks removeAllObjects];
 
-    [_mainSerialQueue performBlockOnMainThread:^{
-      [_delegate dataController:self endUpdatesAnimated:animated completion:completion];
+    [_editingTransactionQueue addOperationWithBlock:^{
```
appleguy: Can you add a comment for why this is necessary, in an intuitive way, In terms of describing a scenario in which it is important to go through the editing transaction queue?

Is there any reason in particular that at the top of this method you use the approach of waiting, where is here use the approach of scheduling?

lkzhao: I can wait here as well. we just need all the editingTransaction to schedule their operations on the _mainSerialQueue before we schedule this. I think that might be cleaner.

PR #1251, opened 2016-02-19
Fix Umbrella Header

ekurutepe:
Adds four missing public headers to the umbrella header in order to prevent warnings when the framework built with the framework target is linked with another project.

Also removes a private header from the umbrella header and from the public headers group.

https://github.com/ekurutepe/ASDKTest is a simple test project which should compile without warnings.

### DIFF
--- a/AsyncDisplayKit/AsyncDisplayKit.h
+++ b/AsyncDisplayKit/AsyncDisplayKit.h
@@ -51,8 +51,11 @@
 #import <AsyncDisplayKit/ASStackLayoutSpec.h>
 
 #import <AsyncDisplayKit/_ASAsyncTransaction.h>
-#import <AsyncDisplayKit/_ASAsyncTransactionContainer+Private.h>
 #import <AsyncDisplayKit/_ASAsyncTransactionGroup.h>
+#import <AsyncDisplayKit/_ASDisplayView.h>
+#import <AsyncDisplayKit/ASDisplayNode+Beta.h>
+#import <AsyncDisplayKit/ASTextNode+Beta.h>
+#import <AsyncDisplayKit/ASTextNodeTypes.h>
 #import <AsyncDisplayKit/ASAvailability.h>
 #import <AsyncDisplayKit/ASCollectionViewLayoutController.h>
 #import <AsyncDisplayKit/ASContextTransitioning.h>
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/AsyncDisplayKit.h
```diff
@@ -51,8 +51,11 @@
 #import <AsyncDisplayKit/ASStackLayoutSpec.h>
 
 #import <AsyncDisplayKit/_ASAsyncTransaction.h>
-#import <AsyncDisplayKit/_ASAsyncTransactionContainer+Private.h>
 #import <AsyncDisplayKit/_ASAsyncTransactionGroup.h>
+#import <AsyncDisplayKit/_ASDisplayView.h>
+#import <AsyncDisplayKit/ASDisplayNode+Beta.h>
```
appleguy: We need to avoid including the +Beta ones here. They are designed to require being explicitly included by the developer, so that they realize they're relying on beta functionality.

There should not be a reason for developers to rely on _ASDisplayView, so this should also not be in the overall .h header. I am actually surprised to see the transaction and transaction group headers here, but we can skip changing that for now, because they are already there.

I only took a very quick look at this, haven't finished reading your description, so let me know if there is a reason we must include the beta headers here… We may need to look at restructuring it a little bit if there is no way around this. But it is important that those methods are not generally available, so maybe we need to add "beta" to their method names or something else.

ekurutepe:
@appleguy We decided to integrate it to our project which uses Carthage. With the current umbrella header, the built framework causes warnings to be thrown when consumers import it as a module. 

Since I'm pretty new to the project and I'm not sure about the purpose of different headers, I made the minimally invasive set of changes to make the framework target build with the correct module headers and get rid of the warnings. 

I actually assumed that the _ headers are not meant to be public the build failed when I removed them from the umbrella header. So it seems like there is some house-keeping work necessary to ensure everything builds fine and with the correct visibility. 

If you can provide me with some pointers on which headers should not be visible publicly, I'm happy to move things around and clean up the headers.

appleguy:
@ekurutepe thank you for providing this additional context. We definitely intend to support Carthage, but the most active maintainers of the framework are not users of it, so we have not yet been as consistent as I would like at cleanly supporting it. Thank you for investigating this and submitting the change.

Now that I understand that, I'm happy to merge this. It would certainly be good to investigate any changes that are necessary to allow the bill to succeed without including any _-prefixed classes in the umbrella header. However, I understand that this Will not help you accomplish your implementation goals in any direct way, so no worries if you don't get around to this or find that only some can be easily eliminated while others require a deeper change that should be attempted later.

appleguy:
@ekurutepe we are supposed to have a project in examples/ that uses Carthage, and all projects in that directory are built as part of the test, but it seems like it has not actually failed when there has been an issue. One of the most valuable changes you could submit would be an integration of the test project you created (rename it something like CarthageBuildTest) to ensure that this does not regress again.

PR #1255, opened 2016-02-21
[ASRelativeSize] Fix function prototype with a different name in header than in implementation.

garrettmoon:


appleguy:
@garrettmoon what is the context here? Is this a change from previous versions, in which case it could break existing apps? Or does it revert an accidental change that was made recently?

garrettmoon:
I don't know about any changes recently, but no one would be able to use this outside the implementation file it is declared in. The function in the implementation file is named what I renamed this to.

So if someone had tried to use the function declared in the header, they would have gotten a linker error.

PR #1256, opened 2016-02-22
[ASDisplayNode] Post-Review Refinements of Recent No-Affinity UIKit/CA Property Bridging

Adlai-Holler:
@appleguy Wow this feels better! Thanks for helping me get the level of quality here up :+1:

### DIFF
--- a/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
+++ b/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
@@ -233,37 +233,53 @@ if (shouldApply) { _layer.layerProperty = (layerValueExpr); } else { ASDisplayNo
 - (void)setFrame:(CGRect)rect
 {
   _bridge_prologue_write;
-  BOOL setFrameDirectly = _flags.synchronous && !_flags.layerBacked;
-  BOOL isMainThread = ASDisplayNodeThreadIsMain();
-  BOOL nodeLoaded = __loaded;
-  if (nodeLoaded && isMainThread && setFrameDirectly) {
-    // For classes like ASTableNode, ASCollectionNode, ASScrollNode and similar - make sure UIView gets setFrame:
-    
-    // Frame is only defined when transform is identity because we explicitly diverge from CALayer behavior and define frame without transform
-#if DEBUG
-    // Checking if the transform is identity is expensive, so disable when unnecessary. We have assertions on in Release, so DEBUG is the only way I know of.
-    ASDisplayNodeAssert(CATransform3DIsIdentity(self.transform), @"-[ASDisplayNode setFrame:] - self.transform must be identity in order to set the frame property.  (From Apple's UIView documentation: If the transform property is not the identity transform, the value of this property is undefined and therefore should be ignored.)");
-#endif
-    _view.frame = rect;
-  } else if (!nodeLoaded || (isMainThread && !setFrameDirectly)) {
-    /**
-     * Sets a new frame to this node by changing its bounds and position. This method can be safely called even if
-     * the transform is a non-identity transform, because bounds and position can be set instead of frame.
-     * This is NOT called for synchronous nodes (wrapping regular views), which may rely on a [UIView setFrame:] call.
-     * A notable example of the latter is UITableView, which won't resize its internal container if only layer bounds are set.
-     */
-    CGRect bounds = CGRectZero;
-    CGPoint position = CGPointZero;
-    ASBoundsAndPositionForFrame(rect, self.bounds.origin, self.anchorPoint, &bounds, &position);
 
-    self.bounds = bounds;
-    self.position = position;
-  } else if (nodeLoaded && !isMainThread) {
-    _ASPendingState *pendingState = ASDisplayNodeGetPendingState(self);
-    if (!pendingState.hasChanges) {
-      [[ASPendingStateController sharedInstance] registerNode:self];
+  // For classes like ASTableNode, ASCollectionNode, ASScrollNode and similar - make sure UIView gets setFrame:
+  struct ASDisplayNodeFlags flags = _flags;
+  BOOL setFrameDirectly = flags.synchronous && !flags.layerBacked;
+
+  BOOL nodeLoaded = __loaded;
+  BOOL isMainThread = ASDisplayNodeThreadIsMain();
+  if (!setFrameDirectly) {
+    BOOL canReadProperties = isMainThread || !nodeLoaded;
+    if (canReadProperties) {
+      // We don't have to set frame directly, and we can read current properties.
+      // Compute a new bounds and position and set them on self.
+      CGRect bounds = CGRectZero;
+      CGPoint position = CGPointZero;
+      ASBoundsAndPositionForFrame(rect, self.bounds.origin, self.anchorPoint, &bounds, &position);
+
+      self.bounds = bounds;
+      self.position = position;
+    } else {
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -235,36 +234,63 @@ - (CGRect)frame
...
-      [ASPendingStateController.sharedInstance registerNode:self];
+      _view.frame = rect;
+    } else {
+      // We do have to set frame directly, but either the node isn't loaded or we're on a non-main thread.
+      // Set the frame on the pending state, and it'll call setFrame: when applied.
+      _ASPendingState *pendingState = ASDisplayNodeGetPendingState(self);
+      if (nodeLoaded && !pendingState.hasChanges) {
+        [[ASPendingStateController sharedInstance] registerNode:self];
+      }
+      pendingState.frame = rect;
```
Adlai-Holler: Here I went with 4 total cases when I might have been able to get away with 3 (combining cases 1 and 3) but I found it damaged readability because it was unclear what possibility space that combined case covered. This way the comments very clearly explain each condition and why we're responding the way we do.

Adlai-Holler:
@appleguy You requested an assertion in ASDisplayNodeShouldApplyBridgedWriteToView that the lock is held. How should I implement that? ASDN::Mutex doesn't seem to have any way to check the lock, or maybe I'm dumb, or both.

Adlai-Holler:
Investigating test failures…

Adlai-Holler:
Huh, a random ASImageDownloader failure and a timeout. Tests are passing locally. Maybe the internet connection on the test machine is bad? @appleguy Mind triggering a rebuild?

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -52,37 +52,36 @@
 /// the property cannot be immediately applied and the node does not already have pending changes.
```
appleguy: Assert the lock is held?  This can get nasty if there is a race with the __loaded optimization

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -235,36 +234,63 @@ - (CGRect)frame
...
   BOOL isMainThread = ASDisplayNodeThreadIsMain();
-  BOOL nodeLoaded = self.nodeLoaded;
-  if (nodeLoaded && isMainThread && setFrameDirectly) {
-    // For classes like ASTableNode, ASCollectionNode, ASScrollNode and similar - make sure UIView gets setFrame:
-    
-    // Frame is only defined when transform is identity because we explicitly diverge from CALayer behavior and define frame without transform
+  if (!setFrameDirectly) {
+    BOOL canReadProperties = isMainThread || !nodeLoaded;
+    if (canReadProperties) {
+      // We don't have to set frame directly, and we can read current properties.
```
appleguy: This is really great.  Dramatic improvement despite some additional verbosity.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -235,36 +234,63 @@ - (CGRect)frame
...
-  } else if (nodeLoaded && !isMainThread) {
-    if (!_pendingViewState.hasChanges) {
-      [ASPendingStateController.sharedInstance registerNode:self];
+      _view.frame = rect;
+    } else {
+      // We do have to set frame directly, but either the node isn't loaded or we're on a non-main thread.
+      // Set the frame on the pending state, and it'll call setFrame: when applied.
+      _ASPendingState *pendingState = ASDisplayNodeGetPendingState(self);
+      if (nodeLoaded && !pendingState.hasChanges) {
+        [[ASPendingStateController sharedInstance] registerNode:self];
```
appleguy: It seems like there is the most minute of race conditions here - scheduling before we have actually set.  I see why, but you might want to read the hasChanges first, set the frame, then register if needed.

Adlai-Holler: Hmm but we're currently holding the node's property lock, which the main thread will have to take at the top of `applyPendingViewState`. Could maybe end up in priority inversion though

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -106,6 +110,30 @@ @implementation _ASPendingState
...
+/**
+ * Apply the state's frame, bounds, and position to layer. This will not
+ * be called on synchronous view-backed nodes which require we directly
+ * call [view setFrame:].
+ *
+ * FIXME: How should we reconcile order-of-operations between setting frame, bounds, position?
+ * Note we can't read bounds and position in the background, so we have to keep the frame
+ * value intact until application time (now).
+ */
+ASDISPLAYNODE_INLINE void ASPendingStateApplyMetricsToLayer(_ASPendingState *state, CALayer *layer) {
```
appleguy: This name seems a bit too general, almost like it would trigger applying the entire pending state.  It is awesome to have this code separated out and shared, however.  Maybe ApplyGeometry ?

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -14,6 +14,10 @@
 #import "ASInternalHelpers.h"
 #import "ASDisplayNodeInternal.h"
 
+#define __shouldSetNeedsDisplay(layer) (flags.needsDisplay \
+  || (flags.setOpaque && opaque != (layer).opaque)\
+  || (flags.setBackgroundColor && !CGColorEqualToColor(backgroundColor, (layer).backgroundColor)))
```
appleguy: We will frequently get to the point of the equality check in evaluating this condition.  it may be worth including a pointer check; Hopefully core graphics does this internally, but it may not. This would probably catch a significant number of typical cases, because of UIColor clearColor, whiteColor, shared default values used by developers, etc.

AsyncDisplayKit/Private/ASPendingStateController.mm
```diff
@@ -98,6 +76,24 @@ - (void)flushNow
...
+ This method is assumed to be called with the lock held.
+ */
+- (void)scheduleFlushIfNeeded
+{
+  if (_flags.pendingFlush) {
+    return;
+  }
+
+  _flags.pendingFlush = YES;
+  dispatch_async(dispatch_get_main_queue(), ^{
```
appleguy: Later, we should consider if there is a more optimal time to run this. I bet you that it would be better to have a runloop observer, strategically times it to occur as close to the transaction commit as possible, specifically after the entire main queue has been drained for that runloop.  This is not essential because any developer relying on this kind of timing is going to experience at least some Visual inconsistency / possibilities of properties applying in different frames, but I think it would be worthwhile to investigate further. The right place to start would probably be with some form of logging facility that allows us to record & number turns of the runloop, and look at the empirical behavior of how these flushes interact with user-triggered inputs, layout passes including a large background remeasure / relayout, etc.

appleguy:
@Adlai-Holler awesome set of improvements! There are a very small number of suggested follow-ups here, but I am impressed to say that our work is essentially done here.

nguyenhuy:
@Adlai-Holler Awesome work on this. Thread affinity is a real pain and thanks to you, it's gone! :dancer: :dancers: :100:

PR #1257, opened 2016-02-22
[ASImageNode:displayDidFinish] Unlock _imageLock Correctly

Adlai-Holler:


AsyncDisplayKit/ASImageNode.mm
```diff
@@ -305,7 +305,7 @@ - (void)displayDidFinish
     
     ASDN::MutexLocker l(_imageLock);
     _displayCompletionBlock = nil;
-    ASDN::MutexLocker u(_imageLock);
```
appleguy: In both of these cases, I would prefer that we use the more common pattern that I recall seeing before -- .lock() and .unlock() calls directly on the _imageLock.  This will also reduce the overhead of these stack allocated objects, which although low, is consistently measurable in hot code like this.

Adlai-Holler:
@appleguy OK I ditched the stack locker objects. Thoughts on the style here? I'd like to use an actual {} scope but I'd have to declare the variables in advance which feels a little cumbersome.

appleguy:
@Adlai-Holler I like the indent - looks great overall!

PR #1261, opened 2016-02-23
ASNetworkImageNode should support a nil cache.

garrettmoon:


appleguy:
@garrettmoon thanks for the very quick turnaround!!

PR #1262, opened 2016-02-23
[Carthage] Build Test project to ensure framework is never left behind by file additions.

ekurutepe:
Integrate a basic hello world project which imports the ASDK framework build by Carthage. 

This project should build without warnings and errors.

See:
https://github.com/facebook/AsyncDisplayKit/pull/1251


ekurutepe:
the travis tests are failing with this: 

```
Building examples/CarthageBuildTest/.
[!] No `Podfile' found in the project directory.
```

¯\_(ツ)_/¯ 

Is it possible to disable `pod install` for the project in this PR?

nguyenhuy:
You may need to mess around with build script ([here](https://github.com/facebook/AsyncDisplayKit/blob/88bfd3e460e57b6c1e7167b2d4d28ac557aec2c8/build.sh#L33)). Thanks for doing this :)

nguyenhuy:
Another approach would be to have a "Life with Carthage" workspace that is somewhat similar to the existing "Life without CocoaPods" one: https://github.com/facebook/AsyncDisplayKit/blob/88bfd3e460e57b6c1e7167b2d4d28ac557aec2c8/build.sh#L50

appleguy:
Restarting the build that failed just to be sure

ekurutepe:
I've update the `.travis.yml` to install carthage from homebrew before build and extended the `build.sh` script to check if we're building with Carthage or CocoaPods. Seems to work on my machine. Let's see if Travis likes it :)

nguyenhuy:
Looks like Travis can't fetch ASDK from GitHub. Can we specify it as a local project in Cartfile? Like this: `git "file:///../.." "master"`? We are doing similarly in other CocoaPods sample projects ([here](https://github.com/facebook/AsyncDisplayKit/blob/master/examples/Kittens/Podfile#L3) for example).

appleguy:
@ekurutepe Yes that would sound correct, we certainly do that for cocoapods and it seems like by far the best approach - can you make that change?  I'm not sure the right format for the cartfile.

ekurutepe:
Hmmm, Cartfiles support using local repositories with `git "file://absolute/path/to/local/repo" "branch"` syntax. 

Do you know how to get the absolute path inside travis @appleguy @nguyenhuy?

nguyenhuy:
I haven't used Carthage actually. My best guess based on the Carthage's documentation is: `git "file:///../.." "master"`

appleguy:
@ekurutepe @erichoracek - hey guys, sorry I am just not going to have a chance to look into this for at least a few weeks. We are just a matter of about 3 weeks from the largest application launch in Pinterest history, and so I have to focus on that before Learning more about Carthage.

If either of you are able to fix this, I would be really thrilled to merge the change. I am very interested in ensuring support for Carthage is both fully integrated and reliable, and can also commit to providing binary builds of future releases, but only if tooling like this can be put in place by the community. Thanks a lot for any help you can provide!

ekurutepe:
:tada: finally! :)

build.sh
```diff
@@ -35,13 +35,35 @@ if [ "$MODE" = "examples" ]; then
...
+              -sdk "$SDK" \
+              -destination "$PLATFORM" \
+              build
+        elif [ -f "${example}/Cartfile" ]; then
+          echo "Using Carthage"
+          local_repo=`pwd`
+          current_branch=`git rev-parse --abbrev-ref HEAD`
+          cd $example
+          
+          echo "git \"file://${local_repo}\" \"${current_branch}\"" > "Cartfile"
```
nguyenhuy: Nifty!

examples/CarthageBuildTest/Cartfile.resolved
```diff
@@ -0,0 +1 @@
+github "facebook/AsyncDisplayKit" "eb6aa28d42f3dd547f048d0a4d5cd4787f0647b0"
```
nguyenhuy: Do we need this file? Looks odd to me.

nguyenhuy:
Great milestone! Thank you @ekurutepe :+1: :dancer: :dancers:

examples/CarthageBuildTest/Cartfile.resolved
ekurutepe: No, we don't need this file for the travis tests. It gets regenerated with `carthage update`

ekurutepe:
@nguyenhuy @appleguy do you think this one is ready to merge?

appleguy:
@ekurutepe what a fantastic surprise for this weekend!  This is IN!

examples/CarthageBuildTest/Cartfile.resolved
nguyenhuy: @ekurutepe Maybe we should add it to gitignore? Not a big deal though.

PR #1263, opened 2016-02-23
Enable animated bounds changes in ASCollectionView

levi:
Talked to @nguyenhuy and it sounds like there wasn't any need for disabling animations during this resize operation. Enabling this fixes a few animation issues we've been seeing in m3 as of late.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -645,7 +645,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     if (_ignoreMaxSizeChange) {
       _ignoreMaxSizeChange = NO;
     } else {
-      [self performBatchAnimated:NO updates:^{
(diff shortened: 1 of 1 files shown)


nguyenhuy:
Man, that's fast :)

nguyenhuy:
LGTM btw

appleguy:
@levi to me it seems important that a) going from zero bounds to regular bounds, which is common, never animates; b ) that the user might need the ability to disable this, and indeed we probably don't want this with for example split screen changes.  Thoughts?

levi:
@appleguy the `animated:YES` flag here is a bit of a red herring. Enabling it doesn't inherently animate the relayout, but prevents ASDK from performing the mass frame calculation of each cell in the data source in the `ASPerformWithoutAnimation` block—which is a bug. When rotating the device or even resizing the app size in split view multitasking, this relayout will be performed within a system animation block and ASDK is currently preventing the cells from animating their layout in that CATransaction.

PR #1265, opened 2016-02-24
[ASTextNode] Stopped trying to cache the currentScaleFactor

rcancro:
Was running into issues where the scale factor would get cleared when setting a new `atributedString` on `ASTextNode`.

I was clearing out `currentScaleFactor` when setting an attributedString into `ASTextNode`. It appears that `_calculateSize` isn't always called when setting a new string into a `ASTextNode`. It can be the case that only `drawInContext:bounds:` is called. With `_currentScaleFactor` cleared out the newly created renderer that calls`drawInContext...` was being created with a scaleFactor of 0.

It could be that the fix could simply be to remove the clearing of `currentScaleFactor` from `setAttributedString`, but clearing the scale factor there seems like the right thing to do. Regardless, leaving scaling all in the renderer (what this PR does) seems like the safer fix to me. It does, however, require an extra run through the font adjuster when enabled.

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -49,9 +49,6 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
     _constrainedSize = constrainedSize;
     _attributes = attributes;
     _sizeIsCalculated = NO;
-    if ([attributes.pointSizeScaleFactors count] > 0) {
-      _currentScaleFactor = attributes.currentScaleFactor;
-    }
   }
   return self;
 }
@@ -141,7 +138,8 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
 - (void)_calculateSize
 {
   [self truncater];
-  if ([_attributes.pointSizeScaleFactors count] > 0) {
+  // if we have no scale factors or an unconstrained width, there is no reason to try to adjust the font size
+  if ([_attributes.pointSizeScaleFactors count] > 0 && isinf(_constrainedSize.width) == NO) {
     _currentScaleFactor = [[self fontSizeAdjuster] scaleFactor];
   }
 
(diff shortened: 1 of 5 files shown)


AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -141,7 +138,8 @@ - (void)setConstrainedSize:(CGSize)constrainedSize
 - (void)_calculateSize
 {
   [self truncater];
-  if ([_attributes.pointSizeScaleFactors count] > 0) {
+  // if we have no scale factors or an unconstrained width, there is no reason to try to adjust the font size
+  if ([_attributes.pointSizeScaleFactors count] > 0 && isinf(_constrainedSize.width) == NO) {
```
appleguy: For both this condition and the one below, we might as well put the scaleFactors count check at the end so we can short circuit evaluation faster - this is particularly true for the second one because it will be more common for _sizeIsCalculated == YES there, and rendering on main is common (this will be enabled by default for visible range of tables / collections as they are presented by tab switch, modal, etc)

appleguy:
@rcancro thanks for the great description. Can you share why the second run through the size adjuster would be required after this? I don't yet have a great understanding of the code, at least not enough to see why that is the case right away.

rcancro:
@appleguy What appeared to be happening to me was that ASTextNode would create a renderer to compute the text size. In the process the renderer would also determine our scale factor. Upon completion, I would save this scale factor away in ASTextNode.  ASTextNode then invalidated (dealloc'ed) the renderer. 

After this `setAttributedString` on ASTextNode was being called, in which I would clear out the saved scale factor. ASTextNode would then create a new renderer with the current scale factor (which was just cleared) to actually draw the text. This renderer was never used to calculate size, so it never updated the scale factor it was created with. By putting the scale computation in both places I ensured that this didn't happen.

TBH, the behavior seems strange to me. Let me describe the situation where I'm seeing this. I have a button that changes text when selected. After I change the `attributedString` on the ASTextNode I never see a text renderer with the new string in `_calculateSize` or `size`. The renderer does have the new string in `drawInContext:bounds:`.

PR #1267, opened 2016-02-24
[ASDisplayNode] Add NS_REQUIRES_SUPER for -layout.

eanagel:
Leaving `[super layout]` out is pretty easy to do (it wasn't required pre-ASLayout) and breaks ASLayout. It's also a big time killer to debug - I can tell you that from experience ;) If you are not using layout this isn't technically required but calling it should be safe. I've also added `[super layout]` to `ASEditableTextNode` to quiet the warning.


appleguy:
@eanagel thanks a lot!  This is far more critical now and that layout specifications are used; earlier versions of the framework really did not depend on this. That further increases the importance of adding it now, because developers adopting layout specifications may have existing code that had forgotten this call.

PR #1268, opened 2016-02-24
[ASTableView] Fix recent regression to ensure scrollViewDidScroll: is forwarded correctly

eanagel:
Initialize `_asyncDelegateImplementsScrollviewDidScroll` so we forward `scrollViewDidScroll:` calls correctly.
@Xinchi


eanagel:
@appleguy Can you take a look at this one when you get a chance? I need this fix to implement virtualization.

appleguy:
@eanagel ouch, thanks for quickly finding this recent regression! You saved it from going out in 1.9.7 in the next day or two...

PR #1270, opened 2016-02-24
[ASMapNode] Remove Live Map in clearFetchedData, Not clearContents

Adlai-Holler:
@appleguy This is pretty critical, since if you have a live map and it leaves the display range but stays in the fetch data range, and you scroll back to it, your map will be gone!

I used ASPerformBlockOnMainThread because I believe `fetchData` and `clearFetchedData` may be called off the main thread. If that's not correct, or some other approach is needed, let me know.


appleguy:
@Adlai-Holler as of today they are never called off main, but I think we should make that API change for 2.0.  cc @maicki, would you be interested in looking at a couple structural upgrades to the range controller that would allow us to reduce main thread overhead (quite significantly, actually) by calling the display / fetchData methods off of main?

aaronschubert0:
@Adlai-Holler Great catch, thanks!

PR #1271, opened 2016-02-25
Example improvements

maicki:
Update examples to use collectionView:nodeBlockForItemAtIndexPath: or tableView:nodeBlockForItemAtIndexPath: to return node blocks in ASCollectionDataSource / ASTableDataSource objects as well as overall code improvements.

### DIFF
--- a/examples/ASCollectionView/Sample/SupplementaryNode.m
+++ b/examples/ASCollectionView/Sample/SupplementaryNode.m
@@ -17,9 +17,11 @@
 
 static CGFloat kInsets = 15.0;
 
-@implementation SupplementaryNode {
-  ASTextNode *_textNode;
-}
+@interface SupplementaryNode ()
+@property (nonatomic, strong) ASTextNode *textNode;
+@end
+
+@implementation SupplementaryNode
 
 - (instancetype)initWithText:(NSString *)text
 {
@@ -37,7 +39,7 @@ static CGFloat kInsets = 15.0;
 {
   ASCenterLayoutSpec *center = [[ASCenterLayoutSpec alloc] init];
   center.centeringOptions = ASCenterLayoutSpecCenteringXY;
-  center.child = _textNode;
+  center.child = self.textNode;
   UIEdgeInsets insets = UIEdgeInsetsMake(kInsets, kInsets, kInsets, kInsets);
   return [ASInsetLayoutSpec insetLayoutSpecWithInsets:insets child:center];
 }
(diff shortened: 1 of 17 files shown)


examples/ASCollectionView/Sample/SupplementaryNode.m
```diff
@@ -17,9 +17,11 @@
 
 static CGFloat kInsets = 15.0;
 
-@implementation SupplementaryNode {
-  ASTextNode *_textNode;
-}
+@interface SupplementaryNode ()
```
appleguy: I believe you can use ASTextCellNode for this.

examples/ASCollectionView/Sample/ViewController.m
```diff
@@ -15,50 +15,48 @@
...
 @end
 
 
 @implementation ViewController
 
-#pragma mark -
-#pragma mark UIViewController.
+- (void)dealloc
+{
+  self.collectionView.asyncDataSource = nil;
```
appleguy: is this actually necessary since these are weak references?  It may actually be, because of the proxy, but not sure.

examples/ASCollectionView/Sample/SupplementaryNode.m
maicki: We could use ASTextCellNode for this, but it is pretty basic in the current state. It defines it's own font size, insets etc. without any way to change that values. Maybe worth it to look into add some more ways to configure ASTextCellNode?

examples/ASCollectionView/Sample/ViewController.m
maicki: I think best practice would be to nil out the asyncDataSource / asyncDelegate as we will nil out the proxy delegate

examples/CollectionViewWithViewControllerCells/Sample/ViewController.m
```diff
@@ -90,16 +92,17 @@ - (void)reloadTapped
...
-  node.layer.borderWidth = 1.0;
-  node.layer.borderColor = [UIColor blackColor].CGColor;
-  
-  return node;
+  UIImage *image = _sections[indexPath.section][indexPath.item];
+  return ^{
+    return [[ASCellNode alloc] initWithViewControllerBlock:^UIViewController *{
+      return [[ImageViewController alloc] initWithImage:image];
+    } didLoadBlock:^(ASDisplayNode * _Nonnull node) {
+      node.layer.borderWidth = 1.0;
```
appleguy: These properties are very inefficient - I believe they cause offscreen rendering (and should be highlighted in yellow).  We should eliminate this code soon even if we don't have an immediate replacement, maybe just remove the border entirely for now.

appleguy:
Awesome, thanks for the broad cleanup! I am a bit worried that many developers will not clear the delegate on the allocation, and so ideally we could get more test coverage running in this configuration, but as far as best practices that still seems like the right advice

Just general principle of the framework, I have tried to prioritize fault tolerance wherever possible, like ensuring that critical behavior does not actually require a call to the superclass even where the API expects that to be done.  -layout is a notable example where we currently require the superclass call, But ideally I would like to see the framework be tolerant against this; fortunately though we are now using NS_REQUIRES_SUPER, which I don't believe existed at the time that the framework was started.

PR #1273, opened 2016-02-25
[tvOS] Add default focus states to ASControlNode & ASImageNode.

aaronschubert0:
Hi @appleguy, apologies for not being as involved lately but have finally gotten some good time for tvOS support. 

Here is the current state of ASDK on tvOS in my opinion (after this PR):
- ASControlNode. The base class has 95% focus engine support. This is basically complete all for the fact that currently I wasn't able to find a way to receive a long press event (gesture recognizer didn't work). So the default behaviour of being able to hold down a control node and cause it to be 'impressed' into the UI isn't supported.

- ASImageNode. This has about 25% focus engine support. This is a lot trickier and would love some of community to help out on this. The built in effects of UIImageView when `.adjustsImageWhenAncestorFocused = YES` are gorgeous and would need to be replicated in order to use it in production. Also sometimes the focus engine effects aren't escaped properly (in the current implementation)

- Since PINRemoteImage is now a subspec of ASDK then we will need to enable tvOS support for that project too, but I doubt that this would be very hard, maybe @garrettmoon has more info on this? Happy to do this piece of work in a new PR.

Once ASImageNode is complete then the next piece of work would be to make it work inside a ASCellNode for use with ASCollection/TableView. That would be mostly it, in my opinion. So not huge amounts of work but just need to perfect certain aspects of ASDK in order for it to be truly viable.

Obviously other nodes will need to be tested when we get to that but once these points are done then I'd be confident with moving ahead. Otherwise I might have to reside to use UICollectionView's with UICollectionViewCells and UIImageView's and only use ASDK for sub views which wouldn't be that great!

cc @levi @nguyenhuy @Adlai-Holler


appleguy:
@aaronschubert0 very cool! I will take a look here in the next few days.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -424,5 +436,75 @@ - (void)cancelTrackingWithEvent:(UIEvent *)touchEvent
...
+    //Not focused
+    [coordinator addCoordinatedAnimations:^{
+      [self setDefaultState];
+    } completion:nil];
+  }
+}
+
+- (void)setFocusedState
+{
+  self.layer.shadowOffset = CGSizeMake(2, 10);
```
appleguy: Can you avoid repeated access to properties, at least all properties used 3 times or more in a method, by creating a local variable?

aaronschubert0: @appleguy Sure thing, is this the preferred way?

garrettmoon:
There's an open PR for tvOS support on PINRemoteImage. I'm looking to land for a 2.1 release.

aaronschubert0:
@garrettmoon Awesome, will track that.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -77,23 +77,35 @@ - (id)init
...
+  
   // As we have no targets yet, we start off with user interaction off. When a target is added, it'll get turned back on.
   self.userInteractionEnabled = NO;
+  
   return self;
 }
 
+- (void)didLoad
+{
+#if TARGET_OS_TV
```
appleguy: Put the gating around the entire method, so we do not even define it / have the call overhead for iOS.

Also, must call super!

AsyncDisplayKit/ASControlNode.h
```diff
@@ -119,7 +122,12 @@ typedef NS_OPTIONS(NSUInteger, ASControlState) {
   @param event The event which triggered these control actions. May be nil.
  */
 - (void)sendActionsForControlEvents:(ASControlNodeEvent)controlEvents withEvent:(nullable UIEvent *)event;
-
+#if TARGET_OS_TV
```
appleguy: Add a space here.  This method name is a bit terse, is there a parallel in Apple's API with similar nomenclature?  I have not yet studied tvOS-specific APIs yet.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -77,23 +77,35 @@ - (id)init
...
+}
+
 #pragma mark - ASDisplayNode Overrides
 - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event
 {
   // If we're not interested in touches, we have nothing to do.
   if (!self.enabled)
     return;
-
+
```
appleguy: All of these space changes seem unnecessary and make the diff a bit less clear.  Keep the ones where indentation did not match the normal 2-spaces, that is helpful - maybe get rid of these.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -424,5 +436,80 @@ - (void)cancelTrackingWithEvent:(UIEvent *)touchEvent
...
+{
+  CALayer *layer = self.layer;
+  layer.shadowOffset = CGSizeZero;
+  layer.shadowColor = [UIColor blackColor].CGColor;
+  layer.shadowRadius = 0;
+  layer.shadowOpacity = 0;
+  layer.shadowPath = [UIBezierPath bezierPathWithRect:self.layer.bounds].CGPath;
+  self.view.transform = CGAffineTransformScale(CGAffineTransformIdentity, 1, 1);
+}
+#endif
```
appleguy: For longer blocks, put // TARGET_OS_TV at the #endif to help clarify what it is ending

AsyncDisplayKit/ASControlNode.mm
aaronschubert0: @appleguy Huh, I was under the assumption Xcode would respect the current spacing when using Ctrl+I. Will revert those space changes.

AsyncDisplayKit/ASControlNode.mm
aaronschubert0: Will do.

AsyncDisplayKit/ASControlNode.mm
aaronschubert0: Great idea, will do that for this and for the future.

AsyncDisplayKit/ASControlNode.h
aaronschubert0: @appleguy Since tvOS handles all it's display states internally, there isn't a similar method. The closest to this would be `setNeedsFocusUpdate` but this implies a different action. Maybe we can make it more descriptive, such as `setDefaultFocusAppearance`

garrettmoon:
tvOS support just landed in PINRemoteImage

AsyncDisplayKit/ASControlNode.h
appleguy: @aaronschubert0 yes, that name sounds much better.  I'm not experienced with tvOS yet, so I can't weigh in, but if that name makes sense to you we should go with it.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -82,9 +82,21 @@ - (id)init
...
   // As we have no targets yet, we start off with user interaction off. When a target is added, it'll get turned back on.
   self.userInteractionEnabled = NO;
+  
   return self;
 }
 
+#if TARGET_OS_TV
+- (void)didLoad
+{
+  //On tvOS all control views, such as buttons, interact with the focus system even if they don't have a target set on them. Here we add our own internal tap gesture to handle this behaviour.
```
appleguy: maybe just say all controls, rather than control views.
Create a second line for the comment rather than long line, and always space after //

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -64,6 +64,11 @@ @implementation ASImageNode
   void (^_displayCompletionBlock)(BOOL canceled);
   ASDN::RecursiveMutex _imageLock;
 
+#if TARGET_OS_TV
+  //tvOS
+  BOOL isDefaultState;
```
appleguy: Good one to update

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -410,6 +415,174 @@ - (void)setImageModificationBlock:(asimagenode_modification_block_t)imageModific
...
+
+  CGSize targetShadowOffset = CGSizeMake(0.0, self.bounds.size.height/8);
+  [layer removeAllAnimations];
+  [CATransaction begin];
+  [CATransaction setCompletionBlock:^{
+    layer.shadowOffset = targetShadowOffset;
+  }];
+  
+  CABasicAnimation *shadowOffsetAnimation = [CABasicAnimation animationWithKeyPath:@"shadowOffset"];
+  shadowOffsetAnimation.toValue = [NSValue valueWithCGSize:targetShadowOffset];
```
appleguy: Ideally abstract this to something like AsyncDisplayKit+tvOS or something.  Not sure what else we could move there but there's quite a bit of animation code that hopefully would be movable.

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -410,6 +415,174 @@ - (void)setImageModificationBlock:(asimagenode_modification_block_t)imageModific
...
+  shadowOpacityAnimation.fillMode = kCAFillModeForwards;
+  shadowOpacityAnimation.timingFunction = [CAMediaTimingFunction functionWithName:@"easeOut"];
+  [layer addAnimation:shadowOpacityAnimation forKey:@"shadowOpacityAnimation"];
+  
+  view.transform = CGAffineTransformScale(CGAffineTransformIdentity, 1.25, 1.25);
+  
+  [CATransaction commit];
+}
+
+- (void)touchesMoved:(NSSet *)touches withEvent:(UIEvent *)event {
```
appleguy: { style

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -410,6 +415,174 @@ - (void)setImageModificationBlock:(asimagenode_modification_block_t)imageModific
...
+  [super touchesMoved:touches withEvent:event];
+
+  if (!isDefaultState) {
+    UIView *view = [self getView];
```
appleguy: what is getView?  Just self.view?

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -410,6 +415,174 @@ - (void)setImageModificationBlock:(asimagenode_modification_block_t)imageModific
...
+
+    UITouch *touch = [touches anyObject];
+    // Get the specific point that was touched
+    // This is quite messy in it's current state so is not ready for production. The reason it is here is for others to contribute and to make it clear what is occuring.
+    // TODO: Clean up, and improve visuals.
+    CGPoint point = [touch locationInView:self.view];
```
appleguy: Is this different than view?

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -410,6 +415,174 @@ - (void)setImageModificationBlock:(asimagenode_modification_block_t)imageModific
...
+      pitch = -15;
+    } else {
+      pitch = (point.y/CGRectGetHeight(self.view.frame))*15;
+    }
+    if (pitch < 0) {
+      topHalf = YES;
+    }
+    
+    if (point.x > CGRectGetWidth(self.view.frame)) {
+      yaw = 10;
```
appleguy: This all is _very_ cool support to have!  We should probably make either ASImageNode+tvOS or find a way to abstract this code out.  Making a folder for tvOS in the project structure and then building +tvOS category files for anything with more than 50 lines of tvOS support seems like a good guideline.

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -410,6 +415,174 @@ - (void)setImageModificationBlock:(asimagenode_modification_block_t)imageModific
...
+  [layer removeAnimationForKey:@"shadowOpacity"];
+  isDefaultState = YES;
+}
+
+- (UIView *)getView
+{
+  UIView *view = self.view;
+  //If we are inside a ASCellNode, then we need to apply our focus effects to the ASCellNode view/layer rather than the ASImageNode view/layer.
+  if (CGSizeEqualToSize(self.view.superview.frame.size, self.view.frame.size) && self.view.superview.superview) {
+    view = self.view.superview.superview;
```
appleguy: Hmm...this isn't the right way to do this.  If we need to walk parents to find a cell, that's one thing.  We can check .hierarchyState for rangeManaged, which means that we're in an ASCellNode.  But what if we're inside another view deep?

There is a Helper function somewhere for the first ancestor of nodeclass or something.

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -410,6 +415,174 @@ - (void)setImageModificationBlock:(asimagenode_modification_block_t)imageModific
...
+  }
+  return view;
+}
+
+- (float)degressToRadians:(float)value
```
appleguy: There is a #define for this provided by the system somewhere.

appleguy:
@aaronschubert0 with the release done, I finally had a chance to review this. I apologize for the delay, as it caught me at the beginning of the longest crunch in some time.

This is very very cool and I am excited to land it soon. There're conflicts that need to be fixed, and I would really appreciate it if you could create some files that are dedicated to tvOS to hide some of the additional complexity added here — mainly because the simplicity and compactness of these key components is important for their readability and ensuring they can be easily reviewed for bugs.

I understand if you won't have time for this soon, but if you do, I am committed to providing a review within a few day turnaround to minimize any further conflicts that come up.

aaronschubert0:
@appleguy Thanks for the thorough review and for that commitment. Will take a look at this at my first chance.

AsyncDisplayKit/ASControlNode.mm.orig
```diff
@@ -0,0 +1,683 @@
...
+#import "ASControlNode.h"
+#import "ASControlNode+Subclasses.h"
+#import "ASThread.h"
+#import "ASDisplayNodeExtras.h"
+#import "ASImageNode.h"
+
+// UIControl allows dragging some distance outside of the control itself during
+// tracking. This value depends on the device idiom (25 or 70 points), so
+// so replicate that effect with the same values here for our own controls.
+#define kASControlNodeExpandedInset (([[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad) ? -25.0f : -70.0f)
```
appleguy: Accidentally swept in the .orig file.  Thanks for the quick resolution of the merge conflicts!  Great to see green tests.  I'm willing to land this before all the feedback is addressed if you're able to do a followup within a few days of landing.  No rush though, we can wait in the PR state if you are busy for a couple weeks.

aaronschubert0: Oops. Thanks @appleguy. I've got time this week to finish this off, will be doing some work today but happy to merge then cleanup.

PR #1274, opened 2016-02-25
[ASMapNode] Promote annotations to a Formal Property

Adlai-Holler:
I decided to make it `nonnull` which incurs a slight performance penalty but it makes for safer code, and it mirrors MKMapView's `annotations` property.

### DIFF
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -43,6 +43,7 @@
   _needsMapReloadOnBoundsChange = YES;
   _liveMap = NO;
   _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASMapNode.mm
```diff
@@ -43,6 +43,7 @@ - (instancetype)init
   _needsMapReloadOnBoundsChange = YES;
   _liveMap = NO;
   _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
+  _annotations = @[];
```
appleguy: Is this necessary, or can we default to nil?

Adlai-Holler: Currently the property is nonnull, see my rationale in the PR's description.

PR #1275, opened 2016-02-25
Fix didCompleteLayoutTransition

nguyenhuy:
In `ASDisplayNode+Beta.h`, the method signature is `didCompleteLayoutTransition`, but in `ASDisplayNode.mm`, the actual implementation uses `didCompleteTransitionLayout`. This causes overriding implementation in my node subclass not being called.


levi:
LGTM! Thanks for finding these!

nguyenhuy:
No problem :)

PR #1279, opened 2016-02-26
[ASDisplayNode] Bridge should trigger display when opacity or bg color change.

appleguy:
A relatively modest regression from the recent thread affinity patch.


appleguy:
cc @garrettmoon, @Adlai-Holler

appleguy:
@Adlai-Holler - if there is a spot in the unit tests you know of that could help test this behavior, sepecifically / especially when a node is already loaded and it's called on main, that would be great to add in the next few weeks.

PR #1280, opened 2016-02-26
Added NSArray+Diffing.m and ASDefaultPlayButton.m to iOS Framework target

davidlawson:
I was getting the following error when including AsyncDisplayKit using Carthage:

```-[__NSArrayM asdk_diffWithArray:insertions:deletions:compareBlock:]: unrecognized selector sent to instance 0x7ff2885cd780```

This pull request adds two missing .m files to the iOS Framework target.


appleguy:
@davidlawson thanks, I am not sure why our build test has not been catching these things? Certainly looking forward to this one land, and hopeful that it will start catching these things in the way that the original "life without cocoapods" was supposed to.

https://github.com/facebook/AsyncDisplayKit/pull/1262

ekurutepe:
That's quite strange. Shouldn't Xcode be throwing incomplete implementation warnings while building the framework? It doesn't…

PR #1281, opened 2016-02-26
[ASVideoNode] AVAsset type cast safety

ejensen:
The current implementation of `setAsset:` does an unsafe cast to `AVURLAsset`. This causes a crash when a AVAsset subclass other than `AVURLAsset` (ex `AVComposition`) is used.

`testVideoNodeDoesNotReplaceAVPlayerItemWhenSameURLIsSet` has also been updated to check if setting different objects with the same URL does not replace the AVPlayerItem.

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -277,7 +277,10 @@
 {
   ASDN::MutexLocker l(_videoLock);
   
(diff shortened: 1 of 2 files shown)


ejensen:
CI test failure looks unrelated to the changeset

appleguy:
Thanks!  Rebuilding.

appleguy:
@ejensen Thank you for also updating the tests!!

PR #1282, opened 2016-02-27
[ASRangeController] API to update current range mode for the next range update

maicki:
Initial implementation for adding an API to update the current range mode of a range controller for the next range update. Currently this is possible through classes that implement the ASRangeControllerUpdateRangeProtocol (ASCollectionNode and ASTableNode).

Furthermore an API to ASViewController was added that allow automatically adjust the current range mode of the containing node if the containing node confirms to the ASRangeControllerUpdateRangeProtocol.

### DIFF
--- a/AsyncDisplayKit/ASCollectionNode.mm
+++ b/AsyncDisplayKit/ASCollectionNode.mm
@@ -10,7 +10,7 @@
 #import "ASCollectionInternal.h"
 #import "ASCollectionViewLayoutFacilitatorProtocol.h"
 #import "ASDisplayNode+Subclasses.h"
-#import "ASRangeController.h"
+#import "ASRangeControllerUpdateRangeProtocol+Beta.h"
 #include <vector>
 
 @interface _ASCollectionPendingState : NSObject
--- a/AsyncDisplayKit/ASCollectionNode.mm
+++ b/AsyncDisplayKit/ASCollectionNode.mm
@@ -219,6 +219,11 @@
   return [self.view.rangeController setTuningParameters:tuningParameters forRangeMode:rangeMode rangeType:rangeType];
 }
 
+- (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode;
+{
+  [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
+}
+
 - (void)reloadDataWithCompletion:(void (^)())completion
 {
   [self.view reloadDataWithCompletion:completion];
--- a/AsyncDisplayKit/ASTableNode.m
+++ b/AsyncDisplayKit/ASTableNode.m
@@ -79,6 +79,10 @@
 
 - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
 {
+    if (!self.isNodeLoaded) {
+        return;
+    }
+    
     [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
 }
 
--- a/AsyncDisplayKit/ASTableNode.m
+++ b/AsyncDisplayKit/ASTableNode.m
@@ -9,7 +9,7 @@
 #import "ASFlowLayoutController.h"
 #import "ASTableViewInternal.h"
 #import "ASDisplayNode+Subclasses.h"
-#import "ASRangeController.h"
+#import "ASRangeControllerUpdateRangeProtocol+Beta.h"
 
 @interface _ASTablePendingState : NSObject
 @property (weak, nonatomic) id <ASTableDelegate>   delegate;
--- a/AsyncDisplayKit/ASTableNode.m
+++ b/AsyncDisplayKit/ASTableNode.m
@@ -77,6 +77,11 @@
   }
 }
 
+- (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
+{
+    [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
+}
+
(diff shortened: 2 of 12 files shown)


AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -130,8 +138,7 @@ - (void)_updateVisibleNodeIndexPaths
   NSMutableOrderedSet<NSIndexPath *> *allIndexPaths = [[NSMutableOrderedSet alloc] initWithSet:visibleIndexPaths];
   
   ASInterfaceState selfInterfaceState = [_dataSource interfaceStateForRangeController:self];
-  ASLayoutRangeMode rangeMode = [ASRangeController rangeModeForInterfaceState:selfInterfaceState
-                                                                 currentRangeMode:_currentRangeMode];
+  ASLayoutRangeMode rangeMode = (_currentRangeMode == ASLayoutRangeModeInvalid) ? ASLayoutRangeModeMinimum : _currentRangeMode;
```
appleguy: would be slightly easier to read as-
rangeMode = _currentRangeMode
if (rangeMode == invalid)
rangeMode = minimum

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -168,7 +175,9 @@ - (void)_updateVisibleNodeIndexPaths
   NSSet<NSIndexPath *> *allCurrentIndexPaths = [[allIndexPaths set] copy];
   [allIndexPaths unionSet:_allPreviousIndexPaths];
   _allPreviousIndexPaths = allCurrentIndexPaths;
-  _currentRangeMode = rangeMode;
+  
+  // Update the current range mode based on interface state
+  _currentRangeMode = [ASRangeController rangeModeForInterfaceState:selfInterfaceState currentRangeMode:_currentRangeMode];
```
appleguy: I realized a flaw with my reasoning, which may require that you bring back the Boolean flag you had earlier. Specifically, although we do want this method to run at least once when the user forces a specific mode, we must also consider that the next time this method runs we may have some relevant new information like a scroll direction. It is important that this information be taken into account on that pass; and that information may not be available at this point, because this occurs potentially much earlier, even before the next update is scheduled...and the thing that schedules the next update may be a scroll event for example.

So, I think it is ultimately important that we query for the automatic range mode at the beginning of the update pass, but actually just skip the query step for the first range update after a mode is manually set.  I apologize for not realizing this when we were discussing earlier today, and asking you to remove the Boolean flag.

AsyncDisplayKit/Details/ASRangeController.h
```diff
@@ -77,6 +82,15 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@protocol ASRangeControllerUpdateRangeProtocol <NSObject>
```
appleguy: This is an interesting idea, but I am a bit concerned about introducing it now, At least this public API. Could you implement this as private API for now, and still have the functionality integration with ASViewController work?  You could put it in a +Beta.h header somewhere.

The reason is that I am planning to add a similar, but larger and differently main protocol to include things like interface state and potentially a few other capabilities that need to be shared between nodes and controllers - or between collection node / collection view and the table versions.  Ultimately I think you have the right intuition here, I am simply interested in ensuring that the API naming remains consistent and stable as we add this functionality in the future

AsyncDisplayKit/ASViewController.h
```diff
@@ -30,6 +30,9 @@ NS_ASSUME_NONNULL_BEGIN
 // Refer to examples/SynchronousConcurrency, AsyncViewController.m
 @property (nonatomic, assign) BOOL neverShowPlaceholders;
 
+/// Automatically adjust the range mode based on view evenets if the containing node confirms to the ASRangeControllerUpdateRangeProtocol
+@property (nonatomic, assign) BOOL automaticallyAdjustRangeModeBasedOnViewEvents;
```
appleguy: For now this should be a private or +Beta.h functionality as we test it, since ASDK public releases go out at a fairly regular cadence (I'm shipping 1.9.7 this weekend :) ).

AsyncDisplayKit/ASCollectionNode.mm
```diff
@@ -219,6 +219,11 @@ - (void)setTuningParameters:(ASRangeTuningParameters)tuningParameters forRangeMo
   return [self.view.rangeController setTuningParameters:tuningParameters forRangeMode:rangeMode rangeType:rangeType];
 }
 
+- (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode;
+{
+  [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
```
appleguy: We should be careful not to trigger view creation...I might say we should entirely skip / ignore the call if someone calls this before self.isNodeLoaded.

appleguy: What do you think, should we skip self.view if not loaded?

AsyncDisplayKit/ASTableNode.m
```diff
@@ -77,6 +77,15 @@ - (void)didLoad
   }
 }
 
+- (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
+{
+    if (!self.isNodeLoaded) {
+        return;
```
appleguy: Maybe instead of early return, just put the actual call in here, and remove the !

PR #1283, opened 2016-02-27
[ASPendingStateController] Relinquish Lock Before Applying Pending States

Adlai-Holler:
@appleguy pointed out a pretty common deadlock case here and suggested this solution. I think it's a winner!


appleguy:
This is actually rare in our app, the only place in the entire beast that has hit it is some weird-ass debug code I added.  This does fix it though, thanks!

PR #1286, opened 2016-02-28
[ASDataController] Restore previous release's behavior to call constrainedSize method serially on one background queue.

appleguy:
New changes to support the nodeBlock... API changed this behavior even for clients that haven't adopted that.  This caused an issue with at least one client, and since there is little value in this being concurrent / it does increase risk for third parties, let's do this.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -123,16 +123,19 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
   }
 }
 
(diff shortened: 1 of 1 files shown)


aaronschubert0:
@appleguy Awesome!! This fixes the issue that I was witnessing over the weekend. Glad you managed to get to the bottom of this so quickly, appreciate the quick turnaround. That's me all good for 1.9.7.

nguyenhuy:
Can you test against master? We have just landed a fix (https://github.com/facebook/AsyncDisplayKit/pull/1288) that is preferred over this one. Thanks!

aaronschubert0:
@nguyenhuy Will give it a shot!

aaronschubert0:
@nguyenhuy Still solid against master!

nguyenhuy:
Awesome. Thanks, @aaronschubert0!

PR #1287, opened 2016-02-29
[ASDisplayNode] Optimize -setNeedsDisplay, deep mutable array copies.

appleguy:
These optimizations are surprisingly impactful.  -setNeedsDisplay being called
for every node triggered cancelAsyncDisplay, locking, and memory management overhead
that is completely avoidable because Core Animation triggers first display automatically.

The mutable array copy optimizations reduced this key cost by over 10x, from 52ms to 5ms
on an iPad Air 2 / A8X with a real-world test case.


appleguy:
screen shot 2016-02-28 at 8 09 50 pm
screen shot 2016-02-28 at 8 04 45 pm

Adlai-Holler:
Wow! This is incredible @appleguy – the new function is just the right tool for the job :zap:

PR #1288, opened 2016-02-29
Introduce ASIndexedNodeContext

nguyenhuy:
- It is a container object that holds enough information to construct and measure a cell node
- All information is gathered on main thread. This allows ASDataController to capture the correct state of its data source before going to background.

Next step is to use ASIndexPath as the primary index type of a context. There will be a lazily loaded NSIndexPath object that is created on demand.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -19,6 +19,8 @@
 #import "ASMainSerialQueue.h"
 #import "ASMultidimensionalArrayUtils.h"
 #import "ASThread.h"
+#import "ASIndexedNodeContext.h"
+#import "ASDataController+Subclasses.h"
 
 //#define LOG(...) NSLog(__VA_ARGS__)
 #define LOG(...)
@@ -110,27 +112,28 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
 
 #pragma mark - Cell Layout
 
-- (void)batchLayoutNodes:(NSArray<ASCellNodeBlock> *)nodes ofKind:(NSString *)kind atIndexPaths:(NSArray<NSIndexPath *> *)indexPaths completion:(void (^)(NSArray<ASCellNode *> *nodes, NSArray<NSIndexPath *> *indexPaths))completionBlock
+- (void)batchLayoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofKind:(NSString *)kind completion:(ASDataControllerCompletionBlock)completionBlock
 {
   NSUInteger blockSize = [[ASDataController class] parallelProcessorCount] * kASDataControllerSizingCountPerProcessor;
+  NSUInteger count = contexts.count;
   
   // Processing in batches
-  for (NSUInteger i = 0; i < indexPaths.count; i += blockSize) {
-    NSRange batchedRange = NSMakeRange(i, MIN(indexPaths.count - i, blockSize));
-    NSArray *batchedIndexPaths = [indexPaths subarrayWithRange:batchedRange];
-    NSArray *batchedNodes = [nodes subarrayWithRange:batchedRange];
-    [self _layoutNodes:batchedNodes ofKind:kind atIndexPaths:batchedIndexPaths completion:completionBlock];
+  for (NSUInteger i = 0; i < count; i += blockSize) {
+    NSRange batchedRange = NSMakeRange(i, MIN(count - i, blockSize));
+    NSArray<ASIndexedNodeContext *> *batchedContexts = [contexts subarrayWithRange:batchedRange];
+    [self _layoutNodesFromContexts:batchedContexts ofKind:kind completion:completionBlock];
   }
 }
 
-- (void)layoutLoadedNodes:(NSArray<ASCellNode *> *)nodes ofKind:(NSString *)kind atIndexPaths:(NSArray<NSIndexPath *> *)indexPaths {
+- (void)layoutLoadedNodes:(NSArray<ASCellNode *> *)nodes fromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofKind:(NSString *)kind
+{
   NSAssert(NSThread.isMainThread, @"Main thread layout must be on the main thread.");
+  ASDisplayNodeAssertTrue(nodes.count == contexts.count);
   
-  [indexPaths enumerateObjectsUsingBlock:^(NSIndexPath *indexPath, NSUInteger idx, __unused BOOL * stop) {
+  [contexts enumerateObjectsUsingBlock:^(ASIndexedNodeContext *context, NSUInteger idx, __unused BOOL * stop) {
     ASCellNode *node = nodes[idx];
     if (node.isNodeLoaded) {
-      ASSizeRange constrainedSize = [self constrainedSizeForNodeOfKind:kind atIndexPath:indexPath];
-      [self _layoutNode:node withConstrainedSize:constrainedSize];
+      [self _layoutNode:node withConstrainedSize:context.constrainedSize];
     }
   }];
 }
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/Details/ASIndexedNodeContext.h
```diff
@@ -0,0 +1,22 @@
...
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import <AsyncDisplayKit/ASDataController.h>
+#import <AsyncDisplayKit/ASDimension.h>
+
+@interface ASIndexedNodeContext : NSObject
+
+@property (nonatomic, readonly, strong) ASCellNodeBlock nodeBlock;
+@property (nonatomic, readonly, assign) NSIndexPath *indexPath;
```
Adlai-Holler: Huh why `assign` here? Wouldn't `strong` make more sense?

nguyenhuy: Nice catch! I attempted to use `ASIndexPath` first with `assign`. Forgot to change to `strong` for `NSIndexPath`.

nguyenhuy: Fix here: https://github.com/facebook/AsyncDisplayKit/pull/1290. Will keep updating that PR to address coming feedbacks.

Adlai-Holler:
@nguyenhuy Thanks for this! Reifying these objects will make it a lot easier to reason about the node-loading process.

nguyenhuy:
@Adlai-Holler Glad you like it!

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -224,9 +224,13 @@ - (void)_layoutNodes:(NSArray *)nodes ofKind:(NSString *)kind atIndexPaths:(NSAr
 
   // Block the _editingTransactionQueue from executing a new edit transaction until layout is done & _editingNodes array is updated.
   dispatch_group_wait(layoutGroup, DISPATCH_TIME_FOREVER);
-  free(nodeBoundSizes);
-
+  
   if (completionBlock) {
+    NSMutableArray *indexPaths = [NSMutableArray arrayWithCapacity:nodeCount];
+    [contexts enumerateObjectsUsingBlock:^(ASIndexedNodeContext * _Nonnull context, NSUInteger idx, BOOL * _Nonnull stop) {
```
Adlai-Holler: It would be a little faster here to use NSFastEnumeration with a for-loop. That way we avoid the dynamic method call and allocating/invoking the dispatch block.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -768,14 +777,20 @@ - (void)insertRowsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDat
     [self accessDataSourceWithBlock:^{
       // sort indexPath to avoid messing up the index when inserting in several batches
       NSArray *sortedIndexPaths = [indexPaths sortedArrayUsingSelector:@selector(compare:)];
-      NSMutableArray *nodes = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
+      NSMutableArray<ASIndexedNodeContext *> *contexts = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
+      
       for (NSUInteger i = 0; i < sortedIndexPaths.count; i++) {
```
Adlai-Holler: Two performance notes:

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -176,13 +178,11 @@ - (void)_layoutNodes:(NSArray *)nodes ofKind:(NSString *)kind atIndexPaths:(NSAr
       dispatch_queue_t queue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
       dispatch_apply(batchCount, queue, ^(size_t i) {
         unsigned long k = j + i;
-        ASCellNodeBlock cellBlock = nodes[k];
-        ASCellNode *node = cellBlock();
+        ASIndexedNodeContext *context = contexts[k];
+        ASCellNodeBlock nodeBlock = context.nodeBlock;
+        ASCellNode *node = nodeBlock();
```
appleguy: We should probably do all this in a single-line expression so that ARC doesn't do anything (I wish its optimizations were less mysterious and I were confident it would not cause retain/release as written - perhaps it can figure this case out?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -417,9 +422,8 @@ - (void)_reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animat
 
     [self accessDataSourceSynchronously:synchronously withBlock:^{
       NSUInteger sectionCount = [_dataSource numberOfSectionsInDataController:self];
-      NSMutableArray *updatedNodes = [NSMutableArray array];
-      NSMutableArray *updatedIndexPaths = [NSMutableArray array];
-      [self _populateFromEntireDataSourceWithMutableNodes:updatedNodes mutableIndexPaths:updatedIndexPaths];
+      NSMutableArray<ASIndexedNodeContext *> *contexts = [NSMutableArray array];
```
appleguy: As a future optimization, we should probably have the method below return the array, so that internally it can create it with a capacity according to the delegate's returned counts.  This may not be practical though, not sure.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -489,29 +493,32 @@ - (void)accessDataSourceSynchronously:(BOOL)synchronously withBlock:(dispatch_bl
...
   [indexSet enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL *stop) {
     NSUInteger rowNum = [_dataSource dataController:self rowsInSection:idx];
     NSIndexPath *sectionIndex = [[NSIndexPath alloc] initWithIndex:idx];
     for (NSUInteger i = 0; i < rowNum; i++) {
       NSIndexPath *indexPath = [sectionIndex indexPathByAddingIndex:i];
-      [indexPaths addObject:indexPath];
-      [nodes addObject:[_dataSource dataController:self nodeBlockAtIndexPath:indexPath]];
+      ASCellNodeBlock nodeBlock = [_dataSource dataController:self nodeBlockAtIndexPath:indexPath];
+      ASSizeRange constrainedSize = [self constrainedSizeForNodeOfKind:ASDataControllerRowNodeKind atIndexPath:indexPath];
+      [contexts addObject:[[ASIndexedNodeContext alloc] initWithNodeBlock:nodeBlock
```
appleguy: This is so great!  Thank you for doing this!

AsyncDisplayKit/Details/ASDataController.mm
Adlai-Holler: Good point. We should also nil out the context's nodeBlock immediately after executing it.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -810,20 +825,24 @@ - (void)reloadRowsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDat
     
     // Reloading requires re-fetching the data.  Load it on the current calling thread, locking the data source.
     [self accessDataSourceWithBlock:^{
-      NSMutableArray *nodes = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
+      NSMutableArray<ASIndexedNodeContext *> *contexts = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
       
       // FIXME: This doesn't currently do anything
```
appleguy: Lol, does anyone know if we can remove this sort?

Adlai-Holler: I was thinking the same thing. We can _definitely_ remove it since it literally doesn't do anything right now except waste time.

As for the deeper question about all the corresponding sorts (whose results are actually used (!)), it's a little tricky. Since in practice we always use `ASChangeSetDataController`, all our changes should already be received in safe order, so all the sorts are redundant (or maybe even dangerous i.e. sorting deletes ascending) but _technically_ ASDataController the superclass was not designed with this guarantee in mind.

nguyenhuy: Yeah, `ASChangeSetDataController` was introduced in a way that it's possible to fallback to `ASDataController` and it can still do its job (reasonably well). With recent work on `ASChangeSetDataController`, I guess it is an essential part now and no one should ever allocate `ASDataController` directly. As soon as we can enforce that, we can get rid of all the redundant sorts. Thoughts?

PR #1292, opened 2016-03-01
[ASCellNode] Update visibility API to support new event types for initial visibility & invisibility (caused by VC transitions, etc)

nguyenhuy:


AsyncDisplayKit/ASCellNode.h
```diff
@@ -14,6 +14,15 @@ NS_ASSUME_NONNULL_BEGIN
 
 typedef NSUInteger ASCellNodeAnimation;
 
+typedef enum : NSUInteger {
+  // Indicates a cell has just became visible
+  ASCellNodeVisibilityEventVisible,
```
appleguy: Maybe DidBecomeVisible?  I wonder if we should drop Visibility from these names, as we may even use ASCellNodeEvent for other things.  Perhaps we should call it ASCellInterfaceEvent or something...or have custom extensions to ASInterfaceState that only apply to ASCellNode.  Dunno, just random ideas, not sure yet what to do hehe.  Directionally though this patch looks like a big improvement.  We do have several clients of this though so I want to make sure to get it right.

chrisdanford: Is there any value to having symmetric `Visible`/`Invisible` events.  I suspect it will be a fairly common need for callers to do something at the point where a cell transitions from invisible to visible.  In the Pinterest case, that action would be "save the current time as the start of a new impression".  Currently, with `EventVisibleRectChanged`, we'll have to store some BOOL `isVisible` state that we use inside of `EventVisibleRectChanged` to detect that transition between invisible and visible.

AsyncDisplayKit/ASCellNode.h
```diff
@@ -14,6 +14,15 @@ NS_ASSUME_NONNULL_BEGIN
...
+  // Indicates a visible cell has a different visible rect. Its position, percent visible or both have changed.
```
appleguy: "Its position (determined by scrollView.contentOffset) has changed while at least 1px remains visible.  It is possible that 100% of the cell is visible both before and after and only its position has changed, or that the position change has resulted in more or less of the cell being visible.  Use CGRectIntersect between cellFrame and scrollView.bounds to get this rectangle"

Issue #1285, opened 2016-02-28
[ASViewController] Should update interfaceState, at least Visible, with VC transitions.


smyrgl:
I'm not sure if this is intentional or not but `interfaceStateDidChange:` is not being called on nodes during VC transitions in `1.9.7`.  I can provide sample code if needed but I figure that it is pretty obvious what is going on.  Is this an intentional thing so as to not muddle VC transitions?

appleguy:
@smyrgl just checking, this is not a change in behavior from previous versions correct?  1.9.7 is so vastly improved over 1.9.6 that I'm hesitant to hold the release any longer, despite the importance of this.

We are actively working on a fix for this, it may not make the next release assuming it is not a regression, that it should be available in master within the next week. This is a tough call, because this is a very important issue to fix, but there are subtleties in doing this correctly (e.g. so that content doesn't flash when returning, and ideally supporting UIViewController rather than just AS*).

cc @levi, @maicki, @nguyenhuy

smyrgl:
@appleguy Actually I had never used this hook in this context before but I doubt very much this is a regression.  This just came up as part of the validation you asked for on 1.9.7.

I tend to agree that this is not a blocker at all but it's probably a very good thing to resolve since visibility hooks on nodes are incredibly useful in table/collection views.

smyrgl:
Plus I'm almost always pulling directly from master anyway so the fact that it won't make the 1.9.7 tag is of little importance to me personally. :)

appleguy:
@smyrgl agreed, we may in fact get this fixed for 1.9.7 as it is a critical issue for Pinterest too.  Thanks a lot for filing it!

appleguy:
Implementation for this is landed. We will continue to refine this behavior, but closing out this task for now. Please reopen if you detect any critical issues, @smyrgl

PR #1293, opened 2016-03-01
[ASInterfaceState] Ensure cell visibility is updated if overall Table or Collection becomes invisible.

maicki:
…llectionView becomes visible or invisible

Adresses issue: #1285

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -532,7 +532,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 
 - (void)collectionView:(UICollectionView *)collectionView willDisplayCell:(_ASCollectionViewCell *)cell forItemAtIndexPath:(NSIndexPath *)indexPath
 {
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1081,6 +1081,9 @@ - (void)didMoveToWindow
   if (!visible && node.inHierarchy) {
     [node __exitHierarchy];
   }
+    
+  // Trigger updating interfaceState for cells in case ASCollectionView becomes visible or invisible
+  [_rangeController visibleNodeIndexPathsDidChangeWithScrollDirection:self.scrollDirection];
```
appleguy: Because we cannot necessarily rely on the call order for the UIkit method, does this always work for nested range controllers? For example, vertically scrolling elements within a pager, or horizontally scrolling elements within a vertical collection?

I'm thinking it could be possible for one of them to re-run the range calculations before it is marked invisible, or something like that.  Note that I think __exitHierarchy may not actually mark it as not visible if the _hierarchyState includes rangeManaged...I am not 100% sure without looking a bit deeper, just have the sense there could be an issue here.

PR #1294, opened 2016-03-01
Account for contentInset when calculating ASScrollDirection

vivqu:
In the Pinterest notifications You tab, the collectionView contentSize height for the first page is always smaller than the bounds height because we need to add a top contentInset so the collectionView doesn't underlap the segmented control node.

screen shot 2016-02-29 at 5 13 28 pm

We need to include the contentInset in the scrollDirection calculation otherwise it returns ASScrollDirectionNone and the collectionView does not fetch any additional pages.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -622,10 +622,10 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 - (ASScrollDirection)nonFlowLayoutScrollableDirections
 {
   ASScrollDirection scrollableDirection = ASScrollDirectionNone;
-  if (self.contentSize.width + self.contentInset.left > self.bounds.size.width) { // Can scroll horizontally.
+  if (self.contentSize.width + self.contentInset.left + self.contentInset.right > self.bounds.size.width) { // Can scroll horizontally.
     scrollableDirection |= ASScrollDirectionHorizontalDirections;
   }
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -622,10 +622,10 @@ - (ASScrollDirection)flowLayoutScrollableDirections:(UICollectionViewFlowLayout
 - (ASScrollDirection)nonFlowLayoutScrollableDirections
 {
   ASScrollDirection scrollableDirection = ASScrollDirectionNone;
-  if (self.contentSize.width > self.bounds.size.width) { // Can scroll horizontally.
+  if (self.contentSize.width + self.contentInset.left > self.bounds.size.width) { // Can scroll horizontally.
```
levi: Include `contentInset.right`

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -622,10 +622,10 @@ - (ASScrollDirection)flowLayoutScrollableDirections:(UICollectionViewFlowLayout
...
     scrollableDirection |= ASScrollDirectionHorizontalDirections;
   }
-  if (self.contentSize.height > self.bounds.size.height) { // Can scroll vertically.
+  if (self.contentSize.height + self.contentInset.top > self.bounds.size.height) { // Can scroll vertically.
```
levi: Include `contentInset.bottom`

appleguy:
@vivqu this really is a great catch.  Thank you - seriously - for diving into the framework to debug, fix, and submit your first PR to this project.  I appreciate it!

PR #1300, opened 2016-03-01
[ASDisplayNode] Don't clear contents when moving between parent nodes.

garrettmoon:


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1227,7 +1227,7 @@ - (void)addSubnode:(ASDisplayNode *)subnode
   if (isMovingEquivalentParents) {
     [subnode __incrementVisibilityNotificationsDisabled];
   }
-  [subnode removeFromSupernode];
+  [subnode removeFromSupernodeMovingBetweenNodes:YES];
```
levi: Will this work when using the `insertSubnode` derivatives?

garrettmoon: Good catch, thanks!

PR #1303, opened 2016-03-02
[ASNetworkImageNode] Dispatch deallocation of UIImage in clearFetchedData: to background thread

maicki:
Destruction of bigger images in ASNetworkImageNode on the main thread can be expensive and can take some time, so we dispatch onto a background queue to actually dealloc.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -19,6 +19,8 @@
 #import "ASPINRemoteImageDownloader.h"
 #endif
 
+static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
+
 @interface ASNetworkImageNode ()
 {
   ASDN::RecursiveMutex _lock;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -248,6 +249,25 @@ - (void)fetchData
...
+{
+  // Destruction of bigger images on the main thread can be expensive
+  // and can take some time, so we dispatch onto a bg queue to
+  // actually dealloc.
+  UIImage *image = self.image;
+  CGSize imageSize = image.size;
+  BOOL shouldReleaseImageOnBackgroundThread = imageSize.width > kMinReleaseImageOnBackgroundSize.width ||
+                                              imageSize.height > kMinReleaseImageOnBackgroundSize.height;
+  if (shouldReleaseImageOnBackgroundThread) {
+    __block UIImage *blockImage = image;
```
appleguy: Technically we should not need __block here, as we don't need to write to the variable but just have it go out of scope.   simply referencing the image with (__unused) or something should be enough - but this is fine too! :)

appleguy:
@maicki it would be great to support this in Multiplex too!

PR #1310, opened 2016-03-03
[ASCollection/TableView] Add -waitUntilAllUpdatesAreCommitted to wait for concurrent measurement to finish

maicki:
The API allows consumer of ASTableView or ASCollectionCiew to block execution of the main thread until all section and row updates are committed.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -462,6 +462,23 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
   }];
 }
 
+- (void)waitUntilAllUpdatesAreCommitted
+{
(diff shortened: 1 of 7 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -462,6 +462,23 @@ - (void)_reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animat
   }];
 }
 
+- (void)waitUntilAllUpdatesAreCommitted
+{
+  ASDisplayNodeAssertMainThread();
```
appleguy: Since we assert main here, probably can remove the assert from the other two places, but not a big deal / could do that with later cleanup (every line counts in huge files with lots of asserts!)

PR #1311, opened 2016-03-03
[ASDisplayNode] Support measureWithSizeRange: on background, and seamless layout spec animation after complete.

nguyenhuy:
- Measurement is performed on the calling thread, which can be a background one.
- Transition context is strictly handled on main.
- ASTextNode uses its bounds to initiate a renderer if needed. It can happen during layout measurement, so it uses a new thread safe bounds.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -658,12 +658,16 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (ASLayout *)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize animated:(BOOL)animated
 {
-  _usesImplicitHierarchyManagement = YES; // Temporary flag for 1.9.x
+  self.usesImplicitHierarchyManagement = YES; // Temporary flag for 1.9.x
+  
   return [self measureWithSizeRange:constrainedSize completion:^{
-    _usesImplicitHierarchyManagement = NO; // Temporary flag for 1.9.x
-    _transitionContext = [[_ASTransitionContext alloc] initWithAnimation:animated delegate:self];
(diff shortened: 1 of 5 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -428,11 +429,13 @@ - (UIView *)_viewToLoad
     ASDisplayNodeAssert(![view isKindOfClass:[_ASDisplayView class]], @"View block should return a synchronously displayed view");
     _viewBlock = nil;
     _viewClass = [view class];
+    _usesDisplayView = [_viewClass isKindOfClass:[_ASDisplayView class]];
```
appleguy: because of the assert above, let's just make this = NO?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -655,12 +658,16 @@ - (ASLayout *)transitionLayoutWithAnimation:(BOOL)animated
...
+  self.usesImplicitHierarchyManagement = YES; // Temporary flag for 1.9.x
+  
   return [self measureWithSizeRange:constrainedSize completion:^{
-    _usesImplicitHierarchyManagement = NO; // Temporary flag for 1.9.x
-    _transitionContext = [[_ASTransitionContext alloc] initWithAnimation:animated delegate:self];
-    [self __implicitlyInsertSubnodes];
-    [self animateLayoutTransition:_transitionContext];
+    self.usesImplicitHierarchyManagement = NO; // Temporary flag for 1.9.x
+
+    ASPerformBlockOnMainThread(^{
```
appleguy: LOL nice!

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: haha, didn't notice that assertion

PR #1313, opened 2016-03-03
Preserve usesImplicitHierarchyManagement when doing animations

levi:
Currently the flag was being reset to NO in all cases. It should ignore turning off the hierarchy management when the user has already enabled it.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -63,7 +63,7 @@ NSString * const ASRenderingEngineDidDisplayNodesScheduledBeforeTimestamp = @"AS
 @synthesize isFinalLayoutable = _isFinalLayoutable;
 @synthesize threadSafeBounds = _threadSafeBounds;
 
-static BOOL usesImplicitHierarchyManagement = FALSE;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -666,10 +666,13 @@ - (ASLayout *)transitionLayoutWithAnimation:(BOOL)animated
 
 - (ASLayout *)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize animated:(BOOL)animated
 {
+  BOOL disableImplicitHierarchyManagement = self.usesImplicitHierarchyManagement == NO;
```
eanagel: Ok this is a bit of a nit, but would it be easier to just capture the original value of usesImplicitHierarchyManagement and restore it unconditionally?

levi: Either way is fine. This flag is going to be removed before 2.0 once I fix some of the ordering issues with the implicit hierarchy operations and also include support for manually managed nodes.

eanagel: Looks great to me, thanks!!!

PR #1317, opened 2016-03-03
[ASDisplayNode] Run loop amortization: Scheduling of view creation / hierarchy assembly across run loops for gaps to handle touches

rahul-malik:
@appleguy : Putting this up for initial review and profiling. We should determine the correct heuristic and/or batch size for each run loop and ultimately if this is a meaningful change in the framework. Feedback welcome as I'm sure this might require a few iterations before we get a solution we feel is an improvement.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -12,6 +12,7 @@
 #import "ASLayoutOptionsPrivate.h"
 
 #import <objc/runtime.h>
+#import <deque>
 
 #import "_ASAsyncTransaction.h"
 #import "_ASAsyncTransactionContainer+Private.h"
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -211,49 +212,72 @@ + (Class)layerClass
...
 + (void)scheduleNodeForRecursiveDisplay:(ASDisplayNode *)node
 {
+
   static ASDN::RecursiveMutex __displaySchedulerLock;
-  static NSMutableArray *__nodesToDisplay = nil;
-  static BOOL __displayScheduled = NO;
-  
-  BOOL scheduleDisplayPassNow = NO;
+  static std::deque<ASDisplayNode *> __renderQueue = std::deque<ASDisplayNode *>();
+  static CFRunLoopObserverRef __mainRunLoopObserver = nil;
```
appleguy: Technical / pedantic nit - initialize CF*Refs to NULL

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -211,49 +212,72 @@ + (Class)layerClass
...
+            __renderQueue.pop_front();
+            displayingNodes.push_back(node);
+          }
+
+          if (__renderQueue.empty()) {
+            isQueueDrained = YES;
+          }
+        }
+
+        CFAbsoluteTime timestamp = isQueueDrained ? CFAbsoluteTimeGetCurrent() : 0;
```
nguyenhuy: Should we get the timestamp while holding `__displaySchedulerLock`? It might be possible that one (or more) nodes are added right after we release the lock, but before we capture the timestamp, no?

rahul-malik: Sounds good to me. Making that change now

AsyncDisplayKit/ASDisplayNode.mm
rahul-malik: will fix.

Issue #1321, opened 2016-03-04
[ASVideoNode] Video node does not support HLS video


gazreese:
I've been playing around with the ASVideoNode functionality for a project and using it to replace the custom functionality I had prototyped. Unfortunately the current ASVideoNode API doesn't support HLS (HTTP live streaming) video.

I've been able to add this functionality in my own fork and I'm raising this issue to capture the change and to see if it can get integrated back into ASDK. I'm no ASDK expert, so you might have to bear with me a little :-)

To add support it would be necessary to initialise the ASVideoNode with either a URL or an AVAsset, analogous to the initialisation of an AVPlayerItem in UIKit. I'm proposing to change the API to make the initialisation with either method explicit:

```
// Mirror the construction of AVPlayerItem with the URL or AVAsset
- (instancetype)initWithURL:(NSURL*)url;
- (instancetype)initWithAsset:(AVAsset*)asset;
```

And in turn to make the current property-based initialisation read-only;

`@property (atomic, strong, readonly) AVAsset *asset;`

appleguy:
@gazreese - I'm excited to see this in the framework! It's great that this is already in a PR and that we are committed to integrating this functionality!

PR #1323, opened 2016-03-04
[ASVideoNode] Add support for HTTP Live Streaming Video

gazreese:
I've been playing around with the ASVideoNode functionality for a project and using it to replace the custom functionality I had prototyped. Unfortunately the current ASVideoNode API doesn't support HLS (HTTP live streaming) video.

I've been able to add this functionality in my own fork and I'm raising this issue to capture the change and to see if it can get integrated back into ASDK. I'm no ASDK expert, so you might have to bear with me a little :-)

To add support it would be necessary to initialise the ASVideoNode with either a URL or an AVAsset, analogous to the initialisation of an AVPlayerItem in UIKit. I'm proposing to change the API to make the initialisation with either method explicit:

```
// Mirror the construction of AVPlayerItem with the URL or AVAsset
- (instancetype)initWithURL:(NSURL*)url;
- (instancetype)initWithAsset:(AVAsset*)asset;
```

And in turn to make the current property-based initialisation read-only;

`@property (atomic, strong, readonly) AVAsset *asset;`

I raised this up in - https://github.com/facebook/AsyncDisplayKit/issues/1321. Wasn't sure what the process was.

Please let me know if you need any changes.


appleguy:
@gazreese this is really fantastic, thanks!  I would love to integrate this kind of functionality.  It may take me a bit to review this, but rest assured the community would be thrilled to have this capability in the master branch.

gazreese:
A code review would be fantastic, thanks!

I'll fix up the tests on Monday. I searched for ASVideoNode in the project and nothing came up, so I assumed there were no tests.

aaronschubert0:
@gazreese Awesome, this will be very useful for us very soon.

gazreese:
The fails don't seem to be anything to do with the code committed, so it would be good to run these again on a working master branch.

appleguy:
@gazreese finally circling back to this. I would like to get it in soon - and in fact we will probably use this in Pinterest :)

Could you replace your changes on the latest? Apologies for the uncharacteristic delay, this change came in at a very intense time for the framework and was just a bit larger than I have had an opportunity to review in detail.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -15,7 +15,7 @@
 // in an issue on GitHub: https://github.com/facebook/AsyncDisplayKit/issues
 
 @interface ASVideoNode : ASControlNode
-@property (atomic, strong, readwrite) AVAsset *asset;
+@property (atomic, strong, readonly) AVAsset *asset;
```
appleguy: @gazreese this is technically a breaking API change. It may be the right thing to do, and now is the time to change it if we are going to. What was your thinking around this part of the change, And do you feel it is important or beneficial?

cc @lappp9, @Xinchi thoughts?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -41,68 +42,106 @@ @interface ASVideoNode ()
...
+    playerLayer.videoGravity = [self gravity];
+    return playerLayer;
+  }];
   
-  if ([[change objectForKey:@"new"] integerValue] == AVPlayerItemStatusFailed) {
-    
+  return playerNode;
+}
+
+- (void) constructCurrentPlayerItemFromInitData {
```
appleguy: Minor, but please ensure all of your methods have no space after the ), and have the { on the next line (see others in the file / other files for reference)

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -126,29 +165,20 @@ - (void)layout
...
-      imageGenerator.appliesPreferredTrackTransform = YES;
-      [imageGenerator generateCGImagesAsynchronouslyForTimes:@[[NSValue valueWithCMTime:CMTimeMake(0, 1)]] completionHandler:^(CMTime requestedTime, CGImageRef  _Nullable image, CMTime actualTime, AVAssetImageGeneratorResult result, NSError * _Nullable error) {
+  // Construct the preview image early on to avoid multiple threads trying to set it
+  if (!_placeholderImageNode)
+    _placeholderImageNode = [[ASImageNode alloc] init];
+  
+  dispatch_async(_previewQueue, ^{
+    AVAssetImageGenerator *imageGenerator = [[AVAssetImageGenerator alloc] initWithAsset:_asset];
+    imageGenerator.appliesPreferredTrackTransform = YES;
+    [imageGenerator generateCGImagesAsynchronouslyForTimes:@[[NSValue valueWithCMTime:CMTimeMake(0, 1)]] completionHandler:^(CMTime requestedTime, CGImageRef  _Nullable image, CMTime actualTime, AVAssetImageGeneratorResult result, NSError * _Nullable error) {
```
appleguy: Maybe break out the NSArray *times =

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -126,29 +165,20 @@ - (void)layout
...
-      return playerLayer;
-    }];
-    
-    [self insertSubnode:_playerNode atIndex:0];
-  } else {
-    dispatch_async(_previewQueue, ^{
-      AVAssetImageGenerator *imageGenerator = [[AVAssetImageGenerator alloc] initWithAsset:_asset];
-      imageGenerator.appliesPreferredTrackTransform = YES;
-      [imageGenerator generateCGImagesAsynchronouslyForTimes:@[[NSValue valueWithCMTime:CMTimeMake(0, 1)]] completionHandler:^(CMTime requestedTime, CGImageRef  _Nullable image, CMTime actualTime, AVAssetImageGeneratorResult result, NSError * _Nullable error) {
+  // Construct the preview image early on to avoid multiple threads trying to set it
```
appleguy: I'm a bit worried about this comment.  Is this method supposed to only be called on the main thread?  If so, let's just add an assertion for that and the comment isn't necessary.  If not main-restricted, can't we just use the lock?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -168,9 +198,54 @@ - (void)didLoad
         dispatch_async(dispatch_get_main_queue(), ^{
           _placeholderImageNode.frame = self.bounds;
           [self insertSubnode:_placeholderImageNode atIndex:0];
+          // [self setNeedsLayout];
```
appleguy: Is this needed?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -168,9 +198,54 @@ - (void)didLoad
...
-      }];
-    });
+      }
+    }];
+  });
+}
+
+- (void)interfaceStateDidChange:(ASInterfaceState)newState fromState:(ASInterfaceState)oldState
+{
+  if (!(newState & ASInterfaceStateVisible)) {
```
appleguy: For readability, make two local BOOLs - nowVisible and wasVisible.

Not sure if they link properly in this file as there are some weird problems with them, but see if you can use ASInterfaceStateIncludesVisible()

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -168,9 +198,54 @@ - (void)didLoad
...
+          _currentPlayerItem.tracks[0].assetTrack &&
+          _currentPlayerItem.tracks[0].assetTrack.asset) {
+        _asset = _currentPlayerItem.tracks[0].assetTrack.asset;
+        [self setPlaceholderImagefromAsset:_asset];
+        [self setNeedsLayout];
+      }
+    }
+  }
+  
+  if ([[change objectForKey:@"new"] integerValue] == AVPlayerItemStatusFailed) {
```
appleguy: I'm not really sure what the @"new" key is for, but can we cache this in a local variable at the top rather than accessing the dictionary and unpacking the integer twice?

AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -42,103 +43,151 @@ @implementation ASVideoNodeTests
 
 - (void)setUp
 {
-  _videoNode = [[ASVideoNode alloc] init];
+  // _videoNode = [[ASVideoNode alloc] init];
```
appleguy: Probably need this for the tests to work well :)

AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -42,103 +43,151 @@ @implementation ASVideoNodeTests
...
-  
-  XCTAssertEqualObjects(item, secondItem);
+- (void)testOnPlayIfVideoIsNotReadyInitializeSpinnerAndAddAsSubnode
+{
+  _videoNode = [[ASVideoNode alloc] initWithAsset:_firstAsset];
+  [self doOnPlayIfVideoIsNotReadyInitializeSpinnerAndAddAsSubnodeWithUrl];
 }
 
-- (void)testSpinnerDefaultsToNil
+- (void)testOnPlayIfVideoIsNotReadyInitializeSpinnerAndAddAsSubnodeWithUrl
```
appleguy: Thanks for taking a look at the tests and making them more descriptive!  If we don't want the init in the method above, then just delete that line.

appleguy:
@gazreese it looks like this is almost ready to merge. My main concern at this point is the breaking API change, so let me know your intention with this and whether it would be appropriate to continue to support a setter method for the asset even if the init methods remain.

It does seem possible to me that a user may want to create a video node and display a placeholder using a traditional image, potentially before enough information has learned from the network to actually create the video asset. This may be an edge case, but it is not clear to me that we should avoid supporting it in the API. That said, there are other positive aspects to change such as not requiring support for the various state updates that have to occur if the asset were to change. We should decide what the appropriate API functionality is based on the total set of reasonable use cases, and then worry about the implementation.

If you could also rebase this one more time, I will push too have it merged soon so that further re-bases are unlikely to be required.

gazreese:
Hi!

I'll be honest, I've been pulled onto another project (hopefully temporarily) and I haven't had chance to look into this just yet.

Regarding the API change with the init, I was basically trying to enforce an either/or for the construction of the class. The user should really be using either a URL or an AVAsset. To construct with both I think would be invalid. I could assert for this but it seems more obvious in the API this way. It's also analogous to the initialisation of an AVPlayerItem in UIKit. I'm more than happy to go with whatever you prefer however.

As for having a placeholder image, this was basically my next task to look at. We're very sensitive to bandwidth costs in our application, and generating placeholders from the video data seemed to me to be a little wasteful. Whether this would be via the AVAsset or just passing in a node I'm not entirely sure, I've not had a look into it yet.

As for rebasing the changes onto the mainline, I did this onto my local a few weeks ago and needed to do a few minor changes that I don't think I've pushed yet. I'll rebase again and have a look at your other comments (which all look valid by the way).

Many thanks for the code review. I'll try to get onto this next week.

Cheers
Gareth

appleguy:
@gazreese thanks for the super quick reply. It sounds like you're well aware, but just a heads-up, make sure you set the tuning parameters for any a table or collection views that you use to reduce the "fetch data" range if your application is particularly sensitive to Data consumption.

For the time being, let's keep the writable property and add a simple assertion if you can Think of the right spot to do this. I think the main reason is that unlike the underlying player item, which there is no use case to desire creating before something like a URL is available, there is the use case for a developer to create a video node potentially before they have a reference to the underlying video. This is a rare use case, and the most important concern at this point is that apps would need to change how they use video nodes to update the framework after this change otherwise.

Let me know if you have any further thoughts or questions, and no worries if you are not able to finish this work. Just leave a comment here to that effect if things get busy for you.

gazreese:
Sounds good, and many thanks for the help. I'm hoping to get a day on it this week, so I'll keep you posted.

AsyncDisplayKit/ASVideoNode.h
gazreese: Done, gone back to the property init method and already had asserts to ensure either URL or AVAsset construction. Seems to work fine.

AsyncDisplayKit/ASVideoNode.mm
gazreese: Used the lock and it appears to be working fine

AsyncDisplayKit/ASVideoNode.mm
gazreese: Done

AsyncDisplayKit/ASVideoNode.mm
gazreese: Nope - removed :-)

AsyncDisplayKit/ASVideoNode.mm
gazreese: Done, looks miles better, cheers.

AsyncDisplayKit/ASVideoNode.mm
gazreese: This was in the original code and to be honest with you I wasn't sure what the score was either. I moved it over to checking that the object is the player item and that we've got a status change. The status can then be taken directly from the player item.

`  if (object == _currentPlayerItem && [keyPath isEqualToString:@"status"]) {
    if (_currentPlayerItem.status == AVPlayerItemStatusReadyToPlay) {`

AsyncDisplayKitTests/ASVideoNodeTests.m
gazreese: I agree, done

gazreese:
@appleguy I've merged with the master and should have all of your comments addressed now. If you could please take a look that would be really helpful.

Review comments are in-line here: https://github.com/facebook/AsyncDisplayKit/pull/1323/files/f7ff8b4173808166d82b6cc5517e5499355ddf29#r58163617

Cheers!
Gareth

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -41,70 +42,81 @@ @interface ASVideoNode ()
...
 @implementation ASVideoNode
 
+//TODO: Have a bash at supplying a preview image node for use with HLS videos as we can't have a priview with those
+
+
+#pragma mark - Construction and Layout
+
 - (instancetype)init
 {
-  if (!(self = [super init])) {
```
yunnan-wu: why deleting these? Should always call super init first.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -34,6 +37,8 @@
 
 @property (atomic, weak, readwrite) id<ASVideoNodeDelegate> delegate;
 
+- (instancetype)init;
```
yunnan-wu: delete. There is always an init method.

appleguy:
Very cool! I plan to merge this tomorrow. If you have a chance to fix the initializer call, that would be great, otherwise I will do it in a separate change after merging miss so that there are no conflicts.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -18,7 +18,10 @@
 //    in an issue on GitHub: https://github.com/facebook/AsyncDisplayKit/issues
 
 @interface ASVideoNode : ASControlNode
+
+@property (atomic, strong, readwrite) NSURL *url;
```
appleguy: Because it's so easy to create an asset with a URL, I'm still not confident we should provide this.  If the user sets an asset, then accesses this URL, it should return something - but what if the asset is local?  Setting both, etc.  For now, I'd like to remove it in a followup, and we can discuss the merits of adding it in continued iteration + make sure it is fully integrated for all the edge cases (and a header comment here about expected usage etc)

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -41,70 +42,81 @@ @interface ASVideoNode ()
...
-    return nil;
-  }
-    
-  _previewQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0);
+	_previewQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0);
```
appleguy: This seems indented too far.  Def. need super init.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -41,70 +42,81 @@ @interface ASVideoNode ()
...
-      }
-      [(UIActivityIndicatorView *)_spinner.view stopAnimating];
-      [_spinner removeFromSupernode];
-    }
-  } else {
-    if (_shouldBePlaying) {
-      [self play];
-    }
-  }
+  ASDisplayNodeAssertNotSupported();
```
appleguy: Very good.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -41,70 +42,81 @@ @interface ASVideoNode ()
...
+    playerLayer.videoGravity = [self gravity];
+    return playerLayer;
+  }];
   
-  if ([change[@"new"] integerValue] == AVPlayerItemStatusFailed) {
-    
+  return playerNode;
+}
+
+- (void)constructCurrentPlayerItemFromInitData
```
appleguy: This can be called multiple times.  We should probably skip the work if _currentPlayerItem exists; should not replace it.  Maybe check if its asset is _asset.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -128,29 +140,22 @@ - (void)layout
...
-        _player = [AVPlayer playerWithPlayerItem:[[AVPlayerItem alloc] initWithAsset:_asset]];
-        _player.muted = _muted;
-      }
-      playerLayer.player = _player;
-      playerLayer.videoGravity = [self gravity];
-      return playerLayer;
-    }];
+  ASDN::MutexLocker l(_videoLock);
+  if (!_placeholderImageNode)
+    _placeholderImageNode = [[ASImageNode alloc] init];
```
appleguy: It seems to me like this class should just subclass ASImageNode.  There's little cost to do so, and it's still an ASControlNode.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -128,29 +140,22 @@ - (void)layout
...
-      imageGenerator.appliesPreferredTrackTransform = YES;
-      [imageGenerator generateCGImagesAsynchronouslyForTimes:@[[NSValue valueWithCMTime:CMTimeMake(0, 1)]] completionHandler:^(CMTime requestedTime, CGImageRef  _Nullable image, CMTime actualTime, AVAssetImageGeneratorResult result, NSError * _Nullable error) {
+    [imageGenerator generateCGImagesAsynchronouslyForTimes:times completionHandler:^(CMTime requestedTime, CGImageRef  _Nullable image, CMTime actualTime, AVAssetImageGeneratorResult result, NSError * _Nullable error) {
+      
+      // Unfortunately it's not possible to generate a preview image for an HTTP live stream asset, so we'll give up here
+      // http://stackoverflow.com/questions/32112205/m3u8-file-avassetimagegenerator-error
+      if (image && _placeholderImageNode.image == nil) {
         UIImage *theImage = [UIImage imageWithCGImage:image];
         
         _placeholderImageNode = [[ASImageNode alloc] init];
```
appleguy: Wait, we should remove the creation call above then.  This seems like a better place to do it.  Should probably also check if it already exists.  Again though using self.image would be better.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -171,8 +176,58 @@ - (void)didLoad
...
+      }
+    }];
+  });
+}
+
+- (void)interfaceStateDidChange:(ASInterfaceState)newState fromState:(ASInterfaceState)oldState
+{
+  [super interfaceStateDidChange:newState fromState:oldState];
+	
+	BOOL nowVisible = ASInterfaceStateIncludesVisible(newState);
```
appleguy: Too much indentation

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -171,8 +176,58 @@ - (void)didLoad
...
+	BOOL wasVisible = ASInterfaceStateIncludesVisible(oldState);
+
+  if (!nowVisible) {
+    if (wasVisible) {
+      if (_shouldBePlaying) {
+        [self pause];
+        _shouldBePlaying = YES;
```
appleguy: This is not necessary, unless the condition is supposed to be !_shouldBePlaying?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -302,6 +353,27 @@ - (AVAsset *)asset
   return _asset;
 }
 
+- (void)setUrl:(NSURL *)url
+{
+	ASDN::MutexLocker l(_videoLock);
```
appleguy: Too much indentation

appleguy:
@gazreese thanks for your work on this!  Will you have time in the next week to do some of the cleanups mentioned here?  There is a lot of active work going on in this file, at least four separate developers submitting changes, so it would be fantastic if you can join our slack room and coordinate live if you have a chance to embark on further improvements so that others can be aware and minimize conflicts.

Email asyncdisplaykit@gmail.com to get a slack invite :)

aaronschubert0:
🎉

AsyncDisplayKit/ASVideoNode.mm
gazreese: @appleguy I couldn't find this one in the master branch, so I assume it's sorted

AsyncDisplayKit/ASVideoNode.mm
gazreese: @appleguy looking at the code I think the only time this could be called multiple times would be through the -play method. it's already guarded in here anyway so I think we're fine. 

Happy to look into this further if you think it's still an issue.

AsyncDisplayKit/ASVideoNode.mm
gazreese: @appleguy Subclassing is going to be a bit beyond my ASDK knowledge if I'm honest! 

I can see where you're coming from, but looking at the external interface of ASImageNode I'm not sure all of the aspects would make sense. For instance, we'd have to implement cropRect. Is display completion relevant? Could we implement tint colours anyway?

Though I can see the benefits of code reuse I'm not sure the external 'is-a ASImageNode' relationship would make complete sense to the user of the class.

AsyncDisplayKit/ASVideoNode.mm
gazreese: @appleguy Now checks for compatible assets at the beginning and drops out if we can't continue:

```
    // Skip the asset image generation if we don't have any tracks available that are capable of supporting it
    NSArray<AVAssetTrack *>* visualAssetArray = [_asset tracksWithMediaCharacteristic:AVMediaCharacteristicVisual];
    if (!visualAssetArray || visualAssetArray.count == 0)
      return;
```

AsyncDisplayKit/ASVideoNode.mm
gazreese: @appleguy This looked fixed in the master branch

AsyncDisplayKit/ASVideoNode.mm
gazreese: @appleguy This isn't one of my additions, but by the looks of things it's intended to pause and play videos that are set to 'auto play' as they come in and out of view. 

Running this with the debugger on the VideoTableView example this does appear to be working as expected, so it doesn't seem to be a problem.

The `_shouldBePlaying = YES` is redundant however and I've removed this.

AsyncDisplayKit/ASVideoNode.mm
gazreese: Not there in master

PR #1327, opened 2016-03-05
[ASCollectionView] scrollableDirections must account for alwaysBounce(Vertical | Horizontal)

samhsiung:
If a UIScrollVIew has alwaysBounce(Vertical|Horizontal) = YES then it will always be scrollable in that direction.  Adding these additional checks will allow us to handle cases where an ASCV without a flow layout has a data source whose first page load doesn't expand contentSize beyond bounds in the desired scroll direction. When the user scrolls in the intended direction, batch fetching will be triggered if there is more content to load.


appleguy:
Great improvement!  @samhsiung should make sure the ASBatchContext is triggering even without scrolling in a scenario like this, though it is certainly good to ensure the user can force a load.

Ideally we would not re-access bounds, contentSize, contentInset properties here - but this is a micro-optimization and not essential.

PR #1328, opened 2016-03-05
[ASCollectionView] Remove the dead code for section inset behavior & mark it as deprecated

lkzhao:
Remove section inset delegate for now as it is doing nothing. Or we can deprecate this. 
But it is good to have support for different section inset for different sections. I will explore on a correct version of the sectionInset implementation this weekend.


levi:
I know we were using this at the NFL. Might be better to wrap the API in a deprecation flag and we remove it when we make the 2.0 jump.

appleguy:
@levi I am also a bit concerned about this, but in detailed discussions with Luke, he reminded me that putting this behavior directly into the constrainedSizeForItem: method at the app level is very straightforward and more likely to result in predictable behavior.

This diff doesn't break the feature - I prior one removed it, this one just flags the fact that it will not work anymore.  I would love to have this feature if we could ensure that it properly only applies the inset to the cells that need it.

PR #1330, opened 2016-03-06
[ASCollectionView] Debug Feature: ASTextNode overlay with # of nodes, views, layers in ASCollectionView

hannahmbanana:
AsyncDisplayKit+Debug:
1. A debug feature for ASCollectionView & ASTableView to visualize the number of nodes (N), views (V) and layers (L) in the collectionView / tableView
- The counts are incremented in the willDisplayCell:foritemAtIndexPath: method for each subnode found in the ASCellNode.
The counts are decremented in the didEndDisplayingCell:forItemAtIndexPath: method for each subnode found in the ASCellNode.
- Note: The debug overlay ASTextNode will only display the count for the number of views in the ASCollectionView / ASTableView if the count is not equal to the count of the number of nodes in the ASCollectionView / ASTableView.
- Note: You must call [ASRangeController setHierarchyCountDebugEnabled:YES] before initializing an
 ASCollectionView / ASTableView you want to debug.
- Note: The layout is currently not configurable. The ASTextNode debug overlay is positioned in the
upper right hand corner. I'll try to follow up with layout that detects nested ASRangeControllers and
makes sure that their debug labels do not overlap.
- Future improvement ideas: NSAttributedString colorizing? Per cell counts?
2. Relocated the ASControlNode debug feature public method into AsyncDisplayKit+Debug as well.


AsyncDisplayKit/ASCollectionView.h
```diff
@@ -336,6 +336,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic) BOOL zeroContentInsets;
 
+/**
+ * Class method to enable a visualization overlay of the number of cells and views in the ASCollectionView.
+ * For app debugging purposes only.
+ *
+ * @param enabled Specify YES to make this debug feature enabled when messaging the ASControlNode class.
+ */
++ (void)setEnableDebugCounts:(BOOL)enable;
```
appleguy: This is fine for now, but soon we should move these kind of features to category definitions in a new file, something like AsyncDisplayKit+Debug.h

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -123,7 +124,15 @@ @interface ASCollectionNode ()
 - (instancetype)_initWithCollectionView:(ASCollectionView *)collectionView;
 @end
 
+static BOOL _enableGlobalDebugCounts = NO;
```
appleguy: In general we use __ (double underbar) for static variable names

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -123,7 +124,15 @@ @interface ASCollectionNode ()
...
+
 @implementation ASCollectionView
+{
+  ASTextNode *_debugCountsTextNode;
+  NSUInteger _globalDebugNodeCount;
```
appleguy: global is not the right term for these -- though the __shouldDisplayDebugCounts variable is truly global, these are per-instance so we should probably just name them _debugNodeCount, _debugViewCount, etc

Let's actually try to eliminate these if possible...

hannahmbanana:
Closing this PR & adding it as an issue it to Ship (ASDK's new feature tracking software) until I have more time to clean it up.

PR #1338, opened 2016-03-07
[ASTextNode] Fix tapping on truncation token for 3D Touch phones

tomizimobile:
In ASTextNode, bail from touchesMoved:withEvent: if the previousLocationInView is the same as the locationInView. On 3D Touch enabled phones, this gets fired with chages in force, and will usually get fired immediately after touchedBegan:withEvent, causing touches to usually be cancelled

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -891,15 +891,16 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 {
   [super touchesMoved:touches withEvent:event];
 
+  UITouch *touch = [touches anyObject];
+  CGPoint locationInView = [touch locationInView:self.view];
   // on 3D Touch enabled phones, this gets fired with changes in force, and usually will get fired immediately after touchesBegan:withEvent:
-  if (CGPointEqualToPoint([[touches anyObject] previousLocationInView:self.view], [[touches anyObject] locationInView:self.view]))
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -891,6 +891,10 @@ - (void)touchesMoved:(NSSet *)touches withEvent:(UIEvent *)event
 {
   [super touchesMoved:touches withEvent:event];
 
+  // on 3D Touch enabled phones, this gets fired with changes in force, and usually will get fired immediately after touchesBegan:withEvent:
+  if (CGPointEqualToPoint([[touches anyObject] previousLocationInView:self.view], [[touches anyObject] locationInView:self.view]))
```
appleguy: @tomizimobile please create a local variable for the touch, as well as a local variable for the location in view. Of course it is fine to avoid caching the previous location, since it is not used elsewhere.

This is important because if more than one touch is ever made available to this method, the set may return a different object when calling any object. Also, these methods are not free, and so it is worth the readability and performance advantages of having the local variables.

tomizimobile: @appleguy Whoops, obvious oversight, thanks for pointing out.

appleguy:
@tomizimobile thanks, this looks great!  I swear I thought we had fixed something in here for 3D touch - maybe it was a more serious issue, but didn't quite fully address it.  Glad to have this detail handled.

PR #1343, opened 2016-03-08
[ASTextNode] Improvements to the implementation of the font scale adjuster.

rcancro:
1) cache the scale in the font adjuster. The adjuster will be dealloc'ed when the renderer that owns it is dealloc'ed (or invalidated). Until that time we can trust the scale that the adjuster has computed.
2) When measuring line count, make sure that we are not bounding the height of the container's size. This will cause the wrong number of lines to be returned.
3) Instead of setting the ascender/descender on an ASTextNode when an attributed string is added, wait until after the renderer calculates size. This way, if there is any need to scale the font to fit we can apply that scale to the ascender/descender.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -337,7 +337,18 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 
   [self setNeedsDisplay];
   
-  return [[self _renderer] size];
+  CGSize size = [[self _renderer] size];
+  if (self.attributedString.length > 0) {
+    CGFloat screenScale = ASScreenScale();
+    self.ascender = round([[_attributedString attribute:NSFontAttributeName atIndex:0 effectiveRange:NULL] ascender] * screenScale)/screenScale;
+    self.descender = round([[_attributedString attribute:NSFontAttributeName atIndex:_attributedString.length - 1 effectiveRange:NULL] descender] * screenScale)/screenScale;
+    if (_renderer.currentScaleFactor > 0 && _renderer.currentScaleFactor < 1.0) {
+      // while not perfect, this is a good estimate of what the ascender of the scaled font will be.
+      self.ascender *= _renderer.currentScaleFactor;
+      self.descender *= _renderer.currentScaleFactor;
+    }
+  }
+  return size;
 }
 
 #pragma mark - Modifying User Text
@@ -374,12 +385,6 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
     self.isAccessibilityElement = NO;
   } else {
     self.isAccessibilityElement = YES;
-  }
-
-  if (attributedString.length > 0) {
-    CGFloat screenScale = ASScreenScale();
-    self.ascender = round([[attributedString attribute:NSFontAttributeName atIndex:0 effectiveRange:NULL] ascender] * screenScale)/screenScale;
-    self.descender = round([[attributedString attribute:NSFontAttributeName atIndex:attributedString.length - 1 effectiveRange:NULL] descender] * screenScale)/screenScale;
   }
 }
 
--- a/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
@@ -20,6 +20,8 @@
   __weak ASTextKitContext *_context;
   ASTextKitAttributes _attributes;
   std::mutex _textKitMutex;
+  BOOL _measured;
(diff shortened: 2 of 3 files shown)


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -337,7 +337,18 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 
   [self setNeedsDisplay];
   
-  return [[self _renderer] size];
+  CGSize size = [[self _renderer] size];
+  if (self.attributedString.length > 0) {
+    CGFloat screenScale = ASScreenScale();
+    self.ascender = round([[_attributedString attribute:NSFontAttributeName atIndex:0 effectiveRange:NULL] ascender] * screenScale)/screenScale;
```
appleguy: This code affects all text nodes, even those that don't have scaling enabled or that aren't inside a baseline-aligned stack.  Although this method is often called off main, there are a lot of critical situations where it is called on main, such as when transitioning to a view controller (modal or nav) and need to layout text positions so content doesn't jump; or in Pinterest, both opening a closeup as well as at the moment of dragging down to relayout the feed.

Is there any other code that uses these values besides the baseline stacks?  What do you think about creating accessor methods that lazily set the local property, or perhaps don't even cache it but just do the attribute / font accesses when requested?

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -104,8 +106,14 @@ - (NSUInteger)lineCountForString:(NSAttributedString *)attributedString
 
 - (CGFloat)scaleFactor
 {
+  if (_measured) {
+    return _scaleFactor;
+  }
+  
   if ([_attributes.pointSizeScaleFactors count] == 0 || isinf(_constrainedSize.width)) {
-    return 1.0;
+    _measured = YES;
```
appleguy: Where is _measured set to NO?  Do we destroy and recreate this object if the attributed string or constrainedSize changes?

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -85,7 +87,7 @@ - (NSUInteger)lineCountForString:(NSAttributedString *)attributedString
   NSLayoutManager *layoutManager = _attributes.layoutManagerCreationBlock ? _attributes.layoutManagerCreationBlock() : [[ASLayoutManager alloc] init];
   layoutManager.usesFontLeading = NO;
   [textStorage addLayoutManager:layoutManager];
-  NSTextContainer *textContainer = [[NSTextContainer alloc] initWithSize:_constrainedSize];
+  NSTextContainer *textContainer = [[NSTextContainer alloc] initWithSize:CGSizeMake(_constrainedSize.width, FLT_MAX)];
```
appleguy: Please consider adding a comment here in your next diff, basically that this code is only used for the purposes of testing whether we fit / learning the true line count so we don't need to limit it.

One question / suggestion: it might actually be faster to set a limit here that is, say, one line greater than what we actually want the string to fit in.  This would allow the layout operation to start earlier if we're trying it at a size where it clearly isn't going to fit.  E.g. if you want something to fit in 3 lines but it's actually 10 at a large starting font size, we should be able to know by line 4 that it isn't going to fit.

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -85,7 +87,7 @@ - (NSUInteger)lineCountForString:(NSAttributedString *)attributedString
   NSLayoutManager *layoutManager = _attributes.layoutManagerCreationBlock ? _attributes.layoutManagerCreationBlock() : [[ASLayoutManager alloc] init];
```
appleguy: These objects, the text storage and the layout manager, are quite expensive to create and even to destroy. They are also required by the renderer. Because many strings will fit without shrinking, it seems quite unfortunate that we must create these objects at least twice.

I don't yet understand all of the mechanics of this awesome system you have put together, so it is possible that we only ever get to this point if we already know that a resize operation is required. However, it would certainly be fantastic and very likely worthwhile if we could either use a text renderer object here for this task, or transfer ownership of these objects after the correct size fit is found so that the drawing operation does not have to re-create them.

screen shot 2016-03-08 at 3 04 51 am

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
rcancro: The renderer owns this object, and the text node will create a new renderer when a new string is set (and in other cases as well). I believe this is ok but it is worth more consideration.

PR #1344, opened 2016-03-09
Ensure that we use the correct default size for collection nodes even if layoutSubviews has not been called yet

eanagel:
This is important when the ASCollectionNode is itself in a node (like a ASTableView of Collections) where the ASCollectionNode is constructed with a zero size and then later resized to the correct size by the layout system.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -780,6 +780,11 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   if (_asyncDataSourceImplementsConstrainedSizeForNode) {
     constrainedSize = [_asyncDataSource collectionView:self constrainedSizeForNodeAtIndexPath:indexPath];
   } else {
+      if (! CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, self.bounds.size)) {
+        _maxSizeForNodesConstrainedSize = self.bounds.size;
+        _ignoreMaxSizeChange = CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, CGSizeZero);
+      }
+
     CGSize maxSize = _maxSizeForNodesConstrainedSize;
     if (ASScrollDirectionContainsHorizontalDirection([self scrollableDirections])) {
       maxSize.width = FLT_MAX;


nguyenhuy:
LGTM. Thanks for fixing this, @eanagel!

appleguy:
@eanagel Great find. The indentation is not quite right here, but I am definitely merging this :)

tomizimobile:
This commit causes a couple issues in my app. 
1) I have an ASCollectionView that has cells which contain an ASPagerNode or an ASCollectionNode. The first time the viewController is loaded, all of the cells contained in the ASPagerNode and ASCollectionNode animate the cells coming from the frame of the first cell going to their final position.
2) I have an ASCollectionView that now doesn't show any cells until I scroll far enough to perform a batch fetch, which then causes cells to appear (I assume after I insert new cells). This is only happening to the first viewController in my UITabBarController that I present after logging in to the app. It's possible that some specific order of events in my app is causing this to happen, but it never happened until this change.

eanagel:
Interesting @tomizimobile, I actually made this change to solve a possibly related problem. In my case I have an ASCollectionNode in a ASTableView cell. The ASCollectionNode cells were being created before the first time layoutSubviews was called which was leaving _maxSizeForNodesConstrainedSize uninitialized, so it was attempting to create the cells with (0,0) constrainedSize and this ultimately caused UICollectionView to hang when loading the invalid cells.

This behavior started for me recently and this commit was my fix. Can you confirm that this specific commit is causing the issue and not another one in the last week or so (ie can you comment out the inserted lines and see if things start working for you?)

tomizimobile:
@eanagel It's definitely this commit causing the issues - I ran against this commit in master vs. the previous commit in master.

eanagel:
Ok, @tomizimobile  I have a less intrusive change that fixes the issue I'm seeing. Before I PR it, would you mind seeing if things work for you with the following change?

Simply replace the following lines in ASCollectionView.mm @ about line 783

      if (! CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, self.bounds.size)) {
        _maxSizeForNodesConstrainedSize = self.bounds.size;
        _ignoreMaxSizeChange = CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, CGSizeZero);
      }

    CGSize maxSize = _maxSizeForNodesConstrainedSize;

With the following line:

    CGSize maxSize = CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, CGSizeZero) ? self.bounds.size : _maxSizeForNodesConstrainedSize;

tomizimobile:
@eanagel I just tested it, and it seems to clear up the issues. Thanks!

PR #1346, opened 2016-03-09
[ASButtonNode] Support the combined [.Selected, .Highlighted] state

tomizimobile:


AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -365,6 +388,30 @@ - (void)setImage:(UIImage *)image forState:(ASControlState)state
...
+        return _selectedHighlightedBackgroundImage;
+      
+      case ASControlStateDisabled:
+        return _disabledBackgroundImage;
+            
+      default:
+        return _normalBackgroundImage;
+    }
+}
+
```
aaronschubert0: @tomizimobile Seems like the spacing for this has reverted to 4 spaces. The standard for ASDK is 2 spaces. You can change this in your project settings. I fell for this a lot initially.

tomizimobile: @aaronschubert0 whoops, the spacing was right but must've gotten blown away after I moved it up to match the order of other getter/setter combos. Fixed.

AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -365,52 +388,58 @@ - (void)setImage:(UIImage *)image forState:(ASControlState)state
...
-- (void)setBackgroundImage:(UIImage *)image forState:(ASControlState)state
+- (UIImage *)backgroundImageForState:(ASControlState)state
 {
   ASDN::MutexLocker l(_propertyLock);
   switch (state) {
     case ASControlStateNormal:
-      _normalBackgroundImage = image;
-      break;
-      
+      return _normalBackgroundImage;
```
appleguy: Thanks!  I definitely prefer the direct return.

appleguy:
@tomizimobile thanks, this is an important feature / subtlety that was not handled in the original!

appleguy:
In the future I would love to have some of this stuff more abstracted so we don't need separate ivars for everything, and the rules can be codified in a more logical way, but this isn't important.

Issue #1295, opened 2016-03-01
[ASTextNode] If maxLines is the first height constraint on a string ending with line break, truncation string not shown.


yxztj:
Repro steps:

1. Set attributed text: "1\n2\n3\n4\n5"
2. Set maximumNumberOfLines to be 3
3. UILabel displays the expected text, the ellipsis should be displayed at last. ASTextNode doesn't.

I noticed #923 could be a related issue, so I checked ASDK 1.9.3, which has the expected ellipsis shown. But after that, the issue appeared till now(also tested 1.9.7)

Sample: https://github.com/yxztj/ASTextNodeTruncationIssue
The left uses UILabel, the right uses ASTextNode
<img width="733" alt="screen shot 2016-03-01 at 11 25 14 am" src="https://cloud.githubusercontent.com/assets/3334458/13416813/58a551a8-dfa0-11e5-8991-f182f696d083.png">

appleguy:
@yxztj thank you for reporting this. In your example, why is the font size different between the label and the node?

yxztj:
@appleguy Thanks for the quick reply.
I was trying to repro the issue using fewer lines, so I didn't set any font attributes for the label and text node. Guess it's the default behavior?

yxztj:
The code is quite simple, so I'd just paste it here.
```Swift
        let testString = "1\n2\n3\n4\n5"
        let attributedString = NSAttributedString(string: testString)
        let maxLines = 3
        
        testLabel.text = testString
        testLabel.backgroundColor = UIColor.lightGrayColor()
        testLabel.numberOfLines = maxLines
        testLabel.frame = CGRect(x: 20, y: 60, width: 40, height: 100)
        testLabel.attributedText = attributedString
        self.view.addSubview(testLabel)
        
        let textNode = ASTextNode()
        textNode.maximumNumberOfLines = UInt(maxLines)
        textNode.backgroundColor = UIColor.lightGrayColor()
        textNode.attributedString = attributedString
        self.view.addSubnode(textNode)
        textNode.frame = CGRect(x: 70, y: 60, width: 40, height: 100)
```

appleguy:
@yxztj have you tried setting a truncationAttributedString on the ASTextNode?  I would expect it to use ... by default, but would be interested to know if even that is insufficient.  I suspect this is caused only by truncations triggered by maximum line cap, rather than a size constraint.

[...]

PR #1347, opened 2016-03-09
[ASTextNode] Fix default ellipsis not showing for line breaks

aaronschubert0:
@appleguy Appreciate your super busy at the moment.

Little fix to address #1295. There is an easy workaround for this but turned out to be a simple fix to correct it to the expected behaviour.


appleguy:
Fantastic, thanks!  Ultra minor nit, but in the future, include a space after //

aaronschubert0:
Thanks @appleguy good point, look cleaner that way.

PR #1348, opened 2016-03-09
Ensure that we use the correct default size for collection nodes even

eanagel:
..if layoutSubviews has not been called yet. This is a more conservative approach than the previous PR #1344 which caused issues for @tomizimobile. This fixes the issues I was seeing and shouldn't break anything else ;) (Previous version could keep the initial relayoutAllNodes from being called in layoutSubviews in some situations.)


eanagel:
Ok @tomizimobile has checked that his use case works with this change and mine does too. I'm comfortable merging once it's been reviewed.

PR #1355, opened 2016-03-09
[ASRangeController] Add ASLayoutRangeModeLowMemory

maicki:


AsyncDisplayKit/Details/ASLayoutRangeType.h
```diff
@@ -20,12 +20,20 @@ typedef NS_ENUM(NSUInteger, ASLayoutRangeMode) {
...
   /**
    * Normal/Full mode that a range controller uses to provide the best experience for end users.
    * This mode is usually used for an active scroll view.
    * A range controller under this requires more resources compare to minimum mode.
    */
   ASLayoutRangeModeFull,
+  
+  /**
+   * Low Memory mode is used when a range controller should limit the amount of work it performs to 0.
+   * Thus, it discards most of the views/layers that are created and it is trying to save as much system
```
appleguy: Technically this does not discard the views / layers themselves, but does eliminate all of the backing stores / decoded images / drawn text - while retaining only the fetched data for the range that would be visible if it comes back onscreen.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -32,6 +33,29 @@ @interface ASRangeController ()
...
+  dispatch_once(&onceToken, ^{
+    [self registerLowMemoryNotification];
+  });
+}
+
+#pragma mark - Class
+
++ (ASWeakSet *)rangeControllers
+{
+  static ASWeakSet<ASRangeController *> *rangeController;
```
appleguy: should use plural.  Also, for a static variable, please prefix with __ (two underscores).  So probably __allRangeControllers.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -32,6 +33,29 @@ @interface ASRangeController ()
 
 @implementation ASRangeController
 
+#pragma mark - NSObject
+
++ (void)load {
```
appleguy: Should use { on new line.  However, please just move this into the method below, where we create the weak set.  No need to observe the notification before the first controller is added to the set.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -42,6 +66,8 @@ - (instancetype)init
   _currentRangeMode = ASLayoutRangeModeInvalid;
   _didUpdateCurrentRange = NO;
   
+  [[self.class rangeControllers] addObject:self];
```
appleguy: let's call this allRangeControllersWeakSet just to make it really clear

PR #1358, opened 2016-03-10
[Tests] Replace sleep() with XCTest asynchronous expectations

ejensen:
This test fails occasionally in Travis CI. I've had better luck with asynchronous test consistency when XCTest expectations are used.

### DIFF
--- a/AsyncDisplayKitTests/ASBasicImageDownloaderTests.m
+++ b/AsyncDisplayKitTests/ASBasicImageDownloaderTests.m
@@ -1,5 +1,5 @@
 //
-//  ASZBasicImageDownloaderTests.m
+//  ASBasicImageDownloaderTests.m
 //  AsyncDisplayKit
 //
 //  Created by Victor Mayorov on 10/06/15.
@@ -10,7 +10,6 @@
 
 #import <AsyncDisplayKit/ASBasicImageDownloader.h>
 
-// Z in the name to delay running until after the test instance is operating normally.
 @interface ASBasicImageDownloaderTests : XCTestCase
 
 @end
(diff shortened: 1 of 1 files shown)


appleguy:
@ejensen I particularly appreciate this one; thank you. The test that is far less reliable is related to asynchronous transactions, which has never once failed for an engineer locally, but fails on the build server around a quarter of the time which is pretty awful.

It only started happening after the excellent changes from @knopp, but he also added a lot of tests and his change was tremendously impactful / very stable from the beginning, so I'm pretty confident it is either a Travis or test structure change.  If you are looking around in that area, definitely let me know if the problem is apparent to you even if you're not able to fix.

Seriously, don't hesitate to email me any time, asyncdisplaykit@gmail.com .  I am super happy to have these unexpected surprise diffs land in our PR queue!!  If you are doing a project with the framework I'd be happy to help advise on architecture or approach (or issues you encounter), or if you are interested in taking on a bigger change, there are a ton of impactful directions.

knopp:
Hey! Just some background about the tests. Some tests occasionally fail on travis because they expect libdispatch to overcommit (create two thread on single core machine). And most of the time it does, until it doesn't. The overcommit behavior seems to have been different when dispatch_group_t was used, but I looked in libdispatch source and was unable to find out why.

In any case, libdispatch makes no guarantees about overcommit. So it is a bit wrong for the tests to rely on having two parallel threads on single core machine. 

Btw. this can be reproduced locally. You can change number of active cores of your computer in Instruments. When you change it to one and then run test that rely on overcommit couple of tens of thunsands of times, you will get to situation where you schedule block with dispat_async, but that block is never called because there is another block one running (and waiting on condition).

Also this should only affect tests. I'm not aware of any other code in framework that actually blocks transaction thread waiting for another transaction to unblock it.

PR #1364, opened 2016-03-11
Use property underscore notation for recursivelyTriggerDisplayAndBlock

levi:
Now that I understand the notation, finding that we're not using it correctly.


appleguy:
@levi thanks!  I will take this one - though ironically it exposes a flaw, which is that I have normally named statics and class methods with __ as well, along with the things that I want to avoid collisions with reasonably-named internal versions (e.g. people who might have a _layout so they can call it manually in an animation).  There is no meaningful collision risk here and we should start getting towards an actual, written-down code style (fortunately most of the style is adhered to in 90%+ the codebase, but DEFINITELY not underscore usage where there are some trends but many inconsistencies)

PR #1367, opened 2016-03-11
Tweak rangeMode update API

binl:
Instead of updating until next rangeUpdate, we update it until we manually clearing it. feels like it's easier to use and makes the SDK less of a black box (not knowing when the next rangeUpdate will occur).

### DIFF
--- a/AsyncDisplayKit/ASTableNode.m
+++ b/AsyncDisplayKit/ASTableNode.m
@@ -79,11 +79,20 @@
 
 - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
 {
-    if (!self.isNodeLoaded) {
-        return;
-    }
-    
-    [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
+  if (!self.isNodeLoaded) {
+    return;
+  }
+  
+  [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
+}
+
+- (void)clearCurrentRangeModeUpdate
+{
+  if (!self.isNodeLoaded) {
+    return;
+  }
+
+  [self.view.rangeController clearCurrentRangeModeUpdate];
 }
 
 - (_ASTablePendingState *)pendingState
--- a/AsyncDisplayKit/ASCollectionNode.mm
+++ b/AsyncDisplayKit/ASCollectionNode.mm
@@ -229,6 +229,11 @@
   [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
 }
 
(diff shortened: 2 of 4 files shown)


AsyncDisplayKit/ASTableNode.m
```diff
@@ -79,11 +79,20 @@ - (void)didLoad
...
-    
-    [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
+  if (!self.isNodeLoaded) {
+    return;
+  }
+  
+  [self.view.rangeController updateCurrentRangeWithMode:rangeMode];
+}
+
+- (void)clearCurrentRangeModeUpdate
```
appleguy: Can we just use updateCurrentRangeWithMode: for this?  Because there are not many modes, I think it is fairly unlikely that an app would really need to be able to return to the prior default value without being able to set one.  Main goal is to reduce API surface area, as although this is still in +Beta, we should converge it towards something that we open up more broadly.

appleguy:
@binl thank you for proposing this. You definitely discovered a real, important issue that was caused by very recent changes related to triggering range updates upon a table / collection moving into or out of a window. In addition to the memory footprint, this was causing back to back calls to visibilityDidChange: separated by just one run loop turn, and although the state always ended up correct, this extra thrash was both inefficient and potentially a problem for certain implementations that might not check to filter out noisy calls.

I feel it is very important to continue to observe this behavior to ensure that we are addressing any root causes of these kind of unnecessary calls, and believe that the system is now behaving in a significantly more calm and efficient way after https://github.com/facebook/AsyncDisplayKit/pull/1382 .  If we find more causes, we should revisit to see if they have as concrete of a solution as was possible in that change.

That said, I would not be to surprise if we end up reopening this change and going in this direction. Overall, the range mode API is very new and Will not B production API for at least another month, possibly more. So we do have sometime to continue to inspect the behavior of the system in a variety of complex circumstances, and this change should continue to apply smoothly if we reopen the PR (no need to create a new one since it's so easy to reopen).

Give the new behavior a try, and let me know which scenarios are still producing noisy calls. I am nearly certain that there are at least a handful of cases that do so, potentially with very significant negative impact, and would be very motivated to start digging into why they happen.

Lastly, I am thrilled to see this new pull request from @hannahmbanana - it allows it visualizing the current range controllers in the system, which will dramatically assist in detecting these kind of problems. It is still a fairly rough implementation, but already amazingly useful, and can be expanded and improved as we learn which capabilities are most valuable.  https://github.com/facebook/AsyncDisplayKit/pull/1390

screen shot 2016-03-16 at 8 15 54 pm

PR #1369, opened 2016-03-11
[ASNetworkImageNode] Add support for attempting to get image synchronously

garrettmoon:


AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -169,6 +171,19 @@ - (void)setDelegate:(id<ASNetworkImageNodeDelegate>)delegate
 - (void)displayWillStart
 {
   [super displayWillStart];
+  
+  if (_cacheSupportsSynchronousFetch) {
+    {
```
appleguy: Don't need the additional lock scope - can contain it with the if.  I think it would be slightly more readable.

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -169,6 +171,19 @@ - (void)setDelegate:(id<ASNetworkImageNodeDelegate>)delegate
...
+      ASDN::MutexLocker l(_lock);
+      if (_URL && _downloadIdentifier == nil) {
+        UIImage *result = [_cache synchronouslyFetchedCachedImageWithURL:_URL];
```
appleguy: Just a sanity check for my understanding, is it guaranteed that this image will always be compressed? Based on our configuration, I assume that is the case, but just checking

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -169,6 +171,19 @@ - (void)setDelegate:(id<ASNetworkImageNodeDelegate>)delegate
 - (void)displayWillStart
 {
   [super displayWillStart];
+  
+  if (_cacheSupportsSynchronousFetch) {
+    {
+      ASDN::MutexLocker l(_lock);
+      if (_URL && _downloadIdentifier == nil) {
```
appleguy: Might be some savings to check _imageLoaded here.

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -29,6 +29,13 @@ + (instancetype)sharedDownloader
 
 #pragma mark ASImageProtocols
 
+- (UIImage *)synchronouslyFetchedCachedImageWithURL:(NSURL *)URL
+{
+  NSString *key = [[PINRemoteImageManager sharedImageManager] cacheKeyForURL:URL processorKey:nil];
```
appleguy: Incredibly minor, but local variable for sharedImageManager makes sense for this - probably immesurable since the iPhone 3G -> 3GS, armv6 -> armv7 switch though :-P

AsyncDisplayKit/ASNetworkImageNode.mm
garrettmoon: As long as PINRemoteImage is only used with ASDK, yes. Basically, if you mix shouldSkipDecode and not skipping decode, you can't tell what is going to end up in the cache (I want to address this in PINRemoteImage). In the default case though, it should always be uncompressed.

AsyncDisplayKit/ASNetworkImageNode.mm
garrettmoon: Doh! Yes :)

AsyncDisplayKit/ASNetworkImageNode.mm
garrettmoon: Good call

PR #1378, opened 2016-03-15
[ASMultidimensionalArrayUtils] Optimization for recursive version, expanded use case of more efficient 2D version

hannahmbanana:
- optimized ASRecursivelyFindIndexPathsForMultidimensionalArray() to remove enumerateUsingBlock:
- replaced single project call to ASIndexPathsForMultidimensionalArray() in ASRangeController with ASIndexPathsForTwoDimensionalArray()


appleguy:
@hannahmbanana Thanks!  Slightly embarassing, I think I added the 2D version recently which is /way/ faster, but then clearly missed a key use case.

It is nice to eliminate the block overhead in the multi dimensional version, although I think after this change we do not use it anywhere? That is fine for now, we can leave it in place for future usage or elimination if it goes unused for long enough.

PR #1381, opened 2016-03-15
[ASImageNode] Annotate imageModificationBlock as nullable

ejensen:
`imageModificationBlock` defaults to nil, and can be set to nil.


MarvinNazari:
this is causing a crash for me

PR #1382, opened 2016-03-15
[ASRangeController] Improvements for automatic range mode

maicki:
- Only update range mode if visibility changes if the node is not range controlled
- Only change explicitly set range mode if ASRangeController becomes visible
- Return interface state for range controller in ASCollectionView and ASTableView based on if the containing node is range managed

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -920,14 +920,14 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   ASCollectionNode *collectionNode = self.collectionNode;
   if (collectionNode && [collectionNode supportsRangeManagedInterfaceState]) {
     // Only use the interfaceState of nodes that are range managed
-    return collectionNode.interfaceState;
+    ASInterfaceState interfaceState = collectionNode.interfaceState;
+    return (self.window == nil ? (interfaceState &= (~ASInterfaceStateVisible)) : interfaceState);
   } else {
     // For not range managed nodes or until we can always create an associated ASCollectionNode
     // without a retain cycle, we might be on our own to try to guess if we're visible.  The node normally
     // handles this even if it is the root / directly added to the view hierarchy.
-    return (self.window != nil ? ASInterfaceStateVisible : ASInterfaceStateNone);
+    return (self.window == nil ? ASInterfaceStateNone : (ASInterfaceStateVisible | ASInterfaceStateDisplay));
   }
-
 }
 
 - (NSArray *)rangeController:(ASRangeController *)rangeController nodesAtIndexPaths:(NSArray *)indexPaths
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1137,7 +1137,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     [node __exitHierarchy];
   }
 
-  // Updating the visible node index paths only for range managed nodes. Not range managed nodes will get their
+  // Updating the visible node index paths only for not range managed nodes. Range managed nodes will get their
   // their update in the layout pass
   if (![node supportsRangeManagedInterfaceState]) {
     [_rangeController visibleNodeIndexPathsDidChangeWithScrollDirection:self.scrollDirection];
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -919,14 +919,15 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 {
   ASCollectionNode *collectionNode = self.collectionNode;
   if (collectionNode && [collectionNode supportsRangeManagedInterfaceState]) {
-    // Only use the interfaceStatate of nodes that are range managed
+    // Only use the interfaceState of nodes that are range managed
     return collectionNode.interfaceState;
+  } else {
+    // For not range managed nodes or until we can always create an associated ASCollectionNode
+    // without a retain cycle, we might be on our own to try to guess if we're visible.  The node normally
+    // handles this even if it is the root / directly added to the view hierarchy.
+    return (self.window != nil ? ASInterfaceStateVisible : ASInterfaceStateNone);
   }
 
-  // For not range managed nodes or until we can always create an associated ASCollectionNode
-  // without a retain cycle, we might be on our own to try to guess if we're visible.  The node normally
-  // handles this even if it is the root / directly added to the view hierarchy.
-  return (self.window != nil ? ASInterfaceStateVisible : ASInterfaceStateNone);
 }
 
 - (NSArray *)rangeController:(ASRangeController *)rangeController nodesAtIndexPaths:(NSArray *)indexPaths
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -918,14 +918,15 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 - (ASInterfaceState)interfaceStateForRangeController:(ASRangeController *)rangeController
 {
   ASCollectionNode *collectionNode = self.collectionNode;
-  if (collectionNode) {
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1986,13 +1986,9 @@ - (void)visibilityDidChange:(BOOL)isVisible
 {
 }
 
-/**
- * We currently only set interface state on nodes in table/collection views. For other nodes, if they are
```
appleguy: I'm not sure there is a reason to remove this comment - let's leave it in, at least until it is replaced by something comparable / useful.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -787,14 +787,15 @@ - (CGSize)viewportSizeForRangeController:(ASRangeController *)rangeController
 - (ASInterfaceState)interfaceStateForRangeController:(ASRangeController *)rangeController
 {
   ASTableNode *tableNode = self.tableNode;
-  if (tableNode) {
+  if (tableNode && [tableNode supportsRangeManagedInterfaceState]) {
+    // Only use the interfaceStatate of nodes that are range managed
```
appleguy: Typo Statate

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -880,14 +880,15 @@ - (CGSize)viewportSizeForRangeController:(ASRangeController *)rangeController
 - (ASInterfaceState)interfaceStateForRangeController:(ASRangeController *)rangeController
 {
   ASCollectionNode *collectionNode = self.collectionNode;
-  if (collectionNode) {
+  if (collectionNode && [collectionNode supportsRangeManagedInterfaceState]) {
+    // Only use the interfaceStatate of nodes that are range managed
     return collectionNode.interfaceState;
-  } else {
```
appleguy: I prefer to keep the if / else rather than an early return.  It makes the code more clear in my opinion that there are exactly 2 cases that the code is handling.

maicki:
@appleguy Addressed comments

AsyncDisplayKit/ASTableView.mm
```diff
@@ -787,11 +787,12 @@ - (CGSize)viewportSizeForRangeController:(ASRangeController *)rangeController
...
+  if (tableNode && [tableNode supportsRangeManagedInterfaceState]) {
+    // Only use the interfaceState of nodes that are range managed
     return self.tableNode.interfaceState;
   } else {
-    // Until we can always create an associated ASTableNode without a retain cycle,
-    // we might be on our own to try to guess if we're visible.  The node normally
+    // For not range managed nodes or until we can always create an associated ASTableNode
+    // without a retain cycle, we might be on our own to try to guess if we're visible.  The node normally
     // handles this even if it is the root / directly added to the view hierarchy.
     return (self.window != nil ? ASInterfaceStateVisible : ASInterfaceStateNone);
```
appleguy: Both here and in collcetion, let's make the first case in the ternary be (ASInterfaceStateVisible | ASInterfaceStateDisplay)

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -918,13 +918,15 @@ - (CGSize)viewportSizeForRangeController:(ASRangeController *)rangeController
 - (ASInterfaceState)interfaceStateForRangeController:(ASRangeController *)rangeController
```
appleguy: This looks like exactly what we need.  In a future diff, let's factor this into the range controller itself, and do something like interfaceStateForDataSource:withWindow: and then deduplicate / share the logic.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -232,7 +234,10 @@ - (void)_updateVisibleNodeIndexPaths
   _allPreviousIndexPaths = allCurrentIndexPaths;
   
   _currentRangeMode = rangeMode;
-  _didUpdateCurrentRange = NO;
+  // Reset the current range mode only if the range controller comes visible
+  if (ASInterfaceStateIncludesVisible(selfInterfaceState)) {
```
appleguy: Is this actually needed?  Looks safe but I think not needed because this is also checked as part of the condition above.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -189,7 +189,9 @@ - (void)_updateVisibleNodeIndexPaths
   
   ASInterfaceState selfInterfaceState = [self interfaceState];
   ASLayoutRangeMode rangeMode = _currentRangeMode;
-  if (!_didUpdateCurrentRange) {
+  // If the range mode is explicitly set via updateCurrentRangeWithMode: it will last in that mode until the
+  // range controller becomes visible again or explicitly changes the range mode again
+  if (!_didUpdateCurrentRange || ASInterfaceStateIncludesVisible(selfInterfaceState)) {
```
appleguy: I think this may need to be &&?  What if we change the range mode but are still visible?  I think we want that to stick for at least one pass.

AsyncDisplayKit/Details/ASRangeController.mm
maicki: If we change the logic from above yes we don't need that anymore in here

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1144,9 +1136,12 @@ - (void)didMoveToWindow
   if (!visible && node.inHierarchy) {
     [node __exitHierarchy];
   }
-    
-  // Trigger updating interfaceState for cells in case ASCollectionView becomes visible or invisible
-  [_rangeController visibleNodeIndexPathsDidChangeWithScrollDirection:self.scrollDirection];
+
+  // Updating the visible node index paths only for range managed nodes. Not range managed nodes will get their
```
appleguy: comment should say only for NOT range managed collections / tables, e.g. root level ones, need this notification?  As range managed ones will get this update through the range controller itself that owns them?

AsyncDisplayKit/ASDisplayNodeExtras.mm
```diff
@@ -10,6 +10,18 @@
 #import "ASDisplayNodeInternal.h"
 #import "ASDisplayNode+FrameworkPrivate.h"
 
+extern ASInterfaceState ASInterfaceStateForDisplayNode(ASDisplayNode *displayNode, UIWindow *window)
+{
+    if (displayNode && [displayNode supportsRangeManagedInterfaceState]) {
+        // Only use the interfaceState of nodes that are range managed
+        ASInterfaceState interfaceState = displayNode.interfaceState;
+        return (window == nil ? (interfaceState &= (~ASInterfaceStateVisible)) : interfaceState);
```
appleguy: // Directly clear the visible bit if we are not in a window.  This means that the interface state is, if not already, about to be set to invisible as it is not possible for an element to be visible while outside of a window.

appleguy: Now that this has been generalized to work for all ASDisplayNodes, we need to assert !layerBacked, as a node that is layerBacked may have a nil window.  Fortunately, the current use cases will never hit this as they must always be view backed, which means that all of their containers must also be view backed.

AsyncDisplayKit/Details/ASRangeControllerUpdateRangeProtocol+Beta.h
```diff
@@ -12,6 +12,20 @@
...
+ * Logic for the automatic range mode:
+ * 1. If there are no visible node paths available nothing is to be done and no range update will happen
+ * 2. The initial range update if the range controller is visible always will be ASLayoutRangeModeCount
+ *    (ASLayoutRangeModeMinimum) as it's the initial fetch
+ * 3. The range mode set explicitly via updateCurrentRangeWithMode: will last at least one range update. After that it
+      the range controller will use the explicit set range mode until it becomes visible and a new range update was
+      triggered or a new range mode via updateCurrentRangeWithMode: is set
+ * 4. If range mode is not explicitly set the range mode is variying based if the range controller is visible or not
+ */
+
```
appleguy: Should this be above the setRangeMode: API?

maicki: I added the comment at the top of the ASRangeControllerUpdateRangeProtocol as the basically applies to the whole feature not only specific for setRangeMode:

PR #1383, opened 2016-03-15
Fix ASVideoNodeTests testPlayerLayerNodeIsNotAddedIfVisibleButShouldN…

maicki:
…otBePlaying test

With a recent change the pre condition for setting an interface state of a display node to so that it should never be possible for a node to be visible but not be allowed / expected to display. In the testPlayerLayerNodeIsNotAddedIfVisibleButShouldNotBePlaying test the interface state is only set to ASInterfaceStateVisible and does not include ASInterfaceStateDisplay. This will let the test to fail.

### DIFF
--- a/AsyncDisplayKitTests/ASVideoNodeTests.m
+++ b/AsyncDisplayKitTests/ASVideoNodeTests.m
@@ -140,7 +140,7 @@
   _videoNode.asset = _firstAsset;
 
   [_videoNode pause];
-  [_videoNode setInterfaceState:ASInterfaceStateVisible];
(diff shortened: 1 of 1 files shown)


AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -140,7 +140,7 @@ - (void)testPlayerLayerNodeIsNotAddedIfVisibleButShouldNotBePlaying
   _videoNode.asset = _firstAsset;
 
   [_videoNode pause];
-  [_videoNode setInterfaceState:ASInterfaceStateVisible];
```
appleguy: This actually needed to be an |, not &!  hehe.  I landed the fix.

PR #1384, opened 2016-03-15
[ASCollectionView] Don't animate size changes if some of the updated cell nodes don't want to

nguyenhuy:


appleguy:
Thanks @nguyenhuy !  @maicki we may want to create that "opposite of ASHierarchyState" thing to signal "upward-hierarchy" information and then port this feature to be based on it.

nguyenhuy:
That would be ideal, @appleguy @maicki.

PR #1385, opened 2016-03-15
[ASDisplayNode] Fix display node pending node property deadlock

maicki:
This should fix a deadlock in ASDisplayNode and it's caused by displayWillStart: and where one thread is recursing down the tree and another thread is recursing up the tree. We remove the lock in _pendingDisplayNodes for the property, but need to guarantee that we only modify the _pendingDisplayNodes state on the main thread.

Furthermore add documentation on what thread displayWillStart and displayDidFinish will be called


nguyenhuy:
@maicki This is awesome. Thanks for fixing!

PR #1392, opened 2016-03-17
[ASDisplayNode] Placeholder Image bug fix

hannahmbanana:
Issue:
- The placeholder image layer does not seem to reload if you scroll past an ASDisplayNode far enough that its contents are cleared and then return.
- E.g. if you scroll past an ASNetworkImageNode node before the image has loaded, (if enabled) the placeholder image will display the first time. If you keep scrolling far enough down so that the node contents clear and then scroll back up past it (again before the image has loaded), the placeholder image will not reload.

Fix:
- When a node is becoming visible, we should check if it’s placeholder layer still exists and if not (and placeholder is enabled), we should recreate it
- check that the node implements display

Notes:
- I’m not sure if this is the correct approach. I dug around in the code, tried this fix in my app and it seems to work.
- I am unsure about whether to call _setupPlaceholderLayer: or _setupPlaceholderLayerContents: or both to handle the recreation of the placeholder image layer.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1676,7 +1676,7 @@ static NSInteger incrementIfFound(NSInteger i) {
   [_pendingDisplayNodes removeObject:node];
 
   // only trampoline if there is a placeholder and nodes are done displaying
-  if ([self _pendingDisplayNodesHaveFinished] && _placeholderLayer.superlayer) {
+  if (self.layer.contents && [self _pendingDisplayNodesHaveFinished] && _placeholderLayer.superlayer) {
     void (^cleanupBlock)() = ^{
       [self _tearDownPlaceholderLayer];
     };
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2260,14 +2260,23 @@ - (void)displayWillStart
...
-  if (_placeholderImage && _placeholderLayer && self.layer.contents == nil) {
-    [CATransaction begin];
-    [CATransaction setDisableActions:YES];
-    [self _setupPlaceholderLayerContents];
-    _placeholderLayer.opacity = 1.0;
-    [CATransaction commit];
-    [self.layer addSublayer:_placeholderLayer];
+  
+  if (_placeholderEnabled && self.layer.contents == nil && [self __implementsDisplay]) {
+    if (!_placeholderImage) {
```
appleguy: This code is almost exactly correct, although doing it at this point Will mean that the sun elements which never become fully visible on screen may do work to re-create their placeholders as they are starting to display because of entering the off screen display range.

Because placeholders are synchronous, it is important to minimize the amount of work that it Is done to support them. Perhaps more importantly, in reviewing this I noticed a few other bugs, most notably that the network image node will remove its placeholder for the image has finished downloading because a display pass is allowed to complete on it even though it has no contents to draw.

Thank you for picking up this task and investigating, although I have started a new branch based off of yours to continue the change and posted it in a different pull request: https://github.com/facebook/AsyncDisplayKit/pull/1398

appleguy:
Thank you @hannahmbanana for picking up this task and investigating!  To investigate further, I actually started a branch based off of this PR, and continue the change -- but posted it in a different pull request: #1398 .

PR #1397, opened 2016-03-18
[ASRunLoopQueue] Add custom run loop source to signal if jobs are still enqueued, to guarantee another runloop turn.

maicki:
It is not guaranteed that the runloop will turn if it has no scheduled work, and this causes processing of the queue to stop. Attaching a custom loop source to the run loop and signal it if new work needs to be done

### DIFF
--- a/AsyncDisplayKit/ASRunLoopQueue.mm
+++ b/AsyncDisplayKit/ASRunLoopQueue.mm
@@ -11,9 +11,14 @@
 
 #import <deque>
 
+static void runLoopSourceCallback(void *info) {
+  // No-op
+}
+
 @interface ASRunLoopQueue () {
   CFRunLoopRef _runLoop;
   CFRunLoopObserverRef _runLoopObserver;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -11,9 +11,14 @@
...
 #import <deque>
 
+static void runLoopSourceCallback(void *info) {
+  // No-op
+}
+
 @interface ASRunLoopQueue () {
   CFRunLoopRef _runLoop;
   CFRunLoopObserverRef _runLoopObserver;
+  CFRunLoopSourceRef _runLoopSource;
```
appleguy: Interesting observation: neither of these actually need to be instance variables, and could instead be static global variables. I don't think this is a big issue right now, but when we support deallocation in more places and find more use cases, it might be nice to eventually consolidate these. Don't worry about it now.

PR #1399, opened 2016-03-19
[ASMapNode] Reduce number of MapKit calls to -cancel by using the .isLoading property.

appleguy:
@aaronschubert0 would love your review - if you have time!

### DIFF
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -134,12 +134,14 @@
 - (void)setOptions:(MKMapSnapshotOptions *)options
 {
   ASDN::MutexLocker l(_propertyLock);
-  _options = options;
-  if (self.isLiveMap) {
-    [self applySnapshotOptions];
-  } else {
-    [self resetSnapshotter];
-    [self takeSnapshot];
+  if (!_options || ![options isEqual:_options]) {
+    _options = options;
+    if (self.isLiveMap) {
+      [self applySnapshotOptions];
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASMapNode.mm
```diff
@@ -160,7 +162,11 @@ - (void)takeSnapshot
   if (!_snapshotter) {
     [self setUpSnapshotter];
   }
-  [_snapshotter cancel];
+  
+  if (_snapshotter.isLoading) {
```
aaronschubert0: @appleguy The only thing that this may have a negative side affect on is if someone is changing the ASMapNode options rapidly or on slow connections. With this implementation the snapshotter simply rejects the new change, rather than aborting the current change and replacing it with this new one. How have you found it in practice?

aaronschubert0:
@appleguy Only have the one comment, apart from that, awesome improvements. Feels good that you've removed any ambiguity that was left in the code or comments!

AsyncDisplayKit/ASMapNode.mm
appleguy: @aaronschubert0 that true? I think the set options method does destroy the snapshot or, but if you could check that the logic feels correct overall for this scenario you described which is certainly an important one, that would be reassuring!

Apologies that I have not gotten to your tvOS diff yet, hope to do so soon.

aaronschubert0:
@appleguy Logic looks all good!

Issue #1366, opened 2016-03-11
Possible bug in the CustomCollectionView example


cc-lam:
Scrolling the collection view always triggers a re-layout.

`shouldInvalidateLayoutForBoundsChange:` always returns true because `collectionView.bounds` is always different from `newBounds`

hannahmbanana:
@cc-lam You're right - this does seem like it would cause a performance hit.

@levi - did you mean to compare bounds.size rather than bounds in the following function in MosaicCollectionViewLayout? Was this meant to support rotation? Are there other reasons this app would want to invalidate layout upon bounds change? I can throw up a quick PR for the one line of code if that's the case. I know you're super busy optimizing the rest of the framework. Btw - great choice on photos!

- (BOOL)shouldInvalidateLayoutForBoundsChange:(CGRect)newBounds
{
  if (!CGRectEqualToRect(self.collectionView.bounds, newBounds)) {
    return YES;
  }
  return NO;
}

appleguy:
@cc-lam thanks for reporting this.  Indeed the contentOffset changing would result in bounds.origin changing frequently.  Fixed by https://github.com/facebook/AsyncDisplayKit/pull/1403.

PR #1403, opened 2016-03-20
[CustomCollectionView Example] Fix shouldInvalidateLayoutForBoundsChange: method

hannahmbanana:
Issue (#1366):
CustomColelctionView example app always returns YES from shouldInvalidateLayoutForBoundsChange: which triggers a re-layout.

Fix:
Comparing CGSize rather than CGRect is enough to  handle rotation support without always triggering a re-layout.


appleguy:
cc @levi -- this looks correct to me, but let me know if not!

PR #1406, opened 2016-03-21
[Accessibility] Basic a11y support for ASButtonNode (adopt text node title as a11y label)

rahul-malik:


AsyncDisplayKit/ASControlNode.mm
```diff
@@ -86,9 +87,11 @@ - (id)init
 
   // As we have no targets yet, we start off with user interaction off. When a target is added, it'll get turned back on.
   self.userInteractionEnabled = NO;
+  self.isAccessibilityElement = YES;
```
appleguy: Does the accessibility system automatically consider views with gesture recognizes to be accessibility elements? If not, we may want to add support for this in _ASDisplayView, where elements with a nonzero number of gesture recognizers are marked as being accessibility elements.

More importantly though, I think we need to add this line only in the addTarget: method.  ASTextNode, ASImageNode, ASMapNode, etc are all ASControlNodes for convenience, and yet many of them are not actually interactive elements. This explains the userInteractionEnabled = NO above as well.

AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -175,6 +181,8 @@ - (void)updateTitle
     _titleNode.attributedString = newTitle;
     [self setNeedsLayout];
   }
+
+  self.accessibilityLabel = _titleNode.accessibilityLabel;
```
appleguy: Not all buttons have a title, and not all titles may have an accessibilityLabel.  This method might get called and overwrite the label in the case of an image-only button where the programmer directly set the accessibilityLabel for the button

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -89,6 +90,13 @@ - (id)init
   return self;
 }
 
+- (void)setUserInteractionEnabled:(BOOL)userInteractionEnabled
+{
+  [super setUserInteractionEnabled:userInteractionEnabled];
+  self.isAccessibilityElement = userInteractionEnabled;
```
appleguy: This is a great solution.

appleguy:
Very personally satisfied to have the opportunity to merge this!  Thank you for caring about every user's experience, @rahul-malik.

PR #1407, opened 2016-03-21
[ASDisplayNode] [Async layout transition] Add method to cancel layout transitions in progress

nguyenhuy:
- Also make sure a transition isn't invalidated right after it passed the validation test and before it proceeds

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -98,4 +98,9 @@ ASDISPLAYNODE_EXTERN_C_END
  */
 - (BOOL)placeholderShouldPersist;
 
+/**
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -98,4 +98,9 @@ ASDISPLAYNODE_EXTERN_C_END
  */
 - (BOOL)placeholderShouldPersist;
 
+/**
+ * @abstract Cancels all performing layout transitions. Can be called on any thread.
```
appleguy: In a later change, we should add some additional detail here to clarify what the point of no return is for committing a new layout, And perhaps give a short example of a scenario of that you would want to or need to use this method.

PR #1410, opened 2016-03-21
[ASRunLoopQueue] Always signal runloop source if queue is not fully drained, to ensure more batches are guaranteed to run.

maicki:


AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -102,6 +126,12 @@ - (void)processQueue
       self.queueConsumer(itemsToProcess[i], isQueueDrained);
     }
   }
+
+  // If the queue is not fully drained yet force another run loop to process next batch of items
+  if (!isQueueDrained) {
+    CFRunLoopSourceSignal(_runLoopSource);
+    CFRunLoopWakeUp(_runLoop);
```
appleguy: Do you need the WakeUp call?  I suspect the answer is no, because at this time we are currently running in an observer callback, and after running all of the BeforeWaiting observers it must surely check for more sources / etc to run before determining if it will sleep?  I could be wrong, but let's remove it if we don't think it is required as it will help keep the code as simple as possible.

AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -47,9 +56,17 @@ - (instancetype)initWithRunLoop:(CFRunLoopRef)runloop andHandler:(void(^)(id deq
     // the queue to stop. Attaching a custom loop source to the run loop and signal it if new work needs to be done
     CFRunLoopSourceContext *runLoopSourceContext = (CFRunLoopSourceContext *)calloc(1, sizeof(CFRunLoopSourceContext));
     runLoopSourceContext->perform = runLoopSourceCallback;
+#if ASRunLoopQueueLoggingEnabled
+    runLoopSourceContext->info = (__bridge void *)self;
```
appleguy: Does this ensure self is not retained?  Not super familiar with __bridge and the other variants of it.  Although some runloop queues are for the full app lifetime, we are planning to make new ones that may have a shorter life, such as pairing with ASDataController

PR #1412, opened 2016-03-22
[ASRangeController] Prevent deallocation of dataSource & delegate in Collection & Table when navigating back, during final range update.

maicki:
Grab a strong reference for asyncDataSource and asyncDelegate in ASCollectionView and ASTableView before executing the range update to be sure they are not going away while executing the range update. This can happen in range updates while going back in the view controller hierarchy

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1139,7 +1139,14 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   // Updating the visible node index paths only for not range managed nodes. Range managed nodes will get their
   // their update in the layout pass
   if (![node supportsRangeManagedInterfaceState]) {
-    [_rangeController visibleNodeIndexPathsDidChangeWithScrollDirection:self.scrollDirection];
+    // Grab a strong reference for data source and delegate to be sure they are not going away while executing
+    // the range update. This can happen in range updates while going back in the view controller hierarchy
+    __block id<ASCollectionDataSource> asyncDataSource = _asyncDataSource;
+    __block id<ASCollectionDelegate> asyncDelegate = _asyncDelegate;
+    [_rangeController scheduleRangeUpdateCompletion:^{
+      asyncDataSource = nil;
+      asyncDelegate = nil;
+    }];
   }
 }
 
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1139,7 +1139,14 @@ - (void)didMoveToWindow
   // Updating the visible node index paths only for not range managed nodes. Range managed nodes will get their
   // their update in the layout pass
   if (![node supportsRangeManagedInterfaceState]) {
-    [_rangeController visibleNodeIndexPathsDidChangeWithScrollDirection:self.scrollDirection];
+    // Grab a strong reference for data source and delegate to be sure they are not going away while executing
+    // the range update. This can happen in range updates while going back in the view controller hierarchy
+    __block id<ASCollectionDataSource> asyncDataSource = _asyncDataSource;
```
appleguy: Technically these don't need to be block variables, just referenced inside the block, although perhaps this is the most explicit way?

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -111,6 +113,15 @@ - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
 
 - (void)scheduleRangeUpdate
 {
+  [self scheduleRangeUpdateCompletion:nil];
+}
+
+- (void)scheduleRangeUpdateCompletion:(void (^)(void))completion
+{
+  if (completion) {
```
appleguy: Should we add a main thread assertion here?

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -320,6 +336,11 @@ - (void)_updateVisibleNodeIndexPaths
   _rangeIsValid = YES;
   _queuedRangeUpdate = NO;
   
+  for (void (^completionBlock)(void) in _scheduledRangeUpdateCompletionBlocks) {
+    completionBlock();
```
appleguy: This does seem like a useful pattern, although it is an interesting question on whether we will keep it long-term after addressing the root need for this change. It could actually turn out to be something we invest in and use more.

AsyncDisplayKit/Details/ASRangeController.mm
maicki: Yes we should add a main thread assertion in the future here

PR #1415, opened 2016-03-22
[ASInterfaceState] Clear the "Visible" bit immediately upon deallocation, rather than waiting for ASRangeController on the next runloop.

appleguy:


maicki:
@appleguy LGTM

PR #1420, opened 2016-03-24
[SocialAppLayout example] update to use ASTextNode's passthroughNonlinkTouches property

hannahmbanana:


appleguy:
Cool!  This allows selecting myself to work with a tap anywhere in the text region, which is currently hooked up to toggle rasterization.

PR #1421, opened 2016-03-24
[ASDisplayNode] When Enumerating Subtree, Copy sublayers to Avoid NSFastEnumerationMutation Exception

Adlai-Holler:
I have some insanely complex table view cells, and as I scroll through them, I eventually get an `NSFastEnumerationMutation` exception when a node leaves the display range. It's crunch time and I'm exhausted, so I don't know what code on my end is modifying the layer tree due to leaving the display range, but I'll attach some snapshots. It's easy to reproduce so if you want to investigate @appleguy let me know. Nothing interesting or consistent going on on other threads.

This also corrects the property declaration for `ASDisplayNode.subnodes`.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -351,7 +351,7 @@ NS_ASSUME_NONNULL_BEGIN
 /** 
  * @abstract The receiver's immediate subnodes.
  */
-@property (nonatomic, readonly, strong) NSArray<ASDisplayNode *> *subnodes;
+@property (nonatomic, readonly, copy) NSArray<ASDisplayNode *> *subnodes;
 
 /** 
  * @abstract The receiver's supernode.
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
screen shot 2016-03-23 at 10 11 45 pm
screen shot 2016-03-23 at 10 11 35 pm

appleguy:
@Adlai-Holler this is a good change, and we have seen a similarly obscure case at Pinterest crop up just today actually.  My main concern is that this will generate a very large number of retain and release calls, in addition to the array copy itself. Can you do

I would significantly prefer if we can use a non-retaining, efficient data structure.  A simple C-array (getBytes from the NSArray copied into a malloc'd memory area?) would be a lot better, if it is safe... Although it does seem at least conceivable that some transformation triggered by one of the elements could actually result in before the allocation of one of the other objects (seems pretty shady to do, but maybe a subnode calling to a supernode via a delegate, and triggering rasterization enable could technically orphan all of the references to a given sublayer of a sibling node?)

I will merge this change now, but if you have an odd moment to consider if a non-retaining approach would work in all worthwhile to support use cases, let me know!

PR #1427, opened 2016-03-24
[ASCellNode] Expanding scrollView callbacks for cells: willBeginDragging & didEndDragging.

Xinchi:
…ac to ASCellNode instances

### DIFF
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -30,6 +30,14 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
    * Indicates a cell is no longer visible
    */
   ASCellNodeVisibilityEventInvisible,
+  /**
+   * Indicates user has started dragging the visible cell
+   */
+  ASCellNodeVisibilityEventWillBeginDragging,
+  /**
+   * Indicates user has ended dragging the visible cell
+   */
+  ASCellNodeVisibilityEventDidEndDragging,
 };
 
 /**
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -107,6 +107,8 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   BOOL _asyncDataFetchingEnabled;
   BOOL _asyncDelegateImplementsScrollviewDidScroll;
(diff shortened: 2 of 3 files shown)


Xinchi:
This pull request is created for accommodating such needs for Pinterest app: Dismiss the play icon overlay in grid for native video pins when user's finger is on screen while the page is scrolling, and reinsert the play icon overlay once the user lifts off his/her finger but the page may still be scrolling.

AsyncDisplayKit/ASCellNode.h
```diff
@@ -30,6 +30,14 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
    * Indicates a cell is no longer visible
    */
   ASCellNodeVisibilityEventInvisible,
+  /**
+   * Indicates user has started dragging the visible cell
+   */
+  ASCellNodeVisibilityEventWillBeginDragging,
```
appleguy: My main concern with this is that these events will be triggered even if the user is not dragging on the cell itself, which is not particularly clear from the API.  Also, if a cell becomes invisible in the process of the user driving on the screen before their fingerless the screen, will we send a did and dragging to that cell?

That would be very important, as well as the reverse, where a cell that is not yet visible gets a begin call even if the Touch is already on screen and continuing to drag at the moment it becomes visible. Making these calls imbalance could create issues for a variety of clients.

Xinchi:
Very reasonable points there, thanks for bringing them up!  I think it's reasonable to reassess if it's worth it to make the new API more robust to accommodate different situations, or stick with the plan to implement dragging-related event on the application level?   If we are going with the former, I think the triggering should be strictly under 2 conditions (which is how it behaves in this PR): 1) The cell is on screen  2) The finger touches down or lifts off the screen.  The call back will only be triggered when both conditions are met (cells that become visible as user drag won't respond to willBeginDragging, and cells that become invisible won't respond to didEndDragging).    Regarding your first point, I think the same rule from scrollViewDidScroll applies here: the finger touch point doesn't have to be in the cell itself,  aka all cells on screen will respond to this method.  (A comment can be added on the API)

appleguy:
@Xinchi send me a note in slack with the current state of this project. Sorry took a while to get around this, I would like to see what the right solution is now especially if we have not already landed the application level workaround. If that workaround is suitable for now, then don't worry about resuming this

Because it is easy to reopen this if we want to proceed, I'm going to close it out for now, As it can't be cleanly merged with the current master without changes anyway.

PR #1428, opened 2016-03-24
[ASDisplayNode] Add ASSubnodesState

maicki:
￼Introducing the counterpart to ASHierarchyState that represents upward communication of options through the hierarchy.

￼

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -670,7 +670,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   int32_t transitionID = [self _newTransitionID];
   constrainedSize.transitionID = transitionID;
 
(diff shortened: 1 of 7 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2230,11 +2238,48 @@ - (void)exitHierarchyState:(ASHierarchyState)hierarchyState
...
+    _subnodesState = newState;
+  }
+  
+  if (newState != oldState) {
+    LOG(@"setSubnodesState: oldState = %lu, newState = %lu", (unsigned long)oldState, (unsigned long)newState);
+  }
+}
+
+- (void)updateSubnodesStates
+{
```
maicki: Only iterate over all of the subnodes is necessary if we disable the property

appleguy:
@maicki thank you for implementing this. Without having built this, I don't think we would have ever had the idea for the environment. We should first focus on completing the adoption in our application, but will come back to implementing this in the environment later. If you could make sure a task is filed in Ship to track this, that would be perfect.

PR #1430, opened 2016-03-25
[ASPagerNode] Prevent calculation of current index path when the collection view is empty

levi:


AsyncDisplayKit/ASPagerFlowLayout.m
```diff
@@ -54,6 +56,18 @@ - (NSIndexPath *)_indexPathForVisiblyCenteredItem
   return nil;
 }
 
+- (BOOL)_dataSourceIsEmpty
+{
+  if ([self.collectionView numberOfSections]) {
+    if ([self.collectionView numberOfItemsInSection:0] == 0) {
```
appleguy: It might be more ideal to use a direct return with a ternary operator here, as right now the fall through condition is a bit hard to understand because the outer if-else case feels like it should be able to handle all possibilities (since it is not else if).  Would it be equivalent to use this same condition, but return ? YES : NO and removes the return at the end of the method?

AsyncDisplayKit/ASPagerFlowLayout.m
```diff
@@ -25,10 +25,8 @@ - (void)invalidateLayout
 - (CGPoint)targetContentOffsetForProposedContentOffset:(CGPoint)proposedContentOffset
 {
   if (self.currentIndexPath) {
-    CGPoint contentOffset = [self _targetContentOffsetForItemAtIndexPath:self.currentIndexPath
+    return [self _targetContentOffsetForItemAtIndexPath:self.currentIndexPath
                                                    proposedContentOffset:proposedContentOffset];
```
appleguy: nit: Indentation / colons

AsyncDisplayKit/ASPagerFlowLayout.m
```diff
@@ -54,6 +56,11 @@ - (NSIndexPath *)_indexPathForVisiblyCenteredItem
   return nil;
 }
 
+- (BOOL)_dataSourceIsEmpty
+{
+  return ([self.collectionView numberOfSections] > 0 && [self.collectionView numberOfItemsInSection:0] > 0);
```
nguyenhuy: Why is data source empty if the collection view has at least 1 section and 1 row?

levi: LOL thanks for catching that. Was an on the fly edit to the review. Need to fix it.

Issue #1181, opened 2016-02-04
[ASDataController] Cancel node measurement after Table/Collection is deallocated


Adlai-Holler:
This caused an assertion failure for me because the data controller was trying to layout nodes to fit constrained size of (0,0). It's also a potential performance hazard if your user opens a giant feed and then immediately closes it and tries to do something else.

I'm honestly too scared to try to touch this one myself. My instinct would be, in `_layoutNode:withConstrainedSize:` to return early if `_dataSource` is nil. @nguyenhuy Thoughts?

appleguy:
@Adlai-Holler Great find, thanks for reporting this. Certainly makes sense to me, and should not be too hard to fix; I had just never thought about this case, even though it is relatively obvious.

@rahul-malik, @lkzhao this is also close to code that you have recently changed, so if you feel motivated, it would be nice to fix this one soon.

appleguy:
cc @maicki this may also fit into your work.

appleguy:
@Adlai-Holler: Let us know if the attached PR addresses this to your satisfaction. I'm not certain that it will prevent background CPU usage after navigating away from a view, but it should address the assertion failure.

Adlai-Holler:
Yep! This is great stuff.

PR #1432, opened 2016-03-25
[ASDataController] Refactor ASDataController layout nodes methods

maicki:
- Refactor ASDataController layout nodes methods
- Don't crash if dataSource becomes nil while layout is happening

@appleguy @nguyenhuy Could you guys take a look, it should address for now: #1181

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -129,7 +129,8 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
 
 - (void)layoutLoadedNodes:(NSArray<ASCellNode *> *)nodes fromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofKind:(NSString *)kind
 {
-  [self _layoutNodes:nodes onMainThread:YES fromContexts:contexts ofKind:kind];
+  NSAssert(ASDisplayNodeThreadIsMain(), @"Layout of loaded nodes must happen on the main thread.");
+  [self _layoutNodes:nodes fromContexts:contexts ofKind:kind];
 }
 
 /**
@@ -155,20 +156,20 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
 /**
  * Perform measurement and layout of loaded or unloaded nodes based if they will be layed out on main thread or not
  */
-- (void)_layoutNodes:(NSArray<ASCellNode *> *)nodes onMainThread:(BOOL)onMainThread fromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofKind:(NSString *)kind
+- (void)_layoutNodes:(NSArray<ASCellNode *> *)nodes fromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofKind:(NSString *)kind
 {
   ASDisplayNodeAssertTrue(nodes.count == contexts.count);
-  NSAssert(NSThread.isMainThread == onMainThread, @"Main thread layout must be on the main thread.");
   
   if (_dataSource == nil) {
     return;
   }
   
+  BOOL isMainThread = ASDisplayNodeThreadIsMain();
   NSInteger idx = 0;
   for (ASIndexedNodeContext *context in contexts) {
     ASCellNode *node = nodes[idx];
     // Only nodes that are loaded should be layout on the main thread
-    if (node.isNodeLoaded == onMainThread) {
+    if (node.isNodeLoaded == isMainThread) {
       [self _layoutNode:node withConstrainedSize:context.constrainedSize];
     }
     idx++;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -160,9 +152,32 @@ - (void)_batchLayoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)context
   }];
 }
 
+/**
+ * Perform measurement and layout of loaded or unloaded nodes based if they will be layed out on main thread or not
+ */
+- (void)_layoutNodes:(NSArray<ASCellNode *> *)nodes onMainThread:(BOOL)onMainThread fromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofKind:(NSString *)kind
```
nguyenhuy: I'm wondering if we can remove `onMainThread` parameter? Instead, we can check `node.isNodeLoaded == NSThread.isMainThread` and only layout a particular node if it passes this test. With that, this method measures loaded nodes if it is called on main thread and unloaded ones if it is called on a background thread. This is exactly the way we are using it.

maicki: I think that makes sense. Changed the behavior.

nguyenhuy:
Great! LGTM

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -129,15 +129,8 @@ - (void)batchLayoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts
 
 - (void)layoutLoadedNodes:(NSArray<ASCellNode *> *)nodes fromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofKind:(NSString *)kind
 {
-  NSAssert(NSThread.isMainThread, @"Main thread layout must be on the main thread.");
-  ASDisplayNodeAssertTrue(nodes.count == contexts.count);
```
appleguy: Should we keep this assertion? Maybe it is not important.

maicki: I moved it in _layoutNodes:fromContexts:ofKind:

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -160,9 +153,32 @@ - (void)_batchLayoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)context
...
+  if (_dataSource == nil) {
+    return;
+  }
+  
+  BOOL isMainThread = ASDisplayNodeThreadIsMain();
+  NSInteger idx = 0;
+  for (ASIndexedNodeContext *context in contexts) {
+    ASCellNode *node = nodes[idx];
+    // Only nodes that are loaded should be layout on the main thread
+    if (node.isNodeLoaded == isMainThread) {
```
appleguy: We should expand this comment. This code is quite creative, but for that reason it is just slightly difficult to read. It is not immediately clear that we intend to perform layout on unloaded nodes when we are not on the main thread.  A comment could make this more clear.

The comment should probably say that for any given layout pass that occurs, this method will be called at least twice, once on the main thread and the background, to result in complete coverage of both loaded and unloaded nodes

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -200,25 +216,27 @@ - (void)_layoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofK
...
-        ASCellNode *node = allocatedNodes[k];
-        // Only measure nodes whose views aren't loaded, since we're in the background.
-        // We should already have measured loaded nodes before we left the main thread, using layoutLoadedNodes:ofKind:atIndexPaths:
-        if (!node.isNodeLoaded) {
-          [self _layoutNode:node withConstrainedSize:contexts[k].constrainedSize];
-        }
-      }
+      // We should already have measured loaded nodes before we left the main thread. Layout the remaining once on a background thread.
+      NSRange range = NSMakeRange(j, batchCount);
+      [self _layoutNodes:[allocatedNodes subarrayWithRange:range]
```
appleguy: Are these two new usages of sub arrays?

appleguy: I think it would be best to have the method support taking the full arrays, And then pass the range as an additional argument. This pattern is followed in a few other areas, if I recall correctly, and is effective at reducing array copy costs which include memory management calls to retain and release each object in the array.

In the near future, we will start adjusting the size of layout batches and may even provide this as an external API for applications to tune based on the average height of their content, and for smaller batch sizes these copies are likely to add up.

Overall though, this is a really fantastic change, and much welcome to this part of the code base that can benefit greatly from this kind of attention to detail.

maicki: Added an extra range parameter to the internal layout nodes method.

AsyncDisplayKit/Details/ASDataController.mm
maicki: Added a new comment

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -200,25 +217,30 @@ - (void)_layoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofK
...
+                  ofKind:kind];
+      
     } else {
       allocationBlock();
       [_mainSerialQueue performBlockOnMainThread:^{
-        [self layoutLoadedNodes:subarray fromContexts:[contexts subarrayWithRange:NSMakeRange(j, batchCount)] ofKind:kind];
+        [self _layoutNodes:subarray
+              fromContexts:contexts
+          inIndexesOfRange:NSMakeRange(0, subarray.count)
+                    ofKind:kind];
```
maicki: @appleguy What do you think creating also a helper method that does not have the index range as parameter that just calls through the one with index range as parameter with the whole nodes range as we here do?

appleguy:
I thought I had merged this!  Sorry, this is definitely a real and material improvement to the code quality and readability of ASDataController :).  Thanks @maicki !

Issue #1434, opened 2016-03-26
[ASEditableTextNode] Support wrapping UITextField (instead of just UITextView).  Required for secureTextEntry.


huyle90:
Hi, I am just starting to use AsyncDisplayKit recently and I want to implement ASEditableNode as a password text field.  But I have a problem with setting securedTextEntry of ASEditableNode ( .textView) Does anyone know why the text view of ASEditableNode does not support UITextInputTraits protocol in this case?

hannahmbanana:
@huyle90 - Thanks for your question! This should already be supported by accessing the textView property because the ASEditableTextNode internally contains a regular UITextView. 

In what way does it not work? Is the textView nil? could you provide the code you are using or a sample project?

appleguy:
@huyle90 it would be pretty easy to add the text input traits protocol to the editable text node. However this would simply forward those properties to the internal text view, and it should be easy for you to apply those properties to that text view directly.

If you can give us any further information on the symptoms you're seeing and what you have tried, that would be useful. By far the best thing would be a small test project that can reproduce the issue, and with that we would be able to fix it quite quickly. It is definitely intended to be supported.

huyle90:
@appleguy  @hannahmbanana textView is not nil since I have instantiated the ASEditableTextNode. I set all properties needed and they are fine except secureTextEntry doesnt show password char (•).  This  somehow didnt work for UITextView I guess.
passwordTextFieldNode.textView.secureTextEntry = true

wpK:
I'm having the same issue. Here is a code sample: https://gist.github.com/wpK/40551dbde865b277cc88 with the following result:

![securetextentry](https://cloud.githubusercontent.com/assets/243156/14091563/f24d3ea0-f50e-11e5-9882-e818fbd90d6e.jpg)

appleguy:
@huyle90 making sure I understand your comment - are you saying you also tried UITextView and that didn't work either?

It is certainly possible that secure text is only supported by UITextField, not UITextView, as the rendering engines are completely different.  But I simply don't know for sure if this limitation is true, even though it seems plausible from what I remember about their implementation back when I worked on that code for iOS 1.0-4.3.

[...]

PR #1437, opened 2016-03-26
[ASEditableTextNode] Allow TextKit component customization

ejensen:
This change opens up the option of using NSLayoutManager & NSTextStorage subclass within ASEditableTextNodes.

### DIFF
--- a/AsyncDisplayKit/ASEditableTextNode.h
+++ b/AsyncDisplayKit/ASEditableTextNode.h
@@ -7,6 +7,7 @@
  */
 
 #import <AsyncDisplayKit/ASDisplayNode.h>
+#import <AsyncDisplayKit/ASTextKitHelpers.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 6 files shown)


appleguy:
@ejensen interesting! What use cases have you envisioned for this?

Would you be interested in adding forwarding methods for UITextInputTraits, potentially resolving https://github.com/facebook/AsyncDisplayKit/issues/1434 for @huyle90 's use case with a cleaner API?

Let me know if you have a chance to look into the build failures. I did rerun the build, but is still not passing.

ejensen:
@appleguy My particular use case is creating editable [un]ordered lists. I use a custom NSLayoutManager to generate the number / bullet glyphs. This is in contrast to inserting the characters and indentation into the attributed string. Which would interfere with typing and selection of the text.

simulator screen shot

Besides my use case, opening up LayoutManager / TextContainer for customization allows for all the TextKit goodies such as exclusion paths.
textkit

I'll look into adding UITextInputTraits to the ASEditableTextNode's interface. It will be in a separate PR if that's OK?

AsyncDisplayKit/ASEditableTextNode.h
```diff
@@ -18,6 +19,17 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface ASEditableTextNode : ASDisplayNode
 
+/**
+ * @abstract Initializes a editable text node with a provided TextKit stack.
```
appleguy: My main concern with this change is that it makes it quite unclear that it is possible to create this object with a standard alloc / init.

Because this method is for a very specialized use case, I would suggest putting it at the very end of the file even though it is against typical convention to put initializers there.  I'm not sure if there is a way to re-declare -init (I tried this recently in a similar case and got a compiler error, but did zero investigation on exactly why)

Let me know if you have any thoughts or opinions on the best way to handle this, but it is definitely a priority for me to make sure the header files remain approachable and intuitive for simple use cases while providing advanced extensibility when desired.

ejensen:
No problem. The standard `init` method is now declared in the header. I didn't run into any errors doing so.

Issue #1436, opened 2016-03-26
[ASDisplayNode] Possible Crash when Responding to Visibility Events Triggered by Dealloc


Adlai-Holler:
In `-[ASDisplayNode dealloc]` we clear the visibility interface state. Updating that interface state will eventually call into client code, which may cause crashes if the client attempts to establish a reference to the deallocating node. 

In my case, ASCellNode deallocated, which sent .Invisible through the cell node callback, where my code called `setNeedsLayout`, which caused a crash inside ASDK when it attempted to establish a new reference to the node. As a workaround I no longer call `setNeedsLayout` on `.Invisible`. 

I don't really know the fix here, but I know it seems weird to clear that bit during dealloc, since before the node is deallocated it must have been removed from its hierarchy, right? How could a node that's visible be deallocated?

appleguy:
Thanks for reporting!  https://github.com/facebook/AsyncDisplayKit/pull/1441

PR #1441, opened 2016-03-27
[ASDisplayNode] Alternative implementation of interfaceState -> invisible that avoids a subclass call from -dealloc.

appleguy:
This resolves https://github.com/facebook/AsyncDisplayKit/issues/1436

cc @Adlai-Holler, @maicki, @nguyenhuy, @garrettmoon, @rahul-malik

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -337,9 +337,9 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (void)dealloc
 {
   ASDisplayNodeAssertMainThread();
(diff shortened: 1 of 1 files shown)


appleguy:
This would greatly benefit from a review, but I am landing it now to be able to included in the current release of the framework in case other clients have code in the visibility handler that is not safe during deallocation.

I've tested it fairly well, and by far my greatest concern is the performance overhead around 100 dispatch blocks during navigation events.  This could be significantly reduced with a smarter implementation, but to prioritize this I will first evaluate its magnitude with a profiler. I'm confident that it will not cause any directly noticeable performance problems, but it is likely within the range that matters for core framework functionality.

Adlai-Holler:
Awesome, thanks for the fix!

PR #1454, opened 2016-03-29
[ASLayoutableContext] Properly lock layoutableContextMap

nguyenhuy:


maicki:
@huy LGTM

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -630,9 +630,12 @@ - (BOOL)shouldMeasureWithSizeRange:(ASSizeRange)constrainedSize
     return NO;
   }
   
-  if (ASHierarchyStateIncludesLayoutPending(_hierarchyState)
-      && _pendingTransitionID != ASLayoutableGetCurrentContext().transitionID) {
-    return NO;
+  if (ASHierarchyStateIncludesLayoutPending(_hierarchyState)) {
+    ASLayoutableContext context =  ASLayoutableGetCurrentContext();
+    ASDisplayNodeAssert(!ASLayoutableContextIsNull(context), @"Context must not be null for a pending layout subnode");
```
appleguy: Is this safe in production, if this case actually occurs? Should we initialize it to a blank structure and then check, Or will the context return a blank structure if it is not set?

AsyncDisplayKit/Layout/ASLayoutable.mm
```diff
@@ -27,14 +52,24 @@ static inline mach_port_t ASLayoutableGetCurrentContextKey()
...
+  ASDN::StaticMutexLocker l(_layoutableContextLock);
+  layoutableContextMap[key] = context;
 }
 
 struct ASLayoutableContext ASLayoutableGetCurrentContext()
 {
-  return layoutableContextMap[ASLayoutableGetCurrentContextKey()];
+  mach_port_t key = ASLayoutableGetCurrentContextKey();
+  ASDN::StaticMutexLocker l(_layoutableContextLock);
+  auto it = layoutableContextMap.find(key);
```
appleguy: Can you comment this code? It is not clear to me why we return the second, but I am not very familiar with this pattern. I assume this is an iterator though. This might be the only use of the auto keyword in the CodeBase

AsyncDisplayKit/Layout/ASLayoutable.mm
```diff
@@ -27,14 +52,24 @@ static inline mach_port_t ASLayoutableGetCurrentContextKey()
...
+  if (it != layoutableContextMap.end()) {
+    return it->second;
+  }
+  return ASLayoutableContextNull;
 }
 
 void ASLayoutableClearCurrentContext() {
```
appleguy: { newline

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: You're right, it's actually not safe assert. Converting to an if-else check.

AsyncDisplayKit/Layout/ASLayoutable.mm
nguyenhuy: Adding...

PR #1456, opened 2016-03-29
[ASEnvironment ] Add support for ASEnvironment

maicki:


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -60,7 +60,7 @@ - (void)_staticInitialize;
 @implementation ASDisplayNode
 
 // these dynamic properties all defined in ASLayoutOptionsPrivate.m
-@dynamic spacingAfter, spacingBefore, flexGrow, flexShrink, flexBasis, alignSelf, ascender, descender, sizeRange, layoutPosition, layoutOptions;
+@dynamic spacingAfter, spacingBefore, flexGrow, flexShrink, flexBasis, alignSelf, ascender, descender, sizeRange, layoutPosition;
```
appleguy: Fantastic to eliminate the most common type of object creation in the framework!

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2664,6 +2668,67 @@ - (void)_enqueueAsyncSizingWithSentinel:(ASSentinel *)sentinel completion:(void(
   return self;
 }
 
+#pragma mark - ASEnvironment
+
+- (ASEnvironmentCollection *)environmentCollection
+{
+  return &_environmentCollection;
```
appleguy: For the next round of work on this, top priority is to make these kind of accesses threadsafe.  The only way I know how to do that is ensure we never return a pointer to the structure (including _any_ sub-structure of the environment), or there isn't a way to lock access to it.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2664,6 +2668,67 @@ - (void)_enqueueAsyncSizingWithSentinel:(ASSentinel *)sentinel completion:(void(
...
+}
+
+- (ASDisplayNode *)parent
+{
+  return self.supernode;
+}
+
+- (void)setParent:(ASDisplayNode *)parent
```
appleguy: I don't think we should support a setParent: method in ASEnvironment.  Will it still work if the protocol only guarantees that an accessor method exists?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2664,6 +2668,67 @@ - (void)_enqueueAsyncSizingWithSentinel:(ASSentinel *)sentinel completion:(void(
...
+{
+  [self __setSupernode:parent];
+}
+
+- (NSArray<ASDisplayNode *> *)children
+{
+  return self.subnodes;
+}
+
+- (BOOL)supportsMultipleChildren
```
appleguy: We should find a better name.  Until then, comment here that this really is intended to reference the fact that upward propogation of ASLayoutable values from ASLayoutSpec to ASDisplayNode is permitted - and that Nodes have "only one layoutspec child", even though clearly may have multiple subnodes.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2664,6 +2668,67 @@ - (void)_enqueueAsyncSizingWithSentinel:(ASSentinel *)sentinel completion:(void(
...
+{
+  return NO;
+}
+
+
+#pragma mark - ASLayoutableExtensibility
+
+- (void)setLayoutOptionExtensionBool:(BOOL)value atIndex:(int)idx
```
appleguy: Should we implement at least these methods with a #define in the same way that the ASLayoutable property accessors are?  In fact, this could go in ASLayoutable.mm ... since ASStackLayoutable is handled the same way, and is comparable to ASLayoutableExtensibility.

AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -0,0 +1,103 @@
...
+
+#import "ASDimension.h"
+#import "ASStackLayoutDefines.h"
+#import "ASRelativeSize.h"
+
+
+ASDISPLAYNODE_EXTERN_C_BEGIN
+NS_ASSUME_NONNULL_BEGIN
+
+static const int kMaxEnvironmentStateBoolExtensions = 1;
```
appleguy: Good, I think these are the right defaults for now.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1853,6 +1856,7 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
   if (_methodOverrides & ASDisplayNodeMethodOverrideLayoutSpecThatFits) {
     ASLayoutSpec *layoutSpec = [self layoutSpecThatFits:constrainedSize];
     layoutSpec.isMutable = NO;
+    layoutSpec.parent = self;
```
maicki: Move that before isMutable is set

AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * ASEnvironment allows objects that conform to the ASEnvironment protocol to be able to propagate specific States
+ * defined in an ASEnvironmentCollection up and down the ASEnvironment tree. To be able to define how merges of
+ * States should happen, specific merge functions can be provided
+ */
+@protocol ASEnvironment <NSObject>
+
+/// The environment collection of an object which class conforms to the ASEnvironment protocol
+- (ASEnvironmentCollection *)environmentCollection;
```
maicki: Rename collection to state (more similar to a struct than array, matches interfacestate etc

AsyncDisplayKit/Private/ASEnvironmentInternal.h
```diff
@@ -0,0 +1,69 @@
...
+
+template <typename ASEnvironmentStateType>
+void ASEnvironmentStatePropagateDown(id<ASEnvironment> object, ASEnvironmentStateType& state) {
+  ASEnvironmentPerformBlockOnObjectAndChildren(object, ^(id<ASEnvironment> node) {
+    ASEnvironmentMergeObjectAndState(object, state, DOWN);
+  });
+}
+
+template <typename ASEnvironmentStateType>
+void ASEnvironmentStatePropagateUp(id<ASEnvironment> object, ASEnvironmentStateType& state) {
```
maicki: Take in a ASEnvironmentState and return a new one

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -128,6 +145,52 @@ - (NSArray *)children
   return self.layoutChildren[kDefaultChildrenKey];
 }
 
+
+#pragma mark - ASEnvironment
+
+- (ASEnvironmentCollection *)environmentCollection
```
maicki: Move this as instance variable and implement the defines in ASLayoutSpec and ASDisplayNode

AsyncDisplayKit/Layout/ASLayoutable.mm
```diff
@@ -38,3 +45,183 @@ struct ASLayoutableContext ASLayoutableGetCurrentContext()
...
+}\
+- (void)propagateUpLayoutOptionsState\
+{\
+  id<ASEnvironment> parent = [self parent];\
+  if (![parent supportsMultipleChildren]) {\
+    ASEnvironmentStatePropagateUp(parent, self.environmentCollection->layoutOptionsState);\
+  }\
+}\
+\
+- (CGFloat)spacingAfter\
```
maicki: All of this properties needs to be locked with the property locks of the classes:

Setters: instance variable .lock() and .unlock()
Getter: MutexLockers

AsyncDisplayKit/ASDisplayNode.mm
maicki: Maybe rename to supportsUpwardPropagation

AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -0,0 +1,103 @@
...
+  unsigned rangeManaged:1; // = NO
+  unsigned transitioningSupernodes:1; // = NO
+  unsigned layoutPending:1; // = NO
+} ASEnvironmentHierarchyState;
+extern ASEnvironmentHierarchyState ASEnvironmentHierarchyStateCreate();
+
+
+#pragma mark - ASEnvironmentCollection
+
+typedef struct ASEnvironmentCollection {
```
maicki: Rename to ASEnvironmentState

AsyncDisplayKit/Details/ASEnvironment.h
maicki: Remove this from the protocol

AsyncDisplayKit/Layout/ASLayoutable.mm
```diff
@@ -38,3 +45,183 @@ struct ASLayoutableContext ASLayoutableGetCurrentContext()
...
+ *  allows a custom spec to set up defaults without needing to alter the layoutable itself. For example,
+ *  image you were creating a custom baseline spec that needed ascender/descender. To assign values automatically
+ *  when a text node's attribute string is set, you would need to subclass ASTextNode and assign the values in the
+ *  override of setAttributeString. However, assigning the defaults via this function allows you to create a
+ *  custom spec without the need to create a subclass of ASTextNode.
+ *
+ *  @param layoutable The layoutable object to inspect for default intrinsic layout option values
+ */
+void ASLayoutableSetValuesForLayoutable(id<ASLayoutable> layoutable)
+{
```
maicki: Move all of this logic in the concrete subclasses like ASDisplayNode, ASTextNode and add it to setPreferredFrameSize (ASDisplayNode) and setAttributedString (ASTextNode)

AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -0,0 +1,103 @@
...
+@protocol ASEnvironment <NSObject>
+
+/// The environment collection of an object which class conforms to the ASEnvironment protocol
+- (ASEnvironmentCollection *)environmentCollection;
+
+/// Returns the parent of an object which class conforms to the ASEnvironment protocol
+- (id<ASEnvironment>)parent;
+
+/// Set the parent of an object which class conforms to the ASEnvironment protocol
+- (void)setParent:(id<ASEnvironment>)parent;
```
maicki: Remove this from protocol

AsyncDisplayKit/Layout/ASLayoutable.mm
maicki: Add recursive mutext to ASLayoutSpec

AsyncDisplayKit/Private/ASEnvironmentInternal.h
maicki: Take in a ASEnvironmentState and return a new one

AsyncDisplayKit/Private/ASEnvironmentInternal.mm
```diff
@@ -0,0 +1,149 @@
...
+
+
+#pragma mark - Merging functions for states
+
+void ASEnvironmentMergeObjectAndState(id<ASEnvironment> object, ASEnvironmentHierarchyState& state, ASEnvironmentStatePropagation propagation) {
+    // Merge object and hierarchy state
+  LOG(@"Merge object and state: %@ - ASEnvironmentHierarchyState", object);
+}
+
+void ASEnvironmentMergeObjectAndState(id<ASEnvironment> object, ASEnvironmentLayoutOptionsState& state, ASEnvironmentStatePropagation propagation) {
```
maicki: Take in a ASEnvironmentState and return a new one

AsyncDisplayKit/Private/ASEnvironmentInternal.h
```diff
@@ -0,0 +1,69 @@
...
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "ASEnvironment.h"
+
+#pragma once
+
+enum ASEnvironmentStatePropagation { DOWN, UP };
```
levi: Use NS_ENUM

maicki: Actually I wanted to use an enum class so it's strongly typed. Will change that to either an enum class or NS_ENUM

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1700,6 +1702,10 @@ - (void)__setSupernode:(ASDisplayNode *)newSupernode
       [self exitHierarchyState:stateToEnterOrExit];
     }
   }
+    
+  if ([newSupernode supportsUpwardPropagation]) {
```
appleguy: This looks great. We may not need to revisit this API, the new name is pretty strong for long-term use unless we encounter a new case that is unusual to implement with it

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1918,6 +1925,7 @@ - (void)setPreferredFrameSize:(CGSize)preferredFrameSize
   ASDN::MutexLocker l(_propertyLock);
   if (! CGSizeEqualToSize(_preferredFrameSize, preferredFrameSize)) {
     _preferredFrameSize = preferredFrameSize;
+    self.sizeRange = ASRelativeSizeRangeMake(ASRelativeSizeMakeWithCGSize(_preferredFrameSize), ASRelativeSizeMakeWithCGSize(_preferredFrameSize));
```
appleguy: What if this value is set directly? Should we definitely override size range, or should it be considered an error to set both of these properties? I filed a ship task number 72 to track clarifying and hopefully unifying these APIs.  @nguyenhuy - not a priority, but in 2 weeks or so it would be great to have your input on where you see the right direction for these APIs, and we can do the implementation for you with our SF team - want to see you focused on the key priorities you have coming up!

AsyncDisplayKit/Details/ASEnvironment.m
```diff
@@ -0,0 +1,41 @@
...
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import "ASEnvironment.h"
+
+ASEnvironmentLayoutOptionsState ASEnvironmentLayoutOptionsStateCreate()
+{
+  return (ASEnvironmentLayoutOptionsState) {
+    .spacingBefore = 0,
```
appleguy: Do we need this default here, And if so, why not spacing after or others? It seems fine to include everything explicitly, but I am not sure that it is required either. Happy to go with whatever your intuition prefers.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -75,7 +81,10 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
 
     id<ASLayoutable> finalLayoutable = [child finalLayoutable];
     if (finalLayoutable != child) {
-      [finalLayoutable.layoutOptions copyFromOptions:child.layoutOptions];
+      // Copy layout options
+      ASEnvironmentState environmentState = finalLayoutable.environmentState;
+      environmentState.layoutOptionsState = child.environmentState.layoutOptionsState;
```
appleguy: I am not completely certain, but I believe we should perform a merge operation here, treating the final layoutable as though it were a parent and merging upward the layout options from the current child.

I think this should have been even in the case of iterating through multiple children, each time a final layoutable is created, check if the object returned from final layoutable is different then the object is called on, then perform an upward merge.

The reason is that there may be properties specified on the final layout double that are specifically intended to override properties of the object it is substituting in for.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -128,6 +146,27 @@ - (NSArray *)children
...
+}
+
+- (void)setEnvironmentState:(ASEnvironmentState)environmentState
+{
+  _environmentState = environmentState;
+}
+
+- (BOOL)supportsUpwardPropagation
+{
+  return YES;
```
appleguy: Add a comment that says that two subclasses override this method to return NO, because upward propagation is not enabled if a layout specification has more than one child, and these are currently the specifications that are known to have more than one.

AsyncDisplayKit/Layout/ASLayoutableExtensibility.h
```diff
@@ -0,0 +1,25 @@
+//
+//  ASLayoutableExtensibility.h
+//  AsyncDisplayKit
+//
+//  Created by Michael Schneider on 3/29/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
```
appleguy: This may not be relevant, but if you can locate the header that edge insets is defined in, let's try using only that header to more specifically document the dependencies

AsyncDisplayKit/Layout/ASLayoutablePrivate.h
```diff
@@ -60,9 +59,202 @@ extern void ASLayoutableClearCurrentContext();
...
+- (CGFloat)spacingAfter\
+{\
+  ASDN::MutexLocker l(_propertyLock);\
+  return _environmentState.layoutOptionsState.spacingAfter;\
+}\
+\
+- (void)setSpacingAfter:(CGFloat)spacingAfter\
+{\
+  _propertyLock.lock();\
+  _environmentState.layoutOptionsState.spacingAfter = spacingAfter;\
```
appleguy: Fantastic, all of these methods are what I had hoped for in terms of a great balance between efficiency and clarity. I think this implementation will last a long time without any modification or optimization.

Indeed, the performance of these methods will greatly exceed that of the node bridged properties, especially because calling into the underlying Apple frameworks adds even more overhead on top of the MutexLockers in all of the setter methods.

AsyncDisplayKit/Layout/ASLayoutablePrivate.h
```diff
@@ -60,9 +59,202 @@ extern void ASLayoutableClearCurrentContext();
...
+  _propertyLock.unlock();\
+}\
+
+
+#pragma mark - ASLayoutableExtensibility
+
+#define ASEnvironmentLayoutExtensibilityForwarding \
+- (void)setLayoutOptionExtensionBool:(BOOL)value atIndex:(int)idx\
+{\
+  _ASEnvironmentLayoutOptionsExtensionSetBoolAtIndex(self, idx, value);\
```
appleguy: We may need locking these methods, I don't have time to confirm right now, but can analyze more closely tomorrow.

AsyncDisplayKit/Details/ASEnvironment.m
maicki: I removed the default values but didn't remove the creation functions in case we want to have default values in the future.

AsyncDisplayKit/Details/ASEnvironment.m
```diff
@@ -0,0 +1,33 @@
...
+}
+
+ASEnvironmentHierarchyState _ASEnvironmentHierarchyStateCreate()
+{
+  return (ASEnvironmentHierarchyState) {
+    // Default values can be defined in here
+  };
+}
+
+ASEnvironmentState ASEnvironmentStateCreate()
```
appleguy: Maybe rename these from Create() -> MakeDefault()?

appleguy:
The performance and architectural scalability advantages this pattern provides are truly impressive.  We're literally bit-packing here, without breaking a sweat, and passing state updates far more efficiently than even the mature UIView does.

The kind of attention to detail that went into this patch is exactly why AsyncDisplayKit is so fast, even though it accomplishes even more than UIView's recursive window and subview notifications that are so costly!

Issue #1431, opened 2016-03-25
[ASStackLayoutSpec] Omit subnodes that are marked .hidden (similar to UIStackView).


wy19901227:
Hi,  In using ASDK's layout I should not to caculate size of a cellNode's subNode,Just like I use autolayout,When I use UIStackView,I set a stackView's contrain or hidden, the stackView will also change frame, So , if I set a cellNode's subNode hidden,can cellNode change in the same time?

appleguy:
@wy19901227 thank you for filing this issue. Is a good suggestion to automatically leave out children in a stack that are marked as hidden.  Was not aware that the new IOS 9 stack component has similar behavior.

For now, there is a good way to accomplish this. We are going to enabled by default (soon / before 2.0) a new framework feature called implicit hierarchy management. You can turn this on for your own node now, and it allows controlling the children in a stack automatically based on what you would turn in the layout spec.

It does not look at the hidden property, but rather automatically adds or removes sub nodes for you according to what is actually in the layout specification.

hannahmbanana:
@wy19901227 - To enable implicit hierarchy management, do the following 3 things:
- `#import "ASDisplayNode+Beta.h"`
- add `self.usesImplicitHierarchyManagement = YES;` to your `init:` method
- any `addSubnode:` calls are no longer needed when using this option, unless you need to manually position the node in `layout` and cannot position the node in `layoutSpecThatFits:`

The file below demonstrates this. `_playgroundNode` is added using implicitHierarchyManagement (no call to `addSubnode:`) while `_resizeHandle` is manually laid out in `layout` and so requires a call to `addSubnode:`

[PlaygroundContainerNode copy.txt](https://github.com/facebook/AsyncDisplayKit/files/190601/PlaygroundContainerNode.copy.txt)

wy19901227:
can offer a demo about this?

hannahmbanana:
@wy19901227 - do you mean offer a demo about implicit hierarchy management? If so, checkout out PR #1457 - it uses implicit hierarchy management in the PhotoCellNode.m file. An official tutorial is coming soon.

Was your question answered? If so, shall we close this issue?

wy19901227:
of course. thank you.

PR #1457, opened 2016-03-29
[New Example] 500pxgram - a social app to directly compare UIKit vs. ASDK performance

hannahmbanana:
Would the ASDK community be interested in a new social app example that directly compares a photo feed implemented with UIKit to an identical feed implemented with ASDK? The app features an infinitely scrolling home feed that demonstrates ASDK's smoother scrolling performance. It also has a significantly sized code base to demonstrate how much less code it takes to design apps using ASDK.

screen shot 2016-03-29 at 3 15 24 pm
If you're interested in including this in the repository, I have a few questions as to how I could improve the sample:

Code Questions:
- could you check the files in the ASDK folders to make sure that I have correctly implemented the ASDK feed?
- do you see any ways to make the shared UIKit and ASDK code implemented in the separate UIKit and ASDK files shared so that we can see a more direct comparison of strictly UIKit code vs ASDK implementation code?

Example App Questions:
- are there any ways to make the comparison between ASDK and UIKit more fair? I'm thinking I should clear the data model when switching between tabs.
- are there any ways to quantify the comparison between ASDK and UIKit?
- are there any ASDK features I could add to make this a better sample app?


wy19901227:
image

wy19901227:
When I set cellNode's subNode hidden,Can the cellNode update itsSelf?

appleguy:
@wy19901227 this was recently discussed an issue, but no, all you need to do is use implicit hierarchy management and ensure that you do not add the sub node to the children array of the stack layout if you would like it to be removed. As long as you do this, a simple one line call for needing layout is enough to completely resize the cell - even animated, or optionally not animated.  Use the transition API if you have to animate the behavior of multiple subnodes.

wy19901227:
I'm sorry ,I try hard，but didn't success,So can you show me the code?

appleguy:
email me and join our slack room!  Hannah can help you there.  You can see examples of implicit hierarchy management in the 500px app that is currently up in a pull request.  It even resizes the cell when the comments asynchronously load for each - calls -setNeedsLayout on itself and adds more items to its layout spec stack.

wy19901227:
Yeah,I have the 500px demo,Just I'm not get how to use it,I'll look again.

hannahmbanana:
@wy19901227 - I'd be happy to help either here or on slack. 

Were you able to follow the three steps I outlined in #1431?

guerciotti:
Hannah this is a great idea, just seeing this now but been thinking of this
sorta thing for a bit. One thought would be to work in a comparison of
sophisticated Text handling.

Congrats, this is a great example.

examples/ASDKgram/Sample/PhotoFeedNodeController.m
```diff
@@ -0,0 +1,158 @@
...
+#pragma mark - ASTableDataSource methods
+
+- (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section
+{
+  return [_photoFeed numberOfItemsInFeed];
+}
+
+- (ASCellNode *)tableView:(ASTableView *)tableView nodeForRowAtIndexPath:(NSIndexPath *)indexPath;
+{
+  PhotoCellNode *cellNode = [[PhotoCellNode alloc] initWithPhotoObject:[_photoFeed objectAtIndex:indexPath.row]];
```
appleguy: Could you use the node block API here?

examples/ASDKgram/Sample/PhotoFeedNodeController.m
```diff
@@ -0,0 +1,158 @@
...
+// table automatic tail loading
+-(void)scrollViewDidScroll:(UIScrollView *)scrollView
+{
+  CGFloat currentOffSetY = scrollView.contentOffset.y;
+  CGFloat contentHeight  = scrollView.contentSize.height;
+  CGFloat screenHeight   = [UIScreen mainScreen].bounds.size.height;
+  
+  CGFloat screenfullsBeforeBottom = (contentHeight - currentOffSetY) / screenHeight;
+  if (screenfullsBeforeBottom < AUTO_TAIL_LOADING_NUM_SCREENFULS) {
+    [self loadPage];
```
appleguy: This is cool, but the framework supports automatic batch loading - please adopt ASBatchContext

examples/ASDKgram/Sample/CommentFeedModel.m
```diff
@@ -0,0 +1,196 @@
...
+          _currentPage = [[response valueForKeyPath:@"current_page"] integerValue];
+          _totalPages  = [[response valueForKeyPath:@"total_pages"] integerValue];
+          _totalItems  = [[response valueForKeyPath:@"total_items"] integerValue];
+          
+          NSArray *comments = [response valueForKeyPath:@"comments"];
+          
+          if ([comments isKindOfClass:[NSArray class]]) {
+            
+            NSUInteger numComments = [comments count];
+            if (numComments > 3) {
```
vtsoup: Can we change this to a defined constant, to match those used in CommentView.m and CommentsNode.m? I noticed that changing the number of comments shown from 3 to 5 to 10 to 15, etc. caused the UIKit version to get progressively worse while the ASDK version stayed the same.

vtsoup: Can we change this to a defined constant, to match those used in CommentView.m and CommentsNode.m? I noticed that changing the number of comments shown from 3 to 5 to 10 to 15, etc. caused the UIKit version to get progressively worse while the ASDK version stayed the same.

hannahmbanana: Thanks for catching this bug! I'll fix it tomorrow. :) Just a warning - I'm new to iOS and ASDK, so there's no guarantee that this works correctly.

vtsoup: Haha. I am in the same boat when it comes to being new to both iOS and ASDK. :) 

I have been ask to present ASDK to the senior IOS devs in two weeks, so your sample came at the right time. The first impression from running it on my 6s+ was that was that UIKit version looked perceivably the same. FPS got to the 30s once in a while, but not too noticeable to the eye. So of course I had to see what I could do to make the app more real world with UIKit being slower than ASDK. ;)

hannahmbanana: If you email @appleguy at AsyncDisplayKit@gmail.com he can add you to the ASDK slack channel. I’m sure he'll be able to give you a much more complete comparison than my app currently does. I’m also helping upgrade the ASDK documentation right now, so if you have any questions as you put together your presentation, feel free to ping me on slack (@hannah).

- As far as my app goes, since it hasn't been fully vetted by @appleguy yet, I'm not sure it is a completely fair comparison of ASDK vs UIKit (but I hope to soon make it so). 
- I think that I need to make it so that the dataModel clears itself when switching between tabs. As it stands, it isn’t a fair comparison, because the app opens on the ASDK tab and by the time you switch to the UIKit tab, the dataModel has already downloaded a bunch of data.  
- I plan on implementing a collectionView in the next few days which will much more dramatically compare ASDK vs UIKit. 
- I also own a 6s+ and yes, it is difficult to perceive the difference on a high speed internet connection with a top of the line apple device. However, if you have a slower device (the 5th Gen iPod is the ideal test device according to @appleguy), like what much of the general population owns, you can much more easily see the difference. 
- Check out https://github.com/kconner/KMCGeigerCounter. With a one line change in the AppDelegate, you can see / hear the frame drops on each home feed. 

Another interesting point that you might want to research for your presentation is this: ASDK was built for performance, however, in doing research for the documentation for the 2.0 launch, the features that come with ASDK are compelling enough set to adopt the framework on their own! There are far too many to mention here, but the ASRangeController (intelligent preloading), automatic batch fetching, concurrent layout, high performance corner rounding, HTTP live video streaming, ridiculous tappable text support (across line breaks), extensible and reusable layout,… the list goes on and on. :)

appleguy: @vtsoup it's great to hear that you are investigating framework for your team. I think every team has different requirements, which depend on the design of the app they're creating, the engineers' interest in learning new things, and the size + current level of tech debt in the codebase.

[Sorry this is long and hard to read - I am getting over some RSI right now and using voice dictation, which reads a bit awkwardly...]

The iPhone 6S+ is _the fastest iPhone ever created_, and nearly the fastest iOS device (iPad Pro beats it by less than 30% in CPU!!!)  That said, many apps still have performance problems with it - but because of its sheer power, you need to be looking at a more complex app to see the quite substantial difference.

Pinterest is currently rolling out the first 100%-ASDK version of the app (if you look at it now, you are almost certainly seeing the old version).  The ASDK version is dramatically faster with such a complex app.  Scrolling smoothness is one thing - the old Pinterest does drop frames, sometimes noticably, on the 6S+ although it is not a big problem (the iPhone 6 is much more stuttery, the 6+ is worse, but the 5S and especially 5C and 5 are far worse yet the most common devices...4S is still very common).  Remember iPads are also very resource constrained compared to iPhones, and even the original iPad Air (very very common) is often very with complex apps.

Speaking performance alone - the biggest advantages of ASDK are non-perf -- for Pinterest and many apps, on a fast device like 6S+ the way the app manages loading new content & images is the most influential, esp. on slower connections.

ASDK is not just for accelerating layout and drawing. Although making both of these operations concurrent will also decrease the time to Display content to the user, sometimes meaningfully shortening the transition to a new you or shaving hundreds of milliseconds off of processing a new server response, it is really the interaction-driven network loading that allows the framework to take advantage of whatever quality of network connection it is given.

For example, out-of-the-box using ASNetworkImageNode with nothing more than an alloc/init call, you get these behaviors:

- Automatic preloading, memory caching, and disk caching that is _managed by sensing the user's intent_
- 100% tunable as to when image downloads begin and when their decode starts.
- Powerful placeholder functionality to improve the look and feel of an application even while it is loading data, as well as during decode or drawing operations. (Pinterest uses this for the colored rounded rectangles - nothing more than implementation of one method, -placeholderImage, and even the fading is automatic).
- Multi-directional scrolling support means that Device resources are primarily focused on preparing and storing data related to the direction of user travel. If the user is paging through full-screen views horizontally, the focus will be on loading pages in the direction of travel while maintaining reasonable memory usage of those behind. If the user begins scrolling vertically within a page, memory is freed from some of the surrounding pages, and devices immediately prioritizes loading data in that direction.
- Image downloads are automatically managed at _three_ priority levels, with no implementation work.  Normal preload (low pri), almost onscreen (within decoding range), mid pri, and visible (extreme priority if an image has not finished rendering).
- Native support for Progressive JPEG, which most .jpg files on the internet already support (all good server JPEG encoders create these because they are actually about 10% smaller than non-progressive for same quality).  This intelligently avoids creating progressive scans for images that are off screen, but if the image comes on screen win only part of the data is available, the data downloaded so far it is decoded to create an intermediate quality version and continuously improve as more data arrives. This makes a huge difference on slow connections and for large images.
- Built in GIF support.
- JSON API fetches can also be driven automatically.

This is really just one example of a performance feature of the framework that will affect even the latest devices, and will also remove complexity from your application code. However, it is important to remember that developers are often surrounded by strong Internet connection quality, even the cell networks of the areas we 10 to work. I'm not sure if this is also true in your area, but it is critical to remember that most people using your app have Devices that are at least five times, often 10 times slower in CPU power and a similar amount slower in network performance...

These differences may seem easy to extrapolate, but it is difficult to predict the full impact that a high latency network connection has on the user experience for an application with only basic mechanisms to preload and manage caching. Although it is difficult to test these things yourself, The network connection tool Apple provides can be helpful, and it is definitely recommended that you add performance instrumentation to your app to see real world performance that your users experience.

Despite the performance capabilities of the framework, in the latest versions the best advantage for a team considering adoption include the ability to use a wide library of components with behaviors that significantly add to the Apple APIs. I can't fully elaborate on this feature set right now, but there are many big and small examples.  A small one to check out is .hitTestSlop - trivially make the tap area of _any_ node type bigger in any direction, without overriding something like -pointInside and doing the frame math.

Hope to see you on Slack and answer any further questions.

examples/ASDKgram/Sample/PhotoTableViewCell.m
```diff
@@ -0,0 +1,248 @@
...
+    _photoCommentsView.backgroundColor                = [UIColor greenColor];
+    _photoDescriptionLabel.backgroundColor            = [UIColor greenColor];
+    _photoLikesLabel.backgroundColor                  = [UIColor greenColor];
+#endif
+  }
+  
+  return self;
+}
+
+- (void)layoutSubviews
```
appleguy: Because the comparison implementation uses layout specifications, which are at least as flexible as Autolayout, we should use Autolayout in the UIKit comparison.

This code is fairly clean and worth including however to extend the example by providing an alternative mechanism, For developers who prefer the manual layout style.  Maybe add a #define USE_UIKIT_AUTOLAYOUT 1, and you can compile out the method by default and then a developer testing the app can easily switch it if they are reading the code and interested

examples/ASDKgram/Sample/AppDelegate.h
```diff
@@ -0,0 +1,21 @@
...
+//  AppDelegate.h
+//  ASDKgram
+//
+//  Created by Hannah Troisi on 2/16/16.
+//  Copyright © 2016 Hannah Troisi. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+
+@protocol PhotoFeedViewControllerProtocol <NSObject>
```
appleguy: Please move this into the header file for the photo view controller. Then, you can import it only in the implementation file and do not need it in this header.

examples/ASDKgram/Sample/AppDelegate.h
```diff
@@ -0,0 +1,21 @@
...
+- (void)resetAllData;
+@end
+
+@interface AppDelegate : UIResponder <UIApplicationDelegate>
+
+@property (strong, nonatomic) UIWindow *window;
```
appleguy: I think we do not need to expose either of these properties, and they can be internal only. In fact, they probably don't need properties and can be just instance variables.

examples/ASDKgram/Sample/AppDelegate.m
```diff
@@ -0,0 +1,116 @@
...
+  
+  self.window.rootViewController = tabBarController;
+  [self.window makeKeyAndVisible];
+  
+  // opaque status bar background
+  self.statusBarOpaqueUnderlayView                 = [[UIView alloc] init];
+  self.statusBarOpaqueUnderlayView.backgroundColor = [UIColor darkBlueColor];
+  self.statusBarOpaqueUnderlayView.frame           = [[UIApplication sharedApplication] statusBarFrame];
+  if ( ![[UIApplication sharedApplication] isStatusBarHidden] ) {
+    [[[UIApplication sharedApplication] keyWindow] addSubview:self.statusBarOpaqueUnderlayView];
```
appleguy: Perhaps this could was moved from another method, but in this method we have direct access to the "window" variable so it should not be accessed through the application object.

examples/ASDKgram/Sample/AppDelegate.m
```diff
@@ -0,0 +1,116 @@
...
+  // UINavigationController does not forward on preferredStatusBarStyle calls to its child view controllers.
+  // Instead it manages its own state...http://stackoverflow.com/questions/19022210/preferredstatusbarstyle-isnt-called/19513714#19513714
+  uikitHomeFeedNavCtrl.navigationBar.barStyle = UIBarStyleBlack;
+  asdkHomeFeedNavCtrl.navigationBar.barStyle  = UIBarStyleBlack;
+  
+  self.window.rootViewController = tabBarController;
+  [self.window makeKeyAndVisible];
+  
+  // opaque status bar background
+  self.statusBarOpaqueUnderlayView                 = [[UIView alloc] init];
```
appleguy: You can get rid of the properties fro both window and the statusBar underlay and just use instance variables

examples/ASDKgram/Sample/AppDelegate.m
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+}
+
+- (void)handleDidChangeStatusBarOrientationNotification:(NSNotification *)notification;
+{
+  NSNumber *newOrientation = [notification.userInfo objectForKey: UIApplicationStatusBarOrientationUserInfoKey];
+  NSInteger newOrientationInt = [newOrientation integerValue];
+  if (UIInterfaceOrientationIsLandscape(newOrientationInt)) {
+    [self.statusBarOpaqueUnderlayView removeFromSuperview];
```
appleguy: Why is it necessary to remove this ever?  Status bar should show in landscape too.

examples/ASDKgram/Sample/AppDelegate.m
```diff
@@ -0,0 +1,116 @@
...
+  } else {
+    [[[UIApplication sharedApplication] keyWindow] addSubview:self.statusBarOpaqueUnderlayView];
+  }
+}
+
+- (void)applicationWillResignActive:(UIApplication *)application {
```
appleguy: Just delete all these default stubs

examples/ASDKgram/Sample/CommentFeedModel.h
```diff
@@ -0,0 +1,27 @@
+//
+//  CommentFeedModel.h
+//  ASDKgram
+//
+//  Created by Hannah Troisi on 3/9/16.
+//  Copyright © 2016 Hannah Troisi. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
```
appleguy: Minor, but maybe put this and all your UIKit.h includes into the precompiled header / .pch

examples/ASDKgram/Sample/Flickrgram.pch
```diff
@@ -0,0 +1,17 @@
...
+//  ASDKgram
+//
+//  Created by Hannah Troisi on 2/26/16.
+//  Copyright © 2016 Hannah Troisi. All rights reserved.
+//
+
+#ifndef Flickrgram_pch
+#define Flickrgram_pch
+
+#import <UIKit/UIKit.h>
```
appleguy: Here is where you could add Foundation and AsyncDisplayKit includes.  Also need to rename this file, although that may require adjusting the project build settinsg to be able to find the file.

examples/ASDKgram/Sample/AppDelegate.h
hannahmbanana: I should have named it more clearly. This is used by both the ViewController and NodeController to reset their data sources when switching between tabs.

examples/ASDKgram/Sample/AppDelegate.m
hannahmbanana: looks like iOS8 defaults to hiding the status bar in landscape orientation. I moved the OpaqueUnderlay to a UIWindow subclass for re-layout ease.

examples/ASDKgram/Sample/CommentFeedModel.m
```diff
@@ -0,0 +1,195 @@
...
+                }
+              }
+            }
+          }
+        }
+      }
+      dispatch_async(dispatch_get_main_queue(), ^{
+        _fetchPageInProgress = NO;
+        if (replaceData) {
+          _comments = [newComments mutableCopy];
```
appleguy: I think you are going to need some form of locking in this class, although it may not be a big issue because I think only one comments request per story is sent in practice...still this code is a lot higher quality than any current project in examples/ !

examples/ASDKgram/Sample/CommentsNode.m
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+
+#pragma mark - Helper Methods
+
+- (void)removeCommentLabels
+{
+  for (ASTextNode *commentLabel in _commentNodes) {
+    [commentLabel removeFromSupernode];
```
appleguy: You should refactor this code - using implicit hierarchy management will be much simpler.  No need for any of the add / remove as far as I know, since no reuse of these items as well.

examples/ASDKgram/Sample/PhotoCellNode.m
```diff
@@ -0,0 +1,188 @@
...
+    _photoTimeIntervalSincePostLabel                  = [[ASTextNode alloc] init];
+    _photoTimeIntervalSincePostLabel.layerBacked      = YES;
+    _photoTimeIntervalSincePostLabel.attributedString = [photo uploadDateAttributedStringWithFontSize:FONT_SIZE];
+    
+    _photoLikesLabel                        = [[ASTextNode alloc] init];
+    _photoLikesLabel.layerBacked            = YES;
+    _photoLikesLabel.attributedString       = [photo likesAttributedStringWithFontSize:FONT_SIZE];
+    
+    _photoDescriptionLabel                  = [[ASTextNode alloc] init];
+    _photoDescriptionLabel.layerBacked      = YES;
```
appleguy: You might have enough text nodes to warrant a -createTextNodeWithString: method, esp because they all set layerBacked.

examples/ASDKgram/Sample/PhotoCellNode.m
```diff
@@ -0,0 +1,188 @@
...
+  }
+  
+  // header stack
+  
+  _userAvatarImageView.preferredFrameSize        = CGSizeMake(USER_IMAGE_HEIGHT, USER_IMAGE_HEIGHT);     // constrain avatar image frame size
+  _photoTimeIntervalSincePostLabel.spacingBefore = HORIZONTAL_BUFFER;                 // to remove double spaces around spacer
+  
+  ASLayoutSpec *spacer      = [[ASLayoutSpec alloc] init];    // FIXME: long locations overflow post time - set max size?
+  spacer.flexGrow           = YES;
+  spacer.flexShrink         = YES;
```
appleguy: flexShrink is probably not needed on this one

examples/ASDKgram/Sample/PhotoCellNode.m
```diff
@@ -0,0 +1,188 @@
...
+  
+  headerStack.flexShrink = YES;
+
+  [headerStack setChildren:@[avatarInset, headerSubStack, spacer, _photoTimeIntervalSincePostLabel]];
+  
+  // header inset stack
+  
+  UIEdgeInsets insets                = UIEdgeInsetsMake(0, HORIZONTAL_BUFFER, 0, HORIZONTAL_BUFFER);
+  ASInsetLayoutSpec *headerWithInset = [ASInsetLayoutSpec insetLayoutSpecWithInsets:insets child:headerStack];
+  headerWithInset.flexShrink = YES;
```
appleguy: Should check if any of these need flexShrink

examples/ASDKgram/Sample/PhotoFeedModel.m
```diff
@@ -0,0 +1,238 @@
...
+  
+  NSString *userString = [NSString stringWithFormat:@"%lu", (long)userID];
+  _urlString = [fiveHundredPX_ENDPOINT_HOST stringByAppendingString:fiveHundredPX_ENDPOINT_USER];
+  _urlString = [[_urlString stringByAppendingString:userString] stringByAppendingString:@"&sort=created_at&image_size=3&include_store=store_download&include_states=voted"];
+  _urlString = [_urlString stringByAppendingString:fiveHundredPX_CONSUMER_KEY_PARAM];
+}
+
+- (void)clearFeed
+{
+  _photos = [[NSMutableArray alloc] init];
```
appleguy: We should take a closer look here in a followup - an @synchronized like you have below may be necessary here and possibly a couple other critical sections as well, like requestPageWithCompletionBlock:

examples/ASDKgram/Sample/UserModel.h
```diff
@@ -0,0 +1,40 @@
...
+//  UserModel.h
+//  ASDKgram
+//
+//  Created by Hannah Troisi on 2/26/16.
+//  Copyright © 2016 Hannah Troisi. All rights reserved.
+//
+
+@interface UserModel : NSObject
+
+@property (nonatomic, assign, readonly) NSDictionary *dictionaryRepresentation;
```
appleguy: ! easy one to miss, but this one must be strong and not assign.

examples/ASDKgram/Sample/WindowWithStatusBarUnderlay.m
```diff
@@ -0,0 +1,40 @@
...
+
+-(void)layoutSubviews
+{
+  [super layoutSubviews];
+  
+  [self bringSubviewToFront:_statusBarOpaqueUnderlayView];
+  
+  _statusBarOpaqueUnderlayView.frame = CGRectMake(0,
+                                                  0,
+                                                  [[UIScreen mainScreen] bounds].size.width,
```
appleguy: Probably worth a comment why this is necessary rather than using the status bar frame - I assume it is not correct during the moment of rotation or something?  Is this one always the right one?  Do you need to switch between width and height depending on orientation?

Might be simpler to read to start by setting rect = statusbarframe, then change rect.size.width =

examples/ASDKgram/Sample/main.m
```diff
@@ -0,0 +1,20 @@
+
+
+
```
appleguy: Remove extra returns

appleguy:
@hannahmbanana thank you for working on this!

Overall, it is by far the cleanest _and_ most full-featured app in examples/ .  Really.  It is also the only app in examples/ with room to grow, both architecturally and in concept.

It would be awesome to see continued work here.  Both the handful of cleanup areas mentioned (all more minor than the issues affecting every other app in examples/), but also additional features like perhaps viewing user profiles for 500px photographers with an ASCollectionView of their photos - and maybe use ASMapNode on a Location page, reusing the ASCollectionView of photos - perhaps wrapping up the example with opening / closing photos and using an ASPagerNode for fullscreen swiping?

If you use ASViewController to contain the user profile or other views, this example would be only one even close to engaging all four ASDisplayNode container types :)  cc @maicki @levi for thinking about developer intuitiveness / documentation, and how we may be able to leverage this code as an easy-to-read example for things like "go check out X for an example of using implicit hierarchy management to resize a cell with -setNeedsLayout".

examples/ASDKgram/Sample/CommentFeedModel.m
hannahmbanana: Yes, I believe I also have a race condition within the PhotoFeedModel.m that I cannot solve myself.  It is triggered when very quickly switching between tabs. I could use some help resolving this.

examples/ASDKgram/Sample/CommentsNode.m
hannahmbanana: I tried to do this by (1) `self.usesImplicitHierchy = YES` in `init:` (2) remove `- (void)removeCommentLabels` and (3) remove `[self addSubnode:commentLabel]` in `- (void)createCommentLabels`. This caused a weird bug where some of the commentViews (but not all) became stretched out:

img_3028

hannahmbanana:
I will definitely continue adding to ASDKgram! I'm glad it could be useful. If you have time in the future, it'd be helpful to know if there is any way we can quantify the performance differences between the UIKit and ASDK -implemented home feeds. E.g. you had mentioned adding the GeigerCounter at some point

FYI - the PhotoFeedNodeController is a ASViewController subclass - to parallel UIKit's UIViewController containing the UITableView.

examples/ASDKgram/Sample/PhotoTableViewCell.m
hannahmbanana: FYI - I had to learn about UIKit's Autolayout to do this (and don't really plan on using it again), so it should be noted that this portion of code is most likely not as well optimized as it could be.

PR #1462, opened 2016-03-30
[WIP] [ASTraitCollection] Add trait collection support to ASDisplayNode

levi:


appleguy:
@levi this is very exciting!  Now that the environment patch has merged, we should make the surgical adaptations required to integrate this functionality, but there is no doubt that most of the work here is reusable.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2049,6 +2052,67 @@ - (BOOL)supportsRangeManagedInterfaceState
...
+- (id<ASTraitCollection>)traitCollection
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if (_traitCollection == nil) {
+    _traitCollection = [[ASDisplayTraitCollection alloc] initWithTraits:_internalTraits];
+  }
+  return _traitCollection;
+}
+
+- (void)setTraitCollection:(id<ASTraitCollection>)traitCollection
```
rcancro: Is setting a trait collection something that we should do? `UITraitEnvironment` exposes `traitCollection` as readonly. 

Instead I wonder if the default behavior should be to determine the traits based upon the current state of the node. Can we try to iterate up the responder chain calling `traitCollection`? If a subclass overrides `traitCollection` to always be a particular set of traits, that would stop the responder iteration and propagate the custom trait collection back to this node. If we reach the top of the responder chain, could we make an `ASTraitCollection` based on that node's `UITraitCollection` (not sure how this will work if the top responder is not view-backed)? If not then we'd need all the nasty logic to determine size classes.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2049,6 +2052,67 @@ - (BOOL)supportsRangeManagedInterfaceState
   return ASHierarchyStateIncludesRangeManaged(_hierarchyState);
 }
 
+- (id<ASTraitCollection>)traitCollection
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if (_traitCollection == nil) {
+    _traitCollection = [[ASDisplayTraitCollection alloc] initWithTraits:_internalTraits];
```
rcancro: I can see wanting to pass around some type of context with the trait collection... in fact I see this type of requirement in the current Pinterest app ;). Is there some way we could have a custom trait collection class so that would support this. 

For example, instead of calling `ASDisplayTraitCollection alloc`, we have some idea of a custom trait collection class so the init would look like `[[[ASDisplayNode traitCollectionClass] alloc] initWithTraits:_internalTraits];` That would allow for a subclass of `ASDisplayTraitCollection` that has a context, though I'm not entirely sure where that context would get set. I don't really want to make an `ASDisplayNode` subclass just for that...

levi: Yes we definitely need it to be custom. I believe this was a stub with the intention of future extension. The problem to consider is how we allow subclassing but also returning a custom struct that can be used for ASEnvironment down propagation.

AsyncDisplayKit/ASDisplayNode.mm
levi: We just need some sort of entry point that UIView hosts can pass into the node hierarchy. The responder chain is difficult to add into the picture here as it requires the node to rely upon the state of the view hierarchy. Ideally, we should design this such that the node is unaware of the view hierarchy and acts pretty self-contained once the trait collection has been received by its issuer.

levi:
@rcancro has a more complete implementation up.

PR #1463, opened 2016-03-31
[AsyncDisplayKit] - Fix Travis issues

marmelroy:
Travis needs to switch to 7.3 in order to build with iOS 9.3


marmelroy:
Okay, tests mode is still failing because of https://github.com/facebook/xctool/issues/666. It's fixed on HEAD but they need to push a new release. Pushing a workaround...

PR #1469, opened 2016-04-02
[ASEnvironment - Layout] Fixes to upward propagation of ASLayoutable properties.

hannahmbanana:
Previously I wasn't able to get the case shown in Social App Layout to work...wrapping an ASLayoutSpec with .flexGrow = YES in an Inset spec would cause the flexGrow to be ignored.  It now appears to work correctly, and I followed in the debugger to check the value was copied.

This also includes a couple small style fixes to SocialAppLayout.  Later I will do a PR to re-indent the files to match the project standard of two spaces.


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1702,10 +1702,6 @@ - (void)__setSupernode:(ASDisplayNode *)newSupernode
       [self exitHierarchyState:stateToEnterOrExit];
     }
   }
-    
-  if ([newSupernode supportsUpwardPropagation]) {
```
appleguy: I think removing this makes sense, because it is not intended for sub nodes to pass layout options to their parents. Indeed, it is common to have multiple subnodes, so the value of supports upward propagation is misleading in this case.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -81,8 +81,8 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
 
     id<ASLayoutable> finalLayoutable = [child finalLayoutable];
     if (finalLayoutable != child) {
-      // Layout options state of child needs to be copied to final layoutable
-      finalLayoutable.environmentState.layoutOptionsState = child.environmentState.layoutOptionsState;
+      // Layout options state of child needs to be copied to final layoutable, but don't override customized values.
+      ASEnvironmentStatePropagateUp(finalLayoutable, child.environmentState.layoutOptionsState);
```
appleguy: This seems like a positive change. I could imagine a final layoutable object with a custom value for a layout options property that is essential not to overwrite, but it does seem like a reasonable default behavior to inherit any properties that aren't overridden.

This does raise a question, which was also present in the prior behavior, on how a final layout of all might be able to override a property to require that it stay the default (e.g. intentionally set flexGrow = NO).  It is quite rare to use finalLayoutable however, so this seems fine for now.

AsyncDisplayKit/Private/ASEnvironmentInternal.mm
```diff
@@ -115,33 +115,33 @@ ASEnvironmentState ASEnvironmentMergeObjectAndState(ASEnvironmentState environme
...
       parentState.spacingAfter = state.spacingAfter;
     }
-    if (parentState.alignSelf != defaultState.alignSelf) {
-      parentState.alignSelf = defaultState.alignSelf;
+    if (parentState.alignSelf == defaultState.alignSelf) {
+      parentState.alignSelf = state.alignSelf;
     }
-    if (parentState.flexGrow != defaultState.flexGrow) {
-      parentState.flexGrow = defaultState.flexGrow;
+    if (parentState.flexGrow == defaultState.flexGrow) {
```
appleguy: @maicki just wanted to check with you - on all the changes in this PR, but these do seem like important ones / look correct to me.

maicki: @appleguy looks good to me and @hannahmbanana thanks for improving this!

PR #1472, opened 2016-04-04
Rename internal headers to conform with + naming convention

levi:
Really like this +Beta.h style. Let's bring it to everything else.


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -12,9 +12,9 @@
 #import <AsyncDisplayKit/_ASDisplayLayer.h>
 #import <AsyncDisplayKit/ASAssert.h>
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
-#import <AsyncDisplayKit/ASDisplayNodeInternal.h>
+#import <AsyncDisplayKit/ASDisplayNode+Internal.h>
 #import <AsyncDisplayKit/ASHighlightOverlayLayer.h>
-#import <AsyncDisplayKit/ASDisplayNodeExtras.h>
+#import <AsyncDisplayKit/ASDisplayNode+Extras.h>
```
appleguy: Some of these headers are currently public, such as this one. I think that means that we need to reserve this change for 2.0.  Unfortunately, Swift and Carthage support at some point require that we make various headers public that were not originally intended to be in order to allow their unusual build systems to succeed.  I only realized this recently, and I am not concerned about changing these names in the near future, but it may not be the best time to do so as framework changes are very rapid right now and it is fairly important that developers can easily update to the latest Core behavior fixes without too many build breaking changes.

appleguy:
Since a conflict has arisen, closing for now and will reopen before 2.0.

PR #1473, opened 2016-04-04
[WIP] Add support for manually inserted nodes in an implicitly managed hierarchy

levi:


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1200,6 +1202,11 @@ - (void)addSubnode:(ASDisplayNode *)subnode
 
   [_subnodes addObject:subnode];
   
+  // Move the layout range when the layout hasn't been generated yet
```
appleguy: I don't quite understand this code. Can you expand the comment a bit?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1501,6 +1515,24 @@ - (void)removeFromSupernode
   }
 }
 
+- (void)_insertLayoutRangeOffsetAtIndex:(NSUInteger)index
```
appleguy: Soon, we should create a file to contain this kind of layout system - it's certainly not the main reason ASDisplayNode is so big, but we might consider a category dedicated to add / remove subnode / measure / layout

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -251,8 +252,9 @@ - (void)_initializeInstance
   [self _staticInitialize];
   _contentsScaleForDisplay = ASScreenScale();
   _displaySentinel = [[ASSentinel alloc] init];
+  _layoutRange = [[ASTranslationRange alloc] initWithLocation:0 length:0];
```
appleguy: Let's find a way to defer the allocation of this. After seeing that ASSentinel is one of the most commonly created and destroyed objects, along with the savings from removing common objects like ASLayoutOptions, it would be ideal to ensure we don't create these for situations that don't need them such as all leaf nodes at the bottom of a hierarchy.

AsyncDisplayKit/Private/ASTranslationRange.mm
```diff
@@ -0,0 +1,158 @@
...
+
+#import "ASAssert.h"
+#import "ASThread.h"
+#import <vector>
+#import <algorithm>
+
+@interface ASTranslationRange () {
+  NSUInteger _length;
+  std::vector<NSRange> _ranges;
+  ASDN::RecursiveMutex _propertyLock;
```
appleguy: It could be a good idea to make sure this object is internally locked, but if all of the usages of it are already locked by the owning object, then this is not needed.

AsyncDisplayKitTests/ASDisplayNodeImplicitHierarchyTests.m
```diff
@@ -122,14 +122,85 @@ - (void)testCalculatedLayoutHierarchyTransitions
...
+  ASDisplayNode *nodeC = [[ASDisplayNode alloc] init];
+  
+  ASDisplayNode *node1 = [[ASDisplayNode alloc] init];
+  ASDisplayNode *node2 = [[ASDisplayNode alloc] init];
+  ASDisplayNode *node3 = [[ASDisplayNode alloc] init];
+  
+  ASDisplayNode *nodeZ = [[ASDisplayNode alloc] init];
+  
+  ASSpecTestDisplayNode *node = [[ASSpecTestDisplayNode alloc] init];
+  node.layoutSpecBlock = ^(ASSizeRange constrainedSize, NSNumber *layoutState) {
```
appleguy: !! layoutSpecBlock !!

Should we make this public API?

AsyncDisplayKit/Private/ASTranslationRange.h
```diff
@@ -0,0 +1,47 @@
...
+//
+//  Created by Levi McCallum on 4/4/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ * A translation range simulates a contiguous range, but allows offsets to be inserted and removed to translate an index
+ * within the contiguous range to the actual index that
```
appleguy: Description cut short.

Love this class.  I don't yet fully understand a simple way to explain what it does, and we might want to consider different names, but just upgrading the description could be enough.

levi:
Thanks for the review, Scott. I've spent enough time away from this patch to realize that I can solve the problem with a better data structure. Going to revise this with a new approach.

levi:
Closing this because there is a better way of doing this

PR #1474, opened 2016-04-04
[ASMapNode] Never Provide Invalid Snapshot Dimensions to MapKit

Adlai-Holler:
Sometimes during the first layout passes for complex hierarchies, my map node will get measured with 0-height or 0-width, and MapKit will throw an exception if you set MKMapSnapshotOptions.size with 0 area. This avoids that exception.

### DIFF
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -279,7 +279,7 @@
 #pragma mark - Layout
 - (void)setSnapshotSizeWithReloadIfNeeded:(CGSize)snapshotSize
 {
-  if (!CGSizeEqualToSize(self.options.size, snapshotSize)) {
+  if (snapshotSize.height > 0 && snapshotSize.width > 0 && !CGSizeEqualToSize(self.options.size, snapshotSize)) {
     _options.size = snapshotSize;
     if (_snapshotter) {
       [self destroySnapshotter];


AsyncDisplayKit/ASMapNode.mm
```diff
@@ -279,7 +279,7 @@ - (void)setAnnotations:(NSArray *)annotations
 #pragma mark - Layout
 - (void)setSnapshotSizeWithReloadIfNeeded:(CGSize)snapshotSize
 {
-  if (!CGSizeEqualToSize(self.options.size, snapshotSize)) {
+  if (snapshotSize.height > 0 && snapshotSize.width > 0 && !CGSizeEqualToSize(self.options.size, snapshotSize)) {
```
appleguy: Is there any possibility that a new snapshot could be triggered using the old size, after the point at which we attempted to set the snapshot size to zero but then avoided entering this condition?

It seems like we may need some kind of flag that invalidates the current size. Maybe a safer fallback is to unconditionally set the size of the options, but guard against calling into the snapshoter with that options if the size is equal to the zero size.

Adlai-Holler: The exception is thrown immediately when you set it on the options object though. =/

appleguy:
@aaronschubert0 not urgent, but if you circle back to this code at some point, it might be worth adding a comment about the fact that ideally we could clear out the options if the size gets set to zero. One good way to implement this might be to set aside the options object given a new instance variable, and if the size is later setback to a nonzero value, reuse the old options object to maintain the other settings it has.

@Adlai-Holler thank you for this fix, I know your schedule is extremely tight right now and so I really value your contributions.

aaronschubert0:
@appleguy That sounds like a good idea, that way we don't need to destroy & re-create the options object. I can circle back to this code soon.

PR #1476, opened 2016-04-05
[ASDisplayNode] Automatically integrate with Accessibility even for layer-backed or rasterized subtrees.

maicki:
Improves Accessibility support especially for ASDisplayNodes with that has layerBacked or shouldRasterizeDescendants enabled should still supported by the accessibility mechanism. To enable this the _ASDisplayView implements the UIAccessibilityContainer methods now and is responsible for creating UIAccessibilityElement for subnodes.

Open Questions:
- [x] Figure out UIAccessibilityContainers in UIAccessibilityContainers

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -711,7 +711,7 @@ NS_ASSUME_NONNULL_END
 @property (atomic, assign)           BOOL accessibilityViewIsModal;
 @property (atomic, assign)           BOOL shouldGroupAccessibilityChildren;
 
-@property (nonatomic) UIAccessibilityNavigationStyle accessibilityNavigationStyle NS_AVAILABLE_IOS(8_0);
+@property (nonatomic) UIAccessibilityNavigationStyle accessibilityNavigationStyle;
 #if TARGET_OS_TV
 @property(nullable, nonatomic, copy) NSArray *accessibilityHeaderElements;
 #endif
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -700,26 +700,25 @@ NS_ASSUME_NONNULL_END
 @interface ASDisplayNode (UIViewBridgeAccessibility)
 
 // Accessibility support
-@property (atomic, assign)           BOOL isAccessibilityElement;
-@property (nullable, atomic, copy)   NSString *accessibilityLabel;
-@property (nullable, atomic, copy)   NSString *accessibilityHint;
-@property (nullable, atomic, copy)   NSString *accessibilityValue;
-@property (atomic, assign)           UIAccessibilityTraits accessibilityTraits;
-@property (atomic, assign)           CGRect accessibilityFrame;
-@property (nullable, atomic, strong) NSString *accessibilityLanguage;
-@property (atomic, assign)           BOOL accessibilityElementsHidden;
-@property (atomic, assign)           BOOL accessibilityViewIsModal;
-@property (atomic, assign)           BOOL shouldGroupAccessibilityChildren;
-
-@property (nonatomic) UIAccessibilityNavigationStyle accessibilityNavigationStyle;
+@property (nonatomic, assign)           BOOL isAccessibilityElement;
+@property (nonatomic, copy, nullable)   NSString *accessibilityLabel;
+@property (nonatomic, copy, nullable)   NSString *accessibilityHint;
+@property (nonatomic, copy, nullable)   NSString *accessibilityValue;
+@property (nonatomic, assign)           UIAccessibilityTraits accessibilityTraits;
+@property (nonatomic, assign)           CGRect accessibilityFrame;
+@property (nonatomic, copy, nullable)   UIBezierPath *accessibilityPath;
+@property (nonatomic, assign)           CGPoint accessibilityActivationPoint;
+@property (nonatomic, copy, nullable)   NSString *accessibilityLanguage;
+@property (nonatomic, assign)           BOOL accessibilityElementsHidden;
+@property (nonatomic, assign)           BOOL accessibilityViewIsModal;
+@property (nonatomic, assign)           BOOL shouldGroupAccessibilityChildren;
+@property (nonatomic, assign)           UIAccessibilityNavigationStyle accessibilityNavigationStyle;
 #if TARGET_OS_TV
-@property(nullable, nonatomic, copy) NSArray *accessibilityHeaderElements;
+@property(nonatomic, copy, nullable) 	NSArray *accessibilityHeaderElements;
 #endif
-@property (nonatomic) CGPoint accessibilityActivationPoint;
-@property (nullable, nonatomic, copy) UIBezierPath *accessibilityPath;
 
 // Accessibility identification support
-@property (nullable, nonatomic, copy) NSString *accessibilityIdentifier;
+@property (nonatomic, copy, nullable)   NSString *accessibilityIdentifier;
 
 @end
 
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -695,6 +695,10 @@ NS_ASSUME_NONNULL_END
 - (nullable UIView *)preferredFocusedView;
 #endif
 
+@end
+
+@interface ASDisplayNode (UIViewBridgeAccessibility)
+
 // Accessibility support
 @property (atomic, assign)           BOOL isAccessibilityElement;
 @property (nullable, atomic, copy)   NSString *accessibilityLabel;
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -711,6 +711,13 @@ NS_ASSUME_NONNULL_END
 @property (atomic, assign)           BOOL accessibilityViewIsModal;
 @property (atomic, assign)           BOOL shouldGroupAccessibilityChildren;
 
+@property (nonatomic) UIAccessibilityNavigationStyle accessibilityNavigationStyle NS_AVAILABLE_IOS(8_0);
+#if TARGET_OS_TV
+@property(nullable, nonatomic, copy) NSArray *accessibilityHeaderElements;
+#endif
+@property (nonatomic) CGPoint accessibilityActivationPoint;
+@property (nullable, nonatomic, copy) UIBezierPath *accessibilityPath;
+
 // Accessibility identification support
 @property (nullable, nonatomic, copy) NSString *accessibilityIdentifier;
 
--- a/AsyncDisplayKit/Details/_ASDisplayView.mm
+++ b/AsyncDisplayKit/Details/_ASDisplayView.mm
@@ -382,22 +382,6 @@
 
 #pragma mark - Accessibility
 
-static BOOL ASNodeValidForAccessibility(ASDisplayNode *node)
-{
-  if (node.isAccessibilityElement) {
-    return YES;
-  }
-  
-  if (node.isLayerBacked) {
-    // Assume for now that all nodes that have an accessibility label or
-    return node.accessibilityLabel.length > 0 ||
-           node.accessibilityValue.length > 0 ||
-           ((node.accessibilityTraits & UIAccessibilityTraitNone) != UIAccessibilityTraitNone);
-  }
-  
-  return NO;
-}
-
 static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
 
 @implementation UIAccessibilityElement (_ASDisplayView)
@@ -449,7 +433,7 @@ static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
       queue.pop();
       
       // Check if we have to add the node to the accessiblity nodes as it's an accessiblity element
-      if (node != selfNode && ASNodeValidForAccessibility(node)) {
+      if (node != selfNode && node.isAccessibilityElement) {
         UIAccessibilityElement *accessibilityElement = [[UIAccessibilityElement alloc] initWithAccessibilityContainer:self];
         accessibilityElement.asyncdisplaykit_node = node;
         [_accessibleElements addObject:accessibilityElement];
@@ -469,7 +453,7 @@ static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
     if (!subnode.isAccessibilityElement && [subnode accessibilityElementCount] > 0) {
       // We are good and the view is an UIAccessibilityContainer so add that
       [_accessibleElements addObject:subnode.view];
-    } else if (ASNodeValidForAccessibility(subnode)) {
+    } else if (subnode.isAccessibilityElement) {
       // Create a accessiblity element from the subnode
       UIAccessibilityElement *accessibilityElement = [[UIAccessibilityElement alloc] initWithAccessibilityContainer:self];
       accessibilityElement.asyncdisplaykit_node = subnode;
--- a/AsyncDisplayKit/Details/_ASDisplayView.mm
+++ b/AsyncDisplayKit/Details/_ASDisplayView.mm
@@ -18,6 +18,8 @@
 #import "ASDisplayNode+FrameworkPrivate.h"
 #import "ASDisplayNode+Subclasses.h"
 
+#import <queue>
+
 @interface _ASDisplayView ()
 @property (nonatomic, assign, readwrite) ASDisplayNode *asyncdisplaykit_node;
 
@@ -380,6 +382,22 @@
 
 #pragma mark - Accessibility
 
+static BOOL ASNodeValidForAccessibility(ASDisplayNode *node)
+{
+  if (node.isAccessibilityElement) {
+    return YES;
+  }
+  
+  if (node.isLayerBacked) {
+    // Assume for now that all nodes that have an accessibility label or
+    return node.accessibilityLabel.length > 0 ||
+           node.accessibilityValue.length > 0 ||
+           ((node.accessibilityTraits & UIAccessibilityTraitNone) != UIAccessibilityTraitNone);
+  }
+  
+  return YES;
+}
+
 static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
 
 @implementation UIAccessibilityElement (_ASDisplayView)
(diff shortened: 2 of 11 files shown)


AsyncDisplayKit/Details/_ASDisplayView.mm
```diff
@@ -375,3 +376,81 @@ - (UIView *)preferredFocusedView
...
+
+@end
+
+@implementation _ASDisplayView (UIAccessibilityContainer)
+
+#pragma mark - UIAccessibility
+
+- (NSArray *)accessibleElements
+{
+  ASDisplayNode *asyncDisplayKitNode = self.asyncdisplaykit_node;
```
maicki: Move down after the if check

AsyncDisplayKit/Details/_ASDisplayView.mm
```diff
@@ -375,3 +376,81 @@ - (UIView *)preferredFocusedView
...
+  if ( _accessibleElements != nil ) {
+    return _accessibleElements;
```
appleguy: How does this get invalidated when the node hierarchy changes?

I'd pretty strongly prefer we do no caching for now, as a) I doubt this method is called frequently, and b) it is very important for both testing and accessibility that we don't have any bugs with this information not getting invalidated.  Particularly since it is a feature that is not tested as frequently, we can layer in optimizations later if needed.

AsyncDisplayKit/Details/_ASDisplayView.mm
```diff
@@ -375,3 +376,81 @@ - (UIView *)preferredFocusedView
...
+      [_accessibleElements addObject:accessibilityElement];
+    }
+  }
+  
+  return _accessibleElements;
+}
+
+- (NSInteger)accessibilityElementCount
+{
+  return [self accessibleElements].count;
```
appleguy: Even though it would be quite suboptimal for the first implementation to regenerate the array on each call to this, I suspect that it will not cause significant performance problems especially because Voice over itself adds fairly significant complexity and does not allow Long scrolling animations where frame drops are essential. There is also essentially no notion of touch latency with voiceovers activation mechanism.

In a future upgrade, it seems to me that the correct way to optimize this is to use the the environment system, doing some form of upward propagation of the count of accessibility elements at each level.  Changing a property such as is accessibility element may trigger a new propagation to update the state of the tree, interestingly this may be a use case where both upward and downward propagation is required in certain situations.

I don't think we need to schedule this optimization work to occur in the next few months however, and it would be interesting to do some profiling on the voiceover mode just to check where the bottlenecks are anyway, as there could be other parts of the framework about are even more useful to optimize for this mode

AsyncDisplayKit/Details/_ASDisplayView.mm
```diff
@@ -375,3 +376,81 @@ - (UIView *)preferredFocusedView
...
+}
+
+- (id)accessibilityElementAtIndex:(NSInteger)index
+{
+  UIAccessibilityElement *element = [self accessibleElements][index];
```
appleguy: If we do cache this, I wonder if there is a way to keep the array for only the amount of time it takes to complete a certain batch of calls to these APIs?

Even with this call point though, because the correct way to invalidate the cash would be to use the environment, it probably makes sense to have no optimization for now and then go directly to the environment based implementation if we can measure the overhead in a way that affects the user experience

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -841,6 +841,12 @@ - (void)setAccessibilityIdentifier:(NSString *)accessibilityIdentifier
   _setToViewOnly(accessibilityIdentifier, accessibilityIdentifier);
 }
 
+- (NSInteger)accessibilityElementCount
```
appleguy: Does this need to be defined in header as a bridge property? Otherwise the compiler may not allow it to be called directly on a node instance

maicki:
@appleguy @rahul-malik Can you take another look over the pull request please

AsyncDisplayKit/Private/ASDisplayNodeInternal.h
```diff
@@ -129,6 +129,19 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
...
+  NSString *_accessibilityLabel;
+  NSString *_accessibilityHint;
+  NSString *_accessibilityValue;
+  UIAccessibilityTraits _accessibilityTraits;
+  CGRect _accessibilityFrame;
+  NSString *_accessibilityLanguage;
+  BOOL _accessibilityElementsHidden;
+  BOOL _accessibilityViewIsModal;
+  BOOL _shouldGroupAccessibilityChildren;
+  NSString *_accessibilityIdentifier;
```
rahul-malik: I think there are more a11y properties to support that have been added in iOS8/9 and TV_OS.

/*
 Some assistive technologies allow the user to select a parent view or container to navigate its elements.
 This property allows an app to specify whether that behavior should apply to the receiver.
 Currently, this property only affects Switch Control, not VoiceOver or other assistive technologies.
 See UIAccessibilityConstants.h for the list of supported values.
 default == UIAccessibilityNavigationStyleAutomatic
 */
@property (nonatomic) UIAccessibilityNavigationStyle accessibilityNavigationStyle NS_AVAILABLE_IOS(8_0);

/*
 The elements considered to be the headers for this element. May be set on an instance of
 UIAccessibilityElement, a View or a View Controller. The accessibility container chain,
 and associated view controllers where appropriate, will be consulted.
 To avoid retain cycles, a weak copy of the elements will be held.
 */
@property(nullable, nonatomic, copy) NSArray *accessibilityHeaderElements UIKIT_AVAILABLE_TVOS_ONLY(9_0);

/*
 Returns the activation point for an accessible element in screen coordinates.
 default == Mid-point of the accessibilityFrame.
 */
@property (nonatomic) CGPoint accessibilityActivationPoint NS_AVAILABLE_IOS(5_0);
/*
 Returns the path of the element in screen coordinates.
 default == nil
 Setting the property, or overriding the method, will cause the assistive technology to prefer the path over the accessibility.
 frame when highlighting the element.
 */
@property (nullable, nonatomic, copy) UIBezierPath *accessibilityPath NS_AVAILABLE_IOS(7_0);

maicki: @rahul-malik They should be supported now

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -707,6 +711,13 @@ NS_ASSUME_NONNULL_END
 @property (atomic, assign)           BOOL accessibilityViewIsModal;
 @property (atomic, assign)           BOOL shouldGroupAccessibilityChildren;
 
+@property (nonatomic) UIAccessibilityNavigationStyle accessibilityNavigationStyle NS_AVAILABLE_IOS(8_0);
+#if TARGET_OS_TV
+@property(nullable, nonatomic, copy) NSArray *accessibilityHeaderElements;
```
rahul-malik: Should we control if this property exists by the #if TARGET_OS_TV or by adding UIKIT_AVAILABLE_TVOS_ONLY(9_0); to the declaration?

AsyncDisplayKit/Details/_ASDisplayView.mm
```diff
@@ -375,3 +378,126 @@ - (UIView *)preferredFocusedView
...
+- (void)setAsyncdisplaykit_node:(ASDisplayNode *)node
+{
+  objc_setAssociatedObject(self, ASDisplayNodeAssociatedNodeKey, node, OBJC_ASSOCIATION_ASSIGN); // Weak reference to avoid cycle, since the node retains the layer.
+  
+  // Update UIAccessibilityElement properties from node
+  self.accessibilityIdentifier = node.accessibilityIdentifier;
+  self.accessibilityLabel = node.accessibilityLabel;
+  self.accessibilityHint = node.accessibilityHint;
+  self.accessibilityValue = node.accessibilityValue;
+  self.accessibilityTraits = node.accessibilityTraits;
```
rahul-malik: Need to update this for the new properties we set?

rahul-malik:
@maicki - Looks great and thanks for adding tests. This seems ready to land now

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -707,6 +711,13 @@ NS_ASSUME_NONNULL_END
 @property (atomic, assign)           BOOL accessibilityViewIsModal;
 @property (atomic, assign)           BOOL shouldGroupAccessibilityChildren;
 
+@property (nonatomic) UIAccessibilityNavigationStyle accessibilityNavigationStyle;
```
appleguy: Properties above should probably be changed to be nonatomic.  One of the strings is strong instead of copy, too.  For alignment, maybe we should put , nullable at the end.

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -707,6 +711,13 @@ NS_ASSUME_NONNULL_END
...
+#if TARGET_OS_TV
+@property(nullable, nonatomic, copy) NSArray *accessibilityHeaderElements;
+#endif
+@property (nonatomic) CGPoint accessibilityActivationPoint;
```
appleguy: Great to have these.  Include , assign for the CGPoint to match CGRect above.  Put the TVOS stuff at the end and group the other ones together.  Let's either list them in the same order Apple does, or group them by variable type.  Feel free to modify the original properties.

AsyncDisplayKit/Details/_ASDisplayView.mm
```diff
@@ -375,3 +378,126 @@ - (UIView *)preferredFocusedView
...
+
+#pragma mark - Accessibility
+
+static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
+
+@implementation UIAccessibilityElement (_ASDisplayView)
+
+- (void)setAsyncdisplaykit_node:(ASDisplayNode *)node
+{
+  objc_setAssociatedObject(self, ASDisplayNodeAssociatedNodeKey, node, OBJC_ASSOCIATION_ASSIGN); // Weak reference to avoid cycle, since the node retains the layer.
```
appleguy: I'm worried about the possibility of bad access references / stale pointers...do we really need this?  In what case are we not dealing with a UIView?  Does this code ever get run, since UIAccessibilityElement is an optional object type for clients to use in their apps?

maicki: We do not deal with UIView's if the node is layer backed or for the subnodes if shouldRasterizeSubnodes is true and to be able to convert the frame from a UIAccessibilityElement to screen coordinates in accessibilityElementAtIndex:. Please see also comments in accessibilityElementAtIndex:.

AsyncDisplayKit/Details/_ASDisplayView.mm
```diff
@@ -375,3 +378,126 @@ - (UIView *)preferredFocusedView
...
+  // Handle not rasterize case
+  // Create UI accessiblity elements for each subnode that represent an elment within the accessibility container
+  for (ASDisplayNode *subnode in selfNode.subnodes) {
+      // Check if this subnode is a UIAccessibilityContainer
+    if (!subnode.isAccessibilityElement && [subnode accessibilityElementCount] > 0) {
+      // We are good and the view is an UIAccessibilityContainer so add that
+      [_accessibleElements addObject:subnode.view];
+    } else if (subnode.isAccessibilityElement) {
+      // Create a accessiblity element from the subnode
+      UIAccessibilityElement *accessibilityElement = [[UIAccessibilityElement alloc] initWithAccessibilityContainer:self];
```
appleguy: I'm not sure I understand the need for this - can't the ASDisplayNode objects implement the UIAccessibilityElement protocol, as they already have most of the properties?  This would also avoid object creation, but more importantly, the additional code to support associated objects etc.

maicki:
@rahul-malik @appleguy Ready for another review

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,153 @@
...
+    
+    // Queue used to keep track of subnodes while traversing this layout in a BFS fashion.
+    std::queue<ASDisplayNode *> queue;
+    queue.push(selfNode);
+    
+    while (!queue.empty()) {
+      ASDisplayNode *node = queue.front();
+      queue.pop();
+      
+      // Check if we have to add the node to the accessiblity nodes as it's an accessiblity element
```
appleguy: This is well done. Perhaps we should provide a convenient method that allows running a block on each node in BFS search fashion?

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,153 @@
...
+
+#import <objc/runtime.h>
+#import <queue>
+
+
+#pragma mark - UIAccessibilityElement
+
+static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
+
+@implementation UIAccessibilityElement (_ASDisplayView)
```
appleguy: See original review comments - I don't think we need this category or associated object implementation, but let me know if I'm wrong - or if there is a good benefit to this way

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,153 @@
...
+}
+
+- (NSInteger)accessibilityElementCount
+{
+  return [self accessibleElements].count;
+}
+
+- (id)accessibilityElementAtIndex:(NSInteger)index
+{
+  if (_accessibleElements == nil) {
```
appleguy: In this case, maybe we should actually trigger the creation of accessibleElements?  I am not sure what to expect of the access pattern for this method, but especially because I think it is no longer used in iOS 9 / thus we will get very little testing, it might be safer to trigger creating.  I honestly don't know though / am not familiar with when the system must call this method and what it would do if nil.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
appleguy: I see one reason - since we are keeping the array caching, it would need to be a weak reference.  

I remember being frustrated this class wasn't available in iOS when I needed it while working at Apple :)   -- but now we can use it as of iOS 6.0!

https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSPointerArray_Class/

AsyncDisplayKit/ASDisplayNode.h
maicki: This changed was made in a previous commit

AsyncDisplayKit/ASDisplayNode.h
maicki: This changed was made in a previous commit

AsyncDisplayKit/Details/_ASDisplayView.mm
maicki: We don't need to have it invalidated anymore as as we talked about this, we don't cache the accessibleElements so aggressive anymore.

AsyncDisplayKit/Details/_ASDisplayView.mm
maicki: We regenerate the _accessibleElements under certain circumstances now. Agree that the upward propagation of accessibility elements would be a great exercise for the ASEnvironment system as well as improving the timing of the accessibleElements caching.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: After logging `accessibilityElementCount` and `accessibilityElementAtIndex:` calls, `accessibilityElementCount` seems to be called always before `accessibilityElementAtIndex`, so the `_accessibleElements` should be available in this situation. I could definitely poke a bit deeper in UIKit to look up what's the definite order.

appleguy: In that case, this should probably be an assertion that the scenario should never happen?

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: For every subnode of a node with shouldRasterizeDescendants enabled or layer backed nodes we have to create a UIAccessibilityElement as for both of these nodes there is no view available. Furthermore we cannot just use an ASDisplayNode as an "accessibleElement" as UIAccessibilityContainer only can contain either UIView's or UIAccessibilityElements as "accessibleElements"

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: Agree fully, already added that just not pushed yet.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
appleguy: Is there any documentation about that?  I was skimming it about a week ago - might have missed something - and my understanding was that the elements are of type id <UIAccessibilityElement>

maicki: If you look into UIAccessiblity.h under UIAccessibilityContainer:

>  Sub-elements of a container that are not represented by concrete UIView
 instances (perhaps painted text or icons) can be represented using instances
 of UIAccessibilityElement class (see UIAccessibilityElement.h).

Furthermore take a look into UIAccessibilityElement.h:
>  Instances of this class can be used as "fake" accessibility elements.
 An accessibility container (see UIAccessibility.h) can create and vend instances
 of UIAccessibilityElement to cover for user interface items that are not
 backed by a UIView (for example: painted text or icon).

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: Added a new function that iterates a ASDisplayNode subtree in BFS search fashion.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
appleguy: @maicki looks like you're correct. I had seen those comments, but "can be represented" didn't clarify for me that there is no protocol that can be used. However, there does not seem to be a way.

It looks like this is an artifact of the API being design around 3.0, which predates buy a few years the time that Apple started with widely using protocols for many important features such as dynamic animation items.  Based on the functionality of these APIs, it is actually quite surprising they're not supported as a protocol!

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,122 @@
...
+#import "ASDisplayNode+FrameworkPrivate.h"
+
+#import <objc/runtime.h>
+
+
+#pragma mark - UIAccessibilityElement
+
+@implementation UIAccessibilityElement (_ASDisplayView)
+
++ (UIAccessibilityElement *)accessibilityElementWithContainer:(id)container node:(ASDisplayNode *)node
```
appleguy: This is probably not required, but maybe it would add additional type safety to use `id < UIAccessibilityContainer>`

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,122 @@
+/* Copyright (c) 2014-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "_ASDisplayViewAccessiblity.h"
```
appleguy: Though a few areas break this, let's follow the regular pattern of _ASDisplayView+Accessibility.h etc

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,122 @@
...
+#import "_ASDisplayView.h"
+#import "ASDisplayNodeExtras.h"
+#import "ASDisplayNode+FrameworkPrivate.h"
+
+#import <objc/runtime.h>
```
appleguy: Can remove this and some of the extra spaces

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,122 @@
...
+#pragma mark - UIAccessibility
+
+- (NSArray *)accessibleElements
+{
+  ASDisplayNode *selfNode = self.asyncdisplaykit_node;
+  if (selfNode == nil) {
+    return nil;
+  }
+  
+  _accessibleElements = [[NSMutableArray alloc] init];
```
appleguy: Let's make the ivar NSArray *, and create a local NSMutableArray * here.  It helps show that there are no areas that are allowed to edit teh array outside this method.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -0,0 +1,122 @@
...
+        [_accessibleElements addObject:accessibilityElement];
+      }
+    });
+    return _accessibleElements;
+  }
+  
+  // Handle not rasterize case
+  // Create UI accessiblity elements for each subnode that represent an elment within the accessibility container
+  for (ASDisplayNode *subnode in selfNode.subnodes) {
+    if (subnode.isAccessibilityElement) {
```
appleguy: How does this work if there is an accessibility element at the bottom of a layer-backed tree?  E.g. 3 levels of layer-backed subnodes deep.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -709,141 +709,223 @@ - (void)setEdgeAntialiasingMask:(unsigned int)edgeAntialiasingMask
...
+#pragma mark - UIViewBridgeAccessibility
+
+// ASDK supports accessibility for view or layer backed nodes. To be able to provide support for layer backed
+// nodes, properties for all of the UIAccessibility protocol defined properties need to be provided an held in sync
+// between node and view
+
+// Helper function with following logic:
+// - If the node is not loaded yet use the property from the pending state
+// - In case the node is loaded
+//  - Check if the node has a view and get the
```
appleguy: "and get the..." not finished

appleguy:
@maicki Great work on this! I am urgent now because it is very close and I know it is challenging to continue modifying a diff of this size. Feel free to address the small areas of follow-up in a later change in the next week or two.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: UIAccessibilityContainer is a class extension no protocol, so we cannot add `id < UIAccessibilityContainer>` to it

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: Addressed in #1513

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: Addressed in #1513

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: Addressed and more information in #1513

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
maicki: Addressed in #1513

maicki:
@appleguy New pull request: #1513. This adds accessibilityElements support for layer backed nodes within layer backed nodes and addressed the areas you are commented on

Issue #341, opened 2015-02-27
[ASBatchContext] Batch fetch call won't occur for content smaller than bounds, unless user scrolls.


lxcid:
I am excited about c657411a2e99e3a2195a03b1382c40815ff986a2 but after I tried it out it doesn't seems to work yet.

For example, if you set `kLitterSize` to 1 and/or `kLitterBatchSize` to 1 to simulate small dataset, after the first batch is loaded, the content size at this moment is smaller than the viewport but it doesn't call `handleBatchFetchScrollingToOffset:` which will trigger the small batch size logic. The `scrollViewWillEndDragging:withVelocity:targetContentOffset:` doesn't get trigger when you add/remove rows though. :(

Another thing I noticed is that

```
  // target offset will always be 0 if the content size is smaller than the viewport
  BOOL hasSmallContent = offset == 0.0 && contentLength < viewLength;
```

For the following code, target offset might not be zero if the content inset of the scroll view is being modified, which is often the case as view controller automatically adjust it post iOS 7 unless you opt out (Such bad API, I know)...

secretiverhyme:
@rnystrom, while we're at it, can we revisit the `-[ASBatchContext completeBatchFetching:]` API?  Passing `NO` is a no-op, and the presence of `-[ASBatchContext cancelBatchFetching]` makes the argument seem superfluous.

appleguy:
@lxcid I just confirmed that this is still happening. Disappointed that it took a year to fix this bug, but at least I can say with confidence that the framework development has been focused on some really big things in the meantime. It is now absolutely time to clean up and improve the batch fetching APIs though. This will be fixed in the next release.

@maicki, this change reproduces the issue. If the user initiates a drag, even though it will rubber band back into place, that will trigger the load. I think that is why this issue has not been widely reported. Probably the right thing to do is to use a new Hook from the ranch controller to check whether a batch fetch needs to be triggered, because the range controller runs its update pass in a variety of situations including edit operations and the appearance or disappearance of cells.

-static const NSInteger kLitterSize = 20;            // intial number of kitten cells in ASTableView
+static const NSInteger kLitterSize = 1;            // intial number of kitten cells in ASTableView

appleguy:
@maicki: thanks for fixing this long standing issue. @lxcid - Sorry this took so long to get around too. I hope you are still using ASDK. Let us know if we can help you out with any more issues.

lxcid:
[...]

PR #1477, opened 2016-04-05
Fix fetch call won't occur for content smaller than bounds unless user scrolls

maicki:
Initial attempt to fix batch fetch call won't occur for content smaller than bounds, unless user scrolls. Addresses and further information at issue #341

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -91,7 +91,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 #pragma mark -
 #pragma mark ASCollectionView.
 
-@interface ASCollectionView () <ASRangeControllerDataSource, ASRangeControllerDelegate, ASDataControllerSource, ASCellNodeLayoutDelegate, ASDelegateProxyInterceptor> {
+@interface ASCollectionView () <ASRangeControllerDataSource, ASRangeControllerDelegate, ASDataControllerSource, ASCellNodeLayoutDelegate, ASDelegateProxyInterceptor, ASBatchFetchingScrollView> {
   ASCollectionViewProxy *_proxyDataSource;
   ASCollectionViewProxy *_proxyDelegate;
   
@@ -605,8 +605,45 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   cellNode.scrollView = nil;
 }
 
-#pragma mark -
-#pragma mark Scroll Direction.
+
+- (void)scrollViewDidScroll:(UIScrollView *)scrollView
+{
+  // If a scroll happenes the current range mode needs to go to full
+  ASInterfaceState interfaceState = [self interfaceStateForRangeController:_rangeController];
+  if (ASInterfaceStateIncludesVisible(interfaceState)) {
+    [_rangeController updateCurrentRangeWithMode:ASLayoutRangeModeFull];
+  }
+  
+  for (_ASCollectionViewCell *collectionCell in _cellsForVisibilityUpdates) {
+    // Only nodes that respond to the selector are added to _cellsForVisibilityUpdates
+    [[collectionCell node] cellNodeVisibilityEvent:ASCellNodeVisibilityEventVisibleRectChanged
+                                      inScrollView:scrollView
+                                     withCellFrame:collectionCell.frame];
+  }
+  if (_asyncDelegateImplementsScrollviewDidScroll) {
+    [_asyncDelegate scrollViewDidScroll:scrollView];
+  }
+}
+
+- (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoint)velocity targetContentOffset:(inout CGPoint *)targetContentOffset
+{
+  _deceleratingVelocity = CGPointMake(
+    scrollView.contentOffset.x - ((targetContentOffset != NULL) ? targetContentOffset->x : 0),
+    scrollView.contentOffset.y - ((targetContentOffset != NULL) ? targetContentOffset->y : 0)
+  );
+
+  if (targetContentOffset != NULL) {
+    ASDisplayNodeAssert(_batchContext != nil, @"Batch context should exist");
+    [self _beginBatchFetchingIfNeededWithScrollView:self forScrollDirection:[self scrollDirection] contentOffset:*targetContentOffset];
+  }
+  
+  if ([_asyncDelegate respondsToSelector:@selector(scrollViewWillEndDragging:withVelocity:targetContentOffset:)]) {
+    [_asyncDelegate scrollViewWillEndDragging:scrollView withVelocity:velocity targetContentOffset:targetContentOffset];
+  }
+}
+
+
+#pragma mark - Scroll Direction.
 
 - (ASScrollDirection)scrollDirection
 {
@@ -700,58 +737,14 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 }
 
 
-#pragma mark -
-#pragma mark Batch Fetching
+#pragma mark - Batch Fetching
 
-- (void)_checkForBatchFetching
+- (ASBatchContext *)batchContext
 {
-  // Dragging will be handled in scrollViewWillEndDragging:withVelocity:targetContentOffset:
-  if ([self isDragging] || [self isTracking] || ![self _shouldBatchFetch]) {
-    return;
-  }
-  
-  // Check if we should batch fetch
-  if (ASDisplayShouldFetchBatchForContext(_batchContext, [self scrollableDirections], self.bounds, self.contentSize, self.contentOffset, _leadingScreensForBatching)) {
-    [self _beginBatchFetching];
-  }
+  return _batchContext;
 }
 
-- (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoint)velocity targetContentOffset:(inout CGPoint *)targetContentOffset
-{
-  _deceleratingVelocity = CGPointMake(
-    scrollView.contentOffset.x - ((targetContentOffset != NULL) ? targetContentOffset->x : 0),
-    scrollView.contentOffset.y - ((targetContentOffset != NULL) ? targetContentOffset->y : 0)
-  );
-
-  if (targetContentOffset != NULL) {
-    [self _handleBatchFetchScrollingToOffset:*targetContentOffset];
-  }
-  
-  if ([_asyncDelegate respondsToSelector:@selector(scrollViewWillEndDragging:withVelocity:targetContentOffset:)]) {
-    [_asyncDelegate scrollViewWillEndDragging:scrollView withVelocity:velocity targetContentOffset:targetContentOffset];
-  }
-}
-
-- (void)scrollViewDidScroll:(UIScrollView *)scrollView
-{
-  // If a scroll happenes the current range mode needs to go to full
-  ASInterfaceState interfaceState = [self interfaceStateForRangeController:_rangeController];
-  if (ASInterfaceStateIncludesVisible(interfaceState)) {
-    [_rangeController updateCurrentRangeWithMode:ASLayoutRangeModeFull];
-  }
-  
-  for (_ASCollectionViewCell *collectionCell in _cellsForVisibilityUpdates) {
-    // Only nodes that respond to the selector are added to _cellsForVisibilityUpdates
-    [[collectionCell node] cellNodeVisibilityEvent:ASCellNodeVisibilityEventVisibleRectChanged
-                                      inScrollView:scrollView
-                                     withCellFrame:collectionCell.frame];
-  }
-  if (_asyncDelegateImplementsScrollviewDidScroll) {
-    [_asyncDelegate scrollViewDidScroll:scrollView];
-  }
-}
-
-- (BOOL)_shouldBatchFetch
+- (BOOL)canBatchFetch
 {
   // if the delegate does not respond to this method, there is no point in starting to fetch
   BOOL canFetch = [_asyncDelegate respondsToSelector:@selector(collectionView:willBeginBatchFetchWithContext:)];
@@ -762,25 +755,41 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   }
 }
 
-- (void)_handleBatchFetchScrollingToOffset:(CGPoint)targetOffset
+- (void)_scheduleCheckForBatchFetching
 {
-  ASDisplayNodeAssert(_batchContext != nil, @"Batch context should exist");
-  
-  if (![self _shouldBatchFetch]) {
+  // Push this to the next runloop to be sure the UITableView has the right content size
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [self _checkForBatchFetching];
+  });
+}
+
+- (void)_checkForBatchFetching
+{
+  // Dragging will be handled in scrollViewWillEndDragging:withVelocity:targetContentOffset:
+  if (self.isDragging || self.isTracking) {
     return;
   }
   
-  if (ASDisplayShouldFetchBatchForContext(_batchContext, [self scrollDirection], self.bounds, self.contentSize, targetOffset, _leadingScreensForBatching)) {
+  [self _beginBatchFetchingIfNeededWithScrollView:self forScrollDirection:[self scrollableDirections] contentOffset:self.contentOffset];
+}
+
+- (void)_beginBatchFetchingIfNeededWithScrollView:(UIScrollView<ASBatchFetchingScrollView> *)scrollView forScrollDirection:(ASScrollDirection)scrollDirection contentOffset:(CGPoint)contentOffset
+{
+  if (ASDisplayShouldFetchBatchForScrollView(self, scrollDirection, contentOffset)) {
     [self _beginBatchFetching];
   }
 }
 
 - (void)_beginBatchFetching
 {
+  NSLog(@"begin batch fetching");
+    
   [_batchContext beginBatchFetching];
-  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-    [_asyncDelegate collectionView:self willBeginBatchFetchWithContext:_batchContext];
-  });
+  if ([_asyncDelegate respondsToSelector:@selector(collectionView:willBeginBatchFetchWithContext:)]) {
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+      [_asyncDelegate collectionView:self willBeginBatchFetchWithContext:_batchContext];
+    });
+  }
 }
 
 
@@ -972,7 +981,10 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
       for (dispatch_block_t block in _batchUpdateBlocks) {
         block();
       }
-    } completion:completion];
+    } completion:^(BOOL finished){
+      [self _scheduleCheckForBatchFetching];
+      if (completion) { completion(finished); }
+    }];
   });
   
   [_batchUpdateBlocks removeAllObjects];
@@ -993,11 +1005,10 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     }];
   } else {
     [_layoutFacilitator collectionViewWillEditCellsAtIndexPaths:indexPaths batched:NO];
-    ASPerformBlockWithoutAnimationCompletion(YES, ^{
+    [UIView performWithoutAnimation:^{
       [super insertItemsAtIndexPaths:indexPaths];
-    }, ^{
-      
-    });
+      [self _scheduleCheckForBatchFetching];
+    }];
   }
 }
 
@@ -1017,6 +1028,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     [_layoutFacilitator collectionViewWillEditCellsAtIndexPaths:indexPaths batched:NO];
     [UIView performWithoutAnimation:^{
       [super deleteItemsAtIndexPaths:indexPaths];
+      [self _scheduleCheckForBatchFetching];
     }];
   }
 }
@@ -1037,6 +1049,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     [_layoutFacilitator collectionViewWillEditSectionsAtIndexSet:indexSet batched:NO];
     [UIView performWithoutAnimation:^{
       [super insertSections:indexSet];
+      [self _scheduleCheckForBatchFetching];
     }];
   }
 }
@@ -1057,6 +1070,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     [_layoutFacilitator collectionViewWillEditSectionsAtIndexSet:indexSet batched:NO];
     [UIView performWithoutAnimation:^{
       [super deleteSections:indexSet];
+      [self _scheduleCheckForBatchFetching];
     }];
   }
 }
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -757,7 +757,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 
 - (void)_scheduleCheckForBatchFetching
 {
-  // Push this to the next runloop to be sure the UITableView has the right content size
+  // Push this to the next runloop to be sure the scroll view has the right content size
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/Private/ASInternalHelpers.h
```diff
@@ -33,6 +33,26 @@ BOOL ASRunningOnOS7();
...
+ @summary Conditionally performs UIView geometry changes in the given block without animation and call completion block afterwards.
+ 
+ Used primarily to circumvent UITableView forcing insertion animations when explicitly told not to via
+ `UITableViewRowAnimationNone`. More info: https://github.com/facebook/AsyncDisplayKit/pull/445
+ 
+ @param withoutAnimation Set to `YES` to perform given block without animation
+ @param block Perform UIView geometry changes within the passed block
+ @param completion Call completion block if UIView geometry changes within the passed block did complete
+ */
+ASDISPLAYNODE_INLINE void ASPerformBlockWithoutAnimationCompletion(BOOL withoutAnimation, void (^block)(), void (^completion)()) {
```
appleguy: Should this function name include WITHOUT or WITH completion?

This seems like a bit risky of an approach, but something that could be extremely beneficial for reducing the massive overhead in collection views that we are seeing from animations. Is it necessary to include this in this change? It seems like it should not be required for the upgrade to the loading system to work correctly, but if it is, please share more details so that I can properly understand the interaction

AsyncDisplayKit/ASTableView.mm
```diff
@@ -675,6 +703,19 @@ - (void)tableView:(UITableView *)tableView didEndDisplayingCell:(_ASTableViewCel
 #pragma mark - 
 #pragma mark Batch Fetching
 
+- (void)_checkForBatchFetching
+{
+  // Dragging will be handled in scrollViewWillEndDragging:withVelocity:targetContentOffset:
+  if ([self isDragging] || [self isTracking] || ![self _shouldBatchFetch]) {
```
appleguy: Can we move this logic into ASBatchFetching.m, thus removing the duplication with the table and collection implementations? I know there are many other things that are duplicated in these files, which we can resolve with the plan to create a shared superclass, but that refactoring will be easier if it is clear that this logic is already unified.

Probably the best thing is to change the expected arguments of the ASBatchFetching.m call, to accept they scroll view instead of accessing the bounds, content size, and content offset. Because the bounds includes the content offset anyway, we should only request bounds and not call into the Apple framework to access content offset.

It would be fine to create a new function type to start splitting up some of the overall logic, but combining it into the single longer function that decides if it is the right time to batch fetch would be fine as well.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -749,22 +762,27 @@ - (BOOL)shouldBatchFetch
...
+- (void)_handleBatchFetchScrollingToOffset:(CGPoint)targetOffset
 {
   ASDisplayNodeAssert(_batchContext != nil, @"Batch context should exist");
   
-  if (![self shouldBatchFetch]) {
+  if (![self _shouldBatchFetch]) {
     return;
   }
   
   if (ASDisplayShouldFetchBatchForContext(_batchContext, [self scrollDirection], self.bounds, self.contentSize, targetOffset, _leadingScreensForBatching)) {
```
appleguy: This call point is very similar to _checkForBatchFetching.  Let's unify them, maybe remove _checkForBatchFetching and this method and replace with _batchFetchIfNeededForContentOffset:.  You could just pass as the argument self.contentOffset when not using targetOffset.

appleguy:
@maicki this is actually a very important chains that I have been wanting to do for a long time, but never had a chance to investigate. The changes you have made are a substantial improvement and also demonstrate an opportunity to further simplify the implementation without much work. Hopefully this will help us ensure that the incrementation is now more robust than it has ever been.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -587,18 +587,46 @@ - (ASScrollDirection)scrollDirection
...
+- (ASScrollDirection)_scrollDirectionForVelocity:(CGPoint)scrollVelocity
 {
   ASScrollDirection direction = ASScrollDirectionNone;
-  if (velocity.y < 0.0) {
-    direction = ASScrollDirectionDown;
-  } else if (velocity.y > 0.0) {
-    direction = ASScrollDirectionUp;
+  ASScrollDirection scrollableDirections = [self scrollableDirections];
+  
+  if (ASScrollDirectionContainsHorizontalDirection(scrollableDirections)) { // Can scroll horizontally.
```
appleguy: Do we really want this behavior in table? I am not aware of a case that it can be made to scroll horizontally, at least in a scenario that we would want to trigger either a range controller or batch operation.

For example, you could transform by rotation 90º and have it scroll horizontally, but from the perspective of the table I believe it would still have only a Y value change and internally behave as though vertical

It is a great point that we can enable scroll bouncing, but even in this case I don't think we need a batch or range controller?

AsyncDisplayKit/ASTableView.mm
```diff
@@ -848,10 +894,15 @@ - (void)rangeController:(ASRangeController *)rangeController didInsertNodes:(NSA
   if (!self.asyncDataSource) {
     return; // if the asyncDataSource has become invalid while we are processing, ignore this request to avoid crashes
   }
-
+  
   BOOL preventAnimation = animationOptions == UITableViewRowAnimationNone;
-  ASPerformBlockWithoutAnimation(preventAnimation, ^{
+  ASPerformBlockWithoutAnimationCompletion(preventAnimation, ^{
     [super insertRowsAtIndexPaths:indexPaths withRowAnimation:(UITableViewRowAnimation)animationOptions];
+  }, ^{
```
appleguy: Instead of this occurring after the animation, does it work to dispatch it after the insert call within the blocked provided to the no animation API? Even in the case that an animation occurs, I believe all of the values are supposed to be updated to their final value. This would be important to check, but I think even the content side should reflect the new value. If true, it would also provide a performance advantage to do this, as slow network loads can be initiated about 300 ms sooner. But I think the biggest advantage would be lower complexity in understanding how this code works and not immediately requiring a adoption of a new type of no-animation handler

This is not a blocker for landing the change, so prioritize other updates to the change first and we can come back to this after fully testing the environment and accessibility APIs.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -869,8 +920,13 @@ - (void)rangeController:(ASRangeController *)rangeController didDeleteNodes:(NSA
   }
 
   BOOL preventAnimation = animationOptions == UITableViewRowAnimationNone;
-  ASPerformBlockWithoutAnimation(preventAnimation, ^{
+  ASPerformBlockWithoutAnimationCompletion(preventAnimation, ^{
     [super deleteRowsAtIndexPaths:indexPaths withRowAnimation:(UITableViewRowAnimation)animationOptions];
```
appleguy: insert and delete are certainly the most important, but it is definitely possible for a move operation or even a reload operation to resulted in an effective changing the distance to the bottom of the content and require a batch fetch.

It is also necessary to check for changes to the sections. Is there any override point that exists where we can check for batch fetching after any form of editing? I think it should be relatively efficient to do so, and so more calls are better than fewer in this case, but we really should have a way to avoid putting the call manually in each method

If there is no other override point, we should rename the no-animation function similarly to how are you have done here, but make it specific to the table or collection. Then in that function, used for all of the editing operations in these positions, we can put the batch fetch call and also call the block inside the no animation API if the provided argument indicates that we don't want animation.

AsyncDisplayKit/Details/ASBatchContext.h
```diff
@@ -34,6 +34,14 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)completeBatchFetching:(BOOL)didComplete;
 
 /**
+ * Let the context object know that a batch fetch was completed.
+ *
+ * @discussion For instance, when a table has reached the end of its data, a batch fetch will be attempted unless the context
+ * object thinks that it is still fetching.
+ */
+- (void)completeBatchFetching;
```
appleguy: Is this simply a convenience for the method above? If so, the discussion should probably just mention that it is equivalent to the above call passing YES, and there should be a @see annotation for the other method.

AsyncDisplayKit/Private/ASInternalHelpers.h
maicki: If you look at the whole diff you will see the only thing I changed is adding a new function with completion block. Before there already was this kind of function just without the completion block. To be able to use a completion block we have to move from the UIView performWithoutAnimation API to CATransaction based API as we can set the completion block there. So theoretically nothing should have changed.

AsyncDisplayKit/Details/ASBatchContext.h
maicki: See the discussion in the initial issue #341 to deprecate the old method where we pass in a BOOL value. I don't see a value for the old method to be honest as if you pass in NO nothing will basically happen. I didn't mark the old method as deprecated yet in the header file though, but we should think about that for 2.0

AsyncDisplayKit/ASTableView.mm
```diff
@@ -683,15 +655,78 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
...
+  ASScrollDirection scrollDirection = [self _scrollDirectionForVelocity:scrollVelocity];
+  return ASScrollDirectionApplyTransform(scrollDirection, self.transform);
+}
+
+- (ASScrollDirection)_scrollDirectionForVelocity:(CGPoint)scrollVelocity
+{
+  ASScrollDirection direction = ASScrollDirectionNone;
+  ASScrollDirection scrollableDirections = [self scrollableDirections];
+  
+  if (ASScrollDirectionContainsHorizontalDirection(scrollableDirections)) { // Can scroll horizontally.
```
appleguy: Check out my earlier comments on this part - we don't need anything related to horizontal right?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -605,8 +605,45 @@ - (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(
   cellNode.scrollView = nil;
 }
 
-#pragma mark -
-#pragma mark Scroll Direction.
+
+- (void)scrollViewDidScroll:(UIScrollView *)scrollView
+{
+  // If a scroll happenes the current range mode needs to go to full
```
appleguy: happens*.  Is it necessary to move these methods?  Although we should definitely look for ways to improve the grouping of methods, it is somewhat important to avoid doing this in the same commit as careful changes to Core logic.  e.g.:

- Unable to see in code review, or subsequently when looking at blame, how the code changed at this point
- Makes overall size of diff larger / more careful cross-diff analysis to understand change so it is tougher to confidently accept

If these methods are _only_ moved with no changes, that is totally fine — just leave a comment to that effect when the diff goes up.  Also, in general I think it is awesome to include small cleanup fixes with other changes (line spacng or typo corrections etc, variable renames, etc especally if they are not closely intertangled with the critical sections of the change)

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -749,16 +755,37 @@ - (BOOL)shouldBatchFetch
   }
 }
 
-- (void)handleBatchFetchScrollingToOffset:(CGPoint)targetOffset
+- (void)_scheduleCheckForBatchFetching
 {
-  ASDisplayNodeAssert(_batchContext != nil, @"Batch context should exist");
-  
-  if (![self shouldBatchFetch]) {
+  // Push this to the next runloop to be sure the UITableView has the right content size
```
appleguy: UICollectionView.  But we should probably update this to be more generic so the comment can remain the same in table as well.

Can this logic be moved into the shared area and the methods defined in the protocol for batch fetching scroll view? Even if there are four or five methods in the protocol, if the logic is indeed identical between table and collection, and if the methods seem to make sense to share in this way which at first look it appears that they may be a good fit for, this would certainly be a Great cleanup of the code and help move additional complexity out of these very large files.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -749,16 +755,37 @@ - (BOOL)shouldBatchFetch
...
+  dispatch_async(dispatch_get_main_queue(), ^{
+    [self _checkForBatchFetching];
+  });
+}
+
+- (void)_checkForBatchFetching
+{
+  // Dragging will be handled in scrollViewWillEndDragging:withVelocity:targetContentOffset:
+  if (self.isDragging || self.isTracking) {
```
appleguy: My last feedback mentioned moving these particular checks into the shared function for should batch fetch. Perhaps there is a reason it is not best to unify the logic there, but if you consider this and it seems to have no negative consequences, we can proceed in that direction.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -749,16 +755,37 @@ - (BOOL)shouldBatchFetch
...
+  }
+}
+
+- (void)_beginBatchFetching
+{
+  NSLog(@"begin batch fetching");
+    
+  [_batchContext beginBatchFetching];
+  if ([_asyncDelegate respondsToSelector:@selector(collectionView:willBeginBatchFetchWithContext:)]) {
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
appleguy: It seems a bit strange to me that we are dispatching to a background queue for this call. Most delicate calls are on the main thread, Although perhaps there is a common use case for this one where it is more convenient to avoid overhead by dispatching so the developer does not have to. I know this has not change, but if you could check that it is comment properly in the header, that would be great.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -997,6 +1028,7 @@ - (void)rangeController:(ASRangeController *)rangeController didDeleteNodes:(NSA
     [_layoutFacilitator collectionViewWillEditCellsAtIndexPaths:indexPaths batched:NO];
     [UIView performWithoutAnimation:^{
       [super deleteItemsAtIndexPaths:indexPaths];
```
appleguy: Please check out my feedback on the last change, which described how move and reload operations can also trigger the need for a Batch request.  Is this maybe an incorrect assumption?

We should probably have a mechanism to defer checking for batching if currently handling a batch EDIT operations, until the commit / endEditing occurs.  It seems like having these checks in every added method may add up with some overhead unless we have a simple strategy for coalescing them. This is probably not essential at this stage but is worth thinking about.

The feedback on the earlier change also asked that we investigate a simpler way to include all of these calls, because I believe there will be at least eight, And described one approach we could use for this. This is not a blocker for the change landing, but it would be great to hear your thoughts on each area feedback written - so I know if perhaps I missed something and did not realize that the suggestion is not applicable at all, or have a sense if you would suggest that we wait until a future improvement of the code to adopted that change.

AsyncDisplayKit/Details/ASBatchContext.h
```diff
@@ -59,6 +56,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ * Let the context object know that a batch fetch was completed.
+ *
+ * @param didComplete A boolean that states whether or not the batch fetch completed.
+ *
+ * @discussion Only by passing YES will the owner of the context know to attempt another batch update when necessary.
+ * For instance, when a table has reached the end of its data, a batch fetch will be attempted unless the context
+ * object thinks that it is still fetching.
+ */
+- (void)completeBatchFetching:(BOOL)didComplete;
```
appleguy: I think I asked about this in my last feedback as well, but could you explain to me why this is now deprecated?

AsyncDisplayKit/Private/ASInternalHelpers.h
appleguy: One question, which I'm not sure of, is if the UIKit Version has any other Side effects. I would hope it is a pure convenience for the Core animation method, but it also seems possible that it could affect how the table internally performs operations

That said, I am fine with us changing to this approach now that I better understand it. However, referencing my comments on the first revision it seems like the best approach to support the need to check for batch fetching on every edited operation could be to have a custom replacement for this method that is specific to table and collection and handles calling the Batch check API.

It would be a good idea to introduce a protocol, Like the batch fetching scroll view, to expose this kind of API and allow a shared piece of code to rely on it. It may make sense to make the shared protocol something more general, such as ASRangeView or something, as a precursor to the larger project to create a shared superclass for the table and collection node. The protocol might allow us to start creating shared method names that can be used externally I objects that are referencing only a pointer that conforms to that protocol rather than a table or collection specifically. This protocol stuff is not required for the first version of this change, although it may be very quick to use such an approach to ensure the editing methods are able to disable animation and schedule a batch call in one convenient line rather than eight separate call points

AsyncDisplayKit/ASCollectionView.mm
maicki: I wondered about it too! The header comment is in fact documenting that properly, but in my opinion this callback should happen on the main thread.

AsyncDisplayKit/ASTableView.mm
maicki: Yes it's not necessary ... missed the comment.

AsyncDisplayKit/ASCollectionView.mm
maicki: Yes I changed the grouping of the methods to have the same grouping in ASCollectionView and ASTableView. Before this was not the case. Furthermore this is also in preparation for further consolidating of logic of ASCollectionView and ASTableView.

AsyncDisplayKit/ASCollectionView.mm
maicki: Added a more generic comment

AsyncDisplayKit/ASCollectionView.mm
maicki: We cannot really move this check into the shared function for "should batch fetch": `_beginBatchFetchingIfNeededWithScrollView:forScrollDirection:contentOffset:` as this method will not only be called from `_checkForBatchFetching` but also from `scrollViewWillEndDragging:withVelocity:targetContentOffset:` and in there "isDragging" is actually true so it would return and not start the batch fetching.

AsyncDisplayKit/ASCollectionView.mm
maicki: To point 1: Look into `rangeController:didEndUpdatesAnimated:completion:` there you see we call `_scheduleCheckForBatchFetching` in the completion handler. This should be the behavior you mention in your comment, it's deferring the check for batch fetching until all the `_batchUpdateBlocks` have finished.

AsyncDisplayKit/Details/ASBatchContext.h
maicki: I revert the code for now. In my opinion would be good to revisit the the `-[ASBatchContext completeBatchFetching:]` API e.g. Passing in `NO` is basically a no op.

AsyncDisplayKit/ASTableView.mm
maicki: Regarding `isDragging` and `isTracking` see my answer to your comment a bit later.

A new function for combining logic around batch fetching for ASCollectionView and ASTableView was created with latest commit.

AsyncDisplayKit/ASCollectionView.mm
maicki: That was done with latest commit

AsyncDisplayKit/ASTableView.mm
maicki: Yes we don't need that. I removed horizontal direction behavior for ASTableView.

AsyncDisplayKit/ASTableView.mm
maicki: Added the dispatch calls after the inserts, deletes etc. and the content size reflects the new size after inserting so if a new check if a batch should occur happens shortly after it should have the updated value to check if a batch should happen or not.

This would also remove the need of a new no animation handler.

AsyncDisplayKit/Private/ASInternalHelpers.h
maicki: The UIKit version should just be a pure convenience for the Core Animation method. As we don't need the completion block function anymore I reverted the code back from the Core Animation to the UIKit one to prevent any regressions.

I will look into adding a protocol for need to check for shared behavior around check for batch fetching on every edited operations or similar functionality.

AsyncDisplayKit/ASCollectionView.mm
appleguy: Out of curiosity, because I'm not sure, is it important to check for dragging and to not allow the batch to be triggered if the drag is in progress? It seems like if the drag does reached the threshold, we should trigger the load immediately at the point that it does reach it

AsyncDisplayKit/ASCollectionView.mm
appleguy: I did see the dispatch back to the main queue, but I believe we dispatch every time even if a dispatch has already occurred, So the check will end up running many times. This could be solved in the same way that coalescing is done by the range controller, having a simple Boolean flag that indicates that a scheduling has occurred but not yet executed, And so we skip scheduling in this case.

AsyncDisplayKit/ASCollectionView.mm
appleguy: Okay, that sounds great. Just to confirm, there're no logic changes here, Correct? No need to reply if that is true and I'm glad that the method order now matches.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -683,15 +655,71 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
...
+{
+  CGPoint scrollVelocity;
+  if (self.isTracking) {
+    scrollVelocity = [self.panGestureRecognizer velocityInView:self.superview];
+  } else {
+    scrollVelocity = _deceleratingVelocity;
+  }
+  
+  ASScrollDirection scrollDirection = [self _scrollDirectionForVelocity:scrollVelocity];
+  return ASScrollDirectionApplyTransform(scrollDirection, self.transform);
```
appleguy: This code is probably pre-existing, but do we actually want to apply the transform? I think there could be a confusion here, and I believe we never want to apply to transform, because the loading logic is based to work on the directions as they apply without the transform and it doesn't actually matter if that is flipped from the the what the user sees

AsyncDisplayKit/ASTableView.mm
```diff
@@ -683,15 +655,71 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
...
+  return direction;
+}
+
+- (ASScrollDirection)scrollableDirections
+{
+  ASScrollDirection scrollableDirection = ASScrollDirectionNone;
+  CGFloat totalContentWidth = self.contentSize.width + self.contentInset.left + self.contentInset.right;
+  CGFloat totalContentHeight = self.contentSize.height + self.contentInset.top + self.contentInset.bottom;
+  
+  if (self.alwaysBounceHorizontal || totalContentWidth > self.bounds.size.width) { // Can scroll horizontally.
```
appleguy: New need for checking the horizontal or the variable above for content with, correct? Even if there is horizontal scrolling possible, I don't believe this has an impact on any System integrated with the table

AsyncDisplayKit/ASTableView.mm
```diff
@@ -702,16 +730,35 @@ - (BOOL)shouldBatchFetch
   }
 }
 
-- (void)handleBatchFetchScrollingToOffset:(CGPoint)targetOffset
+- (void)_scheduleCheckForBatchFetching
 {
-  ASDisplayNodeAssert(_batchContext != nil, @"Batch context should exist");
+  // Push this to the next runloop to be sure the scroll view has the right content size
+  dispatch_async(dispatch_get_main_queue(), ^{
```
appleguy: This is the case I was referring to, where multiple calls will schedule multiple calls to the Block, Meaning they won't be actually coalesced.

appleguy:
@maicki I'm going to take this change now as it is definitely an important improvement compared to where we were before. However, there are a few important follow-ups, and please quickly review my comments to make sure that I did not forget to summarize any of them here.

- The transform logic being applied to the scroll direction is **very likely incorrect / should not be applied**, with potentially significant results.  We should check who originally added this.  If it was before my diffs that flipped / made consistent the definitions of Up, Down, etc...then the author probably thought it was necessary because of another bug, when it is in fact not needed and will cause problems like potentially no batch fetching or a flipped working range for transformed views.

- Coalescing should not schedule multilpe call to check the batch, if one is already pending.

- Remove additional logic around horizontal for tables

- Consider if batching should actually be skipped in the case of dragging, and if not, move the other check into the shared code

- **handle moving and reloading of both rows and sections** which can trigger the need to batch fetch.  While I will definitely accept / merge the diff by just adding additional calls, it might make sense to have a shared override point that does both the "no animation" call and the "schedule check for batch update"

Issue #1468, opened 2016-04-01
[ASVideoNode] Mute only works if accessing through to the underlying .player object.


klivin:
Hi there, incredible work guys!  I would like to use the muted boolean on ASVideoNode to mute and unmute an already playing video.

Muted only seems to be respected on initialization, but does not work if set while the video is already playing.

MarvinNazari:
yeah this is a bug, a workaround is to change the `muted` property on player itself
```swift
videoNode.player.muted = true
```

klivin:
Thank you!  Works like a charm :)

appleguy:
@klivin @MarvinNazari @lappp9 thanks for the report and workaround! Does anyone have time to create a small PR to address the bug?

lappp9:
Yeah I can look at it tonight or tomorrow!

lappp9:
@klivin put in a PR for this today!

appleguy:
@lappp9 - thanks for the PR!
@klivin - please try out the fix on master and let us know if you have any issues with it.

PR #1478, opened 2016-04-05
[ASVideoNode] Ensure mute property is passed onto the internal player

lappp9:
Addresses #1468 and mutes the player when the node is muted.


AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -21,6 +21,7 @@ @interface ASVideoNodeTests : XCTestCase
 
 @interface ASVideoNode () {
```
appleguy: In the category () probably put Tests or something if these vars should never be used otherwise

appleguy:
Thanks for adding tests!!

PR #1482, opened 2016-04-05
Fix setting and add merging ASEnvironmentStateExtensions

maicki:


AsyncDisplayKit/Private/ASEnvironmentInternal.h
```diff
@@ -35,6 +35,8 @@ void ASEnvironmentPerformBlockOnObjectAndParents(id<ASEnvironment> object, void(
 
 #pragma mark - Merging
 
+static const struct ASEnvironmentStateExtensions ASEnvironmentDefaultStateExtensions = {};
```
appleguy: I assume the answer is yes, but is this format defined to initialize each Field to the correct zero value?

PR #1484, opened 2016-04-05
[ASControlNode] Deduplicate Target-Action-Events Entries, Fix Exception When Removing Nil Target

Adlai-Holler:
This matches the UIControl behavior based on: [DuplicateActionsTest.playground.zip]( https://github.com/facebook/AsyncDisplayKit/files/205466/DuplicateActionsTest.playground.zip)


appleguy:
Thank you, @Adlai-Holler!

PR #1485, opened 2016-04-06
[ASImageNode] Add support for animated GIFs, planning for future animated formats as well.

garrettmoon:
Adds support for animated images (GIFs at this time) to ASDK.


AsyncDisplayKit/ASImageNode+AnimatedImage.h
```diff
@@ -0,0 +1,23 @@
+//
+//  ASImageNode+AnimatedImage.h
+//  Pods
```
appleguy: Probably shouldn't say Pods - copy another header, feel free to keep your name and date :)

AsyncDisplayKit/ASImageNode+AnimatedImage.h
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Garrett Moon on 3/22/16.
+//
+//
+
+#import "ASImageNode.h"
+
+#import "ASImageProtocols.h"
+
+@interface ASImageNode ()
```
appleguy: Is it needed to have two different categories?  Nit: remove space between imports and above/below properties, IMO the more distinct groupings are easier to read (might just be tired)

garrettmoon:
What's the best way to indicate that this is Beta to people?

garrettmoon:
FYI, examples will fail until we merge in the PINRemoteImage side of things. Before I do that though, I want to be sure the integration here is correct.

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,255 @@
...
+}
+
+- (void)startAnimating
+{
+  ASDisplayNodeAssertMainThread();
+  if (ASInterfaceStateIncludesVisible(self.interfaceState) == NO) {
+    return;
+  }
+  
+  if (self.animatedImagePaused == YES) {
```
levi: no need to check for yes

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,255 @@
...
+    NSLog(@"invalidating display link");
+  }
+#endif
+  [_displayLink invalidate];
+  _displayLink = nil;
+}
+
+@end
+
+@implementation ASWeakProxy
```
levi: Pull ASWeakProxy out into a separate file

levi:
Reviewed and looks great to me. @appleguy?

AsyncDisplayKit/ASImageNode+AnimatedImage.h
```diff
@@ -0,0 +1,15 @@
...
+//  AsyncDisplayKit
+//
+//  Created by Garrett Moon on 3/22/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import "ASImageNode.h"
+#import "ASImageProtocols.h"
+
+@interface ASImageNode ()
```
appleguy: Does this need to be in a separate file?  To make it more discoverable, we might want to put it in the main header - at most perhaps a separately named category ?

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+
+- (void)setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
+{
+  ASDN::MutexLocker l(_animatedImageLock);
+  if (!ASObjectIsEqual(_animatedImage, animatedImage)) {
+    _animatedImage = animatedImage;
+  }
+  if (animatedImage != nil) {
+    __weak ASImageNode *weakSelf = self;
+    if ([animatedImage respondsToSelector:@selector(setCoverImageReadyCallback:)]) {
```
appleguy: Could you comment about what case(s) do or don't have this selector, whereas apparently all have the playback ready?  Just a short hint to help readability.

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+#import "ASWeakProxy.h"
+
+@implementation ASImageNode (AnimatedImage)
+
+#pragma mark - GIF support
+
+- (void)setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
+{
+  ASDN::MutexLocker l(_animatedImageLock);
+  if (!ASObjectIsEqual(_animatedImage, animatedImage)) {
```
appleguy: Shouldn't we early return in the case the image is the same?

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+
+- (id <ASAnimatedImageProtocol>)animatedImage
+{
+  ASDN::MutexLocker l(_animatedImageLock);
+  return _animatedImage;
+}
+
+- (void)setAnimatedImagePaused:(BOOL)animatedImagePaused
+{
+  ASDN::MutexLocker l(_animatedImagePausedLock);
```
appleguy: Do we need a separate lock for this?  Reduced contention is one thing, but not likely an issue here, and multiple locks can get pretty complicated.

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+  return _animatedImagePaused;
+}
+
+- (void)coverImageCompleted:(UIImage *)coverImage
+{
+  BOOL setCoverImage = YES;
+  {
+    ASDN::MutexLocker l(_displayLinkLock);
```
appleguy: :) MOAR LOKZ

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+  }
+  
+  if (setCoverImage) {
+    self.image = coverImage;
+  }
+}
+
+- (void)animatedImageFileReady
+{
+  dispatch_async(dispatch_get_main_queue(), ^{
```
appleguy: Probably use the ASPerform version that runs synchronously if already on main

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+    [self startAnimating];
+  } else {
+    [self stopAnimating];
+  }
+}
+
+- (void)__enterHierarchy
+{
+  [super __enterHierarchy];
+  [self startAnimating];
```
appleguy: I think these two methods are unnecessary.  If we do need them, we should probably comment.  Also they should probably be the variants like -didEnterHierarchy rather than the __ internal versions.

But, the intent is certainly that visibility be cleared & set appropriately at these points so I'm thinking we don't need it.

examples/ASAnimatedImage/ASAnimatedImage/ViewController.m
```diff
@@ -0,0 +1,37 @@
...
+  ASNetworkImageNode *imageNode = [[ASNetworkImageNode alloc] init];
+  imageNode.URL = [NSURL URLWithString:@"https://s-media-cache-ak0.pinimg.com/originals/07/44/38/074438e7c75034df2dcf37ba1057803e.gif"];
+  imageNode.frame = self.view.bounds;
+  imageNode.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
+  imageNode.contentMode = UIViewContentModeScaleAspectFit;
+  
+  [self.view addSubnode:imageNode];
+}
+
+- (void)didReceiveMemoryWarning {
```
appleguy: I wonder if we should call this project AnimatedGIF for now to make it stand out for discovering how to do that?  You could remove this method override.

examples/ASAnimatedImage/ASAnimatedImage/AppDelegate.m
```diff
@@ -0,0 +1,45 @@
...
+
+@implementation AppDelegate
+
+
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+  // Override point for customization after application launch.
+  return YES;
+}
+
+- (void)applicationWillResignActive:(UIApplication *)application {
```
appleguy: Might as well delete all this boilerplate

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+}
+
+- (void)visibilityDidChange:(BOOL)isVisible
+{
+  [super visibilityDidChange:isVisible];
+  
+  ASDisplayNodeAssertMainThread();
+  if (isVisible) {
+    if (self.animatedImage.coverImageReady) {
+      self.image = self.animatedImage.coverImage;
```
appleguy: Setting self.image still results in an async display pass, even if the image is already decoded...because of the way _ASDisplayLayer integrates with CA display.  We may want to force a synchronous display here.  It's tricky though because it is possible in certain timings that layout may not be fully applied yet here (e.g. -setNeedsLayout called, but -layoutSubviews might run next, depending on whether the range controller dispatches to the next runloop - it does dispatch today, but in the future we don't want it to).

All around complex shittiness here, let's discuss more soon.  This isn't a big problem for now.

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+  [self startAnimating];
+}
+
+- (void)__exitHierarchy
+{
+  [super __exitHierarchy];
+  [self stopAnimating];
+}
+
+- (void)displayLinkFired:(CADisplayLink *)displayLink
```
appleguy: This code seems appropriate to abstract one day, not essential now though.

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+{
+  ASDisplayNodeAssertMainThread();
+
+  CFTimeInterval timeBetweenLastFire;
```
appleguy: you could initialize this to 0, then have one if() that checks if lastDisplayLinkFire > 0

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+  if (self.lastDisplayLinkFire == 0) {
+    timeBetweenLastFire = 0;
+  } else {
+    timeBetweenLastFire = CACurrentMediaTime() - self.lastDisplayLinkFire;
+  }
+  self.lastDisplayLinkFire = CACurrentMediaTime();
```
appleguy: This time is fairly efficient to get, but does use very precise mach time - it may be different between the calls.  Since you will normally need to use it twice and always need it once, a local variable for currentTime would be good

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+  
+  _playHead += timeBetweenLastFire;
+  
+  while (_playHead > self.animatedImage.totalDuration) {
+    _playHead -= self.animatedImage.totalDuration;
```
appleguy: This code needs  a comment.  Why is it a loop?  Why not set the playhead back to 0 rather than subtracting duration?

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -390,9 +395,13 @@ - (void)_lazilyLoadImageIfNecessary
               return;
           }
 
-          if (responseImage != NULL) {
+          if (imageContainer != nil) {
             strongSelf->_imageLoaded = YES;
-            strongSelf.image = responseImage;
+            if ([imageContainer asdk_animatedImageData] && _downloaderImplementsAnimatedImage) {
```
appleguy: very minor, but can you swap the order of the condition?

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -415,16 +424,16 @@ - (void)_lazilyLoadImageIfNecessary
...
 
-        void (^cacheCompletion)(UIImage *) = ^(UIImage *image) {
+        void (^cacheCompletion)(id <ASImageContainerProtocol>) = ^(id <ASImageContainerProtocol> imageContainer) {
           // If the cache UUID changed, that means this request was cancelled.
           if (![_cacheUUID isEqual:cacheUUID]) {
             return;
           }
           
-          if (image == NULL && _downloader != nil) {
+          if ([imageContainer asdk_image] == NULL && _downloader != nil) {
```
appleguy: It seems a bit silly that the original conditions were comparing against NULL in the first place, I guess this code used to use CGImageRef?  Time for nil!

AsyncDisplayKit/Details/ASImageContainerProtocolCategories.m
```diff
@@ -0,0 +1,37 @@
+//
+//  ASImageContainerProtocolCategories.m
```
appleguy: Great file!! <3 simplicity.  Less loving the category naming convention, but I agree it is necessary.

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -1,44 +1,114 @@
...
+#if PIN_ANIMATED_AVAILABLE
+@interface ASPINRemoteImageDownloader () <PINRemoteImageManagerAlternateRepresentationProvider>
+
+@end
+
+@interface PINAnimatedImage (ASPINRemoteImageDownloader) <ASAnimatedImageProtocol>
+
+@end
+
+@implementation PINAnimatedImage (ASPINRemoteImageDownloader)
```
appleguy: Is compilation works equivalently, I think it would be best to move this category implementation to beneath the primary @implementation block below.

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -53,23 +123,39 @@ - (void)fetchCachedImageWithURL:(NSURL *)URL
...
 
 - (nullable id)downloadImageWithURL:(NSURL *)URL
                       callbackQueue:(dispatch_queue_t)callbackQueue
-                   downloadProgress:(void (^)(CGFloat progress))downloadProgressBlock
-                         completion:(void (^)(UIImage *image, NSError * error, id downloadIdentifier))completion
+                   downloadProgress:(ASImageDownloaderProgress)downloadProgress
+                         completion:(ASImageDownloaderCompletion)completion;
 {
-  return [[PINRemoteImageManager sharedImageManager] downloadImageWithURL:URL options:PINRemoteImageManagerDownloadOptionsSkipDecode completion:^(PINRemoteImageManagerResult *result) {
+  return [[self sharedPINRemoteImageManager] downloadImageWithURL:URL options:PINRemoteImageManagerDownloadOptionsSkipDecode completion:^(PINRemoteImageManagerResult *result) {
```
appleguy: I might suggest making this simply [self imageManager], as this file requires PIRI and this method is called in many long lines

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -53,23 +123,39 @@ - (void)fetchCachedImageWithURL:(NSURL *)URL
...
     if (ASDisplayNodeThreadIsMain() && callbackQueue == dispatch_get_main_queue()) {
-      completion(result.image, result.error, result.UUID);      
+#if PIN_ANIMATED_AVAILABLE
+      if (result.alternativeRepresentation) {
+        completion(result.alternativeRepresentation, result.error, result.UUID);
+      } else {
+        completion(result.image, result.error, result.UUID);
+      }
+#else
+      completion(result.image, result.error, result.UUID);
```
appleguy: I think you can have just one call to completion for all cases, and do:

`id <ASImageContainerProtocol> image = (PIN_ANIMATED_AVAILABLE ? result.alternativeRepresentation : result.image);`

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -114,7 +200,19 @@ - (void)setPriority:(ASImageDownloaderPriority)priority withDownloadIdentifier:(
       pi_priority = PINRemoteImageManagerPriorityVeryHigh;
       break;
   }
-  [[PINRemoteImageManager sharedImageManager] setPriority:pi_priority ofTaskWithUUID:downloadIdentifier];
+  [[self sharedPINRemoteImageManager] setPriority:pi_priority ofTaskWithUUID:downloadIdentifier];
+}
+
+#pragma mark - PINRemoteImageManagerAlternateRepresentationProvider
+
+- (id)alternateRepresentationWithData:(NSData *)data options:(PINRemoteImageManagerDownloadOptions)options
```
appleguy: Overall, this seems like a good system. It is a very complex situation for an API to support, but I think you provided enough generality with the structure that a pretty wide array of formats and use cases could be supported!

AsyncDisplayKit/Details/ASWeakProxy.h
```diff
@@ -0,0 +1,17 @@
...
+//  ASWeakProxy.h
+//  AsyncDisplayKit
+//
+//  Created by Garrett Moon on 4/12/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface ASWeakProxy : NSObject
```
appleguy: why not NSProxy?

AsyncDisplayKit/Private/ASImageNode+AnimatedImagePrivate.h
```diff
@@ -0,0 +1,27 @@
+//
+//  ASImageNode+AnimatedImagePrivate.h
```
appleguy: I would actually really prefer if we named this or put it in a +Internal or +FrameworkPrivate, some generally-named thing as a core class of this sort is pretty likely to have additions (and creating files for each is suboptimal for organizing the structure of the project)

AsyncDisplayKit/Details/ASImageContainerProtocolCategories.h
```diff
@@ -0,0 +1,19 @@
+//
+//  ASImageContainerProtocolCategories.h
```
appleguy: Although this file is short and simple, it might be worth merging into one of the other files such as ASImageNode+Internal.  If we should keep it, I think we would find greater parallelism in the name +Categories

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -0,0 +1,226 @@
...
+    return;
+  }
+  
+#if ASAnimatedImageDebug
+  NSLog(@"starting animation: %p", self);
+#endif
+  ASDN::MutexLocker l(_displayLinkLock);
+  if (_displayLink == nil) {
+    _playHead = 0;
+    _displayLink = [CADisplayLink displayLinkWithTarget:[ASWeakProxy weakProxyWithTarget:self] selector:@selector(displayLinkFired:)];
```
appleguy: Why does this require a weak proxy? I think this requires a comment at this location, but ideally we could eliminate it with some other construct.

If we keep the new proxy class, there should also be at least a brief description at the beginning of the class of an example use case. If we can't think of another use case to document it with, I'm not certain we should define it as a generalized primitive.

 Although it is true that it would be harder to find if it is eventually needed by another developer.

appleguy:
@garrettmoon Great work on this change! This feature has been requested many times over the life of the framework, and is even something that the Paper team wanted to do (we mused about adding it on several occasions before the framework was released publicly). It's quite fun and exciting to see something like that finally added three years later.

The implementation is quite awesome. Nearly all of the follow-up comments here are minor refinements, but it would be great to integrate them sometime in the next month or two.  It will just further improve a solid integration of this new capability.

garrettmoon:
@appleguy Thanks for helping me get this in and for the review! I'll follow up with a PR to address the rest of the comments here.

codeOfRobin:
@garrettmoon Couldn't get it to work (image loaded from local app bundle). Any ideas?

garrettmoon:
@codeOfRobin I'd be willing to take a look if you put up a sample project!

codeOfRobin:
@garrettmoon  Here you go: https://github.com/codeOfRobin/localGIFExample

garrettmoon:
@codeOfRobin Oh yeah, I forgot about this! Sadly the GIF system isn't exercised when shouldCacheImage is true. Try adding

`loader.shouldCacheImage = false;` before `loader.url = bundleURL`

codeOfRobin:
@garrettmoon Yay! It worked. Made a PR for the documentation.

garrettmoon:
@codeOfRobin thanks for updating the docs! Wanted to propose an adjustment: https://github.com/AsyncDisplayKit/Documentation/pull/39

PR #1486, opened 2016-04-07
[ASInsetLayoutSpec] Prevent from creating a computed layout when child is nil

levi:


AsyncDisplayKit/Layout/ASInsetLayoutSpec.mm
```diff
@@ -88,6 +88,12 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
       MAX(0, constrainedSize.max.height - insetsY),
     }
   };
+  
+  if (self.child == nil) {
+    ASDisplayNodeAssert(NO, @"Inset spec measured without a child. The spec will do nothing.");
```
appleguy: Keeping an assertion is fine for now, but it will be an interesting consideration to decide if perhaps it is desirable behavior for this condition to be a permitted no-op.

Issue #1401, opened 2016-03-19
[ASVideoNode] new delegate methods and ASVideoNodePlayerState


Eke:
Hello guys,

I have application based on video content, so there is video player in every corner :smile_cat: 

The whole app except video player is made with AsyncDisplayKit and i want to change this.

We have custom video player, like ASVideoNode and it has a lot of delegate methods, for many cases which we use.

I am trying to implement such thing in `ASVideoNode`, so i forked it and made few changes listed bellow.

It will be super if i can hear some ideas from you. If my changes will be useful and it's not bad what i did, i will continue and at the will make pull request :blush:

Here is [link](https://github.com/Eke/AsyncDisplayKit/commit/d8e752f3fafd697506428199639f49469030a9be) to my `ASVideoNode` branch latest commit, which implements everything listed bellow

Added `ASVideoNodePlayerState` and two delegate methods, `videoNode:willChangePlayerState:toState:` and `videoNode:shouldChangePlayerStateTo:`

 `videoNode:willChangePlayerState:toState:`  notifies delegate that `ASVideoNode` is going to change player state.

`videoNode:shouldChangePlayerStateTo:` asks delegate if `ASVideoNode` should change state to new one. for example in some cases you can disable video pause or play.

In `ASVideoNode` class replaced : 
`[self play]` with `self.playerState = ASVideoNodePlayerStatePlaying;` 
`[self pause]` with `self.playerState = ASVideoNodePlayerStatePaused;`

Added new method `setPlayerState:` which decides must `ASVideoNode` call play or pause methods.

If this changes are ok, must remove public play and pause methods from header file. Must use `setPlayerState:` instead

@appleguy Can you take a look?

appleguy:
@Eke - sorry for the late response it's been a busy few weeks! 

This sounds like a definite improvement to the ASVideoNode API, and is something we would likely build upon at Pinterest as we are also interested in things such as a full screen playback state!

Please do proceed to a pull request. Please note, however, that we should include the play and pause methods near the bottom of the header and comment that they are deprecated. Internally, these can simply call the state change API, and we will remove them in a few months.

Eke:
Hello @appleguy, happy to hear this. I have also few more ideas how to improve `ASVideoNode` with more delegate methods, will add them too. In few days will submit pull  request.

[...]

PR #1489, opened 2016-04-07
[ASVideoNode] new delegate methods and ASVideoNodePlayerState

Eke:
details are in #1401


appleguy:
Thanks @Eke for putting up the PR! I think it is a step in the right direction.

That said, speaking with a few coworkers today, they pointed out that the scope of possible states are not all things that should be possible for a developer to set. For example, when playing is finished, this is not necessarily a state that should be available to set on the Player and I believe we will add other states in the future such as paused due to running out of video buffer.

So, for the time being I think it would make sense if we keep the play and pause methods, and then introduce the new state variable as a read only property. We're this still support most of your needs? It would still be fine to have a delegate method that can control whether or not the player can make certain state transitions, Although those state transitions would probably be initiated by a more direct method such as play or pause.

Another option is to introduce a different type of enumeration that has actions that can be set, and include play and pause in those

Eke:
Updated everything following your ideas.

Here is the final list of states i am going to implement (right now using them in my app):

````
typedef enum {
    MVVideoPlayerStateUnknown,
    MVVideoPlayerStateContentLoading,
    MVVideoPlayerStateContentPlaying,
    MVVideoPlayerStateContentPaused,
    MVVideoPlayerStateSuspend,
    MVVideoPlayerStateDismissed,
    MVVideoPlayerStateError
} MVVideoPlayerState;
```` 

The idea of `setPlayerState:` was to control all state changes from one place and avoid code duplication, like calling `isStateChangeValid:` in multiple places, maybe you will have some idea about this.

File: examples/ASVideoNodeAdvanced/Sample/ViewController.m [hunk 1]
```diff
@@ -0,0 +1,58 @@
+//
+//  ViewController.m
+//  Sample
```
appleguy: Thanks for thinking about examples!  Instead of adding a new project, would you mind adding to one of the existing ones?  In fact, it would probably be best to combine both of the Video examples into a single project (perhaps with a tab bar controller to separate the table one from the regular one).  Since pod install and building is required it's much easier for community members to check out fewer, bigger sample apps.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -178,8 +181,8 @@ - (void)didLoad
 
 - (void)tapped
 {
-  if (self.delegate && [self.delegate respondsToSelector:@selector(videoNodeWasTapped:)]) {
-    [self.delegate videoNodeWasTapped:self];
+  if (_delegate && [_delegate respondsToSelector:@selector(videoNodeWasTapped:)]) {
```
appleguy: I do prefer this style, thanks :)

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -217,6 +220,21 @@ - (void)fetchData
...
       _player.muted = _muted;
     }
+    __weak __typeof(self) weakSelf = self;
+    _timeObserver = [_player addPeriodicTimeObserverForInterval:CMTimeMake(1, 1) queue:NULL usingBlock:^(CMTime time){
+      [weakSelf periodicTimeObserver:time];
+    }];
+  }
+}
+
+- (void)periodicTimeObserver:(CMTime)time {
```
appleguy: { on next line

appleguy:
@Eke I finally got back to this - gosh there were so many ASVideoNode PRs recently, and so many conflicts!  Very exciting and good though, now most are closed and things are more calm :).

@ejensen, @yunnan-wu, @maicki could you weigh in here with any thoughts on the API improvement?  Any concerns with landing this?  It looks pretty good to me, but I also don't extensively use video node.  In the worst case I think we could add an ASVideoNode+Beta.h and expose this there.  But ideally we can commit to a public API direction to allow developers to support building a play head / track that moves as the video plays.

Eke:
Hello @appleguy 

Will get back to this in few days. how do you think will it be good to make one full functional Video App example ? Live video listing and single video page with full screen button and etc?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -217,6 +220,21 @@ - (void)fetchData
       _player = [[AVPlayer alloc] initWithPlayerItem:_currentItem];
       _player.muted = _muted;
     }
+    __weak __typeof(self) weakSelf = self;
+    _timeObserver = [_player addPeriodicTimeObserverForInterval:CMTimeMake(1, 1) queue:NULL usingBlock:^(CMTime time){
```
Xinchi: This is awesome!  Is it possible to decrease the interval of the periodic time observer to like 1/100 sec?   This will help smoothen the scrubber UI update.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -373,6 +407,7 @@ - (void)play
     }
   }
   
+  self.playerState = ASVideoNodePlayerStatePlaying;
```
Xinchi: My concern here is, should the player state be set to playing in cases where the buffer is empty? Should we check the item status ready to play on the player item first?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -446,6 +493,7 @@ - (BOOL)shouldBePlaying
 
 - (void)dealloc
 {
+  _timeObserver = nil;
```
Xinchi: Should remove the observer from the player first right? 

Something like `[_player removeTimeObserver:_imeObserver];`

AsyncDisplayKit/ASVideoNode.mm
ejensen: An alternative approach is to add a delegate method that informs when the player changes. That would allow the receiver can configure any observers as they desire. `addPeriodicTimeObserverForInterval:`, `addBoundaryTimeObserverForTimes:`, etc

Eke: Hello @Xinchi, we can move it to delegate as @ejensen mentioned. How do you think would it be better to create whole observer from delegate or just asking for preferred seconds?

Xinchi: I would prefer to move them to delegate like @ejensen has mentioned, because this way we have the flexibility to register any observer as we desire.

AsyncDisplayKit/ASVideoNode.mm
Xinchi: Btw please rebase from master first before making further changes to this PR. I think this diff is missing some key changes that has landed on master.

AsyncDisplayKit/ASVideoNode.mm
Eke: But we will lose ability to control time in ASVideoNode with this rote, no? or am i missing something?  maybe it will be better to keep it in ASVideoNode for future? for example if we will implement default scrubber UI in ASVideoNode we will need periodic time observer

hannahmbanana:
@appleguy: @Eke and I discussed closing this PR on slack this morning and starting over given the many conflicts that have arisen since the framework is evolving so fast. 

Could you review the strategy and give feedback on whether these look like good API changes?

appleguy:
@Eke: I am closing this PR with the understanding that you are going to post another.

Eke:
@appleguy will send new PR tomorrow, messed up with this one : )) there are a lot of conflicts so it will be faster to make new one, i have local copy of my changes so it will not take too long

PR #1495, opened 2016-04-08
[ASTableView] Fix crash if called updateCurrentRangeWithMode: on an ASTableNode

maicki:
A crash currently occurs if called updateCurrentRangeWithMode: on an ASTableNode. The problem is that the ASTableView within the ASTableNode does not expose the rangeController, but the ASTableNode demands it within updateCurrentRangeWithMode:.


appleguy:
@maicki thanks! I would be interested to hear why the compiler was not able to catch this issue, Is the call point incorrectly casting or forcing the method call?

PR #1501, opened 2016-04-10
[ASControlNode] modify hitTestDebug tool to highlight edges clipped by one or more superviews

hannahmbanana:


AsyncDisplayKit/ASControlNode.mm
```diff
@@ -492,7 +493,60 @@ - (void)layout
...
+      }
+      if (CGRectGetMaxY(originalRect) != CGRectGetMaxY(finalRect)) {
+        clippedEdges |= UIRectEdgeBottom;
+      }
+      if (CGRectGetMaxX(originalRect) != CGRectGetMaxX(finalRect)) {
+        clippedEdges |= UIRectEdgeRight;
+      }
+      
+      const CGFloat borderWidth = 2.0;
+      const UIColor *borderColor = [UIColor colorWithRed:30/255.0 green:90/255.0 blue:50/255.0 alpha:0.7];
```
appleguy: Wondering what this looks like. Could you upload an example screenshot?

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -492,7 +493,60 @@ - (void)layout
       intersectSuperlayer = intersectLayer.superlayer;
     }
     
-    _debugHighlightOverlay.frame = [intersectLayer convertRect:intersectRect toLayer:layer];
+    CGRect finalRect = [intersectLayer convertRect:intersectRect toLayer:layer];
+    UIColor *fillColor = [[UIColor greenColor] colorWithAlphaComponent:0.4];
+  
+    // determine which edges were clipped
```
appleguy: This is cool, but the code is pretty generically applicable and probably should not live in control mode in the long term.

Could you think about a way to generalize it and look around for a file that might be appropriate to contain this kind of functionality, or perhaps we can create an image utility file where we could also implement a simple rounded rectangle stretchable image  generator?

hannahmbanana:
E.g. `_likesNode` (cyan backgroundColor) contains `_likesIcon` which has a `target:action:` pair added (green debugOverlay backgroundColor).  `_likesNode` has `hitTestSlop` extending past superNode's `hitTestSlop` on top, left and bottom. To visualize the fact that the tapable area of the `_likesIcon` is restricted by it's superview's tapable area, there is now a dark green border on the 3 clipped sides. 

hittestdebugclippingfortargetaction

appleguy:
Very cool, this is definitely an important improvement as a developer might otherwise keep trying to increase the slop area and be frustrated when there is no visual change. 

 Thinking about this more, I believe there is another problem however. It may be the case that one of the super views has clips to bounds enabled, which even though it will not affect touch delivery and if the hit test slop is applied in a way that can route events correctly,  it would physically prevent the debug label from being seen as its full-size. This could be a case where the when a clip side is not visible when it should be. 

 I am not sure what the best way is to visually represent the situation, because it is a fairly complex subtlety. Could you think about an adjustment to the implementation that will at least change the color of the hard clip edges do something more obvious, like a bright or clashing color, and perhaps users will work at the header comments  for a clue about what the unusual coloring is indicating?

WildDylan:
Where's a demo for use the hit slop....

hannahmbanana:
@WildDylan - there isn't a demo, but you can find more here:

*Hit Test Slop Visualization*: http://asyncdisplaykit.org/docs/debug-tool-hit-test-visualization.html
*How to use Hit Test Slop*: http://asyncdisplaykit.org/docs/hit-test-slop.html

WildDylan:
@hannahmbanana  thanks

PR #1504, opened 2016-04-10
[ASControlNode] Improved hitTestDebug to account for limiting display area by parent .clipsToBounds

hannahmbanana:
- added orange overlay edge highlighting if tapable area clipped by .clipsToBounds of any parent in the hierarchy
- fixed bug in comparison of child / parent rect edges
- updated +Debug.h header to explain colors
hittestdebugclipstoboundssuper

### DIFF
--- a/AsyncDisplayKit/ASControlNode.mm
+++ b/AsyncDisplayKit/ASControlNode.mm
@@ -517,19 +517,21 @@ void _ASEnumerateControlEventsIncludedInMaskWithBlock(ASControlNodeEvent mask, v
     CGRect finalRect = [intersectLayer convertRect:intersectRect toLayer:layer];
     UIColor *fillColor = [[UIColor greenColor] colorWithAlphaComponent:0.4];
   
-    // determine which edges were clipped
-    if (clippedEdges != UIRectEdgeNone) {
+    // determine if edges are clipped
+    if (clippedEdges == UIRectEdgeNone) {
+      _debugHighlightOverlay.backgroundColor = fillColor;
+    } else {
       const CGFloat borderWidth = 2.0;
-      UIColor *superhitTestSlopClipsBorderColor = [UIColor colorWithRed:30/255.0 green:90/255.0 blue:50/255.0 alpha:0.7];
-      UIColor *superClipsToBoundsBorderColor = [[UIColor orangeColor] colorWithAlphaComponent:0.8];
+      UIColor *borderColor = [[UIColor orangeColor] colorWithAlphaComponent:0.8];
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASControlNode.mm
```diff
@@ -479,14 +480,20 @@ - (void)layout
...
-      if (parentNode) {
+      if (parentNode && !parentClipsToBounds) {
         parentHitRect = UIEdgeInsetsInsetRect(parentHitRect, [parentNode hitTestSlop]);
       }
       
       // Convert our current rectangle to parent coordinates, and intersect with the parent's hit rect.
       CGRect intersectRectInParentCoordinates = [intersectSuperlayer convertRect:intersectRect fromLayer:intersectLayer];
       intersectRect = CGRectIntersection(parentHitRect, intersectRectInParentCoordinates);
+      if (parentClipsToBounds) {
+        if (!CGSizeEqualToSize(parentHitRect.size, intersectRectInParentCoordinates.size)) {
```
appleguy: Could you collapse these two conditions? They don't read that easily to me.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -499,41 +506,19 @@ - (void)layout
...
-      if (originalRect.origin.x != finalRect.origin.x) {
-        clippedEdges |= UIRectEdgeLeft;
-      }
-      if (CGRectGetMaxY(originalRect) != CGRectGetMaxY(finalRect)) {
-        clippedEdges |= UIRectEdgeBottom;
-      }
-      if (CGRectGetMaxX(originalRect) != CGRectGetMaxX(finalRect)) {
-        clippedEdges |= UIRectEdgeRight;
-      }
+      clippedEdges = [self setEdgesOfIntersectionForChildRect:originalRect parentRect:finalRect rectEdge:clippedEdges];
```
appleguy: After reading the code of both, I think it would be best to move this up to be in the same condition. You could check for the on equal size first, then run this, and then only if the parent does clip to bounds, run the check for that.

Although technically this would be less efficient and result in more method calls, the method will run very quickly and I think the code will be more readable. Because this code is a bit complex, The readability benefit is substantial.

With this change, I believe you can also eliminate the variable originalRect, and then the check for whether you should draw the edges becomes clippedEdges != UIRectEdgeNone.

appleguy:
Thanks for working on this edge case! It is definitely a subtle one as pointed out in your last change, but certainly a very important one for developers depending on the accuracy of these overlays.

PR #1511, opened 2016-04-11
[ASLayoutSpec] Improvements for ASLayoutSpec's internal implementation of child / children properties.

maicki:
- Added nullable for returning children
- Store default child and children as properties
- Use vector for creating final children

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayoutSpec.h
+++ b/AsyncDisplayKit/Layout/ASLayoutSpec.h
@@ -25,6 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)init;
 
+/**
+ * Parent of the layout spec
+ */
 @property (nullable, nonatomic, weak) id<ASLayoutable> parent;
 
 /**
@@ -97,8 +100,10 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable id<ASLayoutable>)childForIdentifier:(NSString *)identifier;
 
-/** Returns all children added to this layout spec. */
-- (NSArray<id<ASLayoutable>> *)children;
+/**
+ * Returns all children added to this layout spec.
+ */
+- (nullable NSArray<id<ASLayoutable>> *)children;
 
 @end
 
--- a/AsyncDisplayKit/Layout/ASLayoutSpec.mm
+++ b/AsyncDisplayKit/Layout/ASLayoutSpec.mm
@@ -18,24 +18,23 @@
 #import "ASLayout.h"
 #import "ASThread.h"
 
-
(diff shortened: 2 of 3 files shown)


AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -109,46 +108,53 @@ - (void)setParent:(id<ASLayoutable>)parent
...
-  if ([finalLayoutable isKindOfClass:[ASLayoutSpec class]]) {
-    [(ASLayoutSpec *)finalLayoutable setParent:self]; // This will trigger upward propogation if needed.
-  } else if ([self supportsUpwardPropagation]) {
-    ASEnvironmentStatePropagateUp(self, finalLayoutable.environmentState.layoutOptionsState); // Probably an ASDisplayNode
-  }
+  self.childrenWithIdentifier[identifier] = finalLayoutable;
+  
+  // TODO: Should we propagate up the layoutable at it could happen that multiple children will propagated up their
+  //       layout options and one child will overwrite values from another child
+  // [self propagateUpLayoutable:finalLayoutable];
```
maicki: @appleguy Should we propagate up the layoutable at it could happen that multiple children will propagated up their layout options and one child will overwrite values from another child?

AsyncDisplayKit/Layout/ASLayoutSpec.h
```diff
@@ -97,8 +100,10 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable id<ASLayoutable>)childForIdentifier:(NSString *)identifier;
 
-/** Returns all children added to this layout spec. */
-- (NSArray<id<ASLayoutable>> *)children;
+/**
+ * Returns all children added to this layout spec.
+ */
+- (nullable NSArray<id<ASLayoutable>> *)children;
```
maicki: @appleguy I added the nullable annotation in this case as:
1. If no children where assigned we will return nil in this case
2. We nil out the children's array in the setChildren: method now before checking if children where provided and only than assign a newly created array

Any thoughts in general about it?

AsyncDisplayKit/Layout/ASLayoutSpec.mm
appleguy: @maicki good question, we will definitely revisit this in more detail once we aim to re-enable propagation.

I do think in some cases like a vertical stack of two text nodes with flexShrink = YES, we may even want to propagate up from multiple chilrden.  But before we do this stuff we need to go back to the whiteboard and really think carefully about the exact situations it makes sense, and then add them to documentation.

appleguy:
@maicki some nice stuff here!!

PR #1521, opened 2016-04-12
Add Profile build configuration to ASDK

levi:


appleguy:
Thanks!  This will also be useful for automatic performance tests in the future.

PR #1524, opened 2016-04-13
[ASDisplayNodeExtras] Correct nullability annotations

ejensen:


AsyncDisplayKit/ASDisplayNodeExtras.h
```diff
@@ -102,12 +102,12 @@ extern void ASDisplayNodePerformBlockOnEverySubnode(ASDisplayNode *node, void(^b
...
  Given a display node, traverses up the layer tree hierarchy, returning the first display node that passes block.
  */
-extern id _Nullable ASDisplayNodeFindFirstSupernode(ASDisplayNode * _Nullable node, BOOL (^block)(ASDisplayNode *node));
+extern ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernode(ASDisplayNode * _Nullable node, BOOL (^block)(ASDisplayNode *node));
 
 /**
  Given a display node, traverses up the layer tree hierarchy, returning the first display node of kind class.
  */
-extern id _Nullable ASDisplayNodeFindFirstSupernodeOfClass(ASDisplayNode *start, Class c);
+extern __kindof ASDisplayNode * _Nullable ASDisplayNodeFindFirstSupernodeOfClass(ASDisplayNode *start, Class c);
```
appleguy: Not familiar with __kindof, actually - why is this different than the above method return?

ejensen: Here's a short read covering the gist of `__kindof`: http://www.russbishop.net/objective-c-generics It's more flexible than an explicit class type, but safer than an `id`.

appleguy:
Hmm, interesting.  I wonder where else this would be useful.  Silly world we live in with modern Objective-C.

Thanks for your prolific dilligence, @ejensen !  It is _sincerely_ appreciated!!

PR #1525, opened 2016-04-13
Add caching respondsToSelector calls in ASCollectionView and ASTableView

maicki:


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -333,16 +349,26 @@ - (void)setAsyncDataSource:(id<ASCollectionViewDataSource>)asyncDataSource
   if (asyncDataSource == nil) {
     _asyncDataSource = nil;
     _proxyDataSource = _isDeallocating ? nil : [[ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
-    _asyncDataSourceImplementsConstrainedSizeForNode = NO;
-    _asyncDataSourceImplementsNodeBlockForItemAtIndexPath = NO;
+    
+    _asyncDataSourceFlags.asyncDataSourceConstrainedSizeForNode = NO;
```
appleguy: It would be great to take advantage of the fact that these have been cleanly stored in a single flags variable, and find a way to zero all of them together. Ideally, this would work even if a new flag is added later, Maybe by making use of sizeof().  Not exactly sure how to do this.  Would setting _asyncDataSourceFlags = 0ULL do the trick?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -333,16 +349,26 @@ - (void)setAsyncDataSource:(id<ASCollectionViewDataSource>)asyncDataSource
...
   } else {
     _asyncDataSource = asyncDataSource;
     _proxyDataSource = [[ASCollectionViewProxy alloc] initWithTarget:_asyncDataSource interceptor:self];
-    _asyncDataSourceImplementsConstrainedSizeForNode = [_asyncDataSource respondsToSelector:@selector(collectionView:constrainedSizeForNodeAtIndexPath:)];
-    _asyncDataSourceImplementsNodeBlockForItemAtIndexPath = [_asyncDataSource respondsToSelector:@selector(collectionView:nodeBlockForItemAtIndexPath:)];
+    
+    _asyncDataSourceFlags.asyncDataSourceConstrainedSizeForNode = [_asyncDataSource respondsToSelector:@selector(collectionView:constrainedSizeForNodeAtIndexPath:)];
+    _asyncDataSourceFlags.asyncDataSourceNodeBlockForItemAtIndexPath = [_asyncDataSource respondsToSelector:@selector(collectionView:nodeBlockForItemAtIndexPath:)];
+    _asyncDataSourceFlags.asyncDataSourceNumberOfSectionsInCollectionView = [_asyncDataSource respondsToSelector:@selector(numberOfSectionsInCollectionView:)];
+    _asyncDataSourceFlags.asyncDataSourceCollectionViewLockDataSource = [_asyncDataSource respondsToSelector:@selector(collectionViewLockDataSource:)];
```
appleguy: Technically, these methods are deprecated (or intended to be). It is probably good time to remove them from the implementation, is there anywhere important that we check if the data source is locked?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -333,16 +349,26 @@ - (void)setAsyncDataSource:(id<ASCollectionViewDataSource>)asyncDataSource
...
+    _asyncDataSourceFlags.asyncDataSourceCollectionViewUnlockDataSource = [_asyncDataSource respondsToSelector:@selector(collectionViewUnlockDataSource:)];
+    _asyncDataSourceFlags.asyncDataSourceCollectionViewConstrainedSizeForNodeAtIndexPath = [_asyncDataSource respondsToSelector:@selector(collectionView:constrainedSizeForNodeAtIndexPath:)];;
 
     // Data-source must implement collectionView:nodeForItemAtIndexPath: or collectionView:nodeBlockForItemAtIndexPath:
-    ASDisplayNodeAssertTrue(_asyncDataSourceImplementsNodeBlockForItemAtIndexPath || [_asyncDataSource respondsToSelector:@selector(collectionView:nodeForItemAtIndexPath:)]);
+    ASDisplayNodeAssertTrue(_asyncDataSourceFlags.asyncDataSourceConstrainedSizeForNode || _asyncDataSourceFlags.asyncDataSourceNodeBlockForItemAtIndexPath);
```
appleguy: !!! Error here - this should not check constrainedSize for node, but nodeforitem or nodeblockforitem.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -363,11 +389,25 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
...
+    _asyncDelegateFlags.asyncDelegateShouldBatchFetchForCollectionView = NO;
   } else {
     _asyncDelegate = asyncDelegate;
     _proxyDelegate = [[ASCollectionViewProxy alloc] initWithTarget:_asyncDelegate interceptor:self];
-    _asyncDelegateImplementsScrollviewDidScroll = ([_asyncDelegate respondsToSelector:@selector(scrollViewDidScroll:)] ? 1 : 0);
+    
+    _asyncDelegateFlags.asyncDelegateScrollViewDidScroll = [_asyncDelegate respondsToSelector:@selector(scrollViewDidScroll:)];
+    _asyncDelegateFlags.asyncDelegateScrollViewWillEndDraggingWithVelocityTargetContentOffset = [_asyncDelegate respondsToSelector:@selector(scrollViewWillEndDragging:withVelocity:targetContentOffset:)];
+    _asyncDelegateFlags.asyncDelegateCollectionViewWillDisplayNodeForItemAtIndexPath = [_asyncDelegate respondsToSelector:@selector(collectionView:willDisplayNodeForItemAtIndexPath:)];
+    _asyncDelegateFlags.asyncDelegateCollectionViewDidEndDisplayingNodeForItemAtIndexPathDeprecated = [_asyncDelegate respondsToSelector:@selector(collectionView:didEndDisplayingNodeForItemAtIndexPath:)];
```
appleguy: The deprecated and not-deprecated appear to have an identical selector?  collectionView:didEndDisplayingNodeForItemAtIndexPath

I think one is supposed to pass the node.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1002,8 +1042,8 @@ - (void)rangeController:(ASRangeController *)rangeController didInsertNodes:(NSA
     return; // if the asyncDataSource has become invalid while we are processing, ignore this request to avoid crashes
   }
   
+  [_layoutFacilitator collectionViewWillEditCellsAtIndexPaths:indexPaths batched:_performingBatchUpdates];
   if (_performingBatchUpdates) {
-    [_layoutFacilitator collectionViewWillEditCellsAtIndexPaths:indexPaths batched:YES];
```
appleguy: I think you need to remove the line in the else clause -- it looks like we will call batched:NO twice

appleguy:
@maicki very nice improvements here, I greatly prefer the style.

I apologize it took me a while to review, but I am eager to merge this change in the next day or two once these small revisions are made!

AsyncDisplayKit/ASCollectionView.mm
maicki: Yes we actually lock the data source in the ASDataController if we try to access the data source synchronously. Look into accessDataSourceSynchronously:withBlock in ASDataController. This is used in _reloadDataWithAnimationOptions:synchronously:completion

maicki:
@appleguy Addressed all the issues:

- Use memset to clear flags
- Fix check for asyncDataSourceNodeForItemAtIndexPath
- Fix selector for asyncDelegateCollectionViewDidEndDisplayingNodeForItemAtIndexPath
- Fix multiple collectionViewWillEditCellsAtIndexPaths:batched: calls

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -333,16 +350,22 @@ - (void)setAsyncDataSource:(id<ASCollectionViewDataSource>)asyncDataSource
   if (asyncDataSource == nil) {
     _asyncDataSource = nil;
     _proxyDataSource = _isDeallocating ? nil : [[ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
-    _asyncDataSourceImplementsConstrainedSizeForNode = NO;
-    _asyncDataSourceImplementsNodeBlockForItemAtIndexPath = NO;
+    
+    memset(&_asyncDataSourceFlags, 0, sizeof(_asyncDataSourceFlags));
```
appleguy: This is quite perfect, I think we should use this style in other areas as well when we discover applicable use cases!

appleguy:
Great change!  @zintus is also excited about the possibility of Sharing more code between table and collection, perhaps we can work together in the next month to continue progress on re-factoring of this sort :)

zintus:
@appleguy happy to see job being done on this problem 👍

PR #1526, opened 2016-04-13
[ASImageNode] Improvements to progressive decoding and cache accesses.

Adlai-Holler:
- In Network/Multiplex image nodes, update the downloader's progress image block both when visibility changes and when a download has started. This resolves the case where the node becomes visible while checking the cache and no progress renders would occur.
- Resolve a couple unintentional retains in image node fetch completion blocks 37c7f6f
- Update the PINRemoteImageDownloader to use the new cache fetching method.

There's plenty more stuff I'd like to do with image nodes but I'm bandwidth-constrained. Here's what I'm looking at:
- Add support for Photos framework URLs to ASNetworkImageNode. Would obviously need to factor that fetching code out from where it currently lives in ASMultiplexImageNode.
- Currently MultiplexImageNode does not perform the synchronous cache check, although NetworkImageNode does. Multiplex tries to synchronously get an image from its data source, though. Now that the cache is more robust we may want to ditch the data source-cache overlap.
- The ASPINRemoteImageDownloader sends nil from its fetchFromCache method, because we used to start the download synchronously anyway, and we wanted to avoid checking the cache twice in the miss case. Now, however, we start the download asynchronously, so this potential cache fast-path may be worth reopening. Ideally, if we make the cache method actually do the multi-layer cache check, we would add a PINRemoteImageDownloadOption to skip both the in-memory and disk-cache checks since presumably we just did both and got nothing.
- MultiplexImageNode currently cancels all inflight downloads whenever you call `fetchData`, and it calls `fetchData` during `displayWillStart` so you often end up canceling in-progress downloads.


appleguy:
@Adlai-Holler Thanks!  I'd be happy to merge this if you can fix the conflict.

@garrettmoon, could you review this and take mental note of the other issues mentioned?  If you have enough context to tackle any of them easily, I'm sure it would be appreciated - although I know we won't invest much work in Multiplex as we don't use it at Pinterest and there are few clients.  You might want to sync with Adlai about how Photos.framework can be most easily supported in Network without adding much complexity.

AsyncDisplayKit/ASMultiplexImageNode.mm
```diff
@@ -293,9 +294,8 @@ - (void)displayWillStart
 - (void)visibilityDidChange:(BOOL)isVisible
 {
   [super visibilityDidChange:isVisible];
-  
+  ASDN::MutexLocker l(_downloadIdentifierLock);
```
garrettmoon: Does this need to be moved up? Is the idea to make sure visibilityDidChange occurs in its entirety? Otherwise this just means the lock is held longer since it's also locked in _updateProgressImageBlockOnDownloaderIfNeeded.

AsyncDisplayKit/ASMultiplexImageNode.mm
```diff
@@ -478,6 +450,37 @@ - (UIImage *)_bestImmediatelyAvailableImageFromDataSource:(id *)imageIdentifierO
 }
 
 #pragma mark -
+
+- (void)_updateProgressImageBlockOnDownloaderIfNeeded
+{
+    // Read our interface state so that we don't lock super while holding our lock.
+    ASInterfaceState interfaceState = self.interfaceState;
```
garrettmoon: If this is called while the _downloadIdentifierLock is held, won't that result in the issue you're commenting about? It's called with the lock held in visibilityDidChange:?

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -211,9 +212,9 @@ - (void)displayWillStart
 - (void)visibilityDidChange:(BOOL)isVisible
 {
   [super visibilityDidChange:isVisible];
-  
+
+  ASDN::MutexLocker l(_lock);
```
garrettmoon: Same question here, this lock will be held throughout the call to _updateProgress… yes?

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -36,9 +36,9 @@ - (UIImage *)synchronouslyFetchedCachedImageWithURL:(NSURL *)URL
   return result.image;
 }
 
-- (void)fetchCachedImageWithURL:(NSURL *)URL
-                  callbackQueue:(dispatch_queue_t)callbackQueue
-                     completion:(void (^)(CGImageRef imageFromCache))completion
+- (void)cachedImageWithURL:(NSURL *)URL
```
garrettmoon: Ha, I wrote the new protocol and forgot to adopt it, good catch!

appleguy:
@Adlai-Holler although you should not Rush if you're busy, you make sure to send me a text message when Mrs. updated and can be emerged, so that I don't accidentally merge something else and introduce another conflict.

Unfortunately the user interface of this site does not

AsyncDisplayKit/ASMultiplexImageNode.mm
Adlai-Holler: Ah good catch!

Adlai-Holler:
@garrettmoon @appleguy Thanks for the feedback guys! I think this one's ready for love. I tested it pretty thorough with Tripstr.

AsyncDisplayKit/ASMultiplexImageNode.mm
```diff
@@ -22,6 +22,7 @@
 #import "ASPhotosFrameworkImageRequest.h"
 #import "ASEqualityHelpers.h"
 #import "ASInternalHelpers.h"
+#import "ASDisplayNodeExtras.h"
```
appleguy: What change makes this necessary, @Adlai-Holler ?

Adlai-Holler: The use of `ASInterfaceStateIncludesVisible`

Issue #1334, opened 2016-03-07
[ASTextNode] App freezes when calling trailingRect


ChrisSchneider:
Hey there,

in AsyncDisplayKit 1.9.6 or master calling trailingRect on a ASTextNode freezes the app.
The same Code works with ADK 1.9.2

Thanks a lot for your help!

Sample Code:
```
final class TestNode: ASDisplayNode {
    
    let textNode: ASTextNode = {
        let node = ASTextNode()
        node.attributedString = NSAttributedString(string: "Foooo")
        return node
    }()
    
    override func layout() {
        super.layout()
        
        addSubnode(textNode)
        textNode.frame = bounds
        
        print("IN")
        print(textNode.trailingRect())
        print("OUT")
    }
    
}
```

Backtrace:
```
* thread #1: tid = 0x21cf9e, 0x0000000105bbeee2 libsystem_kernel.dylib`__psynch_mutexwait + 10, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x0000000105bbeee2 libsystem_kernel.dylib`__psynch_mutexwait + 10
    frame #1: 0x0000000105b86a49 libsystem_pthread.dylib`_pthread_mutex_lock_wait + 89
    frame #2: 0x0000000101b1c09f libc++.1.dylib`std::__1::mutex::lock() + 9
    frame #3: 0x000000010186aacc TextNodeBugTest`-[ASTextKitContext performBlockWithLockedTextKitComponents:] [inlined] std::__1::lock_guard<std::__1::mutex>::lock_guard(this=0x00007fff5e4a8580, __m=0x00007ff648e14348) + 11 at __mutex_base:83
    frame #4: 0x000000010186aac1 TextNodeBugTest`-[ASTextKitContext performBlockWithLockedTextKitComponents:] [inlined] std::__1::lock_guard<std::__1::mutex>::lock_guard(this=0x00007fff5e4a8580, __m=0x00007ff648e14348) + 12 at __mutex_base:83
[...]

PR #1527, opened 2016-04-13
[ASTextNode] Fix deadlock calling trailingRect

maicki:
Fixes issue: #1334

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer+Positioning.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer+Positioning.mm
@@ -22,128 +22,136 @@ static const CGFloat ASTextKitRendererTextCapHeightPadding = 1.3;
 
 @implementation ASTextKitRenderer (Tracking)
 
-- (NSArray *)rectsForTextRange:(NSRange)textRange
-                 measureOption:(ASTextKitRendererMeasureOption)measureOption
+- (NSArray *)rectsForTextRange:(NSRange)textRange measureOption:(ASTextKitRendererMeasureOption)measureOption
 {
-  __block NSArray *textRects = @[];
+  __block NSArray *textRects = nil;
   [self.context performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
-    NSRange clampedRange = NSIntersectionRange(textRange, NSMakeRange(0, [textStorage length]));
-    if (clampedRange.location == NSNotFound || clampedRange.length == 0) {
-      return;
-    }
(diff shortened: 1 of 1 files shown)


garrettmoon:
Nice! Fixes the issue and likely even more performant :)

maicki:
@rcancro @garrettmoon Can you guys take another look over this please

AsyncDisplayKit/TextKit/ASTextKitRenderer+Positioning.mm
```diff
@@ -333,22 +333,30 @@ - (void)enumerateTextIndexesAtPosition:(CGPoint)externalPosition usingBlock:(as_
...
     }
-
-    // Take everything after our final character as trailing space.
-    NSArray *finalRects = [self rectsForTextRange:NSMakeRange([textStorage length] - 1, 1) measureOption:ASTextKitRendererMeasureOptionLineHeight];
-    CGRect finalGlyphRect = [[finalRects lastObject] CGRectValue];
-    CGPoint origin = CGPointMake(CGRectGetMaxX(finalGlyphRect), CGRectGetMinY(finalGlyphRect));
-    CGSize size = CGSizeMake(calculatedSize.width - origin.x, calculatedSize.height - origin.y);
-    trailingRect = (CGRect){origin, size};
+      
+    textRange = NSMakeRange([textStorage length] - 1, 1);
```
garrettmoon: Now that I think about it more, I don't think it's safe to unlock between getting the textRange and calling rectsForTextRange:measureOption:

maicki:
@garrettmoon @rcancro Good for review again. I created a helper function now that does not lock and this can be used if already in a locked state.

garrettmoon:
I think this is the correct solution all though it sucks to have an unlocked and locked version of the method. rectsForTextRange should still behave exactly the same way as before right @maicki?

Looks good to me!

rcancro:
The github diff makes it look like a lot change, but from what I can tell you just moved things around, right? Can you name the unlocked method something really scary sounding so people don't misuse it?

maicki:
@garrettmoon rectsForTextRange works the same way as before now, it just calls the unlocked version of the method.

@rcancro Currently the unlocked version is not exposed and can just be used within the method were it is defined. I hope the comment at the top makes it clear or is enough that developers don't use it without locking.

appleguy:
@rcancro I do agree.  @maicki since this code has changed a lot since I last was familiar with it, it's pretty difficult to understand the change here because of intermixing of pretty complex code segments.  I'm not sure there would have been a way with less change, but particularly for super-core code like this, it might be the one exception to the otherwise-actively-encouraged idea of making cleanup changes as part of all diffs; for locking correctness we should probably keep it to a minimized change, then a different one for improved formatting.

Because @garrettmoon also took a look and I do really trust @maicki 's quality of work + testing, I'm definitely comfortable landing this despite getting a bit lost in the review.

This is a very important fix and definitely a tough, little-known area of the code base.  Thank you @maicki for jumping in to fix it!!

PR #1529, opened 2016-04-13
[Build] ASVideoNode should be in the iOS framework too.

garrettmoon:


appleguy:
@garrettmoon  sorry, I was a bit slow and accepting this, now ready to merge but a conflict has come up. If you could update this I will make sure to merge it promptly.

appleguy:
@garrettmoon thanks for finding this!  I rolled it into https://github.com/facebook/AsyncDisplayKit/pull/1545 .

PR #1530, opened 2016-04-14
[ASDKgram example update] overrode -fetchData in PhotoCellNode & addressed PR comments

hannahmbanana:
- overrode the -(void)fetchData method in PhotoCellNode.m to download the photo’s comments. This method gets called with the PhotoCellNode enters ASInterfaceStateFetchData, which is set by the rangeController.
    - UIKIT COMPARISON: I left the comment bulk download (for all photos in a page load) in the PhotoFeedViewController side for UIKit because when implemented in the PhotoTableViewCell, each cell jumped around as it changed size when it came on screen.
- minor appearance updates
    - updated color scheme
    - fixed status bar style to darkBackgroundColor
- cleaned up layoutSpecThatFits: in PhotoCellNode

### DIFF
--- a/examples/ASDKgram/Sample/PhotoCellNode.m
+++ b/examples/ASDKgram/Sample/PhotoCellNode.m
@@ -74,17 +74,9 @@
       }
     }];
     
-    _photoTimeIntervalSincePostLabel                  = [[ASTextNode alloc] init];
-    _photoTimeIntervalSincePostLabel.layerBacked      = YES;
-    _photoTimeIntervalSincePostLabel.attributedString = [photo uploadDateAttributedStringWithFontSize:FONT_SIZE];
-    
-    _photoLikesLabel                        = [[ASTextNode alloc] init];
-    _photoLikesLabel.layerBacked            = YES;
-    _photoLikesLabel.attributedString       = [photo likesAttributedStringWithFontSize:FONT_SIZE];
-    
-    _photoDescriptionLabel                  = [[ASTextNode alloc] init];
-    _photoDescriptionLabel.layerBacked      = YES;
-    _photoDescriptionLabel.attributedString = [photo descriptionAttributedStringWithFontSize:FONT_SIZE];
+    _photoTimeIntervalSincePostLabel = [self createLayerBackedTextNodeWithString:[photo uploadDateAttributedStringWithFontSize:FONT_SIZE]];
+    _photoLikesLabel                 = [self createLayerBackedTextNodeWithString:[photo likesAttributedStringWithFontSize:FONT_SIZE]];
+    _photoDescriptionLabel           = [self createLayerBackedTextNodeWithString:[photo descriptionAttributedStringWithFontSize:FONT_SIZE]];
     _photoDescriptionLabel.maximumNumberOfLines = 3;
     
     _photoCommentsView = [[CommentsNode alloc] init];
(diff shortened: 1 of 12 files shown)


examples/ASDKgram/Sample/PhotoCellNode.m
```diff
@@ -125,56 +116,65 @@ - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
...
   return verticalStack;
 }
 
 #pragma mark - Instance Methods
 
+- (void)fetchData
+{
+  [super fetchData];
+  
+  [_photoModel.commentFeed refreshFeedWithCompletionBlock:^(NSArray *newComments) {
```
appleguy: This is a very cool example of using fetch data!

It is probably worth adding a comment here, either linking to the documentation about intelligent preloading and overriding methods like this, or at least a call out to why this is simpler and also much more efficient then what can be easily implemented with UIKit.

examples/ASDKgram/Sample/PhotoFeedNodeController.m
```diff
@@ -93,30 +93,30 @@ - (void)loadPageWithContext:(ASBatchContext *)context
   [_photoFeed requestPageWithCompletionBlock:^(NSArray *newPhotos){
     
     [self insertNewRowsInTableView:newPhotos];
-    [self requestCommentsForPhotos:newPhotos];
+//    [self requestCommentsForPhotos:newPhotos];
```
appleguy: We should also comment at these two locations to indicate that this is not an ideal implementation, and highlight the fact that it is not trivial to do better here. In this case, this is a one-shot call for all of the photos that have just loaded, which can just the network and slows down image  loading . The only easy alternative is to perform the load when  A cell becomes visible, which results in poor user perceive performance of  comment loading  and can cause jumpy and flashy transitions as the comments always change the height of the cells once they are visible.

examples/ASDKgram/Sample/PhotoFeedNodeController.m
```diff
@@ -93,30 +93,30 @@ - (void)loadPageWithContext:(ASBatchContext *)context
...
-      
-      if (cell) {
-        [cell loadCommentsForPhoto:photo];
-        [_tableNode.view beginUpdates];
-        [_tableNode.view endUpdates];
-      }
-    }];
-  }
-}
+//- (void)requestCommentsForPhotos:(NSArray *)newPhotos
```
appleguy: Actually, I mean that the comment is required in the Apple framework version of this file. This code should be deleted from this file as it should no longer be necessary with the more efficient handling of this at the cell level.

examples/ASDKgram/Sample/Utilities.m
```diff
@@ -14,12 +14,12 @@ @implementation UIColor (Additions)
...
 + (UIColor *)darkBlueColor
 {
-  return [UIColor colorWithRed:18.0/255.0 green:86.0/255.0 blue:136.0/255.0 alpha:1.0];
+  return [UIColor colorWithRed:70.0/255.0 green:102.0/255.0 blue:118.0/255.0 alpha:1.0];
 }
 
 + (UIColor *)lightBlueColor
 {
-  return [UIColor colorWithRed:0.0 green:122.0/255.0 blue:1.0 alpha:1.0];
+  return [UIColor colorWithRed:70.0/255.0 green:165.0/255.0 blue:196.0/255.0 alpha:1.0];
```
appleguy: Probably should update the name of these methods to reflect the colors they now are present, or potential he rename them to be more generic such as primary color and secondary text color

appleguy:
@hannahmbanana  Nice refinements overall, thank you! We should definitely delete the comment loading code that is no longer needed, but because it has no impact on the operation and it looks like this is one step towards long-term improvement of this example, this change is good to merge now.

PR #1534, opened 2016-04-15
[ASMapNode] Defer snapshots triggered before layout occurs to reduce unnecessary snapshots.

hannahmbanana:


hannahmbanana:
@maicki @aaronschubert0 - could you help review? 

`.setAnnotations` in the `init:` of an ASCellNode was triggering takeSnapshot before layout. 

screen shot 2016-04-14 at 10 20 45 pm

screen shot 2016-04-14 at 10 20 58 pm

aaronschubert0:
@hannahmbanana Thanks for getting involved with ASMapNode. I appreciate this PR but I don't think this is the way this should be solved, I've put some comments on this PR to specify why.

This assert is usually called if you haven't specified a layout to ASMapNode correctly, usually I would wrap it in a `ASRatioLayoutSpec` or set the `preferredFrameSize` property. What UI are you trying to construct, and maybe we can solve this a different way?

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -159,6 +161,15 @@ - (void)setRegion:(MKCoordinateRegion)region
 
 - (void)takeSnapshot
 {
+  // do not trigger snapshot before layout
+  ASLayout *layout = self.calculatedLayout;
+  if (layout == nil || CGSizeEqualToSize(CGSizeZero, layout.size)) {
```
aaronschubert0: This would mean that ASMapNode would only ever take one snapshot, since layout will never nil again after the first pass.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -307,6 +318,15 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
   return size;
 }
 
+- (void)calculatedLayoutDidChange
+{
+  [super calculatedLayoutDidChange];
+  
+  if (_snapshotNeeded) {
+    [self takeSnapshot];
```
aaronschubert0: This also degrades the functionality of ASMapNode to only take a snapshot on a layout change. Whereas we may want to take a new snapshot without a layout change.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -14,11 +14,13 @@
 #import <AsyncDisplayKit/ASCenterLayoutSpec.h>
 #import <AsyncDisplayKit/ASThread.h>
 #import <AsyncDisplayKit/ASInternalHelpers.h>
+#import <AsyncDisplayKit/ASLayout.h>
 
 @interface ASMapNode()
 {
   ASDN::RecursiveMutex _propertyLock;
   MKMapSnapshotter *_snapshotter;
+  BOOL _snapshotNeeded;
```
aaronschubert0: I'd rather not introduce another ivar to handle this, it introduces managed state, and I'd rather ASMapNode had constraints where it could figure out if it needed a new snapshot without manually having to set a flag.

AsyncDisplayKit/ASMapNode.mm
hannahmbanana: Once the layout is set, whether or not it changes, this condition should always proceed into the snapshot taking process, therefore it shouldn't have any impact except for when the layout is nil or zero sized.

AsyncDisplayKit/ASMapNode.mm
hannahmbanana: Can you explain why that is the case? 

As far as I can tell, this would only increase the frequency with which the snapshot is taken. The base class implementation is empty, so the largest impact that this method could have is additional calls to snapshot. 

As far as I can tell, there is no case where this would be able to prevent a `takeSnapshot` call that would have occurred previously.

AsyncDisplayKit/ASMapNode.mm
hannahmbanana: This seems like a good suggestion. Could you recommend a way to replace this? I am not very familiar with this class, but I assume this might involve if the snapshotter is loading or if self.image is set? How should these be combined and is it relevant to check `isLiveMap`?

hannahmbanana:
@aaronschubert0: thanks for reviewing this!

I believe I am using the ASMapNode API correctly; `.setAnnotations` should be okay to call before the layout pass. If I move the `.setAnnotations` code to `didLoad` everything works smoothly.  However, the former implementation seems like it should be a valid use of the API.

```
@implementation MapCellNode
{
  ASMapNode *_mapNode;
  CLLocationCoordinate2D _coordinates;
}

- (instancetype)initWithCoordinates:(CLLocationCoordinate2D)coordinates
{
  self = [super init];
  
  if (self) {
    self.usesImplicitHierarchyManagement = YES;
    
    _coordinates                = coordinates;
    _mapNode                    = [[ASMapNode alloc] init];

    // set map span and region
    MKCoordinateSpan span       = MKCoordinateSpanMake(5, 5);
    MKCoordinateRegion region   = MKCoordinateRegionMake(coordinates, span);
    _mapNode.region             = region;
    
    // add annotation for coordinate
    MKPointAnnotation *photoLocationAnnotation = [[MKPointAnnotation alloc] init];
    photoLocationAnnotation.coordinate         = _coordinates;
    [_mapNode setAnnotations:@[photoLocationAnnotation]];
  }
  
  return self;
}

{
  _mapNode.preferredFrameSize = constrainedSize.max;
  
  return [ASStaticLayoutSpec staticLayoutSpecWithChildren:@[_mapNode]];
}

@end
```

aaronschubert0:
@hannahmbanana In that case, why don't we wrap `setAnnotations:` with `self.nodeLoaded`:
```
if (self.nodeLoaded) {
 if (self.isLiveMap) {
    [_mapView removeAnnotations:_mapView.annotations];
    [_mapView addAnnotations:annotations];
  } else {
    [self takeSnapshot];
  }
}
```

So that the annotations are set, but it won't trigger a new layout pass unless it's been loaded or maybe we should actually change that condition to `self.calculatedLayout` that way we can be certain this behaviour won't occur.

AsyncDisplayKit/ASMapNode.mm
garrettmoon: @hannahmbanana this is a bit confusing to read. If my understanding is correct, it'd be nice to have a comment which says, set _snapshotNeeded to YES so if layout changes in the future, we'll try snapshotting again.

appleguy: @aaronschubert0    Let me know if Hannah's description doesn't address your concern. This approach seems to make sense to me, or at least I am not sure what a better route would be to addressing the very real crash reported here.

@garrettmoon   Do you think it would be more clear with a different instance variable name? Perhaps mentioning something about a deferred snapshot?

I agree, the comment could definitely be improved to describe that the condition is critical to avoid crashes, although it doesn't seem that opaque to me -- from the perspective that, at least as a performance optimization, a zero sized snapshot won't do any good.  The level of criticality is less clear though.

AsyncDisplayKit/ASMapNode.mm
appleguy: @aaronschubert0 also curious -- I am moderately familiar with this code, but want to make sure I'm not overlooking a problem here.

It does seem to me like this could only increase the number of cases where a snapshot is taken.  Unless your comment on this line is actually referring to the lines above that gate inside the -takeSnapshot method, in which case let's focus on that thread.

AsyncDisplayKit/ASMapNode.mm
garrettmoon: Yeah, _snapshotAfterLayout perhaps?

hannahmbanana: @garrettmoon: Thanks for the suggestion! I've renamed the variable.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -159,6 +161,15 @@ - (void)setRegion:(MKCoordinateRegion)region
 
 - (void)takeSnapshot
 {
+  // set _snapshotNeeded to YES so if layout changes in the future, we'll try snapshotting again.
```
appleguy: Update ivar name in comment; it should also mention something like "if we are zero-sized, ..."

aaronschubert0:
@hannahmbanana This is looking good to me! Just FYI seems like some documentation files and changes to ASTextNode have leaked into this PR.

hannahmbanana:
@aaronschubert0 - thanks for catching this! I removed the _site files. 

The modifications to ASTextNode were intentional. While looking for `calculatedLayoutDidChange` in `ASDisplayNode+Subclasses.h`, I noticed that the header file was inconsistent and missing a few requires super flags. This exposed the fact that `ASTextNode` was missing a call to super.

appleguy:
@hannahmbanana thanks for fixing the missing super call in ASTextNode!

PR #1536, opened 2016-04-15
[ASTextNode] Add locking around NSTextContainer's constrainedSize

garrettmoon:
Summary: Looks like this isn't being locked when it should be.

Reviewers: ricky, levi, scottg, schneider

Reviewed By: scottg, schneider

Subscribers: jenkins

Differential Revision: https://phabricator.pinadmin.com/D84952

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitContext.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitContext.mm
@@ -8,16 +8,15 @@
  *
  */
 
-#import <mutex>
-
 #import "ASTextKitContext.h"
+#import "ASThread.h"
 
 #import "ASLayoutManager.h"
 
 @implementation ASTextKitContext
 {
   // All TextKit operations (even non-mutative ones) must be executed serially.
-  std::mutex _textKitMutex;
+  ASDN::Mutex _textKitMutex;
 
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/TextKit/ASTextKitContext.mm
```diff
@@ -36,8 +35,8 @@ - (instancetype)initWithAttributedString:(NSAttributedString *)attributedString
 {
   if (self = [super init]) {
     // Concurrently initialising TextKit components crashes (rdar://18448377) so we use a global lock.
-    static std::mutex __static_mutex;
-    std::lock_guard<std::mutex> l(__static_mutex);
+    static ASDN::Mutex __staticMutex;
+    ASDN::MutexLocker l(__staticMutex);
```
nguyenhuy: Not a blocking issue but in the future we may want to switch to [`ASDN::StaticMutex`](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASThread.h#L167). I believe [ASLayoutable](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASLayoutable.mm#L51) is the only place it is used as of now, although this thing has existed since the legendary "Initial commit" :P

PR #1537, opened 2016-04-15
Fix ASTableNode / ASCollectionNode backgroundColor does not apply correctly.

maicki:
If the background color is applied via the pending state it's applied to the layer of the UICollectionView / UITableview. Unfortunately UITableView / UICollectionView does not consider using the layer backgroundColor property as it's background color, so it needs to be applied to the view after the ASCollectionNode / ASTableNode did load and the view is available

### DIFF
--- a/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
+++ b/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
@@ -19,6 +19,7 @@
 #import "ASPendingStateController.h"
 #import "ASThread.h"
 #import "ASTextNode.h"
+#import "ASTableNode.h"
 
 /**
  * The following macros are conveniences to help in the common tasks related to the bridging that ASDisplayNode does to UIView and CALayer.
@@ -239,11 +240,11 @@ if (shouldApply) { _layer.layerProperty = (layerValueExpr); } else { ASDisplayNo
 
   // For classes like ASTableNode, ASCollectionNode, ASScrollNode and similar - make sure UIView gets setFrame:
   struct ASDisplayNodeFlags flags = _flags;
-  BOOL setFrameDirectly = flags.synchronous && !flags.layerBacked;
+  BOOL specialPropertiesHandling = ASDisplayNodeNeedsSpecialPropertiesSettingHandlingForFlags(flags);
 
   BOOL nodeLoaded = __loaded(self);
   BOOL isMainThread = ASDisplayNodeThreadIsMain();
-  if (!setFrameDirectly) {
+  if (!specialPropertiesHandling) {
     BOOL canReadProperties = isMainThread || !nodeLoaded;
     if (canReadProperties) {
       // We don't have to set frame directly, and we can read current properties.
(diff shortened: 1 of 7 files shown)


appleguy:
@maicki interesting.  So, there is actually a similar workaround for frame.  UITableView frame does not work properly if it is set only on the layer, unlike almost all other UIViews.

I think the right fix here is actually to change the property bridge, and always use -[UIView setBackgroundColor:] for a view-backed node.  Otherwise, initWithViewBlock: on certain elements like perhaps a custom table (that a developer doesn't need to be async) will have the same issue.  There could also be other UIKit classes that override setBackgroundColor: -- because it is difficult for UIKit to intercept the CALayer calls in the way our architecture allows.  Lastly, it would reduce the code change.

We would need to change the _ASPendingState to always store a UIColor as regenerating it is too costly.  This is fine as the node.backgroundColor is already of this type.  It would actually make some drawing operations faster as capturing it as a draw parameter would be quicker that regenerating.  For layer-backed we'd call CGColor.

If you'd prefer to land this as-is and file a task for the other change, that's totally fine.  I'm not certain how complex the diff would be to make, but it may be simple enough to do now.  @maicki just let me know and I will merge this now if you suggest.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -19,6 +19,7 @@
 #import "ASPendingStateController.h"
 #import "ASThread.h"
 #import "ASTextNode.h"
+#import "ASTableNode.h"
```
appleguy: Let's remove this if possible

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -583,6 +584,12 @@ - (void)setBackgroundColor:(UIColor *)newBackgroundColor
   if (shouldApply) {
     CGColorRef oldBackgroundCGColor = _layer.backgroundColor;
     _layer.backgroundColor = newBackgroundCGColor;
+    
+    BOOL specialPropertiesHandling = ASDisplayNodeNeedsSpecialPropertiesSettingHandlingForFlags(_flags);
+    if (specialPropertiesHandling) {
+        _view.backgroundColor = newBackgroundColor;
```
appleguy: I think in this case we should probably not set the backgroundColor on the layer.  Setting it on the view should always do that, unless perhaps the view class intentionally wants to do something else with it (perhaps it needs to redirect it to be set on a sublayer or such).

So let's fork the code inside shouldApply to the view and the layer case.

AsyncDisplayKit/Private/ASDisplayNodeInternal.h
```diff
@@ -25,8 +25,10 @@
 @class _ASDisplayLayer;
 @class _ASPendingState;
 @class ASSentinel;
+struct ASDisplayNodeFlags;
 
 BOOL ASDisplayNodeSubclassOverridesSelector(Class subclass, SEL selector);
+BOOL ASDisplayNodeNeedsSpecialPropertiesSettingHandlingForFlags(ASDisplayNodeFlags flags);
```
appleguy: Remove Setting from name - variable name above calls it SpecialPropertiesHandling

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -789,9 +789,16 @@ - (void)applyToView:(UIView *)view setFrameDirectly:(BOOL)setFrameDirectly
...
     view.clipsToBounds = clipsToBounds;
 
-  if (flags.setBackgroundColor)
+  if (flags.setBackgroundColor) {
+    // Set the background color to the layer as in the UIView bridge we use this value as background color
     layer.backgroundColor = backgroundColor;
 
+    // We have to make sure certain nodes get the background color call directly
+    if (specialPropertiesHandling) {
+      view.backgroundColor = [UIColor colorWithCGColor:backgroundColor];
```
appleguy: Here we should probably only set it on the view as well.

appleguy:
Thanks, this is a great fix because it is so generalized!  We can make some tweaks in a followup.

maicki:
@appleguy Here is the follow up PR: #1563

PR #1539, opened 2016-04-15
[ASDisplayNode][Refactoring] Remove internal and unused async sizing methods

nguyenhuy:
- Also add a small improvement to async transition: Bail early if a scheduled transition was invalidated before its block executes, thus save some CPU cycles that would otherwise be spent on an unnecessary measurement. Idea borrowed from `_enqueueAsyncSizingWithSentinel`.


appleguy:
Thank you @nguyenhuy !

Issue #875, opened 2015-11-20
c++11 in podspec v.s. gnu++0x in xcodeproj


yilei:
`AsyncDisplayKit.podspec` says it's `'CLANG_CXX_LANGUAGE_STANDARD' => 'c++11'`

But `AsyncDisplayKit.xcodeproj/project.pbxproj` says `CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";`

Should they be consistent with each other?

appleguy:
@yilei wow, great find - this is certainly esoteric!  :)  Thank you for filing the task.

If I'm honest, I'm not familiar with the finer points of the differences between the standards.  @chpwn @toulouse I know you are more expert in language / compiler details than me; do you have a suggestion for which ASDK should be referencing?  I'd certainly assume c++11...

toulouse:
Those environment variables directly reflect the `-std` flag in clang.

`gnu++0x` is an old name for `gnu++11`, or to be more pedantic, `c++0x` was renamed to `c++11`, and the `gnu` convention followed. There are no differences between `0x` and `11`; it's a naming thing only.

`gnu++11` is identical to `c++11 ` except with the `GNUMode` flag enabled. The `LangStandards.def` file in the clang frontend directory is an exhaustive list of what the different standards mean.

Since clang succeeded gcc in Xcode, it was built with a strong GNU compatibility mode. I find that the GNU extensions are rarely if ever noticeable in code largely dominated by Objective-C/Objective-C++. They are exhaustively documented here: https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Extensions.html#C_002b_002b-Extensions.

Without any specific requirement for GNU extensions, it would seem reasonable to leave them disabled, and besides, projects making use of ASDK via CocoaPods should already be compiling with `c++11` and not `gnu++0x`, thanks to the podspec. The lack of build failures on that account would indicate that there isn't any code relying on the GNU extensions.

Switching to `c++11` is a good idea. For that matter, even `c++14` should be pretty safe for any recent codebase, but `c++11` most assuredly is.

tl;dr: `s/gnu++0x/c++11/` :white_check_mark:

appleguy:
@yilei @toulouse - The community is planning an exciting long term road map for the project and getting organized around how to deliver these feature requests. We are moving to a new issue tracker with more sophisticated planning capabilities (check out www.realArtists.com and their new product: Ship!). 

[...]

PR #1540, opened 2016-04-15
Fix podspec vs xcodeproj inconsistency with c++ build flag

maicki:
Addresses issue: #875 

As we don't use any C++ features that need the GNU compatibility mode I changed the xcodeproj to use the c++11 flag instead of using the gnu++0x flag which had the GNU compatibility mode features enabled.


appleguy:
@maicki very cool, I'm so happy with the awesome improvements in C++ over time, and am both a bit disappointed and excited to realize that this flag is finally set properly!

PR #1543, opened 2016-04-16
[ASMapNode] Remove assertion that checks calculatedSize rather than options.size.

aaronschubert0:
@hannahmbanana @appleguy @garrettmoon 

Thanks for waiting a bit longer while I spent a deeper time looking at this issue. Apologies @hannahmbanana after re-reading your implementation another time it's clear that the issue would be fixed your way also.

However upon reproducing your initial crash (I created a sample project with your example code) and walking through the code, I realised that this assert is no longer needed and is actually a misguiding assert, YES - `self.calculatedSize` may be zero *but* this is actually ok. This is because previously we used to set `self.options.size` to `self.calculatedSize` without any check but this is now in place L:127 & L:281 so this assert is no longer needed. I'm confident that this will solve your issue but please put it to test fully and let me know of your results.

### DIFF
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -209,7 +209,6 @@
 
 - (void)setUpSnapshotter
 {
-  ASDisplayNodeAssert(!CGSizeEqualToSize(CGSizeZero, self.calculatedSize), @"self.calculatedSize can not be zero. Make sure that you are setting a preferredFrameSize or wrapping ASMapNode in a ASRatioLayoutSpec or similar.");
   _snapshotter = [[MKMapSnapshotter alloc] initWithOptions:self.options];
 }
 


appleguy:
@aaronschubert0 I'm pretty sure this was added recently, as @Adlai-Holler discovered that MapKit will internally crash if provided a zero size options.

I'm still not understanding though.  If the options has never been set to a non-zero size, and setAnnotations: is called (eg before the view is created, and before layout), we trigger a snapshot.  And it will be zero sized.

How does this address that issue?

aaronschubert0:
Hey @appleguy, I actually added that assert in when I first implemented ASMapNode and the reason was to combat options.size being set to zero. That's why I can be so certain that we can remove this. Since following the changes that @Adlai-Holler made, there are no code paths possible where options.size can be set to zero (and it's set to 256x256 from MapKit by default)

Hannah's issue wasn't that options.size was zero sized. It was that her code hit this false assert, options.size was never set to zero. If Hannah tries out this fix then I'm certain her issue will be resolved. 

I also made a sample project based upon Hannah's example code that I can post if you want to look at that before Hannah gets a chance to try out this fix.

hannahmbanana:
@aaronschubert0 @appleguy - Sorry that I am a bit behind on the conversation. I will try out this fix tonight.

hannahmbanana:
@aaronschubert0 @appleguy - If I disable the assertion as proposed here, I am able to set annotations on an `ASMapNode` in the `init:` method of an `ASCellNode` as desired.  

However, calling `setAnnotations` in `init:` means that `takeSnapshot` will always get called twice (once before and once after layout). The first snapshotter is canceled when `calculateSizeThatFits` is called:

`setAnnotations:` --> `takeSnapshot` (1st snapshotter begins)
`calculateSizeThatFits:` --> `setSnapshotSizeWithReloadIfNeeded:` 
--> `destroySnapshotter` (1st snapshotter canceled)
--> `takeSnapshot` (2nd snapshotter begins)

By checking that layout hasn't happened yet we can avoid the unnecessary 1st call. Thus, I still believe my solution is more performant.

aaronschubert0:
@hannahmbanana Fair enough, let's remove this assert and then clean up your PR. Thanks!

PR #1544, opened 2016-04-18
[AsyncDisplayKit.org] Fixed title on home page, added multiplex image node to list

lappp9:


hannahmbanana:
@lappp9 @appleguy - I don't think that we don't need to check in any _site files for the website to work, since it appears that GitHub Pages automatically generates the site from our source code. The _site folder is generated when running it locally with Jekyll. 

We should remove this from the _site folder from the current repository as well. 

Also, we have a docs folder under ASDK master (not updated in 2 years) which duplicates everything in gh-pages. Can we get rid of that @appleguy?

PR #1545, opened 2016-04-18
[ASVideoNode] Cleanup from recent diffs, standardizing on .asset rather than .url.

appleguy:


appleguy:
FYI @gazreese, @yunnan-wu, @lappp9, @ekurutepe

lappp9:
👍

PR #1546, opened 2016-04-18
[ASVideoNode] Review comments implemented for previous pull request

gazreese:
The review comments are outlined here - https://github.com/facebook/AsyncDisplayKit/pull/1323#discussion_r59999979

The only addition here is a check to avoid processing where no assets are available that could be processed.


AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -149,6 +149,11 @@ - (void)setPlaceholderImagefromAsset:(AVAsset*)asset
   ASPerformBlockOnBackgroundThread(^{
     ASDN::MutexLocker l(_videoLock);
     
+    // Skip the asset image generation if we don't have any tracks available that are capable of supporting it
```
appleguy: Assuming the comment below starting with // Unfortunately is one of the main cases where this early return is hit, could you move those comment lines up here?  I think it is fine to still have the check for if (image), though.

Also below: it looks like it might be incorrect to check _placeholderImageNode.image == nil.  What if it already has an image but now needs a new one applied to it?

Does kCMTimeZero work for the times array or is 0, 1 necessary?

appleguy:
@gazreese I re-ran the tests a couple times, but it looks like this is blocked on video-related test failures :(  https://travis-ci.org/facebook/AsyncDisplayKit/jobs/123922821

appleguy:
@gazreese all the other large ASVideoNode PRs are cleared, so feel free to update this one when you have time!

appleguy:
@gazreese ping!

gazreese:
Hey @appleguy, only got chance to look at this yesterday. It's looking good though and should be able to merge in today.

gazreese:
@appleguy There were basically two changes for me to merge here. The first main change was this:

```
    // Skip the asset image generation if we don't have any tracks available that are capable of supporting it
    NSArray<AVAssetTrack *>* visualAssetArray = [_asset tracksWithMediaCharacteristic:AVMediaCharacteristicVisual];
    if (!visualAssetArray || visualAssetArray.count == 0)
      return;
```

This seems to have already been cherry-picked into the mainline, so there was no need to merge.

The other was a change to the `-interfaceStateDidChange` method, but this method seems to have been removed so there was nothing to change.

I've merged the mainline into the branch and pushed up to my fork just so that I can continue to develop on my own project. The pull request should be identical to your master, so I think you can probably kill this PR now?

appleguy:
Funny, no sure how all that happened exactly, but it looks like we converged to correctness!  Thanks for your help @gazreese :)  Look for 1.9.73, hoping to push it tonight.

gazreese:
Sounds good!

I’m just raising a bug for a potential issue that I discovered. Doesn’t look like anything that might stop a release but thought it was worth opening the discussion.

PR #1550, opened 2016-04-18
[NSArray+Diffing] Static analyzer issue with common indexes algorithm in _asdk_commonIndexesWithArray:

maicki:
The issue was: The left operand of '>=' is a garbage value in else if check for lengths[i+1][j] ... I had to rewrite the algorithm to quiet the static analyzer.

I also added a couple of tests for _asdk_commonIndexesWithArray:

@levi @appleguy Could you guys take a look over it please.

### DIFF
--- a/AsyncDisplayKitTests/ArrayDiffingTests.m
+++ b/AsyncDisplayKitTests/ArrayDiffingTests.m
@@ -10,13 +10,58 @@
 
 #import "NSArray+Diffing.h"
 
+@interface NSArray (ArrayDiffingTests)
+- (NSIndexSet *)_asdk_commonIndexesWithArray:(NSArray *)array compareBlock:(BOOL (^)(id lhs, id rhs))comparison;
+@end
+
 @interface ArrayDiffingTests : XCTestCase
 
 @end
 
 @implementation ArrayDiffingTests
 
-- (void)testDiffing {
+- (void)testDiffingCommonIndexes
+{
+  NSArray<NSArray *> *tests = @[
+    @[
+      @[@"bob", @"alice", @"dave"],
+      @[@"bob", @"alice", @"dave", @"gary"],
+      @[@0, @1, @2]
+    ],
+    @[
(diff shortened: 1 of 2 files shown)


AsyncDisplayKitTests/ArrayDiffingTests.m
```diff
@@ -10,13 +10,58 @@
 
 #import "NSArray+Diffing.h"
 
+@interface NSArray (ArrayDiffingTests)
+- (NSIndexSet *)_asdk_commonIndexesWithArray:(NSArray *)array compareBlock:(BOOL (^)(id lhs, id rhs))comparison;
```
levi: Would recommend not writing test for `_asdk_commonIndexesWithArray:`, as it's a private method. Should only test the output of the public methods.

levi:
Looks good! Just a note about the test.

AsyncDisplayKitTests/ArrayDiffingTests.m
maicki: @levi I see your point. @appleguy any thoughts on removing the tests?

appleguy: @maicki @levi I know there are other areas where we test internal methods. I think it depends mainly on whether they have a contained set of expected behaviors, for example if they are a functional style and algorithmic like this case, I don't see a disadvantage in having tests. If the tests ever break, it will be a reminder that the behavior has changed in a material way, and if that is intended then we can simply delete the tests.

I don't feel strongly about this as a general point, and am both interested and open to the community refining are testing practices in this respect or others. Just mainly for this specific case, it seems like the tests have value at the moment, whereas we can always delete them later if they no longer have value.

appleguy:
@maicki thanks!  I had hoped this would be a simpler fix, but I appreciate you diving into it and very thoroughly addressing the issue. Great benefit to walk away with tests as well!

PR #1552, opened 2016-04-19
[ASVideoNode] Re-enables HLS support

chourobin:
According to AVFoundation [guide](https://developer.apple.com/library/ios/documentation/AudioVideo/Conceptual/AVFoundationPG/Articles/02_Playback.html) on playback, we can't use an AVAsset directly for HTTP live stream media. The latest changes unifying the API under `AVAsset` broke HLS for me.

This change: 
1. first checks if there are tracks loaded on the `AVAsset`
2. if not, creates an `AVPlayerItem` directly from URL

Let me know your thoughts on this change. Thanks!

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -89,7 +89,11 @@
   [self removePlayerItemObservers];
   
   if (_asset) {
-    _currentPlayerItem = [[AVPlayerItem alloc] initWithAsset:_asset];
+    if ([_asset.tracks count]) {
+      _currentPlayerItem = [[AVPlayerItem alloc] initWithAsset:_asset];
+    } else {
+      _currentPlayerItem = [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];
+    }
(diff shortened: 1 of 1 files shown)


appleguy:
@chourobin thanks for this, I really appreciate putting up the minimal fix for now. It seems like a good approach to me, as this issue seems more like a strange idiosyncrasy of Apple's implementation than an intentional or beneficial API design.

What is on is that in one of the test project where we use hey streaming URL, it did work after my change. Maybe the URL is not a completely rigorous test because it is not a true live stream, rather a streaming playback of recorded content.

cc @lappp9, @gazreese, @yunnan-wu, @ejensen

appleguy:
@chourobin is there any way we can include either a unit test, or a video in one of the example projects that would have broken with the previous code and succeeds with this implementation?

Especially because there is a lot of interest and activity in the video node, I want to be sure that this lesser used but very important feature is always maintained in a correct state.

Thanks again for your investigation and work here!

chourobin:
Thanks for pulling this in so quickly! I'll include some tests over the weekend 👍

PR #1555, opened 2016-04-19
Update ASSentinel to use OSAtomicAdd32 to support iOS7.0

vivqu:
OSAtomicIncrement32 does not work on OS < 7.1


appleguy:
Thank you, @vivqu !

Issue #1492, opened 2016-04-08
ASDelegateProxy - methodSignatureForSelector: called!


bogosmer:
Hi there!

We're experiencing crashes because of the methodSignatureForSelector: method is called on ASDelegateProxy. Unfortunately we haven't been able to reproduce it. We need to help it along a bit to get into a scenario that will cause that crash. From the stacktrace it looks like the delegate method that UIKit wants to call is scrollViewDidScroll:
```
Thread : Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x180a42e38 __exceptionPreprocess
1  libobjc.A.dylib                0x1800a7f80 objc_exception_throw
2  CoreFoundation                 0x180a42d80 -[NSException initWithCoder:]
3  Foundation                     0x1814061e4 -[NSProxy methodSignatureForSelector:]
4  CoreFoundation                 0x180a469e0 ___forwarding___
5  CoreFoundation                 0x180944d1c _CF_forwarding_prep_0
6  UIKit                          0x185eb5478 -[UIScrollView(UIScrollViewInternal) _notifyDidScroll]
7  UIKit                          0x185baaf84 -[UIScrollView setContentOffset:]
8  UIKit                          0x185c698e8 -[UITableView setContentOffset:]
9  UIKit                          0x185eb662c -[UIScrollView(UIScrollViewInternal) _adjustContentOffsetIfNecessary]
10 UIKit                          0x185eb0408 -[UIScrollView _updateForChangedScrollRelatedInsets]
11 UIKit                          0x185c4e91c -[UIScrollView setContentInset:]
12 UIKit                          0x185c783ac -[UITableView setContentInset:]
13 Some app                       0x1004532f0 -[SomeASTableViewSubclass adjustsContentInsetMethod] (SomeASTableViewSubclass.m:XXX)
```
The theory is that the delegate proxy returns YES to respondsToSelector:@selector(scrollViewDidScroll:) and then when forwardingTargetForSelector:@selector(scrollViewDidScroll:) is executed the _target instance variable is nil somehow. However I can't trigger it, even with stress testing. I also read about a previous issue that was related to accessibility, but turning that on changes nothing. 

We're using version 1.9.4. Since I can't reproduce it haven't tried updating to newest version. I looked in the changes and didn't seem like there were any direct changes to proxy delegate handling.

So in the hope that anyone has seen this before or got some good ideas, heres a fresh issue on the topic :)

Thanks

PR #1562, opened 2016-04-19
[ASDelegateProxy] Fix "[NSProxy methodSignatureForSelector:] called!" crash

maicki:
Currently if the _interceptor and the _target object in ASDelegateProxy are nil and an object calls a method on the ASDelegateProxy it will raise a [NSProxy methodSignatureForSelector:] exception.

The reason for that is that we are not implementing -methodSignatureForSelector that will be called in case -respondsToSelector: will return NO and -forwardingTargetForSelector: will return nil.

Unfortunately just returning nil from -methodSignatureForSelector will not help in this case we have to return some kind of NSMethodSignature. If we would return nil in -methodSignatureForSelector it would raise an selector not found exception. In order to get this object to work properly, the use of a method which creates an NSMethodSignature from a C string and just returns that if no method signature was found at _interceptor and _target. After that -forwardInvocation: will be called and we just don't do anything in there.

Should fix: #1492

### DIFF
--- a/AsyncDisplayKit/Details/ASDelegateProxy.m
+++ b/AsyncDisplayKit/Details/ASDelegateProxy.m
@@ -131,6 +131,31 @@
   }
 }
 
+- (NSMethodSignature *)methodSignatureForSelector:(SEL)aSelector
+{
+  // Check for a compiled definition for the selector
+  NSMethodSignature *methodSignature = nil;
+  if ([self interceptsSelector:aSelector]) {
+    methodSignature = [[_interceptor class] instanceMethodSignatureForSelector:aSelector];
+  } else {
+    methodSignature = [[_target class] instanceMethodSignatureForSelector:aSelector];
+  }
+  
+  // Unfortunately, in order to get this object to work properly, the use of a method which creates an NSMethodSignature
+  // from a C string. -methodSignatureForSelector is called when a compiled definition for the selector cannot be found.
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Details/ASDelegateProxy.m
```diff
@@ -131,6 +131,31 @@ - (id)forwardingTargetForSelector:(SEL)aSelector
...
+  }
+  
+  // Unfortunately, in order to get this object to work properly, the use of a method which creates an NSMethodSignature
+  // from a C string. -methodSignatureForSelector is called when a compiled definition for the selector cannot be found.
+  // This is the place where we have to create our own dud NSMethodSignature. This is necessary because if this method
+  // returns nil, a selector not found exception is raised. The string argument to -signatureWithObjCTypes: outlines
+  // the return type and arguments to the message. To return a dud NSMethodSignature, pretty much any signature will
+  // suffice. Since the -forwardInvocation call will do nothing if the delegate does not respond to the selector,
+  // the dud NSMethodSignature simply gets us around the exception.
+  return methodSignature ?: [NSMethodSignature signatureWithObjCTypes:"@^v^c"];
```
appleguy: What does this signature represent?  A void return type?  What is the c?

maicki: For the _cmd

PR #1571, opened 2016-04-20
[ASPagerNode] Fix ASPagerNode data source and delegate inconsistencies

maicki:
- Change ASPagerNodeDataSource -> ASPagerDataSource
- Add ASPagerDelegate

### DIFF
--- a/AsyncDisplayKit/ASPagerNode.h
+++ b/AsyncDisplayKit/ASPagerNode.h
@@ -11,7 +11,8 @@
 @class ASPagerNode;
 @class ASPagerFlowLayout;
 
-@protocol ASPagerNodeDataSource <NSObject>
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASPagerNode.h
```diff
@@ -65,25 +66,30 @@
 
 @end
 
+@protocol ASPagerDelegate <NSObject, ASCollectionDelegate>
```
appleguy: This doesn't need NSObject if the one above doesn't, right?

maicki: Yep, ASCollectionDelegate already extends it

maicki:
@appleguy Addressed your comments

levi:
LGTM

PR #1572, opened 2016-04-21
[ASDisplayNode] Add a block API to provide an ASLayoutSpec without having to subclass ASDisplayNode

maicki:
- Add layoutSpecBlock to ASDisplayNode
- Add assert so subclasses do not implement layoutSpecThatFits: and provide a layoutSpecBlock
- Move assert for checking if a layoutSpecBlock or one of the three layout methods are overridden. This needs to happen on runtime now as we cannot figure out if a layoutSpecBlock was provided in the class +initialize method
- Fix ASDisplayNodeImplicitHierarchyTests that already had a class with a layout spec block

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -143,6 +143,8 @@ NS_ASSUME_NONNULL_BEGIN
  * encouraged.
  *
  * @note This method should not be called directly outside of ASDisplayNode; use -measure: or -calculatedLayout instead.
+ *
+ * @warning Overwriting layoutSpecThatFits: in a subclass and providing a layoutSpecBlock block is currently not supported
  */
 - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize;
 
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -43,6 +43,11 @@ typedef void (^ASDisplayNodeDidLoadBlock)(ASDisplayNode * _Nonnull node);
 typedef void (^ASDisplayNodeContextModifier)(_Nonnull CGContextRef context);
 
 /**
+ * ASDisplayNode layout spec block. This block can be used instead of implementing layoutSpecThatFits: in subclass
+ */
+typedef ASLayoutSpec * _Nonnull(^ASLayoutSpecBlock)(ASSizeRange constrainedSize);
+
+/**
  Interface state is available on ASDisplayNode and ASViewController, and
  allows checking whether a node is in an interface situation where it is prudent to trigger certain
  actions: measurement, data fetching, display, and visibility (the latter for animations or other onscreen-only effects).
@@ -251,6 +256,17 @@ NS_ASSUME_NONNULL_BEGIN
  * @see [ASDisplayNode(Subclassing) calculateLayoutThatFits:]
  */
 - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize;
+
+
+/**
+ * @abstract Provides a way to declare a block to provide an ASLayoutSpec without having to subclass ASDisplayNode and
+ * implement layoutSpecThatFits:
+ *
+ * @return The block used to provide a ASLayoutSpec
+ *
+ * @warning Overwriting layoutSpecThatFits: in a subclass and providing a layoutSpecBlock block is currently not supported
+ */
+@property (nonatomic, readwrite, copy, nullable) ASLayoutSpecBlock layoutSpecBlock;
 
 /** 
  * @abstract Return the calculated size.
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -167,7 +167,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   return overrides;
 }
 
-     // At most a layoutSpecBlock or one of the three layout methods is overridden
+  // At most a layoutSpecBlock or one of the three layout methods is overridden
 #define __ASDisplayNodeCheckForLayoutMethodOverrides \
     ASDisplayNodeAssert(_layoutSpecBlock != nil || \
     (ASDisplayNodeSubclassOverridesSelector(self.class, @selector(calculateSizeThatFits:)) ? 1 : 0) \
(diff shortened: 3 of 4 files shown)


AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -143,6 +143,8 @@ NS_ASSUME_NONNULL_BEGIN
  * encouraged.
  *
  * @note This method should not be called directly outside of ASDisplayNode; use -measure: or -calculatedLayout instead.
+ *
+ * @warning Overwriting layoutSpecThatFits: in a subclass and providing a layoutSpecBlock block is currently not supported
```
appleguy: Overriding*.  You might say: "Subclasses that implement -layoutSpecThatFits: must not also use .layoutSpecBlock.  Doing so will trigger an exception.  A future version of the framework may support using both, calling them serially, with the .layoutSpecBlock superseding any values set by the method override."

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -252,6 +257,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize;
 
+
+/**
+ * @abstract Provides a way to declare a block to provide an ASLayoutSpec without having to subclass ASDisplayNode and
+ * implement layoutSpecThatFits:
+ *
+ * @return The block used to provide a ASLayoutSpec
+ *
+ * @warning Overwriting layoutSpecThatFits: in a subclass and providing a layoutSpecBlock block is currently not supported
```
appleguy: Same thing.  For the @return, maybe "A block that takes a constrainedSize ASSizeRange argument, and must return an ASLayoutSpec that includes all of the subnodes to position in the layout.  This input-output relationship is identical to the subclass override method -layoutSpecThatFits:"

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -178,12 +186,6 @@ + (void)initialize
     ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(measureWithSizeRange:)), @"Subclass %@ must not override measureWithSizeRange method", NSStringFromClass(self));
     ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(recursivelyClearContents)), @"Subclass %@ must not override recursivelyClearContents method", NSStringFromClass(self));
     ASDisplayNodeAssert(!ASDisplayNodeSubclassOverridesSelector(self, @selector(recursivelyClearFetchedData)), @"Subclass %@ must not override recursivelyClearFetchedData method", NSStringFromClass(self));
-
-    // At most one of the three layout methods is overridden
-    ASDisplayNodeAssert((ASDisplayNodeSubclassOverridesSelector(self, @selector(calculateSizeThatFits:)) ? 1 : 0)
```
appleguy: I think we should also put __ASDisplayNodeCheckForLayoutMethodOverrides here.  This way it can trigger upon application startup / first usage even if the node doesn't go through a layout process.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1876,13 +1880,22 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
...
 }
 
 - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
 {
+  __ASDisplayNodeCheckForLayoutMethodOverrides;
+
   ASDN::MutexLocker l(_propertyLock);
+  
+  if (_layoutSpecBlock != nil) {
+    return _layoutSpecBlock(constrainedSize);
```
appleguy: Very cool, extremely clean!  I think you should compare the block against NULL rather than nil.  It's probably fine to just do if (_layoutSpecBlock), but != NULL is fine too (there are some style benefits either way so it's your choice)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1904,6 +1917,14 @@ - (ASSizeRange)constrainedSizeForCalculatedLayout
   return _constrainedSize;
 }
 
+- (void)setLayoutSpecThatFitsBlock:(ASLayoutSpecBlock)layoutSpecBlock
```
appleguy: I think this needs to be called setLayoutSpecBlock:

AsyncDisplayKitTests/ASDisplayNodeImplicitHierarchyTests.m
```diff
@@ -109,8 +102,9 @@ - (void)testCalculatedLayoutHierarchyTransitions
   ASDisplayNode *node3 = [[ASDisplayNode alloc] init];
   
   ASSpecTestDisplayNode *node = [[ASSpecTestDisplayNode alloc] init];
-  node.layoutSpecBlock = ^(ASSizeRange constrainedSize, NSNumber *layoutState){
-    if ([layoutState isEqualToNumber:@1]) {
+  __weak ASSpecTestDisplayNode *weakNode = node;
+  node.layoutSpecBlock = ^(ASSizeRange constrainedSize){
+    if ([weakNode.layoutState isEqualToNumber:@1]) {
```
appleguy: Hmm, this raises an interesting question.  Should the block take an argument of the node it is laying out?  Because self is not always available like it normally is in layoutSpecThatFits, and there is a substantial risk of retain cycles if a developer ever makes a small mistake and uses an instance variable inside one...

Maybe the header documentation should include a warning about the risk of retain cycles...

appleguy:
@maicki very exciting to integrate this feature!

We should do a follow-up here to improve the comments a little bit, and to think very carefully about the risk of retained cycles and how we might be able to reduce that risk. I think including the node as an argument, although not strictly necessary if the user correctly uses a weak pointer, maybe a good idea to encourage the pattern of accessing through that object and as a reminder to consider memory management. We might even call the variable for the argument something like "weakNode" or "weakSelf" to remind people that it is important to use a weak relationship there.

AsyncDisplayKit/ASDisplayNode.mm
maicki: The problem having __ASDisplayNodeCheckForLayoutMethodOverrides in here if the subclass does not override one of the three methods as the consumer always want's to use it via a block. So we would assert in here although it's not really wrong to do it that way. Any thoughts?

maicki:
I created a follow up PR that should address your comments: #1583

PR #1577, opened 2016-04-21
[ASCellNode] Repair Layout Delegate Behavior When Triggered Off-Main

### LINKED ISSUE #1574: [ASCellNode] didRelayoutFromOldSize:toNewSize: Doesn't Use Correct Values When Layed Out Off-Main
Looks like the sweet auto-table-row-height update behavior may have suffered a regression =/. It used to work very well, but in 1.9.7.2 it's stopped working.

To reproduce on current master dd4853b, use the Kittens sample app. Replace `-[KittenNode toggleImageEnlargement]` implementation:

```objectivec
- (void)toggleImageEnlargement
{
  _isImageEnlarged = !_isImageEnlarged;
  [self setNeedsLayout];
}
```

with

```objectivec
- (void)toggleImageEnlargement
{
  _isImageEnlarged = !_isImageEnlarged;
  dispatch_async(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
    [self setNeedsLayout];
  });
}
```

Within 1-3 taps I'll usually see the cell fail to update its height to match the new layout. Instead of calling `- didRelayoutFromOldSize:toNewSize:` just once, with the old and new values, the cell node will call it twice, first with the old value twice, then with the new value twice, so `sizeChanged` is never determined as true.

Adlai-Holler:
Resolves #1574. `setNeedsLayout` is just a figurehead – the real magic happens in `__setNeedsLayout` so that's where the cell needs to involve itself.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -34,6 +34,10 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 
 /**
  * Generic cell node.  Subclass this instead of `ASDisplayNode` to use with `ASTableView` and `ASCollectionView`.
+ 
+ * @note When a cell node is contained inside a collection view (or table view),
+ * calling `-setNeedsLayout` will also notify the collection on the main thread
+ * so that the collection can update its item layout if the cell's size changed.
  */
 @interface ASCellNode : ASDisplayNode
 
@@ -84,17 +88,6 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 - (void)touchesMoved:(NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 - (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 - (void)touchesCancelled:(nullable NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
-
-/**
- * Marks the node as needing layout. Convenience for use whether the view / layer is loaded or not.
- *
- * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
- * and the supernode is notified or (if this node is the root one) a full measurement pass is executed using the old constrained size.
- * The delegate will then be notified on main thread.
- *
- * This method can be called inside of an animation block (to animate all of the layout changes).
- */
-- (void)setNeedsLayout;
 
 /**
  * @abstract Initializes a cell with a given view controller block.
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -10,6 +10,7 @@
 
 #import "ASInternalHelpers.h"
 #import "ASEqualityHelpers.h"
+#import "ASDisplayNodeInternal.h"
 #import <AsyncDisplayKit/_ASDisplayView.h>
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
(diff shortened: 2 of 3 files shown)


AsyncDisplayKit/ASCellNode.mm
```diff
@@ -120,10 +121,10 @@ - (void)setLayerBacked:(BOOL)layerBacked
   ASDisplayNodeAssert(!layerBacked, @"ASCellNode does not support layer-backing.");
 }
 
-- (void)setNeedsLayout
+- (void)__setNeedsLayout
```
appleguy: Is this intended to be overriding the ASDisplayNode base class method?  Seems fine, just curious how this operates at runtime / why it is essential.

Adlai-Holler: Yeah it's intended. Consider, if they call `-setNeedsDisplay` off-main, then basically nothing will happen during `[super setNeedsLayout]` so `self.calculatedSize` won't change. Then at some point in the future, when the base class implementation actually does the remeasure, the cell node won't be in the loop anymore.

Adlai-Holler: Actually, I just found this `calculatedSizeDidChange` override point. Let me play with that for a minute and see if we can do better.

Adlai-Holler: Hrm that may be a no-go since we don't get the old value in there.

AsyncDisplayKit/ASCellNode.h
```diff
@@ -86,17 +90,6 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 - (void)touchesCancelled:(nullable NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
- * Marks the node as needing layout. Convenience for use whether the view / layer is loaded or not.
- *
- * If this node was measured, calling this method triggers an internal relayout: the calculated layout is invalidated,
```
appleguy: We lost some of the detail here around re-measuring.  This description wasn't great, so we should rewrite it and yours is closer, but: it's not currently clear that -setNeedsLayout can cause the cell to /change size/, as that is not normally true for any type of UIView or UIKit cell.  Then on top of that, but only once you understand that triggering does a re-measure pass and it may result in a different size, there is the detail that the table or collection will be notified of any size change.

It might be worth mentioning that a table's cell node will be given the fixed width with a min height of 0 and Infinity to measure itself, and collections get 0, 0 and Inf, Inf (unless the constrainedSizeFor... method is overridden).  Just not sure what the best way is to emphasize these critical differences to users.

cc @Adlai-Holler, @hannahmbanana, @lappp9

AsyncDisplayKit/ASCellNode.mm
appleguy: There is also calculated layout to change, although I'm not entirely sure what the situation is in this change. Is this related to what people have been discussing on the open-source slack channel?

Adlai-Holler: Ah yeah that's the one I meant. Not sure if that's what's being discussed, but in theory it would be easy to modify `calculatedLayoutDidChange` (and its antecedent `_completeLayoutCalculation`) to accept the previous layout as an argument, and then have ASCellNode use that to drive its layout delegate calling, where `sizeChanged <= (self.calculatedSize != oldLayout.size)`.

appleguy:
@Adlai-Holler I think there is a chance that this is related to, or the root issue may have been caused by, https://github.com/facebook/AsyncDisplayKit/pull/1578

Would you mind re-testing master with that fix applied and make sure this one is still necessary?

Adlai-Holler:
@appleguy I gave it a try and it's still busted on master =/. This PR is rebased and ready to go though.

appleguy:
@Adlai-Holler thanks for your efforts here even when you're so short on time!

@maicki could you file a followup / cleanup task, in Ship, Priority 2 for 2.0, and link to this?

PR #1581, opened 2016-04-22
[ASNetworkImageNode] API to notify when each render of progressive image is complete.

wendylu:


AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -246,6 +280,9 @@ - (void)visibilityDidChange:(BOOL)isVisible
           }
           
           strongSelf.image = progressImage;
+          ASPerformBlockOnMainThread(^{
+            strongSelf->_currentImageQuality = progress;
```
appleguy: This and other trampolined blocks I believe are un-locked accesses; is that permissible?  It would have some overhead to grab the lock of course, but it may be necessary to go through the property setter.  I wonder if there is any deadlock risk there.

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -38,17 +38,19 @@ @interface ASNetworkImageNode ()
   id _downloadIdentifier;
 
   BOOL _imageLoaded;
-  
+  CGFloat _currentImageQuality;
```
appleguy: Could you add a block comment, maybe here or in the displayDidFinish, describing the overall management strategy - particularly the subtleties around the thread trampoline?

appleguy:
This is a very cool addition; thanks @wendylu!  This will allow developers to track highly detailed image loading progress for any network image in their application, including pJPEG, which is just fantastic.  Later we could add stuff around network / download performance even for images that are offscreen at the time they finish downloading.

appleguy:
Hmm, we need to figure out how to make sure the example apps are getting the latest PINRemoteImage...@garrettmoon, any ideas?

/Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m:177:44: error: property 'renderedImageQuality' not found on object of type 'PINRemoteImageManagerResult *'
        progressBlock(result.image, result.renderedImageQuality, result.UUID);

wendylu:
@appleguy edited the title because we don't actually render at each scan, but rather it depends on the image downloader how often they want to pass back a new image (for PINRemoteImage this is when we are at 0, 35% and 65% download progress, and it does not have to be at a scan boundary)

Adlai-Holler:
Woot this is awesome @wendylu , would it be too much trouble to add to ASMultiplexImageNode as well?

appleguy:
@wendylu I had to take another diff that went up before this one, and touched the same files - unfortunately it created some conflicts.  I'm not expecting much other work landing here, so rebase when you're ready.

@garrettmoon, without more than half a day or so, is it possible to create a PINRemoteImage release branch with the necessary support, while hiding or +Beta'ing the API changes you'd like to hold back?  I wasn't aware that the GIF support in ASDK isn't usable without engineers knowing about the special branch to point to.  If this isn't feasible, I suppose we need to comment above the APIs in ASDK / rename the category AnimatedImageBeta, and a comment pointing people to the branch would also be reasonable.

garrettmoon:
@appleguy they can use the special branch, a commit hash or master for GIF support. Compilation was working fine with or without it when my commit landed, I'll have to look into what's changed.

PR #1583, opened 2016-04-22
[ASDisplayNode] Add layout spec block follow up

maicki:
Followup for #1572

### DIFF
--- a/AsyncDisplayKitTests/ASDisplayNodeImplicitHierarchyTests.m
+++ b/AsyncDisplayKitTests/ASDisplayNodeImplicitHierarchyTests.m
@@ -76,7 +76,7 @@
   ASDisplayNode *node5 = [[ASDisplayNode alloc] init];
 
   ASSpecTestDisplayNode *node = [[ASSpecTestDisplayNode alloc] init];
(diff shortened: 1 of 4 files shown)


AsyncDisplayKitTests/ASDisplayNodeImplicitHierarchyTests.m
```diff
@@ -102,9 +102,9 @@ - (void)testCalculatedLayoutHierarchyTransitions
   ASDisplayNode *node3 = [[ASDisplayNode alloc] init];
   
   ASSpecTestDisplayNode *node = [[ASSpecTestDisplayNode alloc] init];
-  __weak ASSpecTestDisplayNode *weakNode = node;
-  node.layoutSpecBlock = ^(ASSizeRange constrainedSize){
-    if ([weakNode.layoutState isEqualToNumber:@1]) {
+  node.layoutSpecBlock = ^(ASDisplayNode *weakNode, ASSizeRange constrainedSize){
+    ASSpecTestDisplayNode *strongNode = (ASSpecTestDisplayNode *)weakNode;
```
appleguy: Because the pointer is not /actually/ weak, we don't need to do this, right?

Issue #1588, opened 2016-04-25
[Question] about the website documentation on using ASCollectionNode inside ASViewController


nickvelloff:
In http://asyncdisplaykit.org/docs/containers-ascollectionnode.html#replace-uicollectionviewcontroller-with-asviewcontroller

The following can't be correct (as I'm learning this system)t:
<img width="686" alt="screenshot 2016-04-25 02 34 58" src="https://cloud.githubusercontent.com/assets/885369/14779730/6579d020-0a8e-11e6-8b36-a1dd60ee624c.png">

hannahmbanana:
@nickvelloff: thanks for reviewing our new documentation! It is still a work in progress, so expect a few hiccups. :)

The sentence before the code looks correct. I could probably make it more clear that `initWithNode:` is the ASViewController's designated initializer. This is the strategy for using an ASCollectionNode with an ASViewController, but it is also possible to use an ASCollectionNode without managing it with an ASViewController. 

Is it an issue with the Swift code? I am not familiar with Swift syntax. cc: @lappp9

appleguy:
@nickvelloff - happy to help, could you clarify your question?  The designated initializer for ASVC is initWithNode - the code you have looks largely correct, but I'm not sure how to call that initializer in Swift.  Wouldn't it be super.initWithNode(node: collectionNode) ?

cc @nguyenhuy, if you want to check my Swift here.

If you're not on our Slack yet, please join and we'd love to collaborate with you live!  Just email me at asyncdisplaykit@gmail.com - to request an invite.

maicki:
@appleguy It's should be this way:

```swift
init(coordinates: CLLocationsCoordinate2D) {
        let flowLayout     = UICollectionViewFlowLayout()
        let collectionNode = ASCollectionNode(collectionViewLayout: flowLayout)
        
        super.init(node: collectionNode)
        
        flowLayout.minimumInteritemSpacing  = 1
        flowLayout.minimumLineSpacing       = 1
}
```

lappp9:
@maicki Isn't the first version what the docs has?

And yeah, the initializer should be `super.init(node: whateverNode)`  

Which is actually pretty cool, considering the objc version is `[super initWithNode:whateverNode]`

lappp9:
Oh! The `func initWithCoordinates(...)` part is wrong! My bad, i'll update it real quick.

appleguy:
[...]

PR #1589, opened 2016-04-25
Quick update to the ASCollectionNode docs

lappp9:
Fixes swift initializer and rewords things a little bit.


maicki:
Should address issue: #1588

PR #1590, opened 2016-04-25
[ASNetworkImage] Fix deadlock in ASNetworkImageNode

maicki:
Calling _updateProgressImageBlockOnDownloaderIfNeeded should be called without _lock held. We will lock super to read our interface state and it's best to avoid acquiring both locks.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -217,7 +217,7 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
   [super visibilityDidChange:isVisible];
 
   if (_downloaderImplementsSetPriority) {
-    ASDN::MutexLocker l(_lock);
+    _lock.lock();
     if (_downloadIdentifier != nil) {
       if (isVisible) {
         [_downloader setPriority:ASImageDownloaderPriorityVisible withDownloadIdentifier:_downloadIdentifier];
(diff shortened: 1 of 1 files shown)


maicki:
@Adlai-Holler Can you take a looker over it. Thanks!

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -217,16 +217,18 @@ - (void)visibilityDidChange:(BOOL)isVisible
...
-    ASDN::MutexLocker l(_lock);
+    _lock.lock();
     if (_downloadIdentifier != nil) {
       if (isVisible) {
         [_downloader setPriority:ASImageDownloaderPriorityVisible withDownloadIdentifier:_downloadIdentifier];
       } else {
         [_downloader setPriority:ASImageDownloaderPriorityPreload withDownloadIdentifier:_downloadIdentifier];
       }
     }
+    _lock.unlock();
```
garrettmoon: @maicki Just so I understand what's going on, can you help correct me: It was my impression that the lock would unlock at this point automatically due to its destructor and it falling out of scope.

maicki: @garrettmoon This could be the case. I went just the safe route and lock and unlock it manually to also make it a bit more clear as we call `_updateProgressImageBlockOnDownloaderIfNeeded` afterwards that should not be called with a lock

appleguy: Yes this would have unlocked as written but this change is fine.

Adlai-Holler:
Oops! Sorry for introducing this bug – my app only exercises ASMultiplexImageNode. Thanks @maicki for the fix!

PR #1592, opened 2016-04-26
[ASViewController] Add support for ASTraitCollection

rcancro:
Initial attempt to get display traits working with ASEnvironment.

To get proper ASDisplayTraits support, you must use an ASViewController. The ASViewController implements UITraitCollection-related methods (`traitCollectionDidChange:`, `willTransitionToTraitCollection:withTransitionCoordinator:`, `viewWillTransitionToSize:withTransitionCoordinator`) to update the internal ASDisplayTraits and propagate them to subnodes.

ASTableNode and ASCollectionNode don't actually have their cells as subnodes, so a little bit of trickery is involved (on `setEnvironment:` the table/collection node gets its data controller's completedNodes and propagates the new traits. see `ASDisplayTraitsCollectionTableSetEnvironmentState`). The data controller also has the ability to get the current display traits when creating new cells through a new protocol `ASDataControllerEnvironmentDelegate`.

ASViewController also supports the ability to return a custom set of display traits regardless its `UITraitCollection`. For example, if you have a modal dialog that should always have a compact size classes, you can set the override block before displaying the VC (`overrideDisplayTraitsWithTraitCollection` and `overrideDisplayTraitsWithWindowSize`).

A new example, called Display Traits, has been added. It shows how display traits can be used in a ASViewController with a normal ASDisplayNode as its root, as well as in ASViewControllers hosting a table node or collection node. There is also an example of overriding the default display traits of a VC.

Please provide any suggestions that you have!


AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -93,6 +128,33 @@ ASDISPLAYNODE_EXTERN_C_END
...
+// ASCollection/TableNodes don't actually have ASCellNodes as subnodes. Because of this we can't rely on display trait
+// downward propagation via ASEnvironment. Instead if the new environmentState has displayTraits that are different from
+// the cells', then we propagate downward explicitly and request a relayout.
+//
+// If there is any new downward propagating state, it should be added to this define.
+//
+// This logic is used in both ASCollectionNode and ASTableNode
+#define ASDisplayTraitsCollectionTableSetEnvironmentState \
+- (void)setEnvironmentState:(ASEnvironmentState)environmentState\
+{\
```
maicki: Do we need locking for setting and propagation?

examples/DisplayTraits/Sample/CollectionViewController.m
```diff
@@ -0,0 +1,72 @@
...
+  cell.textNode.maximumNumberOfLines = 3;
+  cell.imageTappedBlock = ^{
+    [KittenNode defaultImageTappedAction:self];
+  };
+  return cell;
+}
+
+- (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
+{
+  ASDisplayTraits displayTraits = self.collectionNode.environmentState.displayTraits;
```
maicki: Maybe we should provide a better API to get to the displayTraits so the consumer does not need to have to know (and I think should not know) about the environmentState

cc @appleguy

examples/DisplayTraits/Sample/KittenNode.m
```diff
@@ -0,0 +1,170 @@
...
+  
+  overrideVC.overrideDisplayTraitsWithTraitCollection = ^(UITraitCollection *traitCollection) {
+    return (ASDisplayTraits) {
+      .displayScale = traitCollection.displayScale,
+      .horizontalSizeClass = UIUserInterfaceSizeClassCompact,
+      .userInterfaceIdiom = traitCollection.userInterfaceIdiom,
+      .verticalSizeClass = UIUserInterfaceSizeClassCompact,
+      .forceTouchCapability = traitCollection.forceTouchCapability,
+    };
+  };
```
maicki: Same for here. Seems pretty difficult for API consumer that are not so familiar with structs or so to use the API. Why about creating an object that basically is a representation for a ASDisplayTraits struct and that will be returned in that block. In ASDK if you call the block we use that object and create a ASDisplayTrait for it. Thoughts?

cc @appleguy

AsyncDisplayKit/Private/ASEnvironmentInternal.mm
```diff
@@ -15,11 +15,17 @@
...
+BOOL ASEnvironmentStateUpwardPropagationEnabled()
 {
-  return AS_SUPPORT_PROPAGATION;
+  return AS_SUPPORT_UPWARD_PROPAGATION;
+}
+
+BOOL ASEnvironmentStateDownwardPropagationEnabled()
+{
+  return AS_SUPPORT_DOWNWARD_PROPAGATION;
 }
```
maicki: As we are currently prevent certain state to propagate should we be more discrete what we support for now? Something like -supportsTraitsCollectionPropagation and ASEnvironmentStateTraitCollectionPropagationEnabled? Maybe we can add that to special header file that if we finally decide to support upward and downward propagation for all states we can just remove it?

maicki:
@rcancro  Overall seems good to me. Just one thing I'm thinking about is that the naming of the ASDisplayTraits is a bit different as all the other states that we start with "ASEnvironment..." Just thinking if we should internally use ASEnvironmentDisplayTraitsState and for the public API we expose it as displayTraits as we already do?

 cc @appleguy

rcancro:
@maicki Thanks for the review! I've addressed most of your comments except for the request to make a NSObject-based class that wraps the struct. I'll look at that tomorrow

AsyncDisplayKit/ASViewController.mm
```diff
@@ -132,4 +142,70 @@ - (ASInterfaceState)interfaceState
...
+
+- (void)willTransitionToTraitCollection:(UITraitCollection *)newCollection withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
+{
+  [super willTransitionToTraitCollection:newCollection withTransitionCoordinator:coordinator];
+  
+  ASEnvironmentDisplayTraits displayTraits = [self displayTraitsForTraitCollection:self.traitCollection];
+  [self progagateNewDisplayTraits:displayTraits];
+}
+
+- (void)viewWillTransitionToSize:(CGSize)size withTransitionCoordinator:(id<UIViewControllerTransitionCoordinator>)coordinator
```
levi: Why do we need to implement all three of these notifications? `willTransitionToTraitCollection:` should suffice, as it is called in `transitionToSize:` call and is invoked before `traitCollectionDidChange:` is called. I'd love to learn more about the reasons behind this implementation.

rcancro: I noticed that on start up of the app, `traitCollectionDidChange:` is called, but not `willTransitionToTraitCollection:`.

I also feel like i've seen the case where a window changes size (but not trait collection) and `viewWillTransitionToSize:` is called but not `willTransitionToTraitCollection:`.

`progagateNewDisplayTraits:` does check to make sure that the traits are different before propagating, so even in the case where all three of these are called we only propagate once.

appleguy:
@rcancro I am re-running, but I think this may be a real failure.  Probably worth opening the CarthageBuildTest to run it locally and see what the detailed failure is, and confirm the build server's result that before this change, it builds.  It is possible that there is a new problem in master that hasn't been noticed yet though, as we did just adopt a new PINRemoteImage which apparently has some issue with Carthage.  cc @garrettmoon 

▸ Building Sample/CarthageExample [Debug]
▸ Check Dependencies
▸ Compiling main.m
▸ Compiling AppDelegate.m
❌  /Users/travis/build/facebook/AsyncDisplayKit/examples/CarthageBuildTest/CarthageExample/AppDelegate.m:9:9: could not build module 'AsyncDisplayKit'
@import AsyncDisplayKit;

appleguy:
@rcancro @levi if ASDisplayTraits is directly analogous to UITraitCollection, can we call it ASTraitCollection instead?  I assume it will be a struct directly inside ASEnvironment struct.

Sincere apologies that I have not been able to review this yet.  I really badly want to, but haven't been able to reach the minimum sleep quota recently even without touching ASDK diffs in the last week :(.  Will plan to review on the plane on Friday or on return flight.

examples/BackgroundPropertySetting/Sample/ViewController.swift
```diff
@@ -24,8 +24,8 @@ final class ViewController: ASViewController, ASCollectionDelegate, ASCollection
 		layout.minimumInteritemSpacing = padding
 		layout.minimumLineSpacing = padding
 		super.init(node: ASCollectionNode(collectionViewLayout: layout))
-		navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Color", style: .Plain, target: self, action: "didTapColorsButton")
-		navigationItem.leftBarButtonItem = UIBarButtonItem(title: "Layout", style: .Plain, target: self, action: "didTapLayoutButton")
+        navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Color", style: .Plain, target: self, action: #selector(didTapColorsButton))
```
appleguy: Needs more indentation?

.editorconfig
```diff
@@ -0,0 +1,19 @@
+# http://editorconfig.org
```
levi: Include this file in a separate patch.

levi:
@rcancro looks like this patch got messed up with a rebase or merge.

AsyncDisplayKit/ASViewController.mm
```diff
@@ -132,4 +142,70 @@ - (ASInterfaceState)interfaceState
...
+}
+
+- (void)progagateNewDisplayTraits:(ASEnvironmentTraitCollection)traitCollection
+{
+  ASEnvironmentState environmentState = self.node.environmentState;
+  ASEnvironmentTraitCollection oldTraitCollection = environmentState.traitCollection;
+  
+  if (ASEnvironmentTraitCollectionIsEqualToASEnvironmentTraitCollection(traitCollection, oldTraitCollection) == NO) {
+    environmentState.traitCollection = traitCollection;
+    [self.node setEnvironmentState:environmentState];
```
levi: Are you able to use dot syntax here?

AsyncDisplayKit/ASViewController.mm
```diff
@@ -132,4 +142,70 @@ - (ASInterfaceState)interfaceState
...
+{
+  if (self.overrideDisplayTraitsWithWindowSize) {
+    ASTraitCollection *traitCollection = self.overrideDisplayTraitsWithWindowSize(windowSize);
+    traitCollection.isMutable = NO;
+    return [traitCollection environmentTraitCollection];
+  }
+  return self.node.environmentState.traitCollection;
+}
+
+- (void)progagateNewDisplayTraits:(ASEnvironmentTraitCollection)traitCollection
```
levi: Is there any value in moving this to the node itself? Seems like it doesn't have a lot of relationship with the view controller.

AsyncDisplayKit/ASViewController.mm
```diff
@@ -132,4 +142,70 @@ - (ASInterfaceState)interfaceState
...
+}
+
+- (ASEnvironmentTraitCollection)displayTraitsForWindowSize:(CGSize)windowSize
+{
+  if (self.overrideDisplayTraitsWithWindowSize) {
+    ASTraitCollection *traitCollection = self.overrideDisplayTraitsWithWindowSize(windowSize);
+    traitCollection.isMutable = NO;
+    return [traitCollection environmentTraitCollection];
+  }
+  return self.node.environmentState.traitCollection;
```
levi: Would be nice to reduce the amount of references here. Perhaps add a shortcut to asdisplaynode to get the underlying environmentState trait collection?

AsyncDisplayKit/ASViewController.mm
```diff
@@ -132,4 +142,70 @@ - (ASInterfaceState)interfaceState
   return _node.interfaceState;
 }
 
+#pragma mark - ASEnvironmentTraitCollection
+
+- (ASEnvironmentTraitCollection)displayTraitsForTraitCollection:(UITraitCollection *)traitCollection
```
levi: Might be a better opportunity to allow the developer to provide a subclass of ASTraitCollection (or ASEnvironmentTraitCollection) with extra properties that `displayTraitsContext` currently provide. For the common case, most people won't need to have `displayTraitsContext`, so figuring out a way to reduce API footprint would be a bonus. Are we constrained introducing a design like that?

AsyncDisplayKit/ASViewController.mm
rcancro: The thinking was that the VC would be the one that knows of any special trait collection it may need based on the circumstances. Like UIViewController has  setOverrideTraitCollection method, not UIView.

AsyncDisplayKit/ASViewController.mm
rcancro: you mean `self.node.environmentState = environmentState`?

AsyncDisplayKit/Details/ASTraitCollection.h
```diff
@@ -0,0 +1,29 @@
...
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#import <UIKit/UIKit.h>
+#import <AsyncDisplayKit/ASEnvironment.h>
+
+@interface ASTraitCollection : NSObject
+
+@property (nonatomic, assign) BOOL isMutable;
```
levi: I'd recommend dropping this in favor of making the ASTraitCollection immutable the moment it's initialized. Will require a larger initializer, but will reduce potential mistakes.

AsyncDisplayKit/ASViewController.mm
levi: yeah

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2710,9 +2713,19 @@ - (BOOL)supportsUpwardPropagation
...
 
+- (BOOL)supportsTraitsCollectionPropagation
+{
+  return ASEnvironmentStateTraitCollectionPropagationEnabled();
+}
+
 ASEnvironmentLayoutOptionsForwarding
 ASEnvironmentLayoutExtensibilityForwarding
 
+- (ASTraitCollection *)asyncTraitCollection
```
levi: What about asdk_traitCollection?

AsyncDisplayKit/ASViewController.mm
rcancro: Ultimately we are passing around trait collections as a struct, not a class. It has to be a struct to live in ASEnvironment. The class is just there to be a nicer API to the end user (that probably doesn't care about the context).

AsyncDisplayKit/ASDisplayNode.mm
rcancro: screen shot 2016-05-10 at 16 33 57

AsyncDisplayKit/Details/ASEnvironment.mm
```diff
@@ -0,0 +1,83 @@
...
+    asyncTraitCollection.verticalSizeClass = traitCollection.verticalSizeClass;
+    asyncTraitCollection.userInterfaceIdiom = traitCollection.userInterfaceIdiom;
+    if (AS_AT_LEAST_IOS9) {
+      asyncTraitCollection.forceTouchCapability = traitCollection.forceTouchCapability;
+    }
+  }
+  return asyncTraitCollection;
+}
+
+BOOL ASEnvironmentTraitCollectionIsEqualToASEnvironmentTraitCollection(ASEnvironmentTraitCollection traitCollection0, ASEnvironmentTraitCollection traitCollection1)
```
levi: lhs, rhs instead of traitCollection0, traitCollection1

AsyncDisplayKit/ASViewController.mm
rcancro: @maicki had the idea to expose the trait collection as ASTraitCollection to the user, but store it as a struct behind the scenes. When providing a custom trait collection the user overrides one of two blocks, both of which return ASTraitCollection* not the struct.

I guess my point is that we are trying to make it difficult for the user to get to the struct trait collection since most obj-c developers are more familiar using classes. By adding a getter to the struct traits we kind of break that.

AsyncDisplayKit/Details/ASTraitCollection.h
rcancro: Sounds good. FWIW, when ASViewController creates a ASTraitCollection via its override blocks, it immediately sets it to immutable (i.e., the user doesn't have to do it). This is similar to how layout specs get set to immutable. 

Though I'm not wild that isMutable is read/write so a user could just set it back to NO.

AsyncDisplayKit/ASViewController.mm
levi: We could introduce it as a private method for these use cases.

AsyncDisplayKit/ASDisplayNode.mm
levi: -_-

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -519,8 +520,17 @@ - (void)accessDataSourceSynchronously:(BOOL)synchronously withBlock:(dispatch_bl
     for (NSUInteger i = 0; i < rowNum; i++) {
       NSIndexPath *indexPath = [sectionIndex indexPathByAddingIndex:i];
       ASCellNodeBlock nodeBlock = [_dataSource dataController:self nodeBlockAtIndexPath:indexPath];
+      
+      // When creating a node, make sure to pass along the current display traits so it will be laid out properly
+      ASCellNodeBlock nodeBlockPropagatingDisplayTraits = ^{
+        ASCellNode *cellNode = nodeBlock();
+        id<ASEnvironment> environment = [self.environmentDelegate dataControllerEnvironment];
+        ASEnvironmentStatePropagateDown(cellNode, [environment environmentState].traitCollection);
```
levi: would be nice to have something like `environment.traitCollection`, even if it's brought in by a category.

AsyncDisplayKit/ASViewController.h
```diff
@@ -9,14 +9,31 @@
...
+typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitCollectionBlock)(UITraitCollection *traitCollection);
+typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitWindowSizeBlock)(CGSize windowSize);
+
 @interface ASViewController<__covariant DisplayNodeType : ASDisplayNode *> : UIViewController
 
 - (instancetype)initWithNode:(DisplayNodeType)node NS_DESIGNATED_INITIALIZER;
 
 @property (nonatomic, strong, readonly) DisplayNodeType node;
 
+@property (nonatomic, strong) id displayTraitsContext;
```
levi: The documentation of this property should be very clear — its usefulness and also the internal memory behavior.

AsyncDisplayKit/ASViewController.mm
rcancro: Oh, you are talking about moving `progagateNewDisplayTraits:` to the node level. That is possible. The VC will still be the one to initiate the propagation since it is the one that knows when the trait collection actually changed. Personally, I like that all the trait collection details are encapsulated inside of the VC, but I definitely see your point.

AsyncDisplayKit/ASViewController.h
```diff
@@ -22,7 +22,17 @@ typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitWindowSizeBlock)(C
...
+ *  An optional context to pass along with an ASTraitCollection.
+ *  This can be used to pass any internal state to all subnodes via the ASTraitCollection that is not
+ *  included in UITraitCollection. This could range from more fine-tuned size classes to a class of
+ *  constants that is based upon the new trait collection.
+ *
+ *  Be aware that internally this context is held by a C struct which cannot retain the pointer. Therefore
+ *  ASVC keeps a strong reference to the context to make sure that it stays alive. If you change this value
+ *  it will propagate the change to the subnodes.
+ */
+@property (nonatomic, strong) id _Nullable traitColectionContext;
```
levi: typo

AsyncDisplayKit/ASViewController.mm
```diff
@@ -50,9 +51,8 @@ - (instancetype)initWithNode:(ASDisplayNode *)node
 
 - (void)dealloc
 {
-  if (_displayTraitsContext != nil) {
-    ASEnvironmentTraitCollectionClearDisplayContext(self.node);
-    _displayTraitsContext = nil;
+  if (_traitColectionContext != nil) {
```
levi: Add a note as to why you're clearing it out in dealloc.

levi:
Ricky, this is straight up poetry.

AsyncDisplayKit/ASViewController.h
rcancro: d'oh! good catch

rcancro:
Rotation, oh my
iPad, 6 plus and pad pro
trait collections yay!

AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -93,6 +127,41 @@ ASDISPLAYNODE_EXTERN_C_END
...
+// downward propagation via ASEnvironment. Instead if the new environmentState has displayTraits that are different from
+// the cells', then we propagate downward explicitly and request a relayout.
+//
+// If there is any new downward propagating state, it should be added to this define.
+//
+// This logic is used in both ASCollectionNode and ASTableNode
+#define ASEnvironmentCollectionTableSetEnvironmentState(lock) \
+- (void)setEnvironmentState:(ASEnvironmentState)environmentState\
+{\
+  ASDN::MutexLocker l(lock);\
```
appleguy: Pretty damn fancy, sir @rcancro

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -168,11 +169,13 @@ - (NSArray *)children
 
 - (ASEnvironmentState)environmentState
 {
+  ASDN::MutexLocker l(_propertyLock);
```
appleguy: I am not certain this is safe.  I believe there is a deadlock risk here - let's reference the blame in this file for the change that had to remove a certain lock.  @rcancro @garrettmoon @maicki

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -193,9 +201,20 @@ - (void)propagateUpLayoutable:(id<ASLayoutable>)layoutable
   }
 }
 
+- (ASEnvironmentTraitCollection)environmentTraitCollection
+{
+  return _environmentState.traitCollection;
```
appleguy: I think the locking may be unsafe in this file, BUT - if we do lock for accessing _environmentState below in asyncTraitCollection, we probably should lock here

appleguy:
Phenomenal work, @rcancro !  Once again, you've fundamentally extended the capabilities of the framework.  I sincerely appreciate your attention to detail both in architecting and executing this.  There's great care to efficiency / performance, but simultaneously a clear priority placed on a friendly object-based API.  Love it.

The only followup that seems important for the next 1-2 days is checking on the locking sites I commented at, but otherwise I think this is a mic drop. :)

AsyncDisplayKit/Layout/ASLayoutSpec.mm
maicki: Agree we have to be careful with adding locks in ASEnvironment. E.g. we didn't lock layout options in the past before ASEnvironment and as we added locks for the ASEnvironmentLayoutOptions deadlocks happened. Especially if we traverse the node tree's from bottom up (e.g. it happens if called `setNeedsLayout`) and the same time the layout happens from top down (e.g. call `measureWithSizeRange:` directly)

Issue #1512, opened 2016-04-11
[ASCellNode] Assertion on cell deallocation due visibility not being cleared


tomizimobile:
Using my trusty test project (https://github.com/tomizimobile/ASDKSearchIssue) on the latest master (a6e66fa5d758d6675ddee982c128762e8a86a31a) I've managed to track down and find an issue with cell deallocation under a very specific circumstance, similar to https://github.com/facebook/AsyncDisplayKit/issues/1436. What I've found is when the tableView is deallocated when there's a section header only partially visible at the bottom of the table, it will crash. To reproduce, just go into the details page of the test project until you see a layout as described, and go back before scrolling. The test project generates random data each time the search view controller is loaded.

Here's a screenshot of an example layout that will cause a crash:

![screen shot 2016-04-11 at 4 57 09 pm](https://cloud.githubusercontent.com/assets/7967500/14442336/239f22ae-0007-11e6-8920-1fc2bc7c8278.png)

appleguy:
@tomizimobile I was able to reproduce by typing, but I cannot get to the screenshot you are showing that has a back button.  Tapping the cells does nothing...

It would be much easier to debug with a one-shot reproduction like you describe, any tips?  cc @maicki FYI

tomizimobile:
@appleguy Maybe you're using an older version of that sample project? The app now starts with a screen that just has a button to push a view controller. To reproduce I would just go back and forth until the UI is set up like that, it happens pretty quickly. I did forget to mention this was on an iPhone 6 - I'm sure with other devices, the UI layout might be harder to get randomly, so maybe playing around with hardcoding the data could yield this UI.

tomizimobile:
@appleguy I updated my test project to use static data, and now the crash should be reproducible every time on an iPhone 6 device/simulator. All you need to do is press the button to push the view controller of names, and then press the back button.

maicki:
@tomizimobile @appleguy Just yesterday we discovered what I think is the same problem and we should fix soon. I will add a link to the issue in here: #1569

As @tomizimobile it is 100% reproducible on an iPhone 6 simulator with the sample app in this issue or in #1569

Issue #1569, opened 2016-04-20
ASTableNode crash -[ASDisplayNode dealloc]


hebertialmeida:
I have a ASViewController with ASTableNode, when navigating back to previous controller the app crashes.
![screen shot 2016-04-20 at 9 47 16 am](https://cloud.githubusercontent.com/assets/103670/14679405/19fc9674-06f0-11e6-97fd-7df76cdf14ba.png)

Here is a sample project:
https://github.com/hebertialmeida/ASDKCrash

Open the project > Tap on Settings > Back to profile and Crash.

Any help would be nice.

------

**Update:** @maicki Found that the problem is related to `ASViewController`.

------

**Update2:** The problem only happen when the ASTableNode is defined with Style Grouped `ASTableNode(style: .Grouped)`, probably not related to `ASViewController `.

yichungwang:
I'm having the same issue too, for now I just remove ASTableView from superview in viewwilldisappear.

maicki:
@hebertialmeida @yichungwang Can you two please verify if you have that problem on the latest master?

hebertialmeida:
@maicki Yes, seems that it fixed for me.

PR #1593, opened 2016-04-26
[ASDisplayNode] Fix assertion on cell deallocation due visibility not being cleared

maicki:
Should fix: #1512 and #1569

Further information:
- We mark every node as visible in the ASRangeController which NSIndexPath is returned from visibleNodeIndexPathsForRangeController:
- In visibleNodeIndexPathsForRangeController: we get the visible index path's via a call to UITableView's "indexPathsForVisibleRows" method.
- Unfortunately in this case we cannot use indexPathsForVisibleRows to get all the visible index paths as apparently in a grouped UITableView it would return index paths for cells that are just a bit over the edge of the visible area.
- But this edge cells will never get a call for -tableView:cellForRowAtIndexPath:, but we will mark them as visible in the range controller
- In tableView:cellForRowAtIndexPath: we call -configureContentView:forCellNode
- Because we never get a -configureContentView:forCellNode call for the edge cells, the _ASDisplayView of the nodes will never be added to the window and get a willMoveToWindow and didMoveToWindow call and it's never get's added to the window for now and so the node is NOT marked as "in the hirarchy"
- If the deallocation of the views are happening without the UITableView ever scrolled, the cells don't get a call to __exitHierarchy as they were never added to the window and stay in the interface state "visible" and an exception will be raised within the dealloc method of the ASDisplayNode

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -991,9 +991,9 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
 - (void)dealloc
 {
   ASDisplayNodeAssertMainThread();
-  [_completedNodes enumerateKeysAndObjectsUsingBlock:^(NSString *kind, NSMutableArray *nodes, BOOL *stop) {
-    [nodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
-      [section enumerateObjectsUsingBlock:^(ASCellNode *node, NSUInteger rowIndex, BOOL *stop) {
+  [_completedNodes enumerateKeysAndObjectsUsingBlock:^(NSString *kind, NSMutableArray *sections, BOOL *stop) {
+    for (NSArray *section in sections) {
+      for (ASCellNode *node in section) {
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -991,18 +991,18 @@ - (NSArray *)completedNodes
 - (void)dealloc
 {
   ASDisplayNodeAssertMainThread();
-  [_completedNodes enumerateKeysAndObjectsUsingBlock:^(NSString *kind, NSMutableArray *nodes, BOOL *stop) {
-    [nodes enumerateObjectsUsingBlock:^(NSMutableArray *section, NSUInteger sectionIndex, BOOL *stop) {
-      [section enumerateObjectsUsingBlock:^(ASCellNode *node, NSUInteger rowIndex, BOOL *stop) {
+  [_completedNodes enumerateKeysAndObjectsUsingBlock:^(NSString *kind, NSMutableArray *sections, BOOL *stop) {
```
maicki: @eanagel Hi Ethan, do we still need to remove all the layer / views from the node in here? Why did we added that in the first place?

cc @appleguy

PR #1596, opened 2016-04-26
[Build / Xcode] Add .editorconfig for coding style

maicki:
More infos at: http://editorconfig.org


appleguy:
Nice, thanks @maicki !

PR #1599, opened 2016-04-27
[ASVideoNode] Fix UI thread blocking with remote assets

ejensen:
## Background
Originally [reported in Slack](https://asyncdisplaykit.slack.com/archives/general/p1461569479000382). `ASVideoNode` is blocking the UI thread while it loads metadata. It's hard to notice when using a local file, but it's obvious with a remote file on a poor network. @yichungwang created a [simple project to demonstrate the issue](https://asyncdisplaykit.slack.com/files/daydreamer/F1429JP26/videonodesample.zip) 👏 

## Cause
`AVAsset`'s tracks metadata properties block the calling thread until it the track metadata is known. If the asset is remote, this blocks the thread while waiting on the network.

1.  PR https://github.com/facebook/AsyncDisplayKit/pull/1552 added a call to the `tracks` property on the main thread. 
2.  The changed pulled in from https://github.com/facebook/AsyncDisplayKit/pull/1546 added a call to `tracksWithMediaCharacteristic:` while `_videoLock` is locked.

## Proposed Fix
1.  Restructure `constructPlayerItem` so `tracks` does not need to be checked.
2.  Limit the locking in `imageAtTime`.

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -99,17 +99,13 @@ static NSString * const kStatus = @"status";
 {
   ASDN::MutexLocker l(_videoLock);
 
-  AVPlayerItem *playerItem = nil;
-  
-  if (_asset != nil) {
-    if (_asset.tracks.count > 0) {
-      playerItem = [[AVPlayerItem alloc] initWithAsset:_asset];
-    } else if ([_asset isKindOfClass:[AVURLAsset class]]) {
-      playerItem = [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];
-    }
+  if ([_asset isKindOfClass:[AVURLAsset class]]) {
+    return [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];
+  } else if (_asset != nil) {
+    return [[AVPlayerItem alloc] initWithAsset:_asset];
   }
 
-  return playerItem;
+  return nil;
 }
 
 - (void)addPlayerItemObservers:(AVPlayerItem *)playerItem
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -173,16 +173,16 @@ static NSString * const kStatus = @"status";
 - (void)imageAtTime:(CMTime)imageTime completionHandler:(void(^)(UIImage *image))completionHandler
 {
   ASPerformBlockOnBackgroundThread(^{
-    ASDN::MutexLocker l(_videoLock);
+    AVAsset *asset = self.asset;
 
     // Skip the asset image generation if we don't have any tracks available that are capable of supporting it
-    NSArray<AVAssetTrack *>* visualAssetArray = [_asset tracksWithMediaCharacteristic:AVMediaCharacteristicVisual];
+    NSArray<AVAssetTrack *>* visualAssetArray = [asset tracksWithMediaCharacteristic:AVMediaCharacteristicVisual];
     if (visualAssetArray.count == 0) {
       completionHandler(nil);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


appleguy:
@ejensen now that I recall some of the issues we had with APIs as simple as .duration at Instagram, I believe I understand the severe impact of this issue and am fully supportive of this change.

However, can you please clarify if you're confident that this new implementation will address the behavior and that the check of the number of tracks originally intended to provide? I believe it related to HTTP live streaming.  cc @gazreese

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -99,17 +99,13 @@ - (AVPlayerItem *)constructPlayerItem
...
-      playerItem = [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];
-    }
+  if ([_asset isKindOfClass:[AVURLAsset class]]) {
+    return [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];
+  } else if (_asset != nil) {
+    return [[AVPlayerItem alloc] initWithAsset:_asset];
   }
 
-  return playerItem;
+  return nil;
```
appleguy: Although I am quite happy with the style to eliminate the local variable here, in this kind of formatting, I would prefer that we have an unconditional else clause at the end that includes the fall through return. It makes it more clear that we expect early returns in all cases. Certainly not an issue for this change, and no need for a change just for this, just an offhanded style musing!

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -177,16 +173,16 @@ - (void)generatePlaceholderImage
 - (void)imageAtTime:(CMTime)imageTime completionHandler:(void(^)(UIImage *image))completionHandler
 {
   ASPerformBlockOnBackgroundThread(^{
-    ASDN::MutexLocker l(_videoLock);
+    AVAsset *asset = self.asset;
```
appleguy: Why is it essential to eliminate this lock? Isn't it possible that the asset could change before the next call, which could mean that code which intends to cancel loading of the media may not be able to? I am very unfamiliar with this area of the API, and so I am not sure if it is plausible to initiate an operation here that may require downloading a significant amount of information from the network, and thus should be cancelable if there is a race with setAsset:nil on another thread.

ejensen: `tracksWithMediaCharacteristic:` blocks the current thread (background thread in this case) until it loads the required metadata from the asset. That loading may be over a network. If `_videoLock` is lock on a background thread that's waiting on the network, other threads, such as the UI thread are blocked attempting to acquire the lock.  In order to prevent the blocking while the metadata is loaded, the lock needs to be limited in scope. 

Repeated access of the `asset` variable is the reason for the original lock. This change makes a local snapshot of the `asset` variable to ensure it's consistent throughout the method. A data consistency check made in the completion block of imageAtTime:

``` obj-c
// Ensure the asset hasn't changed since the image request was made
if (ASAssetIsEqual(weakSelf.asset, asset)) {
  [weakSelf setVideoPlaceholderImage:image];
}
```

If the `asset` is changed while the placeholder image is being generated, the image is ignored. Preventing the incorrect image from displaying.

I can change `imageAtTime` to be a static variable with that accepts an asset as a parameter. That way it doesn't access any instance variables and it's clear it doesn't need to acquire _videoLock.

ejensen:
@appleguy I've verified this implementation fixes the UI blocking issue demonstrated in the demo project.  The HLS workaround (extracting the URL from the asset) should remain unchanged. Though I don't have a great way to test HLS.

AsyncDisplayKit/ASVideoNode.mm
ejensen: Moving `imageAtTime` out of `ASVideoNode` is implemented here: https://github.com/facebook/AsyncDisplayKit/compare/master...ejensen:video-ui-blocking Please let me know if you feel that makes the data consitency logic more or less clear. Any other suggestions are more than welcome.

PR #1605, opened 2016-04-28
[ASDisplayNode] Added callbacks for entering and exiting fetch data and display ranges

lappp9:
So at the moment, the naming that seemed to make the most sense was something similar to the view controller lifecycle callback naming conventions.

- (void)didEnterDisplayRange;
- (void)didExitDisplayRange;

- (void)didEnterFetchDataRange;
- (void)didExitFetchDataRange;

I like that -visibilityDidChange: returns a boolean representing whether it is entering or exiting the visible range, but getting the other ones to work that way seems like it would be tough to do without having some sort of difficult to understand naming going on.

Also, I know there is already -fetchData and -clearFetchedData as well as -clearContents, but having these specifically as override methods that don't necessarily imply that you have to do anything specific could be nice?

### DIFF
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -214,9 +214,9 @@
   // To be overriden by subclasses
 }
 
-- (void)visibileStateDidChange:(BOOL)isVisible
+- (void)visibleStateDidChange:(BOOL)isVisible
 {
-  [super visibileStateDidChange:isVisible];
+  [super visibleStateDidChange:isVisible];
   
   CGRect cellFrame = CGRectZero;
   if (_scrollView) {
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -214,9 +214,9 @@
   // To be overriden by subclasses
 }
 
-- (void)visibilityDidChange:(BOOL)isVisible
+- (void)visibileStateDidChange:(BOOL)isVisible
 {
-  [super visibilityDidChange:isVisible];
+  [super visibileStateDidChange:isVisible];
   
   CGRect cellFrame = CGRectZero;
   if (_scrollView) {
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -243,32 +243,22 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)visibilityDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
- * @abstract Called whenever the the node enters the display range.
+ * @abstract Called whenever the the node has entered or left the display state.
  *
- * @discussion Subclasses may use this to monitor when they enter the display range.
+ * @discussion Subclasses may use this to monitor when a node should be rendering its content.
+ *
+ * @note This method can be called from any thread and should therefore be thread safe.
  */
-- (void)didEnterDisplayRange ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)displayStateDidChange:(BOOL)inDisplayState ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
- * @abstract Called whenever the the node exits the display range.
+ * @abstract Called whenever the the node has entered or left the load state.
  *
- * @discussion Subclasses may use this to monitor when they exit the display range.
- */
-- (void)didExitDisplayRange ASDISPLAYNODE_REQUIRES_SUPER;
-
-/**
- * @abstract Called whenever the the node enters the fetch data range.
+ * @discussion Subclasses may use this to monitor data for a node should be loaded, either from a local or remote source.  
  *
- * @discussion Subclasses may use this to monitor when they enter the fetch data range.
+ * @note This method can be called from any thread and should therefore be thread safe.
  */
-- (void)didEnterFetchDataRange ASDISPLAYNODE_REQUIRES_SUPER;
-
-/**
- * @abstract Called whenever the the node exits the fetch data range.
- *
- * @discussion Subclasses may use this to monitor when they exit the fetch data range.
- */
-- (void)didExitFetchDataRange ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)loadStateDidChange:(BOOL)inLoadState ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
  * Called just before the view is added to a window.
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -242,6 +242,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)visibilityDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
 
+- (void)didEnterDisplayRange ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)didExitDisplayRange ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)didEnterFetchDataRange ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)didExitFetchDataRange ASDISPLAYNODE_REQUIRES_SUPER;
+
 /**
  * Called just before the view is added to a window.
  */
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -240,10 +240,17 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @discussion Subclasses may use this to monitor when they become visible.
  */
-- (void)visibileStateDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)visibilityDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
- * @abstract Called whenever the the node has entered or left the display state.
+ * @abstract Called whenever the visiblity of the node changed.
+ *
+ * @discussion Subclasses may use this to monitor when they become visible.
+ */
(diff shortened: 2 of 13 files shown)


appleguy:
Thanks @lappp9 !  @maicki, @levi, @nguyenhuy I could use your help considering the API here.

We definitely need these call points.  Things like displayWillStart are not actually called for elements that don't have a literal display method, and something like -clearFetchedData doesn't feel parallel to -visibilityDidChange:.

That said, I'm not a fan of these particular names.  They seem a bit too long, at least the "Range" part should probably be removed as we plan to have ASViewController also have an independent ASInterfaceState, which might be in the fetch data state just by being one of the non-selected tabs of a tab bar controller (not a range exactly).

We should see if any names could be found that will allow the BOOL parameter to work, as it fully halves the number of strings that a developer needs to remember to use the API without looking at headers or reference.  ..., displayStateDidChange: is one candidate, not sure it's great though.

lappp9:
@appleguy yeah, I agree that the BOOL versions of these methods would make things quite a bit nicer.
So:
`-visibilityDidChange:`
`-displayStateDidChange:`
`-fetchDataStateDidChange:` 

The display one sounds good. The fetchData one feels a little weird, but I think it would make sense with the other ones.  Does anyone else have opinions?

Maybe?
`-dataFetchStateDidChange:`

appleguy:
It would be a bit painful to rename, but maybe we should use visibilityStateDidChange:.  

The others sound pretty good to me.  I think fetchDataState is probably best because autocomplete is important.

appleguy:
I guess we could actually rename it if we want, since the only other method people use today is -fetchData, which is replaced by this.

Is there anything besides "fetch" that could be used for this, maybe even one word?  I've used variants that are more network-focused but I want it to also encompass the idea of going to a database or disk cache.

cc @maicki  @garrettmoon, @levi  for API naming insights!

levi:
I like the idea of having easily accessible callback methods for these events. However, could we not coalesce them all into a version of InterfaceStateDidChange that is easily useable for the developer subclassing ASDisplayNode? There's a lot of flexibility and power in doing membership checking against the interfaceState bitmask.

appleguy:
@levi I would love to do it that way, but in practice it is quite hard to isolate whether a bit has changed, _and_ whether it is now 1 or 0.  Just checking whether it's set is not enough.  E.g. FetchData may be 1 now, but it may also have been 1 before.

The ASDisplayNode implementation of interfaceStateDidChange: shows some of the patterns that developers would have to individually figure out how to reconstruct.

lappp9:
Alright, i did some tweaking and right now the naming is: 

`-(void)visibleStateDidChange:(BOOL)isVisible`
`- (void)displayStateDidChange:(BOOL)inDisplayState`
`- (void)loadStateDidChange:(BOOL)inLoadState`

Any thoughts?

levi:
@lappp9 I suggest reversing the naming of the methods to have the parameter reflect the new state as an alternative of `didChange:`.
```
- (void)didChangeVisibleState:(BOOL)hasVisibleState;
- (void)didChangeLoadState:(BOOL)hasLoadState;
- (void)didChangeDisplayState:(BOOL)hasDisplayState;
```

maicki:
@levi @appleguy  You don't think we should be consistent with our API in ASDK? Currently methods similar to them are:

```
- (void)interfaceStateDidChange:(ASInterfaceState)newState fromState:(ASInterfaceState)oldState ASDISPLAYNODE_REQUIRES_SUPER;
- (void)visibilityDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
- (void)layoutDidFinish ASDISPLAYNODE_REQUIRES_SUPER;
- (void)calculatedLayoutDidChange ASDISPLAYNODE_REQUIRES_SUPER;
```

As you can see `..didChange..` is mostly at the end. Looking up Apple API's unfortunately they are not consistent with that stuff in UIKit so there is no clear "best practice" :/ ... but I think we should be consistent in ASDK

lappp9:
@levi Is a "visible state" or a "display state" something that a node "has" per se?  I think the alternative naming feels a little weird personally but I could be missing something.

As @maicki  said, I also like these three names because they match
`- interfaceStateDidChange:fromState:`

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2172,6 +2188,12 @@ - (void)setInterfaceState:(ASInterfaceState)newState
         }
       }
     }
+    
+    if (nowDisplay) {
```
appleguy: @lappp9 we could collapse this for now like the visible example below; e.g. displayStateDidChange:nowDisplay

AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -236,7 +236,7 @@ - (void)doVideoStartsPlayingOnDidDidBecomeVisibleWhenShouldAutoplay
   }];
   _videoNode.playerNode.layer.frame = CGRectZero;
   
-  [_videoNode visibilityDidChange:YES];
+  [_videoNode visibileStateDidChange:YES];
```
appleguy: Why is this called manually?

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -210,9 +210,9 @@ - (void)cellNodeVisibilityEvent:(ASCellNodeVisibilityEvent)event inScrollView:(U
   // To be overriden by subclasses
 }
 
-- (void)visibilityDidChange:(BOOL)isVisible
+- (void)visibileStateDidChange:(BOOL)isVisible
```
appleguy: The new name is growing on me!  It's probably the right choice.

@lappp9 I think we need to re-add a copy of the original method and keep it working for now.  There are probably 100 definitions of this method in Pinterest, and while I can certainly land a find-and-replace, I don't want to make a tight timing connection to updating the framework there.  Other clients will no doubt be affected as well.

For now let's just add the old method without any special warnings so I can do a normal release of the framework.  As soon as we go to 2.0 Beta we will add a deprecation warning, and perhaps with 2.0 GM we can remove it entirely.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -240,7 +240,25 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @discussion Subclasses may use this to monitor when they become visible.
  */
-- (void)visibilityDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)visibileStateDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
+
+/**
+ * @abstract Called whenever the the node has entered or left the display state.
```
appleguy: entered or exited* rather than left

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -240,7 +240,25 @@ NS_ASSUME_NONNULL_BEGIN
...
+- (void)displayStateDidChange:(BOOL)inDisplayState ASDISPLAYNODE_REQUIRES_SUPER;
+
+/**
+ * @abstract Called whenever the the node has entered or left the load state.
+ *
+ * @discussion Subclasses may use this to monitor data for a node should be loaded, either from a local or remote source.  
+ *
+ * @note This method can be called from any thread and should therefore be thread safe.
+ */
+- (void)loadStateDidChange:(BOOL)inLoadState ASDISPLAYNODE_REQUIRES_SUPER;
```
appleguy: Let's keep this name for now, as discussed.  I will think about whether we should rename the FetchData state (like the NSOptions value in interface state)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2076,11 +2080,21 @@ - (void)recursivelyClearFetchedData
   });
 }
 
-- (void)visibilityDidChange:(BOOL)isVisible
+- (void)visibileStateDidChange:(BOOL)isVisible
```
appleguy: We'll need a copy of this method so the base class can always respond to the old name.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2180,14 +2202,15 @@ - (void)setInterfaceState:(ASInterfaceState)newState
   BOOL wasVisible = ASInterfaceStateIncludesVisible(oldState);
 
   if (nowVisible != wasVisible) {
-    [self visibilityDidChange:nowVisible];
+    [self visibileStateDidChange:nowVisible];
```
appleguy: Here we can call both methods back to back, unconditionally, as long as the base class has a blank implementation of both.

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -135,9 +135,9 @@ - (void)stopAnimating
   [self.animatedImage clearAnimatedImageCache];
 }
 
-- (void)visibilityDidChange:(BOOL)isVisible
+- (void)visibileStateDidChange:(BOOL)isVisible
```
appleguy: It's fine to keep all of these renamed ones - adopting the new name is great!

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -243,6 +243,31 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)visibilityDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
+ * @abstract Called whenever the visiblity of the node changed.
+ *
+ * @discussion Subclasses may use this to monitor when they become visible.
```
appleguy: @lappp9 we should add that this is always called on the main thread.  Edit the comment for the prior version to say it is planned to be deprecated in a future release, and called in the same cases as the newer visibleStateDidChange: method.

appleguy:
@lappp9 thanks for making this!  @maicki @levi @Adlai-Holler @nguyenhuy @hannahmbanana - make sure you are familiar with these important new APIs, which supercede fetchData, clearFetchedData, clearContents, and visibilityDidChange:.  Please weigh in with any thoughts on naming improvements as they are currently free to change.  Note that previously an equivalent to displayStateDidChange:YES did not exist; displayWillStart is NOT a direct equivalent.  The :NO variant is equivalent to -clearContents, currently.

@hannahmbanana we need to update the documentation for this.

PR #1606, opened 2016-04-28
[ASNetworkImageNode] Fix threading issue in current image quality

wendylu:
Summary:
Because we trampoline to main when setting _currentImageQuality, there would be situations where displayWillStart was called, we get a cached image and set currentImageQuality to 1.0, and then a background thread calling setDefaultImage would enqueue an operation that sets currentImageQuality to 0 and overwrites the correct value.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -124,9 +124,15 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 
   if (reset || _URL == nil) {
     self.image = _defaultImage;
-    ASPerformBlockOnMainThread(^{
-      self.currentImageQuality = 1.0;
-    });
+    if (_URL == nil) {
+      ASPerformBlockOnMainThread(^{
+        self.currentImageQuality = 1.0;
+      });
+    } else {
+      ASPerformBlockOnMainThread(^{
+        self.currentImageQuality = 0.0;
+      });
+    }
   }
 
   if (self.interfaceState & ASInterfaceStateFetchData) {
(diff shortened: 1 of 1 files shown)


facebook-github-bot-7:

wendylu:
I'm wondering if we should use dispatch_async in all instances to make sure the sets get executed in the correct order regardless of the queuing thread

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -124,9 +124,15 @@ - (void)setURL:(NSURL *)URL resetToDefault:(BOOL)reset
 
   if (reset || _URL == nil) {
     self.image = _defaultImage;
-    ASPerformBlockOnMainThread(^{
-      self.currentImageQuality = 1.0;
-    });
+    if (_URL == nil) {
```
appleguy: Let's refactor this to only have one dispatch.

Above the outer if, you can do something like BOOL hasURL = (_URL != nil);.  Then you can eliminate the inner if ' else, and do ... = (hasURL ? 0.0 : 1.0);

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -229,7 +241,9 @@ - (void)displayWillStart
       if (result) {
         self.image = result;
         _imageLoaded = YES;
-        _currentImageQuality = 1.0;
+        dispatch_async(dispatch_get_main_queue(), ^{
```
appleguy: Because it is (apparently?) important to have this dispatch even if already running on main, please add a comment here about why that is.  This is key because a later author could easily refactor it to be an ASPerformBlockOnMainThread() otherwise.

appleguy:
@wendylu thanks for the follow up!  Still excited about this new capability.  Send me a ping when the final revision is up and I'll get it merged ASAP.

Issue #1600, opened 2016-04-27
[ASVideoNode] Use of URL in construction


gazreese:
I'm not sure whether this is a bug, but thought I'd raise it anyway for discussion with @appleguy.

```
- (AVPlayerItem *)constructPlayerItem
{
  ASDN::MutexLocker l(_videoLock);

  AVPlayerItem *playerItem = nil;
  
  if (_asset != nil) {
    if (_asset.tracks.count > 0) {
      playerItem = [[AVPlayerItem alloc] initWithAsset:_asset];
    } else if ([_asset isKindOfClass:[AVURLAsset class]]) {
      playerItem = [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];
    }
  }

  return playerItem;
}
```

In my original changes to add URL / HLS Video support to the ASVideoNode I had explicit construction from either a URL or an AVAsset. This was because during my testing I had problems with the AVPlayerItem if I tried to construct from an HLS AVURLAsset rather than constructing directly from the URL. I can't remember exactly what the problem was if I'm honest, but it seemed prudent to mirror the construction of the AVPlayerItem in the ASVideoNode interface as this was the underlying tech.

Since the original merge the API for ASVideoNode has been simplified to only take an AVAsset. I assume someone has run into similar problems I had however as looking at the code above the internals have been reverted to creating the AVPlayerItem directly from the URL, except this time the URL is extracted from the AVAsset and the player item is constructed without it.

What this does however is drop the use of the AVURLAsset that the user uses for construction without any warning. If the user constructed the video node with a normal AVAsset then after construction videoNode.playerItem.asset would be the same as videoNode.asset. If the user observed anything in the AVAsset before construction or accessed anything inside videoNode.asset everything would work.

Looking at it now however if the user constructs with an AVURLAsset the asset is dropped. I assume the AVPlayerItem makes a new asset if you construct with the raw URL, so videoNode.playerItem.asset would no longer be the same object as videoNode.asset.

The .asset property is also used throughout the class, i.e.

```
- (void)generatePlaceholderImage
{
  ASVideoNode * __weak weakSelf = self;
  AVAsset * __weak asset = self.asset;

  [self imageAtTime:kCMTimeZero completionHandler:^(UIImage *image) {
    ASPerformBlockOnMainThread(^{
      // Ensure the asset hasn't changed since the image request was made
      if (ASAssetIsEqual(weakSelf.asset, asset)) {
[...]

PR #1607, opened 2016-04-29
[ASVideoNode] Prepare asset for playback in a uniform way

chourobin:
Following up with issue #1600, the following PR modifies `-fetchData` so that we can prepare an asset (HLS and non-HLS) for playback in a uniform way.

Namely:
1. Calls [`-loadValuesAsynchronouslyForKeys:completionHandler:`](https://developer.apple.com/library/ios/documentation/AVFoundation/Reference/AVAsynchronousKeyValueLoading_Protocol/index.html#//apple_ref/occ/intfm/AVAsynchronousKeyValueLoading/loadValuesAsynchronouslyForKeys:completionHandler:) in `-fetchData`.
2. Creates the `AVPlayerItem` and `AVPlayer` in `-prepareToPlayAsset:withKeys:`.
3. No longer extracts the URL from an AVAsset.

In doing so, the AVPlayer is not immediately available when `-fetchData` is called. Thanks and let me know your thoughts :)

@gazreese @ejensen

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -106,7 +106,7 @@ static NSString * const kStatus = @"status";
   return nil;
 }
 
-- (void)prepareToPlayAsset:(AVAsset *)asset withKeys:(NSArray *)requestedKeys
+- (void)prepareToPlayAsset:(AVAsset *)asset withKeys:(NSArray<NSString *> *)requestedKeys
 {
   for (NSString *key in requestedKeys) {
     NSError *error = nil;
@@ -122,7 +122,7 @@ static NSString * const kStatus = @"status";
   }
     
   AVPlayerItem *playerItem = [self constructPlayerItem];
-  self.currentItem = playerItem;
+  [self setCurrentItem:playerItem];
   
   if (_player != nil) {
     [_player replaceCurrentItemWithPlayerItem:playerItem];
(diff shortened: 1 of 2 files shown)


appleguy:
@chourobin great, thanks for the change!  I've pinged 4 engineers to help take a look at this, once one of them is able to review & approve I will merge.

Xinchi:
@chourobin LGTM!
Thanks for the Pull Request and adding the prepare-asset phase! This is exactly what has been done on the Pinterest app application level, and would be great to move these logic down to ASVideoNode. 
One concern here: As we replace the current player item and register new observers, is it necessary to remove all the old observers?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -264,17 +289,13 @@ - (void)fetchData
...
-      self.player = [AVPlayer playerWithPlayerItem:playerItem];
-    }
-
-    if (_placeholderImageNode.image == nil) {
-      [self generatePlaceholderImage];
+    if (_asset != nil) {
+      NSArray *requestedKeys = @[ @"playable" ];
+      [_asset loadValuesAsynchronouslyForKeys:requestedKeys completionHandler:^{
+          dispatch_async(dispatch_get_main_queue(), ^{
+            [self prepareToPlayAsset:_asset withKeys:requestedKeys];
```
ejensen: A check may be needed to ensure _asset hasn't changed in the time between this request was made, and the asynchronous completionHandler is invoked.

chourobin:
@Xinchi yes it is, and currently it's being done on `-setCurrentItem:` line 540:

```
- (void)setCurrentItem:(AVPlayerItem *)currentItem
{
  ASDN::MutexLocker l(_videoLock);

  [self removePlayerItemObservers:_currentPlayerItem];

  _currentPlayerItem = currentItem;

  [self addPlayerItemObservers:currentItem];
}
```

AsyncDisplayKit/ASVideoNode.mm
chourobin: Agreed and thanks for the catch! Will try to add it in tonight.

appleguy:
@chourobin thanks for this quite large and very valuable contribution!  Hope to learn more about how you're using the framework and see how we can support one another's work :)

PR #1608, opened 2016-04-29
[ASVideoNode] Several new delegate methods and ASVideoNodePlayerState to monitor and control playback

Eke:
Added new delegate methods : 

##### `videoNode:willChangePlayerState:toState:`
This method is called after each state change. at this moment this states are supported: 
````
typedef enum {
  ASVideoNodePlayerStateUnknown,
  ASVideoNodePlayerStatePlaying,
  ASVideoNodePlayerStateLoading,
  ASVideoNodePlayerStatePaused,
  ASVideoNodePlayerStateFinished
} ASVideoNodePlayerState
````
##### `videoNode:shouldChangePlayerStateTo:`
This method is called before play and pause actions. asks delegate if play and pause are allowed states for ASVideoNode. For example you can disable pause method for some videos.

##### `videoNode:didPlayToSecond:`
This method i called from `periodicTimeObserver` and informs delegate about ASVideoNode playback current time. Can be used for scrubber UI or ads in video. For example you can pause video on 10 second and show some ads. you can control `periodicTimeObserver` timescale by setting `periodicTimeObserverTimescale`

##### ASVideoNodePlayerStatePlaying
Is set from `play` method, when player item status is `AVPlayerItemStatusReadyToPlay` and `_shouldBePlaying` is true.
##### ASVideoNodePlayerStateLoading
Is set when player item status is not equal to `AVPlayerItemStatusReadyToPlay` and spinner is shown
##### ASVideoNodePlayerStatePaused
Is set when `pause` method is called
##### ASVideoNodePlayerStateFinished
Is set when `didPlayToEnd` notification is fired

Will be happy to hear your comments

UPD: 
- added `playbackBufferEmpty` observer to show spinner when buffering; 
- moved show/hide spinner to separate methods

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.h
+++ b/AsyncDisplayKit/ASVideoNode.h
@@ -12,6 +12,14 @@
 @class AVAsset, AVPlayer, AVPlayerItem;
 @protocol ASVideoNodeDelegate;
 
+typedef enum {
+  ASVideoNodePlayerStateUnknown,
+  ASVideoNodePlayerStatePlaying,
+  ASVideoNodePlayerStateLoading,
+  ASVideoNodePlayerStatePaused,
+  ASVideoNodePlayerStateFinished
+} ASVideoNodePlayerState;
+
 NS_ASSUME_NONNULL_BEGIN
 
 // IMPORTANT NOTES:
@@ -40,6 +48,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, assign, readwrite) BOOL muted;
 
+@property (nonatomic, assign, readonly) ASVideoNodePlayerState playerState;
+//! Defaults to 100
+@property (nonatomic, assign, readwrite) int32_t periodicTimeObserverTimescale;
+
 //! Defaults to AVLayerVideoGravityResizeAspect
 @property (atomic) NSString *gravity;
 
@@ -63,6 +75,31 @@ NS_ASSUME_NONNULL_BEGIN
  * @discussion The video's play state is toggled if this method is not implemented.
  */
 - (void)videoNodeWasTapped:(ASVideoNode *)videoNode;
+/**
+ * @abstract Delegate method invoked when player changes state.
+ * @param videoNode The video node that was tapped.
+ * @param state player state before this change.
+ * @param toSate player new state.
+ * @discussion This method is called after each state change
+ */
+- (void)videoNode:(ASVideoNode *)videoNode willChangePlayerState:(ASVideoNodePlayerState)state toState:(ASVideoNodePlayerState)toSate;
+/**
+ * @abstract Ssks delegate if state change is allowed
+ * ASVideoNodePlayerStatePlaying or ASVideoNodePlayerStatePaused.
+ * asks delegate if state change is allowed.
+ * @param videoNode The video node that was tapped.
+ * @param state player state that is going to be set.
+ * @discussion Delegate method invoked when player changes it's state to
+ * ASVideoNodePlayerStatePlaying or ASVideoNodePlayerStatePaused 
+ * and asks delegate if state change is valid
+ */
+- (BOOL)videoNode:(ASVideoNode*)videoNode shouldChangePlayerStateTo:(ASVideoNodePlayerState)state;
+/**
+ * @abstract Delegate method invoked when player playback time is updated.
+ * @param videoNode The video node that was tapped.
+ * @param second current playback time in seconds.
+ */
+- (void)videoNode:(ASVideoNode *)videoNode didPlayToSecond:(NSTimeInterval)second;
 @end
 NS_ASSUME_NONNULL_END
 #endif
--- a/AsyncDisplayKit/ASVideoNode.h
+++ b/AsyncDisplayKit/ASVideoNode.h
@@ -50,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, assign, readonly) ASVideoNodePlayerState playerState;
 //! Defaults to 100
-@property (nonatomic, assign, readwrite) int32_t periodicTimeObserverTimescale;
+@property (nonatomic, assign) int32_t periodicTimeObserverTimescale;
 
 //! Defaults to AVLayerVideoGravityResizeAspect
 @property (atomic) NSString *gravity;
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -101,7 +101,7 @@ static NSString * const kStatus = @"status";
 - (ASDisplayNode *)constructPlayerNode
 {
   ASVideoNode * __weak weakSelf = self;
-  
+
   return [[ASDisplayNode alloc] initWithLayerBlock:^CALayer *{
     AVPlayerLayer *playerLayer = [[AVPlayerLayer alloc] init];
     playerLayer.player = weakSelf.player;
@@ -113,7 +113,7 @@ static NSString * const kStatus = @"status";
 - (AVPlayerItem *)constructPlayerItem
 {
   ASDN::MutexLocker l(_videoLock);
-  
+
   if (_asset != nil) {
     return [[AVPlayerItem alloc] initWithAsset:_asset];
   }
@@ -135,7 +135,7 @@ static NSString * const kStatus = @"status";
     NSLog(@"Asset is not playable.");
     return;
   }
-  
+
   AVPlayerItem *playerItem = [self constructPlayerItem];
   [self setCurrentItem:playerItem];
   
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: AsyncDisplayKit/ASVideoNode.h [hunk 1]
```diff
@@ -40,6 +48,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, assign, readwrite) BOOL muted;
 
+@property (nonatomic, assign, readonly) ASVideoNodePlayerState playerState;
+//! Defaults to 100
+@property (nonatomic, assign, readwrite) int32_t periodicTimeObserverTimescale;
```
Xinchi: nit: readwrite is the default value, don't need to specify here

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -114,6 +122,7 @@ - (void)addPlayerItemObservers:(AVPlayerItem *)playerItem
 {
   [playerItem addObserver:self forKeyPath:kStatus options:NSKeyValueObservingOptionInitial | NSKeyValueObservingOptionNew context:ASVideoNodeContext];
   [playerItem addObserver:self forKeyPath:kPlaybackLikelyToKeepUpKey options:NSKeyValueObservingOptionNew context:ASVideoNodeContext];
+  [playerItem addObserver:self forKeyPath:kplaybackBufferEmpty options:NSKeyValueObservingOptionNew context:ASVideoNodeContext];
```
Xinchi: Very nice! Thanks for adding this

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -229,18 +239,23 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
...
+        self.playerState = ASVideoNodePlayerStatePlaying;
+      }
     }
   } else if ([keyPath isEqualToString:kPlaybackLikelyToKeepUpKey]) {
     if (_shouldBePlaying && [change[NSKeyValueChangeNewKey] boolValue] == true && ASInterfaceStateIncludesVisible(self.interfaceState)) {
       [self play]; // autoresume after buffer catches up
     }
+  } else if ([keyPath isEqualToString:kplaybackBufferEmpty]) {
+    if (_shouldBePlaying && [change[NSKeyValueChangeNewKey] boolValue] == true && ASInterfaceStateIncludesVisible(self.interfaceState)) {
+      [self showSpinner]; // show spinner and set ASVideoNodePlayerStateLoading
```
Xinchi: Change the player state here?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -415,26 +469,18 @@ - (void)play
...
-        spinnnerView.color = [UIColor whiteColor];
-
-        return spinnnerView;
-      }];
-
-      [self addSubnode:_spinner];
-    }
-
-    [(UIActivityIndicatorView *)_spinner.view startAnimating];
+    [self showSpinner];
```
Xinchi: Very nice moving the blob to a separate method!

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -415,26 +469,18 @@ - (void)play
...
+  }else{
+    [self removeSpinner];
+    self.playerState = ASVideoNodePlayerStatePlaying;
```
Xinchi: Should we change the state to 'playing' if [self ready] is false?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -462,11 +540,22 @@ - (BOOL)isPlaying
   return (_player.rate > 0 && !_player.error);
 }
 
+- (BOOL)isStateChangeValid:(ASVideoNodePlayerState)state
```
Xinchi: In stead of checking the delegate respondsToSelector each time we play and pause, can we store this variable away in a property, something like 
_asyncDelegateFlags.asyncDelegateScrollViewDidScroll
Check out ASCollectionView.mm :)

Eke: Hmmm, nice one : )) maybe i should update all delegate methods like this?

Eke:
@Xinchi  did everything mentioned in your comment, thanks for review ! please take a look at delegate setter and getter, not sure if 100% correct

Xinchi:
@Eke They look good to me! Thanks for fixing them! @appleguy is it possible to merge this pull request after your review?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -420,41 +495,37 @@ - (void)setMuted:(BOOL)muted
...
-
-        return spinnnerView;
-      }];
-
-      [self addSubnode:_spinner];
-    }
-
-    [(UIActivityIndicatorView *)_spinner.view startAnimating];
+    [self showSpinner];
+  }else{
```
maicki: Add whitespace between `} else {`

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -482,16 +584,27 @@ - (BOOL)isPlaying
   return (_player.rate > 0 && !_player.error);
 }
 
+- (BOOL)isStateChangeValid:(ASVideoNodePlayerState)state
+{
+  if(_delegateFlags.delegateVideNodeShouldChangePlayerStateTo){
```
maicki: Whitepspaces after `if`

examples/Videos/Sample/ViewController.m
```diff
@@ -12,17 +12,16 @@
 #import "ViewController.h"
 
 @interface ViewController()<ASVideoNodeDelegate>
-@property (nonatomic) ASVideoNode *videoNode;
+@property (nonatomic,strong) ASVideoNode *guitarVideoNode;
```
maicki: Whitespace after `,`

examples/Videos/Sample/ViewController.m
```diff
@@ -94,17 +101,51 @@ - (ASButtonNode *)playButton;
...
   playButton.position = CGPointMake([UIScreen mainScreen].bounds.size.width/4, ([UIScreen mainScreen].bounds.size.height/3)/2);
   [playButton setImage:[UIImage imageNamed:@"playButtonSelected@2x.png"] forState:ASControlStateHighlighted];
-
+  
   return playButton;
 }
 
 - (void)videoNodeWasTapped:(ASVideoNode *)videoNode
 {
+  if(videoNode == self.guitarVideoNode){
```
maicki: Whitespace after `if` and `if else`

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -304,11 +344,11 @@ - (void)clearFetchedData
...
-
-    self.player = nil;
-    self.currentItem = nil;
-    _placeholderImageNode.image = nil;
+  ASDN::MutexLocker l(_videoLock);
+  
+  self.player = nil;
+  self.currentItem = nil;
+  _placeholderImageNode.image = nil;
   }
```
maicki: Can you please indent the lines fro that scope

maicki:
@Eke @Xinchi Added some smaller comments but overall looks fine

Xinchi:
Hi @Eke , can you give some use cases of ASVideoNodePlayerState?  In general we should avoid making breaking API changes in such a large framework,  I am thinking maybe it's a better idea to separate the API change from other changes, including adding the periodic observers..etc.   Would you mind sending a separate PR for adding the observers for now?

appleguy:
@Eke Thanks for this very awesome set of extensions!  We actually need some of these features in Pinterest and are merging with intent to sync up to master and start using it right away :).

Main feedback: it does not make sense to return NO for should enter player state Finished.  So I worry that it is possible that the state is not sufficiently broken up into logically distinct categories that integrate with the delegate methods in a fully consistent way.  I won't have time to think about a better solution soon, and it's not urgent to change, but I do worry about making breaking API changes here once we get around to a refinement so the sooner we think of an improvement, the better!  2.0 is a good time to make the breaking changes of this sort though so we have at least a month, don't rush on your side.

Eke:
@appleguy user can't return NO for player state Finished, because `videoNode:shouldChangePlayerStateTo` method is not event called for this state, it is called only for play/pause action at this moment. I will try to find some better solution for this case. We can discuss this in slack channel with other people too.

Hello again @Xinchi  ! Thanks for your help! I

Will try to write one case here, but can make some example application too.

For example, we have some kind of video content app where we have pre-roll Ads and we don't want user to be able to pause ad. In this case we can implement `videoNode:shouldChangePlayerStateTo` and return NO for ASVideoNodePlayerStatePaused change. Or for example we don't want user to play video if another video is playing at this moment or something like this, there can be a lot of cases to use `videoNode:shouldChangePlayerStateTo`. I am going to make one big `ASDKTube` example and will try to use this cases there.

PR #1618, opened 2016-05-03
[ASDisplayNode] Fix rare deadlock when performing layout on multiple threads with the same node.

garrettmoon:


appleguy:
@garrettmoon could you include a full backtrace of the locked threads, both for documentation and reviewers' understanding?  Thanks a ton for investigating, then fixing this!

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -125,6 +125,8 @@ - (void)__setNeedsLayout
 {
   CGSize oldSize = self.calculatedSize;
   [super __setNeedsLayout];
+  
+  ASDN::MutexLocker l(_propertyLock);
```
nguyenhuy: I'm not sure why we need to lock here. `calculatedSize` setter already acquires the lock internally.

garrettmoon: It's for the didRelayoutFromOldSize call. I'm not sure it's necessary, it's just that a lock was held previously when __setNeedsLayout was called so I didn't want to change anything. I can add a comment to this effect?

nguyenhuy: Yes, please add. Thank you for working on this!

garrettmoon:
screen shot 2016-05-03 at 9 26 58 am
screen shot 2016-05-03 at 9 27 12 am

PR #1619, opened 2016-05-04
[ASRangeController] Update Synchronously More Often

Adlai-Holler:
First off, I know how critical this code is, so I've tried to minimize changes and especially additions.

### Big Changes
- Perform update immediately in `visibleNodeIndexPathsDidChangeWithScrollDirection` and `updateRangeWithMode`. This is the meat of the PR. 
  - The comment at ASRangeController.mm:98 claimed this was already executed immediately, but it wasn't.
  - Unfortunately this means that we will perform a range update for each call to `willDisplayCell:` rather than max once per run loop. Is this an acceptable cost for the guarantee that a cell will not be visible before the ranges are updated? I believe so.
- Replace `scheduleRangeUpdate` and `performRangeUpdate` with `performRangeUpdateSynchronously:`.
- In ASTableView, change how we compute the visible index paths. Once I started updating synchronously out of `willDisplayCell:` UITableView started returning `nil` from `indexPathForCell:` even though the cells were retrieved through `visibleCells`. The new method is more direct and is still aware of the grouped table view issue that the old one was working around.

### Small Changes

- Remove extraneous calls to `performRangeUpdate` in the +ASRangeController application lifecycle methods, because changing the range mode already implies an update. Updated documentation for updateCurrentRangeWithMode: to reflect that fact.
- Stop checking `_queuedRangeUpdate = YES` during the range update. The queuing mechanism lives outside the updating mechanism, and previously we had resorted to simply setting that flag before updating which is silly.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -272,6 +272,10 @@
 {
   [super visibleStateDidChange:isVisible];
   
+  if (isVisible && self.neverShowPlaceholders) {
+    [self recursivelyEnsureDisplaySynchronously:YES];
+  }
+  
   // NOTE: This assertion is failing in some apps and will be enabled soon.
   // ASDisplayNodeAssert(self.isNodeLoaded, @"Node should be loaded in order for it to become visible or invisible.  If not in this situation, we shouldn't trigger creating the view.");
   UIView *view = self.view;
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -23,6 +23,7 @@
 #import "ASLayout.h"
 #import "_ASDisplayLayer.h"
 #import "ASTableNode.h"
+#import "ASEqualityHelpers.h"
 
 static const ASSizeRange kInvalidSizeRange = {CGSizeZero, CGSizeZero};
 static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
@@ -126,6 +127,7 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   BOOL _ignoreNodesConstrainedWidthChange;
   BOOL _queuedNodeHeightUpdate;
   BOOL _isDeallocating;
+  BOOL _performingBatchUpdates;
   NSMutableSet *_cellsForVisibilityUpdates;
   
   struct {
@@ -468,6 +470,7 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   
   // To ensure _nodesConstrainedWidth is up-to-date for every usage, this call to super must be done last
   [super layoutSubviews];
+  [_rangeController updateIfNeeded];
 }
 
 #pragma mark -
@@ -644,35 +647,29 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
     [_asyncDelegate tableView:self willDisplayNodeForRowAtIndexPath:indexPath];
   }
   
-  [_rangeController visibleNodeIndexPathsDidChangeWithScrollDirection:[self scrollDirection]];
-
-  if (cellNode.neverShowPlaceholders) {
-    [cellNode recursivelyEnsureDisplaySynchronously:YES];
-  }
+  [_rangeController setNeedsUpdate];
   
   if (ASSubclassOverridesSelector([ASCellNode class], [cellNode class], @selector(cellNodeVisibilityEvent:inScrollView:withCellFrame:))) {
     [_cellsForVisibilityUpdates addObject:cell];
   }
 }
 
-- (void)tableView:(UITableView *)tableView didEndDisplayingCell:(_ASTableViewCell *)cell forRowAtIndexPath:(NSIndexPath*)indexPath
+- (void)tableView:(UITableView *)tableView didEndDisplayingCell:(_ASTableViewCell *)cell forRowAtIndexPath:(NSIndexPath *)indexPath
 {
-  if ([_pendingVisibleIndexPath isEqual:indexPath]) {
+  if (ASObjectIsEqual(_pendingVisibleIndexPath, indexPath)) {
     _pendingVisibleIndexPath = nil;
   }
   
   ASCellNode *cellNode = [cell node];
 
-  [_rangeController visibleNodeIndexPathsDidChangeWithScrollDirection:[self scrollDirection]];
+  [_rangeController setNeedsUpdate];
 
   if (_asyncDelegateFlags.asyncDelegateTableViewDidEndDisplayingNodeForRowAtIndexPath) {
     ASDisplayNodeAssertNotNil(cellNode, @"Expected node associated with removed cell not to be nil.");
     [_asyncDelegate tableView:self didEndDisplayingNode:cellNode forRowAtIndexPath:indexPath];
   }
 
-  if ([_cellsForVisibilityUpdates containsObject:cell]) {
-    [_cellsForVisibilityUpdates removeObject:cell];
-  }
+  [_cellsForVisibilityUpdates removeObject:cell];
 
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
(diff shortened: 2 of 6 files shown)


Adlai-Holler:
@appleguy @nguyenhuy Thoughts?

appleguy:
This is definitely a very sensitive to change, but also a very important one that we have been intending to make with high priority. As with several such things, no one has been available to work on it in the foreseeable future, so I appreciate you picking up!

I know that Michael Schneider is familiar with some of the timing considerations in this code, which are not intuitive and affect only certain editing patterns or situation such as memory warnings for freeing of memory upon suspend of the app. There're also a few special considerations with changing the range mode.

I will review this as soon as I can, but unfortunately this Will not be for at least a few days, and it would be helpful to see if anyone else in the community has initial feedback.

Adlai-Holler:
OK understood. I definitely don't want to rush this through but I believe it's a critical stepping stone toward getting the best of both worlds with ASDK – guarantees about initial states + smooth lookahead/concurrent rendering.

@maicki If you've got bandwidth, let's work together on this. We'll iterate a little more before @appleguy jumps in.

appleguy:
@maicki I know you have been _very_ busy, for good reasons, in **addition** to being overseas with family! :)  When you have another workday - Monday - let's prioritize getting a round of feedback to Adlai here.  I won't have time due to needing to begin packing for Vietnam, and you are by far the best person on the team to review this (I have a similar level of understanding of this code but don't want to block progress).  Let me know on iMessage if you have any questions or concerns and want to check if I have the same impression about how the existing code works.

maicki:
@Adlai-Holler Overall looks good to me!

I'm just a bit worried about the performance impact we will hit if we update more synchronously instead of batching it once per run loop. Did you do any testing around it?

Would also be interesting to understand more why `self.visibleCells -> [self indexPathForCell:]` will return nil if called synchronously from `-willDisplayCell:forRowAtIndexPath`. Could there be other side effects we could ran into if we synchronously update?

Adlai-Holler:
Great points @maicki! With them in mind, I'm thinking about how to improve this approach. I'm just going to store some thoughts here as I attack this.

### The Goal

We want to update the range controller **synchronously**, just **after** each **batch** of visible node changes is committed.
- _After_ because during `willDisplayCell:` the cell isn't quite in the hierarchy yet, so unexpected behavior (like the indexPathForCell: failing) may occur if we pretend otherwise.
- _Synchronous_ because we want to guarantee that the range state is correct before the next frame is drawn.
- _Batch_ because it's potentially expensive to recalculate the ranges more than once per layout change. **Q** I haven't profiled this yet – it would be good to know the relative cost.

### How UITableView/UICollectionView Change Visible Index Paths

There are three reasons for a visible index paths change:
1. bounds change (scrolling, resizing). 
  - This case is pretty easy to cover. Table view and collection view always perform visible cell changes during `layoutSubviews`.
  - We can simply add a flag like `visibleIndexPathsDidChangeDuringLayoutSubviews`, and check it after `[super layoutSubviews]`.
  - If it's changed, then we tell the range controller to update.
1. A batch of editing operations that intersects the visible area.
1. A single editing operation that intersects the visible area.

### Our Hooks

- `willDisplayCell:` This is called just before a cell is assigned an index path and added into the view. Note that the cell may already be in the view (for some previous item) or it may not.
- `didEndDisplayingCell:` This is called just after a cell is unassigned its index path. Note that the cell may be removed from the view or it may not.
- `super layoutSubviews:` Useful override point.
- `super performBatchUpdates:` Useful override point. UICollectionView synchronously updates visible cells during this, including calls to `willDisplay:` and `didEndDisplaying:`
- `super insert/delete/move/reload:` We call these and I suspect **Q** that collection view & table view synchronously update the visible index paths if these are called outside a batch.


### Other Case: Update Animation End

When a table view (**Q** also collection view?) update animation ends, we may get a `didEndDisplayingCell:` call. This can't be predicted and it's not tied into the UIKit layout passes. Unfortunately, the only case where we get a update animation completion hook is in the UICollectionView batch update case. 

We could expand that hook to cover all updates, if we always wrap isolated update calls (e.g. `insertItemsAtIndexPaths:`) in batch updates but that still leaves us out in the cold for table view updates. As far as I know, there is no table view update animation completion hook.

I'm going to keep playing with this and mulling it over, because it's so valuable and yet so delicate. Feedback/thoughts much appreciated @maicki @appleguy @nguyenhuy

AsyncDisplayKit/ASTableView.mm
```diff
@@ -812,19 +812,25 @@ - (NSArray *)visibleNodeIndexPathsForRangeController:(ASRangeController *)rangeC
...
+  
+  NSMutableArray *visibleIndexPaths = [NSMutableArray arrayWithCapacity:unfilteredIndexPaths.count];
+  for (NSIndexPath *indexPath in unfilteredIndexPaths) {
+    CGRect rowRect = [self rectForRowAtIndexPath:indexPath];
+    if (CGRectIntersectsRect(rowRect, bounds)) {
+      [visibleIndexPaths addObject:indexPath];
+    }
   }
   
   if (_pendingVisibleIndexPath) {
```
appleguy: @Adlai-Holler do you think we could get rid of some of this extra logic?  It has a questionable history and has caused some bugs as well as fixing some past issues.  I think the original cause of the bug it addressed may be fixed by the extensive improvements to data and range controller

AsyncDisplayKit/Details/ASFlowLayoutController.mm
```diff
@@ -90,10 +90,9 @@ - (NSSet *)indexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(
     currPath.row++;
 
     // Once we reach the end of the section, advance to the next one.  Keep advancing if the next section is zero-sized.
-    while (currPath.row >= [(NSArray *)completedNodes[currPath.section] count] && currPath.section < completedNodes.count - 1) {
+    while (currPath.row >= [(NSArray *)completedNodes[currPath.section] count] && currPath.section < endPath.section) {
```
appleguy: Was this a bug or just a readability improvement?  Looks like a possible bug?

AsyncDisplayKit/Details/ASFlowLayoutController.mm
```diff
@@ -90,10 +90,9 @@ - (NSSet *)indexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(
...
       currPath.row = 0;
       currPath.section++;
-      ASDisplayNodeAssert(currPath.section <= endPath.section, @"currPath should never reach a further section than endPath");
```
appleguy: Why remove this?  Esp. if it was added after the original authoring of the code, I think I may have added it when encountering issues from ASTableViewStressTest etc

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -105,53 +105,44 @@ - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
...
 }
 
-- (void)scheduleRangeUpdate
+- (void)performRangeUpdateSynchronously:(BOOL)updateSynchronously
 {
-  if (_queuedRangeUpdate) {
-    return;
+  if (updateSynchronously) {
+    ASDisplayNodeAssertMainThread();
+    [self _updateVisibleNodeIndexPaths];
```
appleguy: In this case, should we set a flag that causes any already-scheduled queued range update to be cancelled?  E.g. something that is checked inside the block below.  Maybe as simple as an if (_queuedRangeUpdate) added inside the block, and setting that flag to NO here.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -105,53 +105,44 @@ - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
...
-    [self performRangeUpdate];
+    [self performRangeUpdateSynchronously:YES];
   }
 }
 
 - (void)_updateVisibleNodeIndexPaths
 {
   ASDisplayNodeAssert(_layoutController, @"An ASLayoutController is required by ASRangeController");
-  if (!_queuedRangeUpdate || !_layoutController || !_dataSource) {
+  if (!_layoutController || !_dataSource) {
```
appleguy: I'm not sure if this is safe or at least efficient.  At least in the case above, a previously scheduled update could run even after a full update pass is done by a sync. call.  These types of multiple calls are actually very common when processing edit transactions etc.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -520,14 +510,11 @@ + (void)didEnterBackground:(NSNotification *)notification
...
     BOOL isVisible = ASInterfaceStateIncludesVisible([rangeController interfaceState]);
     [rangeController updateCurrentRangeWithMode:isVisible ? ASLayoutRangeModeVisibleOnly : ASLayoutRangeModeLowMemory];
+    // N.b. we intentionally don't call `performRangeUpdate` because `updateCurrentRangeWithMode` does so if needed.
   }
   
   // Because -interfaceState checks __ApplicationState and always clears the "visible" bit if Backgrounded, we must set this after updating the range mode.
   __ApplicationState = UIApplicationStateBackground;
-  for (ASRangeController *rangeController in allRangeControllers) {
-    // Trigger a range update immediately, as we may not be allowed by the system to run the update block scheduled by changing range mode.
-    [rangeController performRangeUpdate];
```
appleguy: I am nearly certain this is necessary because the value of __ApplicationState must be updated / changes the behavior of the range update

AsyncDisplayKit/ASTableView.mm
Adlai-Holler: I'd like simply to call `indexPathsForVisibleRows` and be done with it, but the prior comment claimed that grouped-style table views would sometimes return index paths for rows that are not visible.

So I see three viable options here:
- Simply do that. Call indexPathsForVisibleRows and, if grouped-style tables return us extra rows, accept it. Easy, quick, and grouped-style tables typically don't have extremely long/rich content so I doubt the possible extra rows would be a problem in practice.
- Only apply the filtering logic for grouped-style tables. This makes the code even MORE complex, but removes the performance hit in the much more common case of plain-style table views.
- Add a unit test to see if we can reproduce the extra rows in the grouped-style case. Seems like a pain in the ass, and almost certainly not worthwhile. Here for rigor's sake.

I'm going to implement option 1. If you disagree obviously we can discuss it more later.

Adlai-Holler: @appleguy I've addressed it in 2c76112. Take a looky-loo.

AsyncDisplayKit/Details/ASFlowLayoutController.mm
Adlai-Holler: It was a bug, I was able to get it to fail the assertion below. So my rationale was "why have an overbroad condition and then assert against it, rather than setting the condition you want directly?" It serves as a sanity check, I know, but it also seemed silly.

AsyncDisplayKit/Details/ASFlowLayoutController.mm
Adlai-Holler: See the comment above. Now that the condition explicitly specifies that iteration should stop at endPath.section - 1 the assertion could never be hit.

appleguy:
Thanks for the updates!

AsyncDisplayKit/Details/ASRangeController.mm
Adlai-Holler: Great point! Done!

AsyncDisplayKit/Details/ASRangeController.mm
Adlai-Holler: I believe this is handled in 14842b5 . The only code path that reaches `_updateVisibleNodeIndexPaths` is the queueing method `performRangeUpdateSynchronously:` and it now correctly checks/sets the queued flag.

Adlai-Holler:
@appleguy Happy to do it! Feels good to be back in this code, especially with the optimization/simplification I was able to pull off for `-[ASTableView visibleNodeIndexPathsForRangeController:]`

AsyncDisplayKit/Details/ASRangeController.mm
Adlai-Holler: True. Dialed in!

Adlai-Holler:
@appleguy OK I think this one's ready for another review!

Adlai-Holler:
This PR got pretty huge =/. Here's what ended up happening:

- Remove `ASRangeController.scrollDirection` and replace with new data source method `scrollDirectionForRangeController:`
- ASRangeController now exposes 2 update methods `setNeedsUpdate` and `updateIfNeeded`
- setNeedsUpdate sets a flag and (if needed) enqueues an `updateIfNeeded` call onto the main queue
- updateIfNeeded clears the flag and performs the update
- Table and collection view call setNeedsUpdate in willDisplayCell:/didDisplayCell:
- Table and collection view call `updateIfNeeded` after `-layoutSubviews`, after submitting data update operations, and after update animations end (only supported for collection view.)
- Optimize `ASTableView.indexPathsForVisibleNodes:` significantly
- Add conditional code to ASRangeController to log frames that contain multiple range updates.
  - Currently during scrolling we never see this, but we sometimes see 2-5 updates during the same frame when update operations are occurring or when table/collections appear.
  - The code in `-[Collection/Table didMoveToWindow:]` needs scrutiny. Are both calls necessary?
- Move enforcement of cell node neverShowPlaceholders into the cell node's visibility callback, out from willDisplayCell:. This way `ensureDisplay` is called after the cell is already added to the hierarchy, and it also reduces class interdependency.

Notes:
- In -[ASCollectionView layoutSubviews] I had to gate the -updateIfNeeded behind `if (self.window != nil)`. If this isn't present, the client app crashes (~80% reliable) with a collection view data integrity problem under certain conditions. It's a very weird issue, I dug into it a little and bailed because we'll flush the range update in didMoveToWindow anyway.

Questions:
- ASRangeController has a `_rangeIsValid` flag that seems to trigger a re-evaluation of the range states of all nodes when the collection view data is updated. Is that correct? I'm thinking we rename it to avoid confusion between `_rangeIsValid` and the new `_needsRangeUpdate` flag.

This has been tested thoroughly but it's obviously a very significant change to very important code. Sorry for the burden, but I think it'll be worth it! Feel free to meet with me in person as needed. Need review @appleguy @maicki @nguyenhuy

Adlai-Holler:
BTW: I've got a branch that removes the `_pendingVisibleIndexPath` feature from ASTableView. I'd like to remove it but I don't know exactly why it was added. By the time we perform the range update, the row should have been added in and be ready to rock (either in the next run loop or manually after -layoutSubviews or after submitting the edit operation).

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -225,6 +225,10 @@ - (void)visibleStateDidChange:(BOOL)isVisible
 {
   [super visibleStateDidChange:isVisible];
   
+  if (isVisible && self.neverShowPlaceholders) {
+    [self recursivelyEnsureDisplaySynchronously:YES];
```
Adlai-Holler: @appleguy I moved this call. It used to live in the separate `willDisplayCell:` callback methods but I thought this a more appropriate place because in the previous case, the cell node wouldn't be in the hierarchy yet. So at the very least, this change gives us a tiny bit more time before we block the main thread. Thoughts?

appleguy:
@Adlai-Holler sorry I haven't been able to review this yet - if this is blocked on me, could you let me know & also send me a list of any other PRs that you'd say are blocked on me?

@eanagel has the background on the pending index path.  Unfortunately it was very necessary and we were able to prove that UITableView does not necessarily include information that we expected it should be guaranteed to.  I looked into the history around this in detail about a month ago, it can be found by seeking back through various PRs and related issues.  It would be fantastic to eliminate.

@eanagel if you have any mechanism for us to test that original issue - I think you had a test project, or at least ability to reproduce it in your app - that would help.  Running your app against this PR would in fact be of phenomenal value to the community.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -850,61 +847,38 @@ - (NSArray *)visibleNodeIndexPathsForRangeController:(ASRangeController *)rangeC
...
-    }
+  NSIndexPath *pendingVisibleIndexPath = _pendingVisibleIndexPath;
+  if (pendingVisibleIndexPath == nil) {
+    return self.indexPathsForVisibleRows;
   }
   
+  NSMutableArray *visibleIndexPaths = [self.indexPathsForVisibleRows mutableCopy];
+  [visibleIndexPaths sortUsingSelector:@selector(compare:)];
+
+  BOOL isPendingIndexPathVisible = (NSNotFound != [visibleIndexPaths indexOfObject:pendingVisibleIndexPath inSortedRange:NSMakeRange(0, visibleIndexPaths.count) options:kNilOptions usingComparator:^(id  _Nonnull obj1, id  _Nonnull obj2) {
```
maicki: You know of any overhead with the block api?

maicki:
@Adlai-Holler One comment but otherwise looks good

AsyncDisplayKit/ASTableView.mm
Adlai-Holler: It's true there is some overhead, but if we remove the block then we have to do a linear search rather than a binary search which means more comparisons =(. So when we have a lot of rows, the block-based approach is faster but when we have fewer rows, a linear search is faster.  For now I will leave this here, since the number of block invocations is log(numVisibleRows) so the difference is tiny and binary search is so nice.

Issue #1614, opened 2016-05-03
Broken link


sushilthe:
The link to http://asyncdisplaykit.org/docs/automatic-layout.html referenced from http://asyncdisplaykit.org/docs/philosophy.html is broken. The link is under the **Asynchronous Performance Gains Section** and can be visited by clicking the **correctly** link.

sushilthe:
The link to http://asyncdisplaykit.org/docs/automatic-layout.html referenced from http://asyncdisplaykit.org/docs/philosophy.html is broken. The link is under the **Asynchronous Performance Gains Section** and can be visited by clicking the **examples/ASDKgram** text.

sushilthe:
The link to http://asyncdisplaykit.org/docs/automatic-layout.html referenced from http://asyncdisplaykit.org/docs/philosophy.html is broken. The link is under the **A Great App Experience for All Users** sub-section **A Great Developer Experience** and can be visited by clicking the **examples/ASDKgram** link.

maicki:
cc @hannahmbanana

PR #1621, opened 2016-05-04
Fixed links on philosophy page

lappp9:


maicki:
This should fix: #1614

PR #1622, opened 2016-05-04
[ASDisplayNode] Improve Transition ID handling

maicki:
Currently in a race condition situation it can happen that subnodes get transition ids from supernodes where the supernodes transition id was reset but the subnode still has the old transition id value. This will trigger an exception although the assignment of the transition id is totally valid. This PR should address this by not reseting the transition id at all for a node and increase the transition id for new transitions.

cc @nguyenhuy @appleguy @levi

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -373,7 +373,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   _pendingViewState = nil;
 
   _displaySentinel = nil;
-  _transitionSentinel = nil;
 
   _pendingDisplayNodes = nil;
 }
@@ -825,11 +824,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 {
   ASDN::MutexLocker l(_propertyLock);
   _transitionInProgress = YES;
-    
-  if (!_transitionSentinel) {
-    _transitionSentinel = [[ASSentinel alloc] init];
-  }
-  return [_transitionSentinel increment];
+  _transitionID = OSAtomicAdd32(1, &_transitionID);
+  return _transitionID;
 }
 
 - (void)_finishOrCancelTransition
@@ -841,11 +837,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (BOOL)_shouldAbortTransitionWithID:(int32_t)transitionID
 {
   ASDN::MutexLocker l(_propertyLock);
-  if (_transitionInProgress) {
-    return _transitionSentinel == nil || transitionID != _transitionSentinel.value;
-  }
-  
-  return YES;
+  return (!_transitionInProgress || _transitionID != transitionID);
 }
 
 - (void)animateLayoutTransition:(id<ASContextTransitioning>)context
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -654,6 +654,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   return (!_flags.isMeasured || !ASSizeRangeEqualToSizeRange(constrainedSize, _constrainedSize));
 }
 
+#pragma mark - Layout Transition
+
 - (void)transitionLayoutWithAnimation:(BOOL)animated
                    shouldMeasureAsync:(BOOL)shouldMeasureAsync
                 measurementCompletion:(void(^)())completion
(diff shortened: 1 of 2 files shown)


levi:
Really great stuff, @maicki! I can imagine this wasn't easy to find.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -814,6 +814,40 @@ - (void)setUsesImplicitHierarchyManagement:(BOOL)value
...
+}
+
+- (BOOL)_shouldAbortTransitionWithID:(int32_t)transitionID
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if (_transitionInProgress) {
+    return _transitionSentinel == nil || transitionID != _transitionSentinel.value;
+  }
+  
+  return YES;
```
nguyenhuy: Should we do something like this?

```
- (BOOL)_shouldAbortTransitionWithID:(int32_t)transitionID
{
  return !_transitionInProgress || _transitionSentinel == nil || transitionID != _transitionSentinel.value;
}
```

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1696,6 +1733,10 @@ - (void)__setSupernode:(ASDisplayNode *)newSupernode
     if (newSupernode) {
       [self enterHierarchyState:stateToEnterOrExit];
     } else {
+      // If a node will be removed from the supernode it should go out from the layout pending state to remove all
+      // layout pending state related properties on the node
+      stateToEnterOrExit |= ASHierarchyStateLayoutPending;
```
nguyenhuy: Should we do the same when there is a new supernode?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -814,6 +814,40 @@ - (void)setUsesImplicitHierarchyManagement:(BOOL)value
...
+
+/// Starts a new transition and returns the transition id
+- (int32_t)_startNewTransition
+{
+  ASDN::MutexLocker l(_propertyLock);
+  _transitionInProgress = YES;
+    
+  if (!_transitionSentinel) {
+    _transitionSentinel = [[ASSentinel alloc] init];
+  }
```
maicki: @nguyenhuy @appleguy What do you guys think of using an instance variable that we atomically increase for that instead of the sentinel object (who does basically the same but has the overhead of creating and destroying an object ?

nguyenhuy: I think ASSentinel provides a somewhat useful but very thin layer of abstraction. It's tempting to remove it and use an atomic int instead.

maicki: Talked with @nguyenhuy about it on Slack and I will remove the sentinel and we will use an atomic int instead to remove the object overhead

maicki:
@appleguy @nguyenhuy Addressed comments from @nguyenhuy

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -790,18 +790,17 @@ - (void)calculatedLayoutDidChange
 - (void)cancelLayoutTransitionsInProgress
 {
   ASDN::MutexLocker l(_propertyLock);
-  if ([self _hasTransitionsInProgress]) {
-    // Invalidate transition sentinel to cancel transitions in progress
-    [self _invalidateTransitionSentinel];
+  if ([self _hasTransitionInProgress]) {
```
appleguy: Both of these names are more clear than they were before.  Thanks!

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -814,6 +813,33 @@ - (void)setUsesImplicitHierarchyManagement:(BOOL)value
...
+  ASDN::MutexLocker l(_propertyLock);
+  _transitionInProgress = YES;
+  _transitionID = OSAtomicAdd32(1, &_transitionID);
+  return _transitionID;
+}
+
+- (void)_finishOrCancelTransition
+{
+  ASDN::MutexLocker l(_propertyLock);
+  _transitionInProgress = NO;
```
appleguy: It almost feels to me like we should also increment the transitionID here.

Basically, if there were a call point that _only_ checked the transition ID and not transition in progress, it could accidentally succeed even after a transition is cancelled.

This scenario is unlikely, and does not currently occur so it is not a blocker.  Just something to consider for additional robustness.

I do understand we can't just reset the ID to 0, as there is the possibility of multiple calls then arriving at the same ID like "2" and actually passing.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1689,9 +1716,11 @@ - (void)__setSupernode:(ASDisplayNode *)newSupernode
   }
   
   if (supernodeDidChange) {
+    // Hierarchy state
```
appleguy: This comment and the next one could be removed

AsyncDisplayKit/ASDisplayNode.mm
appleguy: @maicki do you have a response to this comment from @nguyenhuy?

appleguy:
@maicki amazing investigatory work here!  It's a testament to your understanding of the code that the lines changed is relatively small.

PR #1624, opened 2016-05-05
[Accessibility / iOS 8] Fix 8.0 and 8.1-only, VoiceOver crash when clearing delegate / dataSource for Table or Collection

maicki:
If a delegate / dataSource get's set to nil in ASTableView or ASCollectionView on iOS 8.0 / 8.1 it triggers an accessibility call that the delegate / dataSource proxy is catching. The problem with that is that it goes into a loop as it tries to call `proxyTargetHasDeallocated` what could again trigger to set the delegate / dataSource to nil and the accessibility call is triggered again.


appleguy:
@levi, @nguyenhuy, anyone else who could give this a review?  Happy to merge as soon as one person gives a close analysis.

nguyenhuy:
@appleguy: LGTM

levi:
@appleguy LGTM. Land away!

PR #1630, opened 2016-05-05
[ASVideoNode] Move layout of ASVideoNode to layout specs

maicki:
This PR will move ASVideoNode to use layout spec to layout all of it's subnodes.

- This should fix assignment and position adjustments of a custom play button
- We can now use it properly within other nodes as subnodes and it throws an assertion if used the wrong way
- Fix sample project use of ASVideoNode

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -64,9 +64,9 @@ static NSString * const kStatus = @"status";
   
   ASImageNode *_placeholderImageNode; // TODO: Make ASVideoNode an ASImageNode subclass; remove this.
   
-  ASButtonNode *_playButton;
+  ASButtonNode *_playButtonNode;
   ASDisplayNode *_playerNode;
-  ASDisplayNode *_spinner;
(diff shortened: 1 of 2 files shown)


maicki:
@appleguy @Xinchi Could you guys please do a code review.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -185,24 +185,55 @@ - (void)removePlayerItemObservers:(AVPlayerItem *)playerItem
...
+- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
+{
+    // All subnodes should taking the whole node frame
+    CGSize maxSize = constrainedSize.max;
+    if (!CGSizeEqualToSize(self.preferredFrameSize, CGSizeZero)) {
+        maxSize = self.preferredFrameSize;
+    }
+    
+    // Prevent crashes through if infinite width or height
+    if (isinf(maxSize.width) || isinf(maxSize.height)) {
```
appleguy: What about the case when the minimum size is nonzero and the max size is inf?  should we use the min size?

PR #1632, opened 2016-05-07
[ASMultiplexImageNode] Do Not Request Already-Loaded Images From Data Source

Adlai-Holler:
In a recent update to PINRemoteImage, its memory cache went from storing UIImage instances to containers of NSData which are used to generate a UIImage instance on demand. Unfortunately, these UIImage instances show `isEqual: NO` compared to one another.

Since the images appear to be different, if you wire `ASMultiplexImageNodeDataSource.imageForImageIdentifier` up to synchronously retrieve an image from the PINRemoteImageManager instance, the image will change every time, which will call setNeedsDisplay, which will call displayWillStart, which will call fetchData, which will generate a new UIImage instance.

To break this cycle, this patch avoids requesting the image for an already-loaded image identifier from the multiplex node's data source. The only issue I can see with this is, if you load an image into a multiplex node and then manually set `image` on the multiplex image node (sidestepping the image loading mechanism), the node may continue using that image rather than attempting to refetch. It's not ideal, but I can't imagine a use case for that.

@garrettmoon @appleguy 

As a separate issue, `ASPINRemoteImageDownloader` uses a private image manager instance which completely surprised me. I think this behavior is pretty misleading; for instance if I want to clear the memory/disk cache, or print the disk cache size for diagnostics, I couldn't and most users probably wouldn't even know that there were two PINRemoteImageManager instances.

### DIFF
--- a/AsyncDisplayKit/ASMultiplexImageNode.mm
+++ b/AsyncDisplayKit/ASMultiplexImageNode.mm
@@ -436,8 +436,12 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   }
 
   // Grab the best available image from the data source.
+  UIImage *existingImage = self.image;
   for (id imageIdentifier in _imageIdentifiers) {
-    UIImage *image = [_dataSource multiplexImageNode:self imageForImageIdentifier:imageIdentifier];
+    // If this image is already loaded, don't request it from the data source again because
+    // the data source may generate a new instance of UIImage that returns NO for isEqual:
+    // and we'll end up in an infinite loading loop.
+    UIImage *image = ASObjectIsEqual(imageIdentifier, _loadedImageIdentifier) ? existingImage : [_dataSource multiplexImageNode:self imageForImageIdentifier:imageIdentifier];
     if (image) {
       if (imageIdentifierOut) {
         *imageIdentifierOut = imageIdentifier;


appleguy:
Thanks for keeping a close eye on this!

We recently made some pretty deep overhauls to PINRemoteImage, in order to add seamless support for Animated GIFs at both the image library and ASDK / UI component level.  I can definitely understand the issue you’re pointing out, and wish we caught it, but am glad you did for us.

These changes were done in a very compressed timeline, and overall I think they were very well architected (protocol setup to ensure that even ASDK can build without PINRemoteImage despite the tight GIF integration), including consideration of future animatable image formats / custom extensions for other image formats.  Nonetheless it is certainly worth another careful look through the cache paths to ensure nothing else was disrupted, but I suspect the fix will be pretty feasible.

@garrettmoon, definitely a good thing to have for 3.0 non-beta (as ASDK is pointing to the Beta right now, even in its production release).  I know it’s a tough time to pull aside cycles to spend on the release, but we should plan out what’s required for 3.0 in the next week and decide the timeline it should be done in (maybe end of month?)

garrettmoon:
This change LGTM.

@Adlai-Holler The regeneration of UIImage's on request *is* an unfortunate side effect :/ I thought it better to store only the NSData for alternate representation support. The only way I can see around this is to also store the UIImage which is essentially a copy of the NSData.

I'm open to suggestions around changing ASPINRemoteImageDownloader to use the public instance. The specific problem with that is the version used by ASDK needs to be the alternative representation provider.

Do you think we should expose sharedPINRemoteImageManager? Can you think of another solution?

Adlai-Holler:
@garrettmoon Would it be possible NSDatas when the image was created with an alternate representation and UIImages otherwise? It's definitely not critically urgent but I think that would work.

Maybe make `alternateRepresentationProvider` publicly mutable and nullable on PINRemoteImageManager.  Then `ASPINRemoteImageDownloader` can dispatch-once set itself on the shared manager, if the user hasn't already set some other provider.

garrettmoon:
@Adlai-Holler The problem with storing the UIImage or NSData depending is that you can opt out of an alternative representation on a per-request basis. This means one request could ask for an alternative representation and then the next opt out. Because there's no NSData, we can't fulfill that request without hitting the disk cache.

The reason I didn't make alternativeRepresentationProvider a mutable property is that we'd need to add some sort of locking mechanism for reading it. Currently, PINRemoteImage never locks on the main thread and I was hoping to continue that :( There may be a way around this using atomic swaps, but I haven't looked into it.

PR #1635, opened 2016-05-07
[ASPINRemoteImageDownloader] Call Download Progress Handler

Adlai-Holler:
I snuck in two other small changes.

### DIFF
--- a/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.h
+++ b/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.h
@@ -9,8 +9,12 @@
 #import <Foundation/Foundation.h>
 #import "ASImageProtocols.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 @interface ASPINRemoteImageDownloader : NSObject <ASImageCacheProtocol, ASImageDownloaderProtocol>
 
 + (ASPINRemoteImageDownloader *)sharedDownloader;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.h
+++ b/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.h
@@ -11,6 +11,6 @@
 
 @interface ASPINRemoteImageDownloader : NSObject <ASImageCacheProtocol, ASImageDownloaderProtocol>
 
-+ (instancetype)sharedDownloader;
++ (ASPINRemoteImageDownloader *)sharedDownloader;
 
 @end
--- a/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
+++ b/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
@@ -96,8 +96,9 @@
 
 - (id <ASImageContainerProtocol>)synchronouslyFetchedCachedImageWithURL:(NSURL *)URL;
 {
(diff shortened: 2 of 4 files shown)


Adlai-Holler:
OK I ended up adding a bunch of other tweaks/fixes. Sorry for the complexity but they're all worth it.

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.h
```diff
@@ -9,8 +9,12 @@
 #import <Foundation/Foundation.h>
 #import "ASImageProtocols.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 @interface ASPINRemoteImageDownloader : NSObject <ASImageCacheProtocol, ASImageDownloaderProtocol>
 
-+ (instancetype)sharedDownloader;
++ (ASPINRemoteImageDownloader *)sharedDownloader;
```
garrettmoon: Is there a reason this shouldn't be an instancetype?

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -133,7 +134,18 @@ - (nullable id)downloadImageWithURL:(NSURL *)URL
                    downloadProgress:(ASImageDownloaderProgress)downloadProgress
                          completion:(ASImageDownloaderCompletion)completion;
 {
-  return [[self sharedPINRemoteImageManager] downloadImageWithURL:URL options:PINRemoteImageManagerDownloadOptionsSkipDecode completion:^(PINRemoteImageManagerResult *result) {
+  return [[self sharedPINRemoteImageManager] downloadImageWithURL:URL options:PINRemoteImageManagerDownloadOptionsSkipDecode progressDownload:^(int64_t completedBytes, int64_t totalBytes) {
+    if (downloadProgress == nil) { return; }
+    
+    /// If we're targeting the main queue and we're on the main thread, call immediately.
```
garrettmoon: I'm not sure if dispatching to the main thread is necessary here? I think the reason it is happening below is to prevent image flashing when the load happens synchronously. However I guess it doesn't hurt either and makes the behavior consistent… Maybe ignore me :)

garrettmoon:
Looks good to me as long as the instancetype change was intentional.

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.h
Adlai-Holler: It can't be instancetype, because if I subclass, then the return type of `+[TRPINRemoteImageDownloader sharedDownloader]` is `TRPINRemoteImageDownloader` but what I'll get is actually an `ASPINRemoteImageDownloader`.

garrettmoon: Ahhhh, right!

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
Adlai-Holler: My thinking here was, I wanted to make sure that progress events were never delivered after result events. So if all events are delivered on the main queue, and callbackQueue is main too, we need to deliver each progress event synchronously or else it might come in out-of-order.

Adlai-Holler:
Awesome! Let's take it to the hoop.

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
garrettmoon: 👍

PR #1637, opened 2016-05-09
Fix to text shrinking

rcancro:
Summary: Previously the first thing we did when calculating the size of text was to truncate it. This could lead to ASFontSizeAdjuster being sent a truncated string which, obviously, never needed to shrink. If we change the order then we first shrink, then truncate if still necessary.

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -156,7 +156,7 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
     }];
   }
   
-  [self truncater];
+  [[self truncater] truncate];
   
   // Force glyph generation and layout, which may not have happened yet (and isn't triggered by
   // -usedRectForTextContainer:).
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -137,25 +137,30 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
 
 - (void)_calculateSize
 {
-  [self truncater];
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -137,25 +137,30 @@ - (void)setConstrainedSize:(CGSize)constrainedSize
...
       [ASTextKitFontSizeAdjuster adjustFontSizeForAttributeString:scaledString withScaleFactor:_currentScaleFactor];
       scaledTextStorage = [[NSTextStorage alloc] initWithAttributedString:scaledString];
       
       [textStorage removeLayoutManager:layoutManager];
       [scaledTextStorage addLayoutManager:layoutManager];
-    }
+    }];
+  }
+  
+  [self truncater];
```
garrettmoon: Should we rename this method to truncateIfNecessary? Something better even? It's weird that calling [self truncater] does something other than return a truncater.

rcancro: I don't disagree.

AsyncDisplayKit/TextKit/ASTextKitTruncating.h
```diff
@@ -33,4 +33,9 @@
      truncationAttributedString:(NSAttributedString *)truncationAttributedString
          avoidTailTruncationSet:(NSCharacterSet *)avoidTailTruncationSet;
 
+/**
+ *  Actually do the truncation.
```
levi: LOL. Might want some more context for this note.

levi:
Looks good.

AsyncDisplayKit/TextKit/ASTextKitTruncating.h
appleguy: Haha, yeah - let's leave out unnecessary comments, though humor is welcome in ones that also provide information to the reader :)

PR #1646, opened 2016-05-11
Add Slack to Contributing file

levi:


nguyenhuy:
We actually have an invite bot: http://asdk-slack-auto-invite.herokuapp.com

PR #1648, opened 2016-05-11
[ASDisplayNode] Prevent placeholder creation for zero-bounds nodes.

### LINKED ISSUE #1601: [ASTextNode] placeholderImage throws invalid context if calculatedSize is CGSizeZero
We have had a situation where placeholderImage would be called when calculatedSize is zero. In this case it throws a lot of invalid context errors.

### LINKED ISSUE #1638: [ASTextNode] Crash when accessing renderer.visibleRanges[0] 1.9.73
After trying to update to `1.9.73` my app started to crash in several places due to `ASTextNode` nodes calling the method `placeholderImage` then inside it crashes when it tries to access `renderer.visibleRanges[0]`. The index does not exist, `visibleRanges` is empty and it crashes with `EXC_BAD_ACCESS`

Here is a screenshot of the stack trace. I can't paste as text because Xcode will crash together with this problem and simply close on me, making it super hard to debug. So I managed to take a screenshot and also to put some log messages to make sure my guess on this issue was right.

asdk-crash-and-invalid-context

Any ideas of what's happening here?

Looking at the screenshot you can also spot another problem, Core Graphics is throwing invalid context here because the `CGSize` is `{0, 0}` but this is another open issue here:  [1601](https://github.com/facebook/AsyncDisplayKit/issues/1601)

levi:
Resolves #1601
Resolves #1638

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1793,7 +1793,8 @@ static NSInteger incrementIfFound(NSInteger i) {
 
 - (BOOL)_shouldHavePlaceholderLayer
 {
-  return (_placeholderEnabled && [self __implementsDisplay]);
+  CGSize layoutSize = (_layout ? _layout.size : CGSizeZero);
+  return (CGSizeEqualToSize(layoutSize, CGSizeZero) == NO && _placeholderEnabled && [self __implementsDisplay]);
 }
 
 - (void)_setupPlaceholderLayerIfNeeded
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1780,7 +1780,9 @@ static NSInteger incrementIfFound(NSInteger i) {
   }
 }
 
-// Helper method to summarize whether or not the node run through the display process
+/**
+ * Helper method to summarize whether or not the node run through the display process
+ */
 - (BOOL)__implementsDisplay
 {
   return _flags.implementsDrawRect || _flags.implementsImageDisplay || _flags.shouldRasterizeDescendants || _flags.implementsInstanceDrawRect || _flags.implementsInstanceImageDisplay;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1793,7 +1793,8 @@ - (BOOL)placeholderShouldPersist
 
 - (BOOL)_shouldHavePlaceholderLayer
 {
-  return (_placeholderEnabled && [self __implementsDisplay]);
+  CGSize layoutSize = (_layout ? _layout.size : CGSizeZero);
```
appleguy: Does this function need a lock?

 I believe it is not sufficient to check the layout size here. An element that is positioned with manual layout may have a zero size here, but a very non-zero bounds size. 

  Is this method always called on the main thread, presumably yes because the placeholder layer must be created there? If so, we should probably assert that as well, and we may be able to rely on the fact that a layout sublayers pass has occurred… Which would mean that it is sufficient to on the check bounds  and not worry about checking the layout as well

levi: What other size would we check? `calculatedSize` is the same value, which is what a manually sized node would use for the final frame size.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1793,7 +1795,9 @@ - (BOOL)placeholderShouldPersist
 
 - (BOOL)_shouldHavePlaceholderLayer
 {
-  return (_placeholderEnabled && [self __implementsDisplay]);
+  ASDisplayNodeAssertMainThread();
+  CGSize calculatedSize = [self calculatedSize];
```
appleguy: Thanks for adding the assert!

What happens in the case you have an ASMapNode (or anything), and set its frame = Rect(5, 5, 100, 100) without measure: being called?  I believe calculatedSize will be zero, even though the element will be fully onscreen and prior to this diff would show a placeholder.

appleguy:
@levi ping - I think we need to add some kind of consideration of the bounds value here.  E.g. there are cases like ASDisplayNode wrapping of synchronous elements where at most there is a preferredFrameSize set, or even nothing is set besides -layout performing the positioning.  These cases may still require a placeholder, as CA's synchronous layout pass / onscreen visibility of the element may occur before async rendering is done.

levi:
@appleguy, @maicki has a better solution for this already landed. Closing out.

PR #1653, opened 2016-05-11
[ASViewController] Add visibility depth management

garrettmoon:
This adds support for the concept of visibility depth.

Visibility essentially defines the number of user actions it would
take a user to have a view controller visible. Knowing a view controllers
visibility depth allows view controllers to take action such as clearing
out memory that can be restored at a later date.

This patch also add two new view controller subclasses which adopt
the ASManagesChildVisibilityDepth protocol. Any view controller
that has child view controllers can adopt this protocol to indicate
to the child what they're visibility is. For example, ASNavigationController
will return a visibility depth of it's own visibilityDepth + 1 for
a view controller that would be revealed by tapping the back button.

### DIFF
--- a/AsyncDisplayKit/ASNavigationController.m
+++ b/AsyncDisplayKit/ASNavigationController.m
@@ -39,8 +39,8 @@ ASVisibilityDepthImplementation;
   NSAssert(viewControllerIndex != NSNotFound, @"childViewController is not in the navigation stack.");
   
   if (viewControllerIndex == self.viewControllers.count - 1) {
-    //view controller is at the top
-    return [self visibilityDepth] + 0;
+    //view controller is at the top, just return our own visibility depth.
+    return [self visibilityDepth];
   } else if (viewControllerIndex == 0) {
     //view controller is the root view controller. Can be accessed by holding the back button.
     return [self visibilityDepth] + 1;
--- a/AsyncDisplayKit/ASNavigationController.m
+++ b/AsyncDisplayKit/ASNavigationController.m
@@ -20,6 +20,8 @@ ASVisibilityViewWillAppear;
 
 ASVisibilityViewDidDisappearImplementation;
 
+ASVisibilitySetVisibilityDepth;
+
 ASVisibilityDepthImplementation;
 
 - (void)visibilityDepthDidChange
--- a/AsyncDisplayKit/ASNavigationController.m
+++ b/AsyncDisplayKit/ASNavigationController.m
@@ -14,43 +14,13 @@
   NSInteger _visibilityDepth;
 }
 
-- (void)didMoveToParentViewController:(UIViewController *)parent
-{
-  [super didMoveToParentViewController:parent];
-  [self visibilityDepthDidChange];
-}
+ASVisibilityDidMoveToParentViewController;
 
-- (void)viewWillAppear:(BOOL)animated
-{
-  [super viewWillAppear:animated];
+ASVisibilityViewWillAppear;
 
-  if (_parentManagesVisibilityDepth == NO) {
-    _visibilityDepth = 0;
-    [self visibilityDepthDidChange];
-  }
-}
+ASVisibilityViewDidDisappearImplementation;
 
-- (void)viewDidDisappear:(BOOL)animated
-{
-  [super viewDidDisappear:animated];
-  
-  if (_parentManagesVisibilityDepth == NO) {
-    _visibilityDepth = 1;
-    [self visibilityDepthDidChange];
-  }
-}
-
-- (NSInteger)visibilityDepth
-{
-  if (self.parentViewController && _parentManagesVisibilityDepth == NO) {
-    _parentManagesVisibilityDepth = [self.parentViewController conformsToProtocol:@protocol(ASManagesChildVisibilityDepth)];
-  }
-  
-  if (_parentManagesVisibilityDepth) {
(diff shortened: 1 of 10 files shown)


garrettmoon:
Looking at build errors

AsyncDisplayKit/ASVisibilityProtocols.m
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Garrett Moon on 4/28/16.
+//
+//
+
+#import <Foundation/Foundation.h>
+
+#import "ASVisibilityProtocols.h"
+
+ASLayoutRangeMode ASLayoutRangeModeForVisibilityDepth(NSUInteger visibilityDepth)
```
levi: Why the separation between an integer value and the enum? Why not use the enum everywhere?

Really wish we could use Swift in these instances. A lot of what you're doing can be encapsulated into a single enum structure.

AsyncDisplayKit/ASVisibilityProtocols.h
```diff
@@ -0,0 +1,89 @@
...
+}
+
+#define ASVisibilityViewWillAppear \
+- (void)viewWillAppear:(BOOL)animated \
+{ \
+  [super viewWillAppear:animated]; \
+  \
+  if (_parentManagesVisibilityDepth == NO) { \
+    _visibilityDepth = 0; \
+    [self visibilityDepthDidChange]; \
```
levi: Why not create a setter that invokes the visibilityDidChange automatically?

AsyncDisplayKit/ASViewController.m
```diff
@@ -113,7 +149,9 @@ - (void)setAutomaticallyAdjustRangeModeBasedOnViewEvents:(BOOL)automaticallyAdju
 - (void)updateCurrentRangeModeWithModeIfPossible:(ASLayoutRangeMode)rangeMode
```
levi: We should shorten this to `updateCurrentRangeModeIfNeeded`

AsyncDisplayKit/ASNavigationController.m
```diff
@@ -0,0 +1,91 @@
...
+}
+
+- (NSInteger)visibilityDepthOfChildViewController:(UIViewController *)childViewController
+{
+  NSUInteger viewControllerIndex = [self.viewControllers indexOfObject:childViewController];
+  NSAssert(viewControllerIndex != NSNotFound, @"childViewController is not in the navigation stack.");
+  
+  if (viewControllerIndex == self.viewControllers.count - 1) {
+    //view controller is at the top
+    return [self visibilityDepth] + 0;
```
levi: what's the point of adding zero?

AsyncDisplayKit/ASVisibilityProtocols.m
garrettmoon: There's a difference between range modes and visibility distance. For example, in high memory pressure environments, you might want to map these differently. Or perhaps the other way around. Anyway, I felt the distinction was important – this measures the number of actions to get a view controller visible. Range modes specify a bunch of different things. They're related but different.

AsyncDisplayKit/ASNavigationController.m
garrettmoon: I thought it a bit more readable, but maybe it's more confusing than clarifying?

AsyncDisplayKit/ASViewController.m
garrettmoon: How about updateCurrentRangeWithModeIfNeeded? It's pedantic, but updateCurrentRangeModeIfNeeded doesn't seem like it would take a parameter.

AsyncDisplayKit/ASVisibilityProtocols.h
garrettmoon: That should be ok if the setter is private.

AsyncDisplayKit/ASViewController.m
levi: In the case of not adding the ifNeeded/ifPossible, would you do `updateCurrentRangeModeWithRangeMode`?

AsyncDisplayKit/ASVisibilityProtocols.m
levi: Thanks for the insight. Makes a lot of sense.

AsyncDisplayKit/ASNavigationController.m
levi: Yeah is a little confusing on first read. Had to think about the additive properties of zero.

AsyncDisplayKit/ASViewController.m
garrettmoon: Deal :)

levi:
👍

appleguy:
@garrettmoon  I really love the concept of visibility distance, even the term. You did a phenomenal job coming up with this philosophical approach to abstracting navigation and memory management.   

Proactive memory management has always been cord to the framework methodology, but if this is a direction that has never been explored by it outside of special systems implemented by focused use cases in apps like Paper and Pinterest.

Abstracting it in this way will bring dramatic benefits to applications that do not have such systems, as well as even more efficient memory usage for large applications, all while reducing complexity. 

 Thanks for taking this on with a very rigorous approach to the problem, and refining it alongside @levi's helpful review feedback!

PR #1656, opened 2016-05-12
[ASCellNode] Adding support for scroll view begin / end drag to VisibilityEvent changes.

Xinchi:


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -670,6 +674,29 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
   }
 }
 
+- (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
+{
+  for (_ASCollectionViewCell *collectionCell in _cellsForVisibilityUpdates) {
+    [[collectionCell node] cellNodeVisibilityEvent:ASCellNodeVisibilityEventWillBeginDragging
+                                          inScrollView:scrollView
```
nguyenhuy: Nit: indentation

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -670,6 +674,29 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
...
+                                         withCellFrame:collectionCell.frame];
+  }
+  if (_asyncDelegateFlags.asyncDelegateScrollViewWillBeginDragging) {
+    [_asyncDelegate scrollViewWillBeginDragging:scrollView];
+  }
+}
+
+- (void)scrollViewDidEndDragging:(UIScrollView *)scrollView willDecelerate:(BOOL)decelerate
+{
+    for (_ASCollectionViewCell *collectionCell in _cellsForVisibilityUpdates) {
```
nguyenhuy: Looks like 4 spaces?

AsyncDisplayKit/ASTableView.mm
```diff
@@ -692,6 +696,29 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
   }
 }
 
+- (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView
+{
+  for (_ASTableViewCell *tableViewCell in _cellsForVisibilityUpdates) {
+    [[tableViewCell node] cellNodeVisibilityEvent:ASCellNodeVisibilityEventWillBeginDragging
+                                          inScrollView:scrollView
```
nguyenhuy: Nit: indentation

AsyncDisplayKit/ASTableView.mm
```diff
@@ -692,6 +696,29 @@ - (void)scrollViewWillEndDragging:(UIScrollView *)scrollView withVelocity:(CGPoi
...
+  if (_asyncDelegateFlags.asyncDelegateScrollViewWillBeginDragging) {
+    [_asyncDelegate scrollViewWillBeginDragging:scrollView];
+  }
+}
+
+- (void)scrollViewDidEndDragging:(UIScrollView *)scrollView willDecelerate:(BOOL)decelerate
+{
+  for (_ASTableViewCell *tableViewCell in _cellsForVisibilityUpdates) {
+    [[tableViewCell node] cellNodeVisibilityEvent:ASCellNodeVisibilityEventDidEndDragging
+                                          inScrollView:scrollView
```
nguyenhuy: Nit: indentation

nguyenhuy:
@Xinchi: Something went wrong in the merge commit and caused missing declarations of `ASCellNodeVisibilityEventWillBeginDragging` and `ASCellNodeVisibilityEventDidEndDragging`. This and nits aside, LGTM!

AsyncDisplayKit/ASCellNode.h
```diff
@@ -30,6 +30,14 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
    * Indicates a cell is no longer visible
    */
   ASCellNodeVisibilityEventInvisible,
+  /**
+   * Indicates user has started dragging the visible cell
+   */
+  ASCellNodeVisibilityEventWillBeginDragging,
```
Xinchi: Feedback needed:

Adding 2 events here to listen to dragging events for visible cells on screen.  I am not sure if they belong in visible events in the first place, but it seems this is the easiest thing to do to make it happen, because they function similarly to scrollviewDidScroll and update the visible cell rect for visible cells (ASCellNodeVisibilityEventVisibleRectChanged).  

Alternatively we can create a separate `UserInteractionEvent` for drag events, but that would create a lot more code and mess.  Would like to hear your feedback on this.

binl:
These are cool additions to the framework! Thanks for working on this!

From an API/Framework design perspective, I'm not too sure if these 2 new visibility events make a lot of sense:
```
ASCellNodeVisibilityEventWillBeginDragging
ASCellNodeVisibilityEventDidEndDragging
```
For the following reasons:
1. They are not really visibility events, which you already pointed out.
2. The drag isn't happening on the cells that are receiving these events, they are scrollView events.

I don't think we should propagate all the drag events from the underlying scrollView to all the visible cells, I felt that if a cell cares about those events, it should listen for it (which might just happen in the App level logic)

My 2 cents :-)

appleguy:
@Xinchi I've thought carefully about this and also discussed with @hannahmbanana about API naming.  My original opinion matched @binl and I think there is some question about the flexibility of the original API name.  Your work here matched it perfectly and is correct, it simply exposed some limitations in the flexibility of the original name choice.

However, after further consideration, I think we should move forward with this as-is.  I am likely to want to rename this to something more generic later, like ASCellNodeInterfaceEvent..., but this would be a breaking API change and is not worth doing at this moment (we should consider it for 2.0 though).

In the end, the reason I'm happy to land this is that I definitely agree the functionality is related to the use cases for the other values.  They are all related to a cell monitoring its relationship with the scrollable content area.  It is always possible to solve problems like this at the application level - however, ASDK's main goal is to provide conveniences wherever they don't get in the way / slow things down, specifically to reduce the complexity of application code and make powerful capabilities easy.  Avoiding the need for application level delegate patterns is definitely a simplification.

Thanks for carefully thinking about this, making a very clean change to core framework code, and advocating for why this is the right solution — as well as thinking through alternatives.

If you can rebase it, I'll merge ASAP.

appleguy:
Superceded by #1712

PR #1660, opened 2016-05-13
Fix passing wrong constrained size to nodes

maicki:
In _layoutNodesFromContexts:ofKind:completion: we pass the full array of contexts to _layoutNodes:fromContexts:atIndexesOfRange:ofKind: but for nodes we pass a subarray of nodes instead based on the batchCount. As batchRange we always start from 0 to batchCount. We now use the same indexes that we use to create the subarray of nodes to get a subarray of contexts that we pass to _layoutNodes:fromContexts:atIndexesOfRange:ofKind:.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -191,14 +191,16 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
   for (NSUInteger j = 0; j < nodeCount; j += kASDataControllerSizingCountPerProcessor) {
     NSInteger batchCount = MIN(kASDataControllerSizingCountPerProcessor, nodeCount - j);
 
-    __block NSArray *subarray;
+    // Get subcontexts for the range of nodes that will be allocated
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -190,27 +190,33 @@ - (void)_layoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofK
...
           ASDisplayNodeAssertNotNil(node, @"Node block created nil node; %@, %@", self, self.dataSource);
           node = [[ASCellNode alloc] init]; // Fallback to avoid crash for production apps.
         }
         allocatedNodeBuffer[i] = node;
+        allocatedContextBuffer[i] = context;
       });
-      subarray = [[NSArray alloc] initWithObjects:allocatedNodeBuffer count:batchCount];
-
+      subarrayOfNodes = [[NSArray alloc] initWithObjects:allocatedNodeBuffer count:batchCount];
+      subarrayOfContexts = [NSArray arrayWithObjects:allocatedContextBuffer count:batchCount];
```
appleguy: would be good to match the alloc / init or both use the + method, but this look really good overall and will land now!

Issue #431, opened 2015-04-18
[Swift] Specifically for Archive builds, #import <AsyncDisplayKit/AsyncDisplayKit.h> 'file not found'


ProjectDent:
I followed the instructions to manually integrate AsyncDisplayKit with my app. It builds and runs fine, but when I go to Archive, it points me to my bridging header file, the line `#import <AsyncDisplayKit/AsyncDisplayKit.h>`, giving me the error `'<AsyncDisplayKit/AsyncDisplayKit.h>' file not found`.

Under Framework Header Paths, I've now added the AsyncDisplayKit folder, but this hasn't solved the problem. I'm not sure what the problem is or what I can do about it.

ProjectDent:
Confirmed by trying this with a test project.

ProjectDent:
I detailed this issue here:

http://stackoverflow.com/questions/29730394/failed-to-import-bridging-header-file-not-found-when-archiving-my-project

habitatus:
@ProjectDent I'm having the same issue. Did you ever resolve it?

ProjectDent:
@habitatus I removed the project files, and just imported all AsyncDisplayKit files into my project manually.

habitatus:
@ProjectDent That seems reasonable. Changing the #import <header> to #import "header" throughout the ASDK did the trick for me.
Thank you!

appleguy:
@ProjectDent thanks for reporting this.  Could you test out locally applying this PR to see if it resolves this particular issue with the #import syntax you had previously?  #449 

Alternatively, if you can upload a sample project, I'll test it out and make sure this works properly.

kusmi:
I'm using 1.2.2 and still see this issue. I'm not using Pods, but import the AsyncDisplayKit directly as written in AsyncDisplayKit documentation.

klivin:
Has there been any progress with this issue as this still exists, and makes for supporting iOS 7.1 impossible as Carthage solution is iOS 8 and later (similar to the cocoapods framework support)

appleguy:
@klivin are you sure it is not an issue such as the one described on this page under "archiving works a little differently?"

http://stackoverflow.com/questions/5543854/xcode-4-cant-locate-public-header-files-from-static-library-dependency

I have not seen any other reports of this, but if you can provide more concrete details such as a simple test project that reproduces the issue, we can certainly fix it. Without a way to reproduce, we have no way to test the fix.

appleguy:
[...]

PR #1663, opened 2016-05-14
Replace Framework-Style Imports with Direct Imports

Adlai-Holler:
CocoaPods was having trouble building ASDK as a dependency of an extension pod I made, and this fixed it right up. 

This fixes https://github.com/CocoaPods/CocoaPods/issues/5230 and probably also #431 . @appleguy @maicki thoughts?

### DIFF
--- a/AsyncDisplayKit/ASButtonNode.h
+++ b/AsyncDisplayKit/ASButtonNode.h
@@ -6,8 +6,8 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-#import <AsyncDisplayKit/ASTextNode.h>
-#import <AsyncDisplayKit/ASImageNode.h>
+#import "ASTextNode.h"
+#import "ASImageNode.h"
 
 @interface ASButtonNode : ASControlNode
 
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -6,7 +6,7 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
-#import <AsyncDisplayKit/ASDisplayNode.h>
+#import "ASDisplayNode.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -11,14 +11,14 @@
 #import "ASInternalHelpers.h"
 #import "ASEqualityHelpers.h"
 #import "ASDisplayNodeInternal.h"
-#import <AsyncDisplayKit/_ASDisplayView.h>
-#import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
-#import <AsyncDisplayKit/ASDisplayNode+Beta.h>
-#import <AsyncDisplayKit/ASTextNode.h>
+#import "_ASDisplayView.h"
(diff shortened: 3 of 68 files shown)


maicki:
@Adlai-Holler You have any idea what can cause that. Is it related to using Swift in the extension pod? Furthermore did this just start happen with Cocoapod 1.0 and was fine with < 1.0?

Overall I would be fine with the change if this resolves the issue to have ASDK as dependency under some circumstances.

appleguy:
@Adlai-Holler also fine with merging this if I understand why it is better.  Historically I was under the impression that it was more correct / preferable to use a framework import for anything that is in a public .h file.

maicki:
@Adlai-Holler I tried to reproduce the error as I want to understand the whole `#import` mechanism and also how it relates to Swift and modules (`@import`) a bit more. I tried you sample project from the CocoaPods issue, but it built without any problems :/

Can you quickly check that example again if it fails for you? Fyi used CocoaPods 1.0

Adlai-Holler:
I checked it yesterday under Xcode 7.2.1 or 7.3.1, CocoaPods 0.39.0 or 1.0.0 and it wouldn't build, with a different failure under each case. To confirm: you were able to run the SwiftKittens app?

maicki:
Yeah running the SwiftKittens app. Xcode 7.3.1 and CocoaPods 1.0.0. See Screenshot.
screen shot 2016-05-15 at 8 55 49 pm

Adlai-Holler:
Huh that's really strange, I wonder why the behavior would be different.

@appleguy is your belief that the framework-style imports are more correct strong enough to justify me investigating the different behavior? I've deintegrated and cleared pod cache and deleted derived data, all the usual tricks already.

maicki:
Change the Podfile.lock at the end from `COCOAPODS: 1.0.0.rc.1` to `COCOAPODS: 1.0.0`. That fixed it for me. If you use version 0.39.0 I actually can reproduce the issue now.

I think we should change the imports to non framework-style imports and accept the PR though.

Adlai-Holler:
Huh, never would’ve thought of that. Sweet!

maicki:
You can also just delete the Podfile.lock and create it new via `pod install` and CococaPods 1.0.0 that should have the same effect.

Adlai-Holler:
@maicki I used `pod deintegrate` which reports "All traces of CocoaPods have been removed" which must include Podfile.lock, but it didn't fix the build.

maicki:
This didn't remove the Podfile.lock for me just look into the project folder. Did you try just deleting the Podfile.lock and call 'pod install' ? Be sure to use CocoaPods 1.0.0 and not the rc.

Adlai-Holler:
No I'm hiking today, only talking about my experiences before I opened this PR.

appleguy:
Very interesting!  Still following this.  Adlai, it doesn’t seem urgent, but if you can reproduce Michael’s comments it seems like the main issue here was some transient problem in Cocoapods rc.

Although I’m not yet convinced we should make a change, if you have any reference that this style is preferred / expected for referencing files within a project (even in public headers), I’d be happy for us to adopt that.  The only reason I had the sense that framework includes were better is that the text of the header file is pulled into code being compiled at the application level, and in that context (no longer inside the framework), referencing other headers is properly done via the framework specifier.

Adlai-Holler:
Yep! Recreating the lock file under 1.0 fixes it. Thanks @maicki and @appleguy for a great review. Closing.

PR #1666, opened 2016-05-16
Change ASTableView's visibleNodes interface to return array of ASCellNodes

rmls:
It is known that the visible nodes of an `ASTableView` are `ASCellNode`s. This diff specialize the `visibleNodes` method in `ASTableView` to return an array of `ASCellNode`s instead of an array of `ASDisplayNode`s. This way, it is both clearer of what the objects of that array truly are and also avoids unnecessary casts when using this method.


facebook-github-bot-3:

PR #1667, opened 2016-05-16
[ASDataController] Small improvements

maicki:
1. Use NSFastEnumeration directly instead of getting a NSEnumerator for iterating through completedNodes which should improve performance
2. Use the sorted array in `reloadRowsAtIndexPaths:withAnimationOptions:`. Before we just sorted it, but didn't use it :/
3. Remove FIXME about sorting index paths for deleting rows. @Adlai-Holler I saw you added the FIXME, why do you think we should order in descending order before deleting?

cc @appleguy

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -815,8 +815,7 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
 
     [_editingTransactionQueue waitUntilAllOperationsAreFinished];
     
-    // sort indexPath in order to avoid messing up the index when deleting
-    // FIXME: Shouldn't deletes be sorted in descending order?
+    // sort indexPath in order to avoid messing up the index when deleting in several batches
     NSArray *sortedIndexPaths = [indexPaths sortedArrayUsingSelector:@selector(compare:)];
 
     [_editingTransactionQueue addOperationWithBlock:^{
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
@maicki My thinking there is, when you delete at an index, you invalidate all subsequent indices, and we never want to use an invalid index in a later delete.

For example if you've got [A, B, C, D] and "delete {2, 3}", if you go in order you'll get [A, B, D] and then crash when you attempt to delete 3. If you do "delete {3, 2}" you'll get [A, B, C] then [A, B] as expected.

appleguy:
@maicki sounds like the FIXME comment should be retained?  Happy to merge this once it's back.  Or - I wasn't able to review in detail - do you think the behavior the FIXME asks for is already implemented?

maicki:
@appleguy Yes the FIXME should be retained if we do not plan to change the behavior. That said I think @Adlai-Holler  has a good point to changing the order in a delete case. What do you think, should we add the FIXME back or should I change the behavior so it aligns with the FIXME comment within this PR?

appleguy:
It would be fantastic if you can implement the behavior that the comment speculated is necessary! As long as you agree the description seems like a correct approach. I am not deeply familiar with this implementation and have not carefully analyze the description that Adlai provided in the comment above.

maicki:
@appleguy Ok I made the changes.

We had a longer discussion about the fastest way to actually reverse sort the index path array. Let me know if you know a faster way as currently implemented. There is also the `sortedArrayUsingComparator:` method on NSArray, but based on the sorting benchmark and reverse benchmark in this two articles it should be faster to first use `sortedArrayUsingSelector:` and after that use `reverseObjectEnumerator` and `allObject`.

https://www.objc.io/issues/7-foundation/collections/ and http://stackoverflow.com/questions/586370/how-can-i-reverse-a-nsarray-in-objective-c

Adlai-Holler:
@maicki I believe there's a method like `as_inverseCompare:` that you could use as the sort selector to directly sort index paths descending.

maicki:
@Adlai-Holler Nice, it's `asdk_inverseCompare:` will update that

maicki:
@appleguy Seems like currently we cannot remove index paths in reverse order. `ASDeleteElementsInMultidimensionalArrayAtIndexPaths` -> `ASRecursivelyUpdateMultidimensionalArrayAtIndexPaths` expects a sorted index path otherwise we assert in `ASDeleteElementsInMultidimensionalArrayAtIndexPaths` with `Indexpath is out of range !`

appleguy:
Ah, I see.  We should be able to fix that method / create another variant that takes a reverse order, but that’s annoying and definitely a blocker until we get around to it.

Feel free to restore the original comment for now, or push through the multidimensional change.  Also wondering gif it makes sense to add a TwoDimensional optimized variant like the others I added, which helped performance a lot.

maicki:
Restored the original FIXME comment I think ready to merge now.

Thinking about the TwoDimensional methods that do not support handling unordered array at the moment and I think it was actually pretty important to use the sorted array in `reloadRowsAtIndexPaths:withAnimationOptions:` now as before we just sorted it but not used it :/ . If an API consumer passes in an `indexPaths` array that is not sorted it can mess up the whole reload process...

Which optimized variants are you referring to?

appleguy:
@maicki the TwoDimensional functions are dramatically faster than the Multidimensional ones.  It seemed like the code touched here might also benefit from going to a hardcoded 2D version since that is currently the only use case, but I colud be wrong

PR #1668, opened 2016-05-16
[ASVideoNode] Allow playback to continue sooner, even if another re-buffering pause will be required.

Xinchi:
Resume video playback even when playbackLikelyToKeepUp is NO.  Main reason for that is in cases where the loading rate is always less than the bitrate of the video in a poor network condition, the video will never be resumed once enough contents have been buffered until close to the entire video is loaded.  Instead of that, we want to resume playback even though further re-buffering is needed.

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -318,8 +318,11 @@ static NSString * const kStatus = @"status";
       }
     }
   } else if ([keyPath isEqualToString:kPlaybackLikelyToKeepUpKey]) {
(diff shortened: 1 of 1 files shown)


hannahmbanana:
@Xinchi @Eke - do you think this would be better as an option rather than the default? It seems like this behavior's desirability would very much depend on the application.

Xinchi:
Hi @hannahmbanana, we can actually close this pull request now because we did something better - allow clients to specify whether the video playback should aggressively resume playback from stall by the "shouldAggressivelyRecoverFromStall" flag in ASVideoNode.h

PR #1673, opened 2016-05-17
[ASDisplayNode] Add automatic measure before layout

maicki:
Normally `measure:` will be called on a `ASDisplayNode` before layout occurs. If this doesn't happen, nothing is going to call it at all.  An experimenting developer probably added a node to the hierarchy directly. We now simply call `measureWithSizeRange:` using a size range equal to whatever bounds were provided to that element. This will make initial experimentation using layoutSpecs much easier.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1074,19 +1074,58 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   ASDisplayNodeAssertMainThread();
   ASDN::MutexLocker l(_propertyLock);
   CGRect bounds = self.bounds;
-  if (CGRectEqualToRect(bounds, CGRectZero)) {
-    // Performing layout on a zero-bounds view often results in frame calculations
-    // with negative sizes after applying margins, which will cause
-    // measureWithSizeRange: on subnodes to assert.
+
+  [self measureNodeWithBoundsIfNecessary:bounds];
+
+  // Performing layout on a zero-bounds view often results in frame calculations
+  // with negative sizes after applying margins, which will cause
+  // measureWithSizeRange: on subnodes to assert.
+  if (!CGRectEqualToRect(bounds, CGRectZero)) {
+    _placeholderLayer.frame = bounds;
+    [self layout];
+    [self layoutDidFinish];
+  }
+}
+
+- (void)measureNodeWithBoundsIfNecessary:(CGRect)bounds
+{
+  // Normally measure will be called before layout occurs. If this doesn't happen, nothing is going to call it at all.
+  // We simply call measureWithSizeRange: using a size range equal to whatever bounds were provided to that element or
+  // try to measure the node with the largest size as possible
+  if (self.supernode == nil && !self.supportsRangeManagedInterfaceState && !_flags.isMeasured) {
+    if (CGRectEqualToRect(bounds, CGRectZero)) {
+      // FIXME: Better log to let developers know that the node was not measured before the layout call and no frame was set
+      NSLog(@"Warning: No size given for node before node was trying to layout itself: %@. Please provide a frame for the node.", self);
+    } else {
+      ASSizeRange measureSizeRange = ASSizeRangeMake(CGSizeZero, bounds.size);
+      if ([self shouldMeasureWithSizeRange:measureSizeRange]) {
+        [self measureWithSizeRange:measureSizeRange];
+      }
+    }
+  }
+}
+
+- (void)layout
+{
+  ASDisplayNodeAssertMainThread();
+  
+  if (!_flags.isMeasured) {
     return;
   }
-  _placeholderLayer.frame = bounds;
-  [self layout];
-  [self layoutDidFinish];
+  
+  [self __layoutSublayouts];
+}
+
+- (void)__layoutSublayouts
+{
+  for (ASLayout *subnodeLayout in _layout.immediateSublayouts) {
+    ((ASDisplayNode *)subnodeLayout.layoutableObject).frame = [subnodeLayout frame];
+  }
 }
 
 - (void)layoutDidFinish
 {
+  // Hook for subclasses
 }
 
 - (CATransform3D)_transformToAncestor:(ASDisplayNode *)ancestor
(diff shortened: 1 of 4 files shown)


examples/Videos/Sample/ViewController.m
```diff
@@ -55,16 +56,6 @@ - (void)viewWillAppear:(BOOL)animated
   [self.view addSubnode:_rootNode];
 }
 
-- (void)viewDidLayoutSubviews
```
appleguy: Being able to delete this is really fantastic!

However, instead of setting the frame in the appearance method, should we instead set it in the viewWillLayoutSubviews method? This may be an important distinction for rotation.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1074,19 +1074,58 @@ - (void)__layout
...
+
+- (void)measureNodeWithBoundsIfNecessary:(CGRect)bounds
+{
+  // Normally measure will be called before layout occurs. If this doesn't happen, nothing is going to call it at all.
+  // We simply call measureWithSizeRange: using a size range equal to whatever bounds were provided to that element or
+  // try to measure the node with the largest size as possible
+  if (self.supernode == nil && !self.supportsRangeManagedInterfaceState && !_flags.isMeasured) {
+    if (CGRectEqualToRect(bounds, CGRectZero)) {
+      // FIXME: Better log to let developers know that the node was not measured before the layout call and no frame was set
+      NSLog(@"Warning: No size given for node before node was trying to layout itself: %@. Please provide a frame for the node.", self);
```
appleguy: This is a pretty good idea, but I believe that we should wrap it in a compile time condition so that it does not run in a production application. If a developer misses this, perhaps in a situation that is common for users but was not encountered during testing, it could result in very significant performance overhead even though the console logs will not be visible.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1074,19 +1074,58 @@ - (void)__layout
...
+    } else {
+      ASSizeRange measureSizeRange = ASSizeRangeMake(CGSizeZero, bounds.size);
+      if ([self shouldMeasureWithSizeRange:measureSizeRange]) {
```
appleguy: Because this is already called in the method implementation, I think we should skip it and unconditionally call the measure method. It will make this code a bit simpler and should have the same behavior.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1074,19 +1074,58 @@ - (void)__layout
...
-    // Performing layout on a zero-bounds view often results in frame calculations
-    // with negative sizes after applying margins, which will cause
-    // measureWithSizeRange: on subnodes to assert.
+
+  [self measureNodeWithBoundsIfNecessary:bounds];
+
+  // Performing layout on a zero-bounds view often results in frame calculations
+  // with negative sizes after applying margins, which will cause
+  // measureWithSizeRange: on subnodes to assert.
+  if (!CGRectEqualToRect(bounds, CGRectZero)) {
```
appleguy: This is probably the right trade-off, but unfortunately it can result in some side effects. Specifically, an element that does not enable clipping would normally be able to show contents it holds even if it is zero sized, but this depends on the layout pass occurring.

This condition should improve efficiency (a bit, in some cases) and is probably best to keep, but we should be aware of this trade-off and consider if it is necessary to document somewhere

appleguy:
@maicki great change!  I'm very excited about this; it's important behavioral polish.  I'm ready to land after a few small updates.

AsyncDisplayKit/ASDisplayNode.mm
maicki: If you refer to the `CGRectEqualToRect` line ... it was already in there, just change the logic around it a bit. What would you recommend to add or where should we document it?

examples/Videos/Sample/ViewController.m
maicki: It should actually not be in `viewWillAppear:` at all. As the get's called every time the view controller appears. I moved the code to `viewDidLoad`.

To your question. As you can see while creating the root node we set the autoresizing mask to flexible width and height so we don't have to adjust the frame after a layout of the view controller subviews. You can also see it if you set a background color to the root node it expands to the whole view also if rotation is happening.

maicki:
@appleguy Addressed all comments and added some comments for things I didn't address in code

AsyncDisplayKit/ASDisplayNode.mm
appleguy: I see, indeed I had missed that it was there before. The issue is pretty subtle and so it is unclear how to effectively documented, so let's allow this one to stay in, until we see at least one example of a real app use case that exposes it.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1074,19 +1077,54 @@ - (void)__layout
...
+    [self layoutDidFinish];
+  }
+}
+
+- (void)measureNodeWithBoundsIfNecessary:(CGRect)bounds
+{
+  // Normally measure will be called before layout occurs. If this doesn't happen, nothing is going to call it at all.
+  // We simply call measureWithSizeRange: using a size range equal to whatever bounds were provided to that element or
+  // try to measure the node with the largest size as possible
+  if (self.supernode == nil && !self.supportsRangeManagedInterfaceState && !_flags.isMeasured) {
```
appleguy: I don't think we should check that there is no parent. It is very possible in practice for an element to be included in manual layout, but not be present in any layout specification for its parent.  @maicki it is possible I am overlooking something that you have considered, feel free to let me know if you think we should keep that part of the condition.

I actually have exactly the same question about the range managed check. Assuming an element is excluded from the layout specification, perhaps accidentally.  It would be better for something to ultimately trigger a measure pass even if it appears in the upper left of the view by default. This could help developers understand that they forgot to position the view with the layout spec

appleguy:
I am jetlagged and pretty tired, I think I may have misunderstood the dynamic here, in terms of whether or not a a inclusion in a layout specification is relevant for this case. There may not be any changes necessary, but I would love your input on them just to make sure. No action or response needed if you're confident that no changes are advisable.

PR #1674, opened 2016-05-17
[ASCollectionView] Add a "Null Object" for a layout inspector if custom layout is given but no layout inspector

maicki:
The purpose of the ASCollectionViewCustomLayoutInspector is to provide a default implementation if a custom layout is given, but no layout inspector. We remove the necessary `if` handling if no layout inspector is given and replace it with a "Null Object".

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -98,7 +98,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   ASCollectionDataController *_dataController;
   ASRangeController *_rangeController;
   ASCollectionViewLayoutController *_layoutController;
-  ASCollectionViewFlowLayoutInspector *_flowLayoutInspector;
+  id<ASCollectionViewLayoutInspecting> _defaultLayoutInspector;
   NSMutableSet *_cellsForVisibilityUpdates;
   id<ASCollectionViewLayoutFacilitatorProtocol> _layoutFacilitator;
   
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -244,11 +244,19 @@ - (instancetype)_initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionV
...
+    UICollectionViewFlowLayout *layout = (UICollectionViewFlowLayout *)self.collectionViewLayout;
+    ASDisplayNodeAssertNotNil(layout, @"Collection view layout must be a flow layout to use the built-in inspector");
+    _defaultLayoutInspector = [[ASCollectionViewFlowLayoutInspector alloc] initWithCollectionView:self flowLayout:layout];
+  } else {
+    // Custom layouts will need to roll their own ASCollectionViewLayoutInspecting implementation and set a layout
+    // delegate. In the meantime ASDK provides a custom layout inspector that does not provide any implementation
+    // and asserts for methods that should be implemented in the <ASCollectionViewLayoutInspecting>
+    _defaultLayoutInspector = [[ASCollectionViewCustomLayoutInspector alloc] init];
   }
+  _layoutInspector = _defaultLayoutInspector;
```
appleguy: Is the _defaultLayoutInspector ivar necessary or can we make it a local var?

maicki: Because `_layoutInspector` is a weak property

appleguy: @maicki aha, that makes total sense.  Maybe let's call it _strongLayoutInspector so the purpose is more clear.

PR #1676, opened 2016-05-17
[ASVideoNode] API updates

Xinchi:


Xinchi:
New APIs:

*Initial loading starts
*Initial loading ends
*Video stalls at time interval
*Video recovered from stall
*Allow client to specify aggressive recover vs recover only when the playback is likely to reach the end without further stall

Fix bugs:
*Stop loading the video asset multiple times (even though it's called asynchronously and won't crank up network usage wastefully, it might bite us in the end. Also the didStartInitialLoading callback will be called multiple times meaninglessly.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -47,6 +48,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, readwrite) BOOL shouldAutorepeat;
 
 @property (nonatomic, assign, readwrite) BOOL muted;
+@property (nonatomic, assign, readwrite) BOOL shouldAggressivelyRecoverFromStall;
```
appleguy: I think this should be named something else.  "Recover" is not clear.  What does this actually mean / do?  Play again before buffering is even finished?  Should it be a number that scales the aggressiveness in some way?

Please add a comment describing what it does, consider if it should be a float / etc, and think about a more technically descriptive name.  @Xinchi this is important to change in the next few days to minimize folks adopting API that we change.

Overall, this PR looks like _really great work!_  Thanks so much for your contributions, @Xinchi :)

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -96,10 +98,37 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ * @abstract Delegate method invoked when the video is done loading the asset and can start the playback
+ * @param videoNode The videoNode
+ */
+- (void)videoNodeDidFinishInitialLoading:(ASVideoNode *)videoNode;
+/**
+ * @abstract Delegate method invoked when the video node has recovered from the stall
+ * @param videoNode The videoNode
+ */
+- (void)videoNodeDidRecoverFromStall:(ASVideoNode *)videoNode;
```
appleguy: This should probably be DidResumeFrom or something else

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -736,4 +799,4 @@ - (void)dealloc
 }
 
 @end
-#endif
+#endif
```
appleguy: Should always have one blank line at the end of the file

appleguy:
@maicki @yunnanwu @levi @garrettmoon on API design / naming!

PR #1671, opened 2016-05-16
ASVideoPlayerNode Beta

Eke:
##### What is this?!
`ASVideoPlayerNode` is `ASVideoNode` UI wrapper. It is highly customisable and gives developer ability to build UI above `ASVideoNode`. It has default template but gives developer ability to build fully custom UI with delegate methods

#### So what about `ASVideoNode`?
`ASVideoNode` will remain as video playback core. So `ASVideoPlayerNode` will be responsible for UI, `ASVideoNode` for playback

##### videoPlayerNodeNeededControls:
Asks delegate which controls does it need in certain `ASVideoPlayerNode`.  By default all controls are enabled but delegate can override this and use only `ASVideoPlayerNodeControlTypePlaybackButton`.
```
- (NSArray *)videoPlayerNodeNeededControls:(ASVideoPlayerNode *)videoPlayer
{
  return @[ @(ASVideoPlayerNodeControlTypePlaybackButton) ];
}
```
##### videoPlayerNodeLayoutSpec:forControls:forMaximumSize:
Asks delegate for `ASLayoutSpec` which will be used to layout  certain `ASVideoPlayerNode`. With this method developer can  whole layout for `ASVideoPlayerNode` except `ASVideoNode`, it's locked.

```
- (ASLayoutSpec *)videoPlayerNodeLayoutSpec:(ASVideoPlayerNode *)videoPlayer forControls:(NSDictionary *)controls forMaximumSize:(CGSize)maxSize
{
  NSMutableArray *bottomControls = [[NSMutableArray alloc] init];

  ASDisplayNode *playbackButtonNode = controls[@(ASVideoPlayerNodeControlTypePlaybackButton)];

  if (playbackButtonNode) {
    [bottomControls addObject:playbackButtonNode];
  }

  ASLayoutSpec *spacer = [[ASLayoutSpec alloc] init];
  spacer.flexGrow = YES;

  ASStackLayoutSpec *controlbarSpec = [ASStackLayoutSpec stackLayoutSpecWithDirection:ASStackLayoutDirectionHorizontal
  spacing:10.0
  justifyContent:ASStackLayoutJustifyContentStart
  alignItems:ASStackLayoutAlignItemsCenter
  children:bottomControls];
  controlbarSpec.alignSelf = ASStackLayoutAlignSelfStretch;

  UIEdgeInsets insets = UIEdgeInsetsMake(10.0, 10.0, 10.0, 10.0);

[...]

PR #1677, opened 2016-05-18
[ASVideoNode] Expose the placeholderImageNode property of ASVideoNode

gazreese:
@appleguy @Xinchi @Eke @ejensen @maicki @chourobin 

The intention of this is to allow the user of ASVideoNode to set placeholder images independently of the video decoding process. The iOS code doesn't appear to be capable of pulling a placeholder image from an HLS video stream (the video node appears blank until you tap play), and for my project I need placeholders. This also allows the user to set a placeholder that isn't necessarily a frame of the video.

I've updated one of the test apps and this appears to be working fine. It's also working fine in my project. Please let me know if you have any comments / concerns / changes and I'll get it sorted.

### DIFF
--- a/examples/Videos/Sample/ViewController.m
+++ b/examples/Videos/Sample/ViewController.m
@@ -36,6 +36,9 @@
   ASVideoNode *nicCageVideoNode = self.nicCageVideoNode;
   [_rootNode addSubnode:nicCageVideoNode];
   
+  ASVideoNode *placeholderImageVideoNode = self.placeholderImageVideoNode;
+  [_rootNode addSubnode:placeholderImageVideoNode];
+  
   // Video node with custom play button
   ASVideoNode *simonVideoNode = self.simonVideoNode;
   simonVideoNode.playButton = self.playButton;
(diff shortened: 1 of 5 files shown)


appleguy:
@gazreese, thanks for this awesome PR!  We're super fortunate to have benefitted from your contributions and are excited to actively develop this class together.

I believe Max and Yunnan (Xinchi / yunnanwu) have been planning to change the superclass of ASVideoNode to ASNetworkImageNode.  Although this may sound nontraditional, it turns out to provide exactly the desired functionality with exceptionally low overhead.

If a user desires a fixed, static asset placeholder image or to apply one themselves, they could set videoNode.image.  If a thumbnail URL, they simply set videoNode.url, which we could mirror to placeholderImageURL or such.  Even the playButton will properly display over the top of the image content, as would the AVPlayerLayer when added.

This should mean that it's not necessary to set .image / .url = nil even to play the video, although we could do so if we decide to (it would save memory, and the PINRemoteImage fully automatic / default-on memory + disk cache should have us covered.

I will be flying out of town in the next day, but will check in by Monday or so.  Please hop on Slack and pull together a few of the contributors to discuss if we should adapt this PR to change the superclass.  If you're interested to just try this out, I think it should be a straightforward change, mostly deleting the placeholderImageNode!

Cheers,

		-- Scott

gazreese:
Hi @applyguy

Sounds good to me. I’m out of the office travelling today so I’ll have a better look into it tomorrow.

examples/Videos/Sample/ViewController.m
```diff
@@ -97,6 +104,26 @@ - (ASVideoNode *)nicCageVideoNode;
   return nicCageVideoNode;
 }
 
+// HLS video with a placeholder image
+- (ASVideoNode *)placeholderImageVideoNode;
```
Eke: ; character at the end on method declaration. typo?

Eke:
This is good idea to let users set placeholder image, but @appleguy is correct , there were some discussions about making `ASVideoNode` subclass to `ASNetworkImageNode` but i think at this moment this PR will be useful for users.

One more thing is that `ASVideoNode` should be responsible only for playback, we are currently working on `ASVideoPlayerNode` (#1671). Maybe it will be better to move placeholder functionality to `ASVideoPlayerNode`?

appleguy:
@gazreese thanks again for building this.  I'd merge it as a next step, but it looks like we'd have to remove / rewrite the test and much of the rest of the code, so it probably makes more sense to do the subclass directly from today's master.

Let us know if you won't have time to do this in the next week so others in the community can jump in!

appleguy:
@gazreese hey-o — checking in.  If you won't be able to get around to this, let us know so another community member can be confident they won't cross paths with your work and post a redundant PR around the same time.  Cheers!

gazreese:
Sorry, I didn’t see this in my email for some reason.

I could probably spend a bit of time on this next week but I’m finishing off another project at the moment, so it might be longer.

appleguy:
@gazreese @hannahmbanana got to this one!  https://github.com/facebook/AsyncDisplayKit/pull/1710

gazreese:
Looks good, will try the new superclass functionality next week

PR #1679, opened 2016-05-18
[ASLayout] Rename ASLayoutable to ASLayoutProducer

levi:
Currently a proposal I want to float to the team. Looks large, but was just a simple find + replace.

The sole job of a ASLayoutable is to create an ASLayout. This is
performed through passing a constrained size to `measureWithSizeRange:`,
which results in an calculated ASLayout. This ASLayout describes the
size and, optionally, position of the ASLayoutable.

Changing the name of ASLayoutable to ASLayoutProducer better describes
the single responsibility of an ASLayoutable. ASLayoutProducers produce
valid ASLayouts through `measureWithSizeRange:`. ASLayout objects now
have a reference to their original ASLayoutProducer through the
`layoutProducer` property, which further clarifies the value of the
currently ambiguous property `layoutableObject`.

`finalLayoutable` and `isFinalLayoutable` have been renamed to
`finalLayoutProducer` and `isFinalLayoutProducer`. Moreover, the
ASLayoutable sub protocols (ie. ASStaticLayoutable) have been renamed
to align with the LayoutProducer naming (ie. ASStaticLayoutProducer).


levi:
Actually, the world is not quite ready for this.

appleguy:
@levi lol!  This is actually very compelling, and I think the name is quite good.  @rcancro what do you think?  Any other name ideas?

rcancro:
`ASLayoutComposer`? Either way I think this is probably an improvement.

nguyenhuy:
Love this diff ❤️

levi:
Yeah it's going I the right direction, but I'm thinking of applying some more ground breaking changes for ASLayout before a new scheme of names.

PR #1680, opened 2016-05-19
[ASDisplayNode] Rename ASDisplayNodeTransitionContext to ASLayoutTransition

levi:
Giving the public context object a little more...context. As we complete the layout transition API, this object will be a key facilitator in not just state, but also behavior during the transition duration.


nguyenhuy:
👍

PR #1683, opened 2016-05-19
expose relayoutAllNodes at ASCollectionView level

nickvelloff:
The purpose of this PR is to expose `relayoutAllNodes` from `ASCollectionDataController` to `ASCollectionView` to allow us to manually trigger size changes. Particularly useful during an interactive transition where cells are resized by some try of gesture.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -168,7 +168,7 @@ NS_ASSUME_NONNULL_BEGIN
  * Triggers a relayout of all nodes.
  *
  */
-- (void)relayoutAllNodes;
+- (void)relayoutItems;
 
 /**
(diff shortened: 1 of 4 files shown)


nickvelloff:
cc @nguyenhuy

nickvelloff:
I signed the CLA BTW 😸

nguyenhuy:
At ASDataController level, relayoutAllNode is fine. However, I'm wondering maybe we should have a different name for this API, to be consistent with UIKit naming convention. What do you think about relayoutData or relayoutItems?

Also, should we expose the same API for table view?

nickvelloff:
@nguyenhuy Yes agree on both. I like `relayoutItems` personally.

nickvelloff:
@nguyenhuy ready for review. 

Of course it's the least I can do 👍

nguyenhuy:
LGTM, thanks!

nickvelloff:
👍 Awesome!

PR #1684, opened 2016-05-19
May as well use the existing visibility function.

garrettmoon:
These are equivalent.


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -941,7 +941,7 @@ - (void)setShouldRasterizeDescendants:(BOOL)shouldRasterize
       [self _addSubnodeViewsAndLayers];
     }
     
-    if (self.interfaceState & ASInterfaceStateVisible) {
+    if (ASInterfaceStateIncludesVisible(self.interfaceState)) {
```
appleguy: Nice.  If we hold the lock here, we should use the instance variable, but nbd.

PR #1686, opened 2016-05-20
Allow nil borderColor when using ASImageNodeRoundBorderModificationBlock

smeis:
When using this block the border is optional. However, currently it is necessary to pass a non-nil value for `borderColor` even if `borderWidth` is 0.


appleguy:
@smeis definitely intended, thanks!

PR #1687, opened 2016-05-20
[ASDisplayNode] Move constrained size to ASLayout

levi:
I have this grand master plan of turning our use of ASLayouts into something like a browser render tree, a tree that every node queries when it wants to know its own size and position on screen. This is a change to move into that direction, as constrainedSize ultimately belongs in ASLayout because ASLayoutables (objects that create ASLayouts) all pass a constrainedSize when they generate a layout.

Moreover, I've removed the legacy concept of a node being "measured". Instead, ASLayouts can now be marked as dirty, which then is used to replace the `isMeasured` bit flag. As we mature the underlying ASLayout tree structure, we should move away from "measurement" and begin to use nomenclature that describes generating layouts and rendering nodes.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -650,10 +650,15 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     }
   }
   
-  // only calculate the size if
-  //  - we haven't already
-  //  - the constrained size range is different
-  return (!_flags.isMeasured || !ASSizeRangeEqualToSizeRange(constrainedSize, _layout.constrainedSizeRange));
+  // Only generate a new layout if:
+  // - The current layout is dirty
+  // - The passed constrained size is different than the layout's constrained size
+  return ([self _dirtyLayout] || !ASSizeRangeEqualToSizeRange(constrainedSize, _layout.constrainedSizeRange));
+}
+
+- (BOOL)_dirtyLayout
+{
+  return _layout == nil || _layout.isDirty;
 }
 
 #pragma mark - Layout Transition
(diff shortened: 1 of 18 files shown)


appleguy:
Cool beans.  @nguyenhuy or @maicki could you review and let me know when ready to merge?

I assume this would be a step in the direction of having a mode where we can re-run an existing ASLayoutSpec with a new constrained size, without calling layoutSpecThatFits?  This is actually the key / main capability needed to unblock @hannahmbanana's layout spec playground test app.

levi:
@appleguy that's definitely something we could support with the design I'm working on. Ideally we should be taking advantage of this backing ASLayout tree for all aspects of a node's position and size.

appleguy:
@levi rebase when you get a chance.  Let's talk about the long term plan soon!

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -648,10 +643,15 @@ - (BOOL)shouldMeasureWithSizeRange:(ASSizeRange)constrainedSize
...
   }
   
-  // only calculate the size if
-  //  - we haven't already
-  //  - the constrained size range is different
-  return (!_flags.isMeasured || !ASSizeRangeEqualToSizeRange(constrainedSize, _constrainedSize));
+  // Only generate a new layout if:
+  // - The current layout is dirty
+  // - The passed constrained size is different than the layout's constrained size
+  return ([self _dirtyLayout] || !ASSizeRangeEqualToSizeRange(constrainedSize, _layout.constrainedSizeRange));
```
appleguy: This should be called _layoutIsDirty or something else; it reads like an action / verb currently

appleguy:
This is a very smart change!

levi:
@appleguy rebased and ready to go!

PR #1688, opened 2016-05-21
[Swift] Add missing nullability annotations to transition methods

ejensen:
There may be more public headers missing nullability annotations, but these are the ones I noticed while using them.


appleguy:
Thanks @ejensen !

PR #1691, opened 2016-05-23
WIP - Two different approaches to trigger an update to nodes constrained size

nickvelloff:
The purpose of this PR is to explore approaches to invalidating and remeasuring a cell node's constrained size. Particularly useful during an interactive transition where cells are resized by some type of gesture.

Approach 1: 
- expose `- (void)relayoutAllNodes` in `ASCollectionDataController` to `ASCollectionView` and `ASTableView`
-  call `relayoutItems()` in `ASCollectionView` to trigger all nodes to force their size to invalidate and remeasure

Approach 2: 
- in `ASDisplayNode`, modify `__setNeedsLayout` so `oldConstrainedSize` checks new `[self __constrainedSizeForLayout]`
- in `ASDisplayNode` private `__constrainedSizeForLayout` returns default `_constrainedSize`
- in `ASCellNode` we override `__constrainedSizeForLayout` to return new `ASSizeRange` calling new `constrainedSizeForLayoutWithOldConstrainedSize` delegate method in `ASCellNodeDelegate`
- `ASCollectionView` implements `ASCellNodeDelegate` `constrainedSizeForLayoutWithOldConstrainedSize` delegate method and returns `_asyncDataSource` `constrainedSizeForNodeAtIndexPath` effectively getting the new constrained size from the user's `collectionView` `constrainedSizeForNodeAtIndexPath` `ASCollectionDataSource` delegate implementation 
- User just needs to call `setNeedsLayout()` on any `ASCellNode` to force it's size to invalidate and remeasure

Findings so far:
Approach 1 works well, but may be a bit coarse a solution and perhaps more expensive than it needs to be.
Approach 2 _almost_ works, and is a more elegant solution but it appears to be fighting something else listening to the invalidation, that is getting it wrong. More research to do here.

Demo:
Points to my ASDK fork with these changes.
git@github.com:the-grid/zoom-demo.git
Branch `relayoutAllNodes` uses approach 1
Branch `setNeedsLayout` uses approach 2

cc/ @nguyenhuy @appleguy for all the guidance. I think we are getting close here.

### DIFF
--- a/AsyncDisplayKit/ASCellNode+Internal.h
+++ b/AsyncDisplayKit/ASCellNode+Internal.h
@@ -23,6 +23,15 @@
  */
 - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged;
 
+/**
+ * Notifies the delegate that the specified cell node has a new constrained size.
+ * The notification is done on main thread.
+ *
+ * @param cellNode A node informing the delegate about the it's size.
+ * @param oldConstrainedSize The previous constrained size of the node.
+ */
+- (ASSizeRange)cellNode:(ASCellNode *)node constrainedSizeForLayoutWithOldConstrainedSize:(ASSizeRange)oldConstrainedSize;
+
 @end
 
 @interface ASCellNode ()
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -121,6 +121,11 @@
   ASDisplayNodeAssert(!layerBacked, @"ASCellNode does not support layer-backing.");
 }
 
+-(ASSizeRange)__constrainedSizeForLayout
+{
+  return [_layoutDelegate cellNode:self constrainedSizeForLayoutWithOldConstrainedSize:_constrainedSize];
+}
+
 - (void)__setNeedsLayout
 {
   CGSize oldSize = self.calculatedSize;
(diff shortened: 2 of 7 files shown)


appleguy:
@levi @maicki @nguyenhuy any thoughts here?  I'm hustling fast on many things, and worry that it could be too long before I can properly understand the tradeoffs here.  We should not have PRs standing open for a long time.

nickvelloff:
@appleguy Approach 1 was already merged in another PR.
Approach 2 was based on a discussion I had with you on Slack, as an alternative. 
This PR was intended to not let the conversation get lost. 
That said happy to close it if its the wrong time to explore.

appleguy:
Thanks for summarizing, Nick!  Let's leave this open for now, then.  I recall the solution now and the performance concern around that, so it's worth evaluating the alternative approach you suggested even though the implementation is incomplete.

nickvelloff:
Thanks @appleguy ... With approach 2 the implementation is incomplete because I'm not exactly sure what might be _fighting_ with the layout. Seems like a scroll offset conflict. I think the approach you suggested may be a better performer, but will require a bit more thought.

Adlai-Holler:
Both approaches are now in! Calling -setNeedsLayout on a cell node now does a good job of ensuring the node gets remeasured and the layout updated, either with animation (if `animatesSizeChanges=YES` all the way down the cell node's tree) or without (otherwise). Thanks @nickvelloff @appleguy

PR #1692, opened 2016-05-23
[ASLayout] Add basic infrastructure for layout validation

maicki:


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1901,9 +1912,13 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
     layoutSpec.isMutable = NO;
     ASLayout *layout = [layoutSpec measureWithSizeRange:constrainedSize];
     // Make sure layoutableObject of the root layout is `self`, so that the flattened layout will be structurally correct.
-    if (layout.layoutableObject != self) {
+    BOOL isRootLayout = (layout.layoutableObject != self);
+    if (isRootLayout) {
```
levi: Was looking at this the other day. In what case will we get an ASLayout returned from the layoutSpec where the layoutableObject is self? Seems like only via returning `self` from ASLayoutSpec that fits.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1901,9 +1912,13 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
...
       layout.position = CGPointZero;
       layout = [ASLayout layoutWithLayoutableObject:self size:layout.size sublayouts:@[layout]];
+#if DEBUG
+      [self validateLayout:layout];
```
levi: why not validate layout on initialization?

AsyncDisplayKit/ASDisplayNode.mm
appleguy: Yeah I think the original code here was meant to make it the case that the layoutableObject is self, even though it isn't when returned by the layout spec.  The other case might occur in cases like returning a new node in finalLayoutable like is done with the layout visualizer or certain BRIO elements?

appleguy:
@maicki this is really fantastic, and will allow critical improvements in the developer friendliness of the layout system, By pointing developers to appropriate documentation when it becomes apparent that they are using properties which are not intended for the situation.

I think it is ready to land, however it needs to be rebased, and I would also ask that you put in a flag that sets it to disabled for now. The reason is that we should run it in Pinterest and ask a handful of other developers to try it out to see how many issues it is catching, and to make a decision on whether some of them should be moved into a strict validation mode.

In other words, as one example, I think it is at least possible that valid code sets a few extraneous values which take effect when an element is used in a stack — even if in some other cases it is used outside of the stack (intentionally).

Very cool idea to build this in an extensible way, and I look forward to thinking about expanding this and integrating it directly with the documentation (Hannah can likely help)

maicki:
@appleguy New changes:
- Assert if layout is invalid
- Improved error messages
- Flag to turn layout validation on and off
- Moved all layout code to `ASLayoutValidation` and created a helper function to invoke a full layout validation

AsyncDisplayKit/Layout/ASLayoutValidation.h
```diff
@@ -0,0 +1,82 @@
...
+#import <AsyncDisplayKit/ASBaseDefines.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class ASLayout;
+
+// Enable or disable automatic layout validation
+#define LAYOUT_VALIDATION 0
+
+extern void ASLayoutableValidateLayout(ASLayout *layout);
```
ocrickard: Sorry, may be missing something but don't you need an ASDISPLAYNODE_EXTERN_C_BEGIN/END around this or the C++ mangling will cause normal .m files that import to fail to find the symbol at link time, right?

AsyncDisplayKit/Layout/ASLayoutValidation.mm
```diff
@@ -0,0 +1,223 @@
...
+
+@implementation ASLayoutableValidation
+
+#pragma mark Lifecycle
+
+- (instancetype)init
+{
+  self = [super init];
+  if (self) {
+    _validators = [NSMutableArray array];
```
ocrickard: Any reason we can't use a NSMutableSet/NSCountedSet instead of an array? Each of the searches below turns into O(N). May not matter.

AsyncDisplayKit/Layout/ASLayoutValidation.h
```diff
@@ -0,0 +1,82 @@
...
+@interface ASLayoutableStaticValidator : NSObject<ASLayoutableValidator>
+
+@end
+
+/*
+ * ASLayoutables that have spacingBefore, spacingAfter, flexGrow, flexShrink, flexBasis, alignSelf, ascender or descender
+ * set needs to be wrapped into a ASStackLayout. This validator checks if sublayouts has set one of this properties and
+ * asserts if it's not wrapped in a ASStackLayout if so.
+ */
+@interface ASLayoutableStackValidator : NSObject<ASLayoutableValidator>
```
ocrickard: This is super cool

AsyncDisplayKit/Layout/ASLayoutValidation.h
```diff
@@ -0,0 +1,82 @@
...
+
+@end
+
+
+#pragma mark - ASLayoutableValidation
+
+@interface ASLayoutableValidation : NSObject
+
+/// Currently registered validators
+@property (copy, nonatomic, readonly) NSMutableArray<id<ASLayoutableValidator>> *validators;
```
ocrickard: Can we avoid exposing a mutable array in the public header of this?

AsyncDisplayKit/ASDisplayNode.mm
levi: finalLayoutable makes sense.

AsyncDisplayKit/Layout/ASLayoutValidation.h
maicki: Yep you are right! Currently it's only imported into `ASDisplayNode.mm` that is compiled with C++, but we should  wrap it in ASDISPLAYNODE_EXTERN_C_BEGIN/END.

AsyncDisplayKit/Layout/ASLayoutValidation.mm
maicki: Currently we just iterate through it below. Iterating through arrays / vectors should be faster as using set's. NSMutableSet/NSCountedSet would be preferable if we do a lot of "contains" checks as it's based on a hash map. Further information also: [Sets: Unordered Collections of Objects](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Collections/Articles/Sets.html#//apple_ref/doc/uid/20000136-CJBDHAJD)

Please let me know in case I'm missing something.

AsyncDisplayKit/Layout/ASLayoutValidation.h
maicki: Definitely! This is an overside and we should not expose a `NSMutableArray` here.

maicki:
@ocrickard Thanks for the review. I addressed your comments in the last commit.

AsyncDisplayKit/Layout/ASLayoutValidation.mm
ocrickard: Yeah, totally, perhaps it was unintentional, but I thought below you're using removeObject: on the array. This will enumerate over all elements in the array and compare before removing. This means that it is O(N). Agreed, your other callsites are fine. If you don't have many objects in the array it shouldn't matter though.

AsyncDisplayKit/Layout/ASLayoutValidation.mm
```diff
@@ -0,0 +1,223 @@
...
+- (id<ASLayoutableValidator>)registerValidatorWithBlock:(ASLayoutableBlockValidatorBlock)block
+{
+  ASLayoutableBlockValidator *blockValidator = [[ASLayoutableBlockValidator alloc] initWithBlock:block];
+  [self.validators addObject:blockValidator];
+  return blockValidator;
+}
+
+- (void)unregisterValidator:(id<ASLayoutableValidator>)validator
+{
+  [self.validators removeObject:validator];
```
ocrickard: here

AsyncDisplayKit/Layout/ASLayoutValidation.mm
appleguy: @ocrickard @maicki we should definitely prefer removeObjectIdenticalTo: where possible.  But yeah, for this use case, arrays are probably fine.  The validator should definitely only default to on in DEBUG because we wouldn't want to crash a production app anyway, but a dev app should throw an exception to warn the engineer about likely unintentional usage.

ocrickard: Sounds great.

PR #1693, opened 2016-05-23
[ASTraitCollection] Bug fixing/LayoutSpec propagation

rcancro:
* Fixed bug in `ASEnvironmentMergeObjectAndState`
* New ASLayoutSpec methods for `setChild`/`setChildren`/`setChild:forIdentifier:` have been added to take in a trait collection.
* Added `setChild:`-like methods for ASLayoutSpecs take a traitCollection
* Fixed instances where nodes in a data controller were not getting their trait collections
* propagate traitCollection in ASDisplayNode on insertSubnode or addSubnode
* Re-added the ASTraitCollection example that mysteriously got deleted.


levi:
@rcancro I'd love to figure out a way to keep `setChild` and `setChildren` as is and somehow have the trait collection passed implicitly through the layout spec hierarchy. Problem is that the layout spec is always user-defined, so there's no good way to have the layout spec associated with the node implicitly when constructed via `calculateLayoutSpecThatFits:`. It's not an easy problem to solve from first glance, but I am thinking through some possible solutions. Will post some thoughts when I have them.

rcancro:
@levi @appleguy Can you take a look at these updates?

appleguy:
I'd love to land as soon as any reviewer gives a green light!  Send me a ping when so.

levi:
👍  LOOKS GOOD! Merge it, @appleguy

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1249,6 +1249,9 @@ - (void)addSubnode:(ASDisplayNode *)subnode
   // If we are a managed hierarchy, as in ASCellNode trees, it will also apply our .interfaceState.
   [subnode __setSupernode:self];
 
+  // propagate our traits to the child we are about to add.
+  ASEnvironmentStatePropagateDown(subnode, [self environmentTraitCollection]);
```
appleguy: would it make more sense to do this in _setSupernode: ?  If we add additional call points, this could be left out accidentally.  It would seem to be an action that should take place whenever a parent is added.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2721,6 +2728,11 @@ - (ASEnvironmentTraitCollection)environmentTraitCollection
   return _environmentState.traitCollection;
 }
 
+- (void)setEnvironmentTraitCollection:(ASEnvironmentTraitCollection)environmentTraitCollection
```
appleguy: The naming seems a bit inconsistent - we have both asyncTraitCollection and environmentTraitCollection.  Each name seems great in the situation it is used, but we should consider if a quick refactoring can make this any clearer.

The line above has a similar issue.  layoutSpec.traitCollection is of a different type than node.traitCollection.

AsyncDisplayKit/ASViewController.mm
```diff
@@ -82,6 +83,13 @@ - (void)loadView
   _node.frame = frame;
   _node.autoresizingMask = autoresizingMask;
   self.view = view;
+  
+  // ensure that self.node has a valid trait collection before a subclass's implementation of viewDidLoad.
+  // Any subnodes added in viewDidLoad will then inherit the proper environment.
```
appleguy: This seems worthwhile as a performance optimization. However, both for clarity of my understanding and of the description, would it not be the case that Setting a trait collection after the children are added would trigger a propagation of the new values?

AsyncDisplayKit/ASViewController.mm
```diff
@@ -82,6 +83,13 @@ - (void)loadView
...
+  if (AS_AT_LEAST_IOS8) {
+    ASEnvironmentTraitCollection traitCollection = [self displayTraitsForTraitCollection:self.traitCollection];
```
appleguy: The terminology here is not clear to me ("display traits") and should be considered with the comment above.  Is this essentially asyncTraitsForTraitCollection: or environmentTraitsForTraitCollection?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -527,16 +538,8 @@ - (void)accessDataSourceSynchronously:(BOOL)synchronously withBlock:(dispatch_bl
...
-      ASCellNodeBlock nodeBlockPropagatingDisplayTraits = ^{
-        ASCellNode *cellNode = nodeBlock();
-        id<ASEnvironment> environment = [self.environmentDelegate dataControllerEnvironment];
-        ASEnvironmentStatePropagateDown(cellNode, [environment environmentTraitCollection]);
-        return cellNode;
-      };
-      
       ASSizeRange constrainedSize = [self constrainedSizeForNodeOfKind:ASDataControllerRowNodeKind atIndexPath:indexPath];
-      [contexts addObject:[[ASIndexedNodeContext alloc] initWithNodeBlock:nodeBlockPropagatingDisplayTraits
+      [contexts addObject:[[ASIndexedNodeContext alloc] initWithNodeBlock:[self nodeBlockPropagatingTraitCollectionFromNodeBlock:nodeBlock]
```
appleguy: These codepaths are extremely hot - thus nasty stuff like manually malloc'ing memory and creating NSArrays :).  We should not wrap an additional block as it adds retain / release overhead (I believe it will not only retain the block, but also everything referenced inside it - not certain about the latter piece).

It seems like we could instead implement the propogation step at the point where the existing nodeBlock is called rather than swapping out the block implementation.

AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -154,13 +156,15 @@ ASDISPLAYNODE_EXTERN_C_END
   [super setEnvironmentState:environmentState];\
   ASEnvironmentTraitCollection currentTraits = environmentState.traitCollection;\
   if (ASEnvironmentTraitCollectionIsEqualToASEnvironmentTraitCollection(currentTraits, oldTraits) == NO) {\
-    NSArray<NSArray <ASCellNode *> *> *completedNodes = [self.view.dataController completedNodes];\
-    for (NSArray *sectionArray in completedNodes) {\
-      for (ASCellNode *cellNode in sectionArray) {\
-        ASEnvironmentStatePropagateDown(cellNode, currentTraits);\
-        [cellNode setNeedsLayout];\
+    dispatch_async(dispatch_get_main_queue(), ^{\
```
appleguy: adding a dispatch_async is a material change; could you elaborate on why this is required?  It might be fine, or it might introduce cases where additional traffic (e.g. layout passes) are caused.  It would definitely prevent the possibility from piggybacking on any current setNeedsLayout that the cellNode already has scheduled, since any currently-dirtied layout will execute before the block does.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -156,14 +162,21 @@ - (void)setChildren:(NSArray *)children
 
 - (id<ASLayoutable>)child
 {
-  return _child;
+  return [_children firstObject];
```
appleguy: The previous code would not crash if _child was nil; this case may not have been permitted by assertions, but it would succeed at least in production builds.  We should ensure this does too.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -113,28 +112,35 @@ - (void)setParent:(id<ASLayoutable>)parent
...
 
 - (void)setChild:(id<ASLayoutable>)child forIdentifier:(NSString *)identifier
 {
   ASDisplayNodeAssert(self.isMutable, @"Cannot set properties when layout spec is not mutable");
+  ASDisplayNodeAssertNotNil(child, @"Child cannot be nil");
   
   id<ASLayoutable> finalLayoutable = [self layoutableToAddFromLayoutable:child];
   self.childrenWithIdentifier[identifier] = finalLayoutable;
+  if (finalLayoutable) {
+    self.children = [self.children arrayByAddingObject:finalLayoutable];
```
appleguy: Hmm, I don't really understand this code.  If I set a child for an existing identifier, it should not be appended to the children array, right?  Do we even need both?

I think we should consider separating out these changes.  I'd love to reduce dependence on the dictionary in this class, as it shows up materially in profiles, but I'm a bit worried about some nonintuitive risks here that only rarer layout code exposes.

Anyway, once we've confirmed that unconditionally adding the object is correct, two quick things:
- If self.children is nil, it's not clear to me this will work - may need to alloc the array
- we should not use self.children and instead use _children, because self.children creates a copy, which we're about to copy again by adding an object.

This code needs to be locked in the near future, right?  Maybe add a FIXME somewhere.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -113,28 +112,35 @@ - (void)setParent:(id<ASLayoutable>)parent
...
   
   // TODO: Should we propagate up the layoutable at it could happen that multiple children will propagated up their
   //       layout options and one child will overwrite values from another child
   // [self propagateUpLayoutable:finalLayoutable];
 }
 
-- (void)setChildren:(NSArray *)children
+- (void)setChildren:(NSArray<id<ASLayoutable>> *)children
 {
   ASDisplayNodeAssert(self.isMutable, @"Cannot set properties when layout spec is not mutable");
```
appleguy: We should probably either add an assertion against setting nil children, or likely better, remove the assertion against setting nil children.  I'm not sure it should be disallowed and is a common issue for developers.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -156,14 +162,21 @@ - (void)setChildren:(NSArray *)children
 
 - (id<ASLayoutable>)child
 {
-  return _child;
+  return [_children firstObject];
 }
 
 - (NSArray *)children
 {
   return [_children copy];
```
appleguy: Is this necessary to copy at all, since the array isn't mutable anyway?

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -204,6 +217,11 @@ - (ASEnvironmentTraitCollection)environmentTraitCollection
   return _environmentState.traitCollection;
 }
 
+- (void)setEnvironmentTraitCollection:(ASEnvironmentTraitCollection)traitCollection
+{
+  _environmentState.traitCollection = traitCollection;
```
appleguy: Shouldn't this also do a state propagation?

In fact, it seems like we should not have both _traitCollection and _environmentState.traitCollection.  Couldn't they get out of sync?  Can we standardize on one?

AsyncDisplayKit/Private/ASEnvironmentInternal.h
```diff
@@ -58,14 +58,14 @@ ASEnvironmentState ASEnvironmentMergeObjectAndState(ASEnvironmentState environme
 template <typename ASEnvironmentStateType>
 void ASEnvironmentStatePropagateDown(id<ASEnvironment> object, ASEnvironmentStateType state) {
   ASEnvironmentPerformBlockOnObjectAndChildren(object, ^(id<ASEnvironment> node) {
-    object.environmentState = ASEnvironmentMergeObjectAndState(object.environmentState, state, ASEnvironmentStatePropagation::DOWN);
+    node.environmentState = ASEnvironmentMergeObjectAndState(node.environmentState, state, ASEnvironmentStatePropagation::DOWN);
```
appleguy: Crazy important find, @rcancro  - I'm sure this was a pain to spot and likely required debugging carefully.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
rcancro: `firstObject` won't crash. It will return nil.
https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSArray_Class/#//apple_ref/occ/instp/NSArray/firstObject

appleguy: Ah, very cool!

AsyncDisplayKit/ASDisplayNode.mm
rcancro: Oh, I like that idea :)

AsyncDisplayKit/ASDisplayNode.mm
rcancro: Good point.

AsyncDisplayKit/ASViewController.mm
rcancro: From what I could tell, `traitCollectionDidChange:` is not called by UIKit until after `viewDidLoad`. However `self.traitCollection` is valid in `viewDidLoad` in UIKit. To try to be consistent, I wanted to make sure that anyone that subclasses `ASViewController` will have a `self.node` with a valid `asyncTraitCollection` in `viewDidLoad`. Without this propagation we'd have to wait for UIKit's `traitCollectionDidChange` call before self.node has a valid `asyncTraitCollection`.

AsyncDisplayKit/ASViewController.mm
rcancro: This is actually an oversight from when I was calling these displayTraits instead of traitCollections. I'll fix that up.

AsyncDisplayKit/Details/ASEnvironment.h
rcancro: The main issue is that I am accessing `[self.view.dataController completedNodes]` which must be on a main thread. Should I get that value on the main thread, then jump back to a bg thread?

AsyncDisplayKit/Details/ASDataController.mm
rcancro: sure thing. Made `ASIndexedNodeContext` take an `ASEnviromentTraitCollection`and propagate on creation.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
rcancro: This isn't great. Now that `ASLayoutSpec` conforms to `ASEnvironment`, `children` is expected to return all of the subspecs. The dictionary is around because subclasses of `ASLayoutSpec` like `ASBackgroundLayoutSpec`. Instead of the convention of `_children[0]` is foreground and `_children[1]` is background, I created a dictionary to save away the children based on static strings defined in the subclass. 

Now that I'm adding all subspecs to `_children` anyway, the dictionary is around for little more than a look up. So, to answer your question, adding to the children array here is correct. BUT you should hound me to put up another PR cleaning this up so that we don't need the dictionary.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
rcancro: No -- I bet the array used to be mutable :/

AsyncDisplayKit/Layout/ASLayoutSpec.mm
rcancro: @maicki and I talked about this. Currently ASLayoutSpec is wrapped with `NS_ASSUME_NONNULL_BEGIN` and we thought that it didn't make a lot of sense to send in nil children.

But interestingly enough, I noticed that ASBackgroundSpec has the background spec as nullable. Soooooo, not sure what we should do here. @maicki ?

AsyncDisplayKit/Layout/ASLayoutSpec.mm
rcancro: Setting self.environmentState does not cause propagation, so I didn't think this should. This method actually gets called when propagating, so propagating here too would lead to crazy double nested propagation.

Originally there was only the struct based traitCollection. @maicki suggested that we have a nice NSObject based class so that the API was friendly to the end user. That's how ASTraitCollection was born. 

ASEnvironmentTraitCollection sticks around because we are using the ASEnvironment propagation which requires a struct. While I personally like dealing with the struct directly, I think Michael's thought is a noble one: To make these new API more accessible so that people that don't want to deal with a struct don't have to.

examples/ASTraitCollection/Podfile
```diff
@@ -0,0 +1,5 @@
+source 'https://github.com/CocoaPods/Specs.git'
```
rcancro: OHHHHHH!!! Did some example projects get moved to an examples_extra dir? Is that why my example app disappeared?! HA! I thought it got deleted by accident on a weird merge or something. 

Before I do anything drastic, I want to confirm: I should remove the example from the `examples` folder and update the one in `examples_extra`?

AsyncDisplayKit/Details/ASEnvironment.h
appleguy: @rcancro ah, one improvement then would be to use the ASPerformBlockOnMain() which will run synchronously if already running on main.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
appleguy: @rcancro because editing is not common, we should probably keep it immutable and remove the copy

AsyncDisplayKit/Layout/ASLayoutSpec.mm
appleguy: @rcancro that all makes sense - I definitely agree that ASTraitCollection should be the public-facing API, and the struct is an implementation technique chosen for a variety of advantages.

However, I am confused at the existence of a dedicated ivar for _traitCollection that appears to be able to get out of sync with the struct.  My assumption with the strategy to create an object-based trait collection was that it would be created on demand when accessed, and if it were cached at all, would at least be nil'd out when the underlying struct changes (then regenerated upon the next access of the object).

As it is written now (not only this method but also looking at the one above), wouldn't it be possible to access the two types of trait collections, compare the values, and find some are not equal?  This seems like it should not be possible.

AsyncDisplayKit/ASDisplayNode.mm
appleguy: :-D @rcancro this one is fine to do in a later PR.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1898,6 +1904,7 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
   if ((_methodOverrides & ASDisplayNodeMethodOverrideLayoutSpecThatFits) || _layoutSpecBlock != NULL) {
     ASLayoutSpec *layoutSpec = [self layoutSpecThatFits:constrainedSize];
     layoutSpec.parent = self; // This causes upward propogation of any non-default layoutable values.
+    layoutSpec.traitCollection = self.asyncTraitCollection;
```
appleguy: @rcancro the idea I had in mind with an object and struct based representation is that the common path internal implementation would always use the struct, and the object would exist almost soley for API surface area that third party code accesses.  This would mean that very few nodes create the object - important because there are a truly vast number of nodes, due to the intentional design decision where ASDataController relies on them being cheap, and creates entire cell subtrees for every datasource element.

If that makes sense, we should probably change this usage point / layoutSpec to use environment.  This is the only clear case I see that would "accidentally" trigger the creation, as the rest are in test project code (perfect & legitimate usages because they are "third party" code simulations).

AsyncDisplayKit/Details/ASEnvironment.h
appleguy: @rcancro does completedNodes currently assert that it should be called on the main thread?  If it does not but that is required, we should add that...

Because any transactions are performed on a different array, it may actually the be straightforward to add a walk and allow multithreaded access to this.  @maicki knows that code very well and should consider if this is indeed a simple change, in which case he may be able to remove this thread trampoline for us later.

Probably not a top priority, but I don't think I will have an opportunity to fully consider the implications of the trampoline, so let me know if there is any hypothesis for how it could introduce a problem.

For example, could there be edit transactions scheduled to occur on the next run loop that completely rewrite the data source and all of the nodes, meaning that we should not perform this operation? What happens if this operation is triggered from multiple threads, Is it possible that they will race in a way that cannot be controlled by Application level synchronization?

AsyncDisplayKit/ASViewController.mm
appleguy: OK, that makes perfect sense.  @rcancro You could slightly tweak the comment to clarify the main value of this is to ensure the trait collection is valid as early as possible rather than waiting for the later UIKit call, but that technically traitCollectionDidChange: will always be called at least once subsequently.

AsyncDisplayKit/Layout/ASLayoutSpec.h
```diff
@@ -23,6 +25,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign) BOOL isMutable;
 
+@property (nonatomic, strong, nullable) ASTraitCollection *traitCollection;
```
appleguy: @rcancro this is probably the main one I'd change usage of.  This could potentially be a create-on-demand property but should probably be readonly.  The writing path could use environment trait collection, which since this already conforms to ASEnvironment, it at least has the struct.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -45,7 +44,7 @@ - (instancetype)init
   }
   _isMutable = YES;
   _environmentState = ASEnvironmentStateMakeDefault();
-  
+  _children = [NSArray array];
```
appleguy: @rcancro lolol.  I think this is cheating a bit instead of just checking for nil in the necessary case, which I think is only one place, right?  In that one case we can simply do = @[object], instead of appendObject:, so it should be clean.  

These empty arrays aren't heavy, but we do sometimes create large layout trees on every frame during a gesture like in the commerce scroller resizing, and this would create many dozens of extra objects that can be skipped (not saying it is an influential performance problem, moreso a best practice for a core base class).  Suggestion: if you like this pattern, create a singleton empty array and always assign to that pointer.  _maybe_ NSArray array already does that?  We could check if the pointer is always the same?

appleguy:
@rcancro really fantastic refinements in the last round of updates, thank you! I now feel very confident landing this.

The most important follow-up in my opinion is eliminating the dependence on the object version of the collection for layout specifications. The other one that comes to mind is the creation of a large number of empty arrays, although I am genuinely curious to hear if Foundation optimizes these to be the same pointer. If you could quickly review my other comments, I don't remember with certainty if there are any other areas that are still pending, but you conclusively addressed a significant majority!

PR #1700, opened 2016-05-30
[ASPINRemoteImageDownloader] Allows setting authentication challenge …

1nput0utput:
…block

As ASDK creates its own instance of PINRemoteImageManager, It is not possible to just use PINRemoteImageManager sharedInstance to set an authentication challenge block. 

Not sure if this is the best solution, I would love to have any feedback or suggestion.


appleguy:
@1nput0utput great change!  @garrettmoon any input on the approach here?  If you review, I can merge.

garrettmoon:
Would it be better to just expose `sharedPINRemoteImageManager`?

1nput0utput:
@garrettmoon, I guess it would be ok to have access to ```sharedPINRemoteImageManager```, I'll change it now.

PR #1702, opened 2016-05-30
[Example] Fix disappearing of social app example divider

maicki:
UITableViewCell traverses through all its descendant views and adjusts their background color accordingly either to [UIColor clearColor], although potentially it could use the same color as the selection highlight itself. After selection, the same trick is performed again in reverse, putting all the backgrounds back as they used to be. But in our case, we don't want to have the background color disappearing so we reset it after highlighting or selection is done.

### DIFF
--- a/examples/SocialAppLayout/Sample/PostNode.m
+++ b/examples/SocialAppLayout/Sample/PostNode.m
@@ -15,6 +15,8 @@
 #import "LikesNode.h"
 #import "CommentsNode.h"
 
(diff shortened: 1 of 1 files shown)


laonayt:
sure

appleguy:
This seems a bit unfortunate.  Should we provide access to the separator properties on the table cell?  Or is there actually a separatorColor property in addition to the separatorStyle property on the UITableView?  Then maybe we could eliminate the separator node and shorten the code.

appleguy:
cc @maicki

Issue #1701, opened 2016-05-30
Crash when adding a ASEditableTextNode to a horizontal stack layout spec


flovouin:
Hi guys,

Following a discussion on Slack, I'm posting this issue concerning the ASEditableTextNode and stack layouts. The problem is simple: the `ASEditableTextNode` seems to return a size with an infinite width by default, and the horizontal stack layout has troubles handling this case. In fact, it will make the layout crash at runtime.

Since the discussion on Slack I found the solution: setting both `flexGrow` and `flexShrink` to `true`. As the `ASEditableTextNode` returns an infinite width, using `flexShrink` kind of makes sense. However `flexGrow` is also required otherwise the crash will occur.
Also, this behaviour is not very intuitive as it implies that the user should know that the `ASEditableTextNode` returns an infinite width by default, which seems more of an internal matter. (See the vertical stack example in the code linked below, which works as expected without ever specifying a size for the `ASEditableTextNode`.)
Another (dirtier) solution is to put the node inside a `ASStaticLayoutSpec` and set its `preferredFrameSize`, but that implies computing its size manually, which defeats the purpose of the layout engine.

More details about the crash and the solution are available in this example :
https://github.com/flovouin/ASEditableTextNode-Layout

Bottom line: given the behaviour of other nodes in the framework, I would expect the code that currently crashes ASDK to work (i.e. only setting `flexGrow` to `true` should make the `ASEditableTextNode` fill the available space).

Please let me know if you need any additional details :)

Flo

WildDylan:
Hi, unfortunately you can’t position that in a layout spec if you are relying on the size of the element, because it is a UITextView

we can improve this warning / try to improve the default behavior

it should work if you make an ASStaticLayoutSpec and apply a sizeRange = ASSizeRangeMakeWithExactCGSize.  put the editable text node inside the static spec

WildDylan:
ASStackLayoutSpec * spec = [ASStackLayoutSpec horizontalStackLayoutSpec];
    spec.spacing = 15;
    spec.verticalAlignment = ASVerticalAlignmentCenter;
    
    NSMutableArray * childArray = [NSMutableArray array];
    
    ASStaticLayoutSpec * inputLayout = [ASStaticLayoutSpec staticLayoutSpecWithChildren:@[self.inputText]];
    inputText.sizeRange = ASRelativeSizeRangeMakeWithExactCGSize(CGSizeMake(100, constrainedSize.min.width - 80));
    
[...]

PR #1704, opened 2016-05-31
[ASEditableTextNode] Fix measurement of ASEditableTextNode

maicki:
- Add ability to measure the ASEditableTextNode based on the content now by calling `measure:`
- ASEditableTextNode considers `preferredFrameSize` now
- Fix crash if ASEditableTextNode is included in a horizontal stack
- Add tests for ASEditableTextNode

Addresses: #1701

### DIFF
--- a/AsyncDisplayKitTests/ASEditableTextNodeTests.m
+++ b/AsyncDisplayKitTests/ASEditableTextNodeTests.m
@@ -1,0 +1,103 @@
+//
+//  ASEditableTextNodeTests.m
+//  AsyncDisplayKit
+//
+//  Created by Michael Schneider on 5/31/16.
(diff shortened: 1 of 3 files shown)


AsyncDisplayKitTests/ASEditableTextNodeTests.m
```diff
@@ -0,0 +1,103 @@
+//
+//  ASEditableTextNodeTests.m
```
appleguy: Very nice to have this!  Previously this class was one of the most highlighted under-tested areas of the codebase so this will help the code coverage metric a lot.

appleguy:
@maicki nice fix.  The original implementation was actually better than I thought it was, but still this fix is very important for the unconstrained case (a common one).

PR #1706, opened 2016-06-03
[ASTextKitRenderer] Fix crash truncating a string of an ASTextNode with a zero size

maicki:
The crash happens in the placeholderImage of ASTextNode. The node is not visible in the time it try to get the `placeholderImage` and so the `visibleRange` has count 0 and a crash happens while accessing the first object of an empty array.

I improved tests around that area too so look into the `testHandleZeroHeightConstrainedSize` in `ASTextKitTruncationTests` for an example how the crash can be reproduced

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer+TextChecking.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer+TextChecking.mm
@@ -59,7 +59,6 @@
   NSAttributedString *truncationAttributedString = self.attributes.truncationAttributedString;
 
   // get the index of the last character, so we can handle text in the truncation token
-  NSRange visibleRange = self.truncater.visibleRanges[0];
   __block NSRange truncationTokenRange = { NSNotFound, 0 };
 
   [truncationAttributedString enumerateAttribute:ASTextKitTruncationAttributeName inRange:NSMakeRange(0, truncationAttributedString.length)
@@ -75,6 +74,7 @@
     truncationTokenRange = { 0, truncationAttributedString.length };
   }
 
+  NSRange visibleRange = self.truncater.firstVisibleRange;
   truncationTokenRange.location += NSMaxRange(visibleRange);
   
   __block CGFloat minDistance = CGFLOAT_MAX;
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -469,7 +469,7 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
                  forHighlighting:(BOOL)highlighting
 {
   ASTextKitRenderer *renderer = [self _renderer];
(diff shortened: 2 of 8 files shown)


AsyncDisplayKit/TextKit/ASTextKitRenderer+TextChecking.mm
```diff
@@ -75,6 +74,7 @@ - (NSTextCheckingResult *)textCheckingResultAtPoint:(CGPoint)point
     truncationTokenRange = { 0, truncationAttributedString.length };
   }
 
+  NSRange visibleRange = self.truncater.firstVisibleRange;
   truncationTokenRange.location += NSMaxRange(visibleRange);
```
appleguy: What does this actually do with NSNotFound, 0 ?  I like the general intent with using NSNotFound, but I'm thinking we should actually use 0, 0 as the default value because it is more likely to behave as a "no-op" in various calculations.

maicki: I'm open to change, just thinking it makes more sense this way and aligns a bit more with Apple APIs. For example:

[NSRangeFromString](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_Functions/Reference/reference.html#//apple_ref/doc/uid/20000055-BAJBHBDJ): Returns a range from a textual representation.
> ... If aString does not contain any integers, this function returns an NSRange struct whose location and length values are both 0.

[-[NSString rangeOfString:]](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSString_Class/Reference/NSString.html#//apple_ref/occ/instm/NSString/rangeOfString%3a): Here you pass in the string you search for and it returns NSNotFound as location. That's imho pretty similar to `firstVisibleRange`
> ... Returns {NSNotFound, 0} if aString is not found or is empty (@"").

appleguy: @maicki yes definitely, this is why I complement the design choice. I agree that the alternative is less elegant and thoughtful, but the reason if comes to mind is that it may be slightly more practical, in the sense that these values are used internally only and are not a public API and that the distinction between 0, 0 and NSNotFound is meaningless in this case… sept many easy to imagine usages of the value would have to explicitly check for the not found to avoid an invalid calculation that would otherwise be a safe non-operation with zero.

I assume NSMaxRange returns 0 for this case?

appleguy:
Merging, feel free to ping me separately with any thoughts

AsyncDisplayKit/TextKit/ASTextKitRenderer+TextChecking.mm
maicki: It returns `NSNotFound` not 0 ... I agree we use it mostly internally and don't have a proper check in place. I will put up a separate PR that uses 0,0 for `visibleRange`

PR #1707, opened 2016-06-03
[ASDisplayNode] Fix placeholder image is not appearing if size of node changed after initial placement

maicki:
Placeholder are not appearing if a node is measured after the initial placeholder was set. This should fix this behavior by creating the placeholder layer and the placeholder image in the right size while layout happens. 

Furthermore put in a workaround for the Placeholder example that was broken. But would be reproducible by running the Placeholder example and comment out the fix in `ASDisplayNode`.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1088,7 +1088,13 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   // with negative sizes after applying margins, which will cause
   // measureWithSizeRange: on subnodes to assert.
   if (!CGRectEqualToRect(bounds, CGRectZero)) {
+    // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
+    // was created
+    if ([self _shouldHavePlaceholderLayer]) {
+      [self _setupPlaceholderLayerIfNeeded];
+    }
     _placeholderLayer.frame = bounds;
+  
     [self layout];
     [self layoutDidFinish];
   }
--- a/examples_extra/Placeholders/Sample/PostNode.m
+++ b/examples_extra/Placeholders/Sample/PostNode.m
@@ -60,6 +60,10 @@
 - (UIImage *)placeholderImage
 {
   CGSize size = self.calculatedSize;
+  if (CGSizeEqualToSize(size, CGSizeZero)) {
+    return nil;
+  }
+  
   UIGraphicsBeginImageContext(size);
   [[UIColor colorWithWhite:0.9 alpha:1] setFill];
   UIRectFill((CGRect){CGPointZero, size});
@@ -78,6 +82,8 @@
 
 - (void)layout
 {
+  [super layout];
+  
   CGSize textSize = _textNode.calculatedSize;
   CGSize needyChildSize = _needyChildNode.calculatedSize;
 
--- a/examples_extra/Placeholders/Sample/SlowpokeTextNode.m
+++ b/examples_extra/Placeholders/Sample/SlowpokeTextNode.m
@@ -13,9 +13,14 @@
 
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
 
+@interface ASTextNode (ForwardWorkaround)
+// This is a workaround until subclass overwritting of custom drawing class methods is fixed
+- (void)drawRect:(CGRect)bounds withParameters:(id<NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelledBlock isRasterizing:(BOOL)isRasterizing;
+@end
+
 @implementation SlowpokeTextNode
 
-+ (void)drawRect:(CGRect)bounds withParameters:(id<NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelledBlock isRasterizing:(BOOL)isRasterizing
+- (void)drawRect:(CGRect)bounds withParameters:(id<NSObject>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelledBlock isRasterizing:(BOOL)isRasterizing
 {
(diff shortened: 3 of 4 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1080,7 +1080,14 @@ - (void)__layout
     // measureWithSizeRange: on subnodes to assert.
     return;
   }
+
+  // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
```
appleguy: @maicki could you explain just a few words more on why this is necessary? I looked around and didn't see an obvious place at which we destroy the placeholder layer, so I assume the main requirement is to add it to the view hierarchy again, or to update the contents of the layer.

However, updating the contents will be skipped if it already has a contents. Is the contents proactively destroyed when the size of the node changes?

examples_extra/Placeholders/Sample/SlowpokeTextNode.m
```diff
@@ -13,9 +13,14 @@
 
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
 
+@interface ASTextNode (ForwardWorkaround)
+// This is a workaround until subclass overwritting of custom drawing class methods is fixed
```
appleguy: overriding*

examples_extra/Placeholders/Sample/PostNode.m
```diff
@@ -60,6 +60,10 @@ - (instancetype)init
 - (UIImage *)placeholderImage
 {
   CGSize size = self.calculatedSize;
+  if (CGSizeEqualToSize(size, CGSizeZero)) {
+    return nil;
+  }
+  
   UIGraphicsBeginImageContext(size);
   [[UIColor colorWithWhite:0.9 alpha:1] setFill];
   UIRectFill((CGRect){CGPointZero, size});
```
appleguy: Hmm, this is very inefficient.

We should consider bringing back placeholderColor, since no backing store would be created.  But until then, we should use a 1px stretchable image for huge memory / drawing time savings.  Then it technically wouldn't even need to check for zero size.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1088,7 +1088,13 @@ - (void)__layout
   // with negative sizes after applying margins, which will cause
   // measureWithSizeRange: on subnodes to assert.
   if (!CGRectEqualToRect(bounds, CGRectZero)) {
+    // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
+    // was created
```
appleguy: @maicki In a followup can you clarify this comment?  Why is it that the placeholder needs to be re-created after the node size changes?  Basically just describe what operation / action takes place that results in the need to call this method again.

I think the condition would be simplified to put _shouldHave... up in the if() at the top that checks for zero bounds.  Then we can unconditionally call _setupPlaceholder... and do the .frame = set.

examples_extra/Placeholders/Sample/PostNode.m
maicki: The first call is:
`-[ASDisplayView willMoveToWindow]` -->  `-[ASDisplayNode __enterHierarchy]` --> `-[ASDisplayNode _setupPlaceholderLayerIfNeeded]`

We setup the _placeholderLayer in this case but we don't assign a placeholderImage as the size is 0.

After layout is done we call `-[ASDisplayNode _completeLayoutCalculation]` and set the `_placeholderImage` in there. The problem is we don't assign it to the `_placeholderLayer` content:

```
if (_placeholderEnabled && [self _displaysAsynchronously] && self.contents == nil) {

  // Zero-sized nodes do not require a placeholder.
  CGSize layoutSize = (_layout ? _layout.size : CGSizeZero);
  if (CGSizeEqualToSize(layoutSize, CGSizeZero)) {
    return;
  }

  if (!_placeholderImage) {
    _placeholderImage = [self placeholderImage];
  }
}

```
The problem in there is we cannot set it in there as it has to be on the main thread.

The fix assign's a new right sized placeholder image after the layout is done that will be set to the already created _placeholderLayer's content

AsyncDisplayKit/ASDisplayNode.mm
maicki: The first call is:
`-[ASDisplayView willMoveToWindow]` -->  `-[ASDisplayNode __enterHierarchy]` --> `-[ASDisplayNode _setupPlaceholderLayerIfNeeded]`

We setup the _placeholderLayer in this case but we don't assign a placeholderImage as the size is 0.

After layout is done we call `-[ASDisplayNode _completeLayoutCalculation]` and set the `_placeholderImage` in there. The problem is we don't assign it to the `_placeholderLayer` content:

```
if (_placeholderEnabled && [self _displaysAsynchronously] && self.contents == nil) {

  // Zero-sized nodes do not require a placeholder.
  CGSize layoutSize = (_layout ? _layout.size : CGSizeZero);
  if (CGSizeEqualToSize(layoutSize, CGSizeZero)) {
    return;
  }

  if (!_placeholderImage) {
    _placeholderImage = [self placeholderImage];
  }
}

```
The problem in there is we cannot set it in there as it has to be on the main thread.

The fix assign's a new right sized placeholder image after the layout is done that will be set to the already created `_placeholderLayer`'s content

examples_extra/Placeholders/Sample/PostNode.m
maicki: That is just a sample code in the PostNode.

examples_extra/Placeholders/Sample/SlowpokeTextNode.m
maicki: Fixed typo

AsyncDisplayKit/ASDisplayNode.mm
appleguy: @maicki very helpful description, thanks!  Not sure the easiest way to summarize for the comment but I think it's still a bit unclear.  You might just link to this PR with "details on the call behavior are at: " or something.

PR #1710, opened 2016-06-04
[ASVideoNode] Change superclass to ASNetworkImageNode so that it can be its own placeholder

hannahmbanana:
Changes:
- remove _placeholderImageNode property of ASVideoNode (uses self.image now instead)
- move layoutSpecThatFits: code to calculateSizeThatFits: & layout: methods as ASImageNode uses calculateSizeThatFits:

As mentioned in #1677.


hannahmbanana:
Travis build failure: _"property type 'id<ASVideoNodeDelegate> _Nullable' is incompatible with type 'id<ASNetworkImageNodeDelegate> _Nullable' inherited from 'ASNetworkImageNode' [-Werror,-Wincompatible-property-type]"_

It looks like there are only 3 ways to solve this:
- Rename the ASVideoNodeDelegate property to something other than delegate (API breaking)
- Modify the ASVideoNodeDelegate protocol to inherit from the ASNetworkImageNodeDelegate protocol (non-API breaking)
- Modify the ASVideoNodeDelegate property definition to contain both protocols (non-API breaking)

The 2nd and 3rd method require that the one required protocol method in ASNetworkImageNodeDelegate to be made optional (this seems reasonable). 

What's the best way to proceed here?

hannahmbanana:
I think that the 3rd option is probably the best, so I'll implement that.

appleguy:
@Xinchi @Eke @ejensen @maicki @chourobin have a look!

appleguy:
The main blocker here is the delegate situation that @hannahmbanana describes above.

Todo includes - better exposing the fact that users can use .url to set a JPEG thumbnail URL provided by the server rather than waiting much longer for video data to load and generate one, and make sure that -clearFetchedData handling is correct for all three image cases (url, generated from video data, direct set of .image property)

appleguy:
@hannahmbanana Very cool!  This is a great change that adds several inherent capabilities, and showcases the power of the flexibility we have by controlling the relationship between these components.

Issue #1640, opened 2016-05-10
[ASLayout] Implement property similar to Android `gone` property.


bimawa:
Guys i hate Android developers for view.gone property. We must add some like that if we haven't it.
>This view is invisible, and it doesn't take any space for layout purposes.

it means if we have nodes and LayoutSpecs for it, then  gone = false removed nodes from supernode, and invalidate all ASLayoutSpecs to up by hierarchy. If gone = true its add node to superview and add all ASLayoutSpecs described in `layoutSpecThatFits:` method.

hannahmbanana:
@bimawa: As discussed on Slack, does Implicit Node Hierarchy Management suffice? Or is there some other functionality that is desired?

http://asyncdisplaykit.org/docs/implicit-hierarchy-mgmt.html

This is a beta feature developed by @levi that will soon be on by default. When enabled, Implicit Node Hierarchy Management means that your nodes no longer require addSubnode: or removeFromSupernode method calls. The presence or absence of the INHM node and its subnodes are determined in the layoutSpecThatFits: method.

bimawa:
@hannahmbanana @levi yes a little bit another functional i want. What we have today:
```
- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
{
  NSMutableArray *mainNodes = [NSMutableArray new];
  ASStackLayoutSpec *firstPriceStack = [ASStackLayoutSpec stackLayoutSpecWithDirection:ASStackLayoutDirectionHorizontal spacing:0
                                                                        justifyContent:ASStackLayoutJustifyContentSpaceBetween alignItems:ASStackLayoutAlignItemsCenter
                                                                              children:@[self.priceOfOrderTitle, self.priceByRowsTotalTextNode]];
  self.checkoutOrderButton.flexBasis = ASRelativeDimensionMakeWithPoints(48);
  ASInsetLayoutSpec *firstStackInsets = [ASInsetLayoutSpec insetLayoutSpecWithInsets:UIEdgeInsetsMake(16, 0, 0, 0) child:firstPriceStack];
  [mainNodes addObject:firstStackInsets];
  if(self.isVisibleDeliverPriceNodes)
  {
    ASStackLayoutSpec *deliverStack       = [ASStackLayoutSpec stackLayoutSpecWithDirection:ASStackLayoutDirectionHorizontal spacing:0
[...]

Issue #953, opened 2015-12-18
[Question] ASEditableTextNode as a child of layer-backed nodes.


bimawa:
Guys i made custom `ASCellNode` and try add two `ASEditableTextNode's`, after that i catch exception on  `ASDisplayNode (UIViewBridge):215`. Its related with `Does not support layer backing.` or it's a bug? 
p.s. i use LyoutSpec

appleguy:
@bimawa Could you please share some additional information?  I can't quite tell what your issue is yet.

- What is the exact, full exception message that is shown?
- What is the exact line number that it is referencing, and what version / commit of the framework are you using?  I looked at line 215 in my local copy, but it has no assertion or exception there.  Please try using the latest master build, or using 1.9.3 is fine too.
- Note that ASEditableTextNode is one of the only nodes that requires being view-backed.  This means you cannot add it to a supernode that is layer-backed.  ASCellNode is the other node that can't be layer-backed.  You didn't add enough information for me to tell what you've set up, but if you have a subnode of the ASCellNode that is marked as layerBacked = YES, and then are adding the editable node to that, that's your bug.  Just remove the layerBacked = YES for any supernode of the editable text node.

bimawa:
@appleguy

>What is the exact, full exception message that is shown?

`Uncaught exception: CALayer position contains NaN: [nan 19.75]`
Method `- (void)__setSafeFrame:(CGRect)rect` try setup a frame with next data: `rect = (origin = (x = 0, y = 0), size = (width = 0, height = 0))` on line ` _layer.position = position;`
If i use ASLayoutSpec!

>of the framework are you using?

last version 1.9.3 from cocoapods

> Just remove the layerBacked = YES

its a ASCellNode. It does what I need set layerBacked to NO for ASCellNode, ASTableNode too?

piece of my code with error:
bellow class is inherit ASCellNode

```objc
@interface TCTrendsFilterPriceCellNode () <ASEditableTextNodeDelegate>
@property (nonatomic, strong) ASEditableTextNode *fromPriceTextNode;
@property (nonatomic, strong) ASEditableTextNode *toPriceTextNode;
@end

@implementation TCTrendsFilterPriceCellNode
{

}


- (id)init
{
    self = [super init];
    if (self)
    {
        self.fromPriceTextNode                           = [[ASEditableTextNode alloc] init];
        self.fromPriceTextNode.placeholderEnabled        = YES;
[...]

PR #1715, opened 2016-06-06
[ASLayoutSpec] Add new .flexGone property to more easily trigger omitting nodes from a layout.

bimawa:
Ok i merge last master state. And ready for contribute this work correctly.   

-------
facebook/AsyncDisplayKit#1640


levi:
@bimawa this is great, thanks for putting it up as a PR. As a suggestion of the core behavior of flex gone, I'd recommend simply checking for the property being true at ASLayout flatten time. If the test passes, you can simply exclude it from the final ASLayout sublayout hierarchy.

The other main question we should answer is do we want the ASLayoutable with FlexGone enabled to still calculate it's layout spec? Generally seems like we should forego it and rebuild the entire ASLayout tree that was disabled when the property was YES. If that's the case, you can reduce the footprint of flex gone by checking if the property is YES in an ASDisplayNode's calculateSizeThatFits: and returning a zeroed out ASLayout instead of entering measureWithSizeRange: on the layoutables and having to provide a bunch of boilerplate guard statements.

bimawa:
@levi Thank you Levi for review. Yes i think similar, and i worry about `Implicit Hierarchy Management` just if we have `gone = YES` we must remove node from supperNode for better? 

>FlexGone enabled to still calculate it's layout spec? 

Yes i think we not need spend a resources on calculation. 
So. I try find way how i can check  ASLayout flatten time. And modify PR.

bimawa:
pls Travis! Let me home!

bimawa:
@levi i need help :( why Travis so angry?

levi:
@bimawa looks like the unit tests are failing. Try running the tests locally to identify what is not passing.

bimawa:
@levi  is't your code `sublayouts:nil`. I'n just add sublayers from context. And i don't sure so it fully correct. Pls. check.

appleguy:
@bimawa thanks for your work on this!  Next time though, don't curse in commit messages :).  Those last a long time...

bimawa:
@appleguy oh .. sry for this, it's all my weak English.

maicki:
@levi @appleguy What do you guys think of having that behavior based on the `hidden` property of an `ASDisplayNode` instead of introducing a completely new property that is not in the flex box standard?

levi:
@maicki I was thinking that too, but would be unexpected to see a node not in the hierarchy for a seasoned iOS dev. Introducing a separate property here would be the way to go. CSS makes a similar distinction with Display: none; and visibility: hidden;

appleguy:
I think it is a good idea to suggest, Michael. However, this behavior would definitely break some existing apps, and I agree with Levi that there would be at least some developers confused by a difference compared to the equally named Apple framework property. 

I'm not sure that the name used in the current patch is the best one, but we should probably introduce a new name besides hidden.

bimawa:
Guys, when researching a code, i saw `flattened` property with behaviour similar like `hidden`.

maicki:
Yeah after looking into how Auto Layout treats hidden views, Auto Layout still considers the view also if it's set to hidden.

bimawa:
Yea `hidden` property very similar like `alpha` :)

maicki:
`flatten` property has nothing to do with `hidden` and is not for public usage. Furthermore this property will be not public anymore in the near future.

bimawa:
Ouch ) thanks for info. Take it notice.

eanagel:
I know there has been a lot of discussion on this topic, so maybe I'm rehashing discussions that people have already had on slack here, but I thought I'd pass on my thoughts in case they could be useful.

1. As someone who has implemented some non-trivial layout code, I see how something like this could be useful. Our messaging app has many (many) optional elements and at times getting the logic around the different use cases  and combinations can be tricky. Being able to return a layout spec that included all possible nodes with some of them effectively "gone" could be helpful.
2. I really like the declarative nature of layout specs as we currently have them. It seems to me that making this new property updatable outside `layoutSpecThatFits:` would break this model (is that the proposal?) I'd suggest that we keep the declarative nature and not support changing the value outside `layoutSpecThatFits:`. For my use cases, it would work perfectly. (I typically call `setNeedsLayout` whenever a value changes that effects layout anyway.) The downside of this is it might cause extra overhead at times (you might not technically need to re-create the layout tree of you are only toggling on of these values), but there is a simplicity and clarity I really like about it. Plus, the layout happens on background threads which helps minimize the impact.
3. How does this work if a layout has a single child and this property is enabled for the child? What happens to the parent layout? Does the flattening handle this somehow? Seems like you could get into some really weird states with something like an `ASOverlayLayoutSpec` where the child was set to `gone` Would we correctly handle this case? What if the overlay was set to `gone`?
4. `flexGone` seems like a non-obvious name (partially because of the `flex` prefix), how about `flexIgnore` instead? or maybe even be more clear and go for `flexIgnoreLayout`?

bimawa:
@appleguy #953 not related with it.

bimawa:
Guys what u think about "Implicit Hierarchy" my property might be included in it or not?

Adlai-Holler:
After more internal discussion about this solution, we're going to keep evolving the layout system to be more declarative and avoid adding more state to the layout properties of nodes. Thanks @bimawa for proposing this and further thanks to everyone for the clear and thorough discussion around it.

bimawa:
@Adlai-Holler Ok. Thanks!

Adlai-Holler:
@bimawa Yes

PR #1721, opened 2016-06-07
[ASLayoutSpec] Remove use of dictionary to hold children

rcancro:
Converted the backing store of children to a std::vector. Subclass now have defined indexes instead of string keys to add children.


levi:
Looks good, @rcancro. Curious of the use of a vector over a map?

rcancro:
No real reason. @appleguy said that the NSDictionary I was using was showing up in profiling, so I had arrays on the brain. Though I do worry about returning `children` from a map. Even with keys of 0,1,2,... do we know that the children will be returned in their index order?

maicki:
LGTM!

levi:
👍🏽

PR #1722, opened 2016-06-07
[ASLayoutSpec] Remove cached ASTraitCollection

rcancro:
If a layoutSpec is asked for its `asyncTraitCollection`, it creates it on demand.

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayoutSpec.h
+++ b/AsyncDisplayKit/Layout/ASLayoutSpec.h
@@ -11,8 +11,6 @@
 #import <AsyncDisplayKit/ASLayoutable.h>
 #import <AsyncDisplayKit/ASAsciiArtBoxCreator.h>
 
-@class ASTraitCollection;
-
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Layout/ASLayoutSpec.h
```diff
@@ -25,8 +23,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign) BOOL isMutable;
 
-@property (nonatomic, strong, nullable) ASTraitCollection *traitCollection;
```
appleguy: @rcancro  We could certainly support an accessor that constructs and returns the nicer object (and then doesn't hold a reference to it).  This seems not required right now though, just a thought if you think it would be useful.

rcancro: The method `asyncTraitCollection` does exist via the `ASEnvironment` protocol. I guess it is kind of hidden and I could have kept the property as readonly.

appleguy:
@rcancro makes sense, thanks!!

PR #1723, opened 2016-06-08
[ASImageNode] Allow configuration of animation runloop

ejensen:
`NSDefaultRunLoopMode` suspends while a scrollview is tracking. This causes animated images in the scrollview to pause while scrolling. This behavior is not always desired. So this change enables the configuration of the animation runloop mode.

### DIFF
--- a/AsyncDisplayKit/ASImageNode+AnimatedImage.mm
+++ b/AsyncDisplayKit/ASImageNode+AnimatedImage.mm
@@ -85,6 +85,23 @@
   }
 }
 
+- (NSString *)animatedImageRunLoopMode
+{
+  ASDN::MutexLocker l(_displayLinkLock);
+  return _animatedImageRunLoopMode;
+}
+
+- (void)setAnimatedImageRunLoopMode:(NSString *)runLoopMode
+{
+  ASDN::MutexLocker l(_displayLinkLock);
+
+  if (_displayLink != nil) {
+    [_displayLink removeFromRunLoop:[NSRunLoop mainRunLoop] forMode:_animatedImageRunLoopMode];
+    [_displayLink addToRunLoop:[NSRunLoop mainRunLoop] forMode:runLoopMode];
+  }
+  _animatedImageRunLoopMode = runLoopMode;
+}
+
 - (void)animatedImageFileReady
 {
   ASPerformBlockOnMainThread(^{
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASImageNode+AnimatedImage.mm
```diff
@@ -85,6 +85,23 @@ - (void)coverImageCompleted:(UIImage *)coverImage
...
+  return _animatedImageRunLoopMode;
+}
+
+- (void)setAnimatedImageRunLoopMode:(NSString *)runLoopMode
+{
+  ASDN::MutexLocker l(_displayLinkLock);
+
+  if (_displayLink != nil) {
+    [_displayLink removeFromRunLoop:[NSRunLoop mainRunLoop] forMode:_animatedImageRunLoopMode];
+    [_displayLink addToRunLoop:[NSRunLoop mainRunLoop] forMode:runLoopMode];
```
appleguy: @ejensen this is a very smart idea! I would be open to changing the default value here as well, but providing the setter method is a good idea. However, could you add a check for at least a nil string?  Presumably in that case we should set it back to the default. I'm not sure if there is a way to validate that a string is a real mode though.

cc @garrettmoon for overall thoughts, as well as ideas on the default.

koheku:
@ejensen Great idea ! I had the same issue and was planning a similar PR 👍
@appleguy [FLAnimatedImage](https://github.com/Flipboard/FLAnimatedImage/blob/master/FLAnimatedImage/FLAnimatedImageView.m) also lets you set either `NSDefaultRunLoopMode` or `NSRunLoopCommonModes`, here is [their default](https://github.com/Flipboard/FLAnimatedImage/blob/master/FLAnimatedImage/FLAnimatedImageView.m) :
```objective-c
+ (NSString *)defaultRunLoopMode
{
    // Key off `activeProcessorCount` (as opposed to `processorCount`) since the system could shut down cores in certain situations.
    return [NSProcessInfo processInfo].activeProcessorCount > 1 ? NSRunLoopCommonModes : NSDefaultRunLoopMode;
}
```

AsyncDisplayKit/ASImageNode+AnimatedImage.mm
garrettmoon: Looks good to me! I don't have a problem changing the default runloop mode, the call should be pretty cheap (though on file bounds it might be a bit more expensive as it has to open and memory map the next file. Might be worth testing on a lower end device though?

ejensen: @appleguy A nil check was added, and the default runloop mode was moved to a constant. 

As far as changing the default: I personally assume that when people see GIFs stop during scrolling/zooming they would assume it's a bug (videos don't display the same behavior). At least, this is what I observed when I implemented a similar runloop configuration in a different animated gif library back in 2013: [https://github.com/dtorres/OLImageView/pull/8] caused issues: [https://github.com/dtorres/OLImageView/issues/27] [https://github.com/dtorres/OLImageView/issues/42]. That being said, I'm fine with whatever default mode is chosen.

appleguy: @ejensen yes, I tend to agree. Let's change the default to the more feature rich option - spending cycles on the animation during scrolling. Even from a user perspective, it would be understandable if an older or mid-level device began dropping frames during scroll tracking if it were visibly animating images.

PR #1724, opened 2016-06-08
[ASImageNode] Fix animated image playback if file is ready

ejensen:
ASImageNode only starts animating when the animated image's `playbackReadyCallback` is called. If the animate image was ready for playback before it was assigned as ASImageNode's animated image, then `playbackReadyCallback` is never called in the context of ASImageNode. This change checks the playbackReady property, and if it's true, starts playback immediately.

### DIFF
--- a/AsyncDisplayKit/ASImageNode+AnimatedImage.mm
+++ b/AsyncDisplayKit/ASImageNode+AnimatedImage.mm
@@ -39,6 +39,10 @@
       };
     }
     
+    if (animatedImage.playbackReady) {
+      [self animatedImageFileReady];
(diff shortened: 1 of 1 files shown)


appleguy:
cc @garrettmoon.  Woo, great find @ejensen !

garrettmoon:
This actually makes sense given the protocol in ASDK, nice!

Separately the behavior you're describing sounds like a bug in PINAnimatedImage – the playbackReadyCallback is supposed to be called no matter what. Would you mind opening an issue in PINRemoteImage and describing how to reproduce?

PR #1725, opened 2016-06-08
[ASDisplayNode] Add automatic measurement before layout

maicki:
Fix original PR ( #1673 ) that was reverted (#1717)

- Layout should happen with size as large as possible
- Fix wrong logic around layout `_hasDirtyLayout` check

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1124,13 +1124,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   // with negative sizes after applying margins, which will cause
   // measureWithSizeRange: on subnodes to assert.
   if (!CGRectEqualToRect(bounds, CGRectZero)) {
-    // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
-    // was created
-    if ([self _shouldHavePlaceholderLayer]) {
-      [self _setupPlaceholderLayerIfNeeded];
-    }
     _placeholderLayer.frame = bounds;
-  
     [self layout];
     [self layoutDidFinish];
   }
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -53,8 +53,11 @@ NSString * const ASRenderingEngineDidDisplayNodesScheduledBeforeTimestamp = @"AS
 
 @end
 
-//#define LOG(...) NSLog(__VA_ARGS__)
-#define LOG(...)
+#if ASDisplayNodeLoggingEnabled
+  #define LOG(...) NSLog(__VA_ARGS__)
+#else
+  #define LOG(...)
+#endif
 
 // Conditionally time these scopes to our debug ivars (only exist in debug/profile builds)
 #if TIME_DISPLAYNODE_OPS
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1094,26 +1097,54 @@ - (void)__layout
...
-    // measureWithSizeRange: on subnodes to assert.
-    return;
+
+  [self measureNodeWithBoundsIfNecessary:bounds];
+
+  // Performing layout on a zero-bounds view often results in frame calculations
+  // with negative sizes after applying margins, which will cause
+  // measureWithSizeRange: on subnodes to assert.
+  if (!CGRectEqualToRect(bounds, CGRectZero)) {
+    _placeholderLayer.frame = bounds;
```
appleguy: @maicki I think this change may be overwriting a recent fix that added the lines _shouldHavePlaceholderLayer ?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1094,26 +1097,54 @@ - (void)__layout
...
+    [self layout];
+    [self layoutDidFinish];
+  }
+}
+
+- (void)measureNodeWithBoundsIfNecessary:(CGRect)bounds
+{
+  // Normally measure will be called before layout occurs. If this doesn't happen, nothing is going to call it at all.
+  // We simply call measureWithSizeRange: using a size range equal to whatever bounds were provided to that element or
+  // try to measure the node with the largest size as possible
```
appleguy: Is the last part of this sentence true, or should it be removed? "or try to measure the node with the largest size as possible" - I don't see any CGFLOAT_MAX defaults or anything, and don't think we should put in that kind of logic, so maybe just shorten the comment.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1094,26 +1097,54 @@ - (void)__layout
...
+    return;
   }
-  _placeholderLayer.frame = bounds;
+  
+  [self __layoutSublayouts];
+}
 
-  [self layout];
-  [self layoutDidFinish];
+- (void)__layoutSublayouts
```
appleguy: we should probably call this __layoutSubnodes, or __applySublayouts..."layout sublayouts" seems a bit wrong, since it's more the application of the already-calculated layouts.

examples/Videos/Sample/ViewController.h
```diff
@@ -14,8 +14,7 @@
 //  ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
 //  CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
-#import <AsyncDisplayKit/AsyncDisplayKit.h>
-#import <AsyncDisplayKit/ASVideoNode.h>
+#import <UIKit/UIKit.h>
 
 @interface ViewController : UIViewController
```
appleguy: Interesting.  Technically this should be an ASViewController, but of course as-is it makes for a useful test case for this...although a bit of a "secret" purpose / functionality!

Adlai-Holler:
I've got nothing to add to Scott's comments – 👌

AsyncDisplayKit/ASDisplayNode.mm
maicki: Yep I adjusted the code based on the recent changes we did

AsyncDisplayKit/ASDisplayNode.mm
maicki: I removed it as the behavior changed

AsyncDisplayKit/ASDisplayNode.mm
maicki: Agree makes more sense, changed to `__layoutSubnodes`

examples/Videos/Sample/ViewController.h
maicki: Agree, but we need that for testing purposes.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1114,26 +1117,61 @@ - (void)__layout
...
+  
   [self layout];
   [self layoutDidFinish];
 }
 
+- (void)measureNodeWithBoundsIfNecessary:(CGRect)bounds
+{
+  // Normally measure will be called before layout occurs. If this doesn't happen, nothing is going to call it at all.
+  // We simply call measureWithSizeRange: using a size range equal to whatever bounds were provided to that element
+  if (self.supernode == nil && !self.supportsRangeManagedInterfaceState && [self _hasDirtyLayout]) {
```
appleguy: @maicki what about ASViewController?  I think there is a way to override the constrainedSize, but the .node has no supernode, is not range managed, etc.  This may be handled by _hasDirtyLayout, not sure.

appleguy: We probably want to grab the lock here, at least while checking these three properties.  If we don't, the value of one or more could change as the condition is being evaluated...  There is also locking overhead in each method as it is re-acquired.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1114,26 +1117,61 @@ - (void)__layout
...
+      [self measureWithSizeRange:ASSizeRangeMake(bounds.size, bounds.size)];
+    }
+  }
+}
+
+- (void)layout
+{
+  ASDisplayNodeAssertMainThread();
+  
+  if ([self _hasDirtyLayout]) {
```
appleguy: This doesn't read correctly; is it correct?  Shouldn't it be == NO?  If the layout is dirty, then we should probably layout our subnodes.

Maybe this means _isLayoutCalculated, or _isMeasured

AsyncDisplayKit/ASDisplayNode.mm
maicki: @appleguy In ASViewController we pass in the overwritten constrained size in `measureWithSizeRange:` so this will calling through to the measure method and dirty layout is false

Will add acquiring a lock for `supportsRangeManagedInterfaceState` and check for has dirty layout. In `[self supernode]` we already grab a lock

AsyncDisplayKit/ASDisplayNode.mm
maicki: I just replaced this method, before the `_hasDirtyLayout` change it was `!_flags.isMeasured` so the logic did not change. I have to talk with @nguyenhuy about this but this could be the same reason as in here: #1902

PR #1726, opened 2016-06-08
[ASNetworkImageNode] Support loading local animated images

ejensen:
Previously, animated images would only work if the URL is to a remote location. This change adds support to file system URLs. 

I'm not keen on the path extension check. Please let me know if there's a better way to ensure URLs to non-animated images aren't treated as if they are animated.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -457,26 +457,27 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
           } else {
             // First try to load the path directly, for efficiency assuming a developer who
             // doesn't want caching is trying to be as minimal as possible.
-            NSData *data = [NSData dataWithContentsOfURL:_URL];
-            if (data == nil) {
+            UIImage *nonAnimatedImage = [UIImage imageWithContentsOfFile:_URL.path];
+            if (nonAnimatedImage == nil) {
               // If we couldn't find it, execute an -imageNamed:-like search so we can find resources even if the
               // extension is not provided in the path.  This allows the same path to work regardless of shouldCacheImage.
               NSString *filename = [[NSBundle mainBundle] pathForResource:_URL.path.lastPathComponent ofType:nil];
               if (filename != nil) {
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -457,15 +457,27 @@ - (void)_lazilyLoadImageIfNecessary
...
+            // If the file may be an animated gif and then created an animated image.
+            id<ASAnimatedImageProtocol> animatedImage = nil;
+            if (_downloaderImplementsAnimatedImage && data != nil && [_URL.pathExtension isEqualToString:@"gif"]) {
+              animatedImage = [_downloader animatedImageWithData:data];
+            }
+
+            if (animatedImage != nil) {
+              self.animatedImage = animatedImage;
+            } else {
+              self.image = [UIImage imageWithData:data];
```
appleguy: This may not have any caching behavior, but UIImage does funny things.  I'm wondering if this does any form of caching that ...contentsoffile does not do.  @garrettmoon how about that internal UIImage decoded image cache that we had to disable with CG options?  Is that important for animated images?  I think even if it only caches the last frame drawn, it's relevant, because decoded image memory is big.

garrettmoon: Sadly you're absolutely correct. This should likely be changed back to `imageWithContentsOfFile:` even though it means loading the image twice – it's documented to *not* cache the image.

garrettmoon: PINAnimatedImage has a method for detecting whether NSData is a GIF. I'd recommend extending ASAnimatedImageProtocol to add an optional method:

```
- (BOOL)isDataSupported:(NSData *)data
```

Or something with a better name :) ASPINRemoteImageManager could implement this by calling `- (BOOL)pin_isGIF` in NSData+ImageDetectors (from the PINRemoteImage project)

ejensen: Thank you for the suggestions!
Commit 1fb3ffc replaces `imageWithData:` with `imageWithContentsOfFile:`.
Commit 91e6df0 implement the `isDataSupported:` protocol suggested by Garrett.

PR #1727, opened 2016-06-08
[ASVideoNode] Prevent setting old poster image if the asset was set to nil

ejensen:
Use a strong reference to the current asset in order to prevent setting a placeholder image if the asset is set to nil while the placeholder image is generating.

Also corrects the documentation of the `periodicTimeObserverTimescale` default.


appleguy:
@ejensen A beautiful day of patches!  Thanks, this is another tiny masterpiece fix :).

Issue #1703, opened 2016-05-31
[ASTextKitContext, ASTextView] Crash reports using ASDK 1.9.73


klivin:
Posting here as a response to conversation started in Slack channel with @maicki and @appleguy about sharing relevant crashes.

Here's a handful of crashes we are getting that all seem to be somewhat related to the ASTextKitContext in our production app.  They are very hard to reproduce, as I have not yet reproduced any of them.

http://crashes.to/s/57615585859

http://crashes.to/s/78e5fd29225

http://crashes.to/s/bc12fcdb09d

http://crashes.to/s/f52975e6375

http://crashes.to/s/7f34274b048

http://crashes.to/s/d4f445cb196

http://crashes.to/s/521acd4737b

http://crashes.to/s/3c246834890

http://crashes.to/s/ac645e3d425

http://crashes.to/s/adfc74114fa

levi:
Related: #1739

maicki:
@klivin We did a total assessment around locking in `ASTextNode` that got merge today (#1749) and we think the lack of locking in ASTextNode is responsible for the majority of this crashes

I will close this issue for now, please reopen or open a new issue if you run into further crashes. Thanks again for providing all of the crashlogs they definitely helped to figure out problems on our side!

PR #1728, opened 2016-06-08
[WIP] Fix custom class drawing methods are not called for ASTextNode and ASImageNode

maicki:
ASTextNode and ASImageNode currently implement drawing methods as instance methods. This breaks the custom class drawing methods. They will never be called for a subclass. Fixing it by adding a delegate object to the text / image node drawing methods to be able to dispatch the call to the instance method in ASDisplayNode.

To be able to work this way the original parameters needs to be passed to the superclass if a subclass calls the superclasses class drawing method.

cc @appleguy As we talked about this is a first proposal how we could tackle the broken custom class drawing methods. As also said we need to look into the locking issue that is already happening and likely still happening with this solution.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -123,8 +123,11 @@ static struct ASDisplayNodeFlags GetASDisplayNodeFlags(Class c, ASDisplayNode *i
   flags.isInHierarchy = NO;
   flags.displaysAsynchronously = YES;
   flags.shouldAnimateSizeChanges = YES;
-  flags.implementsDrawRect = ([c respondsToSelector:@selector(drawRect:withParameters:isCancelled:isRasterizing:)] ? 1 : 0);
-  flags.implementsImageDisplay = ([c respondsToSelector:@selector(displayWithParameters:isCancelled:)] ? 1 : 0);
+  
+  // Checks if subclasses implements custom drawing behavior
+  flags.implementsDrawRect = [c methodForSelector:@selector(drawRect:withParameters:isCancelled:isRasterizing:)] != [ASDisplayNode methodForSelector:@selector(drawRect:withParameters:isCancelled:isRasterizing:)] ? 1 : 0;
+  flags.implementsImageDisplay = [c methodForSelector:@selector(displayWithParameters:isCancelled:)] != [ASDisplayNode methodForSelector:@selector(displayWithParameters:isCancelled:)] ? 1 : 0;
+  
   if (instance) {
     flags.implementsDrawParameters = ([instance respondsToSelector:@selector(drawParametersForAsyncLayer:)] ? 1 : 0);
     flags.implementsInstanceDrawRect = ([instance respondsToSelector:@selector(drawRect:withParameters:isCancelled:isRasterizing:)] ? 1 : 0);
@@ -1079,7 +1082,14 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     // measureWithSizeRange: on subnodes to assert.
     return;
   }
+  
+  // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
+  // was created
+  if ([self _shouldHavePlaceholderLayer]) {
+      [self _setupPlaceholderLayerIfNeeded];
+  }
   _placeholderLayer.frame = bounds;
+
   [self layout];
   [self layoutDidFinish];
 }
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -24,15 +24,15 @@
 #import "ASInternalHelpers.h"
 #import "ASEqualityHelpers.h"
 
-@interface _ASImageNodeDrawParameters : NSObject
+@interface _ASImageNodeDrawParameters : NSObject<_ASDisplayLayerDelegateForward>
 
-@property (nonatomic, retain) UIImage *image;
+@property (nonatomic, strong) UIImage *image;
 @property (nonatomic, assign) BOOL opaque;
 @property (nonatomic, assign) CGRect bounds;
 @property (nonatomic, assign) CGFloat contentsScale;
 @property (nonatomic, strong) UIColor *backgroundColor;
 @property (nonatomic, assign) UIViewContentMode contentMode;
-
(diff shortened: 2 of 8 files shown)


ocrickard:
Really curious what the cases are that we subclass ASImageNode/ASTextNode?

maicki:
I discovered it as the "Placeholder" example was broken due to it. We use it in there for no specific reason, but in my opinion custom drawing in ASImageNode / ASTextNode is currently broken, doesn't really matter in what case it is used.

What we could do not allow allow custom drawing in subclasses of ASImageNode / ASTextNode at all and throw an exception in this case.

ocrickard:
Yeah, fair. It's been a few years since I really worked in this codebase and I'm just getting back into it, so defer to you guys on what you think is best. But from my quick scan of the code, looks like this was switched from class draw methods to instance methods in https://github.com/maicki/AsyncDisplayKit/commit/01c16809044e34ea663fbd52373e44720fe2eca3

Did they work properly before that? I'm also curious to see what types of perf optimizations occurred as a result of moving to class methods?

maicki:
Yep that's the reason, they where changed to instance methods and that's the reason. To be honest I don't have to context to that change either, but @appleguy should have more information about that.

appleguy:
@ocrickard @maicki I agree that there is essentially no use case for subclassing these objects and then implementing new drawing methods. If that were necessary, it is much more likely that the developer would need greater flexibility by creating their own class from the ground up. They would have to rely on quite a few internal implementation details of Text or Image node to get any real benefit from its logic without using its draw method.

That said, it would be awesome to simply throw an exception in the event that a user provides a draw implementation (class method) in a class that already has an instance based draw method.  This should be relatively easy and would guarantee the developer would not be confused or waste time, even in this tricky edge case.

Regarding the conversion to instance methods, this was one of the top three most important framework performance optimizations for the Pinterest app.  The creation of draw parameters, Which had to occur on the main thread, proved so expensive that it was directly responsible for small frame drops (e.g. 20-50ms of work).  This is particularly true because a complex scrolling interface has a large number of text and image elements, And it is also common for a burst of cells to be added to the display range all at once when new content loads.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1079,7 +1082,14 @@ - (void)__layout
     // measureWithSizeRange: on subnodes to assert.
     return;
   }
+  
+  // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
+  // was created
+  if ([self _shouldHavePlaceholderLayer]) {
```
appleguy: @maicki This change is unrelated to the rest of the PR, right?  I recall us committing something like this recently, but was reverted due to another bug?

Anyway, if it's intended to be here - can we add to the comment a brief description of why _setup is helpful in the size change case?  It's not totally clear to me, because we're about to set the .frame below anyway...so why does it matter if the node size changed / what side effect does _setup have that makes this necessary?  Thanks much!

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -181,7 +183,8 @@ - (NSObject *)drawParametersForAsyncLayer:(_ASDisplayLayer *)layer
                                                     opaque:self.opaque
                                              contentsScale:self.contentsScaleForDisplay
                                            backgroundColor:self.backgroundColor
-                                               contentMode:self.contentMode];
+                                               contentMode:self.contentMode
```
appleguy: Hmm, it looks like the work to port away from using the draw parameters for these core, extremely performance sensitive objects was never quite completed. I believe where we left off was the ability to implement instance methods, And we are using some instance variables in the method, but it seems at least some of the remaining work did not go through.

In addition, we should check the locking especially in the text implementation.

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -83,6 +83,34 @@ - (NSObject *)drawParameters
   return nil;
 }
 
++ (void)drawRect:(CGRect)bounds withParameters:(id <_ASDisplayLayerDelegateForward>)parameters isCancelled:(asdisplaynode_iscancelled_block_t)isCancelledBlock isRasterizing:(BOOL)isRasterizing
+{
+  ASDisplayNodeAssert([parameters conformsToProtocol:@protocol(_ASDisplayLayerDelegateForward)], @"Calling super for drawRect:withParameters:isCancelled:isRasterizing: must pass in the original parameters to superclass implementation.");
+  
+  if ([parameters conformsToProtocol:@protocol(_ASDisplayLayerDelegateForward)]) {
+    id <_ASDisplayLayerDelegateForward> internalParameters = (id <_ASDisplayLayerDelegateForward>)parameters;
+    id<_ASDisplayLayerDelegate> delegate = internalParameters.delegate;
```
appleguy: @maicki this is quite a smart, creative implementation! I had not thought of a reasonable way to support a super call triggering an instance method, coming from a class method!

My only concern with this route is that it requires keeping the draw parameters, Specifically in the instance method case, which is where we would like to eliminate them. I am more than happy to run some up-to-date profiles to quantify the current cost of the parameters now that fewer elements are being put into the dictionaries.  This is definitely a solvable problem by perhaps adding just the delegate to the parameters, or something else like how we handled the layout context.

I would suggest we put this patch aside for now, because there are some pressing issues to prioritize for the 2.0 release. This would be fun to revisit in one of the follow-up releases, as refining the developer APIs for drawing definitely has a few other opportunities that we can investigate.

[sorry for the odd english, relying on voice dictation sucks]

AsyncDisplayKit/ASImageNode.mm
ocrickard: I'll volunteer to do an audit of the locking policies of ASTextNode and friends today. I'll file issues in the existing implementation if I find anything.

ocrickard:
@appleguy fascinating, I'd love to hear about what was slow about building those draw parameters.

AsyncDisplayKit/ASImageNode.mm
maicki: This issue is related to locking issue and crashes seen in `ASTextNode` (and I guess it will not be different for `ASImageNode` as it uses the same mechanism now): https://github.com/facebook/AsyncDisplayKit/issues/1703

maicki:
Furthermore as we improved the performance by going from the class drawing method to an instance drawing method with PR #1777 we prevent custom drawing in ASTextNode or ASImageNode now.

PR #1729, opened 2016-06-08
[Tests] Update FBSnapshotTestCase and generate @3x images

### LINKED ISSUE #1129: [Tests] Update snapshot reference images to support 3x testing
Often my default simulator is the iPhone 6s Plus and running tests fail because the snapshot tests are looking for a 3x image. @nguyenhuy mentioned that it's a pretty straight forward process, albeit I have no idea right now how the snapshot library works.

levi:
Something that I've been needing to do since November.

Fixes #1129


levi:
@nguyenhuy, you'll love this.

AsyncDisplayKitTests/ASTextKitTests.mm
```diff
@@ -81,7 +81,7 @@ static BOOL checkAttributes(const ASTextKitAttributes &attributes, const CGSize
   FBSnapshotTestController *controller = [[FBSnapshotTestController alloc] init];
   UIImage *labelImage = UITextViewImageWithAttributes(attributes, constrainedSize, linkTextAttributes);
   UIImage *textKitImage = ASTextKitImageWithAttributes(attributes, constrainedSize);
-  return [controller compareReferenceImage:labelImage toImage:textKitImage error:nil];
+  return [controller compareReferenceImage:labelImage toImage:textKitImage tolerance:0.0 error:nil];
```
appleguy: @levi why is this needed?  What is default tolerance if not 0?

levi: we're calling a controller method that requires the param. if the tolerance is non-0, it'll track percent diff while scanning pixel values. 0 will make this part of the image diff algorithm a no-op.

PR #1735, opened 2016-06-09
[ASLayout] Clean up flattening process of ASLayout

levi:
My aim here is to simplify the flattening process such that it is an accessible piece of code that new readers can understand. Here's an itemized overview of my process:

1. I cleaned up the flatten BFS — this was original in #1730, but I got carried away cleaning up the entire class. There are now less iterations and it now uses the `flattened` ivar as a visitation marker so that a layout doesn't contain sublayouts that aren't direct subnodes defined in the node's layout spec.
2. Removed the predicate block. It's very flexible having a block-based API, but we currently don't use it effectively to warrant the design. The API has been renamed to `filteredNodeLayoutTree`, to simplify the API footprint and better communicate the resulting structure.
3. `immediateSublayouts` is gone. This was an ill-designed property when it was unclear of the design direction of implicit hierarchy management. Now that the `flattened` bool is used more intelligently, a flattened layout will only contain subnodes referenced in its layout spec.
4. `frame` is now a dynamic property. This will make it cleaner to use in Swift and allow dot-syntax in obj-c.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -62,7 +62,9 @@ NSString * const ASRenderingEngineDidDisplayNodesScheduledBeforeTimestamp = @"AS
 @implementation ASDisplayNode
 
 // these dynamic properties all defined in ASLayoutOptionsPrivate.m
-@dynamic spacingAfter, spacingBefore, flexGrow, flexShrink, flexBasis, alignSelf, ascender, descender, sizeRange, layoutPosition;
+@dynamic spacingAfter, spacingBefore, flexGrow, flexShrink, flexBasis,
+         alignSelf, ascender, descender, sizeRange, layoutPosition, layoutableType;
+
 @synthesize name = _name;
 @synthesize preferredFrameSize = _preferredFrameSize;
 @synthesize isFinalLayoutable = _isFinalLayoutable;
@@ -657,6 +659,11 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (BOOL)_hasDirtyLayout
 {
   return _layout == nil || _layout.isDirty;
+}
+
+- (ASLayoutableType)layoutableType
+{
+  return ASLayoutableTypeDisplayNode;
 }
 
 #pragma mark - Layout Transition
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1916,13 +1916,7 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
       ASLayoutableValidateLayout(layout);
 #endif
     }
-    return [layout flattenedLayoutUsingPredicateBlock:^BOOL(ASLayout *evaluatedLayout) {
-      if (self.usesImplicitHierarchyManagement) {
-        return ASObjectIsEqual(layout, evaluatedLayout) == NO && [evaluatedLayout.layoutableObject isKindOfClass:[ASDisplayNode class]];
-      } else {
-        return [_subnodes containsObject:evaluatedLayout.layoutableObject];
-      }
-    }];
+    return [layout filteredNodeLayoutTree];
   } else {
     // If neither -layoutSpecThatFits: nor -calculateSizeThatFits: is overridden by subclassses, preferredFrameSize should be used,
     // assume that the default implementation of -calculateSizeThatFits: returns it.
@@ -2390,7 +2384,7 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 
 - (void)__layoutSublayouts
 {
-  for (ASLayout *subnodeLayout in _layout.immediateSublayouts) {
+  for (ASLayout *subnodeLayout in _layout.sublayouts) {
     ((ASDisplayNode *)subnodeLayout.layoutableObject).frame = [subnodeLayout frame];
   }
 }
--- a/AsyncDisplayKit/Layout/ASLayout.mm
+++ b/AsyncDisplayKit/Layout/ASLayout.mm
@@ -12,7 +12,6 @@
 
 #import "ASAssert.h"
 #import "ASDimension.h"
-#import "ASDisplayNode.h"
 #import "ASInternalHelpers.h"
 #import "ASLayoutSpecUtilities.h"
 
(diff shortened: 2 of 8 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1912,13 +1912,7 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
                                                size:layout.size
                                          sublayouts:@[layout]];
     }
-    return [layout flattenedLayoutUsingPredicateBlock:^BOOL(ASLayout *evaluatedLayout) {
-      if (self.usesImplicitHierarchyManagement) {
-        return ASObjectIsEqual(layout, evaluatedLayout) == NO && [evaluatedLayout.layoutableObject isKindOfClass:[ASDisplayNode class]];
-      } else {
-        return [_subnodes containsObject:evaluatedLayout.layoutableObject];
```
levi: Turns out you can't do a simple check of whether the node is a subnode to determine if it should be in the flattened layout tree. Why? Because when implicit hierarchy management is turned on, the layout pass is performed before the nodes are even inserted into the hierarchy. The solution here is to do a simple identity check instead and use the `flattened` ivar to determine if a particular subtree has already been visited. Visited subtrees belong to subnodes of the current node, which turns out to be a good way to determine that a node isn't a direct subnode.

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -22,53 +25,76 @@ extern BOOL CGPointIsNull(CGPoint point)
   return isnan(point.x) && isnan(point.y);
 }
 
+@interface ASLayout ()
+
+/**
+ * A boolean describing if the current layout has been flattened.
+ */
+@property (nonatomic, getter=isFlattened) BOOL flattened;
```
levi: Flatten is now private, as it's only needed in the flattening traversal process.

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -103,53 +128,54 @@ + (instancetype)flattenedLayoutWithLayoutableObject:(id<ASLayoutable>)layoutable
...
-      for (ASLayout *sublayout in context.layout.sublayouts) {
-        // Mark layout trees that have already been flattened for future identification of immediate sublayouts
-        BOOL flattened = context.flattened ? : context.layout.flattened;
-        queue.push({sublayout, context.absolutePosition + sublayout.position, NO, flattened});
+    Context context = queue.front();
+    queue.pop();
+
+    if (self != context.layout && [context.layout.layoutableObject isKindOfClass:[ASDisplayNode class]]) {
+      ASLayout *layout = [ASLayout layoutWithLayout:context.layout position:context.absolutePosition];
+      layout.flattened = YES;
```
levi: Flattened simply means that the layout was already visited in a previous flatten pass (the entire flattening process of a node is recursive).

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -103,53 +127,61 @@ + (instancetype)flattenedLayoutWithLayoutableObject:(id<ASLayoutable>)layoutable
...
-      }
-      
-      for (ASLayout *sublayout in context.layout.sublayouts) {
-        // Mark layout trees that have already been flattened for future identification of immediate sublayouts
-        BOOL flattened = context.flattened ? : context.layout.flattened;
-        queue.push({sublayout, context.absolutePosition + sublayout.position, NO, flattened});
+    Context context = queue.front();
+    queue.pop();
+
+    if (self != context.layout && context.layout.type == ASLayoutableTypeDisplayNode) {
```
levi: This is the main reason I added `layoutableType` — namely, to eliminate the need to class identity check.

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -103,53 +127,61 @@ + (instancetype)flattenedLayoutWithLayoutableObject:(id<ASLayoutable>)layoutable
...
+      ASLayout *layout = [ASLayout layoutWithLayout:context.layout position:context.absolutePosition];
+      layout.flattened = YES;
```
levi: Flattened simply means that the layout was already visited in a previous flatten pass (the entire flattening process of a node is recursive).

appleguy:
@levi awesome work here! I had been hoping this code would be cleaned up for a while now, but had not mentioned it to anyone. So it is a great surprise to see such improvements!

I will need help to review this, because the code is very core and that's critical to get right.  @nguyenhuy would be the person most familiar with the existing code, but I believe @maicki is also an expert here.  @Adlai-Holler may have some familiarity with it.

Once there is one very detailed review signoff, I am happy to take a final look and get this merged. Send me a note when that happens!

rcancro:
Admittedly, this is the first time I've ever looked at this code, but it seems to make sense to me! Nice work, Levi! This was one of those scary places I tried not to visit.

AsyncDisplayKit/Layout/ASLayout.mm
maicki: That's a great change as some dev's already asked for that property and actually wanted to use it for some reason

levi: Yeah, it's best we keep it private for now. If we decide to keep around the initial layout tree that contains both layout spec and node layouts, I could see this being useful. However, in the filtered node layout tree result, there really isn't a need to expose this property.

maicki:
LGTM

Adlai-Holler:
Fresh!

Adlai-Holler:
Also in my case, you achieved your goal! I didn't know this mechanism before, and after review it makes perfect sense. Would have been a lot nastier with that predicate-block method and the class check.

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -103,53 +127,61 @@ + (instancetype)flattenedLayoutWithLayoutableObject:(id<ASLayoutable>)layoutable
...
-                                            sublayouts:flattenedSublayouts];
+  return [ASLayout layoutWithLayoutableObject:_layoutableObject
+                         constrainedSizeRange:_constrainedSizeRange
+                                         size:_size
+                                   sublayouts:flattenedSublayouts];
+}
+
+#pragma mark - Accessors
+
+- (ASLayoutableType)type
```
appleguy: Do we need this accessor?  If so, should we call it layoutableType like the others?  @levi definitely like the idea of having this, as checking the class type had become necessary in some areas, including @hannahmbanana's layout inspector - so this helps cleanliness and efficiency.

PR #1737, opened 2016-06-09
[Performance] Prevent GCD thread explosion due to object deallocation workloads (serial deallocation queue).

maicki:
Currently it’s very suboptimal how many threads ASDK is spawning at some point. You can see that especially by running the ASCollectionView and reload the collection view a couple of times or ASTableViewStressTest example by just running it and push a couple of view controller. It spawns over 100 threads pretty fast.

I looked into that a bit into this and it seems like the main drivers are the dispatches to the background queue in ASTextNode, ASNetworkImageNode, ASMultiplexImageNode where we destroy the text renderer, images etc. 

Instead of using a global background queue this PR is adding a dedicated serial queue that we use to destroy the objects mentioned above. This reduces the threads we spawn tremendously.

### DIFF
--- a/AsyncDisplayKit/Private/ASInternalHelpers.h
+++ b/AsyncDisplayKit/Private/ASInternalHelpers.h
@@ -17,8 +17,15 @@ ASDISPLAYNODE_EXTERN_C_BEGIN
 
 BOOL ASSubclassOverridesSelector(Class superclass, Class subclass, SEL selector);
 BOOL ASSubclassOverridesClassSelector(Class superclass, Class subclass, SEL selector);
+
+/// Dispatches the given block to the main queue if not already running on the main thread
 void ASPerformBlockOnMainThread(void (^block)());
-void ASPerformBlockOnBackgroundThread(void (^block)()); // DISPATCH_QUEUE_PRIORITY_DEFAULT 
+
+/// Dispatches the given block to a background queue with priority of DISPATCH_QUEUE_PRIORITY_DEFAULT if not already run on a background queue
+void ASPerformBlockOnBackgroundThread(void (^block)()); // DISPATCH_QUEUE_PRIORITY_DEFAULT
+
+/// Dispatches a block on to a serial queue that's main purpose is for deallocation of objects on a background thread
+void ASPerformBlockOnDeallocationQueue(void (^block)());
 
 CGFloat ASScreenScale();
 
--- a/AsyncDisplayKit/Private/ASInternalHelpers.mm
+++ b/AsyncDisplayKit/Private/ASInternalHelpers.mm
@@ -52,6 +52,17 @@ void ASPerformBlockOnBackgroundThread(void (^block)())
   }
 }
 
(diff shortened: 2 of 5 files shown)


maicki:
@appleguy You think we could land that pretty soon?

maicki:
@Adlai-Holler @rcancro You guys can give me code review please.

AsyncDisplayKit/Private/ASInternalHelpers.mm
```diff
@@ -52,6 +52,17 @@ void ASPerformBlockOnBackgroundThread(void (^block)())
   }
 }
 
+void ASPerformBlockOnDeallocationQueue(void (^block)())
+{
+  static dispatch_queue_t queue;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    queue = dispatch_queue_create("org.AsyncDisplayKit.deallocationQueue", DISPATCH_QUEUE_SERIAL);
+  });
```
ocrickard: lgtm

ocrickard:
This looks great to me.

The one tradeoff is that this could potentially result in slightly higher memory profile sticking around under memory pressure, but I doubt that'll realistically be a big problem.

Adlai-Holler:
👍

maicki:
@ocrickard @Adlai-Holler Thank you guys very much!

@ocrickard Yes I guess the tradeoff is real, but I think we have to take the risk as it's still better most of the cases than spawning > 400 threads while just browsing through an app within minutes.

ocrickard:
lol, agreed. (y)

AsyncDisplayKit/Private/ASInternalHelpers.h
```diff
@@ -17,8 +17,15 @@ ASDISPLAYNODE_EXTERN_C_BEGIN
...
 BOOL ASSubclassOverridesClassSelector(Class superclass, Class subclass, SEL selector);
+
+/// Dispatches the given block to the main queue if not already running on the main thread
 void ASPerformBlockOnMainThread(void (^block)());
-void ASPerformBlockOnBackgroundThread(void (^block)()); // DISPATCH_QUEUE_PRIORITY_DEFAULT 
+
+/// Dispatches the given block to a background queue with priority of DISPATCH_QUEUE_PRIORITY_DEFAULT if not already run on a background queue
+void ASPerformBlockOnBackgroundThread(void (^block)()); // DISPATCH_QUEUE_PRIORITY_DEFAULT
+
+/// Dispatches a block on to a serial queue that's main purpose is for deallocation of objects on a background thread
```
appleguy: To match the other language and fix one grammar error:

"Dispatches the given block to a serial queue dedicated to deallocating objects in the background"
"Serializing these operations ensures that GCD does not spawn too many threads (thread explosion) when large"
"numbers of objects are scheduled for deallocation at the same time.  Each block executes quickly."

appleguy:
@maicki thanks, this is really great!  We should re-sync Pinterest master as well as the current release branch, since it was just cut.

PR #1743, opened 2016-06-12
[Travis CI] Build Optimizations

hannahmbanana:


hannahmbanana:
Note: This is my first bash script, and it's a terribly gross hack using -head and -tail that won't run all examples if we add more than 2 more additional examples. I will work on fixing this but I wanted to put up something fast to get us around the _extra_ long run times we experienced today. 

Findings:
- the elapsed time as reported by Travis is incorrect because it ran 5 of the builds in parallel (~20 min) and then the life-without-cocoa-pods afterwards (~3 min), so it only reports the 3 min elapsed time. Total time was really **~25 minutes: this is a reduction from previous build time**. Perhaps there is a 5 build limit per run?
- no matter what order I ran and reran our builds in, this build always was queued last. Supposedly builds are queued in order across all OSS projects, but I'm wondering if the extra resources required to run this more parallel job somehow de-prioritized it?
- 25 min total run time is better than the 1 hr+ we experienced today, except that the lead time 'til the batch starts is now the long pole. **Most of our builds were queued for 3-4 hours before they started today**. 
- It looks like there is a 164 build limit for OSS projects on Travis, which gets maxed out during normal working hours (9am - 7pm). You can see the status here: https://www.traviscistatus.com/ 
- Several options if this becomes unacceptable: pay $129/mo for 2 concurrent builds spots on Travis (supposedly the best github integration and low maintainence), pay $50/mo for Circle CI (Adlai used this before, reports comparing the two services I found online said iOS builds have issues failing on Circle and are hard to debug), Jenkins (issues with plugins to GitHub, LOTS of maintainence required, could run on our own fast hardware, have experts at Pintrest already?)

appleguy:
This is a massive improvement, thank you!

 It looks like it will be possible to deliver further improvements as well by cashing or other divisions of the example projects, but this win will cut down the total bill would wait time by fully half!

hannahmbanana:
6/21 update: Builds today have been sitting for 10 hrs and not started yet.

Issue #1681, opened 2016-05-19
Improve our batch fetching cases


nickvelloff:
There is a use cases that might make sense to support a batch fetching update/enhancement:

If all loaded items are within visible bounds, fetching never triggers because this is triggered by a scroll event. If I have 10 of 100 items loaded, then pinch to zoom a collection view where all 10 fit nicely the trigger to fetch additional items will never get called.

This would commonly occur on pinch/zoom out with a collection view. 

cc /@maicki

flovouin:
I mentioned something similar a few days ago on the ASDK Slack, so I would be interested in any comment from the team. :)
Batch fetching also isn't triggered when the `contentOffset` of the scroll view is changed programmatically, for example simply by setting the `contentOffset` property, or when calling `scrollRectToVisible`. At the moment only item manipulations (insert, remove, etc) seem to trigger a batch fetch.

flovouin:
Hi guys, sorry for uping this issue once again, but I'd rather do this than create a new one. ;-) My use case is slightly different compared to nickvelloff, but it fits under "batch fetching cases". Please tell me if you'd like me to create a separate issue.

Basically I have an `ASTableNode` (or `ASCollectionView`) with `scrollEnabled = false` on the view. The displayed content location is changed programatically by setting the `contentOffset` of the scroll view, or by calling `scrollRectToVisible`. If batch fetching could also work in these cases, it would really help me out. Right now I would have to reimplement some kind of batch fetching mechanism to obtain the same behaviour, which (partly) defeats the purpose of using the `ASTableNode` in the first place.

I do realise that this would imply some changes to the current batch fetching mechanism, so my request is more of a "would that be doable / reasonable changes, or should I look for an alternative way of achieving this?".

As always, thanks for the awesome framework :-)
Cheers

appleguy:
@flovouin This is a completely valid use case, and we should support it.  Until seeing your note, I simply didn't realize this wasn't working correctly.  Thank you for pinging the issue (it is clear you are constructive and respectful, please don't hesitate to report or ping issues if we are slow to respond!)

The team is quite busy right now preparing for our Tuesday event (WWDC-week gathering for ASDK users to discuss the future of the framework and profile apps).  We do have two new full time framework developers starting at Pinterest this week.

[...]

PR #1744, opened 2016-06-13
[Batch Fetching] Check for batch fetching in scrollViewDidScroll

flovouin:
Following #1681 I'm opening this very simple pull request. The idea is that batch fetching isn't triggered when the `contentOffset` of the `ASCollectionView` / `ASTableView` is changed programmatically. This is the case for example when setting the `contentOffset` directly, or when calling `scrollRectToVisible` on the underlying `UIScrollView`.
The following repository shows a simplified version of my use case: https://github.com/flovouin/ASDK-BatchFetchingTests
`scrollEnabled` is set to `false` on the `ASCollectionView`, and its scrolling behaviour is controlled by another `UIScrollView` (this is loosely inspired by WWDC 2012 - Session 223). Obviously this is useless in the above example, but very handy in my real-life application. One could think of other use cases, like scrolling (programmatically) to a specific item in the collection.

The most generic solution I found is simply to check if a batch fetch should be performed when the `contentOffset` is changed, in `scrollViewDidScroll`. Although this method might be called quite often, `_checkForBatchFetching` should introduce very little computations as it returns directly during dragging. This also works for the `scrollRectToVisible`, as obviously the `contentOffset` is changed.

Additional thoughts:
- Although this solves my very simple problem, it does not address nickvelloff's original issue #1681. Unfortunately I've never worked on zooming / scaling a collection view and I'm not sure what would be the best way to solve this problem. Also, I saw there's a somewhat related PR #1697.
- A (very small) limitation is that the check is performed at the current `contentOffset`, contrary to the check in `scrollViewWillEndDragging`, that takes into account the target offset of the animation. When simply setting the `contentOffset`, this is the best we can do, but for an animated `scrollRectToVisible`, it would technically be possible to know the target offset (passed to this method), and directly check if a batch fetch is needed at this location. Instead, the proposed solution will trigger a check at every step of the animation (each time the `contentOffset` is updated).

Please tell me if this is an acceptable change and if I can improve it in any way. :-)
Cheers,
Flo

Edit: Now that I think about it, maybe the same change could be applied to zooming by simply checking for batch fetch in `scrollViewDidZoom`? I don't have a toy example to check though.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -659,6 +659,9 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
                                       inScrollView:scrollView
                                      withCellFrame:collectionCell.frame];
   }
+
+  [self _checkForBatchFetching];
+
   if (_asyncDelegateFlags.asyncDelegateScrollViewDidScroll) {
     [_asyncDelegate scrollViewDidScroll:scrollView];
   }
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -681,6 +681,9 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
                                  inScrollView:scrollView
                                 withCellFrame:tableCell.frame];
   }
+
+  [self _checkForBatchFetching];
+
   if (_asyncDelegateFlags.asyncDelegateScrollViewDidScroll) {
     [_asyncDelegate scrollViewDidScroll:scrollView];
   }


facebook-github-bot-9:

facebook-github-bot-6:

hannahmbanana:
@maicki - could you review? this is a short one and you're an expert in this area. Just put r=<username> somewhere in the comment if you approve.

flovouin:
Happy to report that this would also be solved by the aforementioned #1697. :-)

nickvelloff:
@flovouin this is likely resolved in master https://github.com/facebook/AsyncDisplayKit/pull/1697

flovouin:
Indeed it is! Thanks, closing this one.

PR #1746, opened 2016-06-14
Fix default tuning parameter for range mode ASLayoutRangeModeMinimum and range type ASLayoutRangeTypeFetchData

maicki:


Adlai-Holler:
👌

PR #1748, opened 2016-06-14
[ASVideoNode] Re-enabling HLS video constructed from URL

Xinchi:
Apple claims in the AVFoundationProgramming Guide that HLS videos can be constructed only through URL, but later with iOS 4.3 release notes it claimed to bring updates to how the HLS videos should be initialized, which works with asset too. I’ve tested with both, and it looks like initializing with asset is buggy.


hannahmbanana:
@Xinchi: could you modify this PR to include [ASVideoNode] in front of the title? Thanks!

maicki:
LGTM

Eke:
Tested in my app, LGTM

gazreese:
@Xinchi - Please check over this first - https://github.com/facebook/AsyncDisplayKit/issues/1600

There were issues with the previous method, as outlined above.

One way to avoid the issues above may be to make the construction of the ASVideoNode class more explicit like so:

```
// Mirror the construction of AVPlayerItem with the URL or AVAsset
- (instancetype)initWithURL:(NSURL*)url;
- (instancetype)initWithAsset:(AVAsset*)asset;
```

I'm also finding that the current initialisation method is buggy. I find a lot of HLS videos that won't play. There are some examples here if you're interested:

http://stackoverflow.com/questions/10104301/hls-streaming-video-url-need-for-testing

Xinchi:
Hi @gazreese ! Thanks for chiming in your idea!

What are the "issues with the previous method"? and is there a solution?

gazreese:
The main problem is that the user would construct the ASVideoNode with an AVAsset, but the asset that they want to use wasn't actually used. 

The user would see the same AVAsset in the .asset property, but actually this asset has been thrown away. The user can set observers on the AVAsset if they wish, but they would be pointing to the wrong asset.

Also, I think the AVAsset is responsible for network activity. If the user constructs with an asset but it's not used:
`
playerItem = [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];`

I assume another AVAsset is created inside the AVPlayerItem on our behalf. Does this mean that we now have two AVAsset objects performing network activity for the same network data?

If the user constructed with a URL in the first place they wouldn't expect the .asset property to be set.

`- (instancetype)initWithURL:(NSURL*)url;`

We could even just map the ASVideoNode .asset property directly to the one inside AVPlayerItem, and make it read only. Then we could be certain that the property refers to the correct player item.

Please let me know if you have any questions.

ejensen:
`_asset.tracks.count == 0` potentially blocks the main thread until a network request completes. Original details can be found in https://github.com/facebook/AsyncDisplayKit/pull/1599. `@"tracks"` may need to be added to the array of keys passed into `loadValuesAsynchronouslyForKeys:`

Xinchi:
Closing this pull request due to multiple issues with this.

Xinchi:
@ejensen @gazreese Thanks for chiming in your ideas! I am closing this patch for now because the issue seems to have gone away in the Pinterest use case.  Will re-visit if we are seeing this issue again.

gazreese:
Sounds good :-)

Xinchi:
Reopening this issue as the bug has been observed again.   @gazreese I totally agree with you on the duplicated .asset property.  I will put up a PR that maps the ASVideoNode.asset to the internal asset object.

gazreese:
@Xinchi Sounds like a good idea to me. I'm using this functionality in my current project and don't mind testing for you if you like, just let me know when you've done the PR.

Xinchi:
Closing this PR, creating a new one : #1954

Issue #1739, opened 2016-06-10
[ASTextNode] Thread safety issues


ocrickard:
Did a quick audit for unprotected ivar access on the mutable ASTextNode instance:

https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L159 (description, if only used in debugging probably doesn't matter, but may as well lock).

https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L288 accesses _renderer outside of the mutex. You must use the property accessor to have safe atomic access. This can crash due to another thread changing _renderer in the middle of your access.

Same https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L294, https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L309, https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L347

Probably obvious, but https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L368 isn't threadsafe.

https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASTextNode.mm#L403

I'm not going to call out all the other properties that aren't threadsafe. I'm assuming that the threading model by the owning callsites/frameworks makes them OK?

In the renderer:

All of these lazy attributes are definitely not safe:

https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm#L58

https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm#L70

https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm#L95

If you're going to move these things to be lazily generated, don't we also have to stop directly using the ivars? https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm#L181

appleguy:
@ocrickard thanks for auditing this!  Yes, we recently discovered that the many recent changes to ASTextNode have broken the threading model.  This is urgent to fix.

[...]

PR #1749, opened 2016-06-15
[ASTextNode] Refactor locking to account for recent feature additions and on-demand initialization.

maicki:
- Move lock from ASDN::Mutex to std::mutex in ASTextKitContext
- Small cleanup in ASTextNode
- For now we use a big recursive lock. This needs to be revisited as we revisit the whole ASDK locking strategy.

Related: #1739, #1703

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -236,6 +236,8 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 
 - (ASTextKitAttributes)_rendererAttributes
 {
+  std::lock_guard<std::recursive_mutex> l(_textLock);
+  
   return {
     .attributedString = _attributedText,
     .truncationAttributedString = _composedTruncationText,
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -68,6 +68,7 @@ static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncation
 
   NSString *_highlightedLinkAttributeName;
   id _highlightedLinkAttributeValue;
+  ASTextNodeHighlightStyle _highlightStyle;
   NSRange _highlightRange;
   ASHighlightOverlayLayer *_activeHighlightLayer;
 
@@ -195,7 +196,7 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   
   // If we are view-backed and the delegate cares, support the long-press callback.
   SEL longPressCallback = @selector(textNode:longPressedLinkAttribute:value:atPoint:textRange:);
-  if (!self.isLayerBacked && [self.delegate respondsToSelector:longPressCallback]) {
+  if (!self.isLayerBacked && [_delegate respondsToSelector:longPressCallback]) {
     _longPressGestureRecognizer = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(_handleLongPress:)];
     _longPressGestureRecognizer.cancelsTouchesInView = self.longPressCancelsTouches;
     _longPressGestureRecognizer.delegate = self;
@@ -258,7 +259,7 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 - (void)_invalidateRendererIfNeededForBoundsSize:(CGSize)boundsSize
 {
   if ([self _needInvalidateRendererForBoundsSize:boundsSize]) {
-    // Our bounds of frame have changed to a size that is not identical to our constraining size,
+    // Our bounds have changed to a size that is not identical to our constraining size,
     // so our previous layout information is invalid, and TextKit may draw at the
     // incorrect origin.
     {
@@ -489,6 +490,8 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 {
   ASDisplayNodeAssertMainThread();
   
+  std::lock_guard<std::recursive_mutex> l(_textLock);
+  
   ASTextKitRenderer *renderer = [self _renderer];
   NSRange visibleRange = renderer.firstVisibleRange;
   NSAttributedString *attributedString = _attributedText;
@@ -590,8 +593,8 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
     }
 
     // Ask our delegate if a long-press on an attribute is relevant
-    if ([self.delegate respondsToSelector:@selector(textNode:shouldLongPressLinkAttribute:value:atPoint:)]) {
-      return [self.delegate textNode:self
+    if ([_delegate respondsToSelector:@selector(textNode:shouldLongPressLinkAttribute:value:atPoint:)]) {
+      return [_delegate textNode:self
         shouldLongPressLinkAttribute:_highlightedLinkAttributeName
                                value:_highlightedLinkAttributeValue
                              atPoint:[gestureRecognizer locationInView:self.view]];
@@ -612,6 +615,20 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 
 #pragma mark - Highlighting
 
+- (ASTextNodeHighlightStyle)highlightStyle
+{
+  std::lock_guard<std::recursive_mutex> l(_textLock);
+  
+  return _highlightStyle;
+}
+
+- (void)setHighlightStyle:(ASTextNodeHighlightStyle)highlightStyle
+{
+  std::lock_guard<std::recursive_mutex> l(_textLock);
+  
+  _highlightStyle = highlightStyle;
+}
+
 - (NSRange)highlightRange
 {
   ASDisplayNodeAssertMainThread();
@@ -626,8 +643,6 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 
 - (void)setHighlightRange:(NSRange)highlightRange animated:(BOOL)animated
 {
-  ASDisplayNodeAssertMainThread();
-  
   [self _setHighlightRange:highlightRange forAttributeName:nil value:nil animated:animated];
 }
 
@@ -885,6 +900,9 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
 - (void)touchesBegan:(NSSet *)touches withEvent:(UIEvent *)event
 {
   ASDisplayNodeAssertMainThread();
+  
+  std::lock_guard<std::recursive_mutex> l(_textLock);
+  
   [super touchesBegan:touches withEvent:event];
 
   CGPoint point = [[touches anyObject] locationInView:self.view];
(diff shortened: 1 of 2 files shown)


maicki:
@appleguy @Adlai-Holler Can I have a first code review please

Adlai-Holler:
Sure man, I'll review it first thing tomorrow

appleguy:
@garrettmoon @rcancro @ocrickard @shannonma in case you guys have an opportunity to review in the next few days...this is a top priority for me, but I have no chance to review until Saturday morning, which is much too slow IMO for this diff!

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -76,7 +71,7 @@ @implementation ASTextNode {
   NSRange _highlightRange;
   ASHighlightOverlayLayer *_activeHighlightLayer;
 
-  ASDN::Mutex _rendererLock;
+  std::recursive_mutex _textLock;
```
ocrickard: Why recursive?

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -227,7 +224,8 @@ - (ASTextKitRenderer *)_renderer
 
 - (ASTextKitRenderer *)_rendererWithBounds:(CGRect)bounds
 {
-  ASDN::MutexLocker l(_rendererLock);
+  std::lock_guard<std::recursive_mutex> l(_textLock);
```
ocrickard: Looks fine to me.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -250,9 +248,28 @@ - (ASTextKitAttributes)_rendererAttributes
...
+}
+
+- (void)_invalidateRendererIfNeededForBoundsSize:(CGSize)boundsSize
+{
+  if ([self _needInvalidateRendererForBoundsSize:boundsSize]) {
+    // Our bounds of frame have changed to a size that is not identical to our constraining size,
+    // so our previous layout information is invalid, and TextKit may draw at the
+    // incorrect origin.
+    {
+      std::lock_guard<std::recursive_mutex> l(_textLock);
```
ocrickard: √ Only holding the lock within the context of the brackets here.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -250,9 +248,28 @@ - (ASTextKitAttributes)_rendererAttributes
   };
```
ocrickard: Do we need locking on this method?

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -333,6 +338,8 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
   ASDisplayNodeAssert(constrainedSize.width >= 0, @"Constrained width for text (%f) is too  narrow", constrainedSize.width);
   ASDisplayNodeAssert(constrainedSize.height >= 0, @"Constrained height for text (%f) is too short", constrainedSize.height);
   
+  std::lock_guard<std::recursive_mutex> l(_textLock);
```
ocrickard: Ahh, this is probably why it has to be recursive, we hold it while calling into methods that also lock.

If we can achieve this without a recursive mutex I'll be happier, since you're interacting with the global textkit locks inside the renderer, and as written I *think* there's a potential for lock inversion.

AsyncDisplayKit/ASTextNode.mm
maicki: We only call the `_rendererAttributes` method within the `_rendererWithBounds:` method where we already hold the lock that's the reason I didn't add locking to it for now. Would you recommend to add locking to `_rendererAttributes` too?

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -749,32 +785,28 @@ - (NSArray *)_rectsForTextRange:(NSRange)textRange measureOption:(ASTextKitRende
...
+  
   CGRect rect = [[self _renderer] trailingRect];
-  return [self.class _adjustRendererRect:rect forShadowPadding:self.shadowPadding];
+  return ASTextNodeAdjustRenderRectForShadowPadding(rect, self.shadowPadding);
 }
 
 - (CGRect)frameForTextRange:(NSRange)textRange
 {
+  std::lock_guard<std::recursive_mutex> l(_textLock);
```
ocrickard: I don't think this needs a lock. self _renderer is locked so you're guaranteed an atomic access to the renderer, then the subsequent operations should be OK without locking since ASTextNodeRenderer should be threadsafe.

ocrickard: That is, it should be threadsafe if we fixed it :P

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -749,32 +785,28 @@ - (NSArray *)_rectsForTextRange:(NSRange)textRange measureOption:(ASTextKitRende
...
-}
-
-- (NSArray *)highlightRectsForTextRange:(NSRange)textRange
-{
-  return [self _rectsForTextRange:textRange measureOption:ASTextKitRendererMeasureOptionBlock];
-}
-
 - (CGRect)trailingRect
 {
+  std::lock_guard<std::recursive_mutex> l(_textLock);
```
ocrickard: Same here, probably doesn't need to lock?

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -790,6 +822,8 @@ - (UIImage *)placeholderImage
     return nil;
   }
   
+  std::lock_guard<std::recursive_mutex> l(_textLock);
```
ocrickard: Yeah, holding this lock at this scope while we call below into other methods that also take the same lock makes me nervous.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -470,6 +485,8 @@ - (id)_linkAttributeValueAtPoint:(CGPoint)point
    inAdditionalTruncationMessage:(out BOOL *)inAdditionalTruncationMessageOut
                  forHighlighting:(BOOL)highlighting
 {
+  ASDisplayNodeAssertMainThread();
```
ocrickard: Worth noting that just because something is on the main thread doesn't mean you don't have to lock on ivar access. A background thread could hypothetically be altering it while we're using the ivars in this method and we'll corrupt the heap.

ocrickard:
Thanks for doing this!

This is an improvement on the previous implementation, I do feel like it will be prone to deadlocks through lock inversion in the future though. If we can split the existing methods into a few classes:

1. Public API, callable from any thread. Must lock before calling mutative private helpers or directly accessing ivars. Should comment above each of these methods/move them to a single part of the file that specifies their contract.
2. Private dependent methods. These are higher-level helpers that are responsible for their own locking procedures, [self _renderer] for example. Locks should only be held while mutating/reading ivars, ideally not while ever calling into the renderer.
3. Private mutative methods. These are low-level helpers that should not lock, and should *assume* that a lock is held while calling. They should be marked as such possibly by appending "_locked" to their method names.
4. Private pure methods. These are methods that should require no locking because they are pure. They should be moved to static C functions that don't have access to ivars so they can't become unsafe in the future.

If we can do a little bit more organization I feel like the need for a recursive mutex could be lessened? What do you think?

AsyncDisplayKit/ASTextNode.mm
Adlai-Holler: It might be worth adding a comment indicating that the method is assumed to have the lock held.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -250,9 +248,28 @@ - (ASTextKitAttributes)_rendererAttributes
...
 
+- (void)_invalidateRendererIfNeeded
+{
+  [self _invalidateRendererIfNeededForBoundsSize:self.threadSafeBounds.size];
+}
+
+- (void)_invalidateRendererIfNeededForBoundsSize:(CGSize)boundsSize
+{
+  if ([self _needInvalidateRendererForBoundsSize:boundsSize]) {
+    // Our bounds of frame have changed to a size that is not identical to our constraining size,
```
Adlai-Holler: Nit: typo "bounds of frame"

AsyncDisplayKit/ASTextNode.mm
maicki: We need the locks  as `ASTextNodeRenderer` is not thread safe at the moment. This was a conscious decision to not make `ASTextNodeRenderer` and `ASTextKitContext` (except the access to the text kit components) thread safe for now and provide the locking in the `ASTextNode` for both components.

AsyncDisplayKit/ASTextNode.mm
maicki: We decided to start with a recursive mutex for now. As I wrote in the commit message we need to / will rethink the whole locking strategy starting from `ASDisplayNode` pretty soon and with that I would like to remove our recursive mutexes. Unfortunately this is not possible at the moment in terms of time :/

AsyncDisplayKit/ASTextNode.mm
ocrickard: It's trivial to make ASTextNodeRenderer threadsafe, right?. It has far less surface area than this class.

ocrickard: I could be missing something though, maybe it's harder than I thought.

AsyncDisplayKit/ASTextNode.mm
maicki: @ocrickard What do you think in terms of adding a lock in here or adding a comment?

AsyncDisplayKit/ASTextNode.mm
ocrickard: Got it, sounds good. So far as I could tell your locking with the recursive mutex is correct in this diff, but of course... there be dragons in multithreaded locking with recursive mutexes, so there may be edge cases. I guess I'm a little worried about rushing the threading model of one of the core classes of ASDK, but I guess if this is the interim fix I'm cool with that.

We probably need another few thorough reviews on this to verify the locking order is correct and we don't introduce hard-to-catch deadlocks.

maicki: Yep fully agree!

AsyncDisplayKit/ASTextNode.mm
garrettmoon: Yeah, my opinion is to add a comment. Either that the lock is already held by the calling method but we're locking again for completeness and readability *or* don't call the lock and note above the method that it must be called locked. Would it make sense to adopt some type of method naming scheme for methods that must be called with the lock held?

maicki: I added a lock for it for consistency of this PR now, but I agree and as @ocrickard said in his comment in the future we should introduce a naming for methods that must be called with the lock held?

AsyncDisplayKit/ASTextNode.mm
garrettmoon: Yup, I think accessing the renderer's firstVisibleRange and grabbing _attributedText should be done in the lock? Separately, does it even make sense to have the _renderer method?

maicki: Yep accessing the `_renderer` without holding a lock does not makes sense. I will let it in for now until we have a better naming scheme and kind of similar infrastructure as @ocrickard recommended, that btw I really like.

maicki:
@Adlai-Holler @ocrickard @garrettmoon Could you guys give me another look over the code. I tried to address all of your comments as well as I found some more things.

Changes:
- Add lock to _rendererAttributes
- Add locking to _linkAttributeValueAtPoint:attributeName:range:inAdditionalTruncationMessage:forHighlighting: as we access the attributed text in there
- Add locking to touchesBegan:withEvent: as we are accessing the [ASTextKitRenderer firstVisibleRange]
- Add locking for highlightStyle
- Move accessing delegate property access to instance variable access

PR #1755, opened 2016-06-16
[ASCollectionViewLayoutInspecting] Refactor usage of ASCollectionViewLayoutInspecting

maicki:
- Fix not using itemSize of UICollectionViewFlowLayout
- Move automatic constrained size calculation to the ASCollectionViewFlowLayoutInspector
- Provide a null layout inspector for throwing exceptions if a custom
  UICollectionView is given but no ASCollectionViewLayoutInspecting
- Fix not checking for optional layout inspecting data source methods
  are implemented or not
- Improving tests around ASCollectionViewLayoutInspecting

cc @levi that's the pull request we talked the other day about. Let me know what you think

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -253,9 +253,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   _layoutFacilitator = layoutFacilitator;
   
-  // Trigger creating the layout inspector
-  [self layoutInspector];
-  
   _proxyDelegate = [[ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
   super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
   
@@ -371,7 +368,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   super.dataSource = (id<UICollectionViewDataSource>)_proxyDataSource;
   
   if (_layoutInspectorFlags.layoutInspectorDidChangeCollectionViewDataSource) {
-    [_layoutInspector didChangeCollectionViewDataSource:asyncDataSource];
+    [self.layoutInspector didChangeCollectionViewDataSource:asyncDataSource];
   }
 }
 
@@ -406,7 +403,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
   
   if (_layoutInspectorFlags.layoutInspectorDidChangeCollectionViewDelegate) {
-    [_layoutInspector didChangeCollectionViewDelegate:asyncDelegate];
+    [self.layoutInspector didChangeCollectionViewDelegate:asyncDelegate];
   }
 }
 
@@ -977,7 +974,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 
 - (ASSizeRange)dataController:(ASDataController *)dataController constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
 {
-  return [_layoutInspector collectionView:self constrainedSizeForNodeAtIndexPath:indexPath];
+  return [self.layoutInspector collectionView:self constrainedSizeForNodeAtIndexPath:indexPath];
 }
 
 - (NSUInteger)dataController:(ASDataController *)dataController rowsInSection:(NSUInteger)section
@@ -1017,17 +1014,17 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 
 - (ASSizeRange)dataController:(ASCollectionDataController *)dataController constrainedSizeForSupplementaryNodeOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath
 {
-  return [_layoutInspector collectionView:self constrainedSizeForSupplementaryNodeOfKind:kind atIndexPath:indexPath];
+  return [self.layoutInspector collectionView:self constrainedSizeForSupplementaryNodeOfKind:kind atIndexPath:indexPath];
 }
 
 - (NSUInteger)dataController:(ASCollectionDataController *)dataController supplementaryNodesOfKind:(NSString *)kind inSection:(NSUInteger)section
 {
-  return [_layoutInspector collectionView:self supplementaryNodesOfKind:kind inSection:section];
+  return [self.layoutInspector collectionView:self supplementaryNodesOfKind:kind inSection:section];
 }
 
 - (NSUInteger)dataController:(ASCollectionDataController *)dataController numberOfSectionsForSupplementaryNodeOfKind:(NSString *)kind;
 {
-  return [_layoutInspector collectionView:self numberOfSectionsForSupplementaryNodeOfKind:kind];
+  return [self.layoutInspector collectionView:self numberOfSectionsForSupplementaryNodeOfKind:kind];
 }
 
 #pragma mark - ASRangeControllerDataSource
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -95,6 +95,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   ASRangeController *_rangeController;
   ASCollectionViewLayoutController *_layoutController;
   id<ASCollectionViewLayoutInspecting> _defaultLayoutInspector;
+  id<ASCollectionViewLayoutInspecting> _layoutInspector;
   NSMutableSet *_cellsForVisibilityUpdates;
   id<ASCollectionViewLayoutFacilitatorProtocol> _layoutFacilitator;
   
@@ -156,6 +157,11 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     unsigned int asyncDataSourceNumberOfSectionsInCollectionView:1;
     unsigned int asyncDataSourceCollectionViewConstrainedSizeForNodeAtIndexPath:1;
   } _asyncDataSourceFlags;
+  
+  struct {
+    unsigned int layoutInspectorDidChangeCollectionViewDataSource:1;
+    unsigned int layoutInspectorDidChangeCollectionViewDelegate:1;
+  } _layoutInspectorFlags;
 }
 
 // Used only when ASCollectionView is created directly rather than through ASCollectionNode.
@@ -245,20 +251,10 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   // and should not trigger a relayout.
   _ignoreMaxSizeChange = CGSizeEqualToSize(_maxSizeForNodesConstrainedSize, CGSizeZero);
   
-  if ([layout asdk_isFlowLayout]) {
-    // Register the default layout inspector delegate for flow layouts only
-    UICollectionViewFlowLayout *layout = (UICollectionViewFlowLayout *)self.collectionViewLayout;
-    ASDisplayNodeAssertNotNil(layout, @"Collection view layout must be a flow layout to use the built-in inspector");
-    _defaultLayoutInspector = [[ASCollectionViewFlowLayoutInspector alloc] initWithCollectionView:self flowLayout:layout];
-  } else {
-    // Custom layouts will need to roll their own ASCollectionViewLayoutInspecting implementation and set a layout
-    // delegate. In the meantime ASDK provides a null layout inspector that does not provide any implementation
-    // and throws an exception for methods that should be implemented in the <ASCollectionViewLayoutInspecting>
-    _defaultLayoutInspector = [[ASCollectionViewDefaultCustomLayoutInspector alloc] initWithCollectionView:self];
-  }
-  _layoutInspector = _defaultLayoutInspector;
-  
   _layoutFacilitator = layoutFacilitator;
+  
+  // Trigger creating the layout inspector
+  [self layoutInspector];
   
   _proxyDelegate = [[ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
   super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
@@ -373,8 +369,8 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   }
   
   super.dataSource = (id<UICollectionViewDataSource>)_proxyDataSource;
-    
-  if ([_layoutInspector respondsToSelector:@selector(didChangeCollectionViewDataSource:)]) {
+  
+  if (_layoutInspectorFlags.layoutInspectorDidChangeCollectionViewDataSource) {
     [_layoutInspector didChangeCollectionViewDataSource:asyncDataSource];
   }
 }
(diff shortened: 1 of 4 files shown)


maicki:
Possible improvement / idea: Instead of providing a `ASCollectionViewNullLayoutInspector` we provide a default one that would return a default calculated constrained size. The question is just what would we do with the supplemented view callbacks?

cc @levi

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -248,11 +247,19 @@ - (instancetype)_initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionV
...
-    _layoutInspector = [self flowLayoutInspector];
+    // Register the default layout inspector delegate for flow layouts only
+    UICollectionViewFlowLayout *layout = (UICollectionViewFlowLayout *)self.collectionViewLayout;
+    ASDisplayNodeAssertNotNil(layout, @"Collection view layout must be a flow layout to use the built-in inspector");
+    _defaultLayoutInspector = [[ASCollectionViewFlowLayoutInspector alloc] initWithCollectionView:self flowLayout:layout];
+  } else {
+    // Custom layouts will need to roll their own ASCollectionViewLayoutInspecting implementation and set a layout
+    // delegate. In the meantime ASDK provides a null layout inspector that does not provide any implementation
+    // and throws an exception for methods that should be implemented in the <ASCollectionViewLayoutInspecting>
+    _defaultLayoutInspector = [[ASCollectionViewNullLayoutInspector alloc] init];
```
levi: Could we provide a simple inspector by default? There are a few out there using a custom layout without implementing an inspector.

maicki:
@levi Added a simple layout inspector for custom UICollectionViewLayout's.

Furthermore for documentation purposes attached is a graph how this structure looks like now.
ascollectionviewlayoutinspecting

levi:
@maicki that diagram is fantastic. Thanks for putting that together!

maicki:
Recently a lot of requests for how to integrate a custom `UICollectionViewLayout`. We should improve the documentation for that after this PR landed.

cc @hannahmbanana

maicki:
@levi @Adlai-Holler Could you guys take another look over it when you find some time and let me know your thoughts if you think it would be worth to merge soon?

cc @appleguy

Adlai-Holler:
lgtm!

maicki:
@levi Can you get it a last look before I merge? Thanks!

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
```diff
@@ -47,12 +50,37 @@
...
+- (void)didChangeCollectionViewDataSource:(id<ASCollectionDataSource>)dataSource;
+
 @end
 
+/**
+ * A layout inspector for non-flow layouts that returns a constrained size to let the cells layout itself as
+ * far as possible based on the scrollable direction of the collection view. It throws exceptions for delegate
+ * methods that are related to supplementary node's management.
+ */
+@interface ASCollectionViewDefaultCustomLayoutInspector : NSObject <ASCollectionViewLayoutInspecting>
```
levi: Name of this is pretty long. Why not just `ASDefaultLayoutInspector`?

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
```diff
@@ -14,12 +14,15 @@
 #import <AsyncDisplayKit/ASDimension.h>
```
levi: Probably should rename this file or start pulling out some of the classes.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
```diff
@@ -47,12 +50,37 @@
...
+
+- (instancetype)init NS_UNAVAILABLE;
+- (instancetype)initWithCollectionView:(ASCollectionView *)collectionView NS_DESIGNATED_INITIALIZER;
+
+@end
+
+/**
+ * A layout inspector implementation specific for the sizing behavior of UICollectionViewFlowLayouts
+ */
 @interface ASCollectionViewFlowLayoutInspector : NSObject <ASCollectionViewLayoutInspecting>
```
levi: Rename this to ASFlowLayoutInspector

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -411,7 +409,9 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
 
   super.delegate = (id<UICollectionViewDelegate>)_proxyDelegate;
   
-  [_layoutInspector didChangeCollectionViewDelegate:asyncDelegate];
+  if ([_layoutInspector respondsToSelector:@selector(didChangeCollectionViewDelegate:)]) {
```
levi: Cache this when the inspector is set.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -248,11 +247,19 @@ - (instancetype)_initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionV
...
+    UICollectionViewFlowLayout *layout = (UICollectionViewFlowLayout *)self.collectionViewLayout;
+    ASDisplayNodeAssertNotNil(layout, @"Collection view layout must be a flow layout to use the built-in inspector");
+    _defaultLayoutInspector = [[ASCollectionViewFlowLayoutInspector alloc] initWithCollectionView:self flowLayout:layout];
+  } else {
+    // Custom layouts will need to roll their own ASCollectionViewLayoutInspecting implementation and set a layout
+    // delegate. In the meantime ASDK provides a null layout inspector that does not provide any implementation
+    // and throws an exception for methods that should be implemented in the <ASCollectionViewLayoutInspecting>
+    _defaultLayoutInspector = [[ASCollectionViewDefaultCustomLayoutInspector alloc] initWithCollectionView:self];
   }
+  _layoutInspector = _defaultLayoutInspector;
```
levi: Should set this up such that if you provide your own custom layout inspector, the default is never initialized. Probably best to do this in the layoutInspector getter and lazy initialize a default inspector as the inspector if nil.

levi: Also, collection view layouts can be changed after initialization by setting the collectionViewLayout property directly. This API should be able to handle this case.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -8,69 +8,172 @@
...
-@implementation ASCollectionViewFlowLayoutInspector {
-  BOOL _delegateImplementsReferenceSizeForHeader;
-  BOOL _delegateImplementsReferenceSizeForFooter;
+#define kDefaultItemSize CGSizeMake(50, 50)
+
+#pragma mark - Helper Functions
+
+// Returns a constrained size to let the cells layout itself as far as possible based on the scrollable direction
+// of the collection view
+static ASSizeRange ASDefaultConstrainedSizeForNodeForCollectionView(ASCollectionView *collectionView) {
```
levi: `NodeConstrainedSizeWithCollectionView`  No need for th AS prefix since its private to the implementation.

levi: Also make this inline

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -8,69 +8,172 @@
...
 - (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
 {
-  // TODO: Provide constrained size for flow layout item nodes
-  return ASSizeRangeMake(CGSizeZero, CGSizeZero);
+  if (_dataSourceFlags.implementsConstrainedSizeForNodeAtIndexPath) {
+    return [collectionView.asyncDataSource collectionView:collectionView constrainedSizeForNodeAtIndexPath:indexPath];
+  }
+  
+  CGSize itemSize = _layout.itemSize;
+  if (!CGSizeEqualToSize(itemSize, kDefaultItemSize)) {
```
levi: Use the == NO style we've started using recently.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -8,69 +8,172 @@
...
 
+- (NSUInteger)collectionView:(ASCollectionView *)collectionView numberOfSectionsForSupplementaryNodeOfKind:(NSString *)kind
+{
+  ASDisplayNodeAssert(NO, @"To support supplementary nodes in ASCollectionView, it must have a layoutInspector for layout inspection. (See ASCollectionViewFlowLayoutInspector for an example.)");
+  return 0;
+}
+
+- (NSUInteger)collectionView:(ASCollectionView *)collectionView supplementaryNodesOfKind:(NSString *)kind inSection:(NSUInteger)section
+{
+  ASDisplayNodeAssert(NO, @"To support supplementary nodes in ASCollectionView, it must have a layoutInspector for layout inspection. (See ASCollectionViewFlowLayoutInspector for an example.)");
```
levi: We should provide basic functionality without assertion. A few current ASDK users have custom layouts without supplementary nodes and thus don't have an inspector. We generally shouldn't require someone to implement an inspector unless they're doing unique sizing and supplementary node support that is very unique to their collection view layout.

levi: Nevermind that last comment. I now see what you're doing.

AsyncDisplayKit/ASCollectionView.mm
maicki: It's a pretty cheap call as I guess setting the delegate will not happen that often, but I will move it to a cached flag

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
maicki: I know I should ignore the comment, but just to add my comment so we are on the same page. If ASDK users do not have any supplementary nodes in a custom layout these methods should never be called. In case they want to have supplementary nodes they have to define it's own layout inspector and implement the functions. They could actually also subclass this base layout inspector and just overwrite the supplementary nodes methods and get the automatic constrained size calculation for free.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
maicki: I don't know if we should rename that as it would be an API change right? Do you think ASDK users using it somehow?

levi: It won't be a huge change for someone that might be subclassing this, however, I doubt anyone has needed to subclass this presently -- as you would have to rewrite flow layout to get a different supplementary view data source behavior.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
maicki: What about splitting it up into `ASDimension.h` and `ASSizeRange.h`? We already have `ASRelativeSize.h` so it would make sense to have them separated clearly.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
maicki: I think we should go with `ASCollectionViewLayoutInspector`. Just by reading `ASDefaultLayoutInspector` I would not really figure out why it's there or for what I would use it.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
maicki: As I propose to go with `ASCollectionViewLayoutInspector` I think we should not rename the class.  So it would be `ASCollectionViewLayoutInspector` and `ASCollectionViewFlowLayoutInspector`.

levi: Sounds good to me!
    
  
  
>

AsyncDisplayKit/ASCollectionView.mm
maicki: I'm not pretty sure if we should cache these calls as we have to be careful internally to always use the setter for the layout inspector and not the instance variable to wire up the flags. As I said the call should  not happen that often, frankly it will happen more often as someone could just set the delegate and if we not cache it we call only the delegate `respondToSelector:` call and not the data source one. If we cache it we always make it twice to cache it.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -406,7 +397,47 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
...
+    [self layoutInspector];
+  }
+}
+
+- (void)setLayoutInspector:(id<ASCollectionViewLayoutInspecting>)layoutInspector
+{
+  _layoutInspector = layoutInspector;
+  
+  _layoutInspectorFlags.layoutInspectorDidChangeCollectionViewDataSource = [_layoutInspector respondsToSelector:@selector(didChangeCollectionViewDataSource:)];
+  _layoutInspectorFlags.layoutInspectorDidChangeCollectionViewDelegate = [_layoutInspector respondsToSelector:@selector(didChangeCollectionViewDelegate:)];
```
maicki: @levi I don't know if we should cache the calls in here as I said in the other comment we need to be careful if the `_layoutInspector` changes to call the setter

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -406,7 +397,47 @@ - (void)setAsyncDelegate:(id<ASCollectionViewDelegate>)asyncDelegate
...
+    if ([layout asdk_isFlowLayout]) {
+      // Register the default layout inspector delegate for flow layouts only
+      _defaultLayoutInspector = [[ASCollectionViewFlowLayoutInspector alloc] initWithCollectionView:self flowLayout:layout];
+    } else {
+      // Register the default layout inspector delegate for custom collection view layouts
+      _defaultLayoutInspector = [[ASCollectionViewLayoutInspector alloc] initWithCollectionView:self];
+    }
+    
+    // Explicitly call the setter to wire up the _layoutInspectorFlags
+    self.layoutInspector = _defaultLayoutInspector;
```
maicki: @levi See here for example

maicki:
@levi Can you get it another look. All comments should be address from the last one.

appleguy:
Thanks for the development here, Michael!  Since Levi is overseas on time off, we'll wait until he's back before pushing this through.

maicki:
Some sample project around custom `UICollectionViewLayouts`: https://github.com/maicki/ASDKCustomCollectionViewSwift

maicki:
@levi Can you give it another look whenever you find some time. Thank you!

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -8,69 +8,172 @@
...
-@implementation ASCollectionViewFlowLayoutInspector {
-  BOOL _delegateImplementsReferenceSizeForHeader;
-  BOOL _delegateImplementsReferenceSizeForFooter;
+#define kDefaultItemSize CGSizeMake(50, 50)
+
+#pragma mark - Helper Functions
+
+// Returns a constrained size to let the cells layout itself as far as possible based on the scrollable direction
+// of the collection view
+static inline ASSizeRange NodeConstrainedSizeWithCollectionView(ASCollectionView *collectionView) {
```
levi: Make this `NodeConstrainedSizeForScrollDirection`

AsyncDisplayKitTests/ASCollectionViewTests.m
```diff
@@ -107,11 +107,12 @@ - (void)testThatItSetsALayoutInspectorForFlowLayouts
   XCTAssert([collectionView.layoutInspector isKindOfClass:[ASCollectionViewFlowLayoutInspector class]], @"should have a flow layout inspector by default");
 }
 
-- (void)testThatItDoesNotSetALayoutInspectorForCustomLayouts
+- (void)testThatISetALayoutInspectorForCustomLayouts
```
levi: `testThatADefaultLayoutInspectorIsProvidedForCustomLayouts`

levi:
Two comments. Otherwise, good to land!

maicki:
@appleguy @Adlai-Holler That should be ready to merge. If this will land we have to do a small change within Pinterest too so let me know at the time you would like to merge.

Adlai-Holler:
Lovely!

PR #1756, opened 2016-06-16
[ASTextNode] Move modifier to __weak for the highlighted layer

maicki:
Moving the modifier from __unsafe_unretained to __weak for the layer we use for highlighting.


Adlai-Holler:
👌

ocrickard:
lol, this is probably left from the days when this was a non-ARC file...

appleguy:
@ocrickard yes, that was exactly my suspicion when Michael found this - I think it was a mechanical conversion!  Funny stuff.  We should search the codebase for any others.

PR #1757, opened 2016-06-16
[ASCollectionView] Explicitly Retain CALayer Under iOS < 9 to Workaround Issue

Adlai-Holler:
This works around an issue where ASCollectionView may mysteriously be leaked and outlive its layer under iOS < 9. This only occurs under some view controller containment/transition situations and makes crashes virtually inevitable.

The leak does not seem to originate in ASDK. Manual investigation, plus the fact that it depends on iOS version & view controller transition/containment scheme, plus the fact that even if ASDK simply leaked the view, the layer should survive along with it, all point to an internal UIKit/Core Animation issue that was fixed in iOS 9. @maicki @appleguy


Adlai-Holler:
This issue will be worked-around at the application level.

Issue #1760, opened 2016-06-16
[ASEditableTextNode] calculateSizeThatFits measures small which causes text to bounce (baseline moves as you type)


vadims:
You can see the sample project demoing the issue: https://github.com/vadims/ASDKTextBounce

if you replace

```objc
- (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
{
  ASTextKitComponents *displayedComponents = [self isDisplayingPlaceholder] ? _placeholderTextKitComponents : _textKitComponents;
  CGSize textSize = [displayedComponents sizeForConstrainedWidth:constrainedSize.width];
  textSize = ceilSizeValue(textSize);
  return CGSizeMake(fminf(textSize.width, constrainedSize.width), fminf(textSize.height, constrainedSize.height));
}
```

with

```objc
- (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
{
  ASTextKitComponents *displayedComponents = [self isDisplayingPlaceholder] ? _placeholderTextKitComponents : _textKitComponents;
  CGSize textSize = [displayedComponents sizeForConstrainedWidth:constrainedSize.width];
  return CGSizeMake(fminf(ceilf(textSize.width), constrainedSize.width), fminf(ceilf(textSize.height), constrainedSize.height));
}
```

then the issue goes away. It seems like UITextView contentSize rounds up to the next point instead of device pixel?

maicki:
@vadims Hi Vadim, thanks for looking into that! Would you be so good and create a PR for that? That would be awesome!

hannahmbanana:
@vadims: great catch!

appleguy:
@vadims: thanks for creating a PR for us!

PR #1761, opened 2016-06-16
[ASEditableTextNode] Round up editable text node to the next point

vadims:
Rounding up to the next device pixel was calculating a height smaller
than UITextView's contentSize. This was causing the baseline to move as
the user was typing.


facebook-github-bot-3:

appleguy:
Solution to #1760

appleguy:
@vadims thanks for fixing this!  I'm pretty annoyed by the name of ceilSizeValue - it is not at all clear that it does screen pixel rounding rather than points (e.g. numerical rounding).

@vadims / @maicki it would be nice to include a short comment here clarifying that it is important to use points / UITextView apparently does sizing on a whole-point basis.  We should also rename the other method to more clearly describe its scale-dependent behavior.

Issue #1754, opened 2016-06-16
[ASMapNode] Setting region on ASMapNode doesn't work when liveMap = false


george-gw:
Scenario:
1- Create `ASMapNode`
2- Set region with `mapNode.region = /* new region */`

Expected result: snapshot shows the map at region
Actual result: snapshot shows the world map

Workaround:
1- Create new: `MKMapSnapshotOptions`, let's call it `newOptions`
2- Set same options as `mapNode.options` to it // for consistency's sake
3- Change `newOptions.region = /* new region */`
4- Set new options: `mapNode.options = newOptions`

hannahmbanana:
@george-gw: this also looks like it is a bug. I'll go ahead and submit a PR to fix it.

Issue #1753, opened 2016-06-16
[ASMapNode] Toggling liveMap on does not enable touches


george-gw:
Scenario:
1- Create `ASMapNode`
2- At some point during the app, toggle `liveMap = true`
3- Map does not receive touches
Solution:
when you set `liveMap = true`, set `mapNode.userInteractionEnabled = true`

Looking at the code, I saw that `self.userInteractionEnabled = YES` is only enabled in `didLoad` of `ASMapNode` when the node is created with `liveMap = true`, was there a reason for it to be that way rather than enabling and disabling it in `addLiveMap` and `removeLiveMap` respectively?

hannahmbanana:
@george-gw: Great question - thanks for watching out for the consistency of ASDK's API!

@aaronschubert0: It looks like you were the original author of this feature. Could you weigh in on this? Was this the intended behavior?

hannahmbanana:
@george-gw: I just checked out the code and this does seem to be an oversight rather than a feature. Do you feel confident putting up a pull request for this?

hannahmbanana:
@george-gw: I tested this and userInteraction _is_ enabled after setting liveMap=YES (sometime after creating the map node). However, it seems that the mapNode is capturing the touch events.

aaronschubert0:
@hannahmbanana @george-gw Yeah seems to be an oversight, strange it wasn't flag earlier. @george-gw happy for you to submit a PR for this if you wish (so you can test on your specific scenario), otherwise I can fix it.

george-gw:
@hannahmbanana I have added an Example in the pull request, I don't know if this should be there, but it's a way for you to test out the different solutions, you can actually take that code and run it on the old version (without my changes) to see the differences.
I have also NSLoged the userInteractionEnabled variable so you can see its value as you tap the toggle button to switch liveMap on and off.

PR #1762, opened 2016-06-17
[ASMapNode] bug fixes + added the ability to zoom in on the annotations + example project

george-gw:
Toggle user interaction when liveMap changes (#1753)
Change map snapshot when updating it with a new region (#1754)
Commented out code that is causing inaccurate behavior, this was wrapped with a FIXME
A minor change: Fixed typo in some comment

### DIFF
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -48,7 +48,7 @@
   _liveMap = NO;
 //  _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
   _annotations = @[];
-  _showAnnotationsState = ASMapNodeShowAnnotationsIgnored;
+  _showAnnotationsOptions = ASMapNodeShowAnnotationsOptionsIgnored;
   return self;
 }
 
@@ -278,8 +278,8 @@
     [weakSelf setNeedsLayout];
     [weakSelf.view addSubview:_mapView];
 
-    if (self.showAnnotationsState & ASMapNodeShowAnnotationsZoomed) {
-      BOOL const animated = self.showAnnotationsState & ASMapNodeShowAnnotationsAnimated;
+    if (self.showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsZoomed) {
+      BOOL const animated = self.showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsAnimated;
       [_mapView showAnnotations:_mapView.annotations animated:animated];
     }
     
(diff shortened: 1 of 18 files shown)


facebook-github-bot-3:

facebook-github-bot-1:

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -161,7 +160,18 @@ - (MKCoordinateRegion)region
...
+  MKMapSnapshotOptions * options = [[MKMapSnapshotOptions alloc] init];
+  options.camera = oldOptions.camera;
+  options.mapRect = oldOptions.mapRect;
+  options.mapType = oldOptions.mapType;
+  options.showsPointsOfInterest = oldOptions.showsPointsOfInterest;
+  options.showsBuildings = oldOptions.showsBuildings;
+  options.size = oldOptions.size;
+  options.scale = oldOptions.scale;
+  options.region = region;
+  self.options = options;
```
aaronschubert0: This seems overkill, is there a reason for this?

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -266,16 +277,17 @@ - (void)addLiveMap
     [weakSelf setNeedsLayout];
     [weakSelf.view addSubview:_mapView];
     
-    if (CLLocationCoordinate2DIsValid(_centerCoordinateOfMap)) {
-      [_mapView setCenterCoordinate:_centerCoordinateOfMap];
-    }
+//    if (CLLocationCoordinate2DIsValid(_centerCoordinateOfMap)) {
+//      [_mapView setCenterCoordinate:_centerCoordinateOfMap];
```
aaronschubert0: How come you're getting rid of this?

AsyncDisplayKit/ASMapNode.mm
george-gw: I agree, but I thought that since you're only changing the region, we shouldn't change the configuration that the developer might have had. I suspect that most of this will be left as default values and thus will be irrelevant.

I think the most important ones are the showsPointsOfInterest and showsBuildings, I'm not sure we need the rest, I just put them in for completion's sake.

AsyncDisplayKit/ASMapNode.mm
george-gw: I removed the line in removeLineMap regarding the centerCoordinate because it was causing the map to drift, I didn't actually try it with this line intact, besides, shouldn't ` [weakSelf applySnapshotOptions];` take care of that, since we're setting the region inside that method?

george-gw:
Added the ability to zoom in on annotations, similar to showAnnotations:animated: for MKMapView even while toggling the liveMap between ON/OFF

george-gw:
Refer to: #1792

Issue #853, opened 2015-11-17
[Wrapping Views] `initWithViewBlock:` node works initially, then fails to display


molon:
At first , sorry for my poor engilish.

I know we can use `ASTextNode` to draw text and it will provides more smooth performance.

But we need to use custom third-party lib in same cases. (we dont have much time to subclass `ASDisplayNode` for it)
Follow the docs, we can `addSubview` in `didLoad`, it's fine.
But I dont know how to use `layoutSpecThatFits:` with custom subview. maybe I can set a layout placeholder?

So I must find another way.
I test `initWithViewBlock:` method to create node for custom subview(UILabel for test, maybe another view actually):
```
//In a `ASCellNode`
_customNode = [[ASDisplayNode alloc]initWithViewBlock:^UIView *{
            UILabel *label = [[UILabel alloc]init];
            label.backgroundColor = [UIColor colorWithWhite:0.692 alpha:1.000];
            label.text = @"Hello World!";
            return label;
        }];
        _customNode.preferredFrameSize = CGSizeMake(100, 50);
        [self addSubnode:_customNode];
```
Then layout with `_customNode`.
The ASTableView's row number is 30.

First I scroll down,every would be ok. But  when I scroll back,the backgroundcolor and text in UILabel would be lost.(`Capture View Hierarchy` tells me the label is exist with preferredFrameSize)

I dont know why. So expect help.

Or How create a layout placeholder in `layoutSpecThatFits`?

appleguy:
@molon great question!  You found the right approach; if you want to use ASLayoutSpec with a traditional UIView, wrapping it with an ASDisplayNode is the best option.

I don't see any bugs with the code you just provided, so unfortunately the issue is a bit mysterious to me.  A few questions:

1) Can you try out using master?  There is an issue that is now fixed in master which affected rare cases when inserting rows and then scrolling down, and could cause content to disappear.  I doubt this is your problem, just want to make sure.
2) Could you provide a screenshot of the first display when it's apparently working correctly, and then a screenshot of when you have scrolled back and the content is incorrect?
3) You are not using any special features such as .isLayerBacked or .shouldRasterizeDescendants, right?
4) Understanding that this is not a solution - but still an important question - what features does ASTextNode lack that lead you to require the third-party library?
[...]

PR #1764, opened 2016-06-17
Dont clear view.layer.contents when layer is not kind of `_ASDisplayLayer`

molon:
Because only `_ASDisplayLayer` can auto re-create the contents.

https://github.com/facebook/AsyncDisplayKit/issues/853

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2093,7 +2093,9 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 - (void)clearContents
 {
   // No-op if these haven't been created yet, as that guarantees they don't have contents that needs to be released.
-  _layer.contents = nil;
+  if ([_layer isKindOfClass:[_ASDisplayLayer class]]) {
+    _layer.contents = nil;
+  }
   _placeholderLayer.contents = nil;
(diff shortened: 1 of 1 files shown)


facebook-github-bot-0:

facebook-github-bot-2:

hannahmbanana:
@molon: Thanks for submitting this PR! I think you may be accidentally mixing in some other changes here. Can you explain why you are including the other changes?

molon:
@hannahmbanana Sorry for my mistake. I reset the other changes now.

appleguy:
This is very interesting.

CA is actually capable of replacing contents for regular layers by the normal setNeedsDisplay process.  We may simply need to call that explicitly for non-async layers.  There is some concern with not clearing the contents because even for sync layers, they are using substantial amounts of memory which would never be freed until a memory warning.  Memory warnings are not guaranteed to be delivered.  A difference here is that we intend for nodes to be long lived, unlike views in UIKit reuse, so it is important to minimize their memory footprint when offscreen.

maicki:
This PR resolves the initial problem in a different way: #1776
There is also a longer description why just not clearing the layer contents if it's no ASDisplayLyer is not that optimal in terms of memory footprint.

molon:
Thanks for the comprehensive fix. It's great. https://github.com/facebook/AsyncDisplayKit/pull/1776

PR #1765, opened 2016-06-17
[ASDisplayNode] Resolve Deadlock Caused By removeFromSupernode

Adlai-Holler:
The prior implementation of removeFromSupernode would acquire _propertyLock and then attempt to lock the supernode (e.g. by calling `layerBacked` on it). If another thread is traversing down the node hierarchy at the same time, and holding a supernode's lock while trying to acquire the subnode's lock, we get a deadlock.

This patch never acquires supernode's lock while holding self's lock.

- Remove lock overlapping in `removeFromSupernode`
- Optimize `applySubnodeInsertions/Removals` for good measure
- Add some removeFromSupernode tests

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1518,6 +1518,7 @@ static NSInteger incrementIfFound(NSInteger i) {
 // NOTE: You must not called this method while holding the receiver's property lock. This may cause deadlocks.
 - (void)removeFromSupernode
 {
+  ASDisplayNodeAssertThreadAffinity(self);
   _propertyLock.lock();
     __weak ASDisplayNode *supernode = _supernode;
     __weak UIView *view = _view;
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1515,40 +1515,35 @@ static NSInteger incrementIfFound(NSInteger i) {
   [subnode __setSupernode:nil];
 }
 
+// NOTE: You must not called this method while holding the receiver's property lock. This may cause deadlocks.
 - (void)removeFromSupernode
 {
-  ASDisplayNodeAssertThreadAffinity(self);
-  BOOL shouldRemoveFromSuperviewOrSuperlayer = NO;
+  _propertyLock.lock();
+    __weak ASDisplayNode *supernode = _supernode;
+    __weak UIView *view = _view;
+    __weak CALayer *layer = _layer;
+    BOOL layerBacked = _flags.layerBacked;
+  _propertyLock.unlock();
   
-  {
-    ASDN::MutexLocker l(_propertyLock);
-    if (!_supernode)
-      return;
+  if (supernode == nil) {
+    return;
+  }
 
+  [supernode _removeSubnode:self];
+
+  if (self.nodeLoaded && supernode.nodeLoaded) {
     // Check to ensure that our view or layer is actually inside of our supernode; otherwise, don't remove it.
     // Though _ASDisplayView decouples the supernode if it is inserted inside another view hierarchy, this is
     // more difficult to guarantee with _ASDisplayLayer because CoreAnimation doesn't have a -didMoveToSuperlayer.
-    
-    if (self.nodeLoaded && _supernode.nodeLoaded) {
-      if (_flags.layerBacked || _supernode.layerBacked) {
-        shouldRemoveFromSuperviewOrSuperlayer = (_layer.superlayer == _supernode.layer);
-      } else {
-        shouldRemoveFromSuperviewOrSuperlayer = (_view.superview == _supernode.view);
-      }
-    }
-  }
-  
-  // Do this before removing the view from the hierarchy, as the node will clear its supernode pointer when its view is removed from the hierarchy.
-  // This call may result in the object being destroyed.
-  [_supernode _removeSubnode:self];
-
-  if (shouldRemoveFromSuperviewOrSuperlayer) {
     ASPerformBlockOnMainThread(^{
-      ASDN::MutexLocker l(_propertyLock);
-      if (_flags.layerBacked) {
-        [_layer removeFromSuperlayer];
+      if (layerBacked || supernode.layerBacked) {
+        if (layer.superlayer == supernode.layer) {
+          [layer removeFromSuperlayer];
+        }
       } else {
-        [_view removeFromSuperview];
+        if (view.superview == supernode.view) {
+          [view removeFromSuperview];
+        }
       }
     });
   }
--- a/AsyncDisplayKit/Private/ASLayoutTransition.mm
+++ b/AsyncDisplayKit/Private/ASLayoutTransition.mm
@@ -48,9 +48,12 @@
 {
   ASDN::MutexLocker l(_propertyLock);
   [self calculateSubnodeOperationsIfNeeded];
-  for (NSUInteger i = 0; i < [_insertedSubnodes count]; i++) {
+  
+  NSUInteger i = 0;
+  for (ASDisplayNode *node in _insertedSubnodes) {
     NSUInteger p = _insertedSubnodePositions[i];
-    [_node insertSubnode:_insertedSubnodes[i] atIndex:p];
+    [_node insertSubnode:node atIndex:p];
+    i += 1;
   }
 }
 
@@ -58,8 +61,8 @@
 {
   ASDN::MutexLocker l(_propertyLock);
   [self calculateSubnodeOperationsIfNeeded];
-  for (NSUInteger i = 0; i < [_removedSubnodes count]; i++) {
-    [_removedSubnodes[i] removeFromSupernode];
+  for (ASDisplayNode *subnode in _removedSubnodes) {
+    [subnode removeFromSupernode];
   }
 }
 
--- a/AsyncDisplayKitTests/ASDisplayNodeTests.m
+++ b/AsyncDisplayKitTests/ASDisplayNodeTests.m
@@ -1638,6 +1638,57 @@ static inline BOOL _CGPointEqualToPointWithEpsilon(CGPoint point1, CGPoint point
   [c release];
 }
 
+- (void)testRemoveFromViewBackedLoadedSupernode
+{
+  DeclareNodeNamed(a);
+  DeclareNodeNamed(b);
+  [b addSubnode:a];
+  [a view];
+  [b view];
+  XCTAssertNodesLoaded(a, b);
+  XCTAssertEqual(a.supernode, b);
+  XCTAssertEqual(a.view.superview, b.view);
+  
+  [a removeFromSupernode];
+  XCTAssertNil(a.supernode);
+  XCTAssertNil(a.view.superview);
+}
+
+- (void)testRemoveFromLayerBackedLoadedSupernode
+{
+  DeclareNodeNamed(a);
+  a.layerBacked = YES;
+  DeclareNodeNamed(b);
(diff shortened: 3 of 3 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1515,40 +1515,36 @@ - (void)_removeSubnode:(ASDisplayNode *)subnode
   [subnode __setSupernode:nil];
 }
 
+// NOTE: You must not called this method while holding the receiver's property lock. This may cause deadlocks.
```
Adlai-Holler: Someone correctly mentioned that we need a more comprehensive locking refactor. Comments like this bear that out.

I've manually verified that this condition is currently satisfied. Unfortunately I don't know of any way to assert that the current thread is not holding a given lock, so we don't have great enforcement options here =/

Adlai-Holler:
@maicki @ocrickard @appleguy

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1515,40 +1515,36 @@ - (void)_removeSubnode:(ASDisplayNode *)subnode
...
 - (void)removeFromSupernode
 {
   ASDisplayNodeAssertThreadAffinity(self);
-  BOOL shouldRemoveFromSuperviewOrSuperlayer = NO;
+  _propertyLock.lock();
+    __weak ASDisplayNode *supernode = _supernode;
+    __weak UIView *view = _view;
+    __weak CALayer *layer = _layer;
+    BOOL layerBacked = _flags.layerBacked;
+  _propertyLock.unlock();
```
ocrickard: nit: can we use a scoped locker still? So:
```
{
    __weak ASDisplayNode *supernode = _supernode;
    __weak UIView *view = _view;
    __weak CALayer *layer = _layer;
    BOOL layerBacked = _flags.layerBacked;
}
```

The lockers are exception-safe through RAII semantics, while manual lock/unlock will remain deadlocked if an exception is thrown between the calls (for instance, if something insane happens inside layerBacked). Probably not an issue here, but worth preserving the safety for the future.

AsyncDisplayKit/ASDisplayNode.mm
ocrickard: Yeah, this is needed pretty soon. Seems like a bit of technical debt for locking procedures.

ocrickard:
Out of curiosity how are you guys finding deadlocks? Do you have a watchdog implemented somewhere in ASDK or in the client app?

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: Unfortunately we can't use a scoped locker because we want those variables (supernode, view, layer, etc) to exist in the larger scope. So I could declare them before the scope and then assign them new values but it'd be kind of messy. 

Since I don't call any methods in this scope, no exception can be thrown so I think we're good. Agreed?

ocrickard: lol, right. I need to read better.

You can still declare above in the proper scope, and use a locker when you set the values.

Adlai-Holler:
In this case the deadlock was observed in the client app by rigorous futzing.

ocrickard:
Cool, when we were looking for some deadlocks in ComponentKit we started applying a watchdog in the infra. ASDK is less opinionated about its threading model, so may not be valuable in infra, but here's what we did:

https://github.com/facebook/componentkit/blob/b7089d2c38a520279d6236fbe96af12e11d3b976/ComponentKit/Utilities/CKWatchdogTimer.h

Internally we also use an app-wide watchdog thread that checks the main thread for deadlocks and saves information about the callstacks to disk. Seems like ASDK could provide something similar for users of the framework to catch those rare lock inversion cases.

Adlai-Holler:
It would be great to add a watchdog to ASDK sometime, once we figure out the specifics of its role and implementation. Thanks for the link!

ocrickard:
Merge request looks reasonable to me.

Adlai-Holler:
Since this code is so critical, I need to add some additional context/validation here. After this range controller change I'm working on I'll attempt to create a test project that reproduces this issue and validates it's resolved. For the time being (and so I don't lose it!) here's the simplified backtrace that caused the deadlock during random futzing:

```
  thread #1: tid = 0x6c9bf6, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #6: 0x000000010362c349 Pinterest Enterprise`ASDN::Locker<ASDN::Mutex>::Locker(this=0x00007fff5d473ae8, l=0x00007d6000032828) + 57 at ASThread.h:68
    frame #7: 0x00000001045b48fd Pinterest Enterprise`::-[ASDisplayNode isLayerBacked](self=0x00007d6000032800, _cmd="isLayerBacked") + 93 at ASDisplayNode.mm:597
    frame #8: 0x00000001045c9218 Pinterest Enterprise`::-[ASDisplayNode removeFromSupernode](self=0x00007d6000031800, _cmd="removeFromSupernode") + 1336 at ASDisplayNode.mm:1525
    frame #9: 0x000000010463b196 Pinterest Enterprise`::-[ASLayoutTransition applySubnodeRemovals](self=0x00007d2c00057030, _cmd="applySubnodeRemovals") + 486 at ASLayoutTransition.mm:58
    frame #10: 0x00000001045b8c46 Pinterest Enterprise`::-[ASDisplayNode didCompleteLayoutTransition:](self=0x00007d6000032800, _cmd="didCompleteLayoutTransition:", context=0x00007d0800134420) + 182 at ASDisplayNode.mm:854
    frame #11: 0x00000001033431b2 Pinterest Enterprise`-[PIPinCloseupMetadataNode didCompleteLayoutTransition:](self=0x00007d6000032800, _cmd="didCompleteLayoutTransition:", context=0x00007d0800134420) + 178 at PIPinCloseupMetadataNode.m:282
    frame #12: 0x00000001045b8d9e Pinterest Enterprise`::-[ASDisplayNode transitionContext:didComplete:](self=0x00007d6000032800, _cmd="transitionContext:didComplete:", context=0x00007d0800134420, didComplete=YES) + 174 at ASDisplayNode.mm:863
    frame #13: 0x0000000104504e19 Pinterest Enterprise`-[_ASTransitionContext completeTransition:](self=0x00007d0800134420, _cmd="completeTransition:", didComplete=YES) + 169 at _ASTransitionContext.m:92
    frame #14: 0x00000001032f9938 Pinterest Enterprise`__69+[PIBasicLayoutTransitionController animateLayoutTransition:forNode:]_block_invoke.56((null)=<unavailable>, finished=YES) + 776 at PIBasicLayoutTransitionController.m:83
    frame #15: 0x000000010af19392 UIKit`-[UIViewAnimationBlockDelegate _didEndBlockAnimation:finished:context:] + 668
    frame #16: 0x000000010aeecc87 UIKit`-[UIViewAnimationState sendDelegateAnimationDidStop:finished:] + 222
    frame #17: 0x000000010aeed17e UIKit`-[UIViewAnimationState animationDidStop:finished:] + 136
    frame #18: 0x000000010a99b970 QuartzCore`CA::Layer::run_animation_callbacks(void*) + 316

  thread #48: tid = 0x6ca25b, queue = 'com.pinterest.closeupPinUpdateOperationQueue'
    frame #6: 0x000000010362c349 Pinterest Enterprise`ASDN::Locker<ASDN::Mutex>::Locker(this=0x00007e8000c680f8, l=0x00007d6000031828) + 57 at ASThread.h:68
    frame #7: 0x00000001045d628d Pinterest Enterprise`::-[ASDisplayNode hierarchyState](self=0x00007d6000031800, _cmd="hierarchyState") + 93 at ASDisplayNode.mm:2290
    frame #8: 0x00000001045b83a2 Pinterest Enterprise`::__50-[ASDisplayNode cancelLayoutTransitionsInProgress]_block_invoke((null)=0x00000001051e2780, node=0x00007d6000031800) + 130 at ASDisplayNode.mm:802
    frame #9: 0x00000001045e4234 Pinterest Enterprise`::ASDisplayNodePerformBlockOnEveryNode(layer=0x0000000000000000, node=0x00007d6000031800, block=0x00000001045b8320)(ASDisplayNode *)) + 2420 at ASDisplayNodeExtras.mm:49
    frame #10: 0x00000001045e5604 Pinterest Enterprise`::ASDisplayNodePerformBlockOnEverySubnode(node=0x00007d6000032800, block=0x00000001045b8320)(ASDisplayNode *)) + 660 at ASDisplayNodeExtras.mm:89
    frame #11: 0x00000001045b82e0 Pinterest Enterprise`::-[ASDisplayNode cancelLayoutTransitionsInProgress](self=0x00007d6000032800, _cmd="cancelLayoutTransitionsInProgress") + 224 at ASDisplayNode.mm:801
    frame #12: 0x0000000103340de1 Pinterest Enterprise`-[PIPinCloseupMetadataNode setNodes:](self=0x00007d6000032800, _cmd="setNodes:", nodes=@"9 elements") + 129 at PIPinCloseupMetadataNode.m:98
    frame #13: 0x0000000103342093 Pinterest Enterprise`-[PIPinCloseupMetadataNode buildSubnodesForRelayoutForPin:withOldPin:](self=0x00007d6000032800, _cmd="buildSubnodesForRelayoutForPin:withOldPin:", newPin=0x00007d4c00259740, oldPin=0x00007d4c00259900) + 2499 at PIPinCloseupMetadataNode.m:211
    frame #14: 0x00000001033414e3 Pinterest Enterprise`__39-[PIPinCloseupMetadataNode pinUpdated:]_block_invoke((null)=<unavailable>) + 499 at PIPinCloseupMetadataNode.m:139
    frame #15: 0x0000000106ad9057 libclang_rt.tsan_iossim_dynamic.dylib`__tsan::invoke_and_release_block(void*) + 7

  thread #49: tid = 0x6ca25e, queue = 'com.apple.root.default-qos'
    frame #6: 0x000000010362c349 Pinterest Enterprise`ASDN::Locker<ASDN::Mutex>::Locker(this=0x00007e8000df1c20, l=0x00007d6000032828) + 57 at ASThread.h:68
    frame #7: 0x00000001045b89d6 Pinterest Enterprise`::-[ASDisplayNode _shouldAbortTransitionWithID:](self=0x00007d6000032800, _cmd="_shouldAbortTransitionWithID:", transitionID=2) + 102 at ASDisplayNode.mm:842
    frame #8: 0x00000001045b6de3 Pinterest Enterprise`::__97-[ASDisplayNode transitionLayoutWithSizeRange:animated:shouldMeasureAsync:measurementCompletion:]_block_invoke.229((null)=<unavailable>) + 1011 at ASDisplayNode.mm:719
    frame #9: 0x0000000106ad9057 libclang_rt.tsan_iossim_dynamic.dylib`__tsan::invoke_and_release_block(void*) + 7
```

You can see that thread 48 is traversing down the node hierarchy. `cancelLayoutTransitionsInProgress` locks self, and then updates the hierarchy state of the entire subtree (taking each of those locks). So the locks overlap "downward."

Thread 1 in `removeFromSupernode` locks self, and then attempts to read `supernode.layerBacked` which takes the lock. So the locks overlap "upward."

We should avoid this overlapping behavior as much as possible. If it has to happen, we should do it in a consistent order to avoid deadlocks. "Downward" is the natural default here – it's how trees work – so this PR removes the upward overlapping in `removeFromSupernode`. @appleguy

Adlai-Holler:
I've created a demo project:
[Deadlock1765.zip](https://github.com/facebook/AsyncDisplayKit/files/324362/Deadlock1765.zip)

Simply run it and you'll see it deadlocks. Switch the commenting in `Podfile` and `pod install` to get the revision after the change, re-run it, and you'll see it doesn't deadlock. The code is all in AppDelegate.m

PR #1766, opened 2016-06-17
[ASCollectionView, ASCellNode] Add support applyLayoutAttributes: on ASCellNode

hannahmbanana:
@Eke discovered that we don't support applyLayoutAttributes: on ASCellNode. 

He has already tested this patch on one collection and it worked. He has 4 other collection views that need custom animations, etc and so he'll update us as he tests it on each of those collections.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -99,6 +99,8 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 - (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 - (void)touchesCancelled:(nullable NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 
+- (void)applyLayoutAttributes:(UICollectionViewLayoutAttributes *)layoutAttributes;
+
 /**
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASCellNode.mm
```diff
@@ -211,6 +211,10 @@ - (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event
 
 #pragma clang diagnostic pop
 
+- (void)applyLayoutAttributes:(UICollectionViewLayoutAttributes *)layoutAttributes
+{
+}
```
maicki: Can you add a comment why this empty and add the purpose for it to the header file. Here could be something like:  `// To be overriden by subclasses`

AsyncDisplayKit/ASCellNode.h
```diff
@@ -99,6 +99,8 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 - (void)touchesEnded:(NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 - (void)touchesCancelled:(nullable NSSet<UITouch *> *)touches withEvent:(nullable UIEvent *)event ASDISPLAYNODE_REQUIRES_SUPER;
 
+- (void)applyLayoutAttributes:(UICollectionViewLayoutAttributes *)layoutAttributes;
```
maicki: Please add a comment for what this method is for.

hannahmbanana:
@maicki: You'd think I'd know better than to forget the documentation. :)

Issue #1179, opened 2016-02-04
[Pedantic] Replace Percent with Fraction where Appropriate


Adlai-Holler:
"Percent" refers specifically to the portion out of 100 units. Apple uses the word "fraction" in NSProgress to refer to this [0, 1] kind of value and I think it's very appropriate.

So I propose changing `ASRelativeDimensionTypePercent` to `ASRelativeDimensionTypeFraction` and all the related names, and update the word "percentage" in the documentation to "fraction".

I'm glad to knock the change itself out, but only if y'all are interested. Since it's breaking API it might be appropriate for 2.0.

nguyenhuy:
I think it is a reasonable change. Thanks for bringing it up :+1:

appleguy:
I _love_ this suggestion.  Hadn't considered it but strongly agree.  We should do this before 2.0.

appleguy:
The community is planning an exciting long term road map for the project and getting organized around how to deliver these feature requests. We are moving to a new issue tracker with more sophisticated planning capabilities (check out www.realArtists.com and their new product: Ship!). 

If you are interested in helping contribute to this component or any other, don’t hesitate to send me an email at AsyncDisplayKit@gmail.com. If you would like to contribute for a few weeks, I can also add you to our Ship bug tracker so that you can see what everyone is working on and actively coordinate with us. 

As always, keep filing issues and submitting pull requests here on Github and we will only move things to the new tracker if they require long term coordination.

PR #1767, opened 2016-06-17
[ASRelativeDimensionType] Replace "percent" with "fraction" Where Appropriate

Adlai-Holler:
Because we have the best words! The return of #1179


Adlai-Holler:
@hannahmbanana @maicki @nguyenhuy @appleguy take a looky-loo?

Adlai-Holler:
Oops the snapshot reference files need renaming. One min

maicki:
In general looks good to me. We just have to be careful as I think this will be an API breaking change?

@Adlai-Holler Is the plan to merge that before 2.0? cc @appleguy

Adlai-Holler:
Since this is so harmless, I think we should leave it open and merge just before releasing 2.0. Should be trivial to rebase as needed. I hate to let it hang around but it kept popping into my mind during the day!

hannahmbanana:
Could we label it with a "breaking API" or "2.0 release" type of label? I think we're going to run into some more of these.

hannahmbanana:
When this gets merged, please uncomment the assertions on lines 23-24 in `ASRelativeDimension.mm`

```
    // TODO: Enable this assertion for 2.0.  Check that there is no use case for using a larger value, e.g. to layout for a clipsToBounds = NO element.
    // ASDisplayNodeCAssert( 0 <= value && value <= 1.0, @"ASRelativeDimension percent value (%f) must be between 0 and 1.", value);
```

maicki:
@Adlai-Holler Can you please rebase your branch. Thanks!

hannahmbanana:
Looks like examples/AnimatedGIF needs to be updated

Issue #1242, opened 2016-02-18
[ASTableView] Add Optional constrainedSizeForRowAtIndexPath: Data Source Method


Adlai-Holler:
This would correspond to the `ASCollectionDataSource` method

```objectivec
- (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath;
```

The widths in the returned size range would be ignored. This would make it easy for the data source to specify min/max row heights.

- Add the optional selector to the ASTableDataSource protocol with documentation.
- Add a test that the returned size is enforced.
- Add a flag for whether the data source responds to the selector.
- In `-[ASTableView dataController:constrainedSizeForNodeAtIndexPath:]`, check that flag and consult the data source if possible otherwise use the current unconstrained values.

I can tackle this in the next couple days but if someone else has bandwidth go for it!

levi:
We also need this for collection views if we are going to have the cells self-size before the layout accesses them.

appleguy:
@levi it does exist for collection today, right?  just checking.

@Adlai-Holler thanks for filing this.  @nguyenhuy confirmed it would be easy to add after his recent refactor, so we should try to do this in the next month before 2.0GM.

appleguy:
PR #1769 closes this issue.

PR #1740, opened 2016-06-10
[ASCollectionView] Move constrainedSizeForNodeAtIndexPath from the DataSource to the Delegate

tberman:
Small change w/ some associated cleanup, this feels like a delegate method, not a datasource method.

PR #1769, opened 2016-06-18
[ASTableView] Add constrainedSizeForRowAtIndexPath: to control row heights from delegate

hannahmbanana:
Addresses this suggestion: #1242

### DIFF
--- a/AsyncDisplayKitTests/ASTableViewTests.m
+++ b/AsyncDisplayKitTests/ASTableViewTests.m
@@ -139,10 +139,10 @@
 
 @end
 
-@interface ASTableViewFilledDataSourceWithRowSize : ASTableViewFilledDataSource
+@interface ASTableViewFilledDelegate : NSObject <ASTableViewDelegate>
 @end
 
-@implementation ASTableViewFilledDataSourceWithRowSize
+@implementation ASTableViewFilledDelegate
 
 - (ASSizeRange)tableView:(ASTableView *)tableView constrainedSizeForRowAtIndexPath:(NSIndexPath *)indexPath
 {
@@ -165,9 +165,10 @@
   ASTestTableView *tableView = [[ASTestTableView alloc] __initWithFrame:CGRectMake(0, 0, 100, 400)
                                                                   style:UITableViewStylePlain];
   
-  ASTableViewFilledDataSourceWithRowSize *dataSource = [ASTableViewFilledDataSourceWithRowSize new];
+  ASTableViewFilledDelegate *delegate = [ASTableViewFilledDelegate new];
+  ASTableViewFilledDataSource *dataSource = [ASTableViewFilledDataSource new];
 
-  tableView.asyncDelegate = dataSource;
+  tableView.asyncDelegate = delegate;
   tableView.asyncDataSource = dataSource;
   
   [tableView reloadDataImmediately];
--- a/AsyncDisplayKitTests/ASTableViewTests.m
+++ b/AsyncDisplayKitTests/ASTableViewTests.m
@@ -128,7 +128,6 @@
   return textCellNode;
 }
 
-
 - (ASCellNodeBlock)tableView:(ASTableView *)tableView nodeBlockForRowAtIndexPath:(NSIndexPath *)indexPath
 {
   return ^{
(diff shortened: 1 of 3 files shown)


levi:
Thoughts on adding this to the delegate instead?

levi:
Relevant PR: #1740

hannahmbanana:
@levi @maicki: I put it in the data source to be consistent with the _current_ state of ASCollectionView. Should I switch it to the delegate assuming #1740 will eventually be committed?

appleguy:
I had discussed this with Hannah.  I suggested the Data Source because it is unclear to me if we will land the other PR.  If we do, unfortunately we will need to introduce code that detects the method existing on the data source.  Otherwise it is a very problematic “silent” error that could cause a lot of grief for app developers who update to 2.0, but don’t happen to test a deep corner of their app that uses this method.  I’ve seen in the community it is quite common to not set the .delegate, whether or not this is advisable (it is supported by Apple).

There is also a close parallelism between the nodeFor: and constrainedSizeFor: methods, but ultimately we need to make a decision on the other PR’s course to landing safely before deciding if this should also be in the delegate.

levi:
@appleguy the other diff was updated to introduce the data source detection note. If you're concerned about a breaking change, we can come up with a plan to assert or warn of the deprecation in an upcoming 2.0 beta and introduce a porting guide to 2.0 when we finalize the release.

The benifit I see here of having these methods on the delegate is that they're are optional and geometry-defining. A quick glance at UITableViewDataSource and UICollectionViewDataSource will show that there is a strict set of responsibilities for the data source that break down to defining count, data, and responding to insertion/deletion events. The delegate protocols, on the other hand have a slew of geometry-based methods, defining size, insets — much like our sizeForNode methods. I think there is some benifit here keeping the data source light and moving optional methods like this to the delegate. It will keep the data source simple and clear on what needs to be defined to support a functional list of nodes.

maicki:
@appleguy I agree with @levi here. I would suggest moving this method to the delegate to be aligned with `UITableViewDataSource` as well as to have the responsibility on the right place: A DataSource and Delegate have specific responsibilities and the responsibilities for sizing lays in the delegate.

I also think we should move the same method slowly to `ASCollectionViewDelegate` as the other diff does with 2.0. Like @levi said via a slow deprecation and a porting guide.

I will be a bit confusing before 2.0 to have the same method on different protocols between table and collection view, but it is getting even more confusing by introducing this method in the data source and deprecate it slowly after.

hannahmbanana:
Looks like the consensus is start the table method out in the delegate (non-consistent protocol placements between table & collection) and then later move the collection method to the delegate. I will update accordingly.

AsyncDisplayKitTests/ASTableViewTests.m
```diff
@@ -140,12 +139,51 @@ - (ASCellNodeBlock)tableView:(ASTableView *)tableView nodeBlockForRowAtIndexPath
 
 @end
 
+@interface ASTableViewFilledDataSourceWithRowSize : ASTableViewFilledDataSource
```
maicki: As you moved the methods to the delegate can you also change all of the names in the test from delegate to dataSource please.

AsyncDisplayKitTests/ASTableViewTests.m
```diff
@@ -140,12 +139,52 @@ - (ASCellNodeBlock)tableView:(ASTableView *)tableView nodeBlockForRowAtIndexPath
...
+  
+  [tableView reloadDataImmediately];
+  [tableView setNeedsLayout];
+  [tableView layoutIfNeeded];
+  
+  for (int section = 0; section < NumberOfSections; section++) {
+      for (int row = 0; row < NumberOfRowsPerSection; row++) {
+        NSIndexPath *indexPath = [NSIndexPath indexPathForRow:row inSection:section];
+        CGRect rect = [tableView rectForRowAtIndexPath:indexPath];
+        XCTAssertEqual(rect.size.width, 100);  // specified width should be ignored for table
```
appleguy: Thanks for adding tests!

appleguy:
@levi @maicki you make a good case.  I certainly agree it's more appropriate there, just wondering what the path is to move the other method as it has several considerations.  Let's go with this route, and sign up for full compatibility with the collection view method existing on the .dataSource only.  Thanks @hannahmbanana !

Issue #1771, opened 2016-06-18
[ASCollectionView] Deleting first item - inconsistency with associated supplementary node header


tungvoduc:
I have a supplementary header node in the collection view and when I delete the first node, after that scroll it upward. I get crash immediately. 

![crash](https://cloud.githubusercontent.com/assets/6437873/16172389/77be2204-3590-11e6-9686-0f487a8dd6b7.png)

hannahmbanana:
@tungvoduc: Thanks for reporting this! Would it be possible to submit a sample project showing this behavior? Command + D on the project folder will create a copy of the project. Feel free to email it to AsyncDisplayKit@gmail.com if you don't want to post it here.

tungvoduc:
@hannahmbanana , I actually repro with the sample project CustomCollectionView which is using MosaicCollectionViewLayout. 

Easy way to make is crash to to put these lines into the view controller implementation file:
`
-(void)collectionView:(UICollectionView *)collectionView didSelectItemAtIndexPath:(NSIndexPath *)indexPath {
  [[_sections objectAtIndex:indexPath.section] removeObjectAtIndex:indexPath.row];
  [_collectionView deleteItemsAtIndexPaths:[NSArray arrayWithObjects:indexPath, nil]];
}
`

Then tapping some random item to delete it and scroll up and down and it will crash.

hannahmbanana:
@tungvoduc: I confirmed that this is a regression after merging PR #1629 (cc @erichoracek). 

@maicki: Are you familiar with this change?

erichoracek:
@hannahmbanana I would get an assertion after _any_ item-level mutation when the collection view had supplementary nodes for each row (not just deletes of the last item) before my pull request. Section-level mutations worked without issue. This crash was already documented in #1278 and occurred in the same place as is documented here. ("Supplementary node should exist.")

If this behavior is reverted (#1772), we'd be trading crashes in some scenarios for crashes in other scenarios. Perhaps there is a logical error in the recalculation code I wrote? I'll attempt to clone down the example project and see if I can find out anything.

Apologies for introducing a regression!

appleguy:
Hey Eric - thanks for your note!  Since the patch is fairly large, I doubt we will actually push a revert, but it is concerning that this scenario (easy to reproduce) is indeed an item-level mutation that did not crash previously.

[...]

PR #1772, opened 2016-06-18
Revert "[ASCollectionView] Repopulate supplementary views on item-level changes"

appleguy:
Reverts facebook/AsyncDisplayKit#1629.  Not sure if we should do this yet, otherwise need to fix-forward #1771


hannahmbanana:
@appleguy should we close this in favor of #1773?

PR #1777, opened 2016-06-19
[ASTextNode / ASImageNode] Use structs for drawing parameter

maicki:
The purpose of this PR is to move from an ObjC object to a struct for drawing parameters in ASTextNode and ASImageNode.

Furthermore as we improved the performance by going from the class drawing method to an instance drawing method we prevent custom drawing in ASTextNode or ASImageNode subclasses now.
If we decide to merge that PR we should close: #1728 as this is not necessary anymore.

cc @appleguy @ocrickard @garrettmoon: Curious about your thoughts about this PR.


Adlai-Holler:
Wonderful ⚡

PR #1785, opened 2016-06-20
[ASCollectionView/ASTableView] Warn Users About Index Paths in willDisplayNode:/didEndDisplayingNode: Methods

Adlai-Holler:
The index paths passed into these methods may not correspond to the same item in the user's data source as they expect, if there are pending edit operations in the collection view. This is not obvious to users and we've received reports of issues caused by users passing index paths from these methods into their data source.

This is one part of a broader issue for users of ASDK: the separation between "data source index space" and "completed nodes index space." Given an NSIndexPath that comes from the table view, the user may not be safe to query their data source with it if there are enqueued edit operations. Likewise, the user may not be safe to query the table view with an index path from their data source.

Since these methods don't appear to be widely used in client apps, for now we should deprecate them. I considered also adding new `tableView:willDisplayNode:` etc methods which omit the confusing index paths, but under the bias for subtraction I say we start by doing this. We can easily add those methods in the future.

Ideally we would add API to convert between the index spaces. It would be quite expensive to achieve – we would need a queue of changesets in the data controller (hard) and methods to convert index paths bi-directionally across a given changeset (easy). We would also need to reset this mapping for reloadData etc. Here's how the API could be:

```
-[ASTableView indexPathForDataSourceIndexPath:]
-[ASTableView dataSourceIndexPathForIndexPath:]
-[ASTableView sectionForDataSourceSection:]
-[ASTableView dataSourceSectionForSection:]
```

Thoughts @appleguy @maicki

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -430,19 +430,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @optional
 
-- (void)collectionView:(ASCollectionView *)collectionView willDisplayNodeForItemAtIndexPath:(NSIndexPath *)indexPath;
-
-/**
- * Informs the delegate that the collection view did remove the provided node from the view hierarchy.
- * This may be caused by the node scrolling out of view, or by deleting the item
- * or its containing section with @c deleteItemsAtIndexPaths: or @c deleteSections: .
- * 
- * @param collectionView The sender.
- * @param node The node which was removed from the view hierarchy.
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
Update: On reflecting more, I've decided to apply the bias for subtraction to my subtraction and not deprecate these methods formally. Instead I've just added documentation about the index paths. It is still a somewhat-useful code path and it eases the transition away from UIKit.

maicki:
LGTM let's merge that in.

Issue #563, opened 2015-07-13
[ASNetworkImageNode] placeholder will fadeout twice


yxztj:
Hi,

When I'm using ASNetworkImageNode the placeholder is not working as we expected.
In the Placeholder sample which is included in the repo, the logo image is set to imageNode right after node is initialized. The placeholder is working just fine.
```swift
  // ASImageNode gets placeholders by default
  _imageNode = [[SlowpokeImageNode alloc] init];
  _imageNode.image = [UIImage imageNamed:@"logo"];
```

But if I delay setting the image by several seconds(5 seconds for example) as follows:
```swift
  // ASImageNode gets placeholders by default
  _imageNode = [[SlowpokeImageNode alloc] init];
    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(5.0 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
        _imageNode.image = [UIImage imageNamed:@"logo"];
    });
```
The placeholderlayer will fadeout at start, then removed from superlayer, and when setting image on 5 seconds later, the placeholderlayer is added to node layer again, fadeout, and removed from superlayer:
![image2times](https://cloud.githubusercontent.com/assets/3334458/8649250/f7a89c90-299a-11e5-9882-2b638d9d533b.gif)

After some research on the source code, I found that the placeholderlayer is added to node on each displayWillStart, and fadeout on each displayDidFinished(). 
As a result in ASNetworkImageNode, the placeholder is fading out twice(once on initial display, and another on image downloaded). Is this the expected behavior or am I missing anything? How can I prevent the first fade out from happening, since the network image hasn't finished downloading yet?

appleguy:
@yxztj Just wanted to say that I saw this issue when you submitted it, and still intend to fix.  Sorry for the delay.  The placeholder fade-out feature was added by @rnystrom a while back, and hasn't gotten a lot of usage / testing yet.  That's not a good reason for it to have bugs, but suffice it to say that there are a few cool ways we could improve both the API and the implementation.

Right now I'm working on the display system, which is closely related to how placeholders are shown and removed.  Because it is still possible for a client to implement their own placeholder mechanics, I will have to prioritize this after the other display work, but it may well be possible to fix at the same time.

Thanks for trying out the framework, and let us know if you have any other issues or questions!

yxztj:
@appleguy Thanks for the reply. The framework is working just awesome :)
[...]

PR #1786, opened 2016-06-21
[ASDisplayNode] Handle placeholder layer creation in case the size of the node changes after the initial placeholder layer was created

maicki:
Should fix our placeholder example as well as #563

This fixes the placeholder issue again that was already fixed via #1707, but it was reverted with this PR: #1717 that actually didn't have to do something with this fix.


hannahmbanana:
@maicki: confirmed that this PR fixes Issue #563. Please close the issue if/when this merges.

Issue #1782, opened 2016-06-20
[ASVideoNode] Placeholder images are replaced by a blank placeholder


gazreese:
I need to use placeholder images in my project, so I've been using the HEAD code to get hold of the functionality. It appears that even though I'm setting the placeholder via the .URL property, this quickly gets overwritten with a generated placeholder. This generated placeholder takes just as long as the video to load however, and in the mean time has a blank image that overrides the supplied one in .URL. The result of this is that the user sees, in this order:

- The placeholder image
- A blank box
- The video-derived placeholder
- The video then starts to play

I'll attach a suggested change / pull request that seems to fix this behaviour, along with changes to the Video example app to add an HLS video and placeholder. It's difficult to show the problem in the example app, but if you swap out the HLS video URL for something higher bitrate it's easier to see.

There are a lot of HLS example URLs here:
http://stackoverflow.com/questions/10104301/hls-streaming-video-url-need-for-testing

Not a lot of these URLs work inside ASDK however, which may be indicative of other problems. I'm not sure. The links work fine in QuickTime on a Mac.

gazreese:
The first commit had some project file changes in it that seemed to be caused by a difference in Cocoapods versions. I've done a second branch and pull request https://github.com/facebook/AsyncDisplayKit/pull/1788 that covers the changes cleaned up.

There also appeared to be old versions of the ASVideoNode source files inside the Video example. Didn't look like they were supposed to be there but I've not removed them in this change just in case.

Adlai-Holler:
Closed via #1788.

PR #1788, opened 2016-06-21
[ASVideoNode] issue #1782 Placeholder images are replaced by blanks during loading

gazreese:
Now checks the .URL property of the parent class as well as the .image to ensure that new placeholders aren't generated.

Please see https://github.com/facebook/AsyncDisplayKit/issues/1782 for details.


hannahmbanana:
@Xinchi is going to try this out and let us know if it fixes his issue. Right? :)

hannahmbanana:
This failed because i was using the old .travis.yml / build.sh file - it's trying to build mode = examples, when now we have examples_pt1, etc. 

I reran the build, but I'm not sure if that will resolve it or if we need to do something else.

hannahmbanana:
@Xinchi: were you able to test this out?

hannahmbanana:
@gazreese: this build will always fail because it is using our old script. Could you rebase please?

gazreese:
@hannahmbanana  Looks good now, could you please take a look? Thanks!

Adlai-Holler:
@gazreese Another great change. I noticed that the image is being reset because when we first call `generatePlaceholderImage` the AVAsset has no tracks and so we call completion with `nil` and clear the image. This is certainly not an intended behavior.

I think the ideal behavior here is to clear self.image in -setAsset: or -clearFetchedData _if it was automatically set_ but otherwise the node itself should never `self.image = nil`. Thoughts @gazreese @appleguy ?

Thanks again for this and all your other well-made contributions lately. Landing this immediately!

PR #1790, opened 2016-06-21
[ASVideoNode] Ensure that both ASVideoNode and ASNetworkImageNode delegate methods are called

gazreese:
Before this change the user of ASVideoNode could set the .delegate property but the ASNetworkImageNode delegate methods were not called. This meant that the client of ASVideoNode was not able to observe image progress or errors.

I've tested this both with the test harness and in my own project.

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -453,7 +453,9 @@ static NSString * const kStatus = @"status";
 
 - (void)setDelegate:(id<ASVideoNodeDelegate>)delegate
 {
+  [super setDelegate:delegate];
   _delegate = delegate;
+  
   if (_delegate == nil) {
     memset(&_delegateFlags, 0, sizeof(_delegateFlags));
   } else {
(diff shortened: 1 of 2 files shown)


hannahmbanana:
@gazreese: Thanks for catching this - sorry I missed this!

hannahmbanana:
@Adlai-Holler @maicki: this PR LGTM. It's an easy one-liner (excluding testing code).

Adlai-Holler:
Let's take this to the hoop! 👍

gazreese:
Thanks guys!

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -453,7 +453,9 @@ - (AVPlayer *)player
 
 - (void)setDelegate:(id<ASVideoNodeDelegate>)delegate
 {
+  [super setDelegate:delegate];
```
appleguy: @gazreese ak, thanks for fixing this!!!

AsyncDisplayKitTests/ASVideoNodeTests.m
```diff
@@ -405,4 +413,18 @@ - (void)testClearingFetchedContentShouldClearAssetData
   XCTAssertNil(_videoNode.image);
 }
 
+- (void)testDelegateProperlySetForClassHierarchy
```
appleguy: Love love love that you added this test.  Thank you.

PR #1792, opened 2016-06-21
[ASMapNode] bug fixes + added the ability to zoom in on the annotations + example project

george-gw:
* Toggle user interaction when liveMap changes (#1753)
* Change map snapshot when updating it with a new region (#1754)
* Commented out code that is causing inaccurate behavior, this was wrapped with a FIXME
* Added the ability to zoom in and show annotations, similar to showAnnotations:animated: of MKMapView.
* Added a basic example for ASMapNode to try out the different changes

A minor change: Fixed typo in some comment

### DIFF
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -24,7 +24,6 @@
   MKMapSnapshotter *_snapshotter;
   BOOL _snapshotAfterLayout;
   NSArray *_annotations;
-//  CLLocationCoordinate2D _centerCoordinateOfMap;
 }
 @end
 
@@ -46,7 +45,6 @@
   
   _needsMapReloadOnBoundsChange = YES;
   _liveMap = NO;
-//  _centerCoordinateOfMap = kCLLocationCoordinate2DInvalid;
   _annotations = @[];
   _showAnnotationsOptions = ASMapNodeShowAnnotationsOptionsIgnored;
   return self;
@@ -161,18 +159,9 @@
 
 - (void)setRegion:(MKCoordinateRegion)region
 {
-  MKMapSnapshotOptions * __weak oldOptions = self.options;
-  MKMapSnapshotOptions * options = [[MKMapSnapshotOptions alloc] init];
-  options.camera = oldOptions.camera;
-  options.mapRect = oldOptions.mapRect;
-  options.mapType = oldOptions.mapType;
-  options.showsPointsOfInterest = oldOptions.showsPointsOfInterest;
-  options.showsBuildings = oldOptions.showsBuildings;
-  options.size = oldOptions.size;
-  options.scale = oldOptions.scale;
+  MKMapSnapshotOptions * options = [self.options copy];
   options.region = region;
   self.options = options;
-//  self.options.region = region;
 }
 
 #pragma mark - Snapshotter
@@ -282,18 +271,12 @@
       BOOL const animated = self.showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsAnimated;
       [_mapView showAnnotations:_mapView.annotations animated:animated];
     }
-    
-//    if (CLLocationCoordinate2DIsValid(_centerCoordinateOfMap)) {
-//      [_mapView setCenterCoordinate:_centerCoordinateOfMap];
-//    }
   }
 }
 
 - (void)removeLiveMap
 {
-  self.userInteractionEnabled = false;
-  // FIXME: With MKCoordinateRegion, isn't the center coordinate fully specified?  Do we need this?
-//  _centerCoordinateOfMap = _mapView.centerCoordinate;
+  self.userInteractionEnabled = NO;
   [_mapView removeFromSuperview];
   _mapView = nil;
 }
@@ -336,7 +319,7 @@
   CLLocationCoordinate2D topLeftCoord = CLLocationCoordinate2DMake(-90, 180);
   CLLocationCoordinate2D bottomRightCoord = CLLocationCoordinate2DMake(90, -180);
 
-  for (id<MKAnnotation> annotation in _annotations) {
+  for (id<MKAnnotation> annotation in annotations) {
     topLeftCoord = CLLocationCoordinate2DMake(fmax(topLeftCoord.latitude, annotation.coordinate.latitude),
                                               fmin(topLeftCoord.longitude, annotation.coordinate.longitude));
     bottomRightCoord = CLLocationCoordinate2DMake(fmin(bottomRightCoord.latitude, annotation.coordinate.latitude),
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -33,6 +33,7 @@
 @synthesize mapDelegate = _mapDelegate;
 @synthesize options = _options;
 @synthesize liveMap = _liveMap;
+@synthesize showAnnotationsOptions = _showAnnotationsOptions;
 
 #pragma mark - Lifecycle
 - (instancetype)init
@@ -267,8 +268,9 @@
     [weakSelf setNeedsLayout];
     [weakSelf.view addSubview:_mapView];
 
-    if (self.showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsZoomed) {
-      BOOL const animated = self.showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsAnimated;
+    ASMapNodeShowAnnotationsOptions showAnnotationsOptions = self.showAnnotationsOptions;
+    if (showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsZoomed) {
+      BOOL const animated = showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsAnimated;
       [_mapView showAnnotations:_mapView.annotations animated:animated];
     }
   }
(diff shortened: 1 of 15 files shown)


facebook-github-bot-2:

facebook-github-bot-7:

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -161,7 +161,18 @@ - (MKCoordinateRegion)region
...
+  options.camera = oldOptions.camera;
+  options.mapRect = oldOptions.mapRect;
+  options.mapType = oldOptions.mapType;
+  options.showsPointsOfInterest = oldOptions.showsPointsOfInterest;
+  options.showsBuildings = oldOptions.showsBuildings;
+  options.size = oldOptions.size;
+  options.scale = oldOptions.scale;
+  options.region = region;
+  self.options = options;
+//  self.options.region = region;
```
Adlai-Holler: What's the motivation for this change?

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -265,17 +277,23 @@ - (void)addLiveMap
...
+    
+//    if (CLLocationCoordinate2DIsValid(_centerCoordinateOfMap)) {
+//      [_mapView setCenterCoordinate:_centerCoordinateOfMap];
+//    }
   }
 }
 
 - (void)removeLiveMap
 {
+  self.userInteractionEnabled = false;
```
Adlai-Holler: We may not want to make this change. Since ASMapNode inherits from ASControlNode, the user may very well have attached handlers for control events e.g. tap on the map to open a larger map. Let's instead have ASMapNode default to `userInteractionEnabled = YES` (ASControlNode defaults to NO which is something we may want to change in the future) and let the user change it from there if they want.

AsyncDisplayKit/ASMapNode.mm
Adlai-Holler: Oh I see, it's to trigger a new snapshot. Hmm. Let's use `[self.options copy]` instead.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -295,11 +313,44 @@ - (void)setAnnotations:(NSArray *)annotations
...
 
+-(MKCoordinateRegion)regionToFitAnnotations:(NSArray<id<MKAnnotation>> *)annotations
+{
+  if([annotations count] == 0)
+    return MKCoordinateRegionForMapRect(MKMapRectWorld);
+
+  CLLocationCoordinate2D topLeftCoord = CLLocationCoordinate2DMake(-90, 180);
+  CLLocationCoordinate2D bottomRightCoord = CLLocationCoordinate2DMake(90, -180);
+
+  for (id<MKAnnotation> annotation in _annotations) {
```
Adlai-Holler: You probably meant to use the parameter here, not the ivar.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -24,7 +24,7 @@ @interface ASMapNode()
   MKMapSnapshotter *_snapshotter;
   BOOL _snapshotAfterLayout;
   NSArray *_annotations;
-  CLLocationCoordinate2D _centerCoordinateOfMap;
+//  CLLocationCoordinate2D _centerCoordinateOfMap;
```
appleguy: If this is not needed anymore, could you delete it instead of leaving it commented out?  Same below.

AsyncDisplayKit/ASMapNode.mm
appleguy: @george-gw yes, @Adlai-Holler is right - we should copy the options, then just write to the region value, and set it back to self.options.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -265,17 +277,23 @@ - (void)addLiveMap
...
-    
-    if (CLLocationCoordinate2DIsValid(_centerCoordinateOfMap)) {
-      [_mapView setCenterCoordinate:_centerCoordinateOfMap];
+
+    if (self.showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsZoomed) {
+      BOOL const animated = self.showAnnotationsOptions & ASMapNodeShowAnnotationsOptionsAnimated;
+      [_mapView showAnnotations:_mapView.annotations animated:animated];
     }
+    
+//    if (CLLocationCoordinate2DIsValid(_centerCoordinateOfMap)) {
```
appleguy: Why is this commented out?  Can you explain how you know this is no longer needed, and why it would be kept in commented form?

AsyncDisplayKit/ASMapNode.mm
appleguy: Yeah, that seems fine with me.  We can default to setting it to YES in init.

@george-gw it's important to note that false is not supposed to be used with Objective-C's BOOLs.  Only YES / NO.

appleguy:
Looks like a number of revisions are required here.  Thanks for contributing, @george-gw!  This is an important component, so we should get at least two ASDK accepts to land.

AsyncDisplayKit/ASMapNode.mm
george-gw: @appleguy @Adlai-Holler I had written it with `[self.options copy]` before, but it didn't work for some reason, that's why I did what I did there. But it seems I must have done something wrong somewhere else because it seems to work fine with `[self.options copy]` now.

Thanks!

AsyncDisplayKit/ASMapNode.mm
george-gw: @appleguy You're right, sorry, I've been working with swift a lot lately :)

@Adlai-Holler @appleguy how about if I hold the previous state in some temporary variable and then restore it afterwards?
Because in the state it is now, we are changing the state of self.userInteractionEnabled. The reason why I placed that in removeLiveMap was to go back to its original state and thus not create unexpected behavior.

Imagine this scenario:
1- Create ASMapNode (`userInteractionEnabled = YES`, based on what the discussion is now)
2- Set `mapNode.userInteractionEnabled = NO` for whatever reason the user decides to do that
3- Toggle `mapNode.liveMap = YES` (this triggers `self.userInteractionEnabled = YES`)
4- Done with live map, go back to image. `mapNode.liveMap = NO` (current state is: `userInteractionEnabled = YES`)

In this case we changed the state of `userInteractionEnabled`

AsyncDisplayKit/ASMapNode.mm
george-gw: They're technically the same, but you're right, thanks for noticing :)

AsyncDisplayKit/ASMapNode.mm
george-gw: That particular line is not needed because of line 267 `[weakSelf applySnapshotOptions];` in that method the region is set.

There is no reason to keep it commented out, I will delete it.

AsyncDisplayKit/ASMapNode.mm
george-gw: Will do

george-gw:
@appleguy I fixed everything except setting userInteractionEnabled to YES by default because of the pending discussion, I hope that's ok!

Thank you for the revisions and this library!

AsyncDisplayKit/ASMapNode.mm
Adlai-Holler: @george-gw I believe the simplest thing here, is never to change `self.userInteractionEnabled` except in `-init`. Saving and restoring the previous state would interfere with other use cases and might surprise developers. If we set it to `YES` by default, and then give developers complete control after that, we should get a good balance of good out-of-the-box behavior and flexibility.

george-gw: @Adlai-Holler Ah! That makes perfect sense!
It's done.

hannahmbanana:
@Adlai-Holler @maicki - time for a 2nd review on this.

Adlai-Holler:
Would you be willing to protect the new `showAnnotationsOptions` flag with a lock? I doubt it will be contended often, but better safe than sorry. And then once it's protected, ensure that the property is only read once during `setAnnotations:`.

Once that is in, 👍 from me!

hannahmbanana:
@george-gw

george-gw:
@Adlai-Holler: Wouldn't it cause a problem in `addLiveMap` if I protect it with a mutex, since it requires to be called on the main thread?

george-gw:
That's odd! Why would the test fail based on the new commit?
I tried running the test locally, it worked fine for me.

Adlai-Holler:
@george-gw It should be OK because it's acceptable for the main thread to wait on background threads.

Strange that an arbitrary ASDisplayLayer test failed. I've restarted the tests – let's see if they happen again.

@appleguy @hannahmbanana Maybe we have a fragile test? Here's the failure:

```
_ASDisplayLayerTests
  testDelegateDrawInContextAsync, ((asyncDelegate.drawRectCount) equal to (0u)) failed: ("1") is not equal to ("0")
  /Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKitTests/ASDisplayLayerTests.m:361
    XCTAssertEqual(layer.drawInContextCount, 0u);
    XCTAssertEqual(asyncDelegate.drawRectCount, 0u);
    dispatch_resume([_ASDisplayLayer displayQueue]);
	 Executed 308 tests, with 1 failure (0 unexpected) in 26.641 (27.338) seconds
```

george-gw:
@Adlai-Holler Thanks for restarting the test, seems good now!

Adlai-Holler:
Thanks for following through on this one @george-gw ! Glad to have this in.

george-gw:
I have enjoyed the collaboration, thank you for your input @Adlai-Holler :)

PR #1794, opened 2016-06-21
[ASDataController] Remove asyncDataFetching Option, Cleanup

Adlai-Holler:
I was doing a deep-dive into this code in preparation for some work, and I saw a bunch of possible improvements. So here they are.

- Remove asyncDataFetching logic, deprecating the associated methods. We no longer call the locking methods because we now follow the UIKit synchronization paradigm – submit your edits on the main queue, and we'll query you back immediately on the main queue.
- Add assertions to data controller about which queue we're on
- Make data controller use _editingTransactionQueue even when doing synchronous reload (so that those assertions pass).
- Other small tweaks/optimizations (avoiding needless allocations/method calls)

I've tested this thoroughly by hand.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -461,7 +461,7 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
       NSUInteger editingNodesSectionCount = editingNodes.count;
       
       if (editingNodesSectionCount) {
-        NSMutableIndexSet *indexSet = [[NSMutableIndexSet alloc] initWithIndexesInRange:NSMakeRange(0, editingNodesSectionCount)];
+        NSIndexSet *indexSet = [[NSIndexSet alloc] initWithIndexesInRange:NSMakeRange(0, editingNodesSectionCount)];
         [self _deleteNodesAtIndexPaths:ASIndexPathsForTwoDimensionalArray(editingNodes) withAnimationOptions:animationOptions];
         [self _deleteSectionsAtIndexSet:indexSet withAnimationOptions:animationOptions];
       }
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -457,7 +457,7 @@ static void *kASSizingQueueContext = &kASSizingQueueContext;
       // Allow subclasses to perform setup before going into the edit transaction
       [self prepareForReloadData];
       
-      void (^transactionBlock)() = ^{
+      [_editingTransactionQueue addOperationWithBlock:^{
         LOG(@"Edit Transaction - reloadData");
         
         // Remove everything that existed before the reload, now that we're ready to insert replacements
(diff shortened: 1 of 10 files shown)


Adlai-Holler:
@maicki @appleguy Review?

Adlai-Holler:
Those last two commits were going a little crazy, but here's the evidence for my comment about [NSIndexPath indexPathForItem:inSection:]:

screen shot 2016-06-21 at 3 22 41 pm

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -143,15 +142,17 @@ - (void)layoutLoadedNodes:(NSArray<ASCellNode *> *)nodes fromContexts:(NSArray<A
  */
 - (void)_layoutNode:(ASCellNode *)node withConstrainedSize:(ASSizeRange)constrainedSize
 {
-  [node measureWithSizeRange:constrainedSize];
-  node.frame = CGRectMake(0.0f, 0.0f, node.calculatedSize.width, node.calculatedSize.height);
+  CGSize size = [node measureWithSizeRange:constrainedSize].size;
+  node.frame = { .size = size };
```
appleguy: Hmm, is this guaranteed to properly initialize the .origin values?  Or are they left uninitialized?

AsyncDisplayKit/Private/ASMultidimensionalArrayUtils.mm
```diff
@@ -72,7 +78,12 @@ static BOOL ASElementExistsAtIndexPathForMultidimensionalArray(NSArray *array, N
   NSUInteger indexesLength = indexLength - 1;
   NSUInteger indexes[indexesLength];
   [indexPath getIndexes:indexes range:NSMakeRange(1, indexesLength)];
-  NSIndexPath *indexPathByRemovingFirstIndex = [NSIndexPath indexPathWithIndexes:indexes length:indexesLength];
+  NSIndexPath *indexPathByRemovingFirstIndex;
+  if (indexesLength == 2) {
+    indexPathByRemovingFirstIndex = [NSIndexPath indexPathForItem:indexes[1] inSection:indexes[0]];
```
appleguy: I am not sure I see the value in this particular change. I believe this method may well be slower than the alternative, and it is more complex to read. It is possible that it is faster, but if this is the case, we should confirm with profiling before adding the complexity.

appleguy:
@Adlai-Holler thank you for including the categories and deprecation flags to maintain compatibility. I reviewed this in quite heavy detail, but I am also partially trusting that there are not material changes subtly within some of the code blocks that were removed from the access data source with block sections.

appleguy:
@Adlai-Holler the detail  is quite interesting on using Singleton objects for index paths only when initialized with the to argument method, thanks for including that.

Can you please add a comment where are you added the fork code path, because it is far from obvious why it would be worth the complexity to do so?

AsyncDisplayKit/Details/ASDataController.mm
Adlai-Holler: Yep! C++ will initialize all struct members to 0 by default so `{} => CGRectZero` etc.

Issue #1780, opened 2016-06-20
PinRemoteImage only supported via CocoaPods


nickvelloff:
Issue 1:
ASDK uses preprocessor macros to declare `PIN_REMOTE_IMAGE` that are defined in the Podspec, thus only available to CocoaPods build.

Issue 2:
`PinRemoteImage` must be forked and example projects removed to actually build cleanly. Really old issue for this repo, here: https://github.com/pinterest/PINRemoteImage/issues/25#issuecomment-227062812

Basically every attempt to resolve this leads to a fork. 

@hannahmbanana Thanks!

nickvelloff:
BTW it's perfectly reasonable to say "only supported via CocoaPods"... 
Personally I was just trying to avoid CocoaPods, but rather that than maintain a fork on a rapidly changing framework.

nickvelloff:
Example using manually added dynamic framework:
https://github.com/nickvelloff/ASDK-DynamicLib.git

Example using Carthage build, with PinRemoteImage fork to resolve build issue:
https://github.com/nickvelloff/ASDK-DynamicLib/commit/83b5e4c81411493e05d8023bd710cb9661bd4765

hannahmbanana:
Similar issue at #1464.

hannahmbanana:
Update:  With the addition of the cartfile + work around for the PIN_REMOTE_IMAGE flag for non-Cocoapod users, **AsyncDisplayKit now support the standard way to use Carthage**:

_The standard way to use Carthage is by having a Cartfile listing the dependencies, and then running carthage update to download them in the Cathage/Checkouts folder, and build each of those into frameworks located in the Carthage/Build folder, and finally the developer has to manually integrate in the project._

AsyncDisplayKit does not yet support the lighter way of using Carthage, in which you manually add the project files.
http://www.mokacoding.com/blog/carthage-no-build/

maicki:
![happy](https://cloud.githubusercontent.com/assets/25288/16288883/288fc22e-38a5-11e6-821d-49b0dcce8794.gif)

nickvelloff:
@hannahmbanana woohoo! 🚀 🎸 👍

hannahmbanana:
...but I wasn't done? Still not ready to give up on the "lighter Carthage method" :)

PR #1797, opened 2016-06-22
[Carthage] define PIN_REMOTE_IMAGE in .pch for non-CocoaPod users

hannahmbanana:
Resolves Issue 1 of #1780. 

Users who don't use CocoaPods will have PIN_REMOTE_IMAGE defined in .pch file.

### DIFF
--- a/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
+++ b/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
@@ -85,7 +85,23 @@
   static PINRemoteImageManager *sharedPINRemoteImageManager = nil;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
+  
 #if PIN_ANIMATED_AVAILABLE
+    // Check that Carthage users have linked both PINRemoteImage & PINCache by testing for one file each
+    if (!(NSClassFromString(@"PINRemoteImageManager.m"))) {
+        NSException *e = [NSException
+                          exceptionWithName:@"FrameworkSetupException"
+                          reason:@"Missing the path to the PINRemoteImage framework."
+                          userInfo:nil];
+        @throw e;
+    }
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -85,7 +85,23 @@ - (PINRemoteImageManager *)sharedPINRemoteImageManager
   static PINRemoteImageManager *sharedPINRemoteImageManager = nil;
   static dispatch_once_t onceToken;
   dispatch_once(&onceToken, ^{
+  
 #if PIN_ANIMATED_AVAILABLE
+    // Check that Carthage users have linked both PINRemoteImage & PINCache by testing for one file each
+    if (!(NSClassFromString(@"PINRemoteImageManager.m"))) {
```
maicki: You should not to add an `.m` in this case ... just use `NSClassFromStirng(@"PINRemoteImageManader")`

AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -85,7 +85,23 @@ - (PINRemoteImageManager *)sharedPINRemoteImageManager
...
+        NSException *e = [NSException
+                          exceptionWithName:@"FrameworkSetupException"
+                          reason:@"Missing the path to the PINRemoteImage framework."
+                          userInfo:nil];
+        @throw e;
+    }
+    if (!(NSClassFromString(@"PINCache.m"))) {
```
maicki: I guess same for this case as above, you should remove the `.m` at the end of the class

hannahmbanana:
Thanks @nickvelloff for testing this out!

nickvelloff:
@hannahmbanana You bet! Thanks for the solution!

PR #1802, opened 2016-06-23
[Unit Tests] Add Table View Data Thrash Testing

Adlai-Holler:
Good news and bad news, folks! Good news is we now have a thrash tester to tell us if we have data integrity problems in table view. Bad news is, we have data integrity problems in table view. 

### What I Did

- You can switch between UIKit/ASDK so we can shift blame as needed.
- Base test "testThrashingWildly" runs 100 iterations of random updates to a 20x20 table view.
- When a thrash test fails, it logs what it was attempting to do and a Base64 representation of the failed update. 
- You can copy this Base64 into the "ASThrashTestRecordedCase" file and the "testRecordedThrashCase" method will load it and try again now that you've fixed the issue.

### What It Did

UIKit reliably passes the thrash testing. This is actually good! Because it means that the updates I'm generating are consistently valid. ASDK reliably doesn't pass the thrash testing – it submits invalid updates up to UITableView and triggers assertion failures. This isn't good. I'll find out why shortly.

### What I Will Do

Next step is to migrate to collection view and add supplementary views. That probably won't happen until after this basic hurdle is passed and other stuff is also done too.

@maicki @nguyenhuy @appleguy @levi


hannahmbanana:
Super cool! The thrash testing part...not the data integrity part :)

Adlai-Holler:
I got eaten by a grue! Thanks a lot, Travis.

```
xcodebuild: error: Scheme PINRemoteImage is not currently configured for the build action.
Oh no! You walked directly into the slavering fangs of a lurking grue!
```

Looks like this is happening on #1803 as well. Carthaaaaaaaaaaaaaage! @hannahmbanana

hannahmbanana:
Yikes. This is definitely my fault. :( I'll restart the builds once I figure it out.

PR #1803, opened 2016-06-23
[Example] Resolved ASTextNode bugs (test app to reproduce issues, do regression testing)

hannahmbanana:


examples_extra/TextStressTest/Sample/ViewController.m
```diff
@@ -0,0 +1,214 @@
...
+    case 1:                             // default truncation attributed string color shouldn't match attributed text color (ASDK) vs. match (UIKit)
+      node.maximumNumberOfLines = 3;
+      [string addAttribute:NSForegroundColorAttributeName
+                     value:[UIColor redColor]
+                     range:NSMakeRange(0, [string length])];
+      node.attributedText = string;
+      return node;
+      
+    case 2:
+      node.maximumNumberOfLines = 3;
```
appleguy: Can you remove these additional cases for now?  They're easy to add back in as we implement more.

examples_extra/TextStressTest/Sample/ViewController.m
```diff
@@ -0,0 +1,214 @@
...
+
+
+@implementation ViewController
+
+- (instancetype)init
+{
+  if (!(self = [super init]))
+    return nil;
+  
+  _scrollView = [[UIScrollView alloc] init];
```
appleguy: Move all three of these into -viewDidLoad, because _scrollView being created there would be better anyway, and the arrays aren't needed until that point.  Technically we could just delete -init entirely then.

examples_extra/TextStressTest/Sample/ViewController.m
```diff
@@ -0,0 +1,214 @@
...
+{
+  [super viewWillLayoutSubviews];
+  
+  CGFloat maxWidth = 0;
+  CGFloat maxHeight = 0;
+  
+  CGRect frame = CGRectMake(50, 50, 0, 0);
+  
+  for (int i = 0; i < NUMBER_ELEMENTS; i++) {
+
```
appleguy: Delete extra return at start

examples_extra/TextStressTest/Sample/ViewController.m
```diff
@@ -0,0 +1,214 @@
...
+      
+    case 1:
+      label.numberOfLines = 3;
+      [string addAttribute:NSForegroundColorAttributeName
+                     value:[UIColor redColor]
+                     range:NSMakeRange(0, [string length])];
+      label.attributedText = string;
+      return label;
+      
+    case 2:
```
appleguy: Also delete the extra case statements from these next two methods

appleguy:
@hannahmbanana Thanks!  This is pretty clean.  Certainly more we can do to upgrade it over time, but I'm glad we have this project in as a base to continue improving test coverage (and even during-development example app suites) for ASTextNode.  Later, we should add examples from multiple writing scripts, etc.

PR #1805, opened 2016-06-23
[ASLayoutable] Add Convenience Methods to Enable Chaining

Adlai-Holler:
This is mostly a PoC/proposal to see what people think about this. The goal is, for the most part, to be able to specify a complex layout spec in one gigantic declarative-style call and avoid intermediate variables and order-inversion while building it. I think it's an improvement but still sort of daunting.

I don't expect to merge this necessarily. It didn't take long to create so I'm not married to it. But it proves that something like this is possible and I think beneficial.

- Add methods like `-(instancetype)withFlexGrow:(BOOL)flexGrow` to ASLayoutable to set properties while chaining
- Add methods like `-(ASInsetLayoutSpec *)withInset:(UIEdgeInsets)insets` to ASLayoutable to embed layoutables while chaining
- Implementations for these methods are stored in a ASLayoutableConvenienceImpl macro, which is added to ASDisplayNode.mm and ASLayoutSpec.m
- All indentation is Xcode-correct. There are questions about where newlines should go but I think ASDKgram::PhotoCellNode has a pretty nice style.
- Update a couple sample app layout spec methods as demo.

@appleguy @maicki @levi @hannahmbanana Thoughts?


appleguy:
@Adlai-Holler this is an interesting approach...perhaps best reviewed in person with a team meeting?

Adlai-Holler:
@appleguy I agree. Maybe during today's sync-up

eanagel:
@Adlai-Holler I sure would like something to relive me from the temporary variables you end up when creating layouts and I do like this approach. The square braces get out of hand with fluent-style interfaces in ObjC, though. There is a hack I've seen using properties that return blocks to handle this. Take a look at this project and see if it inspires you: https://github.com/VerbalExpressions/ObjectiveCVerbalExpressions

Adlai-Holler:
Nobody has time to really look at this right now. We might revisit it after 2.0.

PR #1806, opened 2016-06-23
[ASButtonNode] Add support for 'preferredFrameSize' to ASButtonNode

maicki:


maicki:
@Adlai-Holler Can I have some eyes. Small change to support `preferredFrameSize` in `ASButtonNode`

Adlai-Holler:
@maicki Don't we want the inset to be applied before the static size? So that the final size matches the preferred size?

PR #1808, opened 2016-06-24
[ASNetworkImage] Use flags to cache instead of instance variables for caching respond to selector calls

maicki:


maicki:
@hannahmbanana @Adlai-Holler @rcancro Could I get some eyes on that please? Thanks!

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -76,17 +80,17 @@ - (instancetype)initWithCache:(id<ASImageCacheProtocol>)cache downloader:(id<ASI
   
   ASDisplayNodeAssert([downloader respondsToSelector:@selector(downloadImageWithURL:callbackQueue:downloadProgress:completion:)] || [downloader respondsToSelector:@selector(downloadImageWithURL:callbackQueue:downloadProgressBlock:completion:)], @"downloader must respond to either downloadImageWithURL:callbackQueue:downloadProgress:completion: or downloadImageWithURL:callbackQueue:downloadProgressBlock:completion:.");
   
-  _downloaderSupportsNewProtocol = [downloader respondsToSelector:@selector(downloadImageWithURL:callbackQueue:downloadProgress:completion:)];
+  _downloaderFlags.downloaderSupportsNewProtocol = [downloader respondsToSelector:@selector(downloadImageWithURL:callbackQueue:downloadProgress:completion:)];
```
rcancro: on 64-bit iOS BOOL is a C bool which defined as a int8_t. But 32-bit BOOL is an unsigned char. This seems fine to me, but are there any casting concerns from unsigned char to unsigned int? I suppose not ...

rcancro:
LGTM!

AsyncDisplayKit/ASNetworkImageNode.mm
appleguy: @rcancro this is a good question. In this case, we are allocating only a single bit to the unsigned int, which is a bit unusual and I believe will be the same size across architectures. We do however use this pattern in several other places in the codebase, and so it has proven to work in practice, although I think it is somewhat questionable to do the assignment without an explicit cast…

Adlai-Holler:
Dat efficiency! This flags pattern is pretty common (used in UIKit as well) so let's roll with it.

PR #1811, opened 2016-06-24
[AsyncDisplayKit+Utilities.h] Convenience methods for creating resizable circles & rounded rects.

hannahmbanana:
I named this generically to match AsyncDisplayKit+Debug.h

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -314,10 +314,10 @@
 		7AB338661C55B3420055FDE8 /* ASRelativeLayoutSpec.mm in Sources */ = {isa = PBXBuildFile; fileRef = 7A06A7381C35F08800FE8DAA /* ASRelativeLayoutSpec.mm */; };
 		7AB338671C55B3460055FDE8 /* ASRelativeLayoutSpec.h in Headers */ = {isa = PBXBuildFile; fileRef = 7A06A7391C35F08800FE8DAA /* ASRelativeLayoutSpec.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		7AB338691C55B97B0055FDE8 /* ASRelativeLayoutSpecSnapshotTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 7AB338681C55B97B0055FDE8 /* ASRelativeLayoutSpecSnapshotTests.mm */; };
-		8021EC161D2A27C600799119 /* AsyncDisplayKit+Utilities.h in Headers */ = {isa = PBXBuildFile; fileRef = 8021EC141D2A27C600799119 /* AsyncDisplayKit+Utilities.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		8021EC171D2A27C600799119 /* AsyncDisplayKit+Utilities.h in Headers */ = {isa = PBXBuildFile; fileRef = 8021EC141D2A27C600799119 /* AsyncDisplayKit+Utilities.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		8021EC181D2A27C600799119 /* AsyncDisplayKit+Utilities.m in Sources */ = {isa = PBXBuildFile; fileRef = 8021EC151D2A27C600799119 /* AsyncDisplayKit+Utilities.m */; };
-		8021EC191D2A27C600799119 /* AsyncDisplayKit+Utilities.m in Sources */ = {isa = PBXBuildFile; fileRef = 8021EC151D2A27C600799119 /* AsyncDisplayKit+Utilities.m */; };
+		8021EC161D2A27C600799119 /* UIImage+ASConvenience.h in Headers */ = {isa = PBXBuildFile; fileRef = 8021EC141D2A27C600799119 /* UIImage+ASConvenience.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		8021EC171D2A27C600799119 /* UIImage+ASConvenience.h in Headers */ = {isa = PBXBuildFile; fileRef = 8021EC141D2A27C600799119 /* UIImage+ASConvenience.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		8021EC181D2A27C600799119 /* UIImage+ASConvenience.m in Sources */ = {isa = PBXBuildFile; fileRef = 8021EC151D2A27C600799119 /* UIImage+ASConvenience.m */; };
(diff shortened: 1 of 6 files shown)


hannahmbanana:
@maicki: @levi: some really easy API questions:

- I think we should add border width as an argument. Is that okay?
- I think we should remove the as_* prefix (looks weird) and at the next ASDK sync, point to the point to the ASDK versions

maicki:
* It's gonna be a lot of parameters to this method, but I agree that a border width would be handy. What about creating an API where we pass in an object that has properties with default values that you can change before passing in? @levi  thoughts
* I don't think we should remove the `as_` prefix. This is common practice and prevents us from clashing with Apple API's.

AsyncDisplayKit/AsyncDisplayKit+Utilities.m
```diff
@@ -0,0 +1,107 @@
...
+                                   backgroundColor:(UIColor *)backgroundColor
+                                         fillColor:(UIColor *)fillColor
+                                       borderColor:(UIColor *)borderColor
+                                       borderWidth:(CGFloat)borderWidth
+                                     roundedCorner:(UIRectCorner)roundedCorners
+{
+    static NSCache *__pathCache = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        __pathCache = [[NSCache alloc] init];
```
maicki: Should we define some limit: `countLimit` or `totalCostLimit`

AsyncDisplayKit/AsyncDisplayKit+Utilities.m
```diff
@@ -0,0 +1,107 @@
...
+//  LICENSE file in the root directory of this source tree. An additional grant
+//  of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import "AsyncDisplayKit+Utilities.h"
+
+@implementation UIImage (ASDKAdditions)
+
++ (UIImage *)as_resizableRoundedImageWithCornerRadius:(CGFloat)cornerRadius
+                                             scale:(CGFloat)scale
```
appleguy: Fix indentation

AsyncDisplayKit/AsyncDisplayKit+Utilities.m
```diff
@@ -0,0 +1,107 @@
...
+                                   backgroundColor:(UIColor *)backgroundColor
+                                         fillColor:(UIColor *)fillColor
+                                       borderColor:(UIColor *)borderColor
+                                       borderWidth:(CGFloat)borderWidth
+{
+    return [[self class] as_resizableRoundedImageWithCornerRadius:cornerRadius scale:scale backgroundColor:backgroundColor fillColor:fillColor borderColor:borderColor borderWidth:borderWidth roundedCorner:UIRectCornerAllCorners];
```
appleguy: You can just use "self" here - because in a class method, self is the class

AsyncDisplayKit/AsyncDisplayKit+Utilities.h
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Created by Hannah Troisi on 6/24/16.
+//
+//  Copyright (c) 2014-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under the BSD-style license found in the
+//  LICENSE file in the root directory of this source tree. An additional grant
+//  of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import <UIKit/UIKit.h>
```
appleguy: Let's make this just /UIImage.h

appleguy:
@hannahmbanana Thanks for working on this!  Could you add 1 or 2 shorter variants with fewer options, which just internally call the longer one, so they are more convenient to use?

I think we should just keep the explicit arguments because it's a bit less convenient to initialize a separate object to use as the options, even though it simplifies the API surface.

AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1874,12 +1884,12 @@
...
 			buildPhases = (
-				2E61B6A0DB0F436A9DDBE86F /* 📦 Check Pods Manifest.lock */,
+				2E61B6A0DB0F436A9DDBE86F /* [CP] Check Pods Manifest.lock */,
 				058D09B8195D04C000B7D73C /* Sources */,
 				058D09B9195D04C000B7D73C /* Frameworks */,
 				058D09BA195D04C000B7D73C /* Resources */,
-				3B9D88CDF51B429C8409E4B6 /* 📦 Copy Pods Resources */,
-				B130AB1AC0A1E5162E211C19 /* 📦 Embed Pods Frameworks */,
+				3B9D88CDF51B429C8409E4B6 /* [CP] Copy Pods Resources */,
+				B130AB1AC0A1E5162E211C19 /* [CP] Embed Pods Frameworks */,
```
maicki: Do we need this changes here? Which Cocoapods version are you using, I think with 1.0 they changed that ...

AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1922,6 +1932,7 @@
 				TargetAttributes = {
 					057D02BE1AC0A66700C7AC3C = {
 						CreatedOnToolsVersion = 6.2;
+						DevelopmentTeam = 4QK39LR37N;
```
maicki: Same for here should define the dev team?

AsyncDisplayKit/AsyncDisplayKit.h
```diff
@@ -75,15 +75,17 @@
...
-#import <AsyncDisplayKit/CGRect+ASConvenience.h>
-#import <AsyncDisplayKit/NSMutableAttributedString+TextKitAdditions.h>
-#import <AsyncDisplayKit/UICollectionViewLayout+ASConvenience.h>
-#import <AsyncDisplayKit/UIView+ASConvenience.h>
 #import <AsyncDisplayKit/ASRunLoopQueue.h>
 #import <AsyncDisplayKit/ASTextKitComponents.h>
 #import <AsyncDisplayKit/ASTraitCollection.h>
 #import <AsyncDisplayKit/ASVisibilityProtocols.h>
 
+#import <AsyncDisplayKit/CGRect+ASConvenience.h>
```
appleguy: Seems nice to have all the categories next to each other, cool.

appleguy:
@hannahmbanana looks great!  Excited to have this in.  It's perfect for folks to use in -placeholderImage, -defaultImage, etc.

PR #1812, opened 2016-06-24
[ASDataController] Improve Update Handling, Esp. Reloading Sections

Adlai-Holler:
This patch gets our thrash tests passing! 🎉 

- Split section reloads into delete and insert.
- Improve the logic for computing final indexes from initial indexes.
- Improve logging/loggability of update events.
- Add more type annotations.

#### Reloads and Moves

I removed `ASDataController`s implementation of reloadSections/reloadItems. Reloads and moves are "composite" operations – they require knowledge of the entire batch update so they can be decomposed into deletes (at old indexes) and inserts (at new indexes). ASDataController was intentionally set up to receive updates in "safe order" so that it can modify its internal arrays bit-by-bit during updates. There is no such "safe order" for moves and reloads as such. The responsibility falls for `ASChangeSetDataController` to decompose all updates into deletes & inserts.

The logic for moves is not touched by this diff, but there is currently an issue where moves may lead to incorrect behavior when combined with other operations in a batch. Moves are tricky business because you have to perform the deletes stashing the nodes away, and then use those nodes when you perform the inserts later. ASChangeSetDataController could provide its superclass with a structure specifying which nodes should be used when inserting which sections/items, at the beginning of the batch update.

There are currently ASDataController-called subclass override methods like `willReloadSections:`. These are used by ASCollectionDataController to maintain supplementary items.
- Since reloads are necessarily split into two operations which will typically not be performed back-to-back, this kind of callback doesn't totally make sense.
- Since the operations we need to perform on supplementary items are basically direct mirrors of the operations we perform for cell items, this pattern is a little wonky. The ideal solution would be to treat supplementary items and cells the same way. This is how UICollectionView works but it will require pretty major refactoring.
- For this PR I've removed `willReloadSections` and `willReloadItems`. The combination of `willDelete` and `willInsert`, called at the appropriate times, achieves the same result or better.
- Something similar will probably have to be done for moves, if we choose not to refactor the supplementary item handling first.

There's obviously a lot more to be done here – unifying cells and supplementary views, handling moves – but I'm very confident this is an extremely worthwhile step forward. @appleguy @maicki @levi Could I get a review on this?


maicki:
LGTM

PR #1813, opened 2016-06-24
[ASTextNode] Further locking improvements for ASTextNode

maicki:


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -394,26 +400,23 @@ - (void)setAttributedText:(NSAttributedString *)attributedText
 
   // Tell the display node superclasses that the cached layout is incorrect now
   [self invalidateCalculatedLayout];
-
```
appleguy: I wonder if we need to include the renderer invalidation and the setNeedsDisplay in the lock context?  If we don't, what happens if another thread writes to other properties in the meantime?

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -394,26 +400,23 @@ - (void)setAttributedText:(NSAttributedString *)attributedText
...
-  if (ASObjectIsEqual(exclusionPaths, _exclusionPaths)) {
-    return;
+    if (ASObjectIsEqual(exclusionPaths, _exclusionPaths)) {
+      return;
+    }
+  
+    _exclusionPaths = [exclusionPaths copy];
   }
   
-  _exclusionPaths = [exclusionPaths copy];
```
appleguy: Same implication here regarding renderer and display invalidation, these operations may need to occur together as if they were a "transaction".

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -713,6 +713,8 @@ - (void)_setHighlightRange:(NSRange)highlightRange forAttributeName:(NSString *)
       }
 
       if (highlightTargetLayer != nil) {
+        std::lock_guard<std::recursive_mutex> l(_textLock);
```
appleguy: So the lock here was missing?

maicki: Yes for the renderer

appleguy:
@maicki This looks great to me!

PR #1814, opened 2016-06-25
[ASViewController] Fix Modal presented ASViewController don't rotate on iOS 8.3

### LINKED ISSUE #1810: [ASViewController] Modal presented ASViewController don't rotate on iOS 8.3
Hi all!
Minimum supported ios version in my project is 8.3 and I use a lot of AsyncDisplayKit features in it. 
But some time ago I faced this strange problem: modal presented ASViewController don't rotate after changing interface orientation. In version > 9.0 it works perfectly, but not in 8.3.

What I tried (and it doesn't help): 

```objectiveC
- (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation
{
    return YES;
}

- (UIInterfaceOrientationMask)supportedInterfaceOrientations
{
    return UIInterfaceOrientationMaskAll;
}
```

Screenshots:
img_0049
img_00491

Sample project:
[ModalASViewController.zip](https://github.com/facebook/AsyncDisplayKit/files/332274/ModalASViewController.zip)

maicki:
Should fix #1810

### DIFF
--- a/AsyncDisplayKit/ASViewController.mm
+++ b/AsyncDisplayKit/ASViewController.mm
@@ -203,7 +203,14 @@ ASVisibilityDepthImplementation;
 {
   // In modal presentation the view does not have the right bounds in iOS7 and iOS8. As workaround using the superviews
   // view bounds
-  CGSize viewSize = (self.presentingViewController != nil) ? self.view.superview.bounds.size : self.view.bounds.size;
+  UIView *view = self.view;
+  CGSize viewSize = view.bounds.size;
+  if (self.presentingViewController != nil) {
+    UIView *superview = view.superview;
+    if (superview != nil) {
+      viewSize = superview.bounds.size;
+    }
+  }
   return ASSizeRangeMake(viewSize, viewSize);
 }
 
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASViewController.mm
```diff
@@ -188,7 +202,19 @@ - (void)updateCurrentRangeModeWithModeIfPossible:(ASLayoutRangeMode)rangeMode
...
+  } else {
+    return [self _legacyConstrainedSize];
+  }
+}
+
+- (ASSizeRange)_legacyConstrainedSize
+{
+  // In modal presentation the view does not have the right bounds in iOS7 and iOS8. As workaround using the superviews
+  // view bounds
+  CGSize viewSize = (self.presentingViewController != nil) ? self.view.superview.bounds.size : self.view.bounds.size;
```
Adlai-Holler: Should we check if `superview` is nil? Calling `-bounds` on nil has an undefined result since it returns a struct.

Adlai-Holler:
One note and otherwise lgtm

AsyncDisplayKit/ASViewController.mm
maicki: Yep, we definitely should be defense in here. Change it and should be ready now!

maicki:
@appleguy @Adlai-Holler Any thoughts on merging that?

Adlai-Holler:
The client confirmed it fixes the issue and I like the code! Merging

Issue #892, opened 2015-11-29
[ASCellNode] Verify implementation of manual de-selection.


appleguy:
See commentary in:

https://github.com/facebook/AsyncDisplayKit/pull/837

appleguy:
cc @levi, @vitalybaev

hashemp206:
@appleguy  thanks for trying to fix the bug, bug still exists! and another bug created!
here is how I select cells.
``` objective-c
- (void)collectionView:(UICollectionView *)collectionView didSelectItemAtIndexPath:(NSIndexPath *)indexPath
{
    HAMenuItemCell *cell = (HAMenuItemCell *)[self.collectionView nodeForItemAtIndexPath:indexPath];
    if (cell.selected) {
        cell.selected = NO;
    }
    else
        cell.selected = YES;
}
``` 
then when I scroll down or up without selected cell being disappeared, setSelected:Yes gets called multiple times! 
[sample app](https://db.tt/gszgJZUV)


and here is the new bug:
![cell selection](https://cloud.githubusercontent.com/assets/3206748/11455736/716554a2-9682-11e5-8096-d5768da897cf.gif)

levi:
@hashemp206 try depending on the default behavior of the `selected` property of `ASCellNode` automatically being set by the selection behavior of UICollectionView. Setting `cell.selected = YES` is where you multiple call issue is coming from.

levi:
@hannahbannana this is something that I think fell through the cracks last year. Could you take a look when you have a spare moment?

levi:
Oops, @hannahmbanana

hannahmbanana:
There is definitely an issue with `setSelected:` getting called multiple times here even after excluding the aforementioned manual cell selection code. I interpret the UIKit documentation as saying that they don't do anything in -prepareForReuse, however it appears that UICollectionViewCells have their selected and highlighted states reset when preparing for reuse. I'm not sure of the idea way to resolve this without multiple calls being made, but I'll take a deeper look tomorrow.

hannahmbanana:
Related: #860

levi:
@hannahmbanana thanks for looking into this. I recall testing this behavior on a vanilla UICollectionView and multiple call were being made. Definitely double check that when diagnosing this. Thanks!

hannahmbanana:
[...]

PR #1816, opened 2016-06-25
[ASCellNode] Fix selection / highlight implementation

hannahmbanana:
Addresses Issue #892:

The issue is that two truths existed for the selected/highlighted state of the collection / table:
1. UIKit retains the selected / highlighted states of its cells, even as they are reused (e.g. it saves these values as cells move offscreen so they can be reset once a cell is comes back onscreen).
2. ASCellNode hold their selected / highlight state in their own properties as well.

When a UIKit _ASCollectionViewCell has its node reset as it comes back onscreen, it copies the UIKit cell values and overwrites the selected / highlighted state on the ASCellNode. This fact combined with the fact that setting selected / highlighted on the node, had no mechanism to change the UIKit selected state, meant that it is currently not possible to change the actual selected state by setting on the node. 

I believe the reason this hadn't been noticed, is because the most common use of the ASCellNode selection occurring by user touch, not to change the selection state of the node programatically. For programmatic selection, developers usually call the collection / table select / deselect method. So even though this pattern is uncommon, thankfully one community member noticed it!

Implementation:
- Renamed ASCellNodeLayoutDelegate to ASCellNodeInteractionDelegate 
- Added 2 new methods to the delegate which are called by ASCellNode when selected / highlighted are set.
- Delegate (table / collection) then calls selectItemAtIndexPath / deslectItemAtIndexPath
- Added conditions to chekc if the value provided to setSelected: / setHighlighted: has actually changed in order to avoid duplicate calls. 
- Changed ASCellNode getter methods to be consistent with UIKit's naming:  isSelected / isHighlighted (deprecated the other versions)
- yay unit tests!


AsyncDisplayKit/ASCellNode.mm
```diff
@@ -170,14 +170,40 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
...
+    [_interactionDelegate nodeSelectedStateDidChange:self];
+  }
+}
+
+- (void)setHighlighted:(BOOL)highlighted
+{
+  if (_highlighted != highlighted) {
+      _highlighted = highlighted;
+      [_interactionDelegate nodeHighlightedStateDidChange:self];
+    }
```
appleguy: It looks like this method has too much indentation after the if.  The one above is fine.

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -170,14 +170,40 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
...
+      [_interactionDelegate nodeDidRelayout:self sizeChanged:sizeChanged];
     });
   }
 }
 
+- (void)setSelected:(BOOL)selected
+{
+  if (_selected != selected) {
+    _selected = selected;
+    [_interactionDelegate nodeSelectedStateDidChange:self];
```
appleguy: I'm not certain if this will cause a problem, but I just realized that this will call the _interactionDelegate even in the case that the selection is coming from UIKit (rather than a programmer calling setSelected: manually on the node).

One possible result: if a selection had been occurring in an animated way, then passing through the call and defaulting to a non-animated selection could result in overriding the animation / forcing the selection to be instant.

We should test this, and may have to add a way for the _ASTableViewCell / _ASCollectionViewCell to set these values without _interactionDelegate being called.  One way would be to temporarily clear the delegate while setting them, or to have a way to ignore the calls generated during setting them, or to introduce entirely different internal / private ways of setting them that don't do this.  All of them are annoying / shitty :(

AsyncDisplayKitTests/ASCollectionViewTests.m
```diff
@@ -134,4 +150,54 @@ - (void)testCollectionViewController
...
+  [testController.collectionView layoutIfNeeded];
+  XCTAssertTrue(node.isSelected == YES, @"Reloaded cell should preserve state.");
+  
+  // deselecting cell should deselect node
+  cell = [testController.collectionView cellForItemAtIndexPath:indexPath];
+  cell.selected = NO;
+  XCTAssertTrue(node.isSelected == NO, @"Deselecting cell should update node selection.");
+  
+  // check setSelected not called extra times
+  XCTAssertTrue([(ASTextCellNodeWithSetSelectedCounter *)node setSelectedCounter] == 6, @"setSelected: should not be called on node multiple times.");
```
appleguy: Thanks for adding tests to all your recent changes, @hannahmbanana !

Adlai-Holler:
@hannahmbanana Could you rebase and update this when you get a mo?

hannahmbanana:
@Adlai-Holler @maicki @appleguy: rebased & ready for review again

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -42,20 +42,31 @@ @implementation _ASCollectionViewCell
 - (void)setNode:(ASCellNode *)node
 {
   _node = node;
-  node.selected = self.selected;
-  node.highlighted = self.highlighted;
+  if (node.selected != self.selected) {
+    [node __setSelectedFromUIKit:self.selected];
```
appleguy: Cool, this new internal wrapper method helps simplify down some of the complex handling here around the interaction delegate (since we don't want internally-generated calls like this to trigger informing UIKit of changes)>

I think you can remove the additional != if around these calls now that the latest version of the diff includes these inside the __set methods.

AsyncDisplayKit/ASCellNode.h
```diff
@@ -122,6 +124,12 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
 
 @end
 
+@interface ASCellNode (Deprecated)
```
appleguy: Could you add a comment above this saying:

// Previous versions of ASDK did not include "is" in the name of the getter for these properties.
// These older accessor methods don't match UIKit naming, and will be removed in a future version.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -67,20 +67,31 @@ - (void)didTransitionToState:(UITableViewCellStateMask)state
 - (void)setNode:(ASCellNode *)node
 {
   _node = node;
-  node.selected = self.selected;
-  node.highlighted = self.highlighted;
+  if (node.selected != self.selected) {
```
appleguy: Also can remove here

hannahmbanana:
Another mysterious Travis failure. Will rerun this one as well.

hannahmbanana:
@maicki @Adlai-Holler @appleguy: I addressed @appleguy's latest comments. I think this is ready to be merged.

Adlai-Holler:
Lovely!

PR #1821, opened 2016-06-27
[ASDelegateProxy] Add `conformsToProtocol:`

maicki:
Currently we don't forward `conformsToProtocol:` to the `target` in `ASDelegateProxy`. Unfortunately there can be some problems with that. I ran into this as I tried to implement a custom UICollectionViewLayout that checks if it's delegate conforms to a protocol. Currently this will will crash so we should forward that `conformsToProtocol:` call to the target.


Adlai-Holler:
👍

Issue #1362, opened 2016-03-10
[ASNetworkImageNode] Issue with setting defaultImage as a resizable image


tomizimobile:
I have an ASNetworkImageNode that uses a resizable image as the `defaultImage`. The problem is that once the image is loaded, it will sometimes be shown with the wrong content mode. I dug around a little bit, and I think it's because in _ASCoreAnimationExtras.mm, `ASDisplayNodeSetupLayerContentsWithResizableImage`, the `layer.contentsGravity` is set to `kCAGravityResize`, and there's not another point where the `contentsGravity`/`contentMode` are reset when the image is loaded. Maybe if there's some kind of pending state that tracks `contentMode` changes until the image is loaded and then applied, this would fix it. For the time being, I have a workaround that avoids using resizable images as the `defaultImage`.

appleguy:
@tomizimobile - this is a great find and I think your analysis is correct. Do you have any current ideas on how we might fix this? I believe it is probably affecting a few other areas as the function you mentioned is used elsewhere. 

Note that there is very large memory cost to making even a single color image full-size rather than stretchable, so you might consider using -placeholderImage (you have to subclass a node and override this method to return the placeholderImage). This uses a separate placeholder layer and does wait for the image to load from the network. You probably want this behavior anyways as it will decode synchronously as defaultImage is asynchronous. 

The main benefit of defaultImage is that it will remain visible even if no URL is set or if the image load fails. This is not the intended use of the placeholder, which will only exist during loading or decoding. But it should not be difficult to catch this case and show a default profile picture or whatever.

tomizimobile:
@appleguy The only idea to fix this that I have right now is what I mentioned above - to use another variable in `ASImageNode` to track changes in `contentMode`, and apply it in `setImage:`. I'm not sure what other things that might affect though, or what the performance impact would be - I'm sure there's optimizations to be made there. Unfortunately I don't have enough time to try out any fixes at the moment.

I was using a small, single color image along with the `contentMode` to stretch it rather than using a full-size image, although I'm not sure of the memory cost of that vs. stretchable images. I decided to use `defaultImage` so I wouldn't have to worry about what happens when the image load fails, but I definitely see the advantage now to using `placeholderImage` and handling the error case separately - thanks for the tip.

appleguy:
@tomizimobile: would you be willing to post a PR with your suggested fix?

garrettmoon:
This issue was moved to TextureGroup/Texture#126

PR #1822, opened 2016-06-27
[ASNetworkImageNode] fix for contentMode when using resizable defaultImage

tomizimobile:
This is a proposed fix for #1362. It tracks changes in `contentMode` when a resizable image is set for `defaultImage`, and applies the pending content mode before setting the image.


appleguy:
@tomizimobile this seems interesting - I don't quite understand it in detail, though.  Could you add some additional comments, even just a description walkthrough here?

Does this handle cases where the network node loads fully, but then its URL is set to nil, and it needs to revert to the defaultImage?  There are many such paths...

tomizimobile:
@appleguy A walkthrough: as mentioned in the original issue, the layer's `contentsGravity` is set to resize if an image is stretchable, and at no other point is it set to the desired `contentMode`. This code checks if the `defaultImage` is stretchable when set, and if so, will track changes in the `contentMode`. The `_pendingContentModeState` is also set at that time to the current `contentMode` in case it was set before `defaultImage` was. Then, in `-setImage:`, if the image isn't the defaultImage, it'll set the `contentMode` to the pending state first. My flag `_needsToApplyPendingContentMode` ensures the comparison of the image to the `defaultImage` and setting `contentMode` are only done when necessary.

I hadn't considered any case where the image is set to the `defaultImage` after the node loads. I'm pretty sure I'll need to make some changes - I'll take a closer look at it and add some comments in the source when I have a chance.

tomizimobile:
@appleguy I reworked the code a bit, but it doesn't seem to work well with animated images. I have a test project that I was using to test this with that is very basic - just showing an ASNetworkImageNode with a test image url - but animated images only seem to work correctly a small portion of the time (or if there's a delay in setting the url). I can't track down exactly why but maybe someone else can give it a look?

hannahmbanana:
@tomizimobile: could you share your test project? thanks!

tomizimobile:
@hannahmbanana Here's my test project - I've defined `pngURL` and `gifURL` for testing with, and you can see where I set the image URL at line 86 of PostViewController.swift.

[asdk-issue-1822.zip](https://github.com/facebook/AsyncDisplayKit/files/418046/asdk-issue-1822.zip)

keleftheriou:

appleguy:
test this please

appleguy:
^^ command to trigger jenkins to see if tests pass

maicki:
Hey @tomizimobile, could you please rebase your current branch on top of master so the CI can pass? That would be awesome, Thanks!

tomizimobile:
@maicki I've updated the code - I'm still seeing an issue with animated images though. Not sure what I'm missing with those, maybe someone else has input on that.

maicki:
@tomizimobile Thanks for updating! Can we see it in your sample project you attached?

tomizimobile:
@maicki whoops, forgot to attach a modified project, it's been updated for Swift 3 and new layout APIs. You can see lines 87-89 of PostViewController.swift show where i set the URL. With the example gif, you can see it almost always gets stretched still, although curiously doesn't if a delay is applied. The fix seems ok for normal images though.

[asdk-issue-1822.zip](https://github.com/facebook/AsyncDisplayKit/files/604820/asdk-issue-1822.zip)

garrettmoon:
@tomizimobile If you would be willing to get the project building again I promise I'll actually take a look at this (ASDK in the podfile is currently pointing to a repo on your desktop and it doesn't build when pointed at your personal master).

tomizimobile:
@garrettmoon hmmm, that project worked fine for me when pointed at my master. Regardless, I've updated the project and merged my master with the current master, and it seems to be working as expected on the first run. For some reason though, now it doesn't want to load the gif at all on subsequent runs - don't really have the time at the moment to dig into it further to see if it's something I've added or a different bug, but here's the project: [asdk-issue-1822.zip](https://github.com/facebook/AsyncDisplayKit/files/764913/asdk-issue-1822.zip)

garrettmoon:
@tomizimobile GIFs are actually broken in ASDK at the moment :( But fixed if you point to PINRemoteImage master.

garrettmoon:
I'll be trying to get the PINRemoteImage GIF fix into ASDK ASAP. We can circle back on this then.

Adlai-Holler:

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -121,6 +126,17 @@ - (void)setImage:(UIImage *)image
...
+  // this will ensure the contentMode is properly set when the final image is loaded
+  if (_hasStretchableDefaultImage) {
+    if (!_needsToApplyPendingContentMode && ASObjectIsEqual(image, _defaultImage)) {
+      _pendingContentModeState = self.contentMode;
+      _needsToApplyPendingContentMode = YES;
+    } else if (_needsToApplyPendingContentMode && !ASObjectIsEqual(image, _defaultImage)) {
+      _needsToApplyPendingContentMode = NO;
+      self.contentMode = _pendingContentModeState;
+    }
+  }
```
Adlai-Holler: Minor – we should guard the new flags with `__instanceLock__`. Unfortunately that will mean either:

- Holding the lock while calling into super (who will probably reacquire the lock)
- Relinquishing the lock before calling into super (and breaking the atomicity of these operations)

I don't really like either of those options, but I think option #1 is the best for the time being. When it comes to recursive locks, "in for a penny, in for a pound" 😈

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -193,6 +210,23 @@ - (UIImage *)defaultImage
   return _defaultImage;
 }
 
+- (void)setAnimatedImage:(id<ASAnimatedImageProtocol>)animatedImage
+{
+  if (_needsToApplyPendingContentMode) {
```
garrettmoon: As @Adlai-Holler said, these two methods need to be guarded with the instance lock.

garrettmoon:
Hey @tomizimobile we're going to close this for now as it's getting stale, but we'll gladly re-open if you address comments and rebase. Thanks for putting this up!

PR #1823, opened 2016-06-27
[Project] Publicize ASPINRemoteImageDownloader.h

### LINKED ISSUE #1815: [Swift] ASPINRemoteImageDownloader and PINCache public access
Is there anyway I can access to `PINRemoteImageManager` and `PINCache` that I included in `ASDK` in my project? How can I make it public?

Also I need to use PINCache inside my project for other purposes, now I have to duplicate `import PINCache`, one in `ASDK` and other in my project, and handling the cache is a nightmare for me.
I would be pleased to know how can I manage this situation?
Thanks! :)

Adlai-Holler:
The header contents are wrapped in `#if PIN_REMOTE_IMAGE` so we won't compile anything for users that are on our "Core" spec. I've tested this and it resolves #1815 and behaves well for Core & default subspecs. @appleguy @maicki


hannahmbanana:
Whomever merges this, could you also close the issue?

maicki:
@Adlai-Holler Ok cool seems good

PR #1828, opened 2016-06-29
[ASTextNode] Release Lock Sooner in -setAttributedText:

Adlai-Holler:
This reduces the number of calls to -length, and also releases the lock as soon as possible.  

Because we use the local variable, there is a chance that another thread will set _attributedString before me, but self.accessibilityLabel after me, and they'd fall out of sync. Since mitigation would be so ugly, this is considered an acceptable risk. Review? @maicki @appleguy

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -366,22 +366,26 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
 
 - (void)setAttributedText:(NSAttributedString *)attributedText
 {
-  std::lock_guard<std::recursive_mutex> l(_textLock);
   
   if (attributedText == nil) {
     attributedText = [[NSAttributedString alloc] initWithString:@"" attributes:nil];
   }
+  
+  // Don't hold the textLock for too long
+  {
+    std::lock_guard<std::recursive_mutex> l(_textLock);
+    if (ASObjectIsEqual(attributedText, _attributedText)) {
+      return;
+    }
 
-  if (ASObjectIsEqual(attributedText, _attributedText)) {
-    return;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -366,31 +366,35 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 
 - (void)setAttributedText:(NSAttributedString *)attributedText
 {
-  std::lock_guard<std::recursive_mutex> l(_textLock);
   
   if (attributedText == nil) {
     attributedText = [[NSAttributedString alloc] initWithString:@"" attributes:nil];
   }
+  
+  // Don't hold textLock for too long.
```
appleguy: @Adlai-Holler this probably only provides value if we call out that we're interested in avoiding locking the displaynode lock while holding this one.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -366,31 +366,35 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
...
-  // Without this, the size calculation of the text with truncation applied will
-  // not take into account the attributes of attributedText in the last line
-  [self _updateComposedTruncationText];
   
-  // We need an entirely new renderer
-  [self _invalidateRenderer];
+  NSUInteger length = attributedText.length;
+  if (length > 0) {
+    CGFloat screenScale = ASScreenScale();
+    self.ascender = round([[attributedText attribute:NSFontAttributeName atIndex:0 effectiveRange:NULL] ascender] * screenScale)/screenScale;
```
appleguy: Good call, in the future we will have a layout lock on these.

appleguy:
Looks great, let's get this in.

PR #1830, opened 2016-06-29
[ASDataController] Prevent calling endUpdatesAnimated:completion: in an unbalanced way

maicki:


Adlai-Holler:
Hmm this change seems good, and it exposed some other (apparently serious) internal logic issue. Even better! I've looked at the code for those failing tests and they use the API correctly, so the failure may lie in ASTableView -> ASChangeSetDataController. I will look at this today and see if that fix is small enough to address in this PR. If not, we will create a separate issue and temporarily disable these tests.

AsyncDisplayKit/Details/ASChangeSetDataController.m
```diff
@@ -43,6 +40,9 @@ - (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
   ASDisplayNodeAssertMainThread();
   _changeSetBatchUpdateCounter--;
   
+  // Prevent calling endUpdatesAnimated:completion: in an unbalanced way
+  NSAssert(_changeSetBatchUpdateCounter < 0, @"endUpdatesAnimated:completion: called without having a balanced beginUpdates call");
```
Adlai-Holler: Let's do `_changeSetBatchUpdateCount >= 0` instead

Adlai-Holler:
@maicki Once the assertion is flipped, this is ready to 🚢

maicki:
@Adlai-Holler Ready to roll 🚗

PR #1831, opened 2016-06-30
[ASDisplayNode] added asyncTraitCollectionDidChange method

rcancro:
Subclasses can override this method to react to a trait collection change

Also added NSIndexSet+ASHelpers to be part of the pod. The example wouldn't build without it.


appleguy:
@rcancro thanks for making this so efficient :)

PR #1833, opened 2016-06-30
Don't propagate trait collections to cells if the node is not loaded yet

maicki:
This fixes an issue where the propagation of trait collections trigger are creation of the node if the node view was not loaded yet.

cc @rcancro Does this make sense or am I overlooking something?


rcancro:
Looks good!!!!

AsyncDisplayKit/Details/ASEnvironment.h
```diff
@@ -157,6 +157,10 @@ ASDISPLAYNODE_EXTERN_C_END
   ASDN::MutexLocker l(lock);\
   ASEnvironmentTraitCollection oldTraits = self.environmentState.environmentTraitCollection;\
   [super setEnvironmentState:environmentState];\
+\
+   /* Extra Trait Collection Handling */\
```
appleguy: @maicki it looks like there is an extra space in front of this line compared to the next line (maybe?).

Thanks a lot for debugging this!

PR #1834, opened 2016-06-30
[ASVisibility] Don't change the range mode if the node was not loaded yet

maicki:
Changing the range mode needs to access the view within the node. If the view was not loaded at this point it will trigger a view load, but that's not necessarily wanted so prevent changing it if the node is not loaded yet.

cc @garrettmoon Garrett does this change makes sense or do you have other ideas around it?

### DIFF
--- a/AsyncDisplayKit/ASCollectionNode.h
+++ b/AsyncDisplayKit/ASCollectionNode.h
@@ -25,10 +25,10 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithCollectionViewLayout:(UICollectionViewLayout *)layout;
 - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout;
 
+@property (strong, nonatomic, readonly) ASCollectionView *view;
+
 @property (weak, nonatomic) id <ASCollectionDelegate>   delegate;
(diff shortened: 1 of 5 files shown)


garrettmoon:
Should this be handled at the <ASRangeControllerUpdateRangeProtocol>node level instead? I.e. prevent it from loading the node based on this?

Could we, for example, only schedule a rangeUpdate if the node isn't loaded?

maicki:
Yes we could add that check on the node level instead of the `ASViewController` level, but what if we really want to force the range update on the node, that was the thinking behind it.

garrettmoon:
Could the range update happen on node load? I honestly don't know the answer, but in there a reason to force the node to load to update the range? Or would it be better to update the range and have it applied when the node loads?

maicki:
We could have a pending state as in `ASDisplayNode` for the range mode and on node load we could apply it.

So logic would be:
1. If node is loaded just apply
2. If node is not loaded save as pending state
2.1 If node is loading apply pending state

This could also work for range parameter adjustments

garrettmoon:
That seems like the right solution to me!

maicki:
@garrettmoon Added a pending state in ASCollectionNode and ASTableNode for the range mode

cc @Adlai-Holler @appleguy

garrettmoon:
LGTM!

PR #1839, opened 2016-07-01
[ASDisplayNode] Allow measure always be off the main thread

maicki:
Currently measurement always needs to happen on the main thread if implicit hierarchy management is enabled as adding and removing from nodes needs to happen on the main thread. We now will trampoline to the main thread to do the insertion and deletion of nodes.

This also resolves the issue that can occur if a node is already loaded deep in the layout hierarchy in the layout that the node is transforming to. Before insertion or deletion is happening we need to crawl the layout hierarchy to check that though.

*Open things:*
* This theoretically that would enable us always measure nodes in the data controller asynchronous
* Furthermore this also would us allow to get rid of the `shouldMeasureAsync` parameter in `transitionLayoutWithAnimation:shouldMeasureAsync:measurementCompletion:` that I think should not be exposed at all - @levi any thoughts on that?
* Do we need to check somehow the transition id in `_applyLayout:layoutTransition:` to abort the insertion or deletion of nodes if the block runs on the main thread?
* Would it be worth to have it's own transition queue instead of always using one of the global background queues?

@appleguy @Adlai-Holler @levi Would be great if I could get a review and thoughts

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -718,7 +718,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   int32_t transitionID = [self _startNewTransition];
   
   ASDisplayNodePerformBlockOnEverySubnode(self, ^(ASDisplayNode * _Nonnull node) {
-    ASDisplayNodeAssert([node _hasTransitionInProgress] == NO, @"Can't start a transition when one of the subnodes is performing one.");
+    ASDisplayNodeAssert([node _isTransitionInProgress] == NO, @"Can't start a transition when one of the subnodes is performing one.");
     node.hierarchyState |= ASHierarchyStateLayoutPending;
     node.pendingTransitionID = transitionID;
   });
@@ -821,7 +821,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (void)cancelLayoutTransitionsInProgress
 {
   ASDN::MutexLocker l(_propertyLock);
-  if ([self _hasTransitionInProgress]) {
+  if ([self _isTransitionInProgress]) {
     // Cancel transition in progress
     [self _finishOrCancelTransition];
       
@@ -844,7 +844,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   _usesImplicitHierarchyManagement = value;
 }
 
-- (BOOL)_hasTransitionInProgress
+- (BOOL)_isTransitionInProgress
 {
   ASDN::MutexLocker l(_propertyLock);
   return _transitionInProgress;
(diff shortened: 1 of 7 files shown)


Adlai-Holler:
If applyLayout:layoutTransition is called on a background thread, is it possible that a subnode involved in the transition could get its node loaded on the main thread after `ASLayoutCanTransitionAsynchronous` but before `applySubnodeTransition`?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -838,8 +846,8 @@ - (void)setUsesImplicitHierarchyManagement:(BOOL)value
 
 - (BOOL)_hasTransitionInProgress
```
levi: This should be _isTransitionInProgress

AsyncDisplayKit/Private/ASLayoutTransition.h
```diff
@@ -18,16 +18,45 @@
...
-              previousLayout:(ASLayout *)previousLayout;
+/**
+ * Pending layout to transition to
+ */
+@property (nonatomic, readonly, strong) ASLayout *pendingLayout;
 
+/**
+ * Returns if the layout transition can happen asynchronously
+ */
+@property (nonatomic, readonly, assign) BOOL canTransitionAsynchronous;
```
levi: `@property (nonatomic, readonly, assign) BOOL isSynchronous;`

AsyncDisplayKit/Private/ASLayoutTransition.mm
```diff
@@ -18,10 +18,37 @@
...
+#import <queue>
 
 #import "NSArray+Diffing.h"
 #import "ASEqualityHelpers.h"
 
+/**
+ * Search the whole layout stack if at least one layout has a layoutable object that can not be layed out asynchronous.
+ * This can be the case for example if a node was already loaded
+ */
+static BOOL ASLayoutCanTransitionAsynchronous(ASLayout *layout) {
```
levi: Make inline

AsyncDisplayKit/Private/ASLayoutTransition.h
```diff
@@ -18,16 +18,45 @@
...
+/**
+ * Returns a newly initialized layout transition
+ */
+- (instancetype)initWithNode:(ASDisplayNode *)node pendingLayout:(ASLayout *)pendingLayout previousLayout:(ASLayout *)previousLayout NS_DESIGNATED_INITIALIZER;
+- (instancetype)init NS_UNAVAILABLE;
+
+/**
+ * Insert and remove subnodes that where added or removed between the previousLayout and the pendingLayout
+ */
+- (void)applySubnodeTransition;
```
levi: `completeTransition` or something similar. Would be great to add a level of indirection in the naming such that the concept of subnode management isn't exposed to the caller.

maicki: I named it `startTransition` now. We should also look into finding a way to not have to expose `applySubnodeInsertions` and `applySubnodeRemovals` to the consumer.

maicki:
@Adlai-Holler Theoretically this could happen, in fact this can happen with the current implementation too. So in this nothing will change. I don't know at the moment a good way to fix that, but we definitely should consider it in the broader refactoring around layout transitions that I'm working on.

@levi Addressed your comments

@levi @appleguy @Adlai-Holler Can I get another look on it. Thanks!

Adlai-Holler:
Awesome! Time to 🚢

PR #1840, opened 2016-07-01
[ASDataController] Remove Aggregate CGRect Initializer

Adlai-Holler:
This irritates other compilers/settings.


Adlai-Holler:
@maicki

PR #1844, opened 2016-07-04
[ASDisplayNode] Do not start measurement for a transition if we have never been provided a constrainedSize.

appleguy:
Also removes a deprecated CoreText property that fails to build with iOS 10.

In the screenshots below, we see a case where the client called transitionLayout... before the first measurement completed.  It's actually possible that measurement was already in progress.

My biggest question would be for @levi - with the new change to put the constrainedSize on _layout, we can't detect what constrainedSize was provided to start the layout pass that is currently underway.  Technically it might be "most correct" to obey the transition call as a new layout (at least, a setNeedsLayout) and use the same constrainedSize as provided to the in-flight measurement.  I presume the measurement is in flight because ASDataController should be doing that for this element, yet _layout is nil.

We can see there are some pretty bad implications to calling into the layout code with zero size.  @maicki could you document a followup to audit the code for any other points we should check for zero size and bail?  I don't think it could ever result in a useful layout.  Interestingly, at one point the height goes from 0 to +Inf...

screen shot 2016-07-03 at 6 17 39 pm
screen shot 2016-07-03 at 6 17 23 pm


maicki:
@appleguy After the measurement pass would finish do we do a layout again? If we just bail would the node be in a unmeasured state?

maicki:
@appleguy Can you rebase that branch with master as we addressed the CoreText property with a merge for Xcode 8.0 beta 2 fixes.

PR #1845, opened 2016-07-04
Added docs for ASImageNode

lappp9:
This one may need some fact checking.  I'm sure the `cropRect` section could use a bit more information.


appleguy:
Sweet, thanks Luke!

Issue #1848, opened 2016-07-05
[ASNetworkImageNode] Crash when the URL is a local file that doesn't exist


flovouin:
Hi guys,

I just found a bug on `master` (not present in 1.9.80) that makes the `ASNetworkImageNode` crash when the specified `URL` is a file URL that doesn't exist. The error is:
```Assertion failure in -[PINAnimatedImage initWithAnimatedImageData:], PINAnimatedImage.m:54
'NSInternalInconsistencyException', reason: 'animatedImageData must not be nil.'```
The reason is simple: when trying to load the local file as an animated image, the data is not checked to verify if it's not `nil` (see line [482](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/ASNetworkImageNode.mm#L482)).
For this bug to appear, `shouldCacheImage` should be set to `NO`. Here is a simple project that shows what happens: https://github.com/flovouin/ASDK-NetworkImageCrash

I can submit a small PR if needed, that simply checks `data` before forwarding it to the `_downloader`. But maybe someone has a different way of handling this?

Cheers,

flovouin:
Solved by PR #1849, thanks for merging it!

PR #1849, opened 2016-07-05
[ASNetworkImageNode] Fix crash when local file URL doesn't exist

flovouin:
Hi guys,

This is a very simple PR concerning the issue I just opened #1848. A test project can be found here: https://github.com/flovouin/ASDK-NetworkImageCrash
Basically what happens is that when `shouldCacheImage = NO` and the `URL` fed to an `ASNetworkImageNode` points to a file that doesn't exist, `nil` data is passed to the `_downloader`, which causes a crash.

My fix is simply to check that the data is indeed not `nil` before trying to load an animated image. However I get the feeling that something more could be done here. If the image exists but is not animated, it gets loaded two times: once as the `nonAnimatedImage`, and once as the `animatedImage`. Isn't there a way to load the image only once and check whether it's animated?

Tell me what you think :-)
Cheers,

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -480,10 +480,12 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
             id<ASAnimatedImageProtocol> animatedImage = nil;
             if (_downloaderFlags.downloaderImplementsAnimatedImage) {
               NSData *data = [NSData dataWithContentsOfURL:_URL];
-              animatedImage = [_downloader animatedImageWithData:data];
+              if (data != nil) {
+                animatedImage = [_downloader animatedImageWithData:data];
 
-              if ([animatedImage respondsToSelector:@selector(isDataSupported:)] && [animatedImage isDataSupported:data] == NO) {
-                animatedImage = nil;
+                if ([animatedImage respondsToSelector:@selector(isDataSupported:)] && [animatedImage isDataSupported:data] == NO) {
+                  animatedImage = nil;
+                }
               }
             }
 


hannahmbanana:
@garrettmoon: Could you review this PR?

garrettmoon:
This seems like a good bandaid, but I also agree that the _lazilyLoadImageIfNecessary could be refactored to improve.

I think we intentionally used imageWithContentsOfFile because it's documented to not cache the results. However the documentation states the same for imageWithData. So I don't see a reason to not use imageWithData, only after retrieving the data and checking if it's an animatedImage first.

maicki:
@garrettmoon Is `imageWithContentsOfFile:` thread safe? `imageWithData:` is not until iOS 10: https://pspdfkit.com/blog/2016/investigating-thread-saftey-of-UIImage/

appleguy:
Yes, by my knowledge imageWithContentsOfFile: has been threadsafe for a long time, even though imageNamed: has not been.

garrettmoon:
@maicki Hmm, I would guess imageWithContentsOfFile would have the same issues described in that blog post. It sounds like there may be an OS bug. Every image framework I know of relies on creating images on background threads :/

flovouin:
Thanks everyone for your messages!
Would it be acceptable to merge this bandaid, as this would currently solve the crash I'm experiencing in my app?
I'm definitely interested in the rest of the discussion, and am willing to experiment with some `_lazilyLoadImageIfNecessary` tweaking if you want me to (although you guys are probably more knowledgeable than me in this area).
Let me know!

maicki:
@Adlai-Holler @garrettmoon That PR looks good to me ready to merge?

We should open a new issue in case if we want to look into refactoring `_lazilyLoadImageIfNecessary` though, but I think the initial issue would be resolved.

appleguy:
@flovouin thanks for the clearly-written explanation, and the PR submittal - this will be in the 1.9.81 release going out shortly!

PR #1851, opened 2016-07-06
Copy header files using the Copy Files phase

kielgillard:
1e3e5a5 / https://github.com/facebook/AsyncDisplayKit/releases/tag/1.9.80

Our app is dependent on and links against the ASDK static library target. During release builds, the `Headers` phase for the ASDK static library target copies header files into a directory like:

> ~Library/Developer/Xcode/DerivedData/project-elqpjeogvesgbpfqlazseekvpbli/Build/Intermediates/ArchiveIntermediates/target/IntermediateBuildFilesPath/UninstalledProducts/iphoneos/include

rather than a directory like:

>  ~Library/Developer/Xcode/DerivedData/project-elqpjeogvesgbpfqlazseekvpbli/Build/Intermediates/ArchiveIntermediates/target/BuildProductsPath/Release-iphoneos/include

...which is where every other dependency (not to mention Apple's template for static libraries) are copying their header files (using a `Copy Files` phase rather than a `Headers` phase).

This PR copies header files using the `Copy Files` phase and replaces the `Headers` phase.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -580,7 +580,6 @@
 		F7CE6C901D2CDB5800BE4C15 /* _ASTransitionContext.h in CopyFiles */ = {isa = PBXBuildFile; fileRef = DB55C25F1C6408D6004EDCF5 /* _ASTransitionContext.h */; };
 		F7CE6C911D2CDB5800BE4C15 /* _ASTransitionContext.m in CopyFiles */ = {isa = PBXBuildFile; fileRef = DB55C2601C6408D6004EDCF5 /* _ASTransitionContext.m */; };
 		F7CE6C921D2CDB5800BE4C15 /* ASBatchFetching.m in CopyFiles */ = {isa = PBXBuildFile; fileRef = 044285061BAA63FE00D16268 /* ASBatchFetching.m */; };
-		F7CE6C931D2CDB5800BE4C15 /* ASDisplayNode+FrameworkPrivate.h in CopyFiles */ = {isa = PBXBuildFile; fileRef = DE6EA3211C14000600183B10 /* ASDisplayNode+FrameworkPrivate.h */; };
 		F7CE6C941D2CDB5800BE4C15 /* ASDisplayNodeInternal.h in CopyFiles */ = {isa = PBXBuildFile; fileRef = 058D0A0C195D050800B7D73C /* ASDisplayNodeInternal.h */; };
 		F7CE6C951D2CDB5800BE4C15 /* ASLayoutTransition.h in CopyFiles */ = {isa = PBXBuildFile; fileRef = E52405B41C8FEF16004DC8E7 /* ASLayoutTransition.h */; };
 		F7CE6C961D2CDB5800BE4C15 /* ASEnvironmentInternal.h in CopyFiles */ = {isa = PBXBuildFile; fileRef = 69E100691CA89CB600D88C1B /* ASEnvironmentInternal.h */; };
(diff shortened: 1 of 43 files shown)


appleguy:
@kielgillard thanks!  We don't have many clients using the static library directly - although I certainly approve of static linking, a lot of folks are achieving that with Cocoapods which I believe creates this build step.  I really appreciate you contributing the fix.

@Adlai-Holler, @hannahmbanana we should see if the Life Without Cocoapods step in build.sh / Travis can be changed to check that this works properly, e.g. fail on current master and pass after this change.  @kielgillard if you have any thoughts on that, let us know.  I would look in more detail, but it is a packed week...

kielgillard:
I'm checking out the Life Without Cocoapods sample now. There looks to be some issues and I'll fix those.

kielgillard:
I'm still addressing compilation issues with both AsyncDisplayKit (static and framework targets) and Life Without Cocoapods. I've run out of time for today.

I suspect this PR might generate some friction when adding new header files. For framework targets, a developer would use Xcode's File inspector to adjust the role of the header file (choosing public, private or internal). For static library targets, you cannot use the File inspector to make the header public (that is, to ensure it copies into `/built/products/directory/include`). A developer will have to add the header file to the `Copy Files` phase manually.

Is this going to rule out this PR and as such waste any more time that I spend on it?

appleguy:
@kielgillard thanks a lot!  Indeed we normally use the role switcher, but if it is necessary to add it to the Copy Headers build phase in addition, we can definitely do that.  As long as we have a CI build test that fails when someone does not do this, I am totally comfortable enforcing that.  The static library / manual integration path is personally important to me that we support, and I'm pretty sad to hear that it was broken, thus my enthusiasm that you both reported this and submitted a fix!

We have a lot of PRs in flight right now, but I can assure you we'll take the fix as soon as we have a chance to understand everything.  E.g. should be as easy as trying out our own test project and attempting to include a header there, to make sure that failed previously, and now works.

kielgillard:
Ok, I feel like I'm done and consider this PR ready for review now that all the tests are passing.

Off the top of my head, I'm not sure how an automated test would reliably detect regressions with header files missing in the `Copy Files` phase of the static library target. For example, taking a leaf out of the CocoaPods book, maybe such a test could recursively check that a `.h` file within `/source/root/AsyncDisplayKit/` exists in `/built/products/directory/include/AsyncDisplayKit/`. But not every header is appropriate to copy into the `include` folder (perhaps those inside `Private`, for example? Ewww!). Yeah, I don't know.

Since static libraries look like they're on their way out (with Frameworks and Swift Package Manager), maybe copying any and all header files into the `include` folder is concession that could be made for simplicity and convenience. In that case, the `Copy Files` phase I'm proposing could be replaced with one that recursively copies header files from `source/root/AsyncDisplayKit` and for the developer there's no manual effort or chance of error.

Please let me know if there's anything I can do to reduce the fiction of the review process. The reproducibility of the original problem is straightforward but I may be able to prepare a sample project if that helps.

appleguy:
@kielgillard hey - although there was a merge conflict, I checked out your changes and performed a manual merge, and got this into master.  I really appreciate your work on this.  Although I haven't been able to review in as detailed a way as I'd like, I wanted to make sure this got into the 1.9.81 release I'm preparing now.

appleguy:
@kielgillard two things:

- I do think a recursive .h copy would be better, if you know of a quick way to write this script (it seems like the toughest part is knowing the right $VARIABLES to use, just a bit of research required).  This is not mandatory, but as files are slowly added, we'll probably forget to add some of them to copy headers.

- What were the reasons to change the other #import statements?  Just want to understand why this was necessary so that future additions don't introduce issues, because there are some stylistic changes (e.g. using "" instead of <> in .m - I realize this is more formally correct, but I'm curious if there was a build failure caused on your end by <> because I'd assume those files will only be referenced while the static library itself is being compiled, at which point I think the compiler would resolve them correctly anyway.

PR #1852, opened 2016-07-06
[ASVideoPlayerNode] Ensure activity indicator view is transparent

gazreese:
At the moment the activity indicator for ASVideoPlayerNode, shown when the video is loading, is displayed inside a black box. This seems to be due to the background of the containing node not being transparent.

This simple change makes the node transparent, leaving the activity indicator shown in white just like the other playback controls.

### DIFF
--- a/AsyncDisplayKit/ASVideoPlayerNode.mm
+++ b/AsyncDisplayKit/ASVideoPlayerNode.mm
@@ -502,6 +502,7 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
       return spinnnerView;
     }];
     _spinnerNode.preferredFrameSize = CGSizeMake(44.0, 44.0);
+    _spinnerNode.backgroundColor = [UIColor clearColor];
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASVideoPlayerNode.mm
```diff
@@ -502,6 +502,7 @@ - (void)showSpinner
       return spinnnerView;
     }];
     _spinnerNode.preferredFrameSize = CGSizeMake(44.0, 44.0);
+    _spinnerNode.backgroundColor = [UIColor clearColor];
```
appleguy: Because this method accesses the view at the bottom, we should add an Assert Main Thread at the top...

It would be slightly better to have this .backgroundColor inside the initWithViewBlock, not only because the spinner color itself is configured there, but because then the property bridge doesn't need to handle it (this is very very minor though :-P)

appleguy:
Thanks @gazreese for all your contributions!

PR #1853, opened 2016-07-06
[ASCollectionView] Greatly Improve Cell Node Resizing

Adlai-Holler:
This is a modification of #1827 that achieves the same effect in a different way. Let's call 1827 option A and this option B. In this option, to achieve the animated relayout, we submit an empty update just before [super layoutSubviews] rather than wrapping the layout in an animation block.

Reasons to go with option B:
- Doesn't cause the collection view to take any snapshots.
- Update animations follow the layout's normal update style (e.g. moved items slide rather than crossfade). Option A causes all items to crossfade.

Reasons to go with A:
- Option A is completely API-safe. This is Apple's "golden path."
- The resized cell node gets a fancy double-sided crossfade animation as it moves to its new size, rather than simply "jumping" to its new size.
- Option B's submitting an empty update inside -layoutSubviews seems a little risky. May not be much worse than we currently submit empty updates async on main.

Great things about both options:
- Right now if you update size without animation, since the update happens asynchronously you will probably see a frame that shows the cell node at its new size, but the collection view layout not updated.

Videos: 
[Updates Videos.zip](https://github.com/facebook/AsyncDisplayKit/files/350557/Updates.Videos.zip)

I advocate for option B. The crossfading and snapshots of option A are ugly.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -91,6 +91,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   id<ASCollectionViewLayoutFacilitatorProtocol> _layoutFacilitator;
   
   BOOL _performingBatchUpdates;
+  BOOL _superPerformingBatchUpdates;
   NSMutableArray *_batchUpdateBlocks;
 
   BOOL _isDeallocating;
@@ -447,6 +448,24 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   return visibleNodes;
 }
 
+#pragma mark Internal
+
+/**
+ Performing nested batch updates with super (e.g. resizing a cell node & updating collection view during same frame)
+ can cause super to throw data integrity exceptions because it checks the data source counts before
+ the update is complete.
+ 
+ Always call [self _superPerform:] rather than [super performBatch:] so that we can keep our `superPerformingBatchUpdates` flag updated.
+*/
+- (void)_superPerformBatchUpdates:(void(^)())updates completion:(void(^)(BOOL finished))completion
+{
+  ASDisplayNodeAssertMainThread();
+  ASDisplayNodeAssert(_superPerformingBatchUpdates == NO, @"Nested batch updates being sent to UICollectionView. This is not expected.");
+  
+  _superPerformingBatchUpdates = YES;
+  [super performBatchUpdates:updates completion:completion];
+  _superPerformingBatchUpdates = NO;
+}
 
 #pragma mark Assertions.
 
@@ -784,7 +803,9 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   _nextLayoutInvalidationStyle = ASCollectionViewInvalidationStyleNone;
   switch (invalidationStyle) {
     case ASCollectionViewInvalidationStyleWithAnimation:
-      [super performBatchUpdates:^{ } completion:nil];
+      if (!_superPerformingBatchUpdates) {
+        [self _superPerformBatchUpdates:^{ } completion:nil];
+      }
       break;
     case ASCollectionViewInvalidationStyleWithoutAnimation:
       [self.collectionViewLayout invalidateLayout];
@@ -1031,7 +1052,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   ASPerformBlockWithoutAnimation(!animated, ^{
     [_layoutFacilitator collectionViewWillPerformBatchUpdates];
-    [super performBatchUpdates:^{
+    [self _superPerformBatchUpdates:^{
       for (dispatch_block_t block in _batchUpdateBlocks) {
         block();
       }
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -24,7 +24,6 @@
 #import "UICollectionViewLayout+ASConvenience.h"
 #import "ASRangeControllerUpdateRangeProtocol+Beta.h"
 #import "_ASDisplayLayer.h"
-#import "ASAvailability.h"
 
 static const NSUInteger kASCollectionViewAnimationNone = UITableViewRowAnimationNone;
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -25,6 +25,12 @@
 #import "ASRangeControllerUpdateRangeProtocol+Beta.h"
 #import "_ASDisplayLayer.h"
 
+typedef NS_ENUM(NSUInteger, ASCollectionViewInvalidationAction) {
+  ASCollectionViewInvalidationActionNone,
+  ASCollectionViewInvalidationActionInvalidateWithoutAnimation,
```
appleguy: Seems pretty cool to have customization here, but these options are not yet self-explanatory.  Add comments, and also consider if we need to include Invalidate twice in the name.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -793,9 +770,25 @@ - (void)layoutSubviews
       [self performBatchAnimated:YES updates:^{
         [_dataController relayoutAllNodes];
       } completion:nil];
+      // We need to ensure the size requery is done before we update our layout.
+      [self waitUntilAllUpdatesAreCommitted];
```
appleguy: I assume this case greatly helps rotation animations?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1143,57 +1136,32 @@ - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged
...
-      // One single non-animated cell node causes the whole context to be non-animated
-      _queuedNodeSizeInvalidationContext.shouldAnimate = NO;
+      // One single non-animated node causes the whole layout update to be non-animated
+      action = ASCollectionViewInvalidationActionInvalidateWithoutAnimation;
     }
   }
-}
-
-// Cause UICollectionView to requery for the new size of all nodes
-- (void)requeryNodeSizes
```
appleguy: Holy shit I am glad to see this obliterated.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1143,57 +1136,32 @@ - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged
...
-    _queuedNodeSizeInvalidationContext = [[_ASCollectionViewNodeSizeInvalidationContext alloc] init];
-    
-    __weak __typeof__(self) weakSelf = self;
-    dispatch_async(dispatch_get_main_queue(), ^{
-      __typeof__(self) strongSelf = weakSelf;
-      if (strongSelf) {
-        [strongSelf requeryNodeSizes];
-      }
-    });
+  NSIndexPath *indexPath = [self indexPathForNode:node];
```
appleguy: This call is not very efficient.  It shouldn't block this diff, but later we should consider if ASCellNodes should be given their indexPath so they can report it more easily.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1143,57 +1136,32 @@ - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged
...
+  if (indexPath == nil) {
+    return;
   }
+
+  [_layoutFacilitator collectionViewWillEditCellsAtIndexPaths:@[ indexPath ] batched:NO];
   
-  [_queuedNodeSizeInvalidationContext.invalidatedNodes addObject:node];
+  ASCollectionViewInvalidationAction previousAction = _nextLayoutInvalidationAction;
+  ASCollectionViewInvalidationAction action = previousAction;
```
appleguy: The way this is done is confusing.  I think we should probably do action = _nextLayout, and then previousAction = action, although even that is a bit strange.  I see why it is needed though.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1143,57 +1136,32 @@ - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged
...
+    [self setNeedsLayout];
+    action = ASCollectionViewInvalidationActionInvalidateWithAnimation;
+  }
 
-  // Check if this node or one of its subnodes can be animated.
-  // If the context is already non-animated, don't bother checking this node.
-  if (_queuedNodeSizeInvalidationContext.shouldAnimate) {
+  // If we think we're going to animate, check if this node will prevent it.
+  if (action == ASCollectionViewInvalidationActionInvalidateWithAnimation) {
     BOOL (^shouldNotAnimateBlock)(ASDisplayNode *) = ^BOOL(ASDisplayNode * _Nonnull node) {
```
appleguy: It's silly, but this block will cause measurable retain / release overhead in profiles.  Not an issue now, but we should earmark this for integration into ASEnvironment state propagation.  There are nice mechanisms available to be able to answer this kind of question by checking state on one node.  The state propagation occurs whenever a node with the non-default value is added as a subnode; if one is removed as a subnode, only in the case that the one removed has a non-default value does the tree need to be re-queried to check if the state can be returned to default, and since removal of subnodes is less common this turns out to be a win.

AsyncDisplayKit/ASCollectionView.mm
Adlai-Holler: Haven't tested rotation but this was necessary to support an image carousel that resizes while it scrolls. Otherwise the sizeForItemAtIndexPath: that we return would be out-of-date.

AsyncDisplayKit/ASCollectionView.mm
Adlai-Holler: Hmm it would be nice but a major pain and very expensive to update the index paths of all nodes whenever an update is performed. We could maybe benefit from using NSOrderedSet but probably not. According to https://github.com/Necropolis/Set/blob/master/NSSet-PerformanceDemonstration/NSSet-PerformanceDemonstration.md NSOrderedSet.indexOfObject is about 25x faster on indexOfObject: but about 9x slower on addObject. As always Apple is silent about the theoretical performance characteristics.

appleguy:
@Adlai-Holler looks great, thanks!

PR #1860, opened 2016-07-07
	[ASTraitCollection] Remove traitCollectionContext from ASTraitCollection; add containerWindowSize

rcancro:
Passing around a pointer was leading to crashes as the ASVC was the sole owner of the context. There are cases where the VC would dealloc while its subnodes were laying out. This could lead to the subnodes accessing a garbage pointer.


Adlai-Holler:
Good stuff! Landing this

PR #1863, opened 2016-07-07
[ASTextNode] Fix text node truncation

### LINKED ISSUE #1841: Truncated ASTextNode appearing in node that should not be truncated.
ASCollectionNode, with ASCellNodes

Please note there is no cell re-use. They look similar (aside from background images) for testing purposes. This reproduces with _any_ string that gets truncated at smaller cell sizes.

Layout approach is pure `layoutSpecThatFits` from root `ASCellNode` down the view hierarchy.

In overriding `visibilityDidChange` I can see that the `constrainedSize` is correct for every node, including the text nodes that have the incorrect truncation. The `bounds`, however, are incorrect for the truncated text nodes.

After the resize transition, many nodes are properly laid out. When you scroll you _eventually_ hit cells that have the `ASTextNode` truncating the `NSAttributedString` to the same place as it was previous to the resize.

This happens regardless of text formatting or otherwise, and can be reproduced with a plain `NSAttributedString`.

In summary, if a `ASTextNode` has ever been truncated, it can appear again truncated with a larger `constrainedSize`, as if it was cached and reused or something. Important to note that this seems to happen for cells that are well out of any buffer ranges.

errrr-cat

maicki:
Before truncation happens no invalidation of the text storage happened. That had the effect, if the constrained size of a text node increases a new size for the text node will be calculated. While this calculation is happening a truncation on the text will happen. Unfortunately the truncation will happen on the already truncated string.

This PR resets the text storage to the original value before the truncation is happening so it always will truncate on the same string no matter what constrained size is given.

Should fix #1841

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitContext.h
+++ b/AsyncDisplayKit/TextKit/ASTextKitContext.h
@@ -10,6 +10,8 @@
 
 #import <UIKit/UIKit.h>
 
+typedef NSTextStorage *(^ASTextKitContextTextStorageCreationBlock)(NSAttributedString *attributedString);
+
 /**
  A threadsafe container for the TextKit components that ASTextKit uses to lay out and truncate its text.
 
@@ -30,9 +32,18 @@
                          constrainedSize:(CGSize)constrainedSize
               layoutManagerCreationBlock:(NSLayoutManager * (^)(void))layoutCreationBlock
                    layoutManagerDelegate:(id<NSLayoutManagerDelegate>)layoutManagerDelegate
-                textStorageCreationBlock:(NSTextStorage * (^)(NSAttributedString *attributedString))textStorageCreationBlock;
+                textStorageCreationBlock:(ASTextKitContextTextStorageCreationBlock)textStorageCreationBlock;
 
+/**
+ Set the constrained size for the text context.
+ */
 @property (nonatomic, assign, readwrite) CGSize constrainedSize;
+
+/**
+ Resets the text storage to the original value in case it was truncated before. This method is called within
+ a locked context.
+ */
+- (void)resetTextStorage;
 
 /**
  All operations on TextKit values MUST occur within this locked context.  Simultaneous access (even non-mutative) to
--- a/AsyncDisplayKit/TextKit/ASTextKitContext.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitContext.mm
@@ -21,7 +21,11 @@
   NSLayoutManager *_layoutManager;
   NSTextStorage *_textStorage;
   NSTextContainer *_textContainer;
+  
+  NSAttributedString *_attributedString;
 }
+
+#pragma mark - Lifecycle
 
 - (instancetype)initWithAttributedString:(NSAttributedString *)attributedString
                            lineBreakMode:(NSLineBreakMode)lineBreakMode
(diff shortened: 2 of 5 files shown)


maicki:
@appleguy @Adlai-Holler Can I get some look over it please

appleguy:
@nickvelloff is also an interested reviewer!  Thanks @maicki :)

AsyncDisplayKit/TextKit/ASTextKitContext.h
```diff
@@ -30,11 +32,20 @@
...
-                textStorageCreationBlock:(NSTextStorage * (^)(NSAttributedString *attributedString))textStorageCreationBlock;
+                textStorageCreationBlock:(ASTextKitContextTextStorageCreationBlock)textStorageCreationBlock;
 
+/**
+ Set the constrained size for the text context.
+ */
 @property (nonatomic, assign, readwrite) CGSize constrainedSize;
 
 /**
+ Resets the text storage to the original value in case it was truncated before. This method is called within
```
appleguy: Could you clarify in the comment what "the original value" means?  Is that the original, un-truncated string?  It seems like there might be a few possible things that could be reset (e.g. like clearing the calculated size).

AsyncDisplayKit/TextKit/ASTextKitContext.mm
```diff
@@ -58,6 +69,21 @@ - (instancetype)initWithAttributedString:(NSAttributedString *)attributedString
...
+
+- (void)resetTextStorage
+{
+  std::lock_guard<std::mutex> l(_textKitMutex);
+  [self _resetTextStorage];
+}
+
+- (void)_resetTextStorage
+{
+  [_textStorage setAttributedString:_attributedString];
```
appleguy: At this level, it is unclear why _textStorage would ever change to a different string...in other words, it is very strange / not obvious that the truncater may externally modify the text storage but not set it back.  We'd want to comment about this at the least, but I think the fix should probably go in the truncater instead.

AsyncDisplayKit/TextKit/ASTextKitTailTruncater.mm
```diff
@@ -152,9 +152,14 @@ - (NSUInteger)_findTruncationInsertionPointAtOrBeforeCharacterIndex:(NSUInteger)
 
 - (void)truncate
 {
+  // Reset the text storage to start always with the full string
+  [_context resetTextStorage];
```
appleguy: It seems to me that because the code below is responsible for modifying the string, this object should probably also be responsible for ensuring that it is working from the original string before truncating.  Can we have this object obtain the string it truncates from a delegate or something, probably the ASTextNode itself, so the data flow is always in the same direction / at the same level?

Maybe this is a bad idea.  This is largely an existing architectural problem, not just a problem with the solution - we're essentially talking about how to adjust the solution to be as clear as possible considering the architecture is a bit strange here.  Open to any suggestions, happy to land this without changes / just adding comments if you are confident it is the best structure.

AsyncDisplayKitTests/ASTextNodeTests.m
```diff
@@ -125,7 +125,7 @@ - (void)testRecalculationOfSizeIsSameAsOriginallyCalculatedSize
   for (NSInteger i = 10; i < 500; i += 50) {
     CGSize constrainedSize = CGSizeMake(i, i);
     CGSize calculatedSize = [_textNode measure:constrainedSize];
-    CGSize recalculatedSize = [_textNode measure:calculatedSize];
+    CGSize recalculatedSize = [_textNode measure:constrainedSize];
```
appleguy: Does this revised test fail before the fix?  That would be great!

nickvelloff:
@maicki  Please ping me if there is anything I can do to help. Happy to verify fix against my use case when you are ready.

maicki:
@nickvelloff Would be awesome if you could point your ASDK entry in your pod file to my fork and the branch with the fix and try it out on your side:

`pod 'AsyncDisplayKit', :git => 'https://github.com/maicki/AsyncDisplayKit.git', :branch => 'MSFixTextNodeTruncation'`

Please let me know if this would resolve your issue.

nickvelloff:
@maicki Unfortunately this did not resolve my issue.

The example i'm using is https://github.com/the-grid/TruncationBug-AsyncDisplayKitPlayground

It's a playground so updates or changes can be very quickly tested. It's updated to point to your branch.

Thanks a lot. Let me know if there is anything else you would like me to try.

maicki:
@nickvelloff Just tried your example out and it worked as expected. No truncation anymore if you scroll down after zooming. Worth checking out a fresh copy of your example and pod update as well as build ASDK from fresh before trying out the playground.

Furthermore I moved your playground example to a real project as it's not possible to debug within playgrounds so I had to create a project for it. I attached it to this comment. Just extract and run it.

[TruncationBug.zip](https://github.com/facebook/AsyncDisplayKit/files/355381/TruncationBug.zip)

nickvelloff:
@maicki I cleared derived data, and it worked perfectly. Can no longer reproduce. 
Also, I can't reproduce with the project you attached.

So for me this completely resolves the issue. Thanks for resolving!

appleguy:
@maicki @nickvelloff I had to revert this change on master; details are in an email.  There is a performance impact of this change and a possible thread safety issue (unit tests trigger the thread safety crash on my machine, but it is very timing sensitive because some debug options make it not happen, so it may not occur on your machine).

@nickvelloff feel free to continue pointing at the master commit that has the change as it is probably the better option for your app, until we get it re-landed next week after some analysis.

nickvelloff:
@appleguy Thanks for the heads up. Do let me know if there is any way I can assist.

PR #1864, opened 2016-07-08
[ASRangeController] Minimize number of registrations to node display notifications

### LINKED ISSUE #1460: ASRenderingEngineDidDisplayScheduledNodes notification posted, but another ASTableView catches it
Hello,

  I have a ViewController containing an ASTableView, which is on top of the stack. Working fine. When poping this ViewController, it is not deallocated instantly (later by the autoreleasepool), need to figure out why.

  But during this, I have the upfront ASTableView, but also still the not-yet dealloacted ASTableView behind. Scrolling will post the ASRenderingEngineDidDisplayScheduledNodes notification to the default NotificationCenter, where then both ViewControllers will catch them.
  One TableView has 1 section, the other more. So sometimes the wrong one gets the wrong amount of sections, which will cause a crash -[__NSArrayM objectAtIndex:]: index 1 beyond bounds [0 .. 0] in the ASFlowLayoutController findIndexPathAtDistance:fromIndexPath:] + 226.

  Any ideas ? 2 pop in my mind, add the reference to the table in the notification, so it may check if it is actually the needed receiver. Or a way to cancel the notification listener from the outside?

  Hope my case it is not too strange

  Thanks
  Tom

nguyenhuy:
Previously, `ASRangeController` registers to notification center every time a range update is performed, regardless of the number of nodes that need to schedule for a recursive display in response to the update. 

Instead, if this number is taken into account and range controller doesn't observe further notifications if it is zero, we can minimize the number of range updates executed and may fix #1460 (verification needed though).

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -245,9 +245,9 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   dispatch_once(&onceToken, ^{
     renderQueue = [[ASRunLoopQueue<ASDisplayNode *> alloc] initWithRunLoop:CFRunLoopGetMain()
                                                                 andHandler:^(ASDisplayNode * _Nonnull dequeuedItem, BOOL isQueueDrained) {
-      CFAbsoluteTime timestamp = isQueueDrained ?  CFAbsoluteTimeGetCurrent() : 0;
       [dequeuedItem _recursivelyTriggerDisplayAndBlock:NO];
       if (isQueueDrained) {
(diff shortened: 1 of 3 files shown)


appleguy:
@nguyenhuy thanks!!

maicki:
@nguyenhuy LGTM

Other thinking around that is, any opinion about looking into finding a way to not align on `NSNotifications` anymore and remove the `NSNotificationCenter` overhead?

Adlai-Holler:
Without interfering with Michael's question I'm going to merge this so that it's in the bag.

nguyenhuy:
@Adlai-Holler Thanks!

@maicki: Will keep thinking about it.

PR #1870, opened 2016-07-08
[ASTableView] Move setting of tuning parameters to the range controller

maicki:
Follow up PR for #1820.

To align ASTableView with ASCollectionView we set the tuning parameters on the range controller instead of the layout controller. The range controller basically just call's through to the layout controller though.


Adlai-Holler:
BOOM ⚡

PR #1871, opened 2016-07-08
[ASImageNode] Move dealloc method from ASImageNode+AnimatedImage category to ASImageNode

maicki:
Having a dealloc method in a category can be very problematic as the superclass dealloc method will not be called in case it exists.


garrettmoon:
Great catch! Thank you!

Adlai-Holler:
👌

PR #1872, opened 2016-07-08
[ASEnvironmentTraitCollection] Fixed bug where containerSize wasn't being saved

rcancro:
Summary:
The old assignment of `self.node.environmentTraitCollection.containerSize = windowSize;` doesn't because the struct creates a copy and then assigns `windowSize` to that copy.

Also realized that we need to create a new `ASEnvironmentTraitCollection` in `willTransitionToTraitCollection:withTransitionCoordinator`:. If we only update in `viewWillTransitionToSize:` we will only update the `containerSize` value in `self.environmentTraitCollection` as we don't have the new trait collection yet.

Differential Revision: https://phabricator.pinadmin.com/D101807


Adlai-Holler:
🐬

PR #1877, opened 2016-07-10
[ASTextNode, ASImageNode, ASVideoNode] Use ASDisplayNode base class lock for subclass property synchronization

hannahmbanana:


AsyncDisplayKit/ASImageNode.mm
```diff
@@ -110,17 +109,11 @@ - (instancetype)initWithViewBlock:(ASDisplayNodeViewBlock)viewBlock didLoadBlock
   return nil;
 }
 
-- (void)dealloc
```
appleguy: @johnepinterest do you remember why this was removed?  I think we'll add it back in.

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -134,11 +127,11 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 
 - (void)setImage:(UIImage *)image
 {
-  _imageLock.lock();
```
appleguy: I actually remember now why this method had to use .lock() and .unlock() manually.  It was to try to avoid interleaving with the superclass _propertyLock, which was grabbed inside invalidateCalculatedLayout and other methods below.

I think now that this is addressed, we should make this case a simple MutexLocker.  The else case that unlocks can be deleted.

AsyncDisplayKit/ASImageNode.mm
maicki: We need to add that back in! That was a refactor to remove the dealloc from the `ASImageNode` category and move it into `ASImageNode`

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -221,8 +214,8 @@ - (UIImage *)displayWithParameters:(id<NSObject> *)parameter isCancelled:(asdisp
   CGRect cropRect               = CGRectZero;
   asimagenode_modification_block_t imageModificationBlock;
 
+  ASDN::MutexLocker l(_propertyLock);
   {
-    ASDN::MutexLocker l(_imageLock);
```
appleguy: This is due to a recent change on master conflicting with the version John modified.  Let's put it back inside these braces, but still _propertyLock.

appleguy:
@hannahmbanana thank you for doing this!  There is a lot of careful stuff here, and this is a very important change that I didn't have time for tonight.

AsyncDisplayKit/ASImageNode.mm
johnepinterest: Although I haven't tracked it down exactly, I strongly suspect this is probably an artifact of the branch that I originally made these changes too (332ab23), which was current at the time.  Due to the nature in which we tried to get this change in (e.g., I sent you the file as-is directly), this context was lost as a result.

I believe the correct thing to do is remove this part of the diff - restore the `-dealloc` code removed - and verify that the additional `-dealloc` code that is "apparently" deleted here was in fact added as a result of a change / merged PR that happened some time after `332ab23`.

johnepinterest: Ok, tracked it down.  The `-dealloc` deletion by this commit is a mistake and is a result of loosing the context and commit that my original changes were applied to (332ab23).

Commit b2810edb0f52, which took place after 332ab23, is the commit that added this `-dealloc` code.

**_IMPORTANT_**: As @maicki says, this `-dealloc` code needs to go back in (if it isn't already).

For reference, the reason why this showed up as a delete is because I gave @appleguy a copy of `ASImageNode.mm` with my changes (relative to 332ab23) _directly_, and I suspect that @applyguy copied that file in to his repo at a "more current than 332ab23, and after b2810edb0f52" point in time.

hannahmbanana: @johnepinterest: `-dealloc` was added back in before it was merged. If that's the only difference then we are good.

PR #1883, opened 2016-07-11
[ASDisplayNode] Certain animations cause relayout before layout transitions can complete

jellenbogen:
This is a clone of my last PR which was closed because I screwed up a push in my fork. 

Original [https://github.com/facebook/AsyncDisplayKit/pull/1394]

To whomever reviews this, please let me know if I'm missing something here. I've been using the framework for a bit now, but there is a ton to learn here, thanks in advance.

There are certain animations that can cause a layoutSubviews or in this case layoutSubnodes to occur up the animating nodes parent chain. The 2 cases I've observed are frame and bounds.

This is causing the the node that is transitioning to snap to the finalLayout before all animations are complete and before completeTransition(didComplete: Bool) is called.

The proposed fix is to gate relayout by whether or not there is a transition occurring. I've added a check for _transitionContext != nil (which is set to nil on the transitionDidComplete callback) in the layout method on ASDisplayNode.


jellenbogen:
Here is the original conversation

[[ASDisplayNode] Certain animations cause relayout before layout transitions can complete by jellenbogen · Pull Request #1394 · facebook:AsyncDisplayKit.pdf](https://github.com/facebook/AsyncDisplayKit/files/356303/ASDisplayNode.Certain.animations.cause.relayout.before.layout.transitions.can.complete.by.jellenbogen.Pull.Request.1394.facebook.AsyncDisplayKit.pdf)

PR #1885, opened 2016-07-11
[ASTextNode] Refactor ASTextNode

maicki:
The main goal of the refactoring was to reduce state that we are currently have in the ASTextNode and all the Core Text components we use. In the ASTextNode we are holding on the renderer object what can result in weird behavior and we also need to be careful when to invalidate the renderer or pass new state down to the Core Text components.

The basic idea is that neither the ASTextNode nor layout should ever have a strong reference to the renderer object. This is to reduce memory load when loading thousands and thousands of text nodes into memory at once. Instead we maintain a LRU renderer cache that is queried via stack-allocated keys.

The renderer will be shared across all ASTextNode's as well as reused if possible. Currently the renderer cache has a an arbitrary size of 500 renderers and cleans up 100 (20% / 0.2) if the cost goes over that limit.

Addresses: #1841


AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1981,6 +2006,21 @@
...
 
 /* Begin PBXShellScriptBuildPhase section */
+		057DE97CCEF5219E324C2A17 /* 📦 Copy Pods Resources */ = {
+			isa = PBXShellScriptBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "📦 Copy Pods Resources";
```
appleguy: @maicki what version of Cocoapods is this?  The one that rewrote it to show [CP] is 1.0.1 which I think is the latest.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -11,32 +11,69 @@
...
 static const CGFloat ASTextNodeHighlightDarkOpacity = 0.22;
+
 static NSString *ASTextNodeTruncationTokenAttributeName = @"ASTextNodeTruncationAttribute";
 
+#pragma mark - ASTextKitRenderer
+
+static ASDK::TextKit::Renderer::Cache *sharedRendererCache()
+{
+  // This cache is sized arbitrarily
+  static ASDK::TextKit::Renderer::Cache *__rendererCache (new ASDK::TextKit::Renderer::Cache("ASTextKitRendererCache", 500, 0.2));
```
appleguy: Could you add comments on what the arbitrary size actually is?  Is 500 the number of objects?  What is 0.2?

There's a lot of great work here.  No doubt it would take some effort to tune this optimally, just wondering what the starting point is.

AsyncDisplayKit.xcodeproj/project.pbxproj
maicki: I already saw that :( I'm on 1.0. I'm currently in the process to revert the project changes that has nothing to do with the change.

AsyncDisplayKit/TextKit/ASFunctor.h
```diff
@@ -0,0 +1,59 @@
...
+#import <Foundation/Foundation.h>
+
+/* generic functors */
+
+namespace ASDK {
+  
+  template<class T>
+  struct DescribeFunctor {
+    NSString *operator()(const T &t) const {
+      return [NSString stringWithFormat:@"%d", static_cast<int>(t)];
```
appleguy: We can profile this, but I'm confident this will show up as inefficient.

The rest of this diff is quite beautiful in its use of efficient techniques :).  We should look at some kind of itoa implementation or find the code in the Pinterest app that does this more efficiently (I think Jon Parise wrote it)

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -118,22 +118,6 @@ - (CGSize)size
   return _calculatedSize;
 }
 
-- (void)setConstrainedSize:(CGSize)constrainedSize
```
appleguy: I would need to look back at the changelog, but I believe there was a performance reason why I added the ability to set the constrained size on an existing renderer rather than just recreating the renderer.  It may not be a showstopper though.

AsyncDisplayKit/TextKit/ASTextKitRendererCache.h
```diff
@@ -0,0 +1,152 @@
...
+namespace ASDK {
+  namespace TextKit {
+    /*
+     These are wrapper functions for CFNotificationCenter to call into the cache with low memory and backgrounding
+     notifications.
+     */
+    void lowMemoryNotificationHandler(CFNotificationCenterRef center, void *observer, CFStringRef name, const void *object, CFDictionaryRef userInfo);
+    void enteredBackgroundNotificationHandler(CFNotificationCenterRef center, void *observer, CFStringRef name, const void *object, CFDictionaryRef userInfo);
+
+    struct ApplicationObserver {
```
appleguy: Instead of doing stuff like this, could we make the ASTextKitRendererCache an Objective-C object for simplicity / familiarity of code, and it just contains the C++ Cache  object?  Or maybe this is the same as having the ASTextNode class subscribe to these notifications upon +initialize, and then have it manage the global Cache.

AsyncDisplayKit/TextKit/ASTextKitRendererCache.h
```diff
@@ -0,0 +1,152 @@
...
+                                        NULL,
+                                        CFNotificationSuspensionBehaviorDeliverImmediately);
+        CFNotificationCenterAddObserver(CFNotificationCenterGetLocalCenter(),
+                                        this,
+                                        &enteredBackgroundNotificationHandler,
+                                        (__bridge CFStringRef)UIApplicationDidEnterBackgroundNotification,
+                                        NULL,
+                                        CFNotificationSuspensionBehaviorDeliverImmediately);
+      };
+      ApplicationObserver() {
```
appleguy: Is this possibly missing a ~ / should be a destructor?

AsyncDisplayKit/ASTextNode.mm
maicki: Changed the comment and added it explicitly

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
maicki: Currently just for robustness we remove setting the constrain size explicitly as the lookup for a renderer in the cache should be pretty fast. We have to look at the performance though and optimize afterwards if necessary though.

AsyncDisplayKit/TextKit/ASTextKitRendererCache.h
maicki: We could do that.

AsyncDisplayKit/TextKit/ASFunctor.h
maicki: Addressed in latest commit

Adlai-Holler:
After discussion with @maicki, closing this for now. May reopen in the future.

PR #1886, opened 2016-07-11
[Layout Transition API] Simplify applying layout transition

maicki:
Simplify applying layout transition in preparation for bigger Layout Transition API work. Changes are:
- Always go through `_pendingLayoutTransition`
- Move from `[ASDisplayNode _applyLayout:layoutTransition:]` to `[ASDisplayNode _applyLayout:]` and `[ASDisplayNode _applyLayoutTransition:]`
- Change from `[ASLayoutTransition startTransition]` to `[ASLayoutTransition applyTransition]`

@levi @appleguy Can you review. This is the first small change in preparation for the bigger change that will come for the Layout Transition API

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -787,7 +787,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
       _pendingLayoutTransition = [[ASLayoutTransition alloc] initWithNode:self
                                                             pendingLayout:newLayout
                                                            previousLayout:previousLayout];
-      [_pendingLayoutTransition completeSubnodeInsertions];
+      [_pendingLayoutTransition applySubnodeInsertions];
 
       _transitionContext = [[_ASTransitionContext alloc] initWithAnimation:animated
                                                             layoutDelegate:_pendingLayoutTransition
@@ -887,7 +887,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (void)didCompleteLayoutTransition:(id<ASContextTransitioning>)context
 {
-  [_pendingLayoutTransition completeSubnodeRemovals];
+  [_pendingLayoutTransition applySubnodeRemovals];
   [self _completeLayoutCalculation];
   _pendingLayoutTransition = nil;
 }
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -867,7 +867,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
  */
 - (void)animateLayoutTransition:(id<ASContextTransitioning>)context
 {
-  [self __layoutSubnodes];
+  [self __layoutSublayouts];
   [context completeTransition:YES];
 }
 
@@ -901,7 +901,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
  */
 - (void)_completePendingLayoutTransition
 {
-  ASDN::MutexLocker l(_propertyLock);
+  ASDN::MutexLocker l(__instanceLock__);
   if (_pendingLayoutTransition) {
     [self setCalculatedLayout:_pendingLayoutTransition.pendingLayout];
     [self _completeLayoutTransition:_pendingLayoutTransition];
@@ -2118,7 +2118,7 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 
 - (void)setCalculatedLayout:(ASLayout *)layout
 {
-  ASDN::MutexLocker l(_propertyLock);
+  ASDN::MutexLocker l(__instanceLock__);
   
   ASDisplayNodeAssertTrue(layout.layoutableObject == self);
   ASDisplayNodeAssertTrue(layout.size.width >= 0.0);
@@ -2575,13 +2575,13 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 
     // Subnode insertions and removals need to happen always on the main thread if at least one subnode is already loaded
     ASPerformBlockOnMainThread(^{
-      [layoutTransition startTransition];
+      [layoutTransition commitTransition];
     });
     
     return;
   }
   
-  [layoutTransition startTransition];
+  [layoutTransition commitTransition];
 }
 
 - (void)layout
@@ -2604,7 +2604,6 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 
 #pragma mark - Display
 
->>>>>>> Simplify applying layout transition in preparation for bigger layout transition API work
 - (void)displayWillStart
 {
   ASDisplayNodeAssertMainThread();
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -642,12 +642,12 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 {
   ASDN::MutexLocker l(__instanceLock__);
   if (! [self shouldMeasureWithSizeRange:constrainedSize]) {
-    return _layout;
+    return _calculatedLayout;
   }
   
   [self cancelLayoutTransitionsInProgress];
 
-  ASLayout *previousLayout = _layout;
+  ASLayout *previousLayout = _calculatedLayout;
   ASLayout *newLayout = [self calculateLayoutThatFits:constrainedSize];
   
   _pendingLayoutTransition = [[ASLayoutTransition alloc] initWithNode:self
@@ -679,12 +679,12 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   // Only generate a new layout if:
   // - The current layout is dirty
   // - The passed constrained size is different than the layout's constrained size
-  return ([self _hasDirtyLayout] || !ASSizeRangeEqualToSizeRange(constrainedSize, _layout.constrainedSizeRange));
+  return ([self _hasDirtyLayout] || !ASSizeRangeEqualToSizeRange(constrainedSize, _calculatedLayout.constrainedSizeRange));
 }
 
 - (BOOL)_hasDirtyLayout
 {
-  return _layout == nil || _layout.isDirty;
+  return _calculatedLayout == nil || _calculatedLayout.isDirty;
 }
 
 - (ASLayoutableType)layoutableType
@@ -703,14 +703,14 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
                    shouldMeasureAsync:(BOOL)shouldMeasureAsync
                 measurementCompletion:(void(^)())completion
 {
-  if (_layout == nil) {
+  if (_calculatedLayout == nil) {
     // constrainedSizeRange returns a struct and is invalid to call on nil.
     // Defaulting to CGSizeZero can cause negative values in client layout code.
     return;
   }
   
   [self invalidateCalculatedLayout];
-  [self transitionLayoutWithSizeRange:_layout.constrainedSizeRange
+  [self transitionLayoutWithSizeRange:_calculatedLayout.constrainedSizeRange
                              animated:animated
                    shouldMeasureAsync:shouldMeasureAsync
                 measurementCompletion:completion];
@@ -772,7 +772,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
         return;
       }
       
-      ASLayout *previousLayout = _layout;
+      ASLayout *previousLayout = _calculatedLayout;
       [self setCalculatedLayout:newLayout];
       
       ASDisplayNodePerformBlockOnEverySubnode(self, ^(ASDisplayNode * _Nonnull node) {
@@ -944,7 +944,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   if (_placeholderEnabled && [self _displaysAsynchronously] && self.contents == nil) {
     
     // Zero-sized nodes do not require a placeholder.
-    CGSize layoutSize = (_layout ? _layout.size : CGSizeZero);
+    CGSize layoutSize = (_calculatedLayout ? _calculatedLayout.size : CGSizeZero);
     if (CGSizeEqualToSize(layoutSize, CGSizeZero)) {
       return;
     }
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -899,6 +890,51 @@ - (void)transitionContext:(_ASTransitionContext *)context didComplete:(BOOL)didC
...
+  ASDN::MutexLocker l(_propertyLock);
+  if (_pendingLayoutTransition) {
+    [self _applyLayout:_pendingLayoutTransition.pendingLayout];
+    [self _applyLayoutTransition:_pendingLayoutTransition];
+    _pendingLayoutTransition = nil;
+  }
+  [self _completeLayoutCalculation];
+}
+
+- (void)_applyLayout:(ASLayout *)layout
```
nguyenhuy: This method looks a lot like a setter. Should it be named `setCalculatedLayout`?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2475,6 +2476,7 @@ - (void)__layoutSublayouts
     ((ASDisplayNode *)subnodeLayout.layoutableObject).frame = [subnodeLayout frame];
   }
 }
+#pragma mark - Display
```
nguyenhuy: New line above this?

nguyenhuy:
Small questions aside, LGTM (although you didn't ask for my review haha) :P

levi:
I'd like to suggest a slight tweaking of the naming used in this diff.  Layouts can be "applied", as they are context objects awaiting usage and the meaning of the word here describes putting the layout into use. Transitions, however, represent a period of change between two layouts, resulting in either a completion or cancellation. Instead of _applying_ a transition, it would be more accurate to use a verb that defines the end result of this in-flight, transient event. Off the cuff "completeTransition" sounds like a good contender, but maybe there is something more meaningful that describes a transition successfully completing and for it to perform its borne operations?

maicki:
@nguyenhuy @levi I addressed both of you comments can you please give it another look

`_applyLayout:` to `_setCalculatedLayout:`: After looking over the `ASDisplayNode` API again we refer the `_layout` most of the time as `calculatedLayout` so it makes sense to actually call it as a setter method

`apply` to `complete`: I see the point of @levi and so let's go with `complete` although maybe `execute` could be a good candidate too. Curious what you guys think? @levi @appleguy

jellenbogen:
@maicki +1 for 'execute' unless whats actually happening is there are transitions in flight already and this is completing them, I think executeLayoutTransitions is the most descriptive

Adlai-Holler:
For subnode insertions and removals, I think `apply` is still the best name since they are done in one step just like updating the layout values.

Otherwise 👍

nguyenhuy:
👍

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -899,6 +890,51 @@ - (void)transitionContext:(_ASTransitionContext *)context didComplete:(BOOL)didC
...
+  ASDisplayNodeAssertTrue(layout.layoutableObject == self);
+  ASDisplayNodeAssertTrue(layout.size.width >= 0.0);
+  ASDisplayNodeAssertTrue(layout.size.height >= 0.0);
+  
+  _layout = layout;
+}
+
+- (void)_completeTransition:(ASLayoutTransition *)layoutTransition
+{
+  // Layout transition is not supported for non implicit hierarchy managed nodes
```
appleguy: Should this be the case?  Maybe we can assert if the items in the layout spec are not the same as previously, but I think we should support this if it is not a hassle.  As long as items are not added or deleted, it seems fine to use.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -899,6 +890,51 @@ - (void)transitionContext:(_ASTransitionContext *)context didComplete:(BOOL)didC
...
+
+  // Trampoline to the main thread if necessary
+  if (ASDisplayNodeThreadIsMain() == NO && layoutTransition.isSynchronous == NO) {
+
+    // Subnode insertions and removals need to happen always on the main thread if at least one subnode is already loaded
+    ASPerformBlockOnMainThread(^{
+      [layoutTransition completeTransition];
+    });
+    
+    return;
```
appleguy: This early return format is very confusing IMO.  The next line should be in an else clause, with no early return in this clause.

I'd suggest flipping the condition so that the case without the dispatch to main is first, in the if, and then the block dispatch one is in the else.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2427,41 +2463,6 @@ - (void)exitHierarchyState:(ASHierarchyState)hierarchyState
...
-
-  // Trampoline to the main thread if necessary
-  if (ASDisplayNodeThreadIsMain() == NO && layoutTransition.isSynchronous == NO) {
-
-    // Subnode insertions and removals need to happen always on the main thread if at least one subnode is already loaded
-    ASPerformBlockOnMainThread(^{
-      [layoutTransition startTransition];
-    });
-    
-    return;
```
appleguy: I see this was the old style, but we should update it.

AsyncDisplayKit/Private/ASLayoutTransition.h
```diff
@@ -47,7 +47,7 @@
 /**
  * Insert and remove subnodes that where added or removed between the previousLayout and the pendingLayout
```
appleguy: This doesn't make sense to me...don't added subnodes need to be put in at the beginning, but we have to wait until the end to remove the deleted ones?  How can both of this be handled in one call?  I guess by setting up a completion block?  Still, the premise of the method is confusing.

Is this basically commitTransition, like commitAnimations?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -899,6 +890,51 @@ - (void)transitionContext:(_ASTransitionContext *)context didComplete:(BOOL)didC
...
+
+- (void)_completePendingLayoutTransition
+{
+  ASDN::MutexLocker l(_propertyLock);
+  if (_pendingLayoutTransition) {
+    [self _setCalculatedLayout:_pendingLayoutTransition.pendingLayout];
+    [self _completeTransition:_pendingLayoutTransition];
+    _pendingLayoutTransition = nil;
+  }
+  [self _completeLayoutCalculation];
```
appleguy: I find the distinction between _completeTransition:, and _completeLayoutCalculation to be unclear and think these names need to be revised.

Also, _setCalculatedLayout: seems unusual to occur separately / before _completeLayoutCalculation.

Why does _completeLayoutCalculation occur even if _pendingLayoutTransition is nil?

AsyncDisplayKit/Private/ASLayoutTransition.h
levi: Yes, this is basically commitAnimations. @maicki, `commitTransition` seems like a winner.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -899,6 +890,51 @@ - (void)transitionContext:(_ASTransitionContext *)context didComplete:(BOOL)didC
...
+}
+
+- (void)_setCalculatedLayout:(ASLayout *)layout
```
levi: Can we just make this a setter? `- (void)setLayout:`

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -899,6 +890,51 @@ - (void)transitionContext:(_ASTransitionContext *)context didComplete:(BOOL)didC
...
+  ASDisplayNodeAssertTrue(layout.size.width >= 0.0);
+  ASDisplayNodeAssertTrue(layout.size.height >= 0.0);
+  
+  _layout = layout;
+}
+
+- (void)_completeTransition:(ASLayoutTransition *)layoutTransition
+{
+  // Layout transition is not supported for non implicit hierarchy managed nodes
+  if (layoutTransition == nil || self.usesImplicitHierarchyManagement == NO) {
```
levi: Probably don't need to check `self.usesImplicitHierarchyManagement` anymore, since it's just a feature flag and you won't be able to get to this point unless you're invoking `transitionLayout` on the node.

maicki: We need to check it here as in comparison to before we also create a layoutTransition of the node is not uses implicit hierarchy management internally

AsyncDisplayKit/ASDisplayNode.mm
maicki: Changed to if / else. and flipped the logic

AsyncDisplayKit/ASDisplayNode.mm
maicki: Will do

AsyncDisplayKit/Private/ASLayoutTransition.h
maicki: Going to use `commitTransition`.

AsyncDisplayKit/ASDisplayNode.mm
maicki: Unfortunately we cannot support it at the moment. This is one of the two problems we need to look into in the current implementation of implicit hierarchy management / layout transitions. Why we currently cannot support that is that we will run into an assert: A flattened layout must consist exclusively of node sublayouts.

What actually can happen is that in the previousLayout that is referenced in a layout transition the layoutable objects in the sublayouts can become nil. As the layoutable is weak in the layout it can happen that they are deallocated before the actual transition can finish. You can reproduce it very well in the ASDKGram example if you remove the `self.usesImplicitHierarchyManagement == NO` in this code path, rebuild, and just scroll down the feed. The reason in this particular example is that the we remove the Comment Nodes (ASTextNodes) from the supernodes and after that call ‘setNeedsLayout’. This has the effect that the Comment Nodes will be deallocated and become nil in the sublayouts of `pendingLayout` we use for the layout transition.

The solution would be to have some kind of managed nodes (ihm managed) / unmanaged notes (manual managed nodes) and dirty / undirty layout, so we can use this information in the layout transition to go from one hierarchy to another.

AsyncDisplayKit/ASDisplayNode.mm
maicki: Think we should stay with `_setCalculatedLayout`. If you currently look into `ASDisplayNode` as well as the public API around returning a layout we always to `_layout` as `calculatedLayout`, e.g.:  `calculatedLayout`, `calculatedLayoutDidChange` or `constrainedSizeForCalculatedLayout`

I think it would actually makes sense to change the `_layout` instance variable to rename to `_calculatedLayout`. So we would be consistence across the code base ... @levi @appleguy any opinions about that?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -798,33 +796,6 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
...
-  if (_placeholderEnabled && [self _displaysAsynchronously] && self.contents == nil) {
-    
-    // Zero-sized nodes do not require a placeholder.
-    CGSize layoutSize = (_layout ? _layout.size : CGSizeZero);
-    if (CGSizeEqualToSize(layoutSize, CGSizeZero)) {
-      return;
-    }
-
-    if (!_placeholderImage) {
-      _placeholderImage = [self placeholderImage];
```
appleguy: This used to be called after any layout calculation, but at least by the method name, now only after layout transitions? (_pendingLayoutTransitionDidComplete)

This seems unexpected, let me know @maicki

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -635,22 +635,15 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
   ASLayout *previousLayout = _layout;
   ASLayout *newLayout = [self calculateLayoutThatFits:constrainedSize];
   
-  if (ASHierarchyStateIncludesLayoutPending(_hierarchyState)) {
-    _pendingLayoutTransition = [[ASLayoutTransition alloc] initWithNode:self
-                                                          pendingLayout:newLayout
-                                                         previousLayout:previousLayout];
-  } else {
-    ASLayoutTransition *layoutTransition = nil;
-    if (self.usesImplicitHierarchyManagement) {
```
appleguy: @maicki could you tell me your thoughts here, I think this is OK because we want to enable it by default, but it's hard to tell if eliminating this condition will lead to the same behavior with today's implementation / client usage?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -764,11 +758,10 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
       }
       
       ASLayout *previousLayout = _layout;
-      [self _applyLayout:newLayout layoutTransition:nil];
+      [self setCalculatedLayout:newLayout];
```
levi: Is there still a need to call it `calculatedLayout` internally? Since the ivar is `_layout`, could we do `setLayout:`?

AsyncDisplayKit/ASDisplayNode.mm
maicki: It leads to the same behavior as before. The goal of this commit is to remove all of this if else clauses and hopefully simplify the code path to apply a layout transition to a ASDisplayNode. We use now in both cases the `_pendingLayoutTransition` variable for accomplishing applying a layout and layout transition to the node. The if / else clause from this place that is necessary is now in `_completePendingLayoutTransition` where we early return if the node is not hierarchy managed. But as you can see now, it's only one path for applying a transition to a node. After we enable hierarchy management by default the only thing we have to do is to return the early returns we currently have for not hierarchy managed nodes and don't have to remove any special code path's anymore.

AsyncDisplayKit/ASDisplayNode.mm
maicki: For me it seems inconsistence in two ways. We call it `calculatedLayout` thought our public API. There is no place where we call it only `layout`. As this is the case I actually think the right thing to do is to rename our instance variable from `_layout` to `_calculatedLayout`. Now we would be consistence throughout the class and could also call it `setCalculatedLayout:`.

Otherwise we can change it to `setLayout:` it makes totally sense in terms of setting an instance variable that's called `_layout`, but actually if you want to query it from the public API (the getter) we call it `calculatedLayout`. So the setter and getter don't have the same name than?? Calling the getter `- (ASLayout)layout` would not work  as we already have a method that is called `- (void)layout` but as you can see has a different return type and that's not possible in ObjC.

So to summarize I think the instance variable `_layout` is just a unfavorable name and we should rename that to `_calculatedLayout`?

@levi Does that make sense?

@nguyenhuy @appleguy @Adlai-Holler Do you guys have any opinion to rename the `_layout` instance variable to `_calculatedLayout` and add `setCalculatedLayout:` in ASDisplayNode

AsyncDisplayKit/ASDisplayNode.mm
maicki: Layout Calculation = Layout Transition now. Basically both go kind of the same way and finally call `_pendingLayoutTransitionDidComplete` that executes this code block. To show the different routes of layout transitions (1. Layout calculation / measurement and 2. Layout animation) I created this graph. Hope this makes it all a bit more clearer:

transition api flow

maicki:
Adding chart that should show the current flow for layout transitions. The goal of this PR is to simplify the code flow to execute a layout transition based on the current API and prepare us for the broader Layout Transition API changes.

transition api flow

Adlai-Holler:
Time for this one to leave the nest! Thanks @maicki you the man. Thanks everyone else for a great review.

PR #1892, opened 2016-07-11
[ASLayoutSpec] Final layoutable set children

rcancro:


hannahmbanana:
Restarted build. Failure = 
`The command "gem install cocoapods -v 1.0.1" failed and exited with 1 during .`

PR #1894, opened 2016-07-11
[ASDataController] Validate Client-Submitted Update Operations

Adlai-Holler:
This adds assertions like "attempt to delete item 10 in section 0 which only contained 5 items before the update" and "Invalid number of sections: count after the update (5) must equal the count before the update (2) plus or minus the number of sections inserted or deleted (4 inserted, 0 deleted)."

This also adds the identity of the data controller into the name of its editing queue, to help debugging.

This is ready for review! @appleguy @nguyenhuy @maicki @levi

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -273,13 +273,13 @@
 		9CFFC6BE1CCAC52B006A6476 /* ASEnvironment.mm in Sources */ = {isa = PBXBuildFile; fileRef = 9CFFC6BD1CCAC52B006A6476 /* ASEnvironment.mm */; };
 		9CFFC6C01CCAC73C006A6476 /* ASViewController.mm in Sources */ = {isa = PBXBuildFile; fileRef = 9CFFC6BF1CCAC73C006A6476 /* ASViewController.mm */; };
 		9CFFC6C21CCAC768006A6476 /* ASTableNode.mm in Sources */ = {isa = PBXBuildFile; fileRef = 9CFFC6C11CCAC768006A6476 /* ASTableNode.mm */; };
-		9F06E5CD1B4CAF4200F015D8 /* ASCollectionViewTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 9F06E5CC1B4CAF4200F015D8 /* ASCollectionViewTests.m */; };
+		9F06E5CD1B4CAF4200F015D8 /* ASCollectionViewTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 9F06E5CC1B4CAF4200F015D8 /* ASCollectionViewTests.mm */; };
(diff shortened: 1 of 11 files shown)


AsyncDisplayKit/Details/NSIndexSet+ASHelpers.m
```diff
@@ -66,7 +66,7 @@ - (NSString *)as_smallDescription
     if (range.length == 1) {
       [result appendFormat:@"%lu ", (unsigned long)range.location];
     } else {
-      [result appendFormat:@"%lu-%lu ", (unsigned long)range.location, (unsigned long)NSMaxRange(range)];
+      [result appendFormat:@"%lu-%lu ", (unsigned long)range.location, (unsigned long)(NSMaxRange(range) - 1)];
```
maicki: Use `%zu` so you don't have to cast it to `unsigned long`

AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
```diff
@@ -277,6 +301,83 @@ - (void)_sortAndCoalesceChangeArrays
...
+  // Assert that no sections were inserted out of bounds.
+  NSInteger invalidSectionInsert = NSNotFound;
+  if (newSectionCount == 0) {
+    invalidSectionInsert = [_insertedSections firstIndex];
+  } else {
+    invalidSectionInsert = [_insertedSections indexGreaterThanIndex:newSectionCount - 1];
+  }
+  ASDisplayNodeAssert(NSNotFound == invalidSectionInsert, @"Attempt to insert section %ld but there are only %ld sections after the update.", (long)invalidSectionInsert, (long)newSectionCount);
+  
+  [_oldItemCounts enumerateObjectsUsingBlock:^(NSNumber * _Nonnull oldItemCountObj, NSUInteger oldSection, BOOL * _Nonnull stop) {
```
maicki: Should we use fast iteration instead?

appleguy: Please do use fast enumeration whenever possible.  It’s specifically the retain and release overhead of the block that shows up in profiling.

maicki:
Small comments but overall looks good to me.

One thing I'm wondering for optimization purposes is if we can use a C++ vector instead of an NSMutableArray with NSNumbers for `_itemCountsFromDataSource`. Seems there can happen a lot of boxing and unboxing ... ?

AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
Adlai-Holler: @appleguy I thought in this case that the readability benefit was worth the performance cost but I'll change it.

Adlai-Holler:
@maicki Great points. Addressing…

hannahmbanana:
@Adlai-Holler: build error looks like something that needs to be fixed on your side

levi:
This is really awesome for debugging purposes. I suppose you've tested all of these cases, but I would suggest introducing a few unit tests to ensure these fail correctly. Otherwise, land away!

Adlai-Holler:
Thanks for suggesting those unit tests @levi I uncovered a false positive when reloading items. I'll address it tomorrow.

Adlai-Holler:
Sweet thanks for the reviews everyone!

PR #1896, opened 2016-07-11
[ASTableView] Add retainedLayer a la Collection View

Adlai-Holler:
We have reason to believe #1861 affects table view as well. This adds a retained layer pointer for the same reason.


maicki:
LGTM

Issue #1865, opened 2016-07-08
[ASButtonNode] Title is truncated in ASButtonNode


suomalainen9890:
I have a ASButtonNode which has static width and height. Originally the button node has a title, if later title of the button node changes to longer text, new title is truncated even though it takes not even half of the width of the button.

appleguy:
@suomalainen9890 thanks for reporting this!  It should be fixed by this PR: https://github.com/facebook/AsyncDisplayKit/pull/1863

Could you test it out locally?  If you're able to provide a test project, we can verify it for you.  The PR mentioned should land in the next day, and will be in the next release.  cc @maicki

suomalainen9890:
@appleguy : Thanks! Does it mean that I need to use this in the Podfile?
pod 'AsyncDisplayKit', :git => 'https://github.com/facebook/AsyncDisplayKit.git', :commit => '9d1e669'

I have never tried to do similar thing before.

maicki:
@suomalainen9890 Try to use this in your Podfile:

`pod 'AsyncDisplayKit', :git => 'https://github.com/maicki/AsyncDisplayKit.git', :branch => 'MSFixTextNodeTruncation'`

suomalainen9890:
@maicki : thanks! I will try.

suomalainen9890:
@maicki @appleguy It does not fix my problem. I will send a sample project here asap.

suomalainen9890:
@maicki @appleguy : This is sample project, less than 100 lines of code. I could be doing something wrong here. Please take a look: https://github.com/tungvoduc/ADKBugTest

maicki:
@suomalainen9890 Can you try out your example / main project with that fix: #1902 please. Thanks!

Furthermore in your example you have to move some code in the `ViewController.swift` otherwise the layout does fundamentally not really work:

```Swift
class ViewController: ASViewController {
    private let _scrollNode: ASScrollNode = ASScrollNode()
    private let _profileInfoNode = ProfileInfoNode()

    override func viewDidLoad() {
        super.viewDidLoad()
        
        self.node.addSubnode(_profileInfoNode)
    }
    
    override func viewDidLayoutSubviews() {
        super.viewDidLayoutSubviews()
        
        // Do any additional setup after loading the view, typically from a nib.
[...]

PR #1902, opened 2016-07-12
[ASDisplayNode] Fix setNeedsLayout triggered from subnode will not trigger relayout

maicki:
Current if `[ASDisplayNode invalidateCalculatedLayout]` and `[ASDisplayNode setNeedsLayout]` called in a subclass (see` [ASButtonNode setTitle:]`) the call actually does not trigger any relayout. It does not even bubble up the tree as it returns early in `[ASDisplayNode __setNeedsDisplay]`

After talking with @nguyenhuy about the initial intention why we are returning early there we found a better way for it.

This should resolve: #1865

@nguyenhuy @levi @appleguy Can I get a look over please

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1055,13 +1055,15 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (void)__setNeedsLayout
 {
   ASDisplayNodeAssertThreadAffinity(self);
+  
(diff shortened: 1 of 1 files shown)


levi:
👍👍👍

suomalainen9890:
@maicki : It does work now but not 100%. Sometimes the old bug happens, probability is around 1/20.

maicki:
@suomalainen9890 We are aware of it that this is not the full fix and @nguyenhuy will look into that probably later today / this week and follow up. Thanks for testing though!

PR #1903, opened 2016-07-12
[ASButtonNode] Correctly compare and update its title

nguyenhuy:
I found this interesting line while debugging #1865. We're basically trying to avoid unnecessary updates by comparing the old and existing titles. However, we should be comparing the contents of these strings instead of their memory addresses, no?


maicki:
LGTM

@appleguy  or @Adlai-Holler can you take a look over it too

Adlai-Holler:
Love it!

PR #1906, opened 2016-07-12
[Performance] Various Easy Performance Wins

Adlai-Holler:
I got on a bit of a rabbit trail here. A couple common tactics:

- Replace `enumerateIndexes` with `enumerateRanges`, using a for-loop to go through the indexes in each range. This reduces block invocations.
- Replace `.allKeys/.allValues` with plain dictionary enumeration and `objectEnumerator` when possible to avoid making copies. 
- Replace `enumerateObjectsUsingBlock` with fast enumeration when possible.
- Prefer `-[NSIndexPath indexPathForItem:inSection:]` because in 64-bit architectures there's some special mechanism to use singletons for these.
- Don't call the same thing 3 times in a row. Especially if that thing allocates memory!

### DIFF
--- a/AsyncDisplayKit/Details/ASCollectionDataController.mm
+++ b/AsyncDisplayKit/Details/ASCollectionDataController.mm
@@ -117,16 +117,18 @@
 - (void)willMoveSection:(NSInteger)section toSection:(NSInteger)newSection
 {
   for (NSString *kind in [self supplementaryKinds]) {
-    NSArray *indexPaths = ASIndexPathsForMultidimensionalArrayAtIndexSet([self editingNodesOfKind:kind], [NSIndexSet indexSetWithIndex:section]);
-    NSArray *nodes = ASFindElementsInMultidimensionalArrayAtIndexPaths([self editingNodesOfKind:kind], indexPaths);
+    NSMutableArray *editingNodes = [self editingNodesOfKind:kind];
+    NSArray *indexPaths = ASIndexPathsForMultidimensionalArrayAtIndexSet(editingNodes, [NSIndexSet indexSetWithIndex:section]);
+    NSArray *nodes = ASFindElementsInMultidimensionalArrayAtIndexPaths(editingNodes, indexPaths);
     [self deleteNodesOfKind:kind atIndexPaths:indexPaths completion:nil];
     
     // update the section of indexpaths
-    NSIndexPath *sectionIndexPath = [[NSIndexPath alloc] initWithIndex:newSection];
     NSMutableArray *updatedIndexPaths = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
-    [indexPaths enumerateObjectsUsingBlock:^(NSIndexPath *indexPath, NSUInteger idx, BOOL *stop) {
-      [updatedIndexPaths addObject:[sectionIndexPath indexPathByAddingIndex:[indexPath indexAtPosition:indexPath.length - 1]]];
-    }];
+    for (NSIndexPath *indexPath in indexPaths) {
+      NSUInteger newItem = [indexPath indexAtPosition:indexPath.length - 1];
+      NSIndexPath *mappedIndexPath = [NSIndexPath indexPathForItem:newItem inSection:newSection];
+      [updatedIndexPaths addObject:mappedIndexPath];
+    }
     [self insertNodes:nodes ofKind:kind atIndexPaths:indexPaths completion:nil];
   }
 }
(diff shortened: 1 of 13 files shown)


Adlai-Holler:
N.b. It's an open question whether `enumerateKeysAndObjects:` or fast enumeration + objectForKey: is faster when we want both the key and the object. We need to profile to see which one is better. So when that was the choice, I just left it alone for now.

Adlai-Holler:
Review plz @appleguy @maicki @levi

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -88,8 +88,7 @@ - (void)prepareForInsertSections:(NSIndexSet *)sections
 
 - (void)willInsertSections:(NSIndexSet *)sections
 {
-  NSArray *keys = _pendingContexts.allKeys;
-  for (NSString *kind in keys) {
+  for (NSString *kind in _pendingContexts) {
     NSMutableArray<ASIndexedNodeContext *> *contexts = _pendingContexts[kind];
```
maicki: I'm not pretty sure but can we actually use the block based enumeration api for NSDictionary here. As this would safe us a lookup.

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -143,15 +145,14 @@ - (void)prepareForInsertRowsAtIndexPaths:(NSArray<NSIndexPath *> *)indexPaths
 
 - (void)willInsertRowsAtIndexPaths:(NSArray<NSIndexPath *> *)indexPaths
 {
-  NSArray *keys = _pendingContexts.allKeys;
-  for (NSString *kind in keys) {
+  for (NSString *kind in _pendingContexts) {
     NSMutableArray<ASIndexedNodeContext *> *contexts = _pendingContexts[kind];
```
maicki: Same here for NSDictionary block based api

Adlai-Holler: It's true. Scott says it's hard to know which API would be faster and that he wants to profile them head-to-head, so for now I decided to make the smallest possible change.

AsyncDisplayKit/Private/_ASHierarchyChangeSet.m
```diff
@@ -314,8 +314,10 @@ + (void)sortAndCoalesceChanges:(NSMutableArray *)changes
   NSMutableIndexSet *allIndexes = [NSMutableIndexSet new];
   
   for (_ASHierarchySectionChange *change in changes) {
-    [change.indexSet enumerateIndexesUsingBlock:^(NSUInteger idx, __unused BOOL *stop) {
-      animationOptions[@(idx)] = @(change.animationOptions);
+    [change.indexSet enumerateRangesUsingBlock:^(NSRange range, BOOL * _Nonnull stop) {
+      for (NSUInteger i = range.location; i < NSMaxRange(range); i++) {
+        animationOptions[@(i)] = @(change.animationOptions);
```
maicki: What about changing animationOptions to a std::unordered_map? Do we expose it somehow?

AsyncDisplayKit/Details/ASCollectionDataController.mm
maicki: There should be a possible answer: https://www.objc.io/issues/7-foundation/collections/#enumeration-and-higher-order-messaging-2

maicki:
@Adlai-Holler couple of comments but otherwise looks good

AsyncDisplayKit/Details/ASCollectionDataController.mm
Adlai-Holler: Nice find! Through testing, I found that if you subclass NSDictionary then enumerateKeysAndObjects: will call __NSDictionaryEnumerate which calls objectForKey: for each key (as it is required to!). But if you don't subclass, the enumeration method uses its own fast path.

Adlai-Holler: I'm satisfied though, I'm going to migrate these over. I'm glad it turned out this way – Apple gave us a faster path. @appleguy

Adlai-Holler:
Thanks for the review @maicki! Comments addressed. Could I get one more set of eyes on this @levi ?

hannahmbanana:
Restarted the Travis build. Error:

ASVideoNodeTests
  testPlayerIsCreatedAsynchronouslyInFetchDataWithURL, failed: caught "NSInternalInconsistencyException", "Node should always be marked invisible before deallocating; interfaceState: 15, <ASThrashTestNode: 0x7fd191415b90>"
  /Users/travis/build/facebook/AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:367
  ```
  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating; interfaceState: %lu, %@", (unsigned long)_interfaceState, self);
  
  ```
	 Executed 308 tests, with 1 failure (1 unexpected) in 47.698 (48.212) seconds

AsyncDisplayKit/Private/ASLayoutTransition.mm
```diff
@@ -38,8 +38,8 @@ static inline BOOL ASLayoutCanTransitionAsynchronous(ASLayout *layout) {
     }
     
     // Add all sublayouts to process in next step
-    for (int i = 0; i < layout.sublayouts.count; i++) {
-      queue.push(layout.sublayouts[0]);
```
levi: wow, wtf. Did this work?

levi:
👍

AsyncDisplayKit/Private/ASLayoutTransition.mm
Adlai-Holler: Oh wow. Wow! I didn't even notice the 0 there before. Freebie!

PR #1907, opened 2016-07-13
[ASDisplayNode] Always layout nodes on a background thread

maicki:
Some crazy idea and follow up to #1839. We will push the layout of nodes to always be on a background thread as based on the work in #1839 we will trampoline the layout automatically to the main thread if at least one node is loaded.

Furthermore there needs to be some more improvements to it:
- Moving away from dispatching to global dispatch queues in layout transition and ASDataController
- Remove the semaphore in ASDataController
- Adding tests

@Adlai-Holler @nguyenhuy @appleguy @levi What do you guys think about that? We have to carefully think about that and check for any gotchas that could happen, if we really want to go forward with that.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -791,7 +791,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   };
 
   // TODO ihm: Can we always push the measure to the background thread and remove the parameter from the API?
-  if (shouldMeasureAsync) {
+  if (ASDisplayNodeThreadIsMain()) {
     ASPerformBlockOnBackgroundThread(transitionBlock);
   } else {
     transitionBlock();
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -184,13 +184,13 @@ NSString * const ASDataControllerRowNodeKind = @"_ASDataControllerRowNodeKind";
 
   NSUInteger nodeCount = contexts.count;
   NSMutableArray<ASCellNode *> *allocatedNodes = [NSMutableArray<ASCellNode *> arrayWithCapacity:nodeCount];
+  NSMutableArray<NSIndexPath *> *indexPaths = nil;
+  if (completionBlock) {
+    indexPaths = [NSMutableArray arrayWithCapacity:nodeCount];
+  }
 
   for (NSUInteger j = 0; j < nodeCount; j += kASDataControllerSizingCountPerProcessor) {
     NSInteger batchCount = MIN(kASDataControllerSizingCountPerProcessor, nodeCount - j);
-    
-    // Allocate nodes concurrently.
-    __block NSArray *subarrayOfContexts;
-    __block NSArray *subarrayOfNodes;
 
     __strong ASIndexedNodeContext **allocatedContextBuffer = (__strong ASIndexedNodeContext **)calloc(batchCount, sizeof(ASIndexedNodeContext *));
     __strong ASCellNode **allocatedNodeBuffer = (__strong ASCellNode **)calloc(batchCount, sizeof(ASCellNode *));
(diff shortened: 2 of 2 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -188,77 +184,46 @@ - (void)_layoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofK
...
-  // Block the _editingTransactionQueue from executing a new edit transaction until layout is done & _editingNodes array is updated.
-  dispatch_group_wait(layoutGroup, DISPATCH_TIME_FOREVER);
-  
-  if (completionBlock) {
-    NSMutableArray *indexPaths = [NSMutableArray arrayWithCapacity:nodeCount];
-    for (ASIndexedNodeContext *context in contexts) {
-      [indexPaths addObject:context.indexPath];
+    // Nil out buffer indexes to allow arc to free the stored cells.
+    for (int i = 0; i < batchCount; i++) {
+      [allocatedNodes addObject:allocatedNodeBuffer[i]];
```
Adlai-Holler: Let's replace `allocatedNodes` with an immutable array created right here like `NSArray *allocatedNodes = [NSArray arrayWithObjects:allocatedNodeBuffer count:batchCount]`

Adlai-Holler: We can do something like that with the index paths. I think remove allocatedContextBuffer and make allocatedIndexPathBuffer and then create an NSArray directly with that

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -791,7 +791,7 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
   };
 
   // TODO ihm: Can we always push the measure to the background thread and remove the parameter from the API?
-  if (shouldMeasureAsync) {
+  if (ASDisplayNodeThreadIsMain()) {
     ASPerformBlockOnBackgroundThread(transitionBlock);
```
appleguy: I believe this code already checks if it is on a background thread and performs it inline if so - maybe not?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -165,16 +165,12 @@ - (void)_layoutNodes:(NSArray<ASCellNode *> *)nodes fromContexts:(NSArray<ASInde
     return;
   }
   
-  // For any given layout pass that occurs, this method will be called at least twice, once on the main thread and
-  // the background, to result in complete coverage of both loaded and unloaded nodes
-  BOOL isMainThread = ASDisplayNodeThreadIsMain();
+  // Layout node on whatever thread we are on. We handle the trampoline to the main thread in case the node is
```
appleguy: Don't you mean ASDisplayNode (rather than We = ASDataController) handles the trampoline?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -187,78 +183,42 @@ - (void)_layoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts ofK
...
-    if (ASDisplayNodeThreadIsMain()) {
-      dispatch_semaphore_t sema = dispatch_semaphore_create(0);
-      dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        allocationBlock();
-        dispatch_semaphore_signal(sema);
-      });
-      dispatch_semaphore_wait(sema, DISPATCH_TIME_FOREVER);
+        
+      allocatedContextIndexPaths[k] = context.indexPath;
+      allocatedNodeBuffer[k] = node;
```
appleguy: later, seems like we don't need a separate node buffer if we have the contexts, if the context contains the allocated node (maybe there is a retain issue with this, not sure)

appleguy:
@maicki very impressive combination of code simplification **and** performance improvement!  Thank you!

PR #1911, opened 2016-07-13
[ASVideoPlayerNode] Expose the UIActivityIndicatorViewStyle to clients

gazreese:
The default spinner size looked a bit lost in my project, especially on iPad. The default is quite small and the 'large' version isn't exactly big either.

We're already exposing the tint colour, so it seemed only right to allow the user to set the spinner style also.

I've tested this on my own project and it appears to work fine.


maicki:
LGTM

Adlai-Holler:
Thank you @gazreese !

Issue #1905, opened 2016-07-12
[ASViewController] iOS 8 + UIWindow rotation problem


stephenkopylov:
Hi all!
The problem is really similar with this one: https://github.com/facebook/AsyncDisplayKit/issues/1810 (thanks for fixing it!)

Now issue appears with custom UIWindow with ASViewController as "rootViewController" and interface rotation - it not rotates again =)

Screens: 
![img_7375_2](https://cloud.githubusercontent.com/assets/5202281/16781912/6e0b06bc-488e-11e6-96bd-9c641646fff6.jpg)
![img_7375](https://cloud.githubusercontent.com/assets/5202281/16781918/71c6ad10-488e-11e6-922c-845f4a0c5442.jpg)

Sample project:
[UIWindow+AsyncVC.zip](https://github.com/facebook/AsyncDisplayKit/files/360187/UIWindow.AsyncVC.zip)

hannahmbanana:
@stephenkopylov: thanks for reporting! I will look into this today.

hannahmbanana:
@stephenkopylov: I had to call in the expert. @maicki should have a fix for you submitted as a PR in the next day!

stephenkopylov:
@hannahmbanana thank you!
Cant wait for fix =)

hannahmbanana:
@stephenkopylov: PR #1917 (will be merged to master by EOD) should fix this issue. I'm going to close the issue, but please reopen it if it doesn't resolve your issue. 

Thanks for reporting this and please let us know if you find any other bugs in the future!

PR #1917, opened 2016-07-14
[ASViewController] Fix sizing of root node in ASViewController for modal presentation and root view controller

maicki:
Adresses: #1905 and more robust check if ASViewController is modal presented that was fixed previously in regard to: #1810

### DIFF
--- a/AsyncDisplayKit/ASViewController.mm
+++ b/AsyncDisplayKit/ASViewController.mm
@@ -88,19 +88,7 @@
   [super viewWillLayoutSubviews];
   [_node measureWithSizeRange:[self nodeConstrainedSize]];
   
-  if (!AS_AT_LEAST_IOS9) {
-    [self _legacyHandleViewDidLayoutSubviews];
-  }
-}
-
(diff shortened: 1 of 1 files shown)


hannahmbanana:
@levi: this was a fix for an issue with the sign-in node and email account sign up in all iOS 8. I think we need to cherry pick to `releases/p6.7` as well.

Adlai-Holler:
I'm good with this 👍

AsyncDisplayKit/ASViewController.mm
```diff
@@ -212,9 +214,18 @@ - (ASSizeRange)_legacyConstrainedSize
   return ASSizeRangeMake(viewSize, viewSize);
 }
 
-- (ASInterfaceState)interfaceState
+- (void)_legacyHandleViewDidLayoutSubviewsIfNecessary
 {
-  return _node.interfaceState;
+  if (AS_AT_LEAST_IOS9) {
```
levi: Why not move this outside the method call to prevent a objcmsgsend?

levi:
👍

PR #1919, opened 2016-07-14
[ASPagerNode] Invalidate the pager’s layout when its trait collection changes

rcancro:
Is this something that I should have to be doing explicitly? If ASCollectioView (i.e., UICollectionView) does this automatically, shouldn’t the pager as well?

### DIFF
--- a/AsyncDisplayKit/ASPagerNode.m
+++ b/AsyncDisplayKit/ASPagerNode.m
@@ -157,4 +157,12 @@
   [self setDataSource:nil];
 }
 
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
@rcancro I agree this should already be handled by ASCollectionView =/. When ASCollectionView's bounds change, during the next `-layoutSubviews` it remeasures all nodes in a batch update (which should cause UICollectionView to invalidate the layout). Is that what this PR is intended to do?  Feel free to Slack me if the easiest repro is inside Pinterest.

PR #1920, opened 2016-07-14
[ASCollectionNode/ASTableNode] Remeasure cells on trait collection change

rcancro:
It isn’t enough to call `setNeedsLayout` on a cell since the container size will change on a trait collection change. We must call `measure` with the new container size as the constrainted size.

This is all happening on main, which I don’t like, but since we are reacting to a trait collection change does it make sense for it to happen anywhere else?

### DIFF
--- a/AsyncDisplayKit/Details/ASEnvironment.h
+++ b/AsyncDisplayKit/Details/ASEnvironment.h
@@ -141,8 +141,8 @@ ASDISPLAYNODE_EXTERN_C_END
   ASDN::MutexLocker l(lock);\
   ASEnvironmentTraitCollection oldTraits = self.environmentState.environmentTraitCollection;\
   [super setEnvironmentState:environmentState];\
-\
-   /* Extra Trait Collection Handling */\
+  \
+  /* Extra Trait Collection Handling */\
   /* If the node is not loaded  yet don't do anything as otherwise the access of the view will trigger a load*/\
   if (!self.isNodeLoaded) { return; } \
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
I'm not super familiar with this code, but is it correct to use the trait collection's container size for each node? Would it be enough instead to call `[dataController relayoutAllNodes]` just like in `-[ASCollectionView layoutSubviews]`? In fact, any idea why the size-change-handling code in that method isn't handling this?

rcancro:
As I understand, calling `setNeedsLayout` or `relayoutAllNodes` will layout using the cached constrained size. All of the nodes will be asked to layout and say: "Oh, my last layout was for this constrained size. Here use that one." Unless we call measure and pass in the new size, there is no way these nodes get a new constrained size and a new layout for this size.

I agree that using container size for the new constrained size is probably not right. Is there another way I could force a complete re-layout?

nguyenhuy:
`relayoutAllNodes` actually asks for a new constrained size and measures nodes using that size ([here](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASDataController.mm#L825) and [here](https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Details/ASDataController.mm#L845)). Note that the data source needs to be aware of the new size before `relayoutAllNodes` is called.

rcancro:
I tried `relayoutAllNodes` and the cells laid out properly. However, now the layout of the colletion view is wrong. It's like the collection view didn't get the new layout sizes of the cells. If I call measure on the cell then the CV layout is correct.

PR #1921, opened 2016-07-14
[ASRangeController] Fix Major Range Mode Updating Issues

Adlai-Holler:
This morning Michael and I found that our range mode updating is basically not working at all for a couple reasons.

- ASCollectionNode and ASTableNode declare their conformance in a category, which apparently doesn't work and causes the runtime to return NO for conformsToProtocol:
- Pinterest (and I'm sure other apps) have ASViewControllers with non-table, non-collection nodes. Sometimes these are plain old ASDisplayNodes. None of them conform to the update protocol.
- ASTableNode and ASCollectionNode did not implement +setRangeModeForMemoryWarnings: so they actually didn't conform to the updating protocol. This also means none of our users is calling that method or it would have crashed.
- Worst of all, we had no idea that none of this was working.

So

- Move the conformances in ASTableNode/ASCollectionNode to the main @interface instead of the category. This makes them return YES from `conformsToProtocol`.
- Allow ASViewController to conform to the range updating protocol as well as its node.
- Add a one-time-per-VC log if we try to adjust the range mode but can't because neither the VC nor the node conforms to the protocol.
- **BREAKING** Remove +setRangeModeForMemoryWarnings from the protocol and replace with a global method `+[ASDisplayNode setRangeModeForMemoryWarnings:]`.
  - I think we can land this even with the breaking change.
  - Obviously nobody was calling it or they would have gotten crashes.
  - It's not clear what calling it on, say, ASTableNode would have done. It would probably have called through to ASRangeController, which would have affected ASCollectionNode as well.
  - It would be a pain to override for every conforming node/VC.

@appleguy @maicki @levi This is pretty urgent. I think we should cherry-pick this to 6.7 if it's approved.

### DIFF
--- a/AsyncDisplayKit/ASCollectionNode.h
+++ b/AsyncDisplayKit/ASCollectionNode.h
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
  * ASCollectionNode is a node based class that wraps an ASCollectionView. It can be used
  * as a subnode of another node, and provide room for many (great) features and improvements later on.
  */
-@interface ASCollectionNode : ASDisplayNode
+@interface ASCollectionNode : ASDisplayNode <ASRangeControllerUpdateRangeProtocol>
 
 - (instancetype)initWithCollectionViewLayout:(UICollectionViewLayout *)layout;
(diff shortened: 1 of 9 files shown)


levi:
👍

maicki:
LGTM

PR #1924, opened 2016-07-14
[ASTextNode] Fix wrong truncation after constrained size changes

### LINKED ISSUE #1841: Truncated ASTextNode appearing in node that should not be truncated.

maicki:
Throw away the all subcomponents to create them with the new constrained size new as well as let the truncater do it's job again for the new constrained size. This is necessary as after a truncation did happen the context would use the truncated string and not the original string to truncate based on the new constrained size

This should fix #1841 and finally #1865

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -123,14 +123,14 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
   if (!CGSizeEqualToSize(constrainedSize, _constrainedSize)) {
     _sizeIsCalculated = NO;
     _constrainedSize = constrainedSize;
-    // If the context isn't created yet, it will be initialized with the appropriate size when next accessed.
-    if (_context || _fontSizeAdjuster) {
-      // If we're updating an existing context, make sure to use the same inset logic used during initialization.
-      // This codepath allows us to reuse the
-      CGSize shadowConstrainedSize = [[self shadower] insetSizeWithConstrainedSize:constrainedSize];
-      if (_context) _context.constrainedSize = shadowConstrainedSize;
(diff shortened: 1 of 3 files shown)


nguyenhuy:
LGTM! It's sad that we have to throw away these components but it's necessary to avoid nasty bugs that are hard to catch otherwise.

Adlai-Holler:
Love this, merge it at will!

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -123,14 +123,14 @@ - (void)setConstrainedSize:(CGSize)constrainedSize
   if (!CGSizeEqualToSize(constrainedSize, _constrainedSize)) {
     _sizeIsCalculated = NO;
```
appleguy: We should also reset _calculatedSize here, to CGSizeZero I guess.

PR #1925, opened 2016-07-15
Slight introduction tweak

levi:
Not everyone has had the pleasure of working with the notorious @appleguy, so the joke in the intro can be confusing.


maicki:
@hannahmbanana LGTM we should just merge it in whenever you are ready

Adlai-Holler:
True

PR #1929, opened 2016-07-15
[ASTextNode] use accessor for `pointSizeScaleFactors`

rcancro:
I have a subclass of `ASTextNode` that uses `pointSizeScaleFactors`. Currently I have to recompute the scale factors any time the font size changes. If `_rendererAttributes` used the property accesor for the scale factors I would only have to create them when asked. I hope this little change can make it in :)


maicki:
@rcancro LGTM. Can you add a comment why we use the property access explicitly in this case. Thanks!

levi:
👍 Game changer.

PR #1931, opened 2016-07-15
[ASCollectionView] When Resizing, Invalidate Layout After Remeasuring

Adlai-Holler:
This fixes rotation issues @rcancro and I have seen. It is still being tested.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -100,7 +100,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   id<ASCollectionViewLayoutFacilitatorProtocol> _layoutFacilitator;
   
   BOOL _performingBatchUpdates;
-  BOOL _superPerformingBatchUpdates;
+  NSUInteger _superBatchUpdateCount;
   NSMutableArray *_batchUpdateBlocks;
 
   BOOL _isDeallocating;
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
OK this is working super good! Could this get some inspection from @maicki @levi @appleguy

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -857,7 +857,7 @@ - (void)layoutSubviews
   _nextLayoutInvalidationStyle = ASCollectionViewInvalidationStyleNone;
   switch (invalidationStyle) {
     case ASCollectionViewInvalidationStyleWithAnimation:
-      if (!_superPerformingBatchUpdates) {
+      if (0 == _superBatchUpdateCount) {
```
maicki: @appleguy @Adlai-Holler What is the opinion about the code style? I personally would prefer `0 == _superBatchUpdateCount`, but we should find a way that works for all and the future

maicki:
One comment but otherwise looks good

hannahmbanana:
Assuming this passes user testing, it sounds like this should be cherry picked to 6.8

AsyncDisplayKit/ASCollectionView.mm
Adlai-Holler: For BOOLs I prefer `NO == _myBool` so you can immediately see the NO. For all other assignments this "defensive ordering" with constant on the left prevents people from accidentally using a single =, so I think we should standardize on that.

rcancro: FWIW, I'm pretty sure the compiler catches a single = for you now. 
screen shot 2016-07-15 at 4 19 30 pm

Adlai-Holler: Oh neat!

appleguy: @Adlai-Holler I understand the value in this style, but for now let's match the prevailing style, which is == NO

PR #1933, opened 2016-07-16
[ASDisplayNode+AsyncDisplay.mm] Refactor display block creation method for conciseness and readability.

appleguy:
@johnepinterest @maicki @Adlai-Holler @chrisdanford @hannahmbanana 

Hey all - this method had grown a bit over time, as features like instance method display were added, and it was time for cleanup.

I took time to do this carefully enough that I feel regressions are unlikely once it has been fully tested with the app, which I'll do over the weekend.  I am very familiar with the original state of the code.

I'd like to switch the order of the if / else so that rasterization comes second, but didn't want to move it for now so the diff has a slight element of readability lol.

This should improve efficiency by eliminating several lock acquisitions, and also adding a few strategically relevant isCancelled checks.  Particularly on the slowest devices with fast user input / large amounts of content moving (like on iPhone 4 with fast scrolling), I think this could make an incremental difference.

### DIFF
--- a/AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
+++ b/AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
@@ -75,6 +75,15 @@ static void __ASDisplayLayerDecrementConcurrentDisplayCount(BOOL displayIsAsync,
   }
 }
 
+#define DISPLAY_COUNT_INCREMENT() __ASDisplayLayerIncrementConcurrentDisplayCount(asynchronous, rasterizing);
+#define DISPLAY_COUNT_DECREMENT() __ASDisplayLayerDecrementConcurrentDisplayCount(asynchronous, rasterizing);
+#define CHECK_CANCELLED_AND_RETURN_NIL(expr)       if (isCancelledBlock()) { \
+                                                     expr; \
+                                                     __ASDisplayLayerDecrementConcurrentDisplayCount(asynchronous, rasterizing); \
+                                                     return nil; \
+                                                   } \
+
+
 - (NSObject *)drawParameters
 {
   if (_flags.implementsDrawParameters) {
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -179,141 +188,119 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
...
   asyncdisplaykit_async_transaction_operation_block_t displayBlock = nil;
   ASDisplayNodeFlags flags;
+  
   _propertyLock.lock();
+
   flags = _flags;
-  _propertyLock.unlock();
-  
-  ASDisplayNodeAssert(rasterizing || !(_hierarchyState & ASHierarchyStateRasterized), @"Rasterized descendants should never display unless being drawn into the rasterized container.");
+  CGRect bounds = self.bounds;
```
appleguy: I considered going straight to the layer for this, could do so I guess.

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -179,141 +188,119 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
...
+  CGFloat contentsScaleForDisplay = _contentsScaleForDisplay;
+  BOOL opaque = _layer.opaque;
+  BOOL shouldBeginRasterizing = (rasterizing == NO && flags.shouldRasterizeDescendants);
```
appleguy: The most important and subtle thing about this method is the difference between rasterizing and shouldBeginRasterizing.  This method may be called by itself with rasterizing == YES, sometime after it's initially called with rasterizing == NO but shouldBeginRasterizing == YES.

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -179,141 +188,119 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
...
-
+      
       UIImage *image = UIGraphicsGetImageFromCurrentImageContext();
       UIGraphicsEndImageContext();
 
-      __ASDisplayLayerDecrementConcurrentDisplayCount(asynchronous, rasterizing);
-
-      return image;
-    };
-  } else if (flags.implementsInstanceImageDisplay || flags.implementsImageDisplay) {
```
appleguy: I was able to entirely eliminate this third condition, and thus use just one else {

AsyncDisplayKit/Private/ASDisplayNode+AsyncDisplay.mm
```diff
@@ -179,141 +188,119 @@ - (void)_recursivelyRasterizeSelfAndSublayersWithIsCancelledBlock:(asdisplaynode
...
       if (currentContext && _willDisplayNodeContentWithRenderingContext) {
         _willDisplayNodeContentWithRenderingContext(currentContext);
       }
       
-      if (flags.implementsInstanceDrawRect) {
-        [self drawRect:bounds withParameters:drawParameters isCancelled:isCancelledBlock isRasterizing:rasterizing];
-      } else {
-        [[self class] drawRect:bounds withParameters:drawParameters isCancelled:isCancelledBlock isRasterizing:rasterizing];
+      // Decide if we use a class or instance method to draw or display.
+      id object = usesInstanceMethodDisplay ? self : [self class];
```
appleguy: This secretly makes me happy, which is probably bad.

PR #1939, opened 2016-07-16
[ASDisplayNode] Short circuit measure calls that have a zero-area constrainedSize.

appleguy:
// If the constrainedSize is completely zero-area, then there is no possibility for layout calculations to be successful.
  // This also avoids the issue of an inset being applied to 0, creating negative frame values.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -666,7 +666,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   
   // If the constrainedSize is completely zero-area, then there is no possibility for layout calculations to be successful.
   // This also avoids the issue of an inset being applied to 0, creating negative frame values.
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -664,6 +664,12 @@ - (BOOL)shouldMeasureWithSizeRange:(ASSizeRange)constrainedSize
     return NO;
   }
   
+  // If the constrainedSize is completely zero-area, then there is no possibility for layout calculations to be successful.
+  // This also avoids the issue of an inset being applied to 0, creating negative frame values.
+  if (constrainedSize.min.width * constrainedSize.min.height == 0.0 && constrainedSize.max.width * constrainedSize.min.width == 0.0) {
```
maicki: Should we assert this in this place? Is it more a api user problem or ASDK side?

maicki:
One comment but otherwise looks good to me

PR #1946, opened 2016-07-18
[ASDataController] Migrate Editing Queue to GCD for Performance/Debuggability

Adlai-Holler:
NSOperationQueue doesn't support backtrace recording, so it is impossible to tell where operations got enqueued from. 

This patch makes debugging data controller trouble a lot easier by migrating to GCD. I've tested this thoroughly. The only thing that we lose with this, is we lose a somewhat valuable assertion. @maicki @appleguy @levi Needs review

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -23,7 +23,11 @@
 //#define LOG(...) NSLog(__VA_ARGS__)
 #define LOG(...)
 
+#define ASSERT_ON_EDITING_QUEUE ASDisplayNodeAssertNotNil(dispatch_get_specific(&kASDataControllerEditingQueueKey), @"%@ must be called on the editing transaction queue.", NSStringFromSelector(_cmd))
+
 const static NSUInteger kASDataControllerSizingCountPerProcessor = 5;
+const static char * kASDataControllerEditingQueueKey = "kASDataControllerEditingQueueKey";
+const static char * kASDataControllerEditingQueueContext = "kASDataControllerEditingQueueKey";
 
 NSString * const ASDataControllerRowNodeKind = @"_ASDataControllerRowNodeKind";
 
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -450,13 +454,11 @@ - (void)waitUntilAllUpdatesAreCommitted
   // This should never be called in a batch update, return immediately therefore
   if (_batchUpdateCounter > 0) { return; }
   
-  [_editingTransactionQueue waitUntilAllOperationsAreFinished];
+  dispatch_sync(_editingTransactionQueue, ^{});
```
maicki: Do we introduce some possibility for a deadlock here?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -450,13 +454,11 @@ - (void)waitUntilAllUpdatesAreCommitted
...
   
   // Schedule block in main serial queue to wait until all operations are finished that are
   // where scheduled while waiting for the _editingTransactionQueue to finish
-  [_mainSerialQueue performBlockOnMainThread:^{
-    ASDisplayNodeAssert(_editingTransactionQueue.operationCount == 0, @"No operation should be in the _editingTransactionQueue anymore");
-  }];
+  [_mainSerialQueue performBlockOnMainThread:^{ }];
```
maicki: Any reason we removed the assert?

maicki:
Two small comments otherwise looks good.

AsyncDisplayKit/Details/ASDataController.mm
Adlai-Holler: No – the behavior here is exactly the same.

AsyncDisplayKit/Details/ASDataController.mm
Adlai-Holler: There is unfortunately no way to check whether a dispatch queue is empty =(

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -399,7 +403,7 @@ - (void)_reloadDataWithAnimationOptions:(ASDataControllerAnimationOptions)animat
   _initialReloadDataHasBeenCalled = YES;
   [self performEditCommandWithBlock:^{
     ASDisplayNodeAssertMainThread();
-    [_editingTransactionQueue waitUntilAllOperationsAreFinished];
+    dispatch_sync(_editingTransactionQueue, ^{});
```
levi: More descriptive way to do this so it's clear we're waiting for the queue to empty before proceeding? Macro?

levi: On second thoughts, can we wait without having to invoke an empty block?

Adlai-Holler:
Thanks @levi comments addressed. Could I get one more look over this before merge? @levi @appleguy ?

levi:
👍

PR #1950, opened 2016-07-19
Update ASVideoNode.mm

Xinchi:
Cherry-pick of https://phabricator.pinadmin.com/D102927 to upstream

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -123,11 +123,21 @@ static NSString * const kStatus = @"status";
 {
   ASDN::MutexLocker l(__instanceLock__);
 
+  AVPlayerItem *playerItem = nil;
   if (_asset != nil) {
-    return [[AVPlayerItem alloc] initWithAsset:_asset];
+    if ([_asset isKindOfClass:[AVURLAsset class]] && [self hasHLSAsset]) {
+      playerItem = [[AVPlayerItem alloc] initWithURL:((AVURLAsset *)_asset).URL];
+    } else {
+      playerItem = [[AVPlayerItem alloc] initWithAsset:_asset];
+    }
   }
 
-  return nil;
+  return playerItem;
+}
+
+- (BOOL)hasHLSAsset
+{
+  return _asset.tracks.count == 0;
 }
 
 - (void)prepareToPlayAsset:(AVAsset *)asset withKeys:(NSArray<NSString *> *)requestedKeys


Xinchi:
This Pull Request has been tested thoroughly in 6.5, 6.6 production, and 6.7 dev branch.  

Issue: 
Apple claims in the AVFoundationProgramming Guide that HLS videos can be constructed only through URL (quote: To create and prepare an HTTP live stream for playback. Initialize an instance of AVPlayerItem using the URL. (You cannot directly create an AVAsset instance to represent the media in an HTTP Live Stream.)), but later with iOS 4.3 release notes it claimed to bring updates to how the HLS videos should be initialized, which works with asset too. I’ve tested with both, and it looks like initializing with asset is buggy (as demonstrated in the videos above). The first initialization would be successful, however if you re-initialize it again within a few seconds, the asset won’t be loaded successfully. I am reverting this to constructing AVPlayerItem using URL for HLS videos, which solves the bug.

Here’s 2 video recordings of the playback with and without the diff:

AVPlayerItem construct from Asset for HLS video (aka without the diff): 
https://www.dropbox.com/s/5yqn9nvce2rga1m/HLS_constructed_with_asset.mov?dl=0

AVPlayerItem construct from URL for HLS video (with the diff):
https://www.dropbox.com/s/h2b59c2q1jnkb17/HLS_constructed_from_url.mov?dl=0

There is an issue with the current implementation: The internal .asset property for HLS video is different than the public asset property.  However this issue wouldn't affect the Pinterest use case now because we don't use the asset object on the application level.  I am working on a new PR to address this issue and it will be put up against master branch soon.  I am trying not to change the logic in this PR to make it consistent with 6.5 and 6.6, which has been tested on both simulator and device and works well.

hannahmbanana:
(from Max):   I would expect this to land *only* on ASDK's 6.7 branch. I am expecting to have another PR landed in ASDK master and cherry-picked in p6.8.

hannahmbanana:
More context: #1748

hannahmbanana:
@Xinchi: it looks like committing to a release branch doesn't trigger a Travis build. Could you run the tests locally to make sure this passes?

Xinchi:
@hannahmbanana Thanks for the tip!  We might have a master-branch solution that can be merged to releases/p6.7 and releases/p6.8 soon.  #1954 

If #1954 can be landed in the next few hours, I will close this PR

hannahmbanana:
@Xinchi: thanks for the update. I was wondering why we had two PRs.

Xinchi:
@hannahmbanana Closing this patch because #1954 has been landed on master.

appleguy:
@Xinchi thanks for making the newer version!

PR #1951, opened 2016-07-19
Fixes iOS 7 and iOS 8 wrong size in viewWillAppear: while a a rotation is happening

maicki:
Follow up fix for #1810 and #1905


hannahmbanana:
This was tested on iOS 9.3 & 8.4 on Pinterest (Log In, Sign Up, Add Pin) and with community member provided test cases #1810 and #1905.

Adlai-Holler:
LGTM

PR #1952, opened 2016-07-19
[ASVideoNode] don't temporarily clear the preview image in setAsset:

eanagel:
In my project I delay setting the `ASVideoNode`'s `asset` until I need to, setting only the preview image initially. Since `ASVideoNode` starts preloading as soon as you set the asset my guess is this isn't an uncommon use case.

This works fine, unfortunately there is a flicker because we currently clear the image (ultimately in `[super clearFetchedData];`) in `setAsset:`. This PR is a fix for this issue.

@maicki @appleguy I wasn't sure about the method name `__clearFetchedData` (technically __ is reserved.) Happy to change if you have something better in mind.

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -389,17 +389,20 @@ static NSString * const kStatus = @"status";
   }
 }
 
+- (void)__clearFetchedData
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  
+  self.player = nil;
+  self.currentItem = nil;
+}
+
 - (void)clearFetchedData
 {
   [super clearFetchedData];
-  
-  {
-    ASDN::MutexLocker l(__instanceLock__);
-
-    self.player = nil;
-    self.currentItem = nil;
-  }
+  [self __clearFetchedData];
 }
+
 
 - (void)visibleStateDidChange:(BOOL)isVisible
 {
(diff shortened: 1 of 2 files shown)


maicki:
@eanagel Hey, great having this PR! Can you try looking into the failing tests if you have some time? Tests should succeed before we can merge ... Thank you very much!

In terms of naming for our internal methods we tend to go to one underline in the method name like `_clearFetchedData`. Curious what @appleguy or @Adlai-Holler thinks about it.

eanagel:
@maicki so I looked a the test and I think the problem is the test ;) It looks like we are doing the following with `ASVideoNode`:

1. set invalid asset
2. set placeholder image
3. test A: videoNode.image == placeholder
4. set second invalid asset
5. test B: videoNode.image != placeholder

Test B now fails. It was succeeding previously because we used to clear the image briefly when setting an asset. Note that if we had swapped steps 1 and 2 the old code (pre this PR) would have failed ;)

Given that the image always displays the placeholder and shouldn't be effected by setting the asset (and especially an invalid one), the existing test doesn't make much sense. I went ahead and replaced it with a similar test that ensures that the image is not changed (cleared) when setting an asset, which is what I'm fixing.

Xinchi:
Hi @eanagel , can you rebase this patch?  Some recent changes in ASAsset passing and AVPlayerItem construction has been landed in the latest ASVideoNode.

Adlai-Holler:
@eanagel If this is still valuable, please feel free to reopen it.

PR #1954, opened 2016-07-19
Hls issue

Xinchi:
Closing #1748 and opening this Pull Request, as the base in #1748 was still the old commit and after rebasing, it leads to the longest PR.   

@gazreese @ejensen  @hannahmbanana @levi @maicki


Xinchi:
@chourobin Regarding your comment in #1600, 

> hey @gazreese, I'd just like to add that after further investigation, I believe it's no longer required to build the AVPlayerItem from the URL (for HLS). Hence, the check for tracks should no longer be needed.
> 
> Apparently the AVFoundation guide was out of date. In this release note -- under Enhancements for HTTP Live Streaming -- it states that all assets (HLS and non-HLS) may be constructed from an AVURLAsset. We just have to call -loadValuesAsynchronouslyForKeys:completionHandler: first before assigning it to a player item. I've been doing this outside of ASVideoNode for now but I would be happy to submit a PR to bring this async asset loading into ASVideoNode by default. I wanted to check in and see what you guys think.

Seems it's not entirely true that this issue has gone away in the AVFoundation, as @gazreese and I have observed the issue with HLS loading.  So this is what this PR is for.

Xinchi:
@gazreese Mind taking a look at this PR to see if it addressed the internal / public asset inconsistency issue you raised in #1600 ?

Xinchi:
@ejensen Mind taking a look at this PR to see if it addressed the asynchronous loading of track key that you mentioned in #1748 ?

Xinchi:
@gazreese @chourobin @ejensen @Eke 
Do you think it's better for ASDK to tell if the AVUrlAsset specified by the client a HLS vs non-HLS asset by checking the track property, or is it better to provide a separate API for the client to specify which to construct the playeritem with, asset or url?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -126,14 +126,25 @@ - (AVPlayerItem *)constructPlayerItem
...
-  return nil;
+  playerItem.videoComposition = _videoComposition;
+  playerItem.audioMix = _audioMix;
+  return playerItem;
+}
+
+- (BOOL)hasURLAsset
+{
+  // The array of AVAssetTrack objects available via the tracks property of an URL asset is typically empty for streaming-based media
+  return _asset.tracks.count == 0;
```
appleguy: @Xinchi Below, it looks like there is use of the safe / "correct" API to asynchronously load this value.  Can we instead use that form of request / completion handler to restructure how we construct the player item, so the player item is created "as quickly as possible" even if it first goes through a call to loadValues...completionBlock: and the bulk of the work is in the completion block?

gazreese:
Hi @Xinchi. Regarding construction I'm keener on having the explicit construction using either the asset or the URL if I'm honest. This is reflected in ASVideoPlayerNode as well as the standard iOS APIs.

As for giving this a try, I've raised it up on my internal tracker and will have a look at it when I've finished what I'm working on.

AsyncDisplayKit/ASVideoNode.mm
Xinchi: The constructPlayerItem method is called once the requestedKeys are loaded in the prepareToPlayAsset: withKeys: method.. not sure how we can make it faster than that?

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -41,6 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (BOOL)isPlaying;
 
 @property (nullable, nonatomic, strong, readwrite) AVAsset *asset;
+@property (nullable, atomic, strong, readwrite) NSURL *assetURL;
```
maicki: @Xinchi Any reason why this property is atomic?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -69,6 +69,7 @@ @interface ASVideoNode ()
   ASVideoNodePlayerState _playerState;
   
   AVAsset *_asset;
+  NSURL *_assetURL;
```
maicki: Do we need that property at all?

AsyncDisplayKit/ASVideoNode.h
Xinchi: Good catch! fixing..

AsyncDisplayKit/ASVideoNode.mm
Adlai-Holler: Where does this get set?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -126,14 +127,16 @@ - (AVPlayerItem *)constructPlayerItem
...
+  if (_assetURL != nil) {
+    playerItem = [[AVPlayerItem alloc] initWithURL:_assetURL];
+  } else if (_asset != nil) {
+    playerItem = [[AVPlayerItem alloc] initWithAsset:_asset];
   }
 
-  return nil;
+  playerItem.videoComposition = _videoComposition;
+  playerItem.audioMix = _audioMix;
+  return playerItem;
```
maicki: Why not just let the method as it is currently. Any reason we would throw away the asset to create the `AVPlayterItem` and not just use the _asset?

Xinchi: This is what this PR was created for - seems we need to create the playerItem using the URL instead of Asset for streaming-based videos.  Initializing with asset will sometimes cause the video to not load properly.

levi:
@Xinchi strong direction. Looks good to me!

maicki:
@Xinchi Looks good to me!!

giphy

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -437,6 +440,33 @@ - (void)setPlayerState:(ASVideoNodePlayerState)playerState
...
+      return;
+    }
+  }
+
+  [self clearFetchedData];
+
+  _asset = [AVURLAsset assetWithURL:assetURL];
+
+  [self setNeedsDataFetch];
+}
```
Adlai-Holler: It seems that `-assetURL` and `-setAssetURL:` are just convenience accessors for setting and retrieving the asset BY url. So couldn't we simplify with e.g.:

```objectivec
- (void)setAssetURL:(NSURL *)assetURL
{
   self.asset = [AVURLAsset assetWithURL:assetURL];
}
```

etc?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -126,14 +126,17 @@ - (AVPlayerItem *)constructPlayerItem
 {
   ASDN::MutexLocker l(__instanceLock__);
 
-  if (_asset != nil) {
-    AVPlayerItem *playerItem = [[AVPlayerItem alloc] initWithAsset:_asset];
-    playerItem.videoComposition = _videoComposition;
-    playerItem.audioMix = _audioMix;
-    return playerItem;
+  AVPlayerItem *playerItem = nil;
+  if (self.assetURL != nil) {
```
appleguy: @Xinchi I'm not sure this is correct - what if we already have an _asset, which was created from this assetURL?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -437,6 +440,24 @@ - (void)setPlayerState:(ASVideoNodePlayerState)playerState
   _playerState = playerState;
 }
 
+- (void)setAssetURL:(NSURL *)assetURL
+{
+  if (!ASObjectIsEqual(assetURL, self.assetURL)) {
+     self.asset = [AVURLAsset assetWithURL:assetURL];
```
appleguy: What if the player item has been created already?  Does the clearFetchedData / fetch data pass cause this to be updated correctly?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -437,6 +440,24 @@ - (void)setPlayerState:(ASVideoNodePlayerState)playerState
   _playerState = playerState;
 }
 
+- (void)setAssetURL:(NSURL *)assetURL
+{
+  if (!ASObjectIsEqual(assetURL, self.assetURL)) {
```
appleguy: Although it is probably not mandatory because the two methods (-assetURL, -setAsset:) each grab the lock - @Xinchi we should probably add a mutexlocker here because otherwise it is possible that in between the other two lock acquisitions, it could be yielded and that may be an issue with the "transactional" nature of an important write operation like this.

AsyncDisplayKit/ASVideoNode.mm
Xinchi: In this case you described, the method won't be invoked because this method is invoked in 3 cases:

1) fetch data is called from parent node 
2) fetch data is called from setting the AssetURL
3) fetch data is called from setting the asset. 

In #1, It shouldn't load the video asset more than once. 
In #2, if we already have an _asset, which was created from this assetURL, the setter will be returned without setting the Asset and invoking the player item construction. 
In #3, It's very unlikely that the application will set the assetURL first and then set the Asset.  @maicki and I have discussed how we can prevent users from setting both the asset and url, and @maicki suggested we should add clear documentation for our users.

AsyncDisplayKit/ASVideoNode.mm
Xinchi: Yes - setting the asset will invoke [self clearFetchedData] and [self setNeedsDataFetch];  cc @Adlai-Holler who proposed this approach.

AsyncDisplayKit/ASVideoNode.mm
Xinchi: Will add this in a follow-up PR

PR #1956, opened 2016-07-20
[ASCollectionView] Do Not Suppress Empty Updates

### LINKED ISSUE #1945: [ASCollectionView] Cells don't swap while reordering, after update to 1.9.81
Hi everyone!

I had reordering cells feature working well with ASCollectionView in my project. After upgrade to 1.9.81 it had stopped working. 

What I do is:
- Add UIPanGestureRecognizer to ASCollectionView
- on UIPanGestureRecognizer.state == UIGestureRecognizerStateChanged (when user is reordering cells) I make some calculations, change layout attributes a little bit and use ASCollectionView.performBatchUpdates:completion: with empty block to trigger relayout with animation.
```
    [self.collectionView performBatchUpdates:^{} completion:^(BOOL finished) {}];
```
- Result: Cells don't switch positions as they used to with previous version. :(

I checked commits that may cause this issue and found out that this is the one: https://github.com/facebook/AsyncDisplayKit/commit/c62a4d3e79e6a5cdaf4cf5ba14300ed3bdbf9fa9

I also checked if `invalidateLayout` method would work and it works, but there is no animation (as always). 

@Adlai-Holler and others: Do you have any idea how can I fix this and still use latest ASDK version? 
Thank you a lot!

Adlai-Holler:
This was a bit of premature optimization that I introduced in c62a4d3. Even if the user didn't tell us about any updates, a batch update can be useful for triggering an animated relayout of the collection view. 

This resolves #1945

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1080,7 +1080,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 - (void)rangeController:(ASRangeController *)rangeController didEndUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
 {
   ASDisplayNodeAssertMainThread();
-  NSUInteger numberOfUpdateBlocks = _batchUpdateBlocks.count;
   if (!self.asyncDataSource || _superIsPendingDataLoad) {
     if (completion) {
       completion(NO);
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
@maicki @levi Could I get a review on this beast?

levi:
👍

maicki:
LGTM

PR #1957, opened 2016-07-20
[ASWeakSet] Replace with NSHashTable

Adlai-Holler:
Just a bit of housekeeping. It was nice knowing you ASWeakSet.

metal_gear_-_big_boss_salute_ metal_gear_solid_3_snake_eater

Note: It would be tempting to replace a couple of our NSMutableSets with NSHashTables. Namely, `_cellsForVisibilityUpdates` and `_pendingDisplayNodes`. We don't really want to retain the cells/nodes in those cases, but according to https://www.objc.io/issues/7-foundation/collections/#nshashtable NSHashTable inserts are much slower than NSMutableSet inserts, so I opted to leave it. I believe in 99.9% of cases we won't retain those things materially longer using NSMutableSet.

Ready for review @levi @maicki @appleguy


maicki:
LGTM

AsyncDisplayKit/Private/ASPendingStateController.mm
```diff
@@ -65,8 +64,8 @@ - (void)flush
 {
   ASDisplayNodeAssertMainThread();
   _lock.lock();
-    ASWeakSet *dirtyNodes = _dirtyNodes;
-    _dirtyNodes = [[ASWeakSet alloc] init];
+    NSArray *dirtyNodes = [_dirtyNodes allObjects];
+    [_dirtyNodes removeAllObjects];
```
appleguy: I'm not actually sure this is faster than letting the object deallocate itself; plus it will create a memory high watermark (not that this is the biggest issue)

appleguy:
@Adlai-Holler This is pretty cool in terms of code deletion, but I'm a bit worried about this because I'm nearly certain that we had to create ASWeakSet because the count of NSHashTable is totally inaccurate when tracking weak objects.  E.g. they are not actually removed from the count until iterated.  There are comments talking about this in the original implementation I believe.

We shouldn't land this without some further discussion on that, and verifying the details of the important systems impacted by this.

Also, this is an API change.  I believe we are using ASWeakSet in Pinterest, and I have suggested it for use to a few community members.

maicki:
Here are some more details: http://stackoverflow.com/questions/29880121/nshashtable-count-incorrectly-reports-non-empty

Apparently there are "problems" with `NSHashTable`:
- Count returns the number of references currently in the hash table some of which may be null.
- Two call to allObjects can lead to objects being left in the autorelease pool
- The behavior is intended

appleguy:
Yeah, I think we should abandon this :(

I must say this is a bit ironic, since I'm pretty sure you discovered the count issue and implemented this class to fix it @Adlai-Holler :-D

+ (NSHashTable<ObjectType> *)weakObjectsHashTable NS_AVAILABLE(10_8, 6_0); // entries are not necessarily purged right away when the weak object is reclaimed

Adlai-Holler:
I still think this is worth keeping. Here is my rationale about these issues:

- `Count may be wrong`: I knew about this one, but NSMapTable – which powers ASWeakSet – behaves the same way. It uses a cached count. It's virtually impossible to have both O(1) `-count` and also zeroing weak references. Since we currently don't use the count I figured it wasn't worth the performance/maintenance cost. I should have mentioned it in the PR.
- `[Two] call to allObjects leads to objects being left in the autorelease pool`: I assume "Two" was meant to be "the." I don't think this is bad or even unexpected. The NSArray will have to retain the objects, so it's not surprising that they end up in the autorelease pool.
- `"Entries are not necessarily purged right away" comment`: I can confirm from running the ASWeakSet unit tests against NSHashTable that fast enumeration immediately stops returning deallocated objects even though count doesn't necessarily update. So I think this comment is just referring to count.

So if we needed to get the O(n) true count of a hash table, we could add a category method like `as_slowCount` that uses fast enumeration just like ASWeakSet does (but faster). If we needed to get an NSArray and we decide that I'm wrong and there _is_ a bug in NSHashTable's `allObjects` then we could add `as_allObjects` that also uses fast enumeration just like ASWeakSet (but faster).

So at the very least it's an obvious no-lose, solid-win proposition to modify ASWeakSet to be backed by NSHashTable.

But I think since we don't currently use our O(n) count method, and since `-[NSHashTable allObjects]` sounds like it behaves as expected, it's silly to have a separate class at all.

AsyncDisplayKit/Private/ASPendingStateController.mm
Adlai-Holler: Ah you're absolutely right. What was I thinking?

Adlai-Holler:
Ah about the API change – damn. No Pinterest access handy but since it's in the `/private/` folder of ASDK, I'd be willing to put in the Pinterest diff to migrate safely to hash table (including handling count if needed) or if there's concern about other users in the community we could deprecate the class.

Adlai-Holler:
@appleguy @maicki Ping

appleguy:
@Adlai-Holler do we not use the count?  I certainly use it regularly in debugging, although if you want to add a comparable function, that sounds good.

Honestly I'd prefer we keep ASWeakSet even if you want to swap out the implementation, because in comparison, the NSHashTable / NSMapTable classes and their initializers are obscure.  We shouldn't add our own primitive for everything, but particularly given the count pitfall, I think it is worth at least a tiny wrapper class.

Adlai-Holler:
Superseded by #1976

PR #1961, opened 2016-07-21
Adding documentation for assetURL in ASVideoNode, and adding a mutex lock for assetURL setter

Xinchi:


appleguy:
@Xinchi Great improvements, thank you!!

PR #1962, opened 2016-07-21
[ASVideoNode] Fix conditional that prevented local asset loading

levi:


maicki:
Ok looks fine to me

Xinchi:
Looking good to me!

PR #1965, opened 2016-07-21
[ASDisplayNode] Don't crash if inserting a nil node

maicki:
If trying to insert a nil node via `insertSubnode:atIndex:` likely a crash will happen. Add assert in case passing in a nil node or return in production.


Adlai-Holler:
Good change. Any insight on why we're trying to insert a nil subnode?

PR #1969, opened 2016-07-22
Add assert for calling ~layoutSpecThatFits: in ASDisplayNode without _layoutSpecBlock

samhsiung:
This assert will catch cases where an `ASDisplayNode` subclass overrides and calls `[super layoutSpecThatFits:]` which is not intended and will crash during async layout due to adding a nil layout object to an `NSArray`.


maicki:
LGTM

appleguy:
@samhsiung thanks for submitting this!  As the framework is currently implemented, this is a great idea.

@maicki what are your thoughts on automatic or easier "concatenation" of layout specs, so that including the super's layoutspec is trivial?  In that world, we probably want this to be a valid no-op instead.  We could return an empty ASLayoutSpec here for example, which even today may be a better default for users who are putting the super-returned layout into a static spec along with their own layout (the "manual" concatenation).

PR #1970, opened 2016-07-22
[General] Remove Dead Code, Check Optional Protocol Selector

Adlai-Holler:
I am working on the data controller teardown/cancellation and I found some dead code. I wanted to break it out into a separate PR. I also found that range controller is calling an optional protocol selector without checking if it responds.

I removed some optional editing methods from the ASLayoutController protocol. They aren't called anywhere, and not implemented by FlowLayoutController.

@maicki @levi @appleguy Quick review?


appleguy:
@Adlai-Holler looks great, thanks!  I recognize these from when I was rewriting ASRangeController.  They are from a time when more state was maintained internal to the range & layout controller, so it's great to remove them.

PR #1971, opened 2016-07-22
[ASMapNode] create the MKPinAnnotationView on the main thread

tomizimobile:
I noticed that pretty often my app's animations would be disabled. I tracked it down to where the `MKPinAnnotationView` is created, and moving that to the main thread fixes the issue.

### DIFF
--- a/AsyncDisplayKit/ASMapNode.mm
+++ b/AsyncDisplayKit/ASMapNode.mm
@@ -166,6 +166,41 @@
 
 #pragma mark - Snapshotter
 
+static BOOL _hasCreatedPinProperties;
+
++ (MKAnnotationView *)pin
+{
+  static dispatch_once_t onceToken;
+  static MKAnnotationView *__pin = nil;
+  dispatch_once(&onceToken, ^{
+    ASDisplayNodeAssertMainThread();
+    __pin = [[MKPinAnnotationView alloc] initWithAnnotation:nil reuseIdentifier:@""];
+  });
+  return __pin;
+}
+
++ (UIImage *)pinImage
+{
+  static dispatch_once_t onceToken;
+  static UIImage *__pinImage = nil;
+  dispatch_once(&onceToken, ^{
+    ASDisplayNodeAssertMainThread();
+    __pinImage = self.pin.image;
+  });
+  return __pinImage;
+}
+
++ (CGPoint)pinCenterOffset
+{
+  static dispatch_once_t onceToken;
+  static CGPoint __pinCenterOffset;
+  dispatch_once(&onceToken, ^{
+    ASDisplayNodeAssertMainThread();
+    __pinCenterOffset = self.pin.centerOffset;
+  });
+  return __pinCenterOffset;
+}
+
 - (void)takeSnapshot
 {
   // If our size is zero, we want to avoid calling a default sized snapshot. Set _snapshotAfterLayout to YES
@@ -187,7 +222,8 @@
   }
 
   __weak __typeof__(self) weakSelf = self;
-  [_snapshotter startWithQueue:dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0)
+  dispatch_queue_t snapshotQueue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
+  [_snapshotter startWithQueue:snapshotQueue
              completionHandler:^(MKMapSnapshot *snapshot, NSError *error) {
                  __typeof__(self) strongSelf = weakSelf;
                 if (!strongSelf) {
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASMapNode.mm
```diff
@@ -207,7 +207,10 @@ - (void)takeSnapshot
                     [image drawAtPoint:CGPointZero];
                     
                     // Get a standard annotation view pin. Future implementations should use a custom annotation image property.
-                    MKAnnotationView *pin = [[MKPinAnnotationView alloc] initWithAnnotation:nil reuseIdentifier:@""];
+                    __block MKAnnotationView *pin;
+                    dispatch_sync(dispatch_get_main_queue(), ^{
+                      pin = [[MKPinAnnotationView alloc] initWithAnnotation:nil reuseIdentifier:@""];
+                    });
                     UIImage *pinImage = pin.image;
                     CGSize pinSize = pin.bounds.size;
```
Adlai-Holler: Technically `pin.image` and `pin.bounds` and the pin's deallocation all need to happen on the main thread also. 

Can we use `static UIImage *pinImage` and `dispatch_once` inside the main thread block to only create one pin annotation view and capture its image? We could use `pinImage.size` to have thread-safe access to the size rather than the bounds of the view.

Adlai-Holler:
Great catch! One comment, seeing if we can go further while you're at it.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -207,7 +207,10 @@ - (void)takeSnapshot
                     [image drawAtPoint:CGPointZero];
                     
                     // Get a standard annotation view pin. Future implementations should use a custom annotation image property.
-                    MKAnnotationView *pin = [[MKPinAnnotationView alloc] initWithAnnotation:nil reuseIdentifier:@""];
+                    __block MKAnnotationView *pin;
+                    dispatch_sync(dispatch_get_main_queue(), ^{
```
appleguy: Unfortunately, dispatch_sync to the main thread is not allowed by the AsyncDisplayKit threading model, and can result in deadlocks if this node happens to be part of a hierarchy that undergoes synchronous layout or display.  The threading model allows for blocking the main thread on background threads (clearly this is to be avoided when possible, but is sometimes necessary for visual correctness) — but a consequence is that the reverse is never allowed.

Since -takeSnapshot is not itself synchronous, we should be able to (if necessary) dispatch_async to main and then back again.

If we made use of this image in all maps, I'd say we should capture it in +initialize, but since we don't always use this let's avoid that.

appleguy: It looks like we can check for this in setAnnotations:.  We probably want to create a +pinImage method, and in there, have a static UIImage *__pinImage = nil, then dispatch_once to get the image.  It should assert that inside the dispatch once is done on the main thread, but subsequent accesses could be off main if we can find a way to regulate that properly.

Seems doable, but definitely a) annoying to get right, and b) shitty that this code existed this way.  Thanks for pointing it out.

tomizimobile: Thanks for all the comments - I'm taking a look at fixing this right now. However, I did notice that `setAnnotations:` needs to run on the main thread at least in the case of `isLiveMap == true`, and I'm not doing that in my code since I'm setting the annotations during creation of my node. Is this intended to only be used on the main thread and missing documentation? Or should the `isLiveMap` portion be put inside of a `dispatch_async`?

george-gw:
Just a small note, I think we need to have #1890 in mind while we work with this, because ultimately we would probably want to have a custom pin in the static map that mirrors the live map.

AsyncDisplayKit/ASMapNode.mm
george-gw: I believe that only the `isLiveMap` portion needs to be run on the main thread.

tomizimobile:
@appleguy I added a commit that uses some stored properties related to pin drawing, and a flag that makes sure they're created on the main thread initially. I didn't end up creating the properties in `setAnnotations:` since we already check `if (annotations.count > 0)` in `takeSnapshot:` and it's easier to handle in one place.

AsyncDisplayKit/ASMapNode.mm
```diff
@@ -201,33 +237,55 @@ - (void)takeSnapshot
...
-                      CGPoint point = [snapshot pointForCoordinate:annotation.coordinate];
-                      if (CGRectContainsPoint(finalImageRect, point)) {
-                        CGPoint pinCenterOffset = pin.centerOffset;
-                        point.x -= pinSize.width / 2.0;
-                        point.y -= pinSize.height / 2.0;
-                        point.x += pinCenterOffset.x;
-                        point.y += pinCenterOffset.y;
-                        [pinImage drawAtPoint:point];
-                      }
+                    if (_hasCreatedPinProperties)
```
tomizimobile: @george-gw Good call on tracking #1890 - I think your code that calls a block could be added around here. We'd wanna make sure that it runs on the main thread, so we'd have to check the presence of the block here, and use it inside the `dispatch_async` later on.

keleftheriou:

Adlai-Holler:
@tomizimobile This diff was superseded by #1890 which should fix the same issue. Thanks for raising it, and let me know if your apps animations are still getting disabled!

PR #1974, opened 2016-07-23
ASImageNode backing store sharing for memory and CPU reduction

chrisdanford:
This change shares computed layer contents between instances of ASImageNodes that are visually identical.

## Motivation 
- Reduce total memory usage by not computing and retaining multiple, identical content images.
- The caching mechanism should not retain any UIImage (neither the source image nor the computed contents image) for longer than it had previously.  Therefore, in the worst case of a 0% cache hit rate, peak memory usage will be no worse than before (minus the size of the `ASWeakMap` itself, which is not substantial).

## Implementation Details
- ASImageNode's layer contents are cached, and the cache key is an`ASImageNodeContentsKey` which contains all data that is passed to draw calls that generate the contents bitmap.
- Contents are stored in `ASWeakMap`.  The `ASImageNode` must retain the `Entry` object returned by the map in order to keep that key/value pair in the map.  `ASImageNode` should retain the `Entry` at the point where it sets `layer.contents`, and should stop retaining the `Entry` when it clears `layer.contents`.

### Stats from the Pinterest app on an iPod 5G.
Steps: Scroll Home feed for 30s, open any Closeup, then side-swipe for 30s
- overall cache hit rate: 52%
- work added by caching mechanism: 1076.98 ms
- work computing contents avoided because of cache hits: 8447.94 ms

### DIFF
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -24,6 +24,10 @@
 
 #import "ASInternalHelpers.h"
 #import "ASEqualityHelpers.h"
+#import "ASEqualityHashHelpers.h"
+#import "ASWeakMap.h"
+
+#include <functional>
 
 struct ASImageNodeDrawParameters {
   BOOL opaque;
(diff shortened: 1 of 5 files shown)


facebook-github-bot-2:

facebook-github-bot-8:

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -295,25 +370,85 @@ - (UIImage *)displayWithParameters:(id<NSObject> *)parameter isCancelled:(asdisp
...
+    return nil;
+  }
+
+  {
+    ASDN::MutexLocker l(cacheLock);
+    return [cache setObject:contents forKey:key];
+  }
+}
+
++ (UIImage *)createContentsForkey:(ASImageNodeContentsKey *)key isCancelled:(asdisplaynode_iscancelled_block_t)isCancelled
```
Adlai-Holler: These two should have Key capitalized

Adlai-Holler:
This change is totally awesome! Love the specific comments about performance costs etc. I'm down to land this after the one tiny change. @appleguy @maicki @levi thoughts?

AsyncDisplayKit/ASImageNode.mm
chrisdanford: ✅

appleguy:
@chrisdanford Thanks a ton, this is a huge contribution and I'm very excited about it!

Of the cache overhead, where roughly is it distributed?  Could you upload a .zip of a profile with that flow?  Wondering how much is -[UIImage hash] vs other things.

appleguy:
@chrisdanford performance figures are very impressive here, the drop in overhead after switching away from @synchronized is significant enough to be an interesting lesson by itself.

@maicki @Adlai-Holler can one of you review this tomorrow and bias towards landing + future improvements if you don't see any issues?  Although there is key creation overhead here, and we want to work towards a more general solution that can be applied to ASTextNode or other nodes, a large bulk of the win is achieved by this patch (especially because sharing text node backing stores is less common than images).

In the pretty reasonable test flow cited, we're looking at a ~7500ms **net** drop with ~1000ms left for potential optimization, where likely any key system will still take ~500ms.  I think the overhead is low enough & the probability of enough cache hits in most apps (to pay for the overhead at least) is very high, to have this enabled by default.

AsyncDisplayKit/ASImageNode.mm
```diff
@@ -323,25 +458,27 @@ - (UIImage *)displayWithParameters:(id<NSObject> *)parameter isCancelled:(asdisp
   // Another option is to have ASDisplayNode+AsyncDisplay coordinate these cases, and share the decoded buffer.
   // Details tracked in https://github.com/facebook/AsyncDisplayKit/issues/1068
   
-  @synchronized(image) {
-    [image drawInRect:imageDrawRect];
+  @synchronized(key.image) {
+    [key.image drawInRect:key.imageDrawRect];
```
Adlai-Holler: In the future we could use associated objects to attach an NSLock to this image and use that instead, since it's allegedly so much faster http://perpendiculo.us/2009/09/synchronized-nslock-pthread-osspinlock-showdown-done-right/

AsyncDisplayKit/Private/ASWeakMap.h
```diff
@@ -0,0 +1,59 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+
+/**
+ * This class is used in conjunction with ASWeakMap.  Instances of this type are returned by an ASWeakMap,
+ * must retain this value for as long as they want the entry to exist in the map.
+ */
+@interface ASWeakMapEntry<Value> : NSObject
+
+@property (nonatomic, retain, readonly) Value value;
```
Adlai-Holler: Can we use the word `strong` here? IIRC `retain` as a property attribute is a holdover from MRR days.

Adlai-Holler:
This is so incredibly excellent! Thanks @chrisdanford!

PR #1975, opened 2016-07-24
Add recursive description debugging method to ASLayout

levi:
Incredibly helpful in debugging a layout tree. I can't believe we've gone this far without this!


appleguy:
@levi very nice indeed!  Really like this and feel we need more in PIModel classes too.

PR #1976, opened 2016-07-24
[ASWeakSet] Refactor to Use NSHashTable

Adlai-Holler:
Supersedes #1957 @appleguy @maicki


appleguy:
@Adlai-Holler this is really good!  I'm annoyed by all of the object forwarding as I'm sure was your intent in eliminating this, but the name is much nicer / more approachable and I think the count is still useful.

I suppose we could consider adding the count logic as a category on NSHashTable, but this seems great for now.

PR #1987, opened 2016-07-26
[ASDataController] Cancel Work if We Lose Data Source, Cleanup

Adlai-Holler:
- Set it up to cancel work if the data source goes away.
- Rename and clean up some methods
- Make `dataSource` readonly, so that once it's gone we can be sure it's time to stop doing anything
- Add a mechanism to see how much work we avoided with this change. I don't have a device handy right now but the results on the simulator aren't great. I'll test it with a device and report back. 

@appleguy I'd like to cancel laying out nodes on reload data, but I'm hesitating because I want your opinion about the risk involved with completion handlers. So if I call insertRowsAtIndexPaths: with 20 items, and in the completion handler I assume those 20 items are there, but I pre-empted it with reloadData, we should probably still call the completion handler but those items may not have been inserted. I think we should go for it in a separate PR, and if it ever becomes an issue deal with it at that time.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -87,7 +87,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 #pragma mark -
 #pragma mark ASCollectionView.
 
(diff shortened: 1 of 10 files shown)


Adlai-Holler:
Ready for review! @maicki @appleguy @levi

Adlai-Holler:
BTW testing shows we don't avoid that much work in practice on an iPhone 4 iOS 7 – 20/1000 nodes skipped. Still, it is working and this PR brings other benefits/simplifications.

maicki:
Nice 👍

PR #2008, opened 2016-07-29
[ASRangeController] Debug overlay to show the size and direction of display + fetchData ranges during scrolling.

hannahmbanana:
This code is a little rough (unnecessarily complex layout code), but it is only run when the debug option is turned on.  I've been sitting on this (previously closed) PR for a while, but wanted to get it posted since it is now rebased with master and works against Pinterest's code.

Working on a follow up which adopts layout specs.

image


AsyncDisplayKit/AsyncDisplayKit+Debug.m
```diff
@@ -191,3 +203,551 @@ - (void)drawEdgeIfClippedWithEdges:(UIRectEdge)rectEdge color:(UIColor *)color b
...
+  NSMutableArray *_rangeControllerViews;
+  NSInteger      _newControllerCount;
+  NSInteger      _removeControllerCount;
+  BOOL           _animating;
+}
+
++ (UIWindow *)keyWindow
+{
+  // hack to work around app extensions not having UIApplication...not sure of a better way to do this?
+  return [[NSClassFromString(@"UIApplication") sharedApplication] keyWindow];
```
hannahmbanana: @Adlai-Holler: how should I handle this? Or does it not matter because it's a debug tool?

hannahmbanana:
@Adlai-Holler: ready for review

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1065,6 +1065,11 @@ - (ASDisplayNode *)rangeController:(ASRangeController *)rangeController nodeAtIn
   return [_dataController nodeAtIndexPath:indexPath];
 }
 
+- (NSString *)nameForRangeControllerDataSource
+{
+  return self.asyncDataSource ? NSStringFromClass([self.asyncDataSource class]) : NSStringFromClass([self class]);
```
hannahmbanana: Known issue: ASPagerNode always returns ASPagerNodeProxy

AsyncDisplayKit/AsyncDisplayKit+Debug.m
Adlai-Holler: I think this is fine for now. If we have any issues, we could add some API like `setShouldShowRangeDebugOverlay:inWindow:` or something but this is fine.

Adlai-Holler:
This is incredibly cool! We are currently slated to remove the `name` field from ASDisplayNode, but things like this would be a great argument for renaming the field to e.g. `displayNodeName` or `debugName` so that we could show that when available.

Time to 🚢 !

PR #2009, opened 2016-07-29
[ASDimension] add assertion for percent values outside 0-1

hannahmbanana:
Until we merge #1767 (Adlai's percent -> fraction naming fix) it is a bit confusing whether the provided value should be 0-100 or 0-1.


hannahmbanana:
If it's easy to do, this could go into `releases/p6.9` as well

Adlai-Holler:
Ding!

PR #2012, opened 2016-07-29
[ASDataController] Always Process In Batches

Adlai-Holler:
This is a reduced version of #1894 . This resolves an issue where non-batched reloads, when decomposed into delete/inserts, would be executed separately one-after-another. This causes crashes in Pinterest 6.7.


Adlai-Holler:
@appleguy @maicki @levi I believe this is very low-risk, but it's also obviously very late in the game.

levi:
👍

PR #2014, opened 2016-07-31
[ASDisplayNode] Ensure that nil can never be returned from -measureWithSizeRange:

appleguy:
This can happen in rare cases when multiple relayouts occur while a transition is being measured.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -643,11 +643,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   ASDN::MutexLocker l(__instanceLock__);
   if (! [self shouldMeasureWithSizeRange:constrainedSize]) {
     ASDisplayNodeAssertNotNil(_calculatedLayout, @"-[ASDisplayNode measureWithSizeRange:] _layout should not be nil! %@", self);
-    if (_calculatedLayout != nil) {
-      return _calculatedLayout;
-    } else {
-      return [ASLayout layoutWithLayoutableObject:self constrainedSizeRange:constrainedSize size:CGSizeZero];
-    }
+    return _calculatedLayout ? : [ASLayout layoutWithLayoutableObject:self constrainedSizeRange:constrainedSize size:CGSizeZero];
   }
   
   [self cancelLayoutTransitionsInProgress];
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -642,7 +642,12 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
 {
   ASDN::MutexLocker l(__instanceLock__);
   if (! [self shouldMeasureWithSizeRange:constrainedSize]) {
-    return _calculatedLayout;
+    ASDisplayNodeAssertNotNil(_calculatedLayout, @"-[ASDisplayNode measureWithSizeRange:] _layout should not be nil! %@", self);
+    if (_calculatedLayout != nil) {
+      return _calculatedLayout;
+    } else {
+      return [ASLayout layoutWithLayoutableObject:self constrainedSizeRange:constrainedSize size:CGSizeZero];
+    }
```
maicki: Just a nitpick, but what about this style:
```objc
return _calculatedLayout ?: [ASLayout layoutWithLayoutableObject:self constrainedSizeRange:constrainedSize size:CGSizeZero];
```

AsyncDisplayKit/Private/ASStackUnpositionedLayout.mm
```diff
@@ -28,7 +28,13 @@
...
   const CGFloat childCrossMin = alignItems == ASStackLayoutAlignItemsStretch ? crossMin : 0;
   const ASSizeRange childSizeRange = directionSizeRange(style.direction, stackMin, stackMax, childCrossMin, crossMax);
-  return [child measureWithSizeRange:childSizeRange];
+  ASLayout *layout = [child measureWithSizeRange:childSizeRange];
+  ASDisplayNodeCAssertNotNil(layout, @"ASLayout returned from measureWithSizeRange: must not be nil: %@", child);
+  if (layout != nil) {
+    return layout;
+  } else {
+    return [ASLayout layoutWithLayoutableObject:child constrainedSizeRange:childSizeRange size:CGSizeZero];
+  }
```
maicki: Same as above ...

maicki:
Just two nitpicks otherwise looks good to me.

appleguy:
@maicki Thanks!  I'd originally done it that way, not sure why I decided to change.  Anyway, put it back to that way.  I do normally put a space like ? : - but we could start doing the other if you prefer.

PR #2018, opened 2016-08-01
[Layout Transition] Fix not propagating updating and propagating down the layout transition id if a subnode is added

maicki:
If a node was added to a supernode, the supernode could be in a layout pending state. All of the hierarchy state properties related to the transition need to be copied over as well as propagated down the subtree. This is especially important as with Implicit Hierarchy Management adding subnodes can happen while a transition is in fly


appleguy:
@maicki This seems like the right fix to me, we should land if we can get an accept from either Levi or Adlai.  Let's get the tests running.  Thanks for the key investigation that went on here to develop this fix!

maicki:
@levi @Adlai-Holler Can you please review. Thanks!

Adlai-Holler:
Looks great. Love how you moved the environment propagation in as well.

PR #2019, opened 2016-08-01
[ASTraitCollection] propagate window container size via ASTraitCollection in iOS7

rcancro:


Adlai-Holler:
This is cool! Sorry for the delay here

PR #2022, opened 2016-08-01
[Automatic Hierarchy Management] Rename IHM to automatic hierarchy management and move out of beta header

maicki:
Renames `Implicit Hierarchy Management` to `Automatic Hierarchy Management`:
- Deprecate `usesImplicitHierarchyManagement`
- New property: `automaticHierarchy`
- Move out of `ASDisplayNode` beta header
- Cleanup headers around `ASDisplayNode` beta header

### DIFF
--- a/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
+++ b/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
@@ -344,7 +344,7 @@ if (shouldApply) { _layer.layerProperty = (layerValueExpr); } else { ASDisplayNo
     // The node is loaded but we're not on main.
     // We will call [self __setNeedsLayout] when we apply
     // the pending state. We need to call it on main if the node is loaded
-    // to support implicit hierarchy management.
+    // to support automatic hierarchy management.
     [ASDisplayNodeGetPendingState(self) setNeedsLayout];
   } else {
     // The node is not loaded and we're not on main.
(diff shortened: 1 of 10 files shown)


appleguy:
automaticHierarchy is not complete (I think grammatically?)  Not sure the language reason why, but it doesn't sound quite finished - like it needs one more word after (Management is one that would complete it).

Should we call this automaticSubnodes?

One that I think we should really consider / try to brainstorm is drawing the clear connection to the layout spec.  layoutSpecifiedSubnodes or something ?

maicki:
@appleguy `automaticHierachy` was just a placeholder. I just pushed an update and the naming is `automaticallyManagesSubnodes`

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -344,7 +344,7 @@ - (void)setNeedsLayout
     // The node is loaded but we're not on main.
     // We will call [self __setNeedsLayout] when we apply
     // the pending state. We need to call it on main if the node is loaded
-    // to support implicit hierarchy management.
+    // to support automatic hierarchy management.
```
Adlai-Holler: Let's be extremely consistent here and use `to support automatic subnode management`

Adlai-Holler:
I'm good with this with the one note about terminology in documentation. Let's land this

hannahmbanana:
@maicki: would `.automaticSubnodeManagement` be apple than `.automaticallyManagesSubnodes`?

maicki:
@hannahmbanana No most of the API uses `automatically...`

maicki:
#2066 will replace this PR that was accidentally closed.

PR #2024, opened 2016-08-02
[ASVideoNode] Improve playerStatus behaviour and image generation

flovouin:
Hi guys,

I looked into how the `playerStatus` is handled in the `ASVideoNode`, and how the `asset` / `playerItem` are loaded. It seems that some unnecessary reloading is happening. For example, `fetchData` loads the `playerItem` if the state is not `ASVideoNodePlayerStateInitialLoading`, but what about if the `playerItem` is already loaded (paused for example)?

This brought me to make the following changes:
- `fetchData` only triggers loading when the state is `ASVideoNodePlayerStateUnknown`. If you think about it, all other states actually indicate that the item is currently loading or has been loaded, so no need to do that again.
- `ASVideoNodePlayerStateInitialLoading` is no longer used. I think it is a bit ambiguous. For example, if someone calls `pause` during the initial loading, the state is changed to `ASVideoNodePlayerStateLoading`. This does not affect the "initial loading" delegate methods, which stay valid. `ASVideoNodePlayerStateInitialLoading` could probably be marked deprecated.
- When clearing data, the state is put back to `ASVideoNodePlayerStateUnknown`. This makes more sense than keeping the old, no longer valid, state. "NotLoaded" or something like that would probably be a better name, but I didn't want to break the API.
- When calling `pause`, the state is only changed to `ASVideoNodePlayerStatePaused` if the player was currently playing. Otherwise we might lose the info that the item is currently loading.
- When receiving a `playbackLikelyToKeepUp` notification, make sure that the value is `true` when setting the `ASVideoNodePlayerStatePlaybackLikelyToKeepUpButNotPlaying` state. Actually I'm not sure of the usefulness of this state, maybe it could be removed altogether and use `ASVideoNodePlayerStateReadyToPlay` instead?
- _EDIT:_ When calling `play`, only change the state if it's not `Unknown`, as it doesn't make sense if the asset hasn't been loaded yet. Thanks maxsz on Slack for reporting this!

In my app, making these changes brought two improvements:
- The `ASVideoNode` no longer flickers when `fetchData` is called several times and the asset was already loaded.
- The video stays loaded and resumes playback at the correct time when the view reappears on the screen (for example when another view controller is popped from the navigation stack).

Please tell me what you think of these changes. I might have missed edge cases in the state handling, so feedback would be nice :-)

Also, there's another commit in this branch that simply make the `AVAssetImageGenerator` use the `videoComposition` if it exists, to make sure that the correct preview is generated. I had forgotten that when implementing the `videoComposition` and `audioMix` properties.

Cheers,

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -37,6 +37,7 @@ static void *ASVideoNodeContext = &ASVideoNodeContext;
 static NSString * const kPlaybackLikelyToKeepUpKey = @"playbackLikelyToKeepUp";
 static NSString * const kplaybackBufferEmpty = @"playbackBufferEmpty";
 static NSString * const kStatus = @"status";
+static NSString * const kRate = @"rate";
 
 @interface ASVideoNode ()
 {
@@ -208,6 +209,25 @@ static NSString * const kStatus = @"status";
   [notificationCenter removeObserver:self name:AVPlayerItemNewErrorLogEntryNotification object:playerItem];
 }
 
+- (void)addPlayerObservers:(AVPlayer *)player
+{
+  if (player == nil) {
+    return;
+  }
+
+  [player addObserver:self forKeyPath:kRate options:NSKeyValueObservingOptionNew context:ASVideoNodeContext];
+}
+
+- (void) removePlayerObservers:(AVPlayer *)player
+{
+  @try {
+    [player removeObserver:self forKeyPath:kRate context:ASVideoNodeContext];
+  }
+  @catch (NSException * __unused exception) {
+    NSLog(@"Unnecessary KVO removal");
+  }
+}
+
 - (void)layout
 {
   [super layout];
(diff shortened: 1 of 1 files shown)


hannahmbanana:
@Eke @ejensen @gazreese: any comments?

gazreese:
Regarding the following:

> When receiving a playbackLikelyToKeepUp notification, make sure that the value is true when setting the ASVideoNodePlayerStatePlaybackLikelyToKeepUpButNotPlaying state. Actually I'm not sure of the usefulness of this state, maybe it could be removed altogether and use ASVideoNodePlayerStateReadyToPlay instead?

To me the `ASVideoNodePlayerStateReadyToPlay` state should be where you are if the video is ready to play but the play button hasn't been pressed. I think re-using it for `ASVideoNodePlayerStatePlaybackLikelyToKeepUpButNotPlaying` could possibly be incorrect.

As for making the state handling a lot better, I'm really pleased to see this go in! I actually had a workaround in my client code for some of the state handling and was hoping to get around to looking at this myself at some point.

I make use of the states quite a bit, so I'll give this a try in my own app and let you know how I get on...

gazreese:
@flovouin @hannahmbanana  OK, so I just tried this with my code. It turns out that the workaround I had in the `videoPlayerNode:willChangeVideoNodeState:toVideoNodeState:` method is still needed as with the release version. This is the bug that I was going to look at but I haven't got around to it yet.

The workaround is as follows:

```
	// For some reason we can be in ready to play state even when we're already playing
	// But we need to captuire ASVideoNodePlayerStateReadyToPlay when the video loads to activate the play button
	if (toState == ASVideoNodePlayerStateReadyToPlay && self.videoPlayerNode.isPlaying == YES) {
		toState = ASVideoNodePlayerStatePlaying;
	}
```

What happens is the last state you get when you start playing the video is 'ready to play' rather than 'playing'. I translate from one to the other so that the following case statement in my code works fine.

This looks incorrect to me. I know this isn't your bug as such but if you're looking at the states we should probably cover this. What do you reckon?

I'll continue to test and let you know how I get on...

gazreese:
The states I get in the callback are as following.

Video starts loading...
- ASVideoNodePlayerStateLoading (isPlaying false)
- ASVideoNodePlayerStateReadyToPlay (isPlaying false)

Then when I click the play button to play it...
- ASVideoNodePlayerStatePlaybackLikelyToKeepUpButNotPlaying (isPlaying true)
- ASVideoNodePlayerStateReadyToPlay (isPlaying true)
- ASVideoNodePlayerStatePlaybackLikelyToKeepUpButNotPlaying (isPlaying true)
- ASVideoNodePlayerStateReadyToPlay (isPlaying true)

hannahmbanana:
@maicki @Adlai-Holler: FYI there is more context / discussion of this PR on #asvideonode in the Slack channel.

flovouin:
@gazreese Indeed, it seems that `ReadyToPlay` and `PlaybackLikelyToKeepUp` have clear different meanings.
I have to admit that I don't really use states in my project, so your feedback is definitely useful!
Yes, if I start tweaking the states, we might as well try to solve other problems. I'll take a look at different use cases. Any feedback / sample projects are welcome. :-)

The issue mentioned in #asvideonode was eventually fixed, but yes, more discussion is available on Slack.

gazreese:
@flovouin OK, sounds good.

Unfortunately I can't give you my code as a test but I expect you can reproduce the same callback states with the Video example app if you wish. If you're playing around with this states this sounds like a useful thing to try anyway.

gazreese:
I should say, this is using ASVideoPlayerNode and not ASVideoNode

flovouin:
@gazreese Yes, I was easily able to reproduce the behaviour you describe. Removing these unwanted state changes wasn't that hard. I also noticed that the same thing happens when the video finishes: `PlaybackLikelyToKeepUpButNotPlaying` can be set after the `Finished` state. This could be the intended behaviour but I'm not sure.

Also, most of the state changes happen through KVO. The state changes in `play` and `pause` seem a bit out of place. Maybe observing `rate` on the `AVPlayer` would provide a more accurate indication of whether the video is playing. I remember doing this a while back on another project.

gazreese:
@flovouin I was just playing around with it myself and had the following change:

```
- (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(NSDictionary *)change context:(void *)context
{

  if (object != _currentPlayerItem) {
    return;
  }

  if ([keyPath isEqualToString:kStatus]) {
    if ([change[NSKeyValueChangeNewKey] integerValue] == AVPlayerItemStatusReadyToPlay) {
		if (self.isPlaying) {
			self.playerState = ASVideoNodePlayerStatePlaying;
		} else {
			self.playerState = ASVideoNodePlayerStateReadyToPlay;
		}
```

Not sure whether this is correct but it seemed to work. I'm just checking the states for when the video finishes playing and I think you're right, we might need to observe the rate, which is also used in `isPlaying`.

flovouin:
@gazreese Yes, I have more or less the same code. I make the same check for the `PlaybackLikelyToKeepUp` notification, such that the state stays `Playing` when it needs to.

I experimented with what I was suggesting: using KVO on `rate` instead of changing the state in `play` and `pause`. It works great, the only noticeable change is that when the video finishes, the `rate` change is notified before `didPlayToEnd`. This results in `Playing -> Paused -> Finished` instead of just `Playing -> Finished`. Would that be a problem?

gazreese:
@flovouin That wouldn't be a problem for me, thanks!

If you would like me to test this just let me know the branch name on your repo and I'll give it a try.

flovouin:
@gazreese I just pushed a new commit on this branch that uses KVO on rate. It also removes unnecessary / incorrect state transitions. So far it works on a sample project provided by @maxsz, and on my project (although I don't make heavy use of the states). As mentioned in my very first post, in addition to fixing the states, it also prevents the `ASVideoNode` from unnecessarily reloading the `playerItem`, hence removing some flickering and "play-back-from-start" issues I was having.

To anyone who uses the `ASVideoNode`, please tell me if it breaks anything on your side, I'd be happy to look into it.

gazreese:
@flovouin Excellent, thanks! I'll have a test now...

gazreese:
@flovouin I just tested this in my app in various situations and the states and transitions between them look perfect to me. Thanks! 👍

flovouin:
The checks were failing because of something unrelated to this PR, so I simply rebased to trigger the travis build and now everything's fine.
I can confirm that this PR improves the behaviour of the `ASVideoNode` in my app. It seems to also work (and remove the need for a work around) for gazreese.
One thing I mentioned is that `ASVideoNodePlayerStateInitialLoading` could be marked deprecated as it is no longer used. I didn't do it just yet as I'm not sure if there's a guideline for this?
Always happy to get more feedback... or merge it if everything is fine with you guys :-)

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -208,6 +209,25 @@ - (void)removePlayerItemObservers:(AVPlayerItem *)playerItem
...
+}
+
+- (void) removePlayerObservers:(AVPlayer *)player
+{
+  @try {
+    [player removeObserver:self forKeyPath:kRate context:ASVideoNodeContext];
+  }
+  @catch (NSException * __unused exception) {
+    NSLog(@"Unnecessary KVO removal");
+  }
```
Adlai-Holler: Is this exception getting thrown? Technically this is safe in Objective-C++ but we should be aware of whether we're observing our player or not.

flovouin: No it's not getting thrown. To be honest I simply followed the example of [removePlayerItemObservers](https://github.com/flovouin/AsyncDisplayKit/blob/f68132c96db12516260755926dc690bee7f673d2/AsyncDisplayKit/ASVideoNode.mm#L194), but I've never seen either of these exceptions being thrown.

flovouin: Should I remove the try block? I have very little ObjC/++ experience so if you see anything that you don't like, it can certainly be changed!

gazreese: @flovouin @Adlai-Holler It looks to me like this is really a separate issue from the playerStatus behaviour improvements, as like flovouin says this pattern is followed elsewhere in the code.

Is this exception handling something we could raise as a new bug and deal with afterwards?

My project is dependent on these changes you see, so I'm keen to see these in mainline or even better a release if possible. I'm more than happy to help with testing.

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -291,29 +312,47 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
...
+        }
       }
-    }
-  } else if ([keyPath isEqualToString:kPlaybackLikelyToKeepUpKey]) {
-    self.playerState = ASVideoNodePlayerStatePlaybackLikelyToKeepUpButNotPlaying;
-    if (_shouldBePlaying && (_shouldAggressivelyRecoverFromStall || [change[NSKeyValueChangeNewKey] boolValue]) && ASInterfaceStateIncludesVisible(self.interfaceState)) {
-      if (self.playerState == ASVideoNodePlayerStateLoading && _delegateFlags.delegateVideoNodeDidRecoverFromStall) {
-        [_delegate videoNodeDidRecoverFromStall:self];
+    } else if ([keyPath isEqualToString:kPlaybackLikelyToKeepUpKey]) {
+      BOOL likelyToKeepUp = [change[NSKeyValueChangeNewKey] boolValue];
```
binl: One thing that I've noticed working with video is that this PlaybackLikelyToKeepUp property is not reliable when the video is fully loaded. it's always NO in that case.

binl:
R+

AsyncDisplayKit/ASVideoNode.mm
Adlai-Holler: Good to know. Filed Ship 181 for this

Adlai-Holler:
Thanks for the work, merging!

flovouin:
Awesome, thanks!

PR #2029, opened 2016-08-03
Because Pods and the Podfile.lock are not checked in, they should be …

garrettmoon:
…removed before installing. Otherwise the state is not consistant.


Adlai-Holler:
Boom!

PR #2031, opened 2016-08-04
[ASDisplayNode] Remove layoutSpecBlock

maicki:
The implementation of `layoutSpecBlock` has currently a couple of issues e.g. if a `layoutSpecBlock` is provided, `layoutSpecThatFits:` will not be called on that node. Furthermore it is make it even harder to resolve the overall issue that currently it's not possible to have multiple `layoutSpecThatFits:` implementation in a class hierarchy.

As it was just recently introduced and does not have that much of a public usage, I would propose to remove it for 2.0.

@appleguy @Adlai-Holler Let me know what you are thinking ...

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -146,9 +146,6 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @note This method should not be called directly outside of ASDisplayNode; use -measure: or -calculatedLayout instead.
  *
- * @warning Subclasses that implement -layoutSpecThatFits: must not also use .layoutSpecBlock. Doing so will trigger
- * an exception. A future version of the framework may support using both, calling them serially, with the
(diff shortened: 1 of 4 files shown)


levi:
@maicki room to bring back a version of #985?

Adlai-Holler:
I believe Google said one of their big complaints was having to create node classes just to specify layouts, and thats a pretty legitimate complaint so we shouldn't make this change without providing them an alternative.

I know of other apps that use this field, so I think we should keep it. We currently throw an exception (and tell them in documentation) if they set layoutSpecBlock and override the method so there should be no surprises.

levi:
@Adlai-Holler #985? *wink* *wink*

Adlai-Holler:
I'm open to #985 in the future, sure, but I think we should not do this diff here. Google just recently complained about having to create node classes just to implement layouts – a problem I had at Tripstr as well – and we shouldn't take this option away until we have a replacement option. @maicki @levi @appleguy @hannahmbanana ?

maicki:
Ok I'm on board with that

PR #2033, opened 2016-08-04
[ASDisplayNode] WIP Measure node before layout if bounds changed

maicki:
The goal of this PR is to automatically trigger a call to "measureWithSizeRange:" in case the bounds of a node did change and is different as the previous. This is necessary to kick creating an`ASLayout`. It's a no op if the current `ASLayout` has the same size as the new #bounds though.


Adlai-Holler:
LGTM. Let's see what is making the tests fail.

Adlai-Holler:
After some internal discussion, we've decided this behavior change isn't worth landing in light of a bigger, more important change to integrate our layout with UIKit's layout system which is in development.

PR #2034, opened 2016-08-04
[ASTableView/ASCollectionView] Enable Scrolling/Selecting Pending Items

Adlai-Holler:
This feature has been requested by Google and others. We could in the future make an asynchronous option that would enqueue this with other edit operations, but that would get messy quickly.

NOTE: It is a known issue that `-waitUntilAllUpdatesAreCommitted` has no effect if the initial reload hasn't been enqueued. Simply adding that line causes other issues that I don't have bandwidth to investigate right now, so this change will only work if the initial reload has already been started. Ship task 156 covers this.

Ready for review @maicki @levi @appleguy


maicki:
LGTM

PR #2035, opened 2016-08-04
Prevent API misuse at compile time

dmcrodrigues:
Using Objective-C attributes, in this case `unavailable`, we can hide unsupported APIs at compile time instead of detecting and warn about it at runtime with a set of asserts.

I think this is not only more safe but also more friendly to newcomers which are able to correctly identify the API that they should be using with each node.


facebook-github-bot-2:

facebook-github-bot-8:

Adlai-Holler:
This is awesome – thank you @dmcrodrigues!

PR #2038, opened 2016-08-05
[ASDisplayNode] Fix if pending layout will be deallocated

maicki:
The pending layout transition needs to stay alive at least until applySubnodeInsertions did finish execute as it can happen that with Implicit Hierarchy Management new nodes gonna be added that internally call setNeedsLayout what will invalidate and deallocate the transition in the middle of inserting nodes

Stack Trace:
[__pendingLayoutTransition__recursively_being_set_while_still_in_use_.m.zip](https://github.com/facebook/AsyncDisplayKit/files/404188/__pendingLayoutTransition__recursively_being_set_while_still_in_use_.m.zip)

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -789,9 +789,12 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
       }
       
       // Setup pending layout transition for animation
-      _pendingLayoutTransition = [[ASLayoutTransition alloc] initWithNode:self
-                                                            pendingLayout:newLayout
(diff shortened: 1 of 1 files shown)


maicki:
Adding sample project that show's the crash and has the fix integrated. Look into ASDisplayNode for `CRASH:` and `FIX:`

[ASDKLayoutPendingCrash.zip](https://github.com/facebook/AsyncDisplayKit/files/403121/ASDKLayoutPendingCrash.zip)

Adlai-Holler:
Awesome! Landing this. If this kind of thing happens often enough in other cases, we may need to use a macro or something to ensure that self is kept alive as long as the MutexLocker is.

PR #1747, opened 2016-06-14
[ASLayoutSpec] Make child and children a property

levi:
Nice dot syntax in obj-c!

PR #2041, opened 2016-08-05
[ASLayoutSpec] Make child and children a property

maicki:
Move to property for `child` and `children` in `ASLayoutSpec` for nice dot syntax


maicki:
@Adlai-Holler What comments changes should be made based on your comment in #1747 ?

PR #2042, opened 2016-08-07
[ASTextNode] Fix ASTextNode shadow is not rendering 

maicki:
We didn't pass through shadow properties via the renderer attributes. That meant that the renderer shadower didn't get those and so setting a shadow on the text node did not have any effect.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -1048,6 +1048,9 @@ static CGRect ASTextNodeAdjustRenderRectForShadowPadding(CGRect rendererRect, UI
     if (shadowColor != NULL) {
       CGColorRetain(shadowColor);
     }
+    if (_shadowColor != NULL) {
+      CGColorRelease(_shadowColor);
+    }
     _shadowColor = shadowColor;
     _cachedShadowUIColor = [UIColor colorWithCGColor:shadowColor];
     [self _invalidateRenderer];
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -1044,6 +1049,7 @@ - (void)setShadowColor:(CGColorRef)shadowColor
       CGColorRetain(shadowColor);
     }
     _shadowColor = shadowColor;
```
Adlai-Holler: Since you're already in this code, could you add `if (_shadowColor != NULL) { CGColorRelease(_shadowColor); }` so we stop leaking them 🚰

Adlai-Holler:
Love this, let's land it ASAP. Thoughts about comment @maicki ?

AsyncDisplayKit/ASTextNode.mm
maicki: On it

maicki:
@Adlai-Holler Fixed the memory leak ready to go

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -1042,8 +1047,10 @@ - (void)setShadowColor:(CGColorRef)shadowColor
   if (_shadowColor != shadowColor) {
     if (shadowColor != NULL) {
       CGColorRetain(shadowColor);
+      CGColorRelease(_shadowColor);
```
Adlai-Holler: Hmmm... I think we need to move this out of the if statement and wrap it in a new one: `if (_shadowColor != NULL)` because no matter what the new value is, if we have an old value we must release it

Adlai-Holler:
Sweet!

Issue #2040, opened 2016-08-05
[ASImageNode] Image blurry despite frame being exact size of image


GitGadd:
After creating an ASImageNode with an image of size 300x62 (600x124 when rasterized using screen scale) with default settings and then sizing it to the exact same size, I noticed that the rendered image was not sharp.

Stepping through the code in ASImageNode.mm I found that this seems to be caused by _ASSizeFillWithAspectRatio flooring the values after using the aspect ratio to compute the size.

In my case, the aspect ratio was computed to be 4.83870983 which in turn resulted in a 'scaledSizeForImage' of 600x123, instead of the expected 600x124.

Attaching the function I think is responsible to ease investigations. My guess would be that this function should be rounding rather than flooring to avoid the case of x.999 snapping to 0.

```
static CGSize _ASSizeFillWithAspectRatio(CGFloat sizeToScaleAspectRatio, CGSize destinationSize)
{
  CGFloat destinationAspectRatio = destinationSize.width / destinationSize.height;
  if (sizeToScaleAspectRatio > destinationAspectRatio) {
    return CGSizeMake(destinationSize.height * sizeToScaleAspectRatio, destinationSize.height);
  } else {
    return CGSizeMake(destinationSize.width, floorf(destinationSize.width / sizeToScaleAspectRatio));
  }
}
```

hannahmbanana:
@GitGadd: this does seem odd. @appleguy: any idea if there is a reason for this (original) code?

hannahmbanana:
Thanks @GitGadd for bringing this to our attention! It's merged into master now.

GitGadd:
Great, thanks for the quick follow up @hannahmbanana !

PR #2045, opened 2016-08-08
[ASImageNode] fix scaling math

hannahmbanana:
Quick fix pointed out by @GitGadd in issue #2040.

This code is original (circa 2014) and appears to be a mistake?


Adlai-Holler:
@hannahmbanana I'm cool with this but we could go even further and start using type-generic math to avoid losing precision with CGFloats. 

Could you do similar to #1994 and import the header and replace "roundf" with "round" and any other function that ends in f with its non-f variant? Also replace usages of `float` in that file with `CGFloat`.

Adlai-Holler:
Cool! Much better

PR #2047, opened 2016-08-09
[ASThread] Add SharedLocker and SharedUnlocker that uses a shared pointer for the mutex

maicki:
This ensures that the mutex is alive at least for the lifetime of `SharedLocker` / `SharedUnlocker`.
I moved `ASLayoutTransition` to use the new `SharedLocker` as first step.


Adlai-Holler:
Tested this and it works great! Once this lands we should revert #2038. I believe it should be landed to both 6.9 and 6.10

PR #2049, opened 2016-08-09
[Layout Transition API] Add support for custom layout transition coordinator

maicki:
- Add new protocol `ASLayoutTransitionCoordinator`
- Add property to `ASDisplayNode` called `layoutTransitionCoordinator`
- Add a default fade in / out transition coordinator for layout transitions: `ASFadeInOutLayoutTransitionCoordinator`

I'm not happy yet with the names of the transition coordinator we provide by default yet: `ASDefaultLayoutTransitionCoordinator` and `ASFadeInOutLayoutTransitionCoordinator` and open for any API improvement suggestions

cc @levi @Adlai-Holler @appleguy

### DIFF
--- a/AsyncDisplayKit/Private/ASLayoutTransition.mm
+++ b/AsyncDisplayKit/Private/ASLayoutTransition.mm
@@ -132,6 +132,11 @@ static inline BOOL ASLayoutCanTransitionAsynchronous(ASLayout *layout) {
 
 #pragma mark - _ASTransitionContextDelegate
 
+- (ASDisplayNode *)containerNode
+{
+    return self.node;
+}
+
 - (NSArray<ASDisplayNode *> *)currentSubnodesWithTransitionContext:(_ASTransitionContext *)context
 {
   ASDN::MutexLocker l(__instanceLock__);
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/Private/ASLayoutTransition.mm
```diff
@@ -226,3 +231,102 @@ static inline void findNodesInLayoutAtIndexesWithFilteredNodes(ASLayout *layout,
 }
 
 @end
+
+@implementation ASDefaultLayoutTransitionCoordinator
+
+- (void)animateLayoutTransition:(id<ASContextTransitioning>)context
+{
+  ASDisplayNode *node = [context containerNode];
```
levi: **Question** Why do we need a container node? Seems that there's no intermediate state where the transition needs a container (eg. UIViewController transitions), since it'll always be `self`.

levi:
Since the plan is to reform this API significantly in 2.X, I'd suggest not adding the TransitionCoordinator abstraction and simply add the animation by default into the ASDisplayNode `animateTransition` implementation. The animation's timing can be set via a property. It's not elegant, I know, but will give us a lot of room to introduce new properties and protocols in 2.x without concern of breaking existing APIs.

maicki:
@levi You think we should just provide the default behavior directly in `ASDisplayNode` without introducing the coordinator for now?

Would be fine with me for now any other thoughts @appleguy @Adlai-Holler ?

levi:
Yeah, I think adding it now will reduce design flexibility for us later. There's some cool stuff we can do with the iOS 10 animation API, that would make this very malleable.

maicki:
Closing for a simpler solution for now over at #2052

PR #2051, opened 2016-08-09
Revert "Fix if pending layout will be deallocated (#2038)"

Adlai-Holler:
Confirmed that #2047 fixes #2038 in the general case


Adlai-Holler:
@maicki @appleguy @levi

maicki:
👍

PR #2052, opened 2016-08-10
[Layout Transition] Add default fade in / out layout transition

maicki:
Provide default fade & scale transition for layout transition API

@levi As discussed over at #2049 this PR implements a simple solution for now. cc @Adlai-Holler @appleguy


Adlai-Holler:
@maicki This diff looks good to me. Could you add a simple example app that demonstrates the default transition behavior so we can easily test and modify this code?

maicki:
@Adlai-Holler Added an example project: ASDKLayoutTransition

Adlai-Holler:
This is great! Let's rename the proper to `defaultLayoutTransitionDuration` and update the comments to say this is the value we will use in the default code. That way users don't mistakenly think that they need to set/read this value in their custom transition implementation.

Once that's in, let's land this!

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -857,14 +859,93 @@ - (BOOL)_shouldAbortTransitionWithID:(int32_t)transitionID
...
+      }
+    }
+  }
+  
+  for (ASDisplayNode *insertedSubnode in insertedSubnodes) {
+    insertedSubnode.frame = [context finalFrameForNode:insertedSubnode];
+    insertedSubnode.alpha = 0;
+  }
+  
+  [UIView animateWithDuration:self.layoutTransitionDuration animations:^{
```
levi: Expose options, duration & delay.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -291,6 +291,10 @@ - (void)_initializeInstance
   
   _environmentState = ASEnvironmentStateMakeDefault();
   
+  _defaultLayoutTransitionDuration = 0.2;
+  _defaultLayoutTransitionDelay = 0.0;
+  _defaultLayoutTransitionOptions = NULL;
```
Adlai-Holler: nit: Use `kNilOptions` here, since NULL is for pointers

Adlai-Holler: Better yet, let's use `UIViewAnimationOptionBeginFromCurrentState` since that's most often what users want.

Adlai-Holler:
Good to merge at will @maicki

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: It is indeed

maicki: Let's add `kNilOptions` and `UIViewAnimationOptionBeginFromCurrentState` explicitly in here

maicki:
@Adlai-Holler @levi  Good point with the layout options. Updated!

Adlai-Holler:
Sweet!

PR #2053, opened 2016-08-10
[ASDataController] Bail on waitUntilUpdatesAreCommitted if batch updating

Adlai-Holler:
We should have already been doing this but ASDataController thought it knew what it was doing.


maicki:
LGTM

PR #2059, opened 2016-08-11
[Examples] Parse RFC 3339 date time strings around 7-8 times faster in ASDKGram

maicki:
Benchmarked RFC 3339 date time string parsing a bit and simple c parsing is around 7-8 times faster than using `NSDateFormatter`.


Adlai-Holler:
Sweet!

PR #2060, opened 2016-08-11
[_ASDisplayViewAccessiblity] Improve accessibility support

maicki:
- Improvements for nodes that position is changing dynamically (like in ASTableView / ASCollectionView)
- Performance improvements calculating accessibility elements
- Add compiler flag: ASDK_ACCESSIBILITY_DISABLE to disable custom accessibility code

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1733,7 +1733,6 @@ static NSInteger incrementIfFound(NSInteger i) {
   BOOL canUseViewAPI = !self.isLayerBacked && !subnode.isLayerBacked;
   if (canUseViewAPI) {
     [_view addSubview:subnode.view];
-    [(_ASDisplayView *)_view setAccessibleElements:nil];
   } else {
     // Disallow subviews in a layer-backed node
     ASDisplayNodeAssert(subnode.isLayerBacked, @"Cannot add a subview to a layer-backed node; only sublayers permitted.");
@@ -1789,7 +1788,6 @@ static NSInteger incrementIfFound(NSInteger i) {
         [layer removeFromSuperlayer];
       } else {
         [view removeFromSuperview];
-        [(_ASDisplayView *)_view setAccessibleElements:nil];
       }
     });
   }
@@ -3250,7 +3248,6 @@ static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
       [selfNode addSubnode:subnode];
     } else {
       [self addSubview:subnode.view];
-      [(_ASDisplayView *)self setAccessibleElements:nil];
     }
   }
 }
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1554,7 +1554,6 @@ static bool disableNotificationsForMovingBetweenParents(ASDisplayNode *from, ASD
         // If we can use view API, do. Due to an apple bug, -insertSubview:atIndex: actually wants a LAYER index, which we pass in
         if (canUseViewAPI && sublayerIndex != NSNotFound) {
           [_view insertSubview:subnode.view atIndex:sublayerIndex];
-          [(_ASDisplayView *)_view setAccessibleElements:nil];
         } else if (sublayerIndex != NSNotFound) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wconversion"
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1554,6 +1554,7 @@ static bool disableNotificationsForMovingBetweenParents(ASDisplayNode *from, ASD
         // If we can use view API, do. Due to an apple bug, -insertSubview:atIndex: actually wants a LAYER index, which we pass in
         if (canUseViewAPI && sublayerIndex != NSNotFound) {
           [_view insertSubview:subnode.view atIndex:sublayerIndex];
+          [(_ASDisplayView *)_view setAccessibleElements:nil];
         } else if (sublayerIndex != NSNotFound) {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wconversion"
@@ -1732,6 +1733,7 @@ static NSInteger incrementIfFound(NSInteger i) {
   BOOL canUseViewAPI = !self.isLayerBacked && !subnode.isLayerBacked;
   if (canUseViewAPI) {
     [_view addSubview:subnode.view];
+    [(_ASDisplayView *)_view setAccessibleElements:nil];
   } else {
     // Disallow subviews in a layer-backed node
     ASDisplayNodeAssert(subnode.isLayerBacked, @"Cannot add a subview to a layer-backed node; only sublayers permitted.");
@@ -1787,6 +1789,7 @@ static NSInteger incrementIfFound(NSInteger i) {
         [layer removeFromSuperlayer];
       } else {
         [view removeFromSuperview];
+        [(_ASDisplayView *)_view setAccessibleElements:nil];
       }
     });
   }
@@ -3247,6 +3250,7 @@ static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
       [selfNode addSubnode:subnode];
     } else {
       [self addSubview:subnode.view];
+      [(_ASDisplayView *)self setAccessibleElements:nil];
     }
   }
 }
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1456,6 +1456,7 @@ - (void)_insertSubnode:(ASDisplayNode *)subnode atSubnodeIndex:(NSInteger)subnod
         // If we can use view API, do. Due to an apple bug, -insertSubview:atIndex: actually wants a LAYER index, which we pass in
         if (canUseViewAPI && sublayerIndex != NSNotFound) {
           [_view insertSubview:subnode.view atIndex:sublayerIndex];
+          [(_ASDisplayView *)_view setAccessibleElements:nil];
```
appleguy: Should we put these in the view, using:

- (void)didAddSubview:(UIView *)subview;
- (void)willRemoveSubview:(UIView *)subview;

Instead?  Related question, does our current implementation work if the user directly adds a UIView to the hierarchy without a node attached to it?

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -14,6 +14,27 @@
 
 #pragma mark - UIAccessibilityElement
 
+// Sort accessiblity elements by x and y origin. First sort the elements by y and than by x value.
+static NSArray *SortAccessibilityElements(NSArray *accessibleElements) {
```
appleguy: This is great!  Should we use NSMutableArray?  Less functional, but probably more efficient.  If we did, we'd probably make this function return just the comparator block, and then call the sort method on the mutable array directly.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -50,59 +72,84 @@ + (UIAccessibilityElement *)accessibilityElementWithContainer:(id)container node
...
+#pragma mark - UIAccessibility
+
+- (void)setAccessibleElements:(NSArray *)accessibleElements
+{
+  _accessibleElements = nil;
+}
+
+- (NSArray *)accessibleElements
+{
+  ASDisplayNode *viewNode = self.asyncdisplaykit_node;
```
appleguy: Probably just call this node.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -112,11 +159,24 @@ - (id)accessibilityElementAtIndex:(NSInteger)index
...
-  return _accessibleElements[index];
+  ASDisplayNode *viewNode = self.asyncdisplaykit_node;
+  if (viewNode == nil) {
+    return nil;
+  }
+
+  // Recalculate the elements
+  NSArray *elements = SortAccessibilityElements(CollectAccessibilityElementsForViewAndNode(self, viewNode));
+
+  // We have to update the frame of the accessiblity element in case it changed
```
appleguy: This solution seems fine.  Just add something like: // the accessibilityFrame is in screen coordinates, so a scroll view moving is a common case where this value needs to change without this element able to detect this (besides performing the coordinate conversation again)

AsyncDisplayKit/ASDisplayNode.mm
maicki: This currently can crash and will be fixed soon

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1634,6 +1635,7 @@ - (void)_addSubnodeSubviewOrSublayer:(ASDisplayNode *)subnode
   BOOL canUseViewAPI = !self.isLayerBacked && !subnode.isLayerBacked;
   if (canUseViewAPI) {
     [_view addSubview:subnode.view];
+    [(_ASDisplayView *)_view setAccessibleElements:nil];
```
maicki: Fix casting

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1689,6 +1691,7 @@ - (void)removeFromSupernode
         [layer removeFromSuperlayer];
       } else {
         [view removeFromSuperview];
+        [(_ASDisplayView *)_view setAccessibleElements:nil];
```
maicki: Fix casting

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3149,6 +3152,7 @@ - (void)addSubnode:(ASDisplayNode *)subnode
       [selfNode addSubnode:subnode];
     } else {
       [self addSubview:subnode.view];
+      [(_ASDisplayView *)self setAccessibleElements:nil];
```
maicki: Fix casting

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
Adlai-Holler: The comparator block could even be static, using dispatch_once

maicki: I really liked the functional style (and prefer that in general) but it is adding some overhead so we pass in a `NSMutableArray` in all of the collect as well as the sort function.

Furthermore the comparator is static now

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
maicki: Changed to node

maicki:
@appleguy @Adlai-Holler Ready for review again. Although it's not the final solution we should consider merging it as it's an improvement as well as fixes the order of accessibility elements issue.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -32,96 +57,118 @@ + (UIAccessibilityElement *)accessibilityElementWithContainer:(id)container node
...
   ASDisplayNode *viewNode = self.asyncdisplaykit_node;
   if (viewNode == nil) {
-    return nil;
+    return @[];
   }
   
-  // Handle rasterize case
-  if (viewNode.shouldRasterizeDescendants) {
-    _accessibleElements = ASCollectUIAccessibilityElementsForNode(viewNode, viewNode, self);
+  CGRect screenFrame = UIAccessibilityConvertFrameToScreenCoordinates(self.frame, self);
```
Adlai-Holler: I believe this should be `self.bounds`. The docs say `The rect should exist in the view space of the UIView argument` and `self.frame` is in the superview's coordinates.

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -32,96 +57,118 @@ + (UIAccessibilityElement *)accessibilityElementWithContainer:(id)container node
...
+        // No view for layer backed nodes exist. It's necessary to create a UIAccessibilityElement that represents this node
+        accessiblityElement = [UIAccessibilityElement accessibilityElementWithContainer:view node:subnode];
+        
+        CGRect frame = [node convertRect:subnode.bounds fromNode:subnode];
+        [accessiblityElement setAccessibilityFrame:UIAccessibilityConvertFrameToScreenCoordinates(frame, view)];
+        [elements addObject:accessiblityElement];
+      } else {
+        // Accessiblity element is not layer backed just translate the view's frame to the container and add the view
+        accessiblityElement = subnode.view;
+        [accessiblityElement setAccessibilityFrame:UIAccessibilityConvertFrameToScreenCoordinates(subnode.frame, view)];
```
Adlai-Holler: Is this necessary? Will the UIView not have the correct accessibilityFrame by default?

AsyncDisplayKit/Details/_ASDisplayViewAccessiblity.mm
```diff
@@ -32,96 +57,118 @@ + (UIAccessibilityElement *)accessibilityElementWithContainer:(id)container node
...
 }
 
 - (NSInteger)indexOfAccessibilityElement:(id)element
 {
-  if (_accessibleElements == nil) {
-    return NSNotFound;
-  }
-  
-  return [_accessibleElements indexOfObject:element];
+  return [self.accessibleElements indexOfObject:element];
```
Adlai-Holler: Let's use `indexOfObjectIdenticalTo:`

Adlai-Holler:
Awesome. Couple comments

maicki:
@Adlai-Holler All comments are addressed. Can I have another look over it. Thanks!

Adlai-Holler:
Great! This stuff isn't exciting but it's important for us to be viable for AAA apps by Google etc

PR #2067, opened 2016-08-14
[Architecture] Add plans for new Section Info Objects

Adlai-Holler:
This is a pretty important change that addresses a major cause of issues with collection views at Pinterest and surely other companies. 

Hopefully this will give us everything we need to build this feature, and integrate smoothly with Pinterest very soon. @binl Thoughts? Our section controllers themselves could serve directly as the section user info.

@garrettmoon Would you be interested in making this a reality?

NOTE:
 - Change the viewing mode of Overview.md to the pretty file mode. The code isn't very interesting.
 - The sequence diagram links are safe to modify. They're copies, so feel free to modify them and post links to/images of your revised version.

Let's review this thing! @appleguy @maicki @levi @hannahmbanana


appleguy:
@Adlai-Holler very awesome!  Here's a link to the pretty Overview.md:

https://github.com/facebook/AsyncDisplayKit/blob/2ddbb0cc2ea2f4e96ce2af458aa865a861aed573/plans/section-infos-api/Overview.md

Section Info sounds like a pretty good name.  SectionLayoutInfo might make it more clear?

This type of extension object is guaranteed to not be required for regular cell / items in the data source, right, because an ASCellNode class could have properties for them?  Or should we consider something like that now, if it will be needed to be able to specify IDs or something else to enable diffing or virtualization features?

Adlai-Holler:
@appleguy Yeah I think we can indefinitely avoid this kind of object for items because nodes (or perhaps ASIndexedNodeContext if we want to promote that to a more long-lived position) already play that role.

I'm cool with SectionLayoutInfo as the name. Clarity is king.

Adlai-Holler:
On second thought maybe SectionLayoutInfo is too restrictive and value-typeish. We're starting it off with the non-layout field `sectionName` and I can see use cases for other ones in the future e.g. ASDK could use content states for the section to drive placeholders.

So we ought to go whole hog and subtly encourage users to treat this object as the general "node" that represents the section. Any info ASDK needs about that section in the future will be provided by the object. Maybe even a stronger word like `SectionContext` would be appropriate. @appleguy thoughts?

plans/section-infos-api/ASExampleLayoutSectionInfo.m
```diff
@@ -0,0 +1,22 @@
+/**
+ * An example of what a section info class might look like.
+ */
+@interface ASExampleLayoutSectionInfo : NSObject <ASSectionUserInfo>
+
+@property (nonatomic, weak, nullable) ASCollectionView *collectionView;
+@property (nullable, copy) NSString *sectionName;
```
garrettmoon: sectionTitle? Is this meant to be used as a title?

plans/section-infos-api/ASCollectionSection.h
```diff
@@ -0,0 +1,27 @@
...
+@property (strong, nullable) id<ASSectionUserInfo> userInfo;
+
+@end
+
+@protocol ASSectionUserInfo
+// This will be set once, immediately after the object is returned by the data source.
+@property (weak, nonatomic, nullable) ASCollectionView *collectionView;
+
+// Could be optional, but need to cache -respondsToSelector: dynamically.
+@property (nullable, readonly, copy) NSString *sectionName;
```
garrettmoon: sectionTitle? Is this meant to be used as a title?

Adlai-Holler: No it's intended for debugging, we'll put it in crash reports etc.

garrettmoon:
Just so I understand: The overall idea here is to break the communication between the collection view layout and the collection view layout delegate? Instead having the collection view layout reference these new SectionContexts which were captured synchronously when the section was reloaded or inserted?

Adlai-Holler:
Exactly.

garrettmoon:
@Adlai-Holler how do we normally move this from planning to commit? Should we create a new ASDK branch to see how it would work with the Pinterest app?

Adlai-Holler:
We're writing the book on that right now! 

Once this is reviewed, let's land it like normal. Since this isn't a breaking API change, we have total flexibility of when to get Pinterest involved – it might be valuable to do early on, in case it informs an architecture change. If it were a breaking API change, we'd need to prepare both ASDK and Pinterest diffs and land them back-to-back as close as possible.

Adlai-Holler:
I'm going to consider this accepted, and we can debate the names later if there is still disagreement.

PR #2070, opened 2016-08-14
[Layout] Coalescing layout until UIKit callbacks

maicki:
### Description:
Currently there are two problems with ASDK's layout process:
* It’s a must to call `measureWithSizeRange:` that triggers the creation of an ASLayout. If this method will never be called on a node no measure and layout pass will happen and it will appear as it it is blank
* Calling `setNeedsLayout` triggers a measurement and layout pass *synchronously* the following way:
    1. Walks up the whole tree and invalidate all layouts until it hits the root node
    2. Starting from the root node it starts measuring the nodes based on the constrained size AND applies the layout immediately for all nodes
    3. It changes the size of itself based on the calculated layout instead of just layout of subviews!

The behavior of Nodes described above is hugely confusing for developers that are used to use UIKit and e.g. setting a frame will trigger a layout.

The behavior I think we should like to have is the based on the following two passes:

1. **Measurement pass**:
    - Will measure (calling `measure:` or `measureWithSizeRange:`)  the node based on a given constrained size.
    - It will has the side effect to create an internal (pending) `ASLayout`.
    - The measurement pass needs to be thread safe as it should be able to measure from a background thread.
2. **Layout pass**:
    - A measurement pass (calling `measure:` or `measureWithSizeRange:`) should just create the internal cached `ASLayout` and return the size for the given size range. There should no layout happen especially, not setting the bounds for the root container
    - `-setNeedsLayout:` only invalidates the layout for the node and all subnodes and call's through to the backing `UIView` / `CALayer` method to let UIKit / CA. Eventually a `-layoutSubview` / `-layoutSublayer` call will happen
    - Bounds changes or if `-layoutSubview` / `-layoutSublayer` is called from the backing `UIView` / `CALayer` we should either apply the pending / cached layout or do a measurement pass before if the size of the pending layout is not equal to the nodes bounds as it changed in the meantime and do a layout (set all frames for subnodes) afterwards

We have to be careful with this PR though as this will change the behavior in terms of layout nodes.

### TODOS:
- [x] Automatic cell sizing in ASTableView
- [ ] Automatic cell sizing in ASCollectionView
- [ ] Add / Adjust unit tests
- [ ] Change sample apps

<br>
***This PR is one of the initiatives to make the Layout API more clear. Every feedback is more than welcome!***

### DIFF
--- a/AsyncDisplayKit/ASCellNode+Internal.h
+++ b/AsyncDisplayKit/ASCellNode+Internal.h
@@ -25,6 +25,7 @@
  * @param node A node informing the delegate about the relayout.
  * @param sizeChanged `YES` if the node's `calculatedSize` changed during the relayout, `NO` otherwise.
  */
+- (void)nodeDidRelayout:(ASCellNode *)node;
 - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged;
 
 /*
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -104,12 +104,12 @@
   _viewControllerNode.frame = self.bounds;
 }
 
-- (void)__setNeedsLayout
+- (void)__layout
 {
-  [super __setNeedsLayout];
+  [super __layout];
   
   ASDN::MutexLocker l(__instanceLock__);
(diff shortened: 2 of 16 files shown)


Adlai-Holler:
Let's close this for now and reopen after the new layout API is built. This is a critically important task – our current implementation of `setNeedsLayout` has very unexpected side-effects and costs, and I can't wait to make it better.

PR #2071, opened 2016-08-15
[ASStackLayoutable] Set .flexShrink to YES by default

maicki:
To align with the CSS flex-shrink spec, let's set the default value of flexShrink to YES: http://www.w3.org/TR/css-flexbox-1/#propdef-flex-shrink

In PR #851 we would only set it for `ASTextNode`, but I would recommend we will set it for all `ASStackLayoutable` to `YES` by default.

CC @appleguy @nguyenhuy @Adlai-Holler @rcancro Thoughts?
<br>
***This PR is one of the initiatives to make the Layout API more clear. Every feedback is more than welcome!***

### DIFF
--- a/AsyncDisplayKit/Details/ASEnvironment.h
+++ b/AsyncDisplayKit/Details/ASEnvironment.h
@@ -39,7 +39,7 @@ typedef struct ASEnvironmentLayoutOptionsState {
   CGFloat spacingBefore;// = 0;
   CGFloat spacingAfter;// = 0;
   BOOL flexGrow;// = NO;
-  BOOL flexShrink;// = NO;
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
I'm good with this. Not only does it conform better with CSS but it also means that, if there's not enough space, all elements will shrink equally by default, rather than overflowing, which is better default behavior IMO.

Let's land this but pair it with a Pinterest audit/update to make sure we don't break anything.

nguyenhuy:
LGTM! Would be great if you briefly test examples as well.

Adlai-Holler:
Superseded by #2247

PR #2072, opened 2016-08-15
Fix wrong annotation which is causing a warning at compile time

dmcrodrigues:
This wrong annotation was introduced by me in #2035, `- initWithCoder:` which is a failable initialiser needs a return type marked as `nullable`.


maicki:
LGTM

dmcrodrigues:
No problem, I should not have introduced it in the first place 😅

Adlai-Holler:
Beautiful!

PR #2075, opened 2016-08-16
[ASDisplayNode] Don't use the default layout transition animation of no animation is wanted

maicki:
This fixes a regression (found in Pinterest) with #2052.


Adlai-Holler:
Beautiful!

PR #2079, opened 2016-08-16
[_ASDisplayViewAccessiblity] Fix accessibility in scroll views

garrettmoon:
There are two changes here:

1. Because we can't reliably update the screen positions of accessibility elements
contained within a scroll view, we need to calculate them on demand, so we override
the accessibilityFrame method to do that.

2. Throwing away our calculated accessibility elements on frame change seemed to
cause the accessibility system to be confused.

Combining these two fixes together results in success, yay!


maicki:
LGTM

Adlai-Holler:
Awesome, let's land once build passes

PR #2083, opened 2016-08-16
Make Our Snapshot Testing Possible by the Public, Standardize

Adlai-Holler:
Pinterest uses snapshot tests internally, and we need to pave the way for others to do the same.

So this removes the framework-private and Objective-C++ bits from `ASSnapshotTestCase` and standardizes the process. So now all you need to do is create the node, call -measure on it, and you're ready to snapshot.

For right now, consumers of this code should copy the .h and .m files into their app repo. It's not beautiful so in the future we should think through how to get CocoaPods to give them the option to bring in this + FBSnapshotTestCase in their unit tests.


keleftheriou:

maicki:
LGTM

PR #2091, opened 2016-08-18
[ASTableView] Ignore Table Cell Layouts When There's No Node Assigned

Adlai-Holler:
This can cause pointless beginUpdates/endUpdates pairs that interfere with scrolling animations.


maicki:
LGTM

PR #2094, opened 2016-08-18
[ASDisplayNode] Clarify Some Special Logic Around Wrapper Nodes

Adlai-Holler:
This clarifies some behavior that I found confusing when debugging. This does not change any behavior. Worked with @maicki to:

- Rename `canCallNeedsDisplayOfLayer` to `canCallSetNeedsDisplayOfLayer` to be clearer.
- Make the rationale around that flag clearer. The actual behavior of the code is the same.
   - We cannot clear the contents or setNeedsDisplay of UIImageView-wrapping nodes.
   - We can do both of the above on any other kind of node.
   - We must call `setNeedsDisplay` on all non-UIImageView wrapper nodes in `recursivelyTriggerDisplayForLayer`, but there's no need to call it on normal nodes


keleftheriou:
test this please

maicki:
LGTM

PR #2095, opened 2016-08-18
[ASDisplayView] Remove Some Unused Code

Adlai-Holler:
Found this while digging around in our drawing/CA integration. May as well yank it out! @maicki @appleguy


maicki:
Change makes sense. Let's see how many are calling `setNeedsDisplay` on a background thread ;)

PR #2096, opened 2016-08-18
[ASDisplayNode] WIP Fix Never Drawn Layers

Adlai-Holler:
I tried to strip this down as much as possible, but it's currently _too_ stripped down and it's going to setNeedsDisplay more often than it should. Got hung up today on other stuff so I'll pick this up tomorrow.


Adlai-Holler:
Superseded by #2149

PR #2099, opened 2016-08-18
[Layout] Remove ASLayoutValidation

maicki:
## This PR will be replaced by: #2172

`ASLayoutValidation` was never really thought through. There are other ideas / solutions on the horizon that will come in the near future, so we should just get rid of that code.

@Adlai-Holler @garrettmoon @appleguy


Adlai-Holler:
I'm cool with removing this but I don't know much about the history there. Let's have @appleguy weigh in

maicki:
@appleguy @levi any thoughts?

levi:
Seems early to remove this if we don't have a replacement.

Adlai-Holler:
Superseded by #2172

PR #2102, opened 2016-08-18
Remove numberOfSectionsForSupplementaryNodeOfKind: Method

Adlai-Holler:
This adds a test that confirms UICollectionView will throw an exception for supplementary nodes that are outside of the normal `-numberOfSections` range.

This means that there is only one set of sections, so this method is a misconception. Deprecated as a reminder to remove, and we'll never call it.

This is probably the first of many changes to our collection view stuff


Adlai-Holler:
Ready for review @maicki @appleguy @levi

maicki:
LGTM

Issue #2111, opened 2016-08-20
Add attributes to properties in ASImageNodeContentsKey 


Yue-Wang-Google:
to make clang happy:

error: ARC forbids synthesizing a property of an Objective-C object with unspecified ownership or storage attribute

Adlai-Holler:
Closed by #2113.

PR #2113, opened 2016-08-20
set the right attributes for block properties.

Yue-Wang-Google:
See https://github.com/facebook/AsyncDisplayKit/issues/2111

### DIFF
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -54,9 +54,9 @@ struct ASImageNodeDrawParameters {
 @property CGRect imageDrawRect;
 @property BOOL isOpaque;
 @property (nonatomic, strong) UIColor *backgroundColor;
-@property (strong) ASDisplayNodeContextModifier preContextBlock;
-@property (strong) ASDisplayNodeContextModifier postContextBlock;
-@property (strong) asimagenode_modification_block_t imageModificationBlock;
+@property (copy) ASDisplayNodeContextModifier preContextBlock;
+@property (copy) ASDisplayNodeContextModifier postContextBlock;
+@property (copy) asimagenode_modification_block_t imageModificationBlock;
 
 @end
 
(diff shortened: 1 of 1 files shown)


Adlai-Holler:

Adlai-Holler:
I believe these should be `(strong)`, according to http://adcdownload.apple.com/wwdc_2012/wwdc_2012_session_pdfs/session_712__asynchronous_design_patterns_with_blocks_gcd_and_xpc.pdf page 83. 

Additionally http://clang.llvm.org/docs/AutomaticReferenceCounting.html#blocks states that:

> With the exception of retains done as part of initializing a `__strong` parameter variable or reading a `__weak` variable, whenever these semantics call for retaining a value of block-pointer type, it has the effect of a `Block_copy`. The optimizer may remove such copies when it sees that the result is used only as an argument to a call.

So for stored strong properties, ARC will automatically copy the block.

Yue-Wang-Google:
@Adlai-Holler  done. Thanks

Yue-Wang-Google:
wait...

Per Apple:

Note: You should specify copy as the property attribute, because a block needs to be copied to keep track of its captured state outside of the original scope. This isn’t something you need to worry about when using Automatic Reference Counting, as it will happen automatically, but it’s best practice for the property attribute to show the resultant behavior. For more information, see Blocks Programming Topics.


https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html#//apple_ref/doc/uid/TP40011210-CH8-SW12

Yue-Wang-Google:
btw, could these be nonatomic? I haven't dived into the code deeply.

maicki:
@Adlai-Holler Blocks should be copy as properties, see also the documentation @Yue-Wang-Google linked too ...

@Yue-Wang-Google I think we should add nonatomic cc @Adlai-Holler

Adlai-Holler:
Yeah all these properties should be nonatomic

Yue-Wang-Google:
Done.

Adlai-Holler:
Sweet, thank you for the stream of improvements!

PR #2115, opened 2016-08-21
[ASViewController] Propagate the traits on willTransitionToTraitCollection

rcancro:
Previously we were only propagating the trait collection on `viewWillTransitionToSize` since it is called shortly after `willTransitionToTraitCollection`. However, some important things can happen in that time “shortly after” (like collection view layout). As long as nothing changes from `willTransitionToTraitCollection` to `viewWillTransitionToSize` (which it shouldn’t) the traits will not be re-propagated anyhow.

Also make sure to use the `ASEnvironmentTraitCollectionMakeDefault` method when creating new envTraitCollection so the struct isn’t filled with junk.


Adlai-Holler:
Love it! @maicki @levi thoughts?

PR #2118, opened 2016-08-23
[ASDataController] Ensure Supplementary Section Count Always Tracks Item Section Count

Adlai-Holler:
This resolves an issue where, say you have 3 sections, but only 2 of them have headers. Before, we would only store 2 sections' worth of headers, but if you then delete section #2, we are out of bounds when we try to retrieve that section of headers.

This diff makes sure that even if only 2 sections have headers, we always have 3 arrays, so that items & supplementaries stay completely in sync.

This issue recently became more apparent because we removed the sectionsForSupplementaryNodeOfKind: method which, ironically, if implemented correctly would have caused the problem but was virtually always implemented just to return `numberOfSections`, masking it. In the future (ship task 193) we will continue to make changes to our storage of supplementary nodes, to support section-level supplementaries (even for sections that don't exist!) as well as item-level ones.

### DIFF
--- a/AsyncDisplayKit/Details/ASCollectionDataController.mm
+++ b/AsyncDisplayKit/Details/ASCollectionDataController.mm
@@ -43,17 +43,18 @@
   return self;
 }
 
-- (void)prepareForReloadData
+- (void)prepareForReloadDataWithSectionCount:(NSInteger)newSectionCount
 {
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Ready for review & fast-track to p6.10 @appleguy @levi @garrettmoon @maicki

appleguy:
@Adlai-Holler this looks good to me, I read through and believe I understand the issue.  Thanks for the investigation!

PR #2119, opened 2016-08-23
[ASVideoNode] Addition of reset method to return to initial state

gazreese:
In our app we need to return the ASVideoNode back to its original state after playback. This initial state would be with the placeholder image displayed and play button visible. Unfortunately, without changes to ASVideoNode this doesn't appear to be possible.

This PR covers the addition of a 'reset' method to allow the video node or video player node to return to its initial state. Unfortunately it didn't appear possible to add any meaningful unit tests to cover these changes.

In some ways these changes are needed because of the hierarchy of ASVideoNode, which subclasses ASNetworkImageNode. Because the video node _is a_ placeholder image there's no way for users of the class or for the class code itself to manipulate the placeholder. If the video node owned its placeholder this code would look much different, and may not even be needed. 

As it stands the only way I found to return to the placeholder state was actually to remove the player node, which is then recreated when the user plays the video again. This must have at least some performance / bandwidth impact I would have thought.

Any suggestions are much appreciated.


Adlai-Holler:
@gazreese I share your concerns about performance and architecture here. We have the alternative of doing `_playerNode.hidden=YES` and unhiding it in `play`.

That may save us some time during the state transitions, but it does have the drawback of accumulating AVPlayerLayer instances if the user scrolls through a feed of videos, watching each one and letting it reset. The developer may not be aware that the AVPlayerLayers are sticking around. 

I believe that AVPlayer itself, not the player layer, manages the resources for the video so it would be unlikely to cause more network traffic either way we go.

I advocate for showing/hiding the player layer rather than removing/recreating it but I'd like @appleguy @maxsz or @binl to weigh in as they have more expertise around AVFoundation.

gazreese:
Hi @Adlai-Holler, apologies for the delay getting back to you on this. I've been on holiday and just got back.

As my project is exactly as you describe, i.e. a feed of videos that the user scrolls through, I'm quite keen on the original remove/recreate method. I'm working in a memory-constrained environment, so I can't really risk leaking the AVPlayerLayer objects.

@appleguy @maxsz or @binl - if you have any comments they would be much appreciated.

AsyncDisplayKit/ASVideoNode.h
```diff
@@ -39,6 +39,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)play;
 - (void)pause;
 - (BOOL)isPlaying;
+- (void)reset;
```
appleguy: @gazreese thoughts on a more descriptive name?

-resetToPlaceholder;
-cancelPlayback;
-clearPlayer;
-resetToStart;

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -686,6 +686,18 @@ - (BOOL)isStateChangeValid:(ASVideoNodePlayerState)state
   return YES;
 }
 
+- (void)reset {
```
appleguy: Open brace on new line for all methods :)

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -686,6 +686,18 @@ - (BOOL)isStateChangeValid:(ASVideoNodePlayerState)state
...
+  ASDN::MutexLocker l(__instanceLock__);
+  
+  if (_playerNode != nil) {
+    [_playerNode removeFromSupernode];
+    _playerNode = nil;
+  }
+  
+  [_player seekToTime:kCMTimeZero];
```
appleguy: @gazreese should we deallocate the AVPlayer as well?  This seems more consistent from a memory management perspective.

AsyncDisplayKit/ASVideoNode.h
gazreese: I like resetToPlaceholder, so I'll go with that unless anyone else has any ideas?

AsyncDisplayKit/ASVideoNode.mm
gazreese: Done

AsyncDisplayKit/ASVideoNode.mm
gazreese: Hi @appleguy I gave this a try and it turns out it's a bit more complicated than it at first seems. The changes required to support this would actually be quite a lot larger to get things to the point where you could re-start the playing of the video after the reset.

It turns out that the construction of the AVPlayer is tied up in `(void)prepareToPlayAsset:(AVAsset *)asset withKeys:(NSArray<NSString *> *)requestedKeys`. This is mixed in with the construction of the AVPlayerItem (which we'd actually want to keep hanging around) and setting up the delegates. I'd need to split prepareToPlayAsset up to separate the handling of the AVPlayer from the rest of the construction. 

I could have a go at this but I think it's best dealt with as a separate bug, keeping this PR with the seekToTime as above. I've actually been using the code above for a while in a live app and it seems to work fine, so I'm confident in the implementation.

Please let me know what you think.

appleguy: @gazreese ok, thanks for the details!  I think we can move forward without it, although the method's behavior would most logically include this resource-saving measure.

It's not a critical priority so don't sweat about the followup, but if you get time, it would be awesome.  The whole team really appreciates all the contributions you've made!

PR #2120, opened 2016-08-23
[ASVideoNode] Maintain playback position between visibility changes

gazreese:
My current project uses ASVideoNode / ASVideoPlayerNode within a collection view. Unfortunately, when the videos scroll out of view and back in again the playback position is reset to the beginning. I feel that this is contrary to what users would expect.

This change maintains the current playback time in an ivar so that we can reestablish the ASVideoNode display at the same position as it was when it left view.


gazreese:
Merged the master branch into mine again to resolve merge conflicts

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -454,6 +460,9 @@ - (void)didExitVisibleState
   ASDN::MutexLocker l(__instanceLock__);
   
   if (_shouldBePlaying) {
+    if(_player != nil && CMTIME_IS_VALID(_player.currentTime)) {
+      _lastPlaybackTime = _player.currentTime;
+    }
     [self pause];
     _shouldBePlaying = YES;
```
Adlai-Holler: Small things: Space after `if` and let's pause before reading the current time.

Adlai-Holler:
@gazreese Thanks for this improvement, and sorry for the long turnaround time on a review. One comment and we'll take it to the hoop!

Adlai-Holler:
Excellent! Thank you @gazreese for making this happen.

PR #2121, opened 2016-08-23
[ASCollectionView] Relayout Nodes as Soon as Bounds Changes

Adlai-Holler:
We previously waited until `layoutSubviews` but that's too late as @rcancro discovered.

I added a failing test case that ensures everything is up-to-date before the first relayout after a rotation.

I expect this to be somewhat controversial because I did something unusual. In order for ASCollectionView to find out about the bounds change before UICollectionView requeries the layout object, I added an optional `ASCALayerExtendedDelegate` protocol. Basically if your view responds to these methods, ASDisplayLayer will call them. I can't use our _ASDisplayLayerDelegate because collection views may not have nodes.

I considered adding an _ASCollectionLayer class with special functionality but decided this more general solution was better.

I also renamed some ivars in ASCollectionView for clarity:
  - `_maxSizeForNodesConstrainedSize` -> `_lastBoundsSizeUsedForMeasuringNodes`
    - This value is not actually used as the max size for nodes constrained size.
  - `_ignoreMaxSizeChange` -> `_ignoreNextBoundsSizeChangeForMeasuringNodes`
    - The ignore only lasts for one change, and it's about bounds size for the particular purpose of remeasuring nodes.

Ready for review @maicki @levi @appleguy @rcancro !

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -86,7 +86,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 #pragma mark -
 #pragma mark ASCollectionView.
 
-@interface ASCollectionView () <ASRangeControllerDataSource, ASRangeControllerDelegate, ASCollectionDataControllerSource, ASCellNodeInteractionDelegate, ASDelegateProxyInterceptor, ASBatchFetchingScrollView, ASDataControllerEnvironmentDelegate> {
+@interface ASCollectionView () <ASRangeControllerDataSource, ASRangeControllerDelegate, ASCollectionDataControllerSource, ASCellNodeInteractionDelegate, ASDelegateProxyInterceptor, ASBatchFetchingScrollView, ASDataControllerEnvironmentDelegate, ASCALayerExtendedDelegate> {
   ASCollectionViewProxy *_proxyDataSource;
   ASCollectionViewProxy *_proxyDelegate;
   
@@ -106,8 +106,8 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   ASBatchContext *_batchContext;
   
-  CGSize _maxSizeForNodesConstrainedSize;
-  BOOL _ignoreMaxSizeChange;
+  CGSize _lastBoundsSizeUsedForMeasuringNodes;
+  BOOL _ignoreNextBoundsSizeChangeForMeasuringNodes;
   
   NSMutableSet *_registeredSupplementaryKinds;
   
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/Details/_ASDisplayLayer.h
```diff
@@ -70,6 +70,20 @@ typedef BOOL(^asdisplaynode_iscancelled_block_t)(void);
 @end
 
 /**
+ * Optional methods that the view associated with an _ASDisplayLayer can implement. 
+ * This is distinguished from _ASDisplayLayerDelegate in that it points to the _view_
+ * not the node. Unfortunately this is required by ASCollectionView, since we currently
+ * can't guarantee that an ASCollectionNode exists for it.
+ */
+@protocol ASCALayerExtendedDelegate
```
levi: `ASLayerExtendedDelegate`

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1309,6 +1289,40 @@ - (void)didMoveToWindow
...
+
+  // First size change occurs during initial configuration. An expensive relayout pass is unnecessary at that time
+  // and should be avoided, assuming that the initial data loading automatically runs shortly afterward.
+  if (_ignoreNextBoundsSizeChangeForMeasuringNodes) {
+    _ignoreNextBoundsSizeChangeForMeasuringNodes = NO;
+  } else {
+    // This actually doesn't perform an animation, but prevents the transaction block from being processed in the
+    // data controller's prevent animation block that would interrupt an interrupted relayout happening in an animation block
+    // ie. ASCollectionView bounds change on rotation or multi-tasking split view resize.
+    [self performBatchAnimated:YES updates:^{
```
garrettmoon: The answer is likely no, but should we instead add another internal API that has an option to prevent the prevent animation block instead?

garrettmoon: My thought being the behavior could change in the future and then the above would make a lot less sense.

AsyncDisplayKit/Details/_ASDisplayLayer.h
Adlai-Holler: I think `ASCALayerExtendedDelegate` is more appropriate because we're literally piggy-backing on top of CALayerDelegate here, and I want to distinguish it from `_ASDisplayLayerDelegate`.

AsyncDisplayKit/ASCollectionView.mm
Adlai-Holler: We certainly could in the future, but to keep this change as minimal as possible, I just copied this code block directly out of `layoutSubviews`

Adlai-Holler:
Ricky confirmed this fixes his issue in a big way – merging.

PR #2122, opened 2016-08-23
[ASViewController] Add asyncTraitCollectionDidChange to ASViewController

rcancro:
This is consistent with UIViewController but also gets called in more cases than traitCollectionDidChange since our trait collection registers a change as any size change.


Adlai-Holler:
@rcancro feel free to reopen this if needed.

PR #2107, opened 2016-08-19
Add ability to force the size which an image node is rendered.

garrettmoon:
Putting this up to hopefully facilitate discussion.

PR #2123, opened 2016-08-23
Attempt to respect edgesForExtendedLayout on ASViewController

garrettmoon:
This is a pretty hacky way to achieve this, but I can't find a better
one. I specifically tried looking at the topLayoutGuide.length, however
it's zero when edgesForExtendedLayout doesn't include the top (of course
it's valid when it does :/)


Adlai-Holler:
Could we get a failing unit test and some context to go with this? It's expected that the topLayoutGuide will be 0 if the top edge isn't extended, because the actual view will be laid out so that its top edge aligns to the bottom of the nav bar – so the overlap is 0.

garrettmoon:
Ah, sorry, the issue is that currently, the root node is sized to be full screen regardless of edgesForExtendedLayout. I think I can build a failing unit test.

Adlai-Holler:
@keleftheriou @garrettmoon  this should have failed the Jenkins build. See the failing test shown in the Travis build. That test case was merged in #2107 and contained an error that was masked by other issues with Travis at the time. Other open PRs show Jenkins passing but Travis correctly failing. The test should pass in #2127

PR #2124, opened 2016-08-24
[ASDisplayNode] Remove Display Semaphore

Adlai-Holler:
Now that ASAsyncTransaction manages a thread pool, we don't need to use a semaphore to limit concurrent displays.

@appleguy should probably review this since he's the only one with super deep knowledge about this code. Obviously this isn't urgent, it'll be here whenever you have time.


maicki:
LGTM. Let's get that in.

PR #2125, opened 2016-08-24
[ComponentKit] Added a CKComponent hosting node

GotEmB:
This let's you host a [`CKComponent`](https://github.com/facebook/componentkit/) within an `ASDisplay` node in a performant way. This is more or less based on [`CKComponentHostingView`](https://github.com/facebook/componentkit/blob/master/ComponentKit/HostingView/CKComponentHostingView.h).

`examples/ComponentHosting` contains a sample project that shows how this can be used within an node tree.

Note: `ASComponentHostingNode` is compiled only if `ComponentKit` is included in the project.

### DIFF
--- a/AsyncDisplayKit/ASComponentHostingNode.mm
+++ b/AsyncDisplayKit/ASComponentHostingNode.mm
@@ -1,0 +1,216 @@
+//
+//  ASComponentHostingNode.mm
+//  AsyncDisplayKit
+//
+//  Copyright (c) 2014-present, Facebook, Inc.  All rights reserved.
+//  This source code is licensed under the BSD-style license found in the
+//  LICENSE file in the root directory of this source tree. An additional grant
+//  of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#if __has_include(<ComponentKit/ComponentKit.h>)
+
+#import "ASComponentHostingNode.h"
+
+#import <ComponentKit/CKComponentLayout.h>
+#import <ComponentKit/CKComponentProvider.h>
+#import <ComponentKit/CKComponentRootView.h>
+#import <ComponentKit/CKComponentScopeRoot.h>
+#import <ComponentKit/CKComponentSizeRangeProviding.h>
+
+#import "ASAssert.h"
+#import "ASDisplayNode+Subclasses.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+struct ASComponentHostingViewInputs {
+  CKComponentScopeRoot *scopeRoot;
+  id<NSObject> _Nullable model;
+  id<NSObject> _Nullable context;
+  CKComponentStateUpdateMap stateUpdates;
+};
+
+@interface ASComponentHostingNode () <CKComponentStateListener>
+{
+  Class<CKComponentProvider> _componentProvider;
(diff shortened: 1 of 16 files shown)


examples/ComponentHosting/Podfile
```diff
@@ -0,0 +1,6 @@
+source 'https://github.com/CocoaPods/Specs.git'
+platform :ios, '7.0'
+target 'Sample' do
+  pod 'ComponentKit', :git => 'https://github.com/facebook/componentkit.git', :commit => '115d2012580f17276b5383cd1dad2b306b24aeb1'
```
nlutsenko: Does it specifically need this commit, if yes, why?

nlutsenko: Per our discussion: We probably should reference the latest public release (latest tag) and check-in `Podfile.lock` here as well, instead of referencing a specific commit.

GotEmB: Yeah, master on ComponentKit has a different API than the latest tag (0.14). I'll try to get them to push out a new tag.

Adlai-Holler:
Hey guys sorry about the test failure, if you rebase on master it should be good. For the example failures, it looks like it doesn't like the __has_include macro.

GotEmB:
@Adlai-Holler Looks like it's still failing due to the Sample App not having a `Sample` scheme?

Adlai-Holler:
Ah yeah, all our example projects have to be cloned from the same template. @hannahmbanana knows more about it than I do – could you take a look?

GotEmB:
We actually feel that it'd be better to have this in its own top level repo. The main reasons being able to require particular versions of both `ComponentKit

GotEmB:
Closing this PR in favor of maintaining a separate top level repo.

AsyncDisplayKit/ASComponentHostingNode.mm
```diff
@@ -0,0 +1,216 @@
...
+
+- (void)_asynchronouslyUpdateComponentIfNeeded
+{
+  if (_scheduledAsynchronousComponentUpdate) {
+    return;
+  }
+  _scheduledAsynchronousComponentUpdate = YES;
+
+  // Wait until the end of the run loop so that if multiple async updates are triggered we don't thrash.
+  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
ocrickard: I think this is the same as hosting view, but worth noting that if *many* hosting views are used together, we don't control their queue bandwidth at all. One block per-hosting view may execute simultaneously. Probably not an issue unless you've got 60+ hosting views all rendering at once (which you may need to deal with in ASDK?).

ocrickard:
Thanks for moving this to another repo! Reviewing a bit here for now. Looks pretty straight-forward to me.

AsyncDisplayKit/ASComponentHostingNode.mm
GotEmB: @ocrickard: 👍

How about moving to an `NSOperationQueue` with a cap on concurrency? Or just a single serial dispatch queue that's off the main thread?

PR #2128, opened 2016-08-24
[ASDisplayNode] Add `onDidLoad` Method to Perform Work When Loaded

Adlai-Holler:
Obviously we already have `didLoad`, but engineers still do view-related stuff in `init`.

This provides engineers with an easy way to capture arguments or state from `-init` and run it safely after the view is loaded. 

What do you guys think? Worthwhile?

P.S. I also realized that if you make a wrapper node with `initWithViewBlock:` and then we unload it because it becomes rasterized, we actually can't get it back because we dispose of the view block. I added assertions against rasterizing wrapper nodes for this reason and since it's a very bad idea.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -40,7 +40,7 @@ typedef CALayer * _Nonnull(^ASDisplayNodeLayerBlock)();
 /**
  * ASDisplayNode loaded callback block. This block is called BEFORE the -didLoad method and is always called on the main thread.
  */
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
PPS I considered supporting calling this after the node is loaded, but I think we shouldn't. If the method is thread-safe we'd have to loosen the semantics and increase the complexity dramatically.

Adlai-Holler:
@maicki @levi @hannahmbanana Thoughts?

hannahmbanana:
@Adlai-Holler: Sorry I missed this the first time. It looks like I should add this to the documentation. Could you give me an example snippet of it being used?

PR #2130, opened 2016-08-24
[ASDisplayNode] Renamed range update callbacks

garrettmoon:
We finally settled on

didEnter/ExitDisplayState
didEnter/ExitPreloadState
didEnter/ExitVisibleState

This change is meant to unify the range update methods to relate to each
other and hopefully be a bit more self explanatory.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -177,6 +177,7 @@
 		636EA1A41C7FF4EC00EE152F /* NSArray+Diffing.m in Sources */ = {isa = PBXBuildFile; fileRef = DBC452DA1C5BF64600B16017 /* NSArray+Diffing.m */; };
 		636EA1A51C7FF4EF00EE152F /* ASDefaultPlayButton.m in Sources */ = {isa = PBXBuildFile; fileRef = AEB7B0191C5962EA00662EF4 /* ASDefaultPlayButton.m */; };
 		680346941CE4052A0009FEB4 /* ASNavigationController.h in Headers */ = {isa = PBXBuildFile; fileRef = 68FC85DC1CE29AB700EDD713 /* ASNavigationController.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		683489281D70DE3400327501 /* ASDisplayNode+Deprecated.h in Headers */ = {isa = PBXBuildFile; fileRef = 683489271D70DE3400327501 /* ASDisplayNode+Deprecated.h */; };
 		68355B311CB5799E001D4E68 /* ASImageNode+AnimatedImage.mm in Sources */ = {isa = PBXBuildFile; fileRef = 68355B2E1CB5799E001D4E68 /* ASImageNode+AnimatedImage.mm */; };
 		68355B341CB579B9001D4E68 /* ASImageNode+AnimatedImage.mm in Sources */ = {isa = PBXBuildFile; fileRef = 68355B2E1CB5799E001D4E68 /* ASImageNode+AnimatedImage.mm */; };
 		68355B3A1CB57A5A001D4E68 /* ASPINRemoteImageDownloader.m in Sources */ = {isa = PBXBuildFile; fileRef = 68355B361CB57A5A001D4E68 /* ASPINRemoteImageDownloader.m */; };
@@ -943,6 +944,7 @@
 		4640521B1A3F83C40061C0BA /* ASFlowLayoutController.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ASFlowLayoutController.h; sourceTree = "<group>"; };
 		4640521C1A3F83C40061C0BA /* ASFlowLayoutController.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; path = ASFlowLayoutController.mm; sourceTree = "<group>"; };
 		4640521D1A3F83C40061C0BA /* ASLayoutController.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ASLayoutController.h; sourceTree = "<group>"; };
+		683489271D70DE3400327501 /* ASDisplayNode+Deprecated.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = "ASDisplayNode+Deprecated.h"; sourceTree = "<group>"; };
 		68355B2E1CB5799E001D4E68 /* ASImageNode+AnimatedImage.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; path = "ASImageNode+AnimatedImage.mm"; sourceTree = "<group>"; };
 		68355B361CB57A5A001D4E68 /* ASPINRemoteImageDownloader.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = ASPINRemoteImageDownloader.m; sourceTree = "<group>"; };
 		68355B371CB57A5A001D4E68 /* ASImageContainerProtocolCategories.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ASImageContainerProtocolCategories.h; sourceTree = "<group>"; };
@@ -1260,6 +1262,7 @@
 				058D09D8195D050800B7D73C /* ASDisplayNode.h */,
 				058D09D9195D050800B7D73C /* ASDisplayNode.mm */,
 				68B027791C1A79CC0041016B /* ASDisplayNode+Beta.h */,
+				683489271D70DE3400327501 /* ASDisplayNode+Deprecated.h */,
 				058D09DA195D050800B7D73C /* ASDisplayNode+Subclasses.h */,
 				058D09DB195D050800B7D73C /* ASDisplayNodeExtras.h */,
 				058D09DC195D050800B7D73C /* ASDisplayNodeExtras.mm */,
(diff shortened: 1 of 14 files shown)


AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -238,36 +238,58 @@ NS_ASSUME_NONNULL_BEGIN
...
-- (void)displayStateDidChange:(BOOL)inDisplayState ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)didEnterDisplayState ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
- * @abstract Called whenever the the node has entered or left the load state.
+ * @abstract Called whenever the the node has exited the display state.
+ *
+ * @discussion Subclasses may use this to monitor when a node should no longer be rendering its content.
+ *
+ * @note This method can be called from any thread and should therefore be thread safe.
```
garrettmoon: All of these methods have these notes. Should they? It would be materially easier for clients if we guaranteed they were called on the main thread.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -238,36 +238,58 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)interfaceStateDidChange:(ASInterfaceState)newState fromState:(ASInterfaceState)oldState ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
- * @abstract Called whenever the visiblity of the node changed.
+ * @abstract Called whenever the node becomes visible.
  *
  * @discussion Subclasses may use this to monitor when they become visible.
+ *
+ * @note This method can be called from any thread and should therefore be thread safe.
```
appleguy: @garrettmoon this one isn't right — we should make a strong guarantee that the visible state methods are called on the main thread.  It's the only way to guarantee their timing is relevant (e.g., actually called before it is visible onscreen), and work performed here is very often related to animations or other main thread things

garrettmoon: I'll update to remove, thanks!

AsyncDisplayKit/ASDisplayNode+Subclasses.h
garrettmoon: My thought is we should guarantee that these will be called on main, but what does everyone else think about signing ourselves up for maintaining that?

hannahmbanana:
When we merge this, we need to update http://asyncdisplaykit.org/docs/upgrading.html with the following:

Updated Interface State callback method names. The new names are meant to unify the range update methods to show how they relate to each other & hopefully be a bit more self explanatory:

- `-didEnter/ExitPreloadState`
- `-didEnter/ExitDisplayState`
- `-didEnter/ExitVisibleState`

These new methods replace the following:

- `-loadStateDidChange:(BOOL)inLoadState`
- `-displayStateDidChange:(BOOL)inDisplayState`
- `-visibleStateDidChange:(BOOL)isVisible`

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -472,6 +494,39 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)asyncTraitCollectionDidChange;
 
+/**
+ * @abstract Called whenever the visiblity of the node changed.
+ *
+ * @discussion Subclasses may use this to monitor when they become visible.
+ */
+- (void)visibilityDidChange:(BOOL)isVisible ASDISPLAYNODE_REQUIRES_SUPER ASDISPLAYNODE_DEPRECATED;
```
appleguy: Are these still supposed to exist in the +Subclasses header?  Should they be in a deprecated category?

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -238,36 +238,58 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * @discussion Subclasses may use this to monitor data for a node should be preloaded, either from a local or remote source.
+ *
+ * @note this method is guaranteed to be called on main.
+ */
+- (void)didEnterPreloadState ASDISPLAYNODE_REQUIRES_SUPER;
+
+/**
+ * @abstract Called whenever the the node has exited the preload state.
+ *
```
appleguy: @garrettmoon I don't think it's standard in our current header docs to have extra lines between these - could you densify them a bit?  Having twice as many methods makes it harder to consume the info IMO relative to shorter block (would still be 5 full lines of comments)

AsyncDisplayKit/ASDisplayNode+Subclasses.h
garrettmoon: @appleguy I can make a new ASDisplayNode+Deprecated category. Are you thinking we should move all of these there?

AsyncDisplayKit/ASDisplayNode+Subclasses.h
garrettmoon: I agree that they're nicer to read without the line breaks, however it appears that there are breaks between all the '@' comments in every header I looked at (though sometimes the param ones are stuck together). Should we standardize 'em all? If so, probably best to do in a different PR?

appleguy: @garrettmoon ok, that's fine.  IIRC these were reformatted a couple months ago and it might have came in then...

AsyncDisplayKit/ASDisplayNode+Subclasses.h
appleguy: @garrettmoon yeah, that seems best.  They can stay in the same file at the bottom.

AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1881,12 +1885,15 @@
...
 			buildPhases = (
+				A5E2A5BB10F93572445AC844 /* 📦 Check Pods Manifest.lock */,
 				2E61B6A0DB0F436A9DDBE86F /* [CP] Check Pods Manifest.lock */,
 				058D09B8195D04C000B7D73C /* Sources */,
 				058D09B9195D04C000B7D73C /* Frameworks */,
 				058D09BA195D04C000B7D73C /* Resources */,
 				3B9D88CDF51B429C8409E4B6 /* [CP] Copy Pods Resources */,
 				B130AB1AC0A1E5162E211C19 /* [CP] Embed Pods Frameworks */,
+				76FDAFA9FB106E27BE016772 /* 📦 Embed Pods Frameworks */,
+				58AF99A9E1C99C30BB2F5335 /* 📦 Copy Pods Resources */,
```
garrettmoon: Shoot, this was added because I'm on cocoa pods 1.0. Will see if I can fix.

AsyncDisplayKit/ASDisplayNode+Deprecated.h
```diff
@@ -0,0 +1,51 @@
+//
+//  ASDisplayNode+Deprecated.h
+//  AsyncDisplayKit
+//
+//  Created by Garrett Moon on 8/26/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
```
maicki: Can you change the header to the default one we use in other places.

AsyncDisplayKit/ASDisplayNode+Deprecated.h
```diff
@@ -0,0 +1,51 @@
...
+
+#ifndef ASDisplayNode_Deprecated_h
+#define ASDisplayNode_Deprecated_h
```
maicki: Use #pragma once

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -238,36 +238,58 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)interfaceStateDidChange:(ASInterfaceState)newState fromState:(ASInterfaceState)oldState ASDISPLAYNODE_REQUIRES_SUPER;
 
 /**
- * @abstract Called whenever the visiblity of the node changed.
+ * @abstract Called whenever the node becomes visible.
  *
  * @discussion Subclasses may use this to monitor when they become visible.
+ *
+ * @note this method is guaranteed to be called on main.
```
maicki: Should we start with uppercase the "this ..."

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2432,26 +2432,38 @@ - (void)recursivelyClearFetchedData
   });
 }
 
-- (void)visibilityDidChange:(BOOL)isVisible
+- (void)didEnterVisibleState
 {
-  // subclass override
+  //subclass override
```
maicki: Why remove the " "?

garrettmoon: Was just trying to match them up. I can add one in all of them instead?

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -281,21 +281,31 @@ - (void)displayWillStart
...
  in ASMultiplexImageNode as well. */
-- (void)visibleStateDidChange:(BOOL)isVisible
+- (void)didEnterVisibleState
 {
-  [super visibleStateDidChange:isVisible];
-
+  [super didEnterVisibleState];
+  
   if (_downloaderFlags.downloaderImplementsSetPriority) {
     ASDN::MutexLocker l(__instanceLock__);
```
maicki: If we access the `_downloaderFlags` do we not also need to hold the `__instanceLock__ `?. Also in `didExitVisibleState`

AsyncDisplayKit/ASMultiplexImageNode.mm
```diff
@@ -293,20 +293,30 @@ - (void)displayWillStart
...
+/* didEnterVisibleState / didExitVisibleState in ASNetworkImageNode has a very similar implementation. Changes here are likely necessary
  in ASNetworkImageNode as well. */
-- (void)visibleStateDidChange:(BOOL)isVisible
+- (void)didEnterVisibleState
 {
-  [super visibleStateDidChange:isVisible];
+  [super didEnterVisibleState];
   
   if (_downloaderImplementsSetPriority) {
     ASDN::MutexLocker l(_downloadIdentifierLock);
```
maicki: Do we not hold the lock while accessing `_downloaderImplementsSetPriority`?

AsyncDisplayKit/ASNetworkImageNode.mm
garrettmoon: _downloaderFlags is only ever set on init.

AsyncDisplayKit/ASMultiplexImageNode.mm
garrettmoon: Also only set on init.

Adlai-Holler:
💯

maicki:
LGTM

Adlai-Holler:
Travis failure is spurious. Landing

PR #2132, opened 2016-08-24
Ensure subclasses of ASCollectionViewFlowLayoutInspector call super f…

garrettmoon:
…or didChange methods.


garrettmoon:
@maicki this should work, right?

Adlai-Holler:
I think the goal is to enforce it for non-flow inspectors – for any object that conforms to ASCollectionViewLayoutInspecting which is tough.

maicki:
Agree with Adlai, I think it's not a full solution and has some downsides.

The methods `didChangeCollectionViewDelegate:` and `didChangeCollectionViewDataSource:` are optional in the protocol. If we would add them in this way they become necessary to overwrite. We could provide a default implemenetation in the superclass, but as said it would not solve the problem for layout inspector that are not inherit from these two classes.

I think we have to think about issue from a different angle and put some more thinking into the whole architecture around `UICollectionViewLayout` and layout inspectors.

garrettmoon:
Yup, closing this for now.

PR #2133, opened 2016-08-24
Enable ASRangeControllerUpdateRangeProtocol by default if view contro…

garrettmoon:
…ller or node support it.


Adlai-Holler:
The old way was set up so that we don't emit the warning until the first range mode update, because Pinterest sets YES in its base view controller class, and then sets NO later for any VC subclass that doesn't use it. I'm cool with landing this but we should be careful to remove the application-level cases where it's explicitly set to avoid a bunch of warnings

Adlai-Holler:
Cool!

PR #2135, opened 2016-08-25
[Transition API] Add old transition API back but allow call from background thread

maicki:
We need a way to force a synchronous measurement in a layout transition. As the API we wanted to introduce in 2.0 does not didn't took that in consideration I think it makes the most sense to go back to the old API but not to force calling the layout transition API from the main thread.

@appleguy @levi @Adlai-Holler Would be great if you could take a look


Adlai-Holler:
LGTM

PR #2137, opened 2016-08-25
[ASViewController] Layout root node AFTER propagating all the traits

rcancro:
Wait until all subnodes have the new trait collection before laying out the root node. The way I’m using `asyncTraitCollectionDidChange`, I sometimes call methods that call `setNeedsLayout` which causes a new layout pass to begin before the new trait collection has been propagated to all nodes. This leads to weird behavoir, like nodes being laid out twice in the same call stack or being laid out with the node's previous constrainedSize (instead of the new size from the bounds). 

This change won’t stop possibly unnecessary layout passes because of `asyncTraitCollectionDidChange`, but it will at least do a complete layout pass once we have propagated all traits.

Ultimately the solution to this problem (as Adlai pointed out) is to wait until the CA transaction commit to perform the actual layout.


garrettmoon:
Seems like a good change regardless of any changes to setNeedsLayout.

Adlai-Holler:
Beautiful!

PR #2139, opened 2016-08-25
[ASStackLayoutSpec] Create Children NSArray Less Often

Adlai-Holler:
`self.children` causes the creation of an NSArray, so for vertical stacks, we were creating 4. This cuts us down to 1. Per @maicki in the future we should be even more aggressive but this is a start.

Ready for review @maicki


maicki:
LGTM! Get that in

PR #2142, opened 2016-08-25
Remove textStorageCreationBlock API

Adlai-Holler:
This API was a hack. In the future we will be reusing NSTextStorage objects and we need to have control over how they are created. 

Pinterest is no longer using this API.


maicki:
LGTM

PR #2143, opened 2016-08-26
[ASViewController] Remeasure Node With New Constrained Size When Propagating Traits

Adlai-Holler:
This fixes an issue where view controllers that participate in custom transitions may be the wrong size at the end. It comes with a unit test!

The issue was, view controllers in custom transitions are fullscreen in viewWillAppear: and when they get their correct size, we then overwrite it by calling -setNeedsLayout in -propagateNewEnvironmentTraitCollection, which remeasures us against the old bad size.

The solution is to measure the VC against the new constrained size when the new environment trait collection goes down.

This fixes Martin's issue, and it should go into 6.10. @garrettmoon @maicki @rcancro


Adlai-Holler:
Worked with @maicki on this – landing.

PR #2145, opened 2016-08-26
[ASDataController] Use 2 Threads Per CPU When Measuring Nodes

Adlai-Holler:
`dispatch_apply` always spawns only one thread per core. That's not enough for us!! So I made a variant that allows you to specify how many threads you want.

@appleguy @maicki @levi

### DIFF
--- a/AsyncDisplayKit/Private/ASDispatch.h
+++ b/AsyncDisplayKit/Private/ASDispatch.h
@@ -11,6 +11,9 @@
 
 /**
  * Like dispatch_apply, but you can set the thread count. 0 means 2*active CPUs.
+ *
+ * Note: The actual number of threads may be lower than threadCount, if libdispatch
+ * decides the system can't handle it. In reality this rarely happens.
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/Private/ASDispatch.h
```diff
@@ -0,0 +1,33 @@
+//
+//  ASDispatch.h
+//  AsyncDisplayKit
+//
+//  Created by Adlai Holler on 8/25/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "ASBaseDefines.h"
```
maicki: Do we need the import?

Adlai-Holler: Ah good catch. I had it ASDISPLAYNODE_INLINE before hence the import.

maicki:
LGTM

PR #2146, opened 2016-08-26
[ASTextNode] Factor in constrained size for final size calculation of ASTextNode

maicki:
cc @Adlai-Holler @garrettmoon


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -397,11 +401,12 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
...
   }
   
-  //add textContainerInset
+  // Add the constrained size back textContainerInset
   size.width += (_textContainerInset.left + _textContainerInset.right);
   size.height += (_textContainerInset.top + _textContainerInset.bottom);
   
-  return size;
+  return CGSizeMake(fminf(size.width, originalConstrainedSize.width),
+                    fminf(size.height, originalConstrainedSize.height));
```
Adlai-Holler: Use std::min or whatever here, so that it's correct on 64-bit

maicki: Done

Adlai-Holler:
Cool!

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -397,11 +401,12 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
     }
   }
   
-  //add textContainerInset
+  // Add the constrained size back textContainerInset
```
garrettmoon: This is actually adding back in the textContainerInset to the returned text size.

PR #2147, opened 2016-08-26
Fixes a deadlock caused by walking up the heirarchy.

garrettmoon:


Adlai-Holler:
Let's go with this. Could you put back the MutexLocker, and use a MutexUnlocker around __setNeedsLayout? That way if an exception gets thrown we'll unlock. Also for good measure, let's add a documentation comment above the __setNeedsLayout method to be careful to call this without the lock held.

garrettmoon:
Will update.

Adlai-Holler:
Would you add a comment above the actual __setNeedsLayout method saying it must be called with the lock NOT held?

PR #2148, opened 2016-08-26
[ASDisplayNode] Relinquish Lock Before Calling setNeedsLayout When Applying Pending State

Adlai-Holler:
Alternative to #2147 . We need to hold the lock the entire time that we're accessing the pending state.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1176,23 +1176,28 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (void)applyPendingViewState
 {
   ASDisplayNodeAssertMainThread();
-  ASDN::MutexLocker l(__instanceLock__);
+  BOOL shouldSetNeedsLayout = NO;
+  {
+    ASDN::MutexLocker l(__instanceLock__);
+    shouldSetNeedsLayout = _pendingViewState.hasSetNeedsLayout;
 
+    if (self.layerBacked) {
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1176,23 +1176,28 @@ - (void)setContentsScaleForDisplay:(CGFloat)contentsScaleForDisplay
...
 
+    [_pendingViewState clearChanges];
+  }
+
+  // NOTE: It's important to call this without holding the lock.
   // FIXME: Ideally we'd call this as soon as the node receives -setNeedsLayout
   // but automatic subnode management would require us to modify the node tree
   // in the background on a loaded node, which isn't currently supported.
-  if (_pendingViewState.hasSetNeedsLayout) {
+  if (shouldSetNeedsLayout) {
```
garrettmoon: Is it ok that setNeedsLayout has moved to after the pending state has been applied? Currently setNeedLayout sometimes lays out synchronously?

Adlai-Holler: Yeah it's OK – in fact it's probably better to have all the view/layer attributes updates before the layout pass. Any thoughts about risk here?

garrettmoon: I think the biggest risk is moving the call ordering, otherwise, LGTM!

Adlai-Holler: You're right, we should call __setNeedsLayout before potentially calling setNeedsLayout on the view/layer to be safe.

PR #2149, opened 2016-08-26
[ASDisplayNode] If We Skipped Rendering Due to Zero-Area, Re-Render When We Get a Real Area

Adlai-Holler:


Adlai-Holler:
This should be brought to Pinterest ASAP. Ready for review @maicki @appleguy

maicki:
LGTM

PR #2150, opened 2016-08-26
I wrote a bunch of code with the Mutex::Unlocker and didn't understan…

garrettmoon:
…d it.


Adlai-Holler:
Great catches – let's get these in Pinterest ASAP.

PR #2151, opened 2016-08-26
[ASTextNode] Actually Deprecate attributedString Property

Adlai-Holler:
This was intended to be deprecated but someone wimped out because we never actually did it.

Ready for review. Obviously we'll have to put in some effort to bring to Pinterest.


maicki:
LGTM

PR #2155, opened 2016-08-27
[ASDisplayNode] Dealloc _pendingViewState if range managed

garrettmoon:
_ASPendingState objects can add up very quickly when adding
many nodes. This is especially an issue in large collection views
and table views. This needs to be weighed against the cost of
reallocing a _ASPendingState. So in range managed nodes we
delete the pending state, otherwise we just clear it.

@Adlai-Holler @appleguy Does this patch make sense? Is it safe?

Profiling Pinterest and just scrolling down these start to add up.
With ~500 pin nodes and a total of 115MiB, pending state made
up 3.52MiB. With this patch, no matter how far you scroll, pending
state stays at about 500KiB.

Separately, I looked at allocation of nodes vs pending states
created when scrolling and they approximately matched when
scrolling the homefeed suggesting there's no cost to recreating them?

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -106,15 +106,7 @@ _ASPendingState *ASDisplayNodeGetPendingState(ASDisplayNode *node)
   ASDN::MutexLocker l(node->__instanceLock__);
   _ASPendingState *result = node->_pendingViewState;
   if (result == nil) {
-    if (node.isNodeLoaded) {
-      if (node->_flags.layerBacked)
-        result = [_ASPendingState pendingViewStateFromLayer:node->_layer];
-      else
-        result = [_ASPendingState pendingViewStateFromView:node->_view];
-    } else {
-      result = [[_ASPendingState alloc] init];
-    }
-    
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
@garrettmoon I like this change, but we should add ASDisplayNodeAssertMainThread to `pendingStateFromView` and `pendingStateFromLayer`. I think `ASDisplayNodeGetPendingState` is called on arbitrary threads =/

Adlai-Holler:
We should replace https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm#L57 with ASDisplayNodeGetPendingState as well.

garrettmoon:
I think you're right about the arbitrary threads… need to think this through some more…

Adlai-Holler:
@garrettmoon On second thought, this may be doable – we don't need to use pendingStateFromView/Layer except when the view is being unloaded. ASPendingState only tracks fields that have been changed, so it's OK to put back [[ASPendingState alloc] init] here.

garrettmoon:
Awesome, yeah that makes sense! I'll update.

Adlai-Holler:
Awesome!

PR #2158, opened 2016-08-28
[Testing] Remove header for ASXCTExtensions

maicki:


Adlai-Holler:
Hmm GitHub must've screwed up here – the header content that this diff removes has already been removed from ASXTExtensions.h…

PR #2159, opened 2016-08-28
[ASViewController] Allow block to determine if traits should propagate

rcancro:
By default, traits propagate when `ASEnvironmentTraitCollectionIsEqualToASEnvironmentTraitCollection` returns NO. Because of the way the ASDK traits are constructed (we know the new container size at `viewWillTransitionToSize` but we don’t know the size classes until `traitCollectionDidChange`) using `ASEnvironmentTraitCollectionIsEqualToASEnvironmentTraitCollection` can result in the traits being propagated multiple times for 1 trait change. By creating a comparison block, the user can decide when the traits propagate. For example, I only want to propagate traits when the `containerSize` changes — I don’t care about the size classes because I don’t use them. I can set this block to only propagate in this case and avoid unnecessary propagation.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -174,6 +174,7 @@
 		509E68641B3AEDB7009B9150 /* ASCollectionViewLayoutController.mm in Sources */ = {isa = PBXBuildFile; fileRef = 205F0E1C1B373A2C007741D0 /* ASCollectionViewLayoutController.mm */; };
 		509E68651B3AEDC5009B9150 /* CGRect+ASConvenience.h in Headers */ = {isa = PBXBuildFile; fileRef = 205F0E1F1B376416007741D0 /* CGRect+ASConvenience.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		509E68661B3AEDD7009B9150 /* CGRect+ASConvenience.m in Sources */ = {isa = PBXBuildFile; fileRef = 205F0E201B376416007741D0 /* CGRect+ASConvenience.m */; };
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
What's the motivation for this change – aside from performance, are the extra propagations causing problems? Also would you be willing to add a unit test for the new feature?

rcancro:
It probably isn't completely necessary. But I've noticed some inconsistent behavior with UIKit when using split screen that fixing on my end out leads to more propagations than we are already doing. (In a case where size classes change we usually end up propagating twice -- once in `willTransitionToTraitCollection` where we don't have the right container size, and again in `viewWillTransitionToSize` where we do.)

For example, I've noticed that UIKit will sometimes send some strange sizes to VCs when resizing. When transitioning from 50/50 to my VC being full screen with a pull out on the side I got a size of `{1541, 768}` from UIKit. This leads to VCs being the wrong size after transitioning. In order to combat this, I've changed our tab bar VC to pass down its size to its childVCs to make sure they have the right size. This can lead to yet another propagation.

In my implementation I don't care about size classes at all, and only want to propagate when the container size changes. I thought adding a block to let the user decide when to propagate could be useful. But I understand if it seems too special case.

Adlai-Holler:
It looks like a bunch of CocoaPods changes got pulled in with this. Could you revert those?

Rather than an equality block, how about an overridable method like 

```
/**
 * Returns YES if `environmentTraitCollection` should be propagated down the node hierarchy.
 * 
 * The default implementation returns YES if the new trait collection does not equal the current trait collection. Subclasses can override this method to suppress trait collection changes that
 * are not worth propagating e.g. only propagate container size changes.
 */
- (BOOL)shouldPropagateNewEnvironmentTraitCollection:(ASEnvironmentTraitCollection)environmentTraitCollection;
```

Adlai-Holler:
Superseded by #2202

PR #2163, opened 2016-08-28
[Layout] Automatic measure pass in layout pass if not happened before

maicki:
If no measure pass happened or the bounds changed between layout passes we manually trigger a measurement pass for the node using a size range equal to whatever bounds were provided to the node

### DIFF
--- a/AsyncDisplayKitTests/ASDisplayNodeLayoutTests.mm
+++ b/AsyncDisplayKitTests/ASDisplayNodeLayoutTests.mm
@@ -99,4 +99,30 @@
 }
 #endif
 
+- (void)testMeasureOnLayoutIfNotHappenedBeforeNoRemeasureForSameBounds
+{
+  CGSize nodeSize = CGSizeMake(100, 100);
+  
+  ASStaticSizeDisplayNode *displayNode = [ASStaticSizeDisplayNode new];
+  displayNode.staticSize  = nodeSize;
+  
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
Nice! While you're at it, could you add w test method that ensures that on a second layout pass with same bounds, layoutSpecBlock is not called? Just for good measure

maicki:
@adlai Added a new test. Let me know if you know a better name for it 😃

AsyncDisplayKitTests/ASDisplayNodeLayoutTests.mm
```diff
@@ -99,4 +99,29 @@ - (void)testNotAllowModifyingSubnodesInLayoutSpecThatFits
...
+    return [ASInsetLayoutSpec insetLayoutSpecWithInsets:UIEdgeInsetsZero child:buttonNode];
+  };
+  
+  displayNode.frame = {.size = nodeSize};
+  
+  // Trigger initial layout pass without a measurement pass before
+  [displayNode.view layoutIfNeeded];
+  
+  [displayNode measureWithSizeRange:ASSizeRangeMake(nodeSize, nodeSize)];
+  XCTAssertTrue(numberOfLayoutSpecThatFitsCalls == 1, @"Should not remeasure with same bounds");
```
Adlai-Holler: Does this actually test that we measure on layout? I see that it tests that we only measure once, but I don't think it tests that we measure on layoutIfNeeded.

PR #2164, opened 2016-08-28
[Layout] Assert if subnodes where modified during layoutSpecThatFits:

maicki:
It's not allowed to modify submodes in `layoutSpecThatFits:` or in the `layoutSpecBlock` block. This PR adds an assertion if DEBUG enabled to assert if this is happening.


Adlai-Holler:
Nice!

PR #2166, opened 2016-08-30
[ASTextNode] Add C++ LRU cache for ASTextKitRenderer

maicki:
Next round of ASTextNode optimization ... adding an LRU cache for caching `ASTextKitRenderer` in the `ASTextNode`.

@Adlai-Holler Let's run it against our tests ...


maicki:
@Adlai-Holler Let's adjust / add performance tests to consider the caching of the text kit renderer so we have a bit of a better overview what is going on.

Adlai-Holler:
@maicki Love the performance, love not having the node own the renderer, but there's a whole lot of (C++) code added in this diff. I think we can achieve a similar result with NSCache, with just the attributed string as the key.

NSCache is LRU (libcache) and I doubt that rendering the same attributed string with different constrained sizes is common enough to justify all this effort. Plus, since the constrained size actually changes when text node is measured (which I still think is dumb!) there may be a lot of cache misses that could have been hits.

Thoughts?

maicki:
@Adlai-Holler Changed the cache to NSCache in this PR: #2166 . I don't like that we have to create a key object on the heap to be honest. The C++ implementation let us create the key on the stack. 

Furthermore we have to have to consider the constrained size in the cache key, the attributed string is not enough.

Adlai-Holler:
Superseded by #2199 that is simpler and performs better in our benchmark.

PR #2167, opened 2016-08-30
make inline functions in ASDisplayNodeExtras.h static 

samhsiung:
Resolves linker errors and allows inline functions such as `ASInterfaceStateIncludesVisible` to be used outside of ASDK framework files.

Also according to this:
http://stackoverflow.com/questions/18623392/how-to-write-global-inline-functions-in-objective-c-using-c-syntax

Global inline functions should be static

### DIFF
--- a/AsyncDisplayKit/ASDisplayNodeExtras.h
+++ b/AsyncDisplayKit/ASDisplayNodeExtras.h
@@ -16,27 +16,27 @@
 
 // Because inline methods can't be extern'd and need to be part of the translation unit of code
 // that compiles with them to actually inline, we both declare and define these in the header.
-inline BOOL ASInterfaceStateIncludesVisible(ASInterfaceState interfaceState)
+static inline BOOL ASInterfaceStateIncludesVisible(ASInterfaceState interfaceState)
 {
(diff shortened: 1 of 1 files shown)


maicki:
To be honest I don't really like the idea to expose this functions. We should have public accessors in `ASDisplayNode` where you can get the state of it. @Adlai-Holler thoughts?

samhsiung:
@maicki that sounds good to me. This PR would at least be an upgrade because the inline functions are currently global and accessible, but you will get a linker error if you try to use it.

samhsiung:
Was the point of inlining these functions solely for performance? There might be a very slight performance hit if you wrap the same checks behind a property or method. Originally created this PR due to @appleguy 's suggestion.

Adlai-Holler:
We have a ship task to add the readonly node properties. This header is currently public. If we want to hide these, we should mark the file private – I agree with @samhsiung that these should be static inlines and they're useful for the time being.

@samhsiung could you declare them ASDISPLAYNODE_INLINE?

Adlai-Holler:
Cool thanks @samhsiung !

PR #2168, opened 2016-08-30
[ASDisplayNode] Add Convenience Interface State Accessors

Adlai-Holler:
This also standardizes on Preload as the name of the range.


maicki:
💯

PR #2171, opened 2016-08-31
Fix Case Where Node Is Deallocated While Visible

### LINKED ISSUE #2011: [ASCollectionView] reloadData: Node should always be marked invisible before deallocating
Hello,
i have a problem with my ASCollectionView; when the user taps a button that makes the _collectionView.reloadData() i get the
`*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Node should always be marked invisible before deallocating; interfaceState: 15, <ASImageNode: 0x7ffb8a07a600>'
` error

I simply get fresh data and do a reloadData() . How can i solve it?
It seems similar to #1466 but this time is for ASCollectionView.
I'm using 1.9.81 version.
Can anyone help me please?
Thanks in advance and greetings
Cris

Adlai-Holler:
See #2011 .

The issue is, if the view outlives the node and its subnodes, then its sublayers will not get the order-out action until after willMoveToSuperview.

So I moved the `__exitHierarchy` call from `willMoveToSuperview` to `didMoveToSuperview` which fixes the issue!

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -389,9 +389,11 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (void)dealloc
 {
   ASDisplayNodeAssertMainThread();
+
+  NSLog(@"%@ dealloc.", self.name);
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
Ready for review @maicki ! Once landed we should bring this to everywhere.

maicki:
LGTM

PR #2172, opened 2016-08-31
[Layout] Add way to validate ASLayoutables in a composable way

maicki:
The current problem ASDK in the current layout system abstraction is that developer can set all `ASLayoutable` properties on a `ASDisplayNode` or `ASLayoutSpec`, but only some properties work in `ASStackLayoutSpec` and some in `ASStaticLayoutSpec`. To prevent setting properties to `ASLayoutable` that will not have any effect in the layout spec that they will be inserted we add layoutable validation to our layout system for the time being.

The idea is to have `ASLayoutSpec` to reject every `ASLayoutable` child that has any properties that are related to `ASStackLayoutSpec` or `ASStaticLayoutSpec`. This would allow validation for children that are e.g added to a `ASBackgroundLayoutSpec`. `ASStaticLayoutSpec` and `ASStackLayoutSpec` overwrite the validation behavior and reject children that are not related to itself.

### Changes:
- [x] Add API for validation of `ASLayoutSpec` children
- [x] Add disable validation for `ASLayoutable` and `ASLayoutSpec`
- [x] Add tests for validation

**This is not ready for merge at the moment**


maicki:
@Adlai-Holler Ready for first review.

AsyncDisplayKit/Layout/ASLayoutable.h
```diff
@@ -67,6 +67,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Layout options from the Layoutable Protocols
 
+#pragma mark - ASLayoutValidation
+
+@property (nonatomic, readwrite) BOOL shouldValidate;
```
Adlai-Holler: Add some documentation to this property e.g.:

```
Whether or not this object should validate its configuration. Defaults to YES.

@discussion AsyncDisplayKit can help catch layout issues early by checking that
layout objects (nodes and layout specs) are configured in a way that makes sense against
their environment. For example, setting `layoutPosition` on a display node only makes sense
if that node is used in a static layout spec.
```

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -90,6 +94,21 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
     id<ASLayoutable> finalLayoutable = [child finalLayoutable];
+    
+#if LAYOUT_VALIDATION
+    if (self.shouldValidate && finalLayoutable.shouldValidate) {
+      for (ASLayoutableValidationBlock validatorBlock in [self layoutableValidatorBlocks]) {
+        NSError *error = nil;
+        BOOL success = validatorBlock(finalLayoutable, &error);
+        
+        if (error != nil) {
+          ASDisplayNodeCAssert(success, error.localizedDescription);
```
Adlai-Holler: I think it would be valuable to modify the error message with a way for them to suppress validation if it becomes bothersome in a particular case. E.g. append to the end of the error message `To suppress this issue, set 'shouldValidate=NO on <ASStackLayoutSpec: 0xFFFFFFFF>`. Man we should really support debug names on these specs.

AsyncDisplayKit/Layout/ASLayoutableValidation.mm
```diff
@@ -0,0 +1,105 @@
...
+
+#import "ASLayoutableValidationBlockProvider.h"
+#import "ASStaticLayoutSpec.h"
+#import "ASStackLayoutSpec.h"
+
+NSString * const ASLayoutableValidationErrorDomain = @"ASLayoutableValidationErrorDomain";
+
+#pragma mark - Helpers
+
+static NSString *ASLayoutValidationWrappingErrorMessage(SEL selector, id obj, Class cl)
```
Adlai-Holler: There's some interchanging between the terms `ASLayoutValidation` and `ASLayoutableValidation`. I think `ASLayoutValidation` is the better choice.

Adlai-Holler:
I love the intention here, but I'm somewhat concerned that this will irritate users in full-size production apps. For instance, if a view controller is passed an image node to use after a custom transition, it may get validation errors due to the kind of layout that that node participated in beforehand. It may not be easy to track this down, and the best recourse in that case would be to disable validation for its owning layout spec.

It's tragic, but I think we should close this until our layout API is more declarative i.e. the implementor of layoutSpecThatFits has complete control over the validity of the layout.

Adlai-Holler:
After some discussion, Michael and I decided to close this for now. The API we're validating against is currently changing a lot, and we'd like to provide compile-time validation instead, which may be possible soon.

PR #2173, opened 2016-08-31
Skip Performance Tests By Default

Adlai-Holler:
This will skip the performance tests when you ⌘U or when the CI runs "xcodebuild test".

You can still run/profile them whenever you want by clicking the little diamond. In the future we may remove the expected performance ranges, but for now I'm leaving them.


maicki:
LGTM

PR #2179, opened 2016-09-01
[ASDisplayNode] Fix using ASDisplayNodePerformBlockOnEveryNode if node is loaded and has shouldRasterizeDescendants enabled

maicki:
We don't call the block on every subnode if the node that is passed in ASDisplayNodePerformBlockOnEveryNode has shouldRasterizeDescendants enabled as it tries to iterate through the sublayer hierarchy, but there are no sublayers


hebertialmeida:
Thanks @maicki, you are the best!

Issue #1642, opened 2016-05-10
[ASPagerNode] Convenient way to initialize an ASPagerNode's initial cell


plm75:
Currently, the only way to set an ASPagerNode's initial cell is to call `scrollToItemAtIndex`.  However, this method can only be called after the ASPagerNode's data has been fully loaded and the view laid out.

One workaround consists in wrapping the call to `scrollToItemAtIndex` inside an `async_after`, another is to call it after `reloadDataImmediately`.

The first workaround is just an ugly hack, the second is too expensive.  A better and more convenient way to set an ASPagerNode's initial cell would be nice.

cc @hannahmbanana

maicki:
@plm75 Agree dispatch_after is a hack and I would recommend try to not using it. Just fyi: you can also use `scrollToItemAtIndex:` in `reloadDataWithCompletion:`, this would not be that be expensive as `reloadDataImmediately`.

Jauzee:
@plm75 Have you got a solution for this issue?

Currently I'm use `scrollToPageAtIndex` method in `viewDidAppear` and it looks very ugly, I'm thinking about make some loader & fade in, but maybe it's not best idea in this case.

appleguy:
A much, much better option is to manually calculate the content offset.  You can use setContentOffset: right away.  This would probably require manually setting the contentSize to at least that amount, though; it will then be reset as the collection loads its content for the pager.

A tricker but even better solution would be to initially add just the page you want to show to the datasource (as the /only/ page), so it shows up quickly, but then insert all the other pages and scroll to the correct page.  Done carefully, this could be really slick.

Overall, make sure you try to avoid exposing many hundred or thousands of items to the data source all at once.  If you can limit that to under a hundred, it is usually acceptable to use -waitUntilAllTransactionsAreCommitted in viewWillAppear: and then you can scroll to the correct page immediately.

Jauzee:
@appleguy thank you for detailed answer, I've been used this for a while.

Any updates in ASDK 2.0?

garrettmoon:
Since there seem to be several work arounds for this, I'm marking it as a P3

garrettmoon:
This issue was moved to TextureGroup/Texture#133

PR #2186, opened 2016-09-03
[ASCollectionView] Add Method to Wait for Updates During Next Layout Pass

### LINKED ISSUE #1584: [ASCollectionNode / ASTableNode] Method to Wait for Initial Nodes to Be Loaded
`waitUntilAllUpdatesAreCommitted` is great! Could we go even further and make a variant that waits until the initial nodes are all measured and inserted as well?

I've got a pager node with 3 tabs, and I'd like to push a VC already jumped to the third tab, but I can't. I add `waitUntilAllUpdatedAreCommitted` in `viewWillAppear` and it has no effect, because `_editingTransactionQueue` is empty but numberOfSections is 0. `viewDidLayoutSubviews` is no help either because my view controller's view is a container that holds the pager inside it. Currently my best option is to async out of `viewWillAppear:` and hope that by then the nodes are ready, else to poll repeatedly and jump as soon as I see `numberOfItemsInSection:` is correct but these aren't ideal.

@nguyenhuy Any advice?

Adlai-Holler:
During the normal flow, the initial data is loaded during the first layout pass. This adds a flag for the user to tell us "when the next layout pass happens, block the main thread until the data is up-to-date".

- Simplifies our ASCollectionView testing setup by combining ASCollectionViewTestController and ASCollectionViewTestDelegate into just the controller.
- Fixes an issue where `waitUntilAllUpdatesAreCommitted` fails if the initial reload hasn't been performed yet.
- Minor: makes `reloadData` only be callable on the main thread for ASCollectionView. It's a pretty small API change, since all of our other editing operations are already required to be on the main thread. It was already required for table view.

Resolves #1584 and essentially #1642 also.

Ready for review!

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -52,11 +52,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) id<ASCollectionDataSource> asyncDataSource;
 
 /**
- * Whether the view should block the main thread during the next layout pass while
- * any pending data is loaded. This is useful if you want to ensure that the next frame
- * that is composited with this view will have all the nodes.
+ * Tell the collection view that, during the next layoutSubviews pass, it should block the
+ * main thread and wait for all items to be up-to-date.
  *
- * This method must be called on the main thread.
+ * @discussion This is useful when you want to ensure that the user never sees an empty
+ * collection view. It is better to call this than to explicitly wait using `waitUntilAllUpdatesAreCommitted`
+ * because the collection view's size may not be correct until the layout pass.
  */
 - (void)waitForUpdatesDuringNextLayoutPass;
 
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -52,13 +52,13 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) id<ASCollectionDataSource> asyncDataSource;
 
 /**
- * Whether the view should block the main thread for the initial data load.
+ * Whether the view should block the main thread during the next layout pass while
+ * any pending data is loaded. This is useful if you want to ensure that the next frame
+ * that is composited with this view will have all the nodes.
  *
- * If YES, the view will wait for the initial batch of nodes to be allocated and measured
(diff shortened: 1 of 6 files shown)


Adlai-Holler:
@maicki @levi @appleguy thoughts? It'll be important to raise the visibility of this option because it's very important. We may even want to connect it to ASViewController's `neverShowPlaceholders` flag since this is part of that.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -520,6 +533,13 @@ - (void)selectItemAtIndexPath:(NSIndexPath *)indexPath animated:(BOOL)animated s
   [super selectItemAtIndexPath:indexPath animated:animated scrollPosition:scrollPosition];
 }
 
+- (void)waitForUpdatesDuringNextLayoutPass
+{
+  ASDisplayNodeAssertMainThread();
+  [self setNeedsLayout];
```
appleguy: @Adlai-Holler should this actually set a new layout pass as being needed?  The name doesn't currently imply that.

Adlai-Holler: You're absolutely right – that was a holdover from some previous implementation I believe. Removed.

Adlai-Holler:
Ready for next review @maicki @appleguy @garrettmoon

maicki:
LGTM

maicki:
LGTM

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -52,6 +52,16 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) id<ASCollectionDataSource> asyncDataSource;
 
 /**
+ * Tell the collection view that, during the next layoutSubviews pass, it should block the
+ * main thread and wait for all items to be up-to-date.
+ *
+ * @discussion This is useful when you want to ensure that the user never sees an empty
+ * collection view. It is better to call this than to explicitly wait using `waitUntilAllUpdatesAreCommitted`
```
garrettmoon: Can we add a `@warning` in the waitUntilAllUpdatesAreCommitted method indicating they might want to use this method instead?

Adlai-Holler:
This will be changed so much in light of the 2.0 ASCollectionView/ASTableView -> Node diff that it may as well be closed. Sayonara.

PR #2187, opened 2016-09-03
[ASStackLayoutSpec] Set alignItems to ASStackLayoutAlignItemsStretch

maicki:
To align with the CSS `align-items` property we should set alignItems to `ASStackLayoutAlignItemsStretch` instead of `ASStackLayoutAlignItemsStart`: https://drafts.csswg.org/css-flexbox-1/#propdef-align-items


Adlai-Holler:
Love it!

PR #2195, opened 2016-09-05
[Documentation] Add ASDK 2.0 Layout Documention

maicki:
Created this PR so we have a discussion place about the ASDK 2.0 Layout Documentations.

### DIFF
--- a/docs/ASDK-2.0-Layout-Documention.md
+++ b/docs/ASDK-2.0-Layout-Documention.md
@@ -1,0 +1,802 @@
+# ASDK 2.0 Layout
+AsyncDisplayKit's automatic layout system is pretty powerful and loosely based on the CSS Box Model that a lot of developer are familiar from the web. Furthermore it provides components, so called layout specs, to layout and compose elements, called layoutables, in vastly different ways like in a flexbox or absolute fashion.
+
+## Table of Contents
+* [Differences from web](#differences-from-web)
+  * [Naming properties](#naming-properties)
+  * [No margin / padding properties](#no-margin--padding-properties)
+  * [Missing features](#missing-features)
+* [Layoutables](#layoutables)
+  * [Replaced Layoutables](#replaced-layoutables)
+* [Height and Width](#height-and-width)
+  * [The content area](#the-content-area)
+  * [Setting a size](#setting-a-size)
+  * [Sizing Dimensions](#sizing-dimensions)
+  * [Layout with Flexbox](#layout-with-flexbox)
+  * [Absolute Layout](#absolute-layout)
+* [Layout Properties](#layout-properties)
+  * [The ASLayoutable protocol](#the-aslayoutable-protocol)
+    * [@property (nonatomic, assign, readwrite) ASDimension width;](#property-nonatomic-assign-readwrite-asdimension-width)
+    * [@property (nonatomic, assign, readwrite) ASDimension height;](#property-nonatomic-assign-readwrite-asdimension-height)
+    * [@property (nonatomic, assign, readwrite) ASDimension minHeight;](#property-nonatomic-assign-readwrite-asdimension-minheight)
+    * [@property (nonatomic, assign, readwrite) ASDimension maxHeight;](#property-nonatomic-assign-readwrite-asdimension-maxheight)
+    * [@property (nonatomic, assign, readwrite) ASDimension minWidth;](#property-nonatomic-assign-readwrite-asdimension-minwidth)
+    * [@property (nonatomic, assign, readwrite) ASDimension maxWidth;](#property-nonatomic-assign-readwrite-asdimension-maxwidth)
+  * [The ASStackLayoutSpec protocol](#the-asstacklayoutspec-protocol)
+    * [@property (nonatomic, assign) ASStackLayoutDirection direction;](#property-nonatomic-assign-asstacklayoutdirection-direction)
+    * [@property (nonatomic, assign) CGFloat spacing;](#property-nonatomic-assign-cgfloat-spacing)
+    * [@property (nonatomic, assign) ASHorizontalAlignment horizontalAlignment;](#property-nonatomic-assign-ashorizontalalignment-horizontalalignment)
+    * [@property (nonatomic, assign) ASVerticalAlignment verticalAlignment;](#property-nonatomic-assign-asverticalalignment-verticalalignment)
+    * [@property (nonatomic, assign) ASStackLayoutJustifyContent justifyContent;](#property-nonatomic-assign-asstacklayoutjustifycontent-justifycontent)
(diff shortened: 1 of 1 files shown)


docs/ASDK-2.0-Layout-Documention.md
```diff
@@ -0,0 +1,802 @@
...
+  * [Manual Layout UIKit](#manual-layout-uikit)
+  * [Manual Layout ASDK](#manual-layout-asdk)
+    * [calculateSizeThatFits and <code>layout</code>](#calculatesizethatfits-and-layout)
+    * [Example](#example)
+* [Ressources:](#ressources)
+
+## Introduction
+// TODO: Small introduction with an example
+
+## Differences from web
```
levi: This section spend a lot of time explaining what the layout system _isn't_, but not a lot of what it really is. Best to introduce newcomers on the strengths, benefits and positive elements of the framework before directly comparing it to something else like the CSS Box Model. Here is really the time to sell and convince them of how simple and understandable the whole system is.

docs/ASDK-2.0-Layout-Documention.md
```diff
@@ -0,0 +1,802 @@
...
+  1. the size (but not position) of the layoutable object and
+  2. the size and position of all of its immediate child objects.
+
+`ASDisplayNode` as well as `ASLayoutSpec` classes are conforming to the `ASLayoutable` protocol and provide this functionally out of the box. This allows you to compose nodes and layout specs to build very complex layouts in a declarative way.
+
+### Replaced Layoutables
+A replaced layoutable is any layoutable whose appearance and dimensions are defined by an external resource. Examples include `ASImageNode`, `ASTextNode`, `ASButtonNode` or `ASTextNode`. All other layoutables can be referred to as non-replaced layoutables. This would include `ASVideoNode`, `ASVideoPlayerNode`, `ASNetworkImageNode` or `ASEditableTextNode`.
+
+Replaced layoutables can have intrinsic dimensions—width and height values that are defined by the layoutable itself, rather than by its surroundings. For example, if an `ASImageNode` has a width set to `ASDimensionAuto`, the width of the linked image file will be used. For an `ASImageNode` the intrinsic content size would be calculated based on the text content.
```
levi: I'm not sure if I truly understand "replaced layouts" based on these two paragraphs. It first says that it's a layoutable whose appearance and dimensions are defined by an external resource, then says that it can have intrinsic size that can be defined by itself rather than by its surroundings. I'm at odds on what I'm supposed to understand.

Could you discuss with me some examples that support these two cases and I think we'll be able to identify a simple, common definition that will be able to accurately describe the behavior of these types of layoutables.

docs/ASDK-2.0-Layout-Documention.md
```diff
@@ -0,0 +1,802 @@
...
+
+## Height and Width
+
+### The content area
+The content area is the area containing the real content of a layoutable. The properties `width`, `minWidth`, `maxWidth`, `height`, `minHeight` and `maxHeight` control the content size.
+
+### Setting a size
+A layoutable's height and width dimensions determine its size on the screen. The simplest way to set the dimensions of a layoutable is by setting the fixed `width` and `height` property. All dimensions in AsyncDisplayKit's layout system are unitless, and represent density-independent pixels. Setting dimensions this way is common for layoutables that should always render at exactly the same size, regardless of screen dimensions.
```
levi: `All dimensions in AsyncDisplayKit's layout system are unitless, and represent density-independent pixels. Setting dimensions this way is common for layoutables that should always render at exactly the same size, regardless of screen dimensions.`

Would suggest introducing the percent and point dimensions and go into the differences, rather than describing density-independent pixels. Namely, because developers will think they have already been working with density-independent pixels with UIKit. We should retrain their thinking when representing that you can not just have points, but also percents — which is why the layout system uses ASDimensions and not CGFloat.

docs/ASDK-2.0-Layout-Documention.md
```diff
@@ -0,0 +1,802 @@
...
+
+```Obj-c
+// Set a height of 0.5 points
+layoutable.width = ASDimensionMake(ASDimensionTypePoints, 0.5);
+
+// Set a width of 50%
+layoutable.width = ASDimensionMake(ASDimensionTypePercentage, 0.5);
+```
+
+Besides width and height there is also `minWidth`, `minHeight`, `maxWidth` and `maxHeight` properties. For example an layoutable to which `minWidth` is applied will never be narrower than the minimum width specified, but it will be allowed to grow normally if its content exceeds the minimum width set. For a full description look at the `Layout Properties` section.
```
levi: "In addition to width and height, layoutables provide `minWidth`, `minHeight`, `maxWidth` and `maxHeight` properties."

docs/ASDK-2.0-Layout-Documention.md
```diff
@@ -0,0 +1,802 @@
...
+
+### Sizing Dimensions
+Setting dimensions on layoutables are of the type `ASDimension`. `ASDimension` is essentially a normal CGFloat with support for representing either a point value, or a % value. It allows the same API to take in both fixed values, as well as relative ones:
```
levi: The reader should already understand that dimensions are "essentially a normal CGFloat with support for representing either a point value, or a % value." at this point if it's described in the intro. Focus on the fact that this concept is represented by the `ASDimension` type and all the additional utilities that come with it.

docs/ASDK-2.0-Layout-Documention.md
```diff
@@ -0,0 +1,802 @@
...
+ */
+typedef NS_ENUM(NSInteger, ASDimensionType) {
+  /** This indicates "I have no opinion" and may be resolved in whatever way makes most sense given the circumstances. */
+  ASDimensionTypeAuto,
+  /** Just a number. It will always resolve to exactly this amount. This is the default type. */
+  ASDimensionTypePoints,
+  /** Multiplied to a provided parent amount to resolve a final amount. */
+  ASDimensionTypeFraction,
+};
+```
```
levi: Rather than providing a header dump, add each case as a title and write a little bit about its utility. ie.

**ASDimensionTypeAuto**
This is the default dimension. When defined, it allows the layoutable to define the dimension that will fit the best in the current layout context. For example, a `height` set to `Auto` when a specific dimension is defined for minHeight and maxHeight will ensure the layoutable's height is never fixed.

docs/ASDK-2.0-Layout-Documention.md
```diff
@@ -0,0 +1,802 @@
...
+
+When a relative percentage (%) value is used, it is resolved against the size of the parent. For example, an item with 50% flexBasis will ultimately have a point value set on it at the time that the stack achieves a concrete size.
```
levi: Might be a little confusing to go into this much implementation detail to describe %. Particularly, I was lost at `an item with 50% flexBasis`, since I haven't yet been introduced to the stack layout and what flexBasis means.

hannahmbanana:
I moved Levi's comments to canvas.

PR #2197, opened 2016-09-06
[ASImageNode] Use kCGBlendModeCopy when Possible

Adlai-Holler:
By default, images are drawn with the normal blending mode. kCGBlendModeCopy is a lot faster when you can use it.

~~Compare the following results from 30 seconds of Pinterest, the first using normal and the second using copy. Currently PINRemoteImage returns images that falsely have alpha components so for this test, I ALWAYS used copy. So this is basically a theoretical maximum. But most of our large images are JPEGs so we should be able to get close.~~

|                              | Normal | Copy   |
|------------------------------|--------|--------|
| ImageNode Drawing            | 3361ms | 1813ms |
| TextNode Drawing (reference) | 353ms  | 405ms  |
| Ratio                        | 9.5x   | 4.48x  |

~~So image drawing, which is the most expensive operation we do, is twice as fast using copy.~~

The test results above are probably skewed because the image caches were warmed, and so less progressive rendering happened. But it's way better.

@appleguy @levi @maicki

### DIFF
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -437,15 +437,19 @@ static ASDN::Mutex cacheLock;
   // will do its rounding on pixel instead of point boundaries
   UIGraphicsBeginImageContextWithOptions(key.backingSize, key.isOpaque, 1.0);
   
+  BOOL contextIsClean = YES;
+  
   CGContextRef context = UIGraphicsGetCurrentContext();
   if (context && key.preContextBlock) {
     key.preContextBlock(context);
+    contextIsClean = NO;
   }
   
   // if view is opaque, fill the context with background color
   if (key.isOpaque && key.backgroundColor) {
     [key.backgroundColor setFill];
     UIRectFill({ .size = key.backingSize });
+    contextIsClean = NO;
   }
   
   // iOS 9 appears to contain a thread safety regression when drawing the same CGImageRef on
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASImageNode.mm
```diff
@@ -460,8 +464,12 @@ + (UIImage *)createContentsForkey:(ASImageNodeContentsKey *)key isCancelled:(asd
   // Another option is to have ASDisplayNode+AsyncDisplay coordinate these cases, and share the decoded buffer.
   // Details tracked in https://github.com/facebook/AsyncDisplayKit/issues/1068
   
-  @synchronized(key.image) {
-    [key.image drawInRect:key.imageDrawRect];
+  // We want to use copy, but if our image has alpha and our context is dirty, we have to use normal.
+  UIImage *image = key.image;
+  CGBlendMode blendMode = contextIsClean == NO && CGImageGetAlphaInfo(image.CGImage) != kCGImageAlphaNone ? kCGBlendModeNormal : kCGBlendModeCopy;
```
appleguy: @Adlai-Holler This is a great patch, thank you!

This line is too complex to read.  If you leave it compound, please put parenthesis around the two elements before the ?.

Probably better is to create a BOOL imageHasAlpha or imageIsOpaque.

appleguy:
@Adlai-Holler The numbers here are impressive.  I have certainly seen the benefit of Copy first-hand, although I am interested in details of how you measured.  Does your measurement exclude the image decoding itself somehow?  Decoding is certainly more costly than the blend operation, so unless this were excluded (maybe by unintentional decoding at an earlier stage or something else), I'm not sure we could see such a large gain.

The other question is the scenario of scrolling to test this.  To remove extra variables you should force-disable rendering progressive images because this could add a lot of extra drawing time in a sporadic way.

AsyncDisplayKit/ASImageNode.mm
Adlai-Holler: I just wanted to avoid the alpha check if the context is clean. I'll clean it up.

Adlai-Holler:
It's a pretty unscientific test but probably close enough. I just fired up the app and scrolled down the home feed, opened a closeup, scrolled down to some related pins, opened another closeup, popped back, popped back, and scrolled down the home feed a little more.

The times listed there are the total times spent in `+createContentsForKey` compared to the total time spent in `ASTextNode drawRect:` as a reference. I'll send you the traces.

appleguy:
@Adlai-Holler yay, thank you!  We'll need to take some more accurate measurements — I'm pretty confident the gain is closer to ~10% than ~50% — but it can be guaranteed that there is a meaningful gain here and we should take this patch.

Looking into the image alpha is definitely important because a real alpha channel would increase memory consumption.  I believe the Apple frameworks default to doing this, so we probably need to set up the logic to explicitly specify opaque for JPEG images.

Once you have updated times, please edit your comment for future reference (feel free to include some screenshots of the ripc_RenderImage call or whatever else is applicable).

Adlai-Holler:
@appleguy Struck through the comment. I was wrong about PINRemoteImage – the images have alpha value 5 which is kCGImageAlphaNoneSkipLast which is the same as None so we're good. Filed the follow-up diff to get in on that.

PR #2206, opened 2016-09-07
[Layout] Automatic measurement on layout fixes

maicki:
- Fix measure pass of nodes that are range managed
- Fix measure pass for a node that is mid transition

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1382,17 +1382,19 @@ ASLayoutableSizeHelperForwarding
 
 - (void)measureNodeWithBoundsIfNecessary:(CGRect)bounds
 {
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
@maicki The case that we saw in Pinterest was actually range-managed node =/. A plain old ASDisplayNode node was inserted into the subtree of a collection cell, and a layout transition was manually initiated on that cell node before UIKit ran a layout pass that automeasured the node.

Would it make sense to instead go ahead and measure the node at layout transition start time if it hasn't been measured yet?

I'm going to play around with this diff in a little bit but I wanted to get you that info as soon as possible.

Adlai-Holler:
OK this is good to go. @maicki When you get time, rebase this please.

maicki:
@Adlai-Holler done

PR #2207, opened 2016-09-07
Fixes unsafe use of `asyncdisplaykit_node`.

johnepinterest:
The previous code used either a `@property` with `assign, nonatomic` semantics, or a `__unsafe_unretained` (non-atomic) ivar pointer to an instance for `asyncdisplaykit_node`.

This commit changes those access so that they have the equivalent of `weak, atomic` semantics.

- `_ASDisplayView.mm`
  - Removes the `_node` ivar that was qualified with `__unsafe_unretained`.
  - Removes `@synthesize asyncdisplaykit_node = _node;`.
  - All direct uses of `_node` were replaced with:
    - Creates a strong reference via `ASDisplayNode *node = _asyncdisplaykit_node;`.
    - `s/_node/node/`.
    - These changes were made even if there's a single use of `_asyncdisplaykit_node` as a consistency / defensive measure so that anyone editing this code in the future does not accidentally use a `weak` reference twice.

- `ASDisplayNode.mm`
  - Getters and setters for `asyncdisplaykit_node` were changed from `nonatomic, assign` semantics to the equivalent of `atomic, weak` semantics.
  - `weak` semantics were implemented by using a `ASWeakProxy` object.
  - `objc_setAssociatedObject` was changed from `OBJC_ASSOCIATION_ASSIGN` to `OBJC_ASSOCIATION_RETAIN` (the `atomic` version of retain).

- `ASDisplayNode+FrameworkPrivate.h`
  - Changed the `@property` declarations for `asyncdisplaykit_node` from `nonatomic, assign` to `atomic, weak`.
    - The actual getters and setters are implemented in `ASDisplayNode.mm`, which were changed accordingly.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -33,6 +33,7 @@
 #import "ASLayoutSpec.h"
 #import "ASLayoutValidation.h"
 #import "ASCellNode.h"
+#import "ASWeakProxy.h"
 
 NSInteger const ASDefaultDrawingPriority = ASDefaultTransactionPriority;
 NSString * const ASRenderingEngineDidDisplayScheduledNodesNotification = @"ASRenderingEngineDidDisplayScheduledNodes";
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
@johnepinterest Thanks! Couple notes:

- Will _ASDisplayView have a synthesized atomic accessor for the ivar, even though the getter is declared in a category of the superclass UIView? Should we use that atomic accessor instead of direct ivar access?
- If we only use the variable once during a given scope, what's the benefit in assigning it to a local variable? E.g. `[_asyncdisplaykit_node setTintColor:tintColor]`
- Would you mind rebasing onto master? Should be a trivial change in the `description` method.

johnepinterest:
- Will `_ASDisplayView` have a synthesized atomic accessor for the ivar, even though the getter is declared in a category of the superclass `UIView`?

Yes.  From the [documentation](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html#//apple_ref/doc/uid/TP40011210-CH5-SW5)

> By default, these accessor methods are synthesized automatically for you by the compiler, so you don’t need to do anything other than declare the property using `@property` in the class interface.

- Should we use that atomic accessor instead of direct ivar access?

Hmm.  While not obvious from a plain reading of the code, I believe that this is effectively achieved by using `ASDisplayNode *node = _asyncdisplaykit_node;`.  I don't see any advantage to using `self.asyncdisplaykit_node` vs. direct ivar access in this case as `_asyncdisplaykit_node` is a weak ivar and the weak to strong conversion implicility happens atomically.  One advantage of direct ivar access is that it's faster as it avoids a trip through the Objective C message dispatcher.

My recommendation is to leave it as is.

- If we only use the variable once during a given scope, what's the benefit in assigning it to a local variable? E.g.` [_asyncdisplaykit_node setTintColor:tintColor]`

None, really.  In my commit message I did mention this, though:

> These changes were made even if there's a single use of `_asyncdisplaykit_node` as a consistency / defensive measure so that anyone editing this code in the future does not accidentally use a `weak` reference twice.

- Would you mind rebasing onto master? Should be a trivial change in the `description` method.

Rebasing?  No.  But I did merge in the latest changes.

Adlai-Holler:
Oh so _that's_ what a "property" is! Thanks man!

PR #2209, opened 2016-09-08
Measure performance for ASCellNode layout

shamanskyh:
This is the first step in measuring layout performance in ASCellNode. A future PR will add similar tracking at the ASDataController level.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -23,9 +23,6 @@
 #import <AsyncDisplayKit/ASViewController.h>
 #import <AsyncDisplayKit/ASInsetLayoutSpec.h>
 
-NSString *const kASCellNodeLayoutSpecTime = @"ASCellNodeLayoutSpecTime";
-NSString *const kASCellNodeLayoutGenerationTime = @"ASCellNodeLayoutGenerationTime";
-
 #pragma mark -
 #pragma mark ASCellNode
 
@@ -47,8 +44,6 @@ NSString *const kASCellNodeLayoutGenerationTime = @"ASCellNodeLayoutGenerationTi
 {
   if (!(self = [super init]))
     return nil;
-
-  _measurementOptions = ASCellNodePerformanceMeasurementOptionsNone;
 
   // Use UITableViewCell defaults
   _selectionStyle = UITableViewCellSelectionStyleDefault;
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -23,6 +23,9 @@
 #import <AsyncDisplayKit/ASViewController.h>
 #import <AsyncDisplayKit/ASInsetLayoutSpec.h>
 
+NSString *const kASCellNodeLayoutSpecTime = @"ASCellNodeLayoutSpecTime";
+NSString *const kASCellNodeLayoutGenerationTime = @"ASCellNodeLayoutGenerationTime";
+
 #pragma mark -
 #pragma mark ASCellNode
 
@@ -44,6 +47,8 @@
 {
   if (!(self = [super init]))
     return nil;
+
+  _measurementOptions = ASCellNodePerformanceMeasurementOptionsNone;
 
   // Use UITableViewCell defaults
   _selectionStyle = UITableViewCellSelectionStyleDefault;
@@ -308,6 +313,7 @@
 
   return result;
 }
+
 @end
 
 
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -308,7 +308,6 @@
 
   return result;
 }
-
 @end
 
 
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -18,6 +18,18 @@ void ASPerformBlockOnMainThread(void (^block)());
 void ASPerformBlockOnBackgroundThread(void (^block)()); // DISPATCH_QUEUE_PRIORITY_DEFAULT
 ASDISPLAYNODE_EXTERN_C_END
 
+/**
+ * Bitmask to indicate what performance measurements the cell should record.
+ */
+typedef NS_OPTIONS(NSUInteger, ASDisplayNodePerformanceMeasurementOptions) {
+  ASDisplayNodePerformanceMeasurementOptionsNone = 0,
+  ASDisplayNodePerformanceMeasurementOptionsLayoutSpec = 1 << 0,
+  ASDisplayNodePerformanceMeasurementOptionsLayoutGeneration = 1 << 1
+};
+
+extern NSString *const ASDisplayNodeLayoutSpecTimes;
+extern NSString *const ASDisplayNodeLayoutGenerationTimes;
+
 @interface ASDisplayNode (Beta)
 
 /**
@@ -56,6 +68,17 @@ ASDISPLAYNODE_EXTERN_C_END
  * @abstract allow modification of a context after the node's content is drawn
  */
 @property (nonatomic, copy, nullable) ASDisplayNodeContextModifier didDisplayNodeContentWithRenderingContext;
+
+/**
+ * @abstract A bitmask representing which actions (layout spec, layout generation) should be measured.
+ */
+@property (nonatomic, assign) ASDisplayNodePerformanceMeasurementOptions measurementOptions;
+
+/**
+ * @abstract A dictionary representing performance metrics collected.
+ * @note see the constants above to retrieve relevant performance metrics
+ */
+@property (nonatomic, strong, readonly) NSDictionary *performanceMetrics;
 
 /** @name Layout Transitioning */
 
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -27,8 +27,13 @@ typedef NS_OPTIONS(NSUInteger, ASDisplayNodePerformanceMeasurementOptions) {
   ASDisplayNodePerformanceMeasurementOptionsLayoutGeneration = 1 << 1
 };
 
-extern NSString *const ASDisplayNodeLayoutSpecTimes;
-extern NSString *const ASDisplayNodeLayoutGenerationTimes;
+/**
+ * Keys to retrieve performance entries from the performance dictionary.
+ */
+extern NSString *const ASDisplayNodeLayoutSpecTotalTime;
+extern NSString *const ASDisplayNodeLayoutSpecNumberOfPasses;
+extern NSString *const ASDisplayNodeLayoutGenerationTotalTime;
+extern NSString *const ASDisplayNodeLayoutGenerationNumberOfPasses;
 
 @interface ASDisplayNode (Beta)
 
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -80,10 +80,10 @@ extern NSString *const ASDisplayNodeLayoutGenerationNumberOfPasses;
 @property (nonatomic, assign) ASDisplayNodePerformanceMeasurementOptions measurementOptions;
 
 /**
- * @abstract A dictionary representing performance metrics collected.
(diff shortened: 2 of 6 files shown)


facebook-github-bot-1:

facebook-github-bot-1:

Adlai-Holler:
This looks great! Feedback so far:

- Let's elevate this to ASDisplayNode, with the public API in ASDisplayNode+Beta.
- Let's use `__instanceLock__` around the measurement options and result values.
- Currently this will measure only the most recent layout pass. Would we rather measure the total time spent in all layout passes, plus possibly how many layout passes that is? 
- Let's get a rebase when you have time.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2279,8 +2279,16 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
 
   ASDN::MutexLocker l(__instanceLock__);
   if ((_methodOverrides & ASDisplayNodeMethodOverrideLayoutSpecThatFits) || _layoutSpecBlock != NULL) {
-    ASLayoutSpec *layoutSpec = [self layoutSpecThatFits:constrainedSize];
-    
+    ASLayoutSpec *layoutSpec = nil;
+    // optional performance measurement for cell nodes
+    if ([self isKindOfClass:[ASCellNode class]] &&
```
levi: You can leave this on for all nodes.

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -276,6 +281,12 @@ - (void)handleVisibilityChange:(BOOL)isVisible
                   withCellFrame:cellFrame];
 }
 
+- (NSDictionary *)performanceMetrics
+{
+  return @{kASCellNodeLayoutSpecTime : @(_timeToSpecLayout),
```
levi: Conditionally add the entries based on which measurement options are present

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2279,8 +2279,16 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
...
+        ((ASCellNode *)self).measurementOptions & ASCellNodePerformanceMeasurementOptionsLayoutSpec) {
+      ASDN::ScopeTimer t(_timeToSpecLayout);
+      layoutSpec = [self layoutSpecThatFits:constrainedSize];
```
levi: Better to put the scope timer in the body of these methods

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -276,6 +281,12 @@ - (void)handleVisibilityChange:(BOOL)isVisible
...
+           kASCellNodeLayoutGenerationTime : @(_timeToGenerateLayout)};
+}
```
levi: Use instance lock for these

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: Can't. Subclasses override them and aren't supposed to call super.

Adlai-Holler:
Awesome! Boxing those times and adding them to an array is somewhat expensive. @shamanskyh would it be sufficient to have the total time spent plus the number of data points? It would be really inexpensive to increment the count and sum the time.

shamanskyh:
Cool, wasn't sure, but not surprised. Summing and keeping a count should be sufficient.

AsyncDisplayKit/Private/_ASScopeTimer.h
```diff
@@ -38,3 +38,31 @@ namespace ASDN {
...
+
+ {
+ // some scope
+ ASDisplayNode::ScopeTimerSum t(placeToStoreTiming);
+ DoPotentiallySlowWork();
+ MorePotentiallySlowWork();
+ }
+
+ */
+namespace ASDN {
```
levi: YOu can add this to the same namespace above

AsyncDisplayKit/Private/_ASScopeTimer.h
```diff
@@ -38,3 +38,31 @@ namespace ASDN {
...
+  struct ScopeTimerSum {
```
levi: Call this SumScopeTimer

AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -57,6 +74,17 @@ ASDISPLAYNODE_EXTERN_C_END
...
+/**
+ * @abstract A bitmask representing which actions (layout spec, layout generation) should be measured.
+ */
+@property (nonatomic, assign) ASDisplayNodePerformanceMeasurementOptions measurementOptions;
+
+/**
+ * @abstract A dictionary representing performance metrics collected.
+ * @note see the constants above to retrieve relevant performance metrics
+ */
+@property (nonatomic, strong, readonly) NSDictionary *performanceMetrics;
```
levi: call this `performanceMeasurements`

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -308,6 +308,7 @@ - (void)handleVisibilityChange:(BOOL)isVisible
 
   return result;
 }
+
```
levi: Remove from this PR

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2342,8 +2374,16 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
 
   ASDN::MutexLocker l(__instanceLock__);
   if ((_methodOverrides & ASDisplayNodeMethodOverrideLayoutSpecThatFits) || _layoutSpecBlock != NULL) {
-    ASLayoutSpec *layoutSpec = [self layoutSpecThatFits:constrainedSize];
-    
+    ASLayoutSpec *layoutSpec = nil;
+    // optional performance measurement for cell nodes
+    if (self.measurementOptions & ASDisplayNodePerformanceMeasurementOptionsLayoutSpec) {
```
levi: Don't need to use the setter here since you've already grabbed the instance lock

AsyncDisplayKit/Private/ASDisplayNodeInternal.h
```diff
@@ -163,6 +165,13 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
   CGPoint _accessibilityActivationPoint;
   UIBezierPath *_accessibilityPath;
 
+  // performance measurement
+  ASDisplayNodePerformanceMeasurementOptions _measurementOptions;
+  NSTimeInterval _layoutSpecTotalTime;
+  int _layoutSpecNumberOfPasses;
```
levi: Use NSUInteger

AsyncDisplayKit/ASDisplayNode.mm
levi: I meant getter

AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -18,6 +18,23 @@ void ASPerformBlockOnMainThread(void (^block)());
 void ASPerformBlockOnBackgroundThread(void (^block)()); // DISPATCH_QUEUE_PRIORITY_DEFAULT
 ASDISPLAYNODE_EXTERN_C_END
 
+/**
+ * Bitmask to indicate what performance measurements the cell should record.
+ */
+typedef NS_OPTIONS(NSUInteger, ASDisplayNodePerformanceMeasurementOptions) {
+  ASDisplayNodePerformanceMeasurementOptionsNone = 0,
+  ASDisplayNodePerformanceMeasurementOptionsLayoutSpec = 1 << 0,
+  ASDisplayNodePerformanceMeasurementOptionsLayoutGeneration = 1 << 1
```
Adlai-Holler: Remove `ASDisplayNodePerformanceMeasurementOptionsNone` entry. `kNilOptions` (objc) or `[]` (Swift) is the standard for no options.

AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -18,6 +18,23 @@ void ASPerformBlockOnMainThread(void (^block)());
...
+};
+
+/**
+ * Keys to retrieve performance entries from the performance dictionary.
+ */
+extern NSString *const ASDisplayNodeLayoutSpecTotalTime;
+extern NSString *const ASDisplayNodeLayoutSpecNumberOfPasses;
+extern NSString *const ASDisplayNodeLayoutGenerationTotalTime;
+extern NSString *const ASDisplayNodeLayoutGenerationNumberOfPasses;
```
Adlai-Holler: Add the word `Key` to the end of each of these names.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -308,7 +313,7 @@ - (void)_initializeInstance
   _size = ASLayoutableSizeMake();
   _preferredFrameSize = CGSizeZero;
   _environmentState = ASEnvironmentStateMakeDefault();
-  
+
```
Adlai-Holler: Stray whitespace change

Adlai-Holler:
Alrighty, a few more tweaks and this is ready for primetime @shamanskyh

Adlai-Holler:
Thanks @shamanskyh we've been talking about doing something like this for a while and you did it in a day! Awesome!

AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -18,6 +18,22 @@ void ASPerformBlockOnMainThread(void (^block)());
 void ASPerformBlockOnBackgroundThread(void (^block)()); // DISPATCH_QUEUE_PRIORITY_DEFAULT
 ASDISPLAYNODE_EXTERN_C_END
 
+/**
+ * Bitmask to indicate what performance measurements the cell should record.
+ */
+typedef NS_OPTIONS(NSUInteger, ASDisplayNodePerformanceMeasurementOptions) {
+  ASDisplayNodePerformanceMeasurementOptionLayoutSpec = 1 << 0,
+  ASDisplayNodePerformanceMeasurementOptionLayoutGeneration = 1 << 1
```
appleguy: @shamanskyh these are so awesome to have!  I might suggest the second one be LayoutMeasurement or LayoutComputation.  Not important enough for a follow-up diff now though.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1155,6 +1160,33 @@ - (void)calculatedLayoutDidChange
...
+  ASDN::MutexLocker l(__instanceLock__);
+  return _measurementOptions;
+}
+
+- (NSDictionary *)performanceMeasurements
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  NSMutableDictionary *measurements = [NSMutableDictionary dictionaryWithCapacity:4];
+  if (_measurementOptions & ASDisplayNodePerformanceMeasurementOptionLayoutSpec) {
+    measurements[ASDisplayNodeLayoutSpecTotalTimeKey] = @(_layoutSpecTotalTime);
```
appleguy: @shamanskyh the design of this is efficient insofar as at recording time, cost is very low.  Accessing this property is pretty costly though, as the numbers also have to be boxed to put in the dictionary (and then accessed from it).

Would it be reasonable to define a struct type that all of these numbers can be contained within, and use that both as the instance variable storage and directly returned from this method?  If we want easy printing, maybe a good balance would be a basic NSObject subclass with simple @properties for each of these.

This has the downside (which you probably already considered) of making it more ambiguous which values are set vs zero.  It might be a nastier API, but we could use NaN to represent those defaults.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2352,7 +2392,16 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
     ASEnvironmentStatePropagateDown(layoutSpec, self.environmentTraitCollection);
```
appleguy: @shamanskyh it might be interesting to locally test this on a slow device, to see if it makes sense to include this in the OptionLayoutSpec measurement.  It makes more sense there than as the Generation step, and including it may be helpful as it would allow us to detect regressions in the environment system + overall account for more of the total work going on in these key methods.

PR #2210, opened 2016-09-08
Improve Crash Logging

hannahmbanana:
Many thanks to @johnepinterest for this one:

"This commit makes some changes to the ASKD "dealloc2main" stuff that we are considering to include in 6.10 to potentially help us track down some top crashes.

This changes alters the "dealloc2main" retain / release logic slightly by also including an "over release" check when -retain is called.
The idea being: Catching the error at that point may allow us to determine at what point the incorrect -retain is happening, instead of at -release, at which point the forensic stack information about the error is no longer present.

There is also a small, one line change to -retainCount. It was changed to use __atomic_load_n instead of reading the ivar directly. I overlooked this case the last time I ran the thread sanitizer and cleaned up the problems it reported. No one calls retainCount anyways, but making the read of the ivar an atomic load is more formally correct."


hannahmbanana:
@johnepinterest: is this wanted in 6.10 and 6.11?

levi:
@hannahmbanana thanks for putting this up. Getting this into 6.10 and 6.11 and master is ideal.

Adlai-Holler:
Hurrah for traceability!

PR #2212, opened 2016-09-08
[Layout] Don't use `automaticallyManageSubnodes` in ASTextCellNode

maicki:
Don't use `automaticallyManageSubnodes` in `ASTextCellNode` as all subclasses will inherit this behavior. In ASDKGram the `CommentsNode` is a subclass of `ASTextCellNode` and in there we didn't expect ASM of the superclass so subnodes where added and removed explicitly. Currently asm does not support automatic and manual adding and removing of subnodes yet. This caused crashes in ASDKGram.

A PR will follow that will add error handling (throw exception to the user with information) for this case.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -338,8 +338,8 @@ static const CGFloat kASTextCellNodeDefaultVerticalPadding = 11.0f;
   if (self) {
     _textInsets = textInsets;
     _textAttributes = [textAttributes copy];
-    self.automaticallyManagesSubnodes = YES;
     _textNode = [[ASTextNode alloc] init];
+    [self addSubnode:_textNode];
   }
   return self;
 }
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
The exception that got thrown here was so insanely hard to parse. Let's write a private version of remove/insert subnode. Have the public version assert that ASM is off and then call into the private version. 

I hate to throw exceptions at our users, but if the alternative is crashes with internal error messages, it's better.

A compromise would be modifying the common exception messages for this error. Maybe if the node has ASM on, add "This may be caused by manually adding or removing subnodes with automaticallyManagesSubnodes=YES."

Thoughts @maicki ? I think we should do option 1 but if it turns out that developers are doing this a whole whole lot we'll have to disable the exception. But the investment to make private subnode removal/insertions is likely to pay dividends as we continue developing this.

maicki:
@Adlai-Holler I'm already on implementing exceptions for adding / removing subnodes while asm is enabled on the supernode :) Stay tuned!

hannahmbanana:
Thanks for catching my mistake @maicki!

PR #2214, opened 2016-09-08
[Layout] Throw manually adding / removing subnodes if node / supernode has automaticallySubnodeManagement enabled

maicki:
Initial commit to throw an exception if manually adding / removing subnodes if node has automaticallyManagesSubnodes enabled

TODOS:
- [ ] Consolidate "insertSubnode:" methods
- [ ] More unit tests

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1554,6 +1554,12 @@ static bool disableNotificationsForMovingBetweenParents(ASDisplayNode *from, ASD
 
 - (void)addSubnode:(ASDisplayNode *)subnode
 {
+  ASDisplayNodeAssert(self.automaticallyManagesSubnodes == NO, @"Attempt to manually add subnode to node with automaticallyManagesSubnodes=YES. Node: %@", subnode);
+  [self _addSubnode:subnode];
+}
+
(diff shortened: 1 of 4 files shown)


levi:
Looks good! Small nit that's just personal preference — looks weird to have underscore methods exposed via the internal header in this case. I don't have any better suggestions, however.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1854,10 +1884,18 @@ - (void)_removeSubnode:(ASDisplayNode *)subnode
   [subnode __setSupernode:nil];
 }
 
-// NOTE: You must not called this method while holding the receiver's property lock. This may cause deadlocks.
 - (void)removeFromSupernode
 {
+  ASDisplayNodeAssert(self.supernode.automaticallyManagesSubnodes == NO, @"Attempt to manually remove subnode from node with automaticallyManagesSubnodes=YES. Node: %@", self);
+    
+  [self _removeFromSupernode];
```
Adlai-Holler: This is unfortunate. If you don't really care how the supernode is implemented (which is usually important), this will be very irritating.

Adlai-Holler:
Landing this. When we sync it, let's check that we don't cause a ton more headaches than it's worth.

Issue #2117, opened 2016-08-22
[ASCollectionView] Add/Fix Support for Interactive Movement


ay8s:
I've been exploring ways to enable reordering within ASCollectionView. I was trying out the Interactive Movement option introduced in iOS 9.0. It looks like most of the logic works except iOS isn't able to show the cell when a cell is actually picked up to move causing the rest of the logic and flow to fail.

Example Project: https://cl.ly/3n3C330N1Y10
Simply added the required methods to ASCollectionView Example.

When it works the cell should be "picked up" and move with the gesture location to the new location before collectionView:moveItemAtIndexPath:toIndexPath: is called with the final indexPaths.

The whole flow should look like... https://cl.ly/1R1v1c3c2f3D

Would be amazing if this could function with ASCollectionView and ASCellNodes.

Adlai-Holler:
@ay8s I'd love to get this up and running. We've got a ton of work for ASDK 2 that we're aiming to make ready for release, but here's the info I gathered on a deep-dive into getting collection view reordering working with ASDK (work is on the AHCollectionReordering branch):

- In `updateInteractiveMovementPosition` collection view calls `-moveItemAtIndexPath:toIndexPath:` on itself if the pending target index path has changed.
- It's a special kind of move, where the data source counts are required _not_ to change (since the data source shouldn't know about it). So we'll have to fake the `numberOfItemsInSection:` return value during interactive moves, to pretend we _don't_ know about the pending move.
- At the same time, we DO need the data controller to move the node after each move update or else the range controller will make mistakes in assigning interface states to each node.
- We need to update the range controller after each such move. This is easily accomplished by adding `-setNeedsUpdate` after calling `[super moveItem:]` if this is an interactive move.
- Just before `endInteractiveMove` we need to put the node back where it was and reset our fake data source counts, so that when we get the _real_ `collectionView:moveItemAtIndexPath:toIndexPath:` call, we do the update like normal.
- We'll probably need to add two ivars: `sourceInteractiveMoveIndexPath` and `pendingInteractiveMoveIndexPath`. We can use these to fake our data source item counts after each interactive move update, and to move the dragged node around to follow the cell.
- All of the above would work assuming they don't edit the collection view during interactive movement. I'm not sure how well UICollectionView handles this case, even.

harryworld:
[...]

PR #2221, opened 2016-09-09
[ASTableView] Add Support for Interactive Reordering

Adlai-Holler:


Adlai-Holler:
This is confirmed to work in Pinterest. Ready for review @levi @appleguy @maicki

ay8s:
@Adlai-Holler Awesome work on this, curious if this would be something fairly easy to add support for under UICollectionView interactive movement? #2117

Adlai-Holler:
@ay8s Great idea! I've updated the diff to add support for collection view reordering as well.

ay8s:
Nice @Adlai-Holler. Going to give this a go and see if it fixes the issues I was seeing. Curious if you know if it'd be possible to relayout the items when reordering is initiated to shrink the cells a little to make ordering quicker for the user as some of our cells are quite large so may explore hiding nodes or truncating the ASCellNode to a set height? Might be out of the scope of AsyncDisplayKit.

ay8s:
@Adlai-Holler Curious if you're seeing an assertion failure when moving within the ASCollectionView? 

`Assertion failure in -[BFRUpdateNode dealloc], /Users/andrewyates/Apps/Buffer/buffer-ios/App/Pods/AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:431`

Adlai-Holler:
@ay8s Thanks for testing this out! 

- Resizing all the cells should be possible by changing your cell nodes' layouts, and/or changing the return value for `constrainedSizeForNodeAtIndexPath:` if you implement it, and then calling `relayoutAllItems` on the collection view. If you're using flow layout, or if your custom layout knows about the change, all the items should be resized.
- Could you provide more detail about the assertion failure? Is it `Node should be marked invisible before deallocation?` Would you be willing to provide sample code to demo the issue?

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -898,6 +898,25 @@ - (NSArray *)completedNodes
...
+    [_externalCompletedNodes[newIndexPath.section] insertObject:node atIndex:newIndexPath.item];
+  }
+
+  NSMutableArray *internalCompletedNodes = _completedNodes[ASDataControllerRowNodeKind];
+  if (internalCompletedNodes != nil) {
+    NSMutableArray *oldSection = internalCompletedNodes[indexPath.section];
+    ASCellNode *node = oldSection[indexPath.item];
+    [oldSection removeObjectAtIndex:indexPath.item];
+    [internalCompletedNodes[newIndexPath.section] insertObject:node atIndex:newIndexPath.item];
+  }
```
levi: Can you simply this duplication with an inline function or method or something?

levi:
Looks good and really simple!

ay8s:
@Adlai-Holler Probably could have done a better job at an example here but if you move the items around a few times you should run into that assertion failure. I'm working on implementing it still using a custom flow layout to move the cells as you scroll.

Example: https://cl.ly/2w1Y1b3i3f1c

Adlai-Holler:
@ay8s OK great, your sample app helped me iron out some remaining issues with interactive movement. Thank you!

I never reproduced the assertion failure you mentioned. Now, with the latest code, the interactive movements and reloadDatas seem to be running very well.

ay8s:
@Adlai-Holler Glad I could help with something. Will give it a whirl tomorrow and see how it fairs in the Buffer iOS app. Will try and put together something if I come across any oddness.

Adlai-Holler:
@ay8s Hi! I continued building collection view reordering support, but unfortunately it's going to take more development and testing than I think belongs in this diff. So I reduced this diff back to just table view reordering and I'll land it.

I left the AHCollectionReordering branch up so we don't lose the knowledge gained. Here's an info dump about what will be required for collection view interactive reordering:
- Collection view doesn't call data source `collectionView:moveItemAtIndexPath:toIndexPath` until `-endInteractiveMovement`.
- In `updateInteractiveMovementPosition` collection view calls `-moveItemAtIndexPath:toIndexPath:` on itself if the pending target index path has changed.
- It's a special kind of move, where the data source counts are required _not_ to change (since the data source shouldn't know about it). So we'll have to fake the `numberOfItemsInSection:` return value during interactive moves, to pretend we _don't_ know about the pending move.
- At the same time, we DO need the data controller to move the node after each move update or else the range controller will make mistakes in assigning interface states to each node.
- We need to update the range controller after each such move. This is easily accomplished by adding `-setNeedsUpdate` after calling `[super moveItem:]` if this is an interactive move.
- Just before `endInteractiveMove` we need to put the node back where it was and reset our fake data source counts, so that when we get the _real_ `collectionView:moveItemAtIndexPath:toIndexPath:` call, we do the update like normal.
- We'll probably need to add two ivars: `sourceInteractiveMoveIndexPath` and `pendingInteractiveMoveIndexPath`. We can use these to fake our data source item counts after each interactive move update, and to move the dragged node around to follow the cell.

@ay8s Hopefully I or someone on the team can get collection view interactive reordering built soon. If you're interested in working on it, please ping me on the asyncdisplaykit slack channel!

ay8s:
@Adlai-Holler Sounds good. We actually ended up switching back to the table view as it seems to handle orientation changes and bounds changes a little smoother for our next release. I believe we'll still need the CollectionView reordering in the future however so I'll be sure to ping you if I get chance to take a look. Still early days with ASDK so keen to get to grips with more of it. 

Thanks for exploring it as far as you did! 👍

PR #2223, opened 2016-09-10
[ASDisplayNode] Let Nodes Deallocate Naturally, Deallocate UIKit Components in Run Loop

Adlai-Holler:
This removes ASDealloc2MainObject entirely.
- When a node is deallocated off-main, we read its ivars that are UIKit/CoreAnimation and schedule them for gradual deallocation in a main run loop queue.
- There's a shared NSCache that goes Class -> [UIKit-Ivar]. So we don't spend too much time scanning the objective-C runtime.
- UIImages and UIColors are deallocated normally, since we already deallocate these in the background.
- Anything else that inherits from UIView/CALayer, or is named UI\* CA\* is slated for main queue dealloc.
- ASViewController schedules its node for deallocation on the shared deallocation queue.


Adlai-Holler:
Testing with Pinterest is going great! Everything's working as expected. Ready for review @johnepinterest @appleguy @levi

Adlai-Holler:
Found an issue. CALayer's delegate pointer may dangle if its delegate isn't a UIView. So the property declaration that says `weak` is BS. I'll use a weak proxy to act as the delegate and stay alive until the layer is deallocated.

levi:
!!!

Adlai-Holler:
@levi Yep! To repro on master, just comment out 

```objectivec
  if (_flags.layerBacked)
    _layer.delegate = nil;
```

from `-[ASDisplayNode dealloc]` and run unit test `testThatLayerBackedSubnodesAreMarkedInvisibleBeforeDeallocWhenSupernodesViewIsRemovedFromHierarchyWhileBeingRetained`

It'll crash every time during UIView dealloc. The view is enumerating its layer's sublayers' delegates and it gets a garbage value. @johnepinterest

Adlai-Holler:
This thing is working really well! 15 minutes straight in Pinterest on simulator 9.3 with no issues. About 15% of nodes are deallocated off-main right now. We can probably get that up much higher later.

Adlai-Holler:
One next step I want to take is have that NSCache actually store the ivars for the key class PLUS all the superclasses up to ASDisplayNode. That way instead of ~3 cache accesses per off-main dealloc, we approach 1. NSCache is crazy fast though so it's not a huge deal.

Adlai-Holler:
#2229

Adlai-Holler:
When I was testing this a couple weeks ago, I ran into one issue. I can't remember the details, but it was in `[ASCollectionView dealloc]`, where it calls `setAsyncDataSource:nil`. That triggered a reloadData that was dangerous or something. It'll be easy to fix once we see it again, but heads up.

garrettmoon:
test this please

appleguy:
test this please

tungvoduc:
This fixes my problem with Swift 3.0

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -442,15 +443,119 @@ - (void)dealloc
...
+  dispatch_once(&onceToken, ^{
+    queue = [[ASRunLoopQueue alloc] initWithRunLoop:CFRunLoopGetMain() andHandler:^(id _Nonnull dequeuedItem, BOOL isQueueDrained) { }];
+    queue.batchSize = 10;
+  });
+
+  NSValue *ivarsObj = [[self class] _ivarsThatMayNeedMainDeallocation];
+
+  // Unwrap the ivar array
+  unsigned int count = 0;
+  NSAssert(sscanf(ivarsObj.objCType, "[%u^{objc_ivar}]", &count), @"Unexpected type in NSValue: %s", ivarsObj.objCType);
```
jparise: Won't this line be excluded when `NS_BLOCK_ASSERTIONS=1`, leaving `count = 0`?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -442,15 +443,119 @@ - (void)dealloc
...
+  static const int kMaxDealloc2MainIvarsPerClassTree = 64;
+  Ivar resultIvars[kMaxDealloc2MainIvarsPerClassTree];
+
+  // Get superclass results first.
+  Class c = class_getSuperclass(self);
+  if (c != [NSObject class]) {
+    NSValue *ivarsObj = [c _ivarsThatMayNeedMainDeallocation];
+    // Unwrap the ivar array and append it to our working array
+    unsigned int count = 0;
+    NSAssert(sscanf(ivarsObj.objCType, "[%u^{objc_ivar}]", &count), @"Unexpected type in NSValue: %s", ivarsObj.objCType);
```
jparise: Similar concern about `NS_BLOCK_ASSERTIONS=1` here.

AsyncDisplayKit/Private/ASInternalHelpers.m
```diff
@@ -84,6 +84,48 @@ void ASPerformBackgroundDeallocation(id object)
...
+  }
+
+  if ([class isSubclassOfClass:[UIResponder class]]
+      || [class isSubclassOfClass:[CALayer class]]
+      || [class isSubclassOfClass:[UIGestureRecognizer class]]) {
+    return YES;
+  }
+
+  const char *name = class_getName(class);
+  if (strncmp(name, "UI", 2) == 0 || strncmp(name, "AV", 2) == 0 || strncmp(name, "CA", 2) == 0) {
```
jparise: If this name-based check is faster than the three `isSubclassOfClass` tests above, consider performing this check first.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -442,15 +443,119 @@ - (void)dealloc
...
+  });
 
-  _pendingDisplayNodes = nil;
+  NSValue *result = [ivarsCache objectForKey:self];
+  if (result != nil) {
+    return result;
+  }
+
+  // Cache miss.
+  unsigned int resultCount = 0;
```
maicki: Some NSCache good to know in terms of memory and async: https://twitter.com/_mochs/status/796618264709361664 and https://twitter.com/_mochs/status/796678238844125184

AsyncDisplayKit/ASDisplayNode.mm
appleguy: @Adlai-Holler no wonder my profiling showed this was _really fast!_ (aka didn't show up :) )  In PROFILE, assertions are off.  Will retry with this fixed locally.

appleguy:
@Adlai-Holler As I suspected, you did a good job optimizing this approach from the design phase :).  Thank you for putting care into crafting this with a low-level implementation.

The overhead is definitely acceptable, about 0.1% CPU.  We of course get some bigger gains from having more objects deallocate off main (although more traffic going through the dealloc queue, and the extra retain / release operations, is going to create some complex additional lock contention issues for us to optimize later).

I think there's some more testing to do here, but performance won't be a blocking factor for this diff.

screen shot 2016-11-16 at 3 29 42 pm

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: @garrettmoon @maicki I don't think that guy is correct. Here's the twitter conversation we're in – the test that he claims reproduces the issue doesn't, and the cache keeps caching: https://twitter.com/adlaih/status/799162403652976640

Adlai-Holler: Also using NSMutableDictionary would be great except that Class doesn't do NSCopying, so we'd have to use an NSMapTable which is much slower.

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: D'oh! Good catch – patch applied.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -471,7 +471,8 @@ - (void)_scheduleIvarsForMainDeallocation
 
   // Unwrap the ivar array
   unsigned int count = 0;
-  NSAssert(sscanf(ivarsObj.objCType, "[%u^{objc_ivar}]", &count), @"Unexpected type in NSValue: %s", ivarsObj.objCType);
+  int scanResult = sscanf(ivarsObj.objCType, "[%u^{objc_ivar}]", &count);
+  NSAssert(scanResult, @"Unexpected type in NSValue: %s", ivarsObj.objCType);
```
jparise: You could be even more specific here and explicitly assert `scanResult == 1`, if you're feeling pedantic. :)

Adlai-Holler: I sure am! 👍

appleguy:
test this please

appleguy:
@Adlai-Holler I rebased this to master for you.  There was a strange merge issue — it is showing parts of the new layout code that is on master in this diff, even though they should not be seen as differences compared to the base branch.  Can you take a look?

Feel free to revert my rebase if it's incorrect.  The goal is that I want to live on this diff over the weekend because I'm extensively testing Pinterest with the layout coalescing changes anyway — but need to be able to apply both this diff and master's new inclusion of the coalescing.

AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -235,24 +236,19 @@ - (void)processQueue
...
 
     if (_internalQueue.empty()) {
       isQueueDrained = YES;
     }
   }
 
-  unsigned long numberOfItems = itemsToProcess.size();
-  for (int i = 0; i < numberOfItems; i++) {
-    if (isQueueDrained && i == numberOfItems - 1) {
-      _queueConsumer(itemsToProcess[i], YES);
```
Adlai-Holler: The prior logic seemed incorrect to me. If `isQueueDrained == YES` then this will behave the same as the `else` branch. If it's `NO` then it will also behave the same as the else branch.

AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -235,24 +236,19 @@ - (void)processQueue
...
     // Snatch the next batch of items.
-    NSUInteger totalNodeCount = _internalQueue.size();
-    for (int i = 0; i < MIN(self.batchSize, totalNodeCount); i++) {
-      id node = _internalQueue[0];
-      itemsToProcess.push_back(node);
-      _internalQueue.pop_front();
-    }
+    auto firstItemToProcess = _internalQueue.cbegin();
+    auto lastItemToProcess = MIN(_internalQueue.cend(), firstItemToProcess + self.batchSize);
+    itemsToProcess = std::vector<id>(firstItemToProcess, lastItemToProcess);
```
Adlai-Holler: I've confirmed it's safe to construct the iterator `firstItemToProcess + self.batchSize` even if that goes beyond the end of the queue. Tested by setting batch size to 10000 and then confirming that it safely grabbed the entire queue.

PR #2224, opened 2016-09-10
[Layout] Use `_addSubnode` / `_removeFromSupernode` methods in _ASDisplayView

maicki:
Fix assertion in `_ASDisplayView` if adding a subnode or removing from super node. It needs to be looked into more deeply for nodes with `automaticallyManagesSubnodes=YES` though

### DIFF
--- a/AsyncDisplayKit/Details/_ASDisplayView.mm
+++ b/AsyncDisplayKit/Details/_ASDisplayView.mm
@@ -127,7 +127,8 @@
 
     if (needsSupernodeUpdate) {
       // -removeFromSupernode is called by -addSubnode:, if it is needed.
-      [newSuperview.asyncdisplaykit_node addSubnode:node];
+      // FIXME: Needs rethinking if automaticallySubnodeManagements=YES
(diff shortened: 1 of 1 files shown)


appleguy:
@maicki Could you explain a bit more what the functional difference is / why this is the correct behavior?

Small typo should be fixed so that a find-all locates this instance - I think it is automaticallyManagesSubnodes (in the comments you added)

appleguy:
r=me though, you can land this - just hoping for some additional documentation about why this is the right change.

Adlai-Holler:
@appleguy Once we added exceptions in #2214 against manually manipulating subnodes of nodes using automatic subnode management, these code paths were causing exceptions that surprised users.

This change is a stopgap until we can do a deep-dive into how to combine manual and automatic subnode management more cleanly. However, now that we have internal insert/remove subnode methods, it also makes conceptual sense for us to call them in cases like this.

Adlai-Holler:
Since Michael is out, I'm going to reopen this with the typo fixed and land it.

PR #2235, opened 2016-09-13
[ASDataController] Allocate & Measure Cell Nodes Async on Shared Queue

Adlai-Holler:
In the future we want to support creating ASCellNodes on demand, after inserting the item/row into the UITableView/UICollectionView.

To support this, we need to make our cell node allocation & layout asynchronous. It is critical that this process be wait-onable, and that we can limit the total number of concurrent layouts. It's nice if this process supports changing priorities for tasks on-the-fly.

As a first step, I've built this on NSOperation. NSOperation has an unfortunate amount of overhead, but it supports all the features mentioned above and I don't think the overhead will be very significant in practice. We will likely replace NSOperation for this usage in the future, and we'll keep that in mind as we build.

Obviously this is a small step but I'm confident it moves us in the right direction.

- For laying-out nodes, replace the synchronous ASDispatchApply call with enqueuing onto a shared "node layout" operation queue.
- Have ASIndexedNodeContext own the creation operation and the finished node.
  - Likely that ASIndexedNodeContext evolves into a sort of `ASCollectionItem` object, that survives beyond the initial insertion of the item.

Ready for review @appleguy @levi 

P.S. Sorry for the mess in ASIndexedNodeContext.h but the init method came after the properties so, y'know. Scorched earth policy. 😛

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -127,18 +127,6 @@ NSString * const ASDataControllerRowNodeKind = @"_ASDataControllerRowNodeKind";
   return parallelProcessorCount;
 }
 
-+ (NSOperationQueue *)nodeLayoutQueue
-{
-  static NSOperationQueue *nodeLayoutQueue;
-  static dispatch_once_t onceToken;
-  dispatch_once(&onceToken, ^{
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
This needs a minor change. I'll reopen after.

Issue #2229, opened 2016-09-12
Crash in -[ASDealloc2MainObject retain]


GitGadd:
After updating to master, I'm getting a crash in -[ASDealloc2MainObject Retain] every time I reload a certain screen. The object being deallocated is a subclass of ASDisplayNode written in Swift with nothing special going on. I'm not doing any manual retains/releases.

The exception is 'EXC_BREAKPOINT' and the line being hit is

```
_OBJC_SUPPORTED_INLINE_REFCNT_WITH_DEALLOC2MAIN
```

Attaching a screenshot of the trace, though it is unfortunately very short and uninformative.

Device is running the latest beta of iOS 10 as of now, which is 10.0.1 (14A403).
Using Xcode Version 8.0 (8A218a) (=the GM)

I would love to help in any way I can, but I have way too little knowledge of how ASDK manages references to even begin investigating. Any help is greatly appreciated as I cannot revert to using the stable 1.9.x version because of another bug that was fixed in the 2.x betas.

![screen shot 2016-09-12 at 16 57 16](https://cloud.githubusercontent.com/assets/3338062/18440775/adc35014-790a-11e6-9da0-a6eb3a3d6a0a.png)

hannahmbanana:
@GitGadd: thanks including all of the relevant details here! @Adlai-Holler is currently working to isolate and resolve the underlying cause of the issue. We don't yet have a 100% easy repro case, so it would be awesome if we could see some of your code. I'll send you a message on Slack. 

@Adlai-Holler @appleguy

khashish:
FWIW, this issue happens on the ASDKgram example when scrolling down then up quickly

PR #2208, opened 2016-09-08
[ASDisplayNode] Majorly Improve Descriptions

Adlai-Holler:
I went to debug some issues today and got annoyed with how off-the-mark our object descriptions are.

### `po` ASTextNode
**Before:** `<ASTextNode 0x7ff7dc841a00 [_ASDisplayView : 0x7ff7dad37620]>`

Very little info, nonstandard format.

**After:** `<ASTextNode: 0x7fcdac01ec00; text = "Lick tail flick jump on the table stretching purr …"; frameInWindow = (106 350.5; 198 210); frame = (106 16; 198 210); cellNode = <KittenNode: 0x7fcdaae3f8c0>; interfaceState = { MeasureLayout | Preload | Display | Visible }; view = <_ASDisplayView: 0x7fcdac83fe80>>`

Way more information, including the frame onscreen, text and the nearest ASCellNode ancestor.

--- 

### `po` ASCellNode
**Before:** `<KittenNode 0x7ff7dac12350 name = Node for kitten #144 [_ASDisplayView : 0x7ff7dad1c0a0]>`

Very little info, nonstandard format.

**After:** `<KittenNode: 0x7fcdaae3f8c0; name = "Node for kitten #144"; frameInWindow = (0 334.5; 320 242); frame = (0 0; 320 242); interfaceState = { MeasureLayout | Preload | Display | Visible }; view = <_ASDisplayView: 0x7fcdac83f7e0>; indexPath = (0, 3); tableView = <ASTableView: 0x7fcdad01b000>>`

Way more information, including index path and owning ASTableView/ASCollectionView/ASTableNode/ASCollectionNode.

---

### `po` _ASDisplayView

**Before:** `<<ASTextNode: 0x7ff7dc841a00; text = "Lick quis nunc toss the mousie vel, tortor pellent…"; truncation string = "…"; frame = {{106, 16}, {198, 384.5}}; renderer = 0x7ff7dc399520>, view = <_ASDisplayView: 0x7ff7dad37620; frame = (106 16; 198 384.5); userInteractionEnabled = NO; layer = <<ASTextNode: 0x7ff7dc841a00; text = "Lick quis nunc toss the mousie vel, tortor pellent…"; truncation string = "…"; frame = {{106, 16}, {198, 384.5}}; renderer = 0x7ff7dc399520>, layer = <_ASDisplayLayer: 0x7ff7dad172d0>>>>`

[...]

PR #2241, opened 2016-09-13
Make Description Functions Dealloc-Safe

Adlai-Holler:
I believe that #2229 is caused by #2011 failing the assertion, and attempting to describe the node during dealloc.

The description method has not been dealloc-safe since #2208. This patch makes the description functions dealloc-safe.


Adlai-Holler:
Scott looked at this and said he's cool landing it. Neither of us is 100% clear on the `_autoreleasing` specifier but ARC doesn't complain if these are called in dealloc anymore. We'll keep an eye on it.

PR #2249, opened 2016-09-15
[ASDisplayNode] Use Weak Proxy to Avoid Dangling CALayer.delegate

Adlai-Holler:
CALayer's delegate property is declared `weak` but implemented `assign`.

If the node is marked for deallocation on a background thread, and the layer tries to access its delegate before the node runs `dealloc` (deferred by ASDealloc2MainObject) then you get a crash for trying to retain an object after it's marked for deallocation.

This diff uses a proxy object between the layer and the node. The layer retains the proxy.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -444,8 +444,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
   _view = nil;
   _subnodes = nil;
-  if (_flags.layerBacked)
-    _layer.delegate = nil;
   _layer = nil;
 
   // TODO: Remove this? If supernode isn't already nil, this method isn't dealloc-safe anyway.
@@ -461,7 +459,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (void)__unloadNode
 {
-  ASDisplayNodeAssertThreadAffinity(self);
+  ASDisplayNodeAssertMainThread();
   ASDisplayNodeAssert([self isNodeLoaded], @"Implementation shouldn't call __unloadNode if not loaded: %@", self);
   ASDisplayNodeAssert(_flags.synchronous == NO, @"Node created using -initWithViewBlock:/-initWithLayerBlock: cannot be unloaded. Node: %@", self);
   ASDN::MutexLocker l(__instanceLock__);
@@ -553,8 +551,13 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   if (isLayerBacked) {
     TIME_SCOPED(_debugTimeToCreateView);
     _layer = [self _layerToLoad];
-    // Surpress warning for Base SDK > 10.0
-    _layer.delegate = (id<CALayerDelegate>)self;
+    static int ASLayerDelegateAssociationKey;
+
+    // We cannot make self the layer's delegate directly, because the delegate is
+    // actually `assign` though the docs say weak.
+    ASWeakProxy *instance = [ASWeakProxy weakProxyWithTarget:self];
+    _layer.delegate = (id<CALayerDelegate>)instance;
+    objc_setAssociatedObject(_layer, &ASLayerDelegateAssociationKey, instance, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
   } else {
     TIME_SCOPED(_debugTimeToCreateView);
     _view = [self _viewToLoad];
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -553,8 +553,12 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     _layer = [self _layerToLoad];
     static int ASLayerDelegateAssociationKey;
 
-    // We cannot make self the layer's delegate directly, because the delegate is
-    // actually `assign` though the docs say weak.
+    /**
+     * CALayer's .delegate property is documented to be weak, but the implementation is actually assign.
+     * Because our layer may survive longer than the node (e.g. if someone else retains it, or if the node
+     * begins deallocation on a background thread and it waiting for the -dealloc call to reach main), the only
+     * way to avoid a dangling pointer is to use a weak proxy.
+     */
     ASWeakProxy *instance = [ASWeakProxy weakProxyWithTarget:self];
     _layer.delegate = (id<CALayerDelegate>)instance;
     objc_setAssociatedObject(_layer, &ASLayerDelegateAssociationKey, instance, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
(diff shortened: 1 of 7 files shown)


Adlai-Holler:
OK this diff now includes a failing (now passing) test case. I've tested it with Pinterest and it's working well. Ready for review @levi @johnepinterest @appleguy @maicki

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -461,7 +459,7 @@ - (void)dealloc
 
 - (void)__unloadNode
 {
-  ASDisplayNodeAssertThreadAffinity(self);
```
levi: Why is this main-only now?

Adlai-Holler: It always has been, since we deallocate the view here, and since we assert thread affinity and also assert node is loaded. This just makes it clearer and I hate the phrase "ThreadAffinity."

Adlai-Holler:
Hmmm `Failed to install slather. Failed to build native gem extension.` Retrying build…

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -553,8 +551,13 @@ - (void)_loadViewOrLayerIsLayerBacked:(BOOL)isLayerBacked
   if (isLayerBacked) {
     TIME_SCOPED(_debugTimeToCreateView);
     _layer = [self _layerToLoad];
-    // Surpress warning for Base SDK > 10.0
-    _layer.delegate = (id<CALayerDelegate>)self;
+    static int ASLayerDelegateAssociationKey;
+
+    // We cannot make self the layer's delegate directly, because the delegate is
```
appleguy: Maybe say "CALayer's .delegate property is documented to be weak, but the code emitted is actually assign."
"Because ASDisplayNodes may begin deallocation on a background thread but only execute -dealloc on the main thread, the only way to avoid a dangling pointer is to use a weak proxy as the CALayer .delegate"

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -553,8 +551,13 @@ - (void)_loadViewOrLayerIsLayerBacked:(BOOL)isLayerBacked
...
+    // actually `assign` though the docs say weak.
+    ASWeakProxy *instance = [ASWeakProxy weakProxyWithTarget:self];
+    _layer.delegate = (id<CALayerDelegate>)instance;
+    objc_setAssociatedObject(_layer, &ASLayerDelegateAssociationKey, instance, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
```
appleguy: Hmm, I didn't realize we would need this, although it makes sense.  We should take this diff anyway, but I'm pretty confident this will show up in profiling.  I can think of two costs:

- Object allocation / deallocation of ASWeakProxy (not just total time, but objc runtime locks / malloc contention when the system is running at load)
- Associated object definition and also teardown when layer is deallocated
- All traffic to the delegate needs to access the associated object too, which also includes retain/release traffic
- Finally, proxy-related overhead for checking selectors...

Probably the key observation is that almost all of this work will be confined to the main thread (although still cause micro-contentions with background threads).  layoutSublayers, display, etc.

It seems to me like setting _layer.delegate = _layer is more utilitarian.  I believe we already have a nodeDelegate or the layer just accesses the node directly anyway, so the code should be short.

All that said, we may not use the delegate as much as it could be used.  I think we override -layoutSublayers and -display directly instead of using the delegate method, for example.  So maybe the traffic is very low.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransactionContainer.m
```diff
@@ -85,12 +81,10 @@ - (void)asyncdisplaykit_cancelAsyncTransactions
 
 - (void)asyncdisplaykit_asyncTransactionContainerStateDidChange
 {
-#if ASASYNCTRANSACTIONCONTAINER_FORWARD_STATE_CHANGE
   id delegate = self.delegate;
   if ([delegate respondsToSelector:@selector(asyncdisplaykit_asyncTransactionContainerStateDidChange)]) {
```
appleguy: @Adlai-Holler I had a few moments to look at this, and as I'm sure you were seeing originally, the integration point isn't that deep  (I had incorrectly assumed it was).

Let's go ahead and remove this functionality entirely (either in this diff or another).  We need to make other improvements to the transaction system and I suspect it may return, but it might as well be part of a larger audit.

AsyncDisplayKitTests/ASDisplayNodeTests.m
```diff
@@ -1964,6 +1964,23 @@ - (void)testThatLayerBackedSubnodesAreMarkedInvisibleBeforeDeallocWhenSupernodes
...
+- (void)testThatLayerDelegateDoesntDangleAndCauseCrash
+{
+  NS_VALID_UNTIL_END_OF_SCOPE UIView *host = [[UIView alloc] init];
+
+  __block NS_VALID_UNTIL_END_OF_SCOPE ASDisplayNode *node = [[ASDisplayNode alloc] init];
+  node.layerBacked = YES;
+
+  [host addSubnode:node];
+  [self executeOffThread:^{
+    node = nil;
```
appleguy: Could you comment this code more?  I think the idea is:

// Running on main thread
// Cause retain count of node to fall to zero _synchronously on a background thread_ (pausing main thread)
// ASDealloc2MainObject queues actual call to -dealloc to occur on the main thread
// Continue execution on main, before the dealloc can run, to dealloc the host view
// Node is in an invalid state (about to dealloc, not valid to retain) but accesses to sublayer delegates causes attempted retain — unless weak variable works correctly

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: Yep, it's not free. Still, we need to land this and move on. The performance cost here is almost certainly not worth straining our already-strained engineering resources. Especially when there is incredibly low-hanging performance fruit waiting for us up ahead.

One of the important uses for the layer delegate is observing kCAOrderIn/kCAOrderOut. We cannot have the layer be its own delegate unless we drop support for that or drop support for layer-backed synchronous nodes. Which I would be OK with, although AVPlayerLayer nodes are nice to have.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransactionContainer.m
```diff
@@ -102,19 +88,18 @@ - (_ASAsyncTransaction *)asyncdisplaykit_asyncTransaction
       transactions = [NSHashTable hashTableWithOptions:NSPointerFunctionsObjectPointerPersonality];
       self.asyncdisplaykit_asyncLayerTransactions = transactions;
     }
+    __weak CALayer *weakSelf = self;
     transaction = [[_ASAsyncTransaction alloc] initWithCallbackQueue:dispatch_get_main_queue() completionBlock:^(_ASAsyncTransaction *completedTransaction, BOOL cancelled) {
+      __strong CALayer *self = weakSelf;
+      if (self == nil) {
+        return;
+      }
```
Adlai-Holler: Thoughts on this __weak here @appleguy ? Zeroing weak refs aren't free but I don't like the idea of all of our layers establishing retain cycles during their display. And given the issue we saw today, it definitely causes layers to be retained longer than their nodes in some cases.

Adlai-Holler:
@appleguy Thanks for the review! I'll modify this to use the layer itself in a follow-up diff. ship://Problems/239 <Have ASDisplayLayer Be Its Own Delegate, No Proxy>

PR #2251, opened 2016-09-15
Remove Unneeded Thread Affinity Assertions

Adlai-Holler:
While debugging today, I was trying to `po someLoadedNode` (calls -debugDescription) from a background thread and ran into the `inHierarchy` thread affinity assertion. We don't need it, so let's get rid of it.


maicki:
LGTM

PR #2255, opened 2016-09-15
[Layout] Further deprecation steps for `preferredFrameSize`

maicki:
- Remove all support for preferredFrameSize in ASDK
- preferredFrameSize setter calls through and sets the width and height of the node
- preferredFrameSize getter tries to return a CGSize based on the width and height properties otherwise if this is not possible it throws

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -827,7 +827,7 @@ NS_ASSUME_NONNULL_BEGIN
  * size. For example, this property could be set on an ASImageNode to display at a size different from the underlying
  * image size.
  *
- * @return Try to create a CGSize for preferredFrameSize of this node from the width and height property of this node. It will throw if the dimension unit of width or height is of type ASDimensionUnitFraction.
+ * @return Try to create a CGSize for preferredFrameSize of this node from the width and height property of this node. It will return CGSizeZero if widht and height dimensions are not of type ASDimensionUnitPoints.
  *
  * @deprecated Deprecated in version 2.0: Just calls through to set the height and width property of the node. Convert to use sizing properties instead: height, minHeight, maxHeight, width, minWidth, maxWidth.
  */
--- a/AsyncDisplayKit/ASDisplayNode.h
+++ b/AsyncDisplayKit/ASDisplayNode.h
@@ -827,9 +827,9 @@ NS_ASSUME_NONNULL_BEGIN
  * size. For example, this property could be set on an ASImageNode to display at a size different from the underlying
  * image size.
  *
- * @return The preferred frame size of this node
+ * @return Try to create a CGSize for preferredFrameSize of this node from the width and height property of this node. It will throw if the dimension unit of width or height is of type ASDimensionUnitFraction.
  *
- * @deprecated Deprecated in version 2.0: Use sizing properties instead: height, minHeight, maxHeight, width, minWidth, maxWidth
+ * @deprecated Deprecated in version 2.0: Just calls through to set the height and width property of the node. Convert to use sizing properties instead: height, minHeight, maxHeight, width, minWidth, maxWidth.
  */
 @property (nonatomic, assign, readwrite) CGSize preferredFrameSize ASDISPLAYNODE_DEPRECATED;
 
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -3488,11 +3488,11 @@ ASEnvironmentLayoutExtensibilityForwarding
 {
   ASDN::MutexLocker l(__instanceLock__);
   
-  // We don't support fractions in preferredFrameSize
-  ASDisplayNodeAssert(_style.width.unit != ASDimensionUnitFraction, @"preferredFrameSize.width cannot be of unit type ASDimensionUnitFraction");
-  ASDisplayNodeAssert(_style.height.unit != ASDimensionUnitFraction, @"preferredFrameSize.height cannot be of unit type ASDimensionUnitFraction");
+  if (_style.width.unit == ASDimensionUnitPoints && _style.height.unit == ASDimensionUnitPoints) {
+    return CGSizeMake(_style.width.value, _style.height.value);
+  }
 
-  return CGSizeMake(_style.width.value, _style.height.value);
+  return CGSizeZero;
 }
 
 @end
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2465,8 +2465,7 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   __ASDisplayNodeCheckForLayoutMethodOverrides;
-    
-  ASDN::MutexLocker l(__instanceLock__);
+
   return CGSizeZero;
 }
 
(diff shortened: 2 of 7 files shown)


AsyncDisplayKitTests/ASDisplayNodeTests.m
```diff
@@ -2048,4 +2048,25 @@ - (void)testThatOnDidLoadWorks
...
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+
+  ASDisplayNode *node = [ASDisplayNode new];
+  
+  // Default auto preferred frame size will be CGSizeZero
+  XCTAssert(CGSizeEqualToSize(node.preferredFrameSize, CGSizeZero));
+  
+  // Set a specific preferredFrameSize
+  node.preferredFrameSize = CGSizeMake(100, 100);
+  XCTAssert(CGSizeEqualToSize(node.preferredFrameSize, CGSizeMake(100, 100)));
```
Adlai-Holler: `ASXCTAssertEqualSizes`

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3468,6 +3440,27 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
 }
 
+- (void)setPreferredFrameSize:(CGSize)preferredFrameSize
+{
+  ASDN::MutexLocker l(__instanceLock__);
+
+  // Deprecated preferredFrameSize just calls through to set width and height
+  self.width = ASDimensionMake(preferredFrameSize.width);
+  self.height = ASDimensionMake(preferredFrameSize.height);
+  [self invalidateCalculatedLayout];
```
Adlai-Holler: Can we remove this invalidation? Shouldn't `width` and `height` perform an equality check and invalidate if needed?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3468,6 +3440,27 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
+
+- (CGSize)preferredFrameSize
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  
+  // We don't support fractions in preferredFrameSize
+  ASDisplayNodeAssert(self.width.unit != ASDimensionUnitFraction, @"preferredFrameSize.width cannot be of unit type ASDimensionUnitFraction");
+  ASDisplayNodeAssert(self.height.unit != ASDimensionUnitFraction, @"preferredFrameSize.height cannot be of unit type ASDimensionUnitFraction");
+
+  return CGSizeMake(self.width.value, self.height.value);
```
Adlai-Holler: Let's be lenient here and return CGSizeZero, our old default if needed. People who are reading this deprecated property know that they need to change their code, the last thing they want is an exception in a getter.

Adlai-Holler:
I'll be unavailable soon – plane with no wi-fi. I'm cool with landing this, modulo those comments.

AsyncDisplayKit/ASDisplayNode.mm
maicki: When do you think we should return `CGSizeZero`? We could return zero if width and height are of type `auto`, but that's already happening. What if just one of `width` / `height` properties have a unit of `ASDimensionUnitFraction` and developer expect some size if at least one is of type `ASDimensionUnitPoint`?

AsyncDisplayKit/ASDisplayNode.mm
maicki: With the changes we will do in #2261 this has to change in any case. For now I think it's fine to do it this way to have the same behavior as before.

AsyncDisplayKitTests/ASDisplayNodeTests.m
maicki: Done

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: I think that developers will never expect a partial size like `(640, 0.75)` from `preferredFrameSize`. I think we should return `CGSizeZero` unless `width` and `height` are `ASDimensionUnitPoints` so that we always give them back a size that is measured in points.

Adlai-Holler: (It's wacky that this comment got disconnected from the one I was responding to, in the PR overview page.)

AsyncDisplayKit/ASDisplayNode.mm
maicki: Will put up a PR soon that will enable callbacks for `style` object changes and we can add invalidation in case `width` and `height` changes.

AsyncDisplayKit/ASDisplayNode.mm
maicki: @Adlai-Holler Agree with thinking and addressed

maicki:
@garrettmoon @Adlai-Holler @appleguy Ready for review

AsyncDisplayKit/ASDisplayNode.h
```diff
@@ -827,9 +827,9 @@ NS_ASSUME_NONNULL_BEGIN
  * size. For example, this property could be set on an ASImageNode to display at a size different from the underlying
  * image size.
  *
- * @return The preferred frame size of this node
+ * @return Try to create a CGSize for preferredFrameSize of this node from the width and height property of this node. It will throw if the dimension unit of width or height is of type ASDimensionUnitFraction.
```
garrettmoon: Can you say: "It will throw an error" instead of just "It will throw"

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2469,12 +2467,6 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
   __ASDisplayNodeCheckForLayoutMethodOverrides;
```
garrettmoon: This method is always going to return CGSizeZero now, should it throw an assert as well? Do we need the lock?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3501,6 +3474,27 @@ - (ASLayout *)measureWithSizeRange:(ASSizeRange)constrainedSize
...
+
+- (CGSize)preferredFrameSize
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  
+  if (_style.width.unit == ASDimensionUnitPoints && _style.height.unit == ASDimensionUnitPoints) {
+    return CGSizeMake(_style.width.value, _style.height.value);
+  }
+
+  return CGSizeZero;
```
garrettmoon: Should this assert? Something along the lines of "Could not figure out the preferredFrameSize, it's recommended that you switch to accessing height and width directly."

Adlai-Holler: It asserted previously, and I recommended we just return CGSizeZero. My rationale was that we should be lenient – users who are reading a deprecated property know that they need to update their code, and the last thing they want is an exception being thrown from an accessor.

AsyncDisplayKit/ASDisplayNode.mm
maicki: We don't need the lock, but I think we should not throw.

AsyncDisplayKit/ASDisplayNode.h
maicki: Changed

AsyncDisplayKit/ASDisplayNode.mm
maicki: Agree with @Adlai-Holler

maicki:
@garrettmoon @Adlai-Holler ready for review again

AsyncDisplayKit/ASDisplayNode.mm
garrettmoon: Counterpoint is it's much easier to address an assert than to figure out why something isn't working correctly, possibly in a subtle way. But I also understand y'alls point.

Adlai-Holler:
Nice here we go!

PR #2178, opened 2016-09-01
[ASCollectionView][Architecture] Introduce ASSectionContext

nguyenhuy:
It is a custom object that can be optionally provided by ASCollectionDataSource and later retrieved from the collection view to be used in collection view layout. The idea was first proposed by @Adlai-Holler in #2067. 

I've been playing around with different ways to implement this feature, including having an ASSection that is a proxy of cell nodes array ([here](https://github.com/facebook/AsyncDisplayKit/compare/master...nguyenhuy:ASSectionContext_proxy?expand=1)), or an ASSection that is a subclass of NSMutableArray<ASCellNode *> ([here](https://github.com/facebook/AsyncDisplayKit/compare/master...nguyenhuy:ASSectionContext_array_subclass?expand=1)). I think the implementation in this PR is simpler and thus, more elegant.

Please let me know what you think, @Adlai-Holler @maicki @appleguy @garrettmoon @levi. Thanks!

TODO:
- [ ] Documentation ([Overview.md](https://github.com/facebook/AsyncDisplayKit/blob/master/plans/section-infos-api/Overview.md))
- [x] Unit test

PR #2257, opened 2016-09-15
[ASDataController] No Longer Wait On Prior Updates When Updated

Adlai-Holler:
I don't know why we used to do this, but there's no reason to flush prior updates when we are updated. 

Also it was planned to prevent users from calling `waitUntilAllUpdatesAreCommitted` while submitting batch updates, but that's not necessary either. There are valid use cases for being in the middle of a batch update, and waiting for all prior updates to be finished.

I've tested this on simulator with Pinterest for a couple minutes and it's working well.

### DIFF
--- a/AsyncDisplayKit/Details/ASChangeSetDataController.mm
+++ b/AsyncDisplayKit/Details/ASChangeSetDataController.mm
@@ -84,18 +84,6 @@
   return batchUpdating;
 }
 
-- (void)waitUntilAllUpdatesAreCommitted
-{
-  ASDisplayNodeAssertMainThread();
-  if (self.batchUpdating) {
-    // This assertion will be enabled soon.
-//    ASDisplayNodeFailAssert(@"Should not call %@ during batch update", NSStringFromSelector(_cmd));
-    return;
-  }
-
-  [super waitUntilAllUpdatesAreCommitted];
-}
-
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
I've put up branch `AHRemoveImplicitWait_p6.12` which is based on our p6.12 branch, and we can test Pinterest using that. We'll make sure to test more thoroughly before this is synced.

Ready for review! @chrisdanford @hannahmbanana @maicki @levi @appleguy

appleguy:
@Adlai-Holler A long time ago, I believe this was necessary - but it was also before the ASIndexedNodeContext and other things, so it's good to hear that it isn't now.

You already committed to the testing, but I can't understate how important that is for this PR.  It looks good to me from a code perspective, but before landing it I think we should complete a few lodestone passes with it.  Maybe you can make a Pinterest branch so that everyone on our team / others who opt-in can run on it, and we wouldn't update that build at all, so any Fabric crashes can be aggregated at low volume after a weekend of testing?

Adlai-Holler:
One consideration with this diff is, it would benefit greatly from the section context system. This waiting is not technically necessary, but it can seriously reduce the incidence of section-metrics-out-of-sync issues that the new system is designed to fix.

Adlai-Holler:
Closing until #2178 is in the bag.

PR #2259, opened 2016-09-16
[ASViewController] Fix Accessing traitCollection on iOS 7

Adlai-Holler:
Fixes an iOS 7 regression that would reliably cause crashes, from #2202.


maicki:
LGTM

PR #2261, opened 2016-09-16
[Layout] Add style property to `ASLayoutable`

maicki:
Since the size property (and many others) are modeled after CSS, we add a `style` property to `ASLayoutable` of type `ASLayoutableStyleDeclaration`  that has all the CSS properties like `minWidth`, `width`, `maxWidth`, `minHeight`, `height`, `maxHeight`, `flex properties`, etc.

## TODO:
- [X] Should style alike property also be added  to `ASStackLayoutSpec`?
- [x] Be extensible by client code


maicki:
@Adlai-Holler @appleguy @levi Can you please take a first look over it

Adlai-Holler:
@maicki 

- The class names are unpleasant. How about replace `ASLayoutableStyleDeclaration` with `ASStyle`? It has the word ass in it, which is unfortunate and kind of funny.
- Our dimension creating is also unpleasant.
   - Ideally we could do something clever with macros to enable: `style.minHeight = ASDimensionMake(100%)`. And somehow get the preprocessor to notice the suffix.
   - We could optimize the string-based creation method (currently it's quite expensive) but it's still a little ugly.
   - Now that we have a style object, we could expose: `minHeight: CGFloat, minHeightPercent: CGFloat, minHeightAuto: BOOL` which are all interfaces to the same underlying ASDimensionValue.
- The word `Declaration` in e.g. `ASStackLayoutStyleDeclaration` is baffling to me. If we use `ASStyle` for the main style object, let's keep storing layoutspec-specific style options (like centeringOptions) directly on the layout spec object and remove these intermediate objects.
- What do you think about adding a convenience method like

```objectivec
[titleNode styleWithBlock:^(ASStyle *style) {
   style.flexShrink = NO;
   style.minWidth = ASDimensionMakePoints(255);
   style.maxHeight = ASDimensionMakePoints(200);
}];
```

levi:
LGTM

PR #2263, opened 2016-09-19
[ASDisplayNode] Ensure ASHierarchyState propagation does not jump discontinuities in the node hierarchy.

appleguy:
Scenario: An ASCollectionNode is a subnode of an ASCellNode.  A layout transition is started, resulting in the removal of the ASCollectionNode as a subnode.  As it is removed, the hierarchy state is cleared - including the "range managed" bit - on the ASCollectionNode.  However, the deep recursion traverses the layer hierarchy too, and clears this bit on the ASCellNodes inside the ASCollectionNode.  A moment later, the collection performs its final ASRangeController update to mark its cells as invisible and free memory.  Then an assertion is triggered in ASRangeController, because it is operating on nodes that do not have the "range managed" bit set.

It turns out that ASInterfaceState also propogates in this way, but that behavior is efficient and beneficial in its current configuration (it assists how multi-dimensional preloading works).  However, hierarchy state should never need to jump discontinuities in the node hierarchy.  For now, disabling that case and will revisit other use cases soon.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNodeExtras.h
+++ b/AsyncDisplayKit/ASDisplayNodeExtras.h
@@ -82,16 +82,16 @@ extern ASDisplayNode * _Nullable ASViewToDisplayNode(UIView * _Nullable view);
 extern ASDisplayNode *ASDisplayNodeUltimateParentOfNode(ASDisplayNode *node);
 
 /**
- This function will walk the layer hierarchy, spanning discontinuous sections of the node hierarchy (e.g. the layers
- of UIKit intermediate views in UIViewControllers, UITableView, UICollectionView).
+ If traverseSublayers == YES, this function will walk the layer hierarchy, spanning discontinuous sections of the node hierarchy\
+ (e.g. the layers of UIKit intermediate views in UIViewControllers, UITableView, UICollectionView).
  In the event that a node's backing layer is not created yet, the function will only walk the direct subnodes instead
  of forcing the layer hierarchy to be created.
  */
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASDisplayNodeExtras.h
```diff
@@ -82,24 +82,24 @@ extern ASDisplayNode * _Nullable ASViewToDisplayNode(UIView * _Nullable view);
 extern ASDisplayNode *ASDisplayNodeUltimateParentOfNode(ASDisplayNode *node);
 
 /**
- This function will walk the layer hierarchy, spanning discontinuous sections of the node hierarchy (e.g. the layers
- of UIKit intermediate views in UIViewControllers, UITableView, UICollectionView).
+ If traverseSublayers == YES, this function will walk the layer hierarchy, spanning discontinuous sections of the node hierarchy\
```
maicki: Did you add a `\` at the end?

AsyncDisplayKit/ASDisplayNodeExtras.h
```diff
@@ -82,24 +82,24 @@ extern ASDisplayNode * _Nullable ASViewToDisplayNode(UIView * _Nullable view);
...
+ (e.g. the layers of UIKit intermediate views in UIViewControllers, UITableView, UICollectionView).
  In the event that a node's backing layer is not created yet, the function will only walk the direct subnodes instead
  of forcing the layer hierarchy to be created.
  */
```
maicki: Should we add to the documentation that if `traversingSublayers = NO` we still try to talk the node hierarchy?

PR #2266, opened 2016-09-19
Add support to image alignment in ASButtonNode

rmls:
Currently `ASButtonNode`, when both image and text co-exist, the only possible placement combination is image at the left and text at the right. This is because in the `layoutSpecThatFits:` implementation in `ASButtonNode`, an horizontal `ASStackLayoutSpec` is used and no customization is possible.

Subclassing `ASButtonNode` to solve this is not trivial as well given the fact that the stack spec might be a child of another spec.

### DIFF
--- a/AsyncDisplayKit/ASButtonNode.h
+++ b/AsyncDisplayKit/ASButtonNode.h
@@ -15,10 +15,10 @@
  Image alignment defines where the image will be placed relative to the text.
  */
 typedef NS_ENUM(NSInteger, ASButtonNodeImageAlignment) {
-  /** Places the image at the left side of the text. */
-  ASButtonNodeImageAlignmentLeft  = 0,
-  /** Places the image at the right side of the text. */
-  ASButtonNodeImageAlignmentRight = 1 << 0
+  /** Places the image before the text. */
+  ASButtonNodeImageAlignmentBeginning  = 0,
+  /** Places the image after the text. */
+  ASButtonNodeImageAlignmentEnd = 1 << 0
 };
 
 @interface ASButtonNode : ASControlNode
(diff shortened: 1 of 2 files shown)


levi:
Thanks for putting this up, Rocir! Looks good on my end.

maicki:
LGTM

AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -480,7 +481,11 @@ - (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
   }
   
   if (_titleNode.attributedText.length > 0) {
-    [children addObject:_titleNode];
+    if (_imageAlignment == ASButtonNodeImageAlignmentBeginning) {
+      [children addObject:_titleNode];
+    } else {
+      [children insertObject:_titleNode atIndex:0];
```
Adlai-Holler: This could certainly be changed later, but it's a bit unintuitive that the image alignment option would actually affect the location of the _other_ nodes in the stack.

AsyncDisplayKit/ASButtonNode.h
```diff
@@ -11,6 +11,16 @@
...
 #import <AsyncDisplayKit/ASImageNode.h>
 
+/**
+ Image alignment defines where the image will be placed relative to the text.
+ */
+typedef NS_ENUM(NSInteger, ASButtonNodeImageAlignment) {
+  /** Places the image before the text. */
+  ASButtonNodeImageAlignmentBeginning  = 0,
+  /** Places the image after the text. */
+  ASButtonNodeImageAlignmentEnd = 1 << 0
```
hannahmbanana: cc @rmls

AsyncDisplayKit/ASButtonNode.mm
rmls: I agree. In this case, though, earlier in this method we create the `children` array with capacity set to 2. Although it could still grow, but this spec has been setup to have only two children. What this does, only, is to add the title at the beginning in case the image is not.

PR #2268, opened 2016-09-20
[Layout] Add assertions for creating an invalid ASDimension from type ASDimensionUnitAuto and ASDimensionUnitFraction

maicki:
- Enabling assertions for creating auto dimensions via `ASDimensionMake` and providing a value != 0. There is also a constant that actually makes creating auto dimensions called: `ASDimensionAuto`
- Enabling assertions for creating fraction dimensions within the range of 0.0 and 1.0


levi:
Land awway!

PR #2274, opened 2016-09-22
[ASDataController] Call All Completion Handlers after Nested Batch Updates

Adlai-Holler:
Prior to this diff, we would only call the outermost completion handler.

Ready for review! @appleguy @maicki @levi

### DIFF
--- a/AsyncDisplayKit/Details/ASChangeSetDataController.mm
+++ b/AsyncDisplayKit/Details/ASChangeSetDataController.mm
@@ -49,9 +49,12 @@
   if (_changeSetBatchUpdateCounter == 0) {
     [self invalidateDataSourceItemCounts];
     [_changeSet markCompletedWithNewItemCounts:[self itemCountsFromDataSource]];
+    void (^batchCompletion)(BOOL finished) = _changeSet.completionHandler;
 
     if (!self.initialReloadDataHasBeenCalled) {
-      _changeSet.completionHandler(YES);
+      if (batchCompletion != nil) {
+        batchCompletion(YES);
+      }
       _changeSet = nil;
       return;
     }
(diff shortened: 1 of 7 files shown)


AsyncDisplayKit/Details/ASChangeSetDataController.mm
```diff
@@ -20,6 +20,11 @@ @implementation ASChangeSetDataController {
   _ASHierarchyChangeSet *_changeSet;
 }
 
+- (void)dealloc
+{
+  ASDisplayNodeCAssert(_changeSetBatchUpdateCounter == 0, @"ASChangeSetDataController deallocated in the middle of a batch update.");
```
Adlai-Holler: Used C assert to avoid referencing self during dealloc.

AsyncDisplayKit/Private/_ASHierarchyChangeSet.mm
```diff
@@ -132,6 +133,31 @@ - (instancetype)initWithOldData:(std::vector<NSInteger>)oldItemCounts
...
+- (void (^)(BOOL finished))completionHandler
+{
+  [self _ensureCompleted];
+
+  void (^completionHandler)(BOOL) = _completionHandler;
+  _completionHandler = nil;
+  return completionHandler;
+}
+
+- (void)addCompletionHandler:(void (^)(BOOL))completion
```
maicki: Should we not call it `setCompletionHandler`? `addCompletionHandler` is more like adding an observer instead of replacing the completion handler

Adlai-Holler: The method actually does _add_ a completion handler. I've updated the documentation to explain what's going on.

maicki:
LGTM

PR #2279, opened 2016-09-23
[Layout] Remove styleClass property from ASLayoutable for now

maicki:
Currently the styleClass property does not give us any benefits. Let's remove it for now until we found a more though trough way to add extensibility for styles.


hannahmbanana:
Shouldn't the naming of this be `ASLayoutElementStyle`?

maicki:
@Adlai-Holler That's a cool idea. I will dig into that a bit later though.

PR #2281, opened 2016-09-23
[ASCellNode] Document Cell Node Event Callback Behavior

Adlai-Holler:
This documents some unexpected behavior @binl observed, per discussion with @appleguy.


maicki:
LGTM

PR #2283, opened 2016-09-23
[Layout] Add delegate to ASLayoutableStyle for property changes

maicki:
Add a delegate to be able to get property changes.

Furthermore we move the `size` property from `readwrite`to `readonly`

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayoutable.h
+++ b/AsyncDisplayKit/Layout/ASLayoutable.h
@@ -150,18 +150,47 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+
 #pragma mark - ASLayoutableStyle
+
+extern NSString * const ASLayoutableStyleWidthProperty;
+extern NSString * const ASLayoutableStyleMinWidthProperty;
+extern NSString * const ASLayoutableStyleMaxWidthProperty;
+
+extern NSString * const ASLayoutableStyleHeightProperty;
+extern NSString * const ASLayoutableStyleMinHeightProperty;
+extern NSString * const ASLayoutableStyleMaxHeightProperty;
+
+extern NSString * const ASLayoutableStyleSpacingBeforeProperty;
+extern NSString * const ASLayoutableStyleSpacingAfterProperty;
+extern NSString * const ASLayoutableStyleFlexGrowProperty;
+extern NSString * const ASLayoutableStyleFlexShrinkProperty;
+extern NSString * const ASLayoutableStyleFlexBasisProperty;
+extern NSString * const ASLayoutableStyleAlignSelfProperty;
+extern NSString * const ASLayoutableStyleAscenderProperty;
+extern NSString * const ASLayoutableStyleDescenderProperty;
+
+extern NSString * const ASLayoutableStyleLayoutPositionProperty;
+
+@protocol ASLayoutableStyleDelegate <NSObject>
+- (void)style:(id)style propertyDidChange:(NSString *)propertyName;
+@end
 
 @interface ASLayoutableStyle : NSObject <ASStackLayoutable, ASStaticLayoutable>
 
+#pragma mark - Delegate
 
-#pragma mark - Sizing
+/**
+ * @abstract The object that acts as the delegate of the style.
+ *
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/Layout/ASLayoutable.mm
```diff
@@ -192,4 +235,62 @@ - (void)setExactSizeWithCGSize:(CGSize)size
   self.maxHeight = ASDimensionMakeWithPoints(size.height);
 }
 
+#pragma mark - ASStackLayoutable
+
+- (void)setSpacingBefore:(CGFloat)spacingBefore
+{
+  _spacingBefore = spacingBefore;
+  ASLayoutableStyleCallDelegate(ASLayoutableStyleSpacingBeforeProperty);
```
Adlai-Holler: What's the rationale for not locking these methods?

AsyncDisplayKit/Layout/ASLayoutable.h
```diff
@@ -150,18 +150,52 @@ NS_ASSUME_NONNULL_BEGIN
...
+extern NSString * const ASLayoutableStyleFlexShrinkProperty;
+extern NSString * const ASLayoutableStyleFlexBasisProperty;
+extern NSString * const ASLayoutableStyleAlignSelfProperty;
+extern NSString * const ASLayoutableStyleAscenderProperty;
+extern NSString * const ASLayoutableStyleDescenderProperty;
+
+extern NSString * const ASLayoutableStyleLayoutPositionProperty;
+
+@protocol ASLayoutableStyleDelegate <NSObject>
+- (void)style:(id)style propertyDidChange:(NSString *)propertyName;
```
Adlai-Holler: Would `__kindof ASLayoutableStyle *` be a better type for the sender param?

AsyncDisplayKit/Layout/ASLayoutable.h
```diff
@@ -150,18 +150,52 @@ NS_ASSUME_NONNULL_BEGIN
...
+
 @interface ASLayoutableStyle : NSObject <ASStackLayoutable, ASStaticLayoutable>
 
+#pragma mark - Delegate
 
-#pragma mark - Sizing
+/**
+ * @abstract Initializes the layoutable style with a specified delegate
+ */
+- (instancetype)initWithDelegate:(id<ASLayoutableStyleDelegate>)delegate;
```
Adlai-Holler: Looking to the future a bit, it's likely that `delegate` will have to be `readwrite` because there are plans to reuse these style objects to save resources.

AsyncDisplayKit/Layout/ASLayoutable.mm
```diff
@@ -174,10 +218,9 @@ - (void)setMaxHeight:(ASDimension)maxHeight
 {
   ASDN::MutexLocker l(__instanceLock__);
   _size.maxHeight = maxHeight;
+  ASLayoutableStyleCallDelegate(ASLayoutableStyleMaxHeightProperty);
```
Adlai-Holler: Note for the record: Michael and I discussed whether to hold the lock during this delegate call. It's a tough call, and it's unlikely to matter, but I advocated to do so on the basis that NSCache does the same, and that informing the delegate should be "part of" the same atomic operation.

Since we are currently using a recursive mutex – which may be worth changing in a class as simple as this one – it is almost certain not to make any difference.

hannahmbanana:
test this please

AsyncDisplayKit/Layout/ASLayoutable.h
maicki: Yep agree but I think for now we should go for the "closest" implementation

AsyncDisplayKit/Layout/ASLayoutable.h
maicki: Sounds good

AsyncDisplayKit/Layout/ASLayoutable.mm
maicki: They should be locked

maicki:
@Adlai-Holler comments addressed

PR #2285, opened 2016-09-23
Fix example projects to use Layout API 2.0 style object

hannahmbanana:
the example projects are the current cause of our Travis failures


maicki:
LGTM

PR #2286, opened 2016-09-23
I don't believe it's safe to return a vector by reference created in …

garrettmoon:
…a function.


hannahmbanana:
the title of this PR is very catchy :)

maicki:
Overall PR LGTM too

Yeah @Adlai-Holler and me talked about and it's very likely not the root cause for the crash we are seeing ...

garrettmoon:
test this please

PR #2291, opened 2016-09-24
Add More Nullability Annotations to Public Headers

Adlai-Holler:
I was poking around in the Swift example this morning and I found some missing nullability specifiers. These make for an extremely unpleasant Swift experience.

After this diff, all of our public headers (the ones in the root folder anyway) have full nullability specifications.


appleguy:
test this please

maicki:
LGTM

PR #2292, opened 2016-09-24
[UIImageHelpers] Refactor and Optimize Rounded Image Generation

Adlai-Holler:
I should be enjoying time with my family, but when I see a "freebie" like this I want to get it done so I can stop thinking about it 🤓

- Use `kCGBlendModeCopy` when possible, which should be fairly often.
- Simplify path cache key generation – use `NSValue` instead of the admittedly-excellent `NSNumber` hack.
- Remove `@synchronized` around path cache access. By doing this, it is possible that the same path is generated multiple times on different threads, but it's very unlikely, and not worth adding significant overhead to all cases.

Ready for review @maicki @levi @appleguy @hannahmbanana !

### DIFF
--- a/AsyncDisplayKit/UIImage+ASConvenience.m
+++ b/AsyncDisplayKit/UIImage+ASConvenience.m
@@ -12,11 +12,10 @@
 
 #import "UIImage+ASConvenience.h"
 #import <UIKit/UIKit.h>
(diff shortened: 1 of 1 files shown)


appleguy:
test this please

maicki:
LGTM

PR #2294, opened 2016-09-24
[Layout] Add extensibility support to ASLayoutableStyle

maicki:
With the introduction of `ASLayoutableStyle` that combines all style related properties for a `ASLayoutElement` we need to provide some kind of extensibility support. This is important as it should be possible to add custom layout specs and extend the `ASLayoutableStyle` for usage within the custom layout specs.

The general idea is outline in this gist via examples: https://gist.github.com/maicki/2630b51f62cbedc0cb34e0bcac96d6c9


appleguy:
test this please

maicki:
As a lot of work went into the layout infrastructure that will cause merge conflicts as well as further investigation for this kind of extensibility needs to happen, let's close this PR.

PR #2295, opened 2016-09-25
[Testing] Upgrade snapshot tests to support IOS 10

hannahmbanana:
Our `ASTextNodeSnapshotTests` were causing the Jenkins build to fail.


build.sh
```diff
@@ -2,7 +2,7 @@
 
 # **** Update me when new Xcode versions are released! ****
 PLATFORM="platform=iOS Simulator,OS=9.3,name=iPhone 6"
-SDK="iphonesimulator9.3"
+SDK="iphonesimulator10.0"
```
hannahmbanana: Not sure if we should merge this, just curious to see if this fixes Jenkins.

hannahmbanana: Removed

appleguy:
Thanks for the fix @hannahmbanana — I tried this locally on my iOS 10 instance because Jenkins is failing, and it does succeed, so I think there is a separate Jenkins issue.

PR #2298, opened 2016-09-25
[Layout] Add support for flex factor

maicki:
The Flexbox specification states that flex-grow and flex-shrink should be integer value (non-negative integers only) that set the flex grow factor and flex shrink factor, respectively. Currently ASStackLayoutSpec only support boolean values for these properties.

The [Flexbox specification](https://www.w3.org/TR/css-flexbox-1/) actually relies on the [flex-grow](https://www.w3.org/TR/css-flexbox-1/#flex-flex-grow-factor) and [flex-shrink](https://www.w3.org/TR/css-flexbox-1/#flex-flex-shrink-factor) properties to set the flex grow factor and flex shrink factor, respectively. These "flex factors" actually describe how any positive or negative free space should be distributed across flex items.

**TODO:**
- [x] Improve / Add new tests
- [x] Recreate snapshot images for tests


maicki:
@Adlai-Holler @appleguy @levi Ready for review

garrettmoon:
test this please

appleguy:
test this please

appleguy:
last comment was to trigger a build

maicki:
@Adlai-Holler Yes the reference images change was expected.

PR #2303, opened 2016-09-26
[ASTextKitFontSizeAdjuster] Use the constrainedSize’s height to adjust font scaling

rcancro:
Previously we used `maximumNumberOfLines` and the longest word in a string to scale font size. This adds taking the `constrainedSize`’s height into consideration. If, without scaling, a string’s height is larger than the `constrainedSize` height, apply scaling.

Also fixed up the ASTraitCollection sample to work with the new layout API.

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
@@ -150,7 +150,7 @@
     }
     
     NSUInteger scaleIndex = 0;
-    NSAttributedString *scaledString = [[NSMutableAttributedString alloc] initWithAttributedString:textStorage];
+    NSMutableAttributedString *scaledString = [[NSMutableAttributedString alloc] initWithAttributedString:textStorage];
 
     // find the longest word and make sure it fits in the constrained width
     if ([longestWordNeedingResize length] > 0) {
@@ -196,7 +196,6 @@
           
           if ([self lineCountForString:scaledString] <= _attributes.maximumNumberOfLines) {
             // we fit! we are done
-            [[self class] adjustFontSizeForAttributeString:scaledString withScaleFactor:adjustedScale];
             break;
           }
         }
--- a/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
@@ -45,6 +45,8 @@
 
 + (void)adjustFontSizeForAttributeString:(NSMutableAttributedString *)attrString withScaleFactor:(CGFloat)scaleFactor
 {
+  if (scaleFactor == 1.0) return;
+  
   [attrString beginEditing];
 
   // scale all the attributes that will change the bounding box
@@ -133,7 +135,7 @@
   
   __block CGFloat adjustedScale = 1.0;
   
-  NSArray *scaleFactors = _attributes.pointSizeScaleFactors;
+  NSArray *scaleFactors = [@[@(1)] arrayByAddingObjectsFromArray:_attributes.pointSizeScaleFactors];
   [_context performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
     
     // Check for two different situations (and correct for both)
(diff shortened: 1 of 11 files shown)


hannahmbanana:
Restarting the Travis build for a seemingly random failure.

hannahmbanana:
Review please? @garrettmoon @maicki @Adlai-Holler

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -179,23 +181,42 @@ - (CGFloat)scaleFactor
...
+          if ([self lineCountForString:scaledString] <= _attributes.maximumNumberOfLines) {
             // we fit! we are done
             break;
           }
         }
+      }
+    }
+    
+    // check to see if the string's height is greater that the constrained height
+    CGSize stringSize = [scaledString boundingRectWithSize:CGSizeMake(_constrainedSize.width, CGFLOAT_MAX) options:NSStringDrawingUsesLineFragmentOrigin context:nil].size;
```
garrettmoon: We currently use NSLayoutManager and NSTextContainer via ASTextKitRenderer to size strings. Any way we can use ASTextKitRenderer to do this?

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -179,23 +181,42 @@ - (CGFloat)scaleFactor
...
       // get the number of lines in our possibly scaled string
-      NSUInteger numberOfLines = [self lineCountForString:textStorage];
+      NSUInteger numberOfLines = [self lineCountForString:scaledString];
       if (numberOfLines > _attributes.maximumNumberOfLines) {
         
         for (NSUInteger index = scaleIndex; index < scaleFactors.count; index++) {
-          NSMutableAttributedString *entireAttributedString = [[NSMutableAttributedString alloc] initWithAttributedString:textStorage];
-          [[self class] adjustFontSizeForAttributeString:entireAttributedString withScaleFactor:[scaleFactors[index] floatValue]];
-          
+          scaledString = [[NSMutableAttributedString alloc] initWithAttributedString:textStorage];
```
garrettmoon: Can you create a new var here and add a comment that it needs to be reset each loop because it's being mutated below? Maybe the comment is unnecessary if you create a new var here…

garrettmoon:
It may actually be the right call to switch ASTextKitRenderer to use NSString's sizing methods too. Might resolve some of the bugs we've seen with sizing using `usedRectForTextContainer`

garrettmoon:
Hmm, looks like we were already using boundingRectWithSize, so I'm ok with this going in.

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
rcancro: I keep scaled string alive outside of the loops so that i can check line count/text height on the string that may have been scaled by the previous check. I added a comment to make this clearer.

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -179,23 +181,47 @@ - (CGFloat)scaleFactor
...
+        // save away this scale factor. Even if we don't fit completely we should still scale down
+        adjustedScale = [scaleFactors[index] floatValue];
+        [[self class] adjustFontSizeForAttributeString:scaledString withScaleFactor:adjustedScale];
+        
+        stringSize = [scaledString boundingRectWithSize:CGSizeMake(_constrainedSize.width, CGFLOAT_MAX) options:NSStringDrawingUsesLineFragmentOrigin context:nil].size;
+        if (stringSize.height <= _constrainedSize.height) {
+          // we fit! we are done
+          break;
+        }
       }
```
garrettmoon: Is there any way we can combine the constrainedSize check with the maximum line number check? Looks to be a lot of similar code. Just a question, as I can imagine it actually becomes more convoluted with that change.

appleguy:
test this please

appleguy:
Last comment to trigger build

hannahmbanana:
test this please

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
rcancro: I'll take a look. I'm afraid that it may make the code more bug-prone but agree that it could be cleaner

PR #2309, opened 2016-09-27
[ASTextKitFontSizeAdjuster] Use the constrainedSize’s height to adjust font scaling

rcancro:
Previously we used maximumNumberOfLines and the longest word in a string to scale font size. This adds taking the constrainedSize’s height into consideration. If, without scaling, a string’s height is larger than the constrainedSize height, apply scaling.

Also fixed up the ASTraitCollection sample to work with the new layout API.

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
@@ -154,9 +154,6 @@
       }
     }
     
-    NSUInteger scaleIndex = 0;
-    NSMutableAttributedString *scaledString = [[NSMutableAttributedString alloc] initWithAttributedString:textStorage];
-    
     // check to see if we may need to shrink for any of these things
     BOOL longestWordFits = [longestWordNeedingResize length] ? NO : YES;
     BOOL maxLinesFits = _attributes.maximumNumberOfLines > 0 ? NO : YES;
--- a/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
@@ -158,44 +158,49 @@
     BOOL longestWordFits = [longestWordNeedingResize length] ? NO : YES;
     BOOL maxLinesFits = _attributes.maximumNumberOfLines > 0 ? NO : YES;
     BOOL heightFits = isinf(_constrainedSize.height) ? YES : NO;
+
+    CGSize longestWordSize = CGSizeZero;
+    if (longestWordFits == NO) {
+        NSRange longestWordRange = [str rangeOfString:longestWordNeedingResize];
+        NSAttributedString *attrString = [textStorage attributedSubstringFromRange:longestWordRange];
+        longestWordSize = [attrString boundingRectWithSize:CGSizeMake(CGFLOAT_MAX, CGFLOAT_MAX) options:NSStringDrawingUsesLineFragmentOrigin context:nil].size;
+    }
     
-    if (longestWordFits == NO || maxLinesFits == NO || heightFits == NO) {
-      // we may need to shrink for some reason, so let's iterate through our scale factors to see if we actually need to shrink
-      // Note: the first scale factor in the array is 1.0 so will make sure that things don't fit without shrinking
-      for (NSUInteger index = 0; index < scaleFactors.count && (longestWordFits == NO || maxLinesFits == NO || heightFits == NO); index++) {
+    // we may need to shrink for some reason, so let's iterate through our scale factors to see if we actually need to shrink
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -133,7 +135,7 @@ - (CGFloat)scaleFactor
   
   __block CGFloat adjustedScale = 1.0;
   
-  NSArray *scaleFactors = _attributes.pointSizeScaleFactors;
+  NSArray *scaleFactors = [@[@(1)] arrayByAddingObjectsFromArray:_attributes.pointSizeScaleFactors];
```
garrettmoon: Can you also put an example array in the header (ASTextNode+Beta.h)

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -150,52 +152,47 @@ - (CGFloat)scaleFactor
...
+    NSMutableAttributedString *scaledString = [[NSMutableAttributedString alloc] initWithAttributedString:textStorage];
+    
+    // check to see if we may need to shrink for any of these things
+    BOOL longestWordFits = [longestWordNeedingResize length] ? NO : YES;
+    BOOL maxLinesFits = _attributes.maximumNumberOfLines > 0 ? NO : YES;
+    BOOL heightFits = isinf(_constrainedSize.height) ? YES : NO;
+    
+    if (longestWordFits == NO || maxLinesFits == NO || heightFits == NO) {
+      // we may need to shrink for some reason, so let's iterate through our scale factors to see if we actually need to shrink
+      // Note: the first scale factor is the array is 1.0 so will make sure that things don't fit without shrinking
```
garrettmoon: nit: typo "the first scale factor *in* the array is 1.0"

appleguy:
test this please

rcancro:
I don't understand these build failures. They are complaining about unused variables in ASTextKitFontSizeAdjuster, but it looks like they are variables from before the refactor...

screen shot 2016-09-28 at 8 05 31 am

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -149,53 +154,45 @@ - (CGFloat)scaleFactor
...
-          
-          if (std::ceil(longestWordSize.width * [scaleFactor floatValue])  <= _constrainedSize.width) {
-            // we fit! we are done
-            break;
-          }
+        adjustedScale = [scaleFactors[index] floatValue];
+        
+        if (longestWordFits == NO) {
+          // we need to check the longest word to make sure it fits
+          NSRange longestWordRange = [str rangeOfString:longestWordNeedingResize];
```
Adlai-Holler: `longestWordRange` is an invariant, so we can scan `str` to get it once outside the for loop.

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -149,53 +154,45 @@ - (CGFloat)scaleFactor
...
+          NSMutableAttributedString *attrString = [textStorage attributedSubstringFromRange:longestWordRange].mutableCopy;
```
Adlai-Holler: I think we can remove the `mutableCopy` here.

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
```diff
@@ -149,53 +154,45 @@ - (CGFloat)scaleFactor
...
-          adjustedScale = [scaleFactor floatValue];
-          
-          // adjust here so we start at the proper place in our scale array if we have too many lines
-          scaleIndex++;
-          
-          if (std::ceil(longestWordSize.width * [scaleFactor floatValue])  <= _constrainedSize.width) {
-            // we fit! we are done
-            break;
-          }
+        adjustedScale = [scaleFactors[index] floatValue];
```
Adlai-Holler: We can use fast enumeration here, like

```objectivec
for (NSNumber *adjustedScaleObj in scaleFactors) {
  if (longestWordFits && maxLinesFits && heightFits) {
    break;
  }
  CGFloat adjustedScale = adjustedScaleObj.floatValue;
  // Work it
}
```

This approach has the drawback of always _starting_ enumeration, even if we're going to break immediately, but that's not very costly (doesn't touch the heap) and we achieve a good balance of performance and readability.

AsyncDisplayKit/TextKit/ASTextKitFontSizeAdjuster.mm
rcancro: Good call. I think we can move all of this except the multiplication out of the loop.

PR #2311, opened 2016-09-27
[Layout] Remove legacy layout validation code

maicki:
As a lot changed recently in the layout code I think it's time to remove the legacy layout validation code and come up with a better solution.

In our effort to remove legacy and mostly confusing code let's get rid of it now.


garrettmoon:
I agree, tests feel like a better way to validate layout.

levi:
👍 validation can be done statically with this new layout style API.

appleguy:
test this please

appleguy:
test this please

PR #2315, opened 2016-09-28
[ASTextNode] Disable ASTextKitRenderer cache

maicki:
Due to recent changes in the text system the `ASTextKitRenderer` is not build in a thread safe way anymore and if the same renderer is used in multiple text nodes can cause crashes.

Disable the renderer cache for the time being.


maicki:
@Adlai-Holler @levi @appleguy can i please have a review

PR #2310, opened 2016-09-27
[ASDisplayNode] Add allowsGroupOpacity property

levi:
Since we're disabling it in our app, provide easy access for changing the layer's behavior.

PR #2320, opened 2016-09-29
[ASDisplayNode] Add allowsGroupOpacity property to ASDisplayNode

maicki:
Since we're disabling it in our app, provide easy access for changing the layer's behavior.

Replaces #2310

### DIFF
--- a/AsyncDisplayKit/Private/_ASPendingState.mm
+++ b/AsyncDisplayKit/Private/_ASPendingState.mm
@@ -56,6 +56,7 @@ typedef struct {
   int setBorderWidth:1;
   int setBorderColor:1;
   int setAsyncTransactionContainer:1;
+  int setAllowsGroupOpacity:1;
   int setAllowsEdgeAntialiasing:1;
   int setEdgeAntialiasingMask:1;
   int setIsAccessibilityElement:1;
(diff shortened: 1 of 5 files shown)


maicki:
@levi @Adlai-Holler @appleguy Can I get a look over it please

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -207,6 +209,7 @@ - (instancetype)init
   contents = nil;
   isHidden = NO;
   needsDisplayOnBoundsChange = NO;
+  allowsGroupOpacity = YES;
```
levi: You don't want to set a default for this, since we've disabled it in our app.

maicki: We have to read the default value from a `UIView` or from the `Info.plist`of the project. Addressed

AsyncDisplayKit/Private/_ASPendingState.mm
```diff
@@ -195,6 +199,12 @@ - (instancetype)init
     CFRetain(blackColorRef);
     CGColorSpaceRelease(colorSpace);
     defaultTintColor = [UIColor colorWithRed:0.0 green:0.478 blue:1.0 alpha:1.0];
+    
+    // Create temporary view to read default values that are based on linked SDK and Info.plist values
+    UIView *view = [[UIView alloc] init];
```
Adlai-Holler: This may not be called on the main thread =/

levi:
@maicki we shouldn't add an extra +load method for this case. Let's just read the info.plist dictionary directly from the bundle.

PR #2321, opened 2016-09-29
[ASCellNode] Fix applyLayoutAttributes Not Called, Add Readonly Property

Adlai-Holler:
I came across these issues while working with some excellent engineers from Pinterest yesterday.

- `applyLayoutAttributes` was never being called on `ASCellNode`. See the documentation in the diff for why.
- It is useful to be able to safely get the layout attributes for a node in `willDisplayNode:forItemAtIndexPath:`. At this point you can't convert the frame because the node's not in the hierarchy yet. See the documentation in the diff for why.

The product engineers _do_ have a workaround in place for this, but we should merge this as soon as possible. Ready for review @maicki @appleguy @levi @garrettmoon @rcancro

### DIFF
--- a/AsyncDisplayKit/ASCellNode+Internal.h
+++ b/AsyncDisplayKit/ASCellNode+Internal.h
@@ -12,6 +12,8 @@
 
 #import "ASCellNode.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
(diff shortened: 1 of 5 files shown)


maicki:
LGTM

PR #2323, opened 2016-09-30
[ASButtonNode] Only change state if it's different

hannahmbanana:
To avoid unnecessary layout pass in Pin Closeup. 

While testing the sync candidate branch for Pinterest, I noticed that the pin-to-pin closeup swipe seemed slower than normal. Unnecessary layout passes are being called every time the action bar buttons are updated for a new pin. 

Before
before

After
after

@appleguy: could you review & merge so I can get this into `6.12` and `6.13` tonight?

### DIFF
--- a/AsyncDisplayKit/ASButtonNode.mm
+++ b/AsyncDisplayKit/ASButtonNode.mm
@@ -110,25 +110,31 @@
 
 - (void)setEnabled:(BOOL)enabled
 {
-  [super setEnabled:enabled];
-  if (enabled) {
-    self.accessibilityTraits = UIAccessibilityTraitButton;
-  } else {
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASButtonNode.mm
```diff
@@ -110,25 +110,31 @@ - (void)setLayerBacked:(BOOL)layerBacked
...
+  }
 }
 
 - (void)setSelected:(BOOL)selected
 {
-  [super setSelected:selected];
-  [self updateButtonContent];
+  if (self.selected != selected) {
+    [super setSelected:selected];
+    [self updateButtonContent];
```
appleguy: I see.  I probably wrote the original code here...shorter, but not efficient.

PR #2324, opened 2016-09-30
[Layout] Add helper properties for setting sizes via CGSize and ASRelativeSize

maicki:
After working a bit on the layout documentation as well as thinking more of an API perspective I think the following two methods on `ASLayoutElementStyle` are not well chosen and we should in a first step remove it.

```Swift
func setSizeWithCGSize(size: CGSize)
func setExactSizeWithCGSize(size: CGSize)
```

Furthermore we will introduce helper properties for setting a size (besides `width`, `height`, `minWidth`, `minHeight` ...) of an `ASLayoutElementStyle` via a `CGSize`or `ASRelativeSize`:

**Write-Only Properties:**
```Swift
var preferredSize: CGSize { set }
var minSize: CGSize { set }
var maxSize: CGSize { set }
```

**Read-Write properties:**
```Swift
var preferredRelativeSize: ASRelativeSize { get set }
var minRelativeSize: ASRelativeSize { get set }
var maxRelativeSize: ASRelativeSize { get set }
```

**TODO:**
- [x] Add documentation for new size helper properties
- [x] Adjust examples


appleguy:
@maicki this needs to update the example projects to succeed.

hannahmbanana:
First pass at the header docs.

/**
 * @abstract Provides a suggested size for a layout element. If the optional minSize or maxSize are provided, 
 * and the preferredSize exceeds these, the minSize or maxSize will be enforced. If this optional value is not 
 * provided, the layout element’s size will default to it’s intrinsic content size provided calculateSizeThatFits:
 * 
 * @discussion This method is optional, but one of either preferredSize or preferredRelativeSize is required 
 * for nodes that either have no intrinsic content size or 
 * should be laid out at a different size than its intrinsic content size. For example, this property could be 
 * set on an ASImageNode to display at a size different from the underlying image size. 
 */
@property (nonatomic, assign) CGSize preferredSize;
- (CGSize)preferredSize UNAVAILABLE_ATTRIBUTE;


 /**
 * @abstract An optional property that provides a minimum size bound for a layout element. If provided, this restriction will 
 * always be enforced. If a parent layout element’s minimum size is smaller than its child’s minimum size, the child’s  
 * minimum size will be enforced and its size will extend out of the layout spec’s.  
 * 
 * @discussion For example, if you set a preferred relative width of 50% and a minimum width of 200 points on an
 * element in a full screen container, this would result in a width of 160 points on an iPhone screen. However, 
 * since 160 pts is lower than the minimum width of 200 pts, the minimum width would be used.
 */
@property (nonatomic, assign) CGSize minSize;
- (CGSize)minSize UNAVAILABLE_ATTRIBUTE;
​

/**
 * @abstract An optional property that provides a maximum size bound for a layout element. If provided, this restriction will 
 * always be enforced.  If a child layout element’s maximum size is smaller than its parent, the child’s maximum size will 
 * be enforced and its size will extend out of the layout spec’s.  
 * 
 * @discussion For example, if you set a preferred relative width of 50% and a maximum width of 120 points on an
 * element in a full screen container, this would result in a width of 160 points on an iPhone screen. However, 
 * since 160 pts is higher than the maximum width of 120 pts, the maximum width would be used.
 */
@property (nonatomic, assign) CGSize maxSize;
- (CGSize)maxSize UNAVAILABLE_ATTRIBUTE;
​

/**
 * @abstract Provides a suggested RELATIVE size for a layout element. An ASRelativeSize uses percentages rather 
 * than points to specify layout. E.g. width should be 50% of the parent’s width. If the optional minRelativeSize or 
 * maxRelativeSize are provided, and the preferredRelativeSize exceeds these, the minRelativeSize or maxRelativeSize 
 * will be enforced. If this optional value is not provided, the layout element’s size will default to its intrinsic content size 
 * provided calculateSizeThatFits:
 */
@property (nonatomic, assign, readwrite) ASRelativeSize preferredRelativeSize;
​

 /**
 * @abstract An optional property that provides a minimum RELATIVE size bound for a layout element. If provided, this 
 * restriction will always be enforced. If a parent layout element’s minimum relative size is smaller than its child’s minimum 
 * relative size, the child’s minimum relative size will be enforced and its size will extend out of the layout spec’s.  
 */
@property (nonatomic, assign, readwrite) ASRelativeSize minRelativeSize;
​

 /**
 * @abstract An optional property that provides a maximum RELATIVE size bound for a layout element. If provided, this 
 * restriction will always be enforced. If a parent layout element’s maximum relative size is smaller than its child’s maximum 
 * relative size, the child’s maximum relative size will be enforced and its size will extend out of the layout spec’s.  
 */
@property (nonatomic, assign, readwrite) ASRelativeSize maxRelativeSize;

maicki:
@Adlai-Holler @appleguy Ready for review

PR #2332, opened 2016-10-01
[ASAsyncTransaction] Make System Less Specific to Layer Display, More Readable

Adlai-Holler:
The `ASAsyncTransaction` system is one of the most fundamental components of AsyncDisplayKit. It powers the asynchronous layer display process which is possibly the most reliable & performance-tuned component of the framework.

It provides lightweight, cancellable, waitable, transactionable operation semantics. It seems like it could be useful in other contexts than drawing, such as layout, or updating UIKit/CALayer properties from background threads. There's a ton of promise here.

The only problem with it, is that it works so reliably that @appleguy is the only maintainer who understands how it works and how it could be applied in other contexts.

This diff contains no functional changes. Its only purpose is to help make the system easier to understand and open the door for more general usage:

- Add nullability and generics to the system.
- Rename and move the `asyncdisplaykit_currentAsyncLayerTransaction` property out from the special `CALayer (ASAsyncTransaction)` category up into the `ASAsyncTransaction` protocol itself with the new name `asyncdisplaykit_currentAsyncTransaction`. Matching other cases like this, `UIView` and `ASDisplayNode` pass through to the layer in their conformance.
- Shorten names e.g. `ASDisplayNodeAsyncTransactionContainer` -> `ASAsyncTransactionContainer`
- Reduce underscores in `ASAsyncTransaction` so it's less scary.
- Replace references to `container layers` – which appear to be an old subclass of `CALayer` that's gone now – with either `layer` or `container` based on the context.

Ready for review @maicki @levi @appleguy

### DIFF
--- a/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
+++ b/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
@@ -10,14 +10,16 @@
 
 #import <Foundation/Foundation.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
 #define ASDISPLAYNODE_DELAY_DISPLAY 0
 
 @class _ASAsyncTransaction;
 
 typedef void(^asyncdisplaykit_async_transaction_completion_block_t)(_ASAsyncTransaction *completedTransaction, BOOL canceled);
-typedef id<NSObject>(^asyncdisplaykit_async_transaction_operation_block_t)(void);
-typedef void(^asyncdisplaykit_async_transaction_operation_completion_block_t)(id<NSObject> value, BOOL canceled);
-typedef void(^asyncdisplaykit_async_transaction_complete_async_operation_block_t)(id<NSObject> value);
+typedef id<NSObject> _Nullable(^asyncdisplaykit_async_transaction_operation_block_t)(void);
+typedef void(^asyncdisplaykit_async_transaction_operation_completion_block_t)(id<NSObject> _Nullable value, BOOL canceled);
+typedef void(^asyncdisplaykit_async_transaction_complete_async_operation_block_t)(id<NSObject> _Nullable value);
 typedef void(^asyncdisplaykit_async_transaction_async_operation_block_t)(asyncdisplaykit_async_transaction_complete_async_operation_block_t completeOperationBlock);
 
 /**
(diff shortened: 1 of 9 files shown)


maicki:
@Adlai-Holler you should try to remove the dependencies from ASDK for that component. E.g. :
```
#import "ASAssert.h"
#import "ASThread.h"
```

maicki:
LGTM

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
```diff
@@ -70,18 +72,18 @@ extern NSInteger const ASDefaultTransactionPriority;
...
  */
-@property (nonatomic, readonly, copy) asyncdisplaykit_async_transaction_completion_block_t completionBlock;
+@property (nonatomic, readonly, copy, nullable) asyncdisplaykit_async_transaction_completion_block_t completionBlock;
 
 /**
  The state of the transaction.
  @see ASAsyncTransactionState
  */
-@property (nonatomic, readonly, assign) ASAsyncTransactionState state;
+@property (readonly, assign) ASAsyncTransactionState state;
```
appleguy: @Adlai-Holler Shouldn't this still have nonatomic?

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -8,22 +8,24 @@
 //  of patent rights can be found in the PATENTS file in the same directory.
 //
 
+#import <UIKit/UIApplication.h>
```
appleguy: @Adlai-Holler I may have missed it, but what adds this requirement?  I'd prefer to not have a dependency on UIApplication at this level.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -317,7 +319,8 @@ - (NSString *)description
 @implementation _ASAsyncTransaction
 {
   ASAsyncTransactionQueue::Group *_group;
-  NSMutableArray *_operations;
+  NSMutableArray<ASAsyncTransactionOperation *> *_operations;
+  _Atomic(ASAsyncTransactionState) _state;
```
appleguy: Cool, never seen this before.  Does it enforce using atomic setters / getters?

AsyncDisplayKit/Details/Transactions/_ASAsyncTransactionGroup.h
```diff
@@ -8,19 +8,20 @@
 //  of patent rights can be found in the PATENTS file in the same directory.
 //
 
-#import <UIKit/UIKit.h>
```
appleguy: I see, removing this is a great change.  If the UIApplication dependency is needed, we should probably add a comment next to the import in a future diff to point out its utility.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.h
Adlai-Holler: No, because we implement the accessors to make it atomic. So removing `nonatomic` here is just documentation.

AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
Adlai-Holler: No, but we implemented the accessors to enforce it.

PR #2333, opened 2016-10-01
[ASAsyncTransaction] Remove dependencies for ASAssert and ASThread in the async transaction system

maicki:
To be able to extract our powerful async transaction system with it's own tests and example usage patterns this PR removes the ASDK dependencies.

- Move to C++ 11 mutex
- Replace ASDK assert defines

### DIFF
--- a/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
+++ b/AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
@@ -10,10 +10,12 @@
 
 #import "_ASAsyncTransaction.h"
 #import "_ASAsyncTransactionGroup.h"
-#import "ASAssert.h"
-#import "ASThread.h"
+
 #import <list>
 #import <map>
+#import <mutex>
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Details/Transactions/_ASAsyncTransaction.mm
```diff
@@ -10,10 +10,12 @@
...
 #import "_ASAsyncTransaction.h"
 #import "_ASAsyncTransactionGroup.h"
-#import "ASAssert.h"
-#import "ASThread.h"
+
 #import <list>
 #import <map>
+#import <mutex>
+
+#define ASAsyncTransactionAssertMainThread() NSAssert(0 != pthread_main_np(), nil, @"This method must be called on the main thread");
```
Adlai-Holler: No need for the `nil`

maicki: Addressed

PR #2335, opened 2016-10-03
[ASCellNode] Restore Layout Attributes Fix + iOS 10 Modification

Adlai-Holler:
#2321 was reverted because its test was failing under iOS 10. iOS 10 seems to use a much larger reuse queue for collection cells, which revealed an issue with the previous implementation of this feature.

This restores #2321, plus:
- We used to clear the layout attributes only in `prepareForReuse` but now we clear it in `didEndDisplayingCell:`. 

### DIFF
--- a/AsyncDisplayKit/ASCellNode+Internal.h
+++ b/AsyncDisplayKit/ASCellNode+Internal.h
@@ -12,6 +12,8 @@
 
 #import "ASCellNode.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 @protocol ASCellNodeInteractionDelegate <NSObject>
 
 /**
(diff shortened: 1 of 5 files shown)


AsyncDisplayKitTests/ASCollectionViewTests.mm
```diff
@@ -86,6 +92,16 @@ - (ASCellNodeBlock)collectionView:(ASCollectionView *)collectionView nodeBlockFo
   };
 }
 
+- (void)collectionView:(ASCollectionView *)collectionView willDisplayNode:(ASCellNode *)node forItemAtIndexPath:(NSIndexPath *)indexPath
+{
+  ASDisplayNodeAssertNotNil(node.layoutAttributes, @"Expected layout attributes for node in %@ to be non-nil.", NSStringFromSelector(_cmd));
+}
+
+- (void)collectionView:(ASCollectionView *)collectionView didEndDisplayingNode:(ASCellNode *)node forItemAtIndexPath:(NSIndexPath *)indexPath
```
Adlai-Holler: Nah because these are just test implementations.

AsyncDisplayKit/ASCellNode+Internal.h
```diff
@@ -12,6 +12,8 @@
 
 #import "ASCellNode.h"
 
+NS_ASSUME_NONNULL_BEGIN
```
Adlai-Holler: `weak` always implies `nullable` so we're good

garrettmoon: Oh! Learned something new :)

PR #2337, opened 2016-10-03
[ASCollectionView] Allow User to Submit Updates Before Initial Data Load

Adlai-Holler:
There is an issue with UICollectionView (tested and documented in this diff) where issuing an update before the initial data load will cause an exception to be thrown.

Since #2274 , ASCollectionView has had the same issue. Since it's so infuriating, and since it's relatively easy for us to workaround on the user's behalf, this diff restores the old behavior of ignoring updates before initial data load.

### DIFF
--- a/AsyncDisplayKit/Details/ASChangeSetDataController.mm
+++ b/AsyncDisplayKit/Details/ASChangeSetDataController.mm
@@ -48,7 +48,13 @@
   [_changeSet addCompletionHandler:completion];
   if (_changeSetBatchUpdateCounter == 0) {
     void (^batchCompletion)(BOOL finished) = _changeSet.completionHandler;
-
+    
+    /**
+     * If the initial reloadData has not been called, just bail because we don't have
+     * our old data source counts.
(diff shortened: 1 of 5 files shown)


maicki:
LGTM

AsyncDisplayKit/Details/ASChangeSetDataController.mm
```diff
@@ -58,6 +56,9 @@ - (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
       _changeSet = nil;
       return;
     }
+
+    [self invalidateDataSourceItemCounts];
```
garrettmoon: Can you add a comment about this being a workaround?

garrettmoon: I guess the test is effective at ensuring it stays this way… so maybe it's not necessary.

Adlai-Holler:
Good call @garrettmoon!

PR #2338, opened 2016-10-03
[ASCollectionView] Handle nil Node Blocks Gracefully in Production

Adlai-Holler:


garrettmoon:
Should it assert if the block is nil?

Adlai-Holler:
@garrettmoon It does 🙂

garrettmoon:
Would you look at that, it sure does. LGTM!

PR #2340, opened 2016-10-04
[ASDisplayNode] Change _pendingDisplayNodes from NSMutableSet to ASWeakSet

appleguy:
In at least some cases, _pendingDisplayNodes referencing self can cause the
object to live longer than the application needs, e.g. while tearing down views.

It also appears there may be a cancellation condition where this state is
not balanced properly, which can cause a reference-cycle leak.  This change
safely fixes this problem when tested against a repro case for it.

I'm filing a followup task to make sure all cancellation edge cases have
proper accounting for this codepath, even though the leak is fixed.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2235,7 +2235,7 @@ static NSInteger incrementIfFound(NSInteger i) {
   ASDisplayNodeAssertMainThread();
 
   if (!_pendingDisplayNodes) {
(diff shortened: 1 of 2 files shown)


appleguy:
Before & After

screen shot 2016-10-03 at 10 22 48 pm
screen shot 2016-10-03 at 10 06 20 pm

Adlai-Holler:
Good catch!

garrettmoon:
Man, that new memory tool is awesome.

PR #2342, opened 2016-10-04
[Examples] Fix extra examples

maicki:
Fixes extra examples


maicki:
@Adlai-Holler Can you please review and merge. Fixes all of our example projects with the API changes as well as fixes the carthage build we broke while exposing `NSArray+Diffing.h`

PR #2182, opened 2016-09-02
[ASTableView] Go Straight to Super for Requerying Heights

Adlai-Holler:
We don't need to involve the data controller, range controller, or data source here. We just want UITableView to ask us for our row heights again.

PR #2343, opened 2016-10-04
[ASCollectionView] Only Remeasure All Nodes if Bounds Changed in Fixed Direction

Adlai-Holler:
See comments for details – doing this empty-update-remeasure-all-nodes is both extremely expensive and also somewhat unsafe if the data source has unreported changes. This drastically reduces the instances of it, since many collection views flow vertically and change bounds height due to status bar/nav bar/tab bar changes.

Fixes an issue in Pinterest with low risk. A permanent fix for the data unsafety issue is in #2182 but it caused other issues and I haven't had time to circle back on it yet.

Ready for review with an aim to land into our p6.12 branch.


maicki:
LGTM

PR #2344, opened 2016-10-05
[ASDisplayNode] Add -displayWillStartAsynchronously: method to allow skipping synchronous image cache check.

appleguy:
This is an optimization for displays where synchronous image caching is not needed, e.g. standard display-ahead in a range controller.

The first highlighted line here is the relevant one:

pasted_image_at_2016_10_04_03_28_pm_1024


appleguy:
Please merge this upon review - it's pretty safe and I don't have any further changes planned.

PR #2346, opened 2016-10-05
Implement ivar on _ASDisplayLayer to speed up getting node from layer.

garrettmoon:
Good ideas from @appleguy and @Adlai-Holler turned into sweet, sweet code.


appleguy:
Wow, astoundingly concise optimization!  Thanks @garrettmoon!

PR #2348, opened 2016-10-05
[ASCollectionView] Separate indexPath Spaces More Clearly

Adlai-Holler:
Users are getting confused when we tell them "the index paths are out-of-sync" or "don't line up."

To help tackle this major problem, this diff:
- Add failing unit tests for the new behavior.
- Makes `nodeAtIndexPath` and `indexPathForNode` work in the _data source's_ index space, not in the UIKit index space. So immediately after submitting an update, if you call `nodeAtIndexPath:` you'll get the new node.
- Deprecate `calculatedSizeForNodeAtIndexPath`: we should let them get the node and call `calculatedSize` instead – now that the index space distinction is more clear we should centralize all our logic on nodes instead.
- Removes the `indexPath` argument from `collectionView:willDisplayNode:forItemAtIndexPath:` and friends.
   - The `indexPath` was provided by the collection view and would not be correct for accessing their data source.
   - They can always call `indexPathForNode:` to get the data-source index path for the node, if needed. If the item for the node has been deleted from the collection view i.e. no longer exists in the data source, they'll get nil back. It's a bummer, but c'est la vie. The delegate will probably still want to know that a given node is being displayed, even if it no longer corresponds to an item.
- Add some assertions about what happens on main vs on editing queue. These are all pretty safe I believe, and are mostly there for documentation/regression-proofing.
- Add supplementary nodes into our collection view tests.

**NOTE**: I have added a unit test for similar behavior to be applied to supplementary nodes. This diff is already big enough just handling it for rows/items, so the unit test is disabled and I will enable it + the feature in a follow-up diff.

Ready for review @nguyenhuy @appleguy @maicki @garrettmoon !


Adlai-Holler:
Hmmm I need to adjust this to publicly support UIKit index paths for the sake of the UICollectionViewLayout. Kicking back to `development in progress`.

Adlai-Holler:
@appleguy @maicki @garrettmoon Well, the plan to completely hide UIKit index paths is a no-go  because `UICollectionViewLayout` subclasses will want to get nodes in their index space, which is the UIKit index space.

So if we have to expose them, we should come up with a public-facing name for the "completed" set or the "UIKit" set. The current implementation uses the word `Completed` which is pretty to look at but somewhat vague. We should obviously beef up the documentation about this, but simultaneously come up with good consistent words for these two spaces.

Ideas?

AsyncDisplayKit/ASCollectionView.h
```diff
@@ -358,6 +360,19 @@ NS_ASSUME_NONNULL_BEGIN
...
 @end
 
+@interface ASCollectionView (Deprecated)
+
+/**
+ * Query the sized node at `indexPath` for its calculatedSize.
+ *
+ * @param indexPath The index path for the node of interest.
+ *
+ * @deprecated Call @c calculatedSize on the node of interest instead.
```
maicki: Can you add the version number we deprecated the method

garrettmoon: Ooooh! Good call! We should start doing this.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1003,6 +1023,7 @@ - (ASCellNode *)dataController:(ASCollectionDataController *)dataController supp
   return node;
 }
 
+// TODO: Lock this
```
garrettmoon: This looks like it should be to-done before we land.

AsyncDisplayKit/ASTableView.h
```diff
@@ -272,20 +272,30 @@ NS_ASSUME_NONNULL_BEGIN
...
  * 
  * @param indexPath The index path of the requested node.
+ */
+- (nullable ASCellNode *)nodeForRowAtIndexPath:(NSIndexPath *)indexPath;
+
+/**
+ * Retrieves the node for the row at the given index path, in UITableView's index space.
  *
- * @return a node for display at this indexpath.
+ * @param indexPath The index path of the requested node.
```
garrettmoon: Should this have an `@return`

So this is *not* in the datasource index space? Maybe that should be mentioned?

AsyncDisplayKit/ASTableView.h
```diff
@@ -272,20 +272,30 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
-- (ASCellNode *)nodeForRowAtIndexPath:(NSIndexPath *)indexPath;
+- (nullable ASCellNode *)nodeForRowAtCompletedIndexPath:(NSIndexPath *)indexPath;
```
garrettmoon: `nodeForRowAtAsynchronousIndexPath` ? Not any clearer without documentation but…

AsyncDisplayKit/ASCollectionView.mm
Adlai-Holler: Nah, this was always wrong but I noticed it while building this

Adlai-Holler:
OK I've updated the implementation, going with a much cleaner API. The `nodeAtCompletedIndexPath` language is still used internally, in the data controller. The term `completed` is understood in that context so I'm just going to leave it for now.

appleguy:
@Adlai-Holler here's the Jenkins failure:

10:38:16 ▸ Compiling ASCollectionView+Internal.m
10:38:16 
10:38:16 ❌  /Users/engineering/jenkins_root/workspace/AsyncDisplayKit-PRs/AsyncDisplayKit/Private/ASCollectionView+Internal.m:9:9: 'ASCollectionView+Internal.h' file not found
10:38:16 
10:38:16 #import "ASCollectionView+Internal.h"
10:38:16         ^
10:38:16 
10:38:16 
10:38:16 ** BUILD FAILED **
10:38:16 
10:38:16 
10:38:16 The following build commands failed:
10:38:16 	CompileC /Users/engineering/Build/Intermediates/Pods.build/Debug-iphonesimulator/AsyncDisplayKit.build/Objects-normal/x86_64/ASCollectionView+Internal.o /Users/engineering/jenkins_root/workspace/AsyncDisplayKit-PRs/AsyncDisplayKit/Private/ASCollectionView+Internal.m normal x86_64 objective-c com.apple.compilers.llvm.clang.1_0.compiler
10:38:16 (1 failure)

Adlai-Holler:
Thanks @appleguy! The scope of this has changed radically and it's under re-development. Closing and will be superseded.

PR #2349, opened 2016-10-06
[Layout] Treat flex factors as true factors

maicki:
ASStackUnpositionedLayout was not actually treating flex factors as "true" factors. This meant that `flex-grow` and `flex-shrink` could only be expressed in one way: as whole numbers used to determine how much of the violation to assign to each flexible child.


maicki:
@Adlai-Holler @appleguy @levi Ready for review

AsyncDisplayKitTests/ASStackLayoutSpecSnapshotTests.mm
```diff
@@ -21,6 +21,15 @@ @interface ASStackLayoutSpecSnapshotTests : ASLayoutSpecSnapshotTestCase
 
 @implementation ASStackLayoutSpecSnapshotTests
 
+#pragma mark - XCTestCase
+
+- (void)setUp
+{
+  [super setUp];
+  
+  self.recordMode = NO;
```
garrettmoon: Is this method necessary? recordMode = NO is default, yeah?

AsyncDisplayKit/Private/ASStackUnpositionedLayout.mm
```diff
@@ -363,17 +363,23 @@ static void flexChildrenAlongStackDimension(std::vector<ASStackUnpositionedItem>
...
-  std::function<CGFloat(const ASStackUnpositionedItem &, BOOL)> flexAdjustment = flexAdjustmentInViolationDirection(items,
-                                                                                                                    style,
-                                                                                                                    violation,
-                                                                                                                    flexFactorSum);
+  std::function<CGFloat(const ASStackUnpositionedItem &)> flexAdjustment = flexAdjustmentInViolationDirection(items,
+                                                                                                              style,
+                                                                                                              violation,
+                                                                                                              flexFactorSum);
+
+  // Compute any remaining violation. Any remaining violation it will be allocated to the first flexible child.
```
garrettmoon: nit: remove 'it' in "Any remaining violation it will be allocated…"

AsyncDisplayKit/Private/ASStackUnpositionedLayout.mm
```diff
@@ -363,17 +363,23 @@ static void flexChildrenAlongStackDimension(std::vector<ASStackUnpositionedItem>
...
+  const CGFloat remainingViolation = std::accumulate(items.begin(), items.end(), violation, [&](CGFloat x, const ASStackUnpositionedItem &item) {
+    return x - flexAdjustment(item);
+  });
   BOOL isFirstFlex = YES;
```
garrettmoon: is `isFirstFlex` always YES or am I missing something? If so can we just factor it out or at least make it a const?

AsyncDisplayKit/Private/ASStackUnpositionedLayout.mm
```diff
@@ -174,17 +178,13 @@ static inline CGFloat scaledFlexShrinkFactor(const ASStackUnpositionedItem &item
...
 {
-  const CGFloat violationPerFlexFactor = floorf(violation / flexFactorSum);
-  const CGFloat remainingViolation = violation - (violationPerFlexFactor * flexFactorSum);
   // To compute the flex grow adjustment distribute the violation proportionally based on each item's flex grow factor.
-  // If there happens to be a violation remaining make sure it is allocated to the first flexible child.
-  return [violationPerFlexFactor, remainingViolation](const ASStackUnpositionedItem &item, BOOL isFirstFlex) {
-    // Only apply the remaining violation for the first flexible child that has a flex grow factor.
-    return violationPerFlexFactor * item.child.style.flexGrow + (isFirstFlex && item.child.style.flexGrow > 0 ? remainingViolation : 0);
+  return [violation, flexFactorSum](const ASStackUnpositionedItem &item) {
+    return floorf(violation * (item.child.style.flexGrow / flexFactorSum));
```
Adlai-Holler: Use `std::floor` and friends throughout, so that we get type-generic math on different platforms.

appleguy:
test this please

AsyncDisplayKitTests/ASStackLayoutSpecSnapshotTests.mm
maicki: Yes we finally should add this. By default `recordMode` is `NO` as we don't want to create any snapshot images. Setting it to `YES` will trigger to create the snapshot images.

maicki:
@Adlai-Holler @garrettmoon addressed comments

PR #2351, opened 2016-10-06
[Layout] Improve ASDimensionMake via NSString

maicki:
- Add ASXCTAssertEqualDimensions
- Add support for "auto"
- Assert if given an empty or invalid string


maicki:
@Adlai-Holler @appleguy @garrettmoon Ready for review

AsyncDisplayKit/Layout/ASDimension.mm
```diff
@@ -17,32 +17,35 @@
...
+    
+    // Handle points
+    NSUInteger pointsStringLocation = [dimension rangeOfString:@"pt"].location;
+    if (pointsStringLocation != NSNotFound) {
+      // Check if points is at the end and remove it
+      if (pointsStringLocation == (dimension.length-2)) {
+        dimension = [dimension substringToIndex:(dimension.length-2)];
+        return ASDimensionMake(ASDimensionUnitPoints, dimension.floatValue);
+      }
+    }
```
Adlai-Holler: Style: Use `hasSuffix:@"pt"]

Performance: remove `substringToIndex` – according to the docs, `floatValue` only checks if the string starts with a number.

Same below.

maicki:
@Adlai-Holler Comments addressed

AsyncDisplayKit/Layout/ASDimension.mm
```diff
@@ -17,32 +17,35 @@
...
-  NSUInteger pointsStringLocation = [dimension rangeOfString:@"pt"].location;
-  if (pointsStringLocation != NSNotFound) {
-    // Check if points is at the end and remove it
-    if (pointsStringLocation == (dimension.length-2)) {
-      dimension = [dimension substringToIndex:(dimension.length-2)];
-      return ASDimensionMake(ASDimensionUnitPoints, dimension.floatValue);
+  if (dimension.length != 0) {
+    
+    // Handle points
+    NSUInteger pointsStringLocation = [dimension rangeOfString:@"pt"].location;
```
appleguy: One tricky way to perform this check would be characterAtIndex: length - 1, and then check if it == 't' or == '%' — and if it is neither, then return Auto?  Probably too obtuse, just having fun :)

maicki: We don't do that check anymore, using `hasSuffix` now :)
👍 for fun though

AsyncDisplayKit/Layout/ASDimension.mm
```diff
@@ -10,39 +10,32 @@
...
-      return ASDimensionMake(ASDimensionUnitPoints, dimension.floatValue);
+  if (dimension.length > 0) {
+    
+    // Handle points
+    if ([dimension hasSuffix:@"pt"]) {
+      return ASDimensionMake(ASDimensionUnitPoints, ASCGFloatFromString(dimension));
+    }
+    
+    // Handle auto
+    if ([dimension isEqualToString:@"auto"]) {
```
appleguy: @maicki very minor change, but I would suggest:

- Make auto the last one to check.  It will be the rarest to use since there is an easy constant value, ASDimensionAuto, that feels cleaner for clients to use.
- Check percent first, probably the most common reason to use this?  pt vs percent first doesn't make much difference.

Then, this will be a bit more optimized for anyone who uses it heavily.

AsyncDisplayKit/Layout/ASDimension.mm
```diff
@@ -66,12 +59,12 @@ @implementation NSNumber (ASDimension)
 
 - (ASDimension)as_pointDimension
 {
-  return ASDimensionMake(ASDimensionUnitPoints, self.floatValue);
+  return ASDimensionMake(ASDimensionUnitPoints, ASCGFloatFromNumber(self));
 }
 
 - (ASDimension)as_fractionDimension
```
appleguy: as_dimensionAsFraction?  Hmm this is a strange naming case, not sure it feels right currently...

AsyncDisplayKit/Layout/ASDimension.mm
maicki: Adressed

PR #2353, opened 2016-10-06
[Umbrella] Add warn_unused_result Attribute to Public Headers

Adlai-Holler:
This attribute triggers compiler warnings/errors (in ASDK it's errors) when calling a function or method and not using its result.

This diff also makes `scrollableDirections` and `scrollDirection` properties for a slightly cleaner Swift import. I'm willing to revert this if someone thinks it's bad.


maicki:
LGTM

PR #2366, opened 2016-10-10
[Layout] Add ASLayoutElementStylability protocol

maicki:
Is adding convenience methods to form layout specifications like:

```objc
 [ASInsetLayoutSpec
  insetLayoutSpecWithInsets:UIEdgeInsetsMake(10, 10, 10, 10)
  child:[ASStackLayoutSpec
         stackLayoutSpecWithDirection:ASStackLayoutDirectionVertical
         spacing:10
         justifyContent:ASStackLayoutJustifyContentStart
         alignItems:ASStackLayoutAlignItemsStretch
         children:@[
          [titleTextNode styleWithBlock:^(ASLayoutElementStyle *style) {
            style.width = ASDimensionMake(100);
            style.flexGrow = 1.0;
          }],
          [messageTextNode styleWithBlock:^(ASLayoutElementStyle *style) {
            style.flexGrow = 1.0;
            style.flexShrink = 1.0;
          }]]]];
```


AsyncDisplayKit/Layout/ASLayoutElement.h
```diff
@@ -363,4 +364,10 @@ extern NSString * const ASLayoutElementStyleLayoutPositionProperty;
 
 @end
 
+@protocol ASLayoutElementStylability
+
+- (id<ASLayoutElement>)styleWithBlock:(void (^)(__kindof ASLayoutElementStyle *style))styleBlock;
```
Adlai-Holler: I think the return type here can be `instancetype`

Adlai-Holler: What do you think of the name `styledWithBlock`? That emphasizes the fact that the method has a return value.

Adlai-Holler: Let's add the fancy `__attribute__((noescape))` before the block type.

maicki:
@Adlai-Holler Comments addressed. In fact what do you think about `ASLayoutElementStylability` protocol name at all? I don't really know if this is a great name, but couldn't find a better name for  it yet ...

Adlai-Holler:
@maicki Looks great! To the protocol name, it's not a great name but we can always change it (or even remove it) later with no great penalty so let's move forward with it for now.

PR #2370, opened 2016-10-12
[ASLog] Add new ASProfilingSignpost defines for System Trace flags.

appleguy:
screen shot 2016-10-11 at 9 54 13 pm

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -21,6 +21,8 @@
 #import "ASDataController+Subclasses.h"
 #import "ASDispatch.h"
 
+#import <sys/kdebug_signpost.h>
+
 //#define LOG(...) NSLog(__VA_ARGS__)
 #define LOG(...)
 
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -136,6 +138,8 @@ - (void)batchLayoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts
     [ASDataController _expectToInsertNodes:contexts.count];
 #endif
 
+  ASProfilingSignpostStart(2);
```
Adlai-Holler: What's this 2 mean? Should this be some constant?

appleguy:
Added some more.  @Adlai-Holler the constant values should eventually become defines, but I don't yet have a naming scheme or strategy for assigning them that calls for this yet.  For now, we have these, each with its own color (maxing out the available colors):

0 - ASRunloopQueue
1 - ASRangeController
2 - ASDataController
3 - AsyncDisplay

screen shot 2016-10-12 at 7 56 57 pm

PR #2371, opened 2016-10-12
[Layout] Add locking for ASStackLayoutElement and ASAbsoluteLayoutElement

maicki:


AsyncDisplayKit/Layout/ASLayoutElement.h
```diff
@@ -302,66 +302,6 @@ extern NSString * const ASLayoutElementStyleLayoutPositionProperty;
  */
 @property (nonatomic, assign, readwrite) ASLayoutSize maxLayoutSize;
 
-
-#pragma mark - ASStackLayoutElement
-
-/**
- * @abstract Additional space to place before this object in the stacking direction.
```
garrettmoon: Are we losing all the documentation then?

maicki: No as the documentation is in `ASStackLayoutElement` protocol

garrettmoon:
Great find! LGTM.

PR #2372, opened 2016-10-12
[Umbrella] ASCollectionView -> ASCollectionNode Migration, Separate Index Spaces

Adlai-Holler:
I've tested this thoroughly with Pinterest using the deprecated API.

Supersedes #2348 .

Methods involving the `ASCollectionNode` always use data source index space. Methods on the view always use UIKit index space. I may remove `nodeForRowAtIndexPath` and `indexPathForNode` from `ASTableView` to avoid confusion but I'm still thinking about possibly legitimate use cases for them such as interactive reordering and `scrollToRowAtIndexPath:` or after they call `indexPathForCell:` for some reason, which they shouldn't do but we can't stop them.

### DIFF
--- a/AsyncDisplayKit.xcodeproj/project.pbxproj
+++ b/AsyncDisplayKit.xcodeproj/project.pbxproj
@@ -440,6 +440,7 @@
 		CC7FD9DF1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.m in Sources */ = {isa = PBXBuildFile; fileRef = CC7FD9DD1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.m */; };
 		CC7FD9E11BB5F750005CCB2B /* ASPhotosFrameworkImageRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CC7FD9E01BB5F750005CCB2B /* ASPhotosFrameworkImageRequestTests.m */; };
 		CC7FD9E21BB603FF005CCB2B /* ASPhotosFrameworkImageRequest.h in Headers */ = {isa = PBXBuildFile; fileRef = CC7FD9DC1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		CC87BB951DA8193C0090E380 /* ASCellNode+Internal.h in Headers */ = {isa = PBXBuildFile; fileRef = CC87BB941DA8193C0090E380 /* ASCellNode+Internal.h */; };
 		CC88F7AE1D80AF5E000D6D4E /* ASObjectDescriptionHelpers.h in CopyFiles */ = {isa = PBXBuildFile; fileRef = CC446A2D1D80AAE00071FD03 /* ASObjectDescriptionHelpers.h */; };
 		CC8B05D61D73836400F54286 /* ASPerformanceTestContext.m in Sources */ = {isa = PBXBuildFile; fileRef = CC8B05D51D73836400F54286 /* ASPerformanceTestContext.m */; };
 		CC8B05D81D73979700F54286 /* ASTextNodePerformanceTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CC8B05D71D73979700F54286 /* ASTextNodePerformanceTests.m */; };
@@ -1106,6 +1107,7 @@
 		BDC2D162BD55A807C1475DA5 /* Pods-AsyncDisplayKitTests.profile.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-AsyncDisplayKitTests.profile.xcconfig"; path = "Pods/Target Support Files/Pods-AsyncDisplayKitTests/Pods-AsyncDisplayKitTests.profile.xcconfig"; sourceTree = "<group>"; };
 		CC051F1E1D7A286A006434CB /* ASCALayerTests.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = ASCALayerTests.m; sourceTree = "<group>"; };
 		CC0AEEA31D66316E005D1C78 /* ASUICollectionViewTests.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = ASUICollectionViewTests.m; sourceTree = "<group>"; };
+		CC2E317F1DAC353700EEE891 /* ASCollectionView+Undeprecated.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = "ASCollectionView+Undeprecated.h"; sourceTree = "<group>"; };
 		CC3B20811C3F76D600798563 /* ASPendingStateController.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ASPendingStateController.h; sourceTree = "<group>"; };
 		CC3B20821C3F76D600798563 /* ASPendingStateController.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; path = ASPendingStateController.mm; sourceTree = "<group>"; };
 		CC3B20871C3F7A5400798563 /* ASWeakSet.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ASWeakSet.h; sourceTree = "<group>"; };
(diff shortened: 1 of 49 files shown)


AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -440,6 +440,7 @@
 		CC7FD9DF1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.m in Sources */ = {isa = PBXBuildFile; fileRef = CC7FD9DD1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.m */; };
 		CC7FD9E11BB5F750005CCB2B /* ASPhotosFrameworkImageRequestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CC7FD9E01BB5F750005CCB2B /* ASPhotosFrameworkImageRequestTests.m */; };
 		CC7FD9E21BB603FF005CCB2B /* ASPhotosFrameworkImageRequest.h in Headers */ = {isa = PBXBuildFile; fileRef = CC7FD9DC1BB5E962005CCB2B /* ASPhotosFrameworkImageRequest.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		CC87BB951DA8193C0090E380 /* ASCellNode+Internal.h in Headers */ = {isa = PBXBuildFile; fileRef = CC87BB941DA8193C0090E380 /* ASCellNode+Internal.h */; };
```
Adlai-Holler: This file existed before, but wasn't part of the project for some reason.

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -298,8 +300,16 @@ - (void)handleVisibilityChange:(BOOL)isVisible
   
   ASDisplayNode *owningNode = scrollView.asyncdisplaykit_node;
   if ([owningNode isKindOfClass:[ASCollectionNode class]]) {
+    NSIndexPath *ip = [(ASCollectionNode *)owningNode indexPathForNode:self];
+    if (ip != nil) {
+      [result addObject:@{ @"indexPath" : ip }];
+    }
```
Adlai-Holler: This is sort of unrelated and should have been here before. Really helps logging.

AsyncDisplayKit/ASCollectionNode.h
```diff
@@ -131,20 +241,346 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *   then @c waitUntilAllUpdatesAreCommitted instead.
+ */
+- (void)reloadDataImmediately ASDISPLAYNODE_DEPRECATED;
+
+@end
+
+/**
+ * This is a node-based UICollectionViewDataSource.
+ */
+@protocol ASCollectionDataSource <ASCommonCollectionDataSource>
```
Adlai-Holler: This (and `ASCollectionDelegate` `ASTableDataSource` and `ASTableDelegate`) was kept as-is, but moved from the `View.h` file to the `Node.h` file.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -488,7 +540,18 @@ - (CGSize)calculatedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
 
 - (ASCellNode *)nodeForItemAtIndexPath:(NSIndexPath *)indexPath
 {
-  return [_dataController nodeAtIndexPath:indexPath];
+  return [_dataController nodeAtCompletedIndexPath:indexPath];
+}
+
+- (NSIndexPath *)convertIndexPathFromCollectionNode:(NSIndexPath *)indexPath waitingIfNeeded:(BOOL)wait
```
Adlai-Holler: This is private, and only here for backwards-compatibility reasons. Currently the `scrollToItemAtIndexPath` methods on `ASTableView` and `ASCollectionView` take index paths in the data source space, and so for the time being we use this as a shim to maintain that behavior. In the future, after some testing, we'll remove our override on that method, and remove this with it.

AsyncDisplayKit/ASCollectionNode.h
```diff
@@ -131,20 +241,346 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * @return A constrained size range for layout for the item at this index path.
+ */
+- (ASSizeRange)collectionNode:(ASCollectionNode *)collectionNode constrainedSizeForItemAtIndexPath:(NSIndexPath *)indexPath;
+
+- (void)collectionNode:(ASCollectionNode *)collectionNode willDisplayItemWithNode:(ASCellNode *)node;
+
+- (void)collectionNode:(ASCollectionNode *)collectionNode didEndDisplayingItemWithNode:(ASCellNode *)node;
+
+- (BOOL)collectionNode:(ASCollectionNode *)collectionNode shouldHighlightItemWithNode:(ASCellNode *)node;
+- (void)collectionNode:(ASCollectionNode *)collectionNode didHighlightItemWithNode:(ASCellNode *)node;
```
Adlai-Holler: This was a tough question. I could have gone with `didSelectItemAtIndexPath` etc but since the item/section may have been deleted or reloaded, the argument would have to be `nullable` which would be really wacky, giving users no information at all. I could filter out cases like that, but there could be legitimate arguments for behaviors like "whenever a header node of X class is showing, the background color of my view should be Y" so filtering out these callbacks would break that. Or if the user happened to select an item just before a section is reloaded, the select would get ignored.

Adlai-Holler: Well, I've reversed stance on this. In order to keep things simple, all delegate callbacks except `willDisplayRow` and `didEndDisplayingRow` will pass a `nonnull NSIndexPath *`. In the extremely unlikely case that the user selected an item that is pending deletion/reload, the event won't get sent.

AsyncDisplayKit/ASCollectionNode.h
```diff
@@ -122,6 +122,116 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *  Perform a batch of updates asynchronously, optionally disabling all animations in the batch. This method must be called from the main thread.
+ *  The data source must be updated to reflect the changes before the update block completes.
+ *
+ *  @param animated   NO to disable animations for this batch
+ *  @param updates    The block that performs the relevant insert, delete, reload, or move operations.
+ *  @param completion A completion handler block to execute when all of the operations are finished. This block takes a single
+ *                    Boolean parameter that contains the value YES if all of the related animations completed successfully or
+ *                    NO if they were interrupted. This parameter may be nil. If supplied, the block is run on the main thread.
+ */
+- (void)performBatchAnimated:(BOOL)animated updates:(nullable __attribute((noescape)) void (^)())updates completion:(nullable void (^)(BOOL finished))completion;
```
garrettmoon: Can we add back in performBatchUpdates without the animated flag?

AsyncDisplayKit/ASCollectionNode.h
```diff
@@ -131,20 +241,346 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * This is a node-based UICollectionViewDataSource.
+ */
+@protocol ASCollectionDataSource <ASCommonCollectionDataSource>
+
+@optional
+
+/**
+ * Asks the data source for the number of items in the given section of the collection node.
+ *
+ * @see @c numberOfSectionsInCollectionView:
```
garrettmoon: update

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -209,33 +238,19 @@ + (Class)layerClass
 
 - (instancetype)initWithCollectionViewLayout:(UICollectionViewLayout *)layout
 {
-  return [self _initWithFrame:CGRectZero collectionViewLayout:layout ownedByNode:NO];
+  return [self _initWithFrame:CGRectZero collectionViewLayout:layout layoutFacilitator:nil];
```
garrettmoon: Super nit: can this call through into initWithFrame:

hannahmbanana:
@Adlai-Holler: Want me to put up a separate PR for these example projects needing updates?

- ASCollectionView
- ASDKgram
- ASDKTube
- CatDealsCollectionView
- CustomCollectionView
- HorizontalWithinVerticalScrolling

Adlai-Holler:
@hannahmbanana Sure – feel free to push them onto this branch.

hannahmbanana:
@Adlai-Holler: could you review the example file updates?

examples/ASDKgram/Sample/PhotoFeedNodeController.m
```diff
@@ -192,7 +192,7 @@ - (void)tableView:(ASTableView *)tableView willBeginBatchFetchWithContext:(ASBat
 - (void)resetAllData
 {
   [_photoFeed clearFeed];
```
Adlai-Holler: @hannahmbanana Change `tableView:willBeginBatchFetchWithContext:` above to `tableNode:willBeginBatchFetchWithContext:`. Is that not showing a deprecation warning? Bummer.

Adlai-Holler: Also change the other ASTableDelegate methods to `tableNode:numberOfRowsInSection:` etc.

examples/ASCollectionView/Sample/ViewController.m
```diff
@@ -92,6 +96,8 @@ - (void)viewDidLoad
 
 - (void)reloadTapped
 {
+  // This method is deprecated because we reccommend using ASCollectionNode instead of ASCollectionView.
+  // This functionality & example project remains for users who insist on using ASCollectionView.
```
Adlai-Holler: @hannahmbanana Since `ASCollectionView` outside of the node isn't going to be supported for long, this sample should either be changed majorly or discarded entirely.

AsyncDisplayKit/ASCollectionNode.h
Adlai-Holler: ✅

AsyncDisplayKit/ASCollectionNode.h
Adlai-Holler: ✅

hannahmbanana:
@Adlai-Holler: I think there's an issue with your new code not `nodeForRowAtIndexPath:`
 
`ASTableView.mm` (line 1389) should be this:
 `} else if (_asyncDataSourceFlags.tableNodeNodeForRow) {`

This fixes one of the sample projects, but not completely.

Adlai-Holler:
Here we go!

PR #2373, opened 2016-10-12
This is super minor, but active processor account takes into processo…

garrettmoon:
…rs that may be out of commission due to boot arguments or thermal conditions.


maicki:
LGTM

PR #2376, opened 2016-10-13
[ASTextNode] Go Back to Having Renderer as Ivar

Adlai-Holler:
In 6.11 we introduced a text node renderer cache, and removed the `renderer` ivar from text node. Pinterest started getting a lot of text node related crashes.

In 6.12 we disabled the cache – not a revert, just removed the cache itself – and we still get the same crashes, maybe even with higher frequency. In addition, this means creating a lot more renderers.

This diff is a clean revert back to the original renderer-ivar-invalidation behavior. We can fix forward from here.


Adlai-Holler:
Ready for review @appleguy @maicki

appleguy:
Thanks @Adlai-Holler!  This may re-introduce a very minor, rare edge case sizing bug - but fixing the performance and thread safety issues is far more important.

@hannahmbanana Let's get this merged into p6.13 and Pinterest updated as well.

PR #2378, opened 2016-10-13
[Layout] Return CGSizeZero in case a ASNetworkImageNode is asked for the size and the image did not load yet

maicki:
We are catching the behavior if a developer set an URL and asks for the size of the node but the image was not loaded yet.

This is not the right behavior in my opinion, but we should add it for the transition time. That's also the reason we only added it to `ASNetworkImageNode` as it's the most common case.

We also will flag that and in the future we should think about removing that check.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -333,6 +333,23 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
   }
 }
 
+#pragma mark - Layout and Sizing
+
+- (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
+{
+  ASDN::MutexLocker l(__instanceLock__);
+
+  // If the image node is currently in the loading process and no valid size is applied return CGSizeZero for the time
+  // being.
+  // TODO: After 2.0 is stable we should remove this behavior as a ASNetworkImageNode is a replaced element and the
+  // client code should set the size of an image or it's container it's embedded in
+  if (ASIsCGSizeValidForSize(constrainedSize) == NO && _URL != nil && self.image == nil) {
+    return CGSizeZero;
+  }
+    
+  return [super calculateSizeThatFits:constrainedSize];
+}
+
 #pragma mark - Private methods -- only call with lock.
 
 - (void)_updateProgressImageBlockOnDownloaderIfNeeded
--- a/AsyncDisplayKit/Layout/ASAbsoluteLayoutSpec.mm
+++ b/AsyncDisplayKit/Layout/ASAbsoluteLayoutSpec.mm
@@ -39,8 +39,8 @@
 {
   // TODO: layout: isValidForLayout() call should not be necessary if INFINITY is used
   CGSize size = {
-    (isinf(constrainedSize.max.width) || !ASPointsAreValidForLayout(constrainedSize.max.width)) ? ASLayoutElementParentDimensionUndefined : constrainedSize.max.width,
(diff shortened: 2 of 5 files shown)


garrettmoon:
I think this is definitely the right call for now!

AsyncDisplayKit/Layout/ASAbsoluteLayoutSpec.mm
```diff
@@ -39,8 +39,8 @@ - (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
 {
   // TODO: layout: isValidForLayout() call should not be necessary if INFINITY is used
   CGSize size = {
-    (isinf(constrainedSize.max.width) || !ASPointsAreValidForLayout(constrainedSize.max.width)) ? ASLayoutElementParentDimensionUndefined : constrainedSize.max.width,
-    (isinf(constrainedSize.max.height) || !ASPointsAreValidForLayout(constrainedSize.max.height)) ? ASLayoutElementParentDimensionUndefined : constrainedSize.max.height
+    (isinf(constrainedSize.max.width) || !ASPointsValidForLayout(constrainedSize.max.width)) ? ASLayoutElementParentDimensionUndefined : constrainedSize.max.width,
```
garrettmoon: 👍

maicki:
Will merge as Jenkins did finish and Travis didn't even start :/

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -333,6 +333,23 @@ - (void)fetchData
...
+
+- (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
+{
+  ASDN::MutexLocker l(__instanceLock__);
+
+  // If the image node is currently in the loading process and no valid size is applied return CGSizeZero for the time
+  // being.
+  // TODO: After 2.0 is stable we should remove this behavior as a ASNetworkImageNode is a replaced element and the
+  // client code should set the size of an image or it's container it's embedded in
+  if (ASIsCGSizeValidForSize(constrainedSize) == NO && _URL != nil && self.image == nil) {
```
appleguy: @maicki shouldn't it do this even if _URL == nil?

What about checking the .defaultImage?

maicki: `_URL == nil`: If `_URL` is nil in this case there is something wrong with the client code. The `_URL` should never be nil in this case, also if a `defaultImage` is set a `_URL` should be set.

`.defaultImage`: The `.defaultImage` size wasn't used in the size calculation in `1.x` so I think we should not introduce some new behavior I would like to avoid anyway.

In general: This overall solution should be there for a short period in the first place until the conversion pain points from 1.x to 2 are solved, because an `ASNetworkImageNode` always should have some kind of size set.

maicki: What about adding an assert if it goes into here and no `_URL` is set? I don't want to encourage developers to rely on this workaround we are putting in here.

cc @garrettmoon @appleguy

garrettmoon: Something like: `NSAssert(_URL != nil, @"An NSURL must be set on ASNetworkImageNode before layout unless a valid size is provided.");`. Take out the _URL check in the if statement and add the assert inside there?

maicki: @garrettmoon @appleguy Put up a PR that add's an assertion if the network image node calculates it's size and `_URL == nil` with invalid constrained size or no image set: #2389

PR #2381, opened 2016-10-14
[ASCollectionNode] Add scrollableDirections method to ASCollectionViewLayoutInspecting p…

garrettmoon:
…rotocol

ASViewController's default method will not always return the correct scrollable
directions in cases where size changes may not have been applied. This allows
us to ignore this issue for a little while longer by allowing the layout
inspector to explicitly vend this.

@Adlai-Holler, @maicki ready for review

### DIFF
--- a/AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
+++ b/AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
@@ -12,7 +12,7 @@
 
 #import <Foundation/Foundation.h>
 #import <AsyncDisplayKit/ASDimension.h>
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.h
```diff
@@ -12,6 +12,7 @@
 
 #import <Foundation/Foundation.h>
 #import <AsyncDisplayKit/ASDimension.h>
+#import "ASScrollDirection.h"
```
maicki: Should we import it as `<AsyncDisplayKit/ASScrollDirection.h>` as it's in a public header?

PR #2385, opened 2016-10-14
Changed relative spec position and sizing options to start at 1

lappp9:
Starting NS_ENUM definitions at 0 doesn't automatically convert to Swift 3 correctly.

Doing so will only reveal options 1..n in the autocomplete/auto-generated headers.
screen shot 2016-10-13 at 11 43 54 pm

So if you want to use ASRelativeLayoutSpecPositionStart or ASRelativeLayoutSpecSizingOptionDefault your layout spec definition will look something like this.

screen shot 2016-10-13 at 11 44 12 pm

### DIFF
--- a/AsyncDisplayKit/Layout/ASRelativeLayoutSpec.h
+++ b/AsyncDisplayKit/Layout/ASRelativeLayoutSpec.h
@@ -13,23 +13,25 @@
 #import <AsyncDisplayKit/ASLayoutSpec.h>
 
 /** How the child is positioned within the spec. */
-typedef NS_OPTIONS(NSUInteger, ASRelativeLayoutSpecPosition) {
+typedef NS_ENUM(NSUInteger, ASRelativeLayoutSpecPosition) {
     /** The child is positioned at point 0 relatively to the layout axis (ie left / top most) */
-    ASRelativeLayoutSpecPositionStart = 1 << 0,
+    ASRelativeLayoutSpecPositionStart = 0,
     /** The child is centered along the specified axis */
-    ASRelativeLayoutSpecPositionCenter = 1 << 1,
+    ASRelativeLayoutSpecPositionCenter = 1,
     /** The child is positioned at the maximum point of the layout axis (ie right / bottom most) */
-    ASRelativeLayoutSpecPositionEnd = 1 << 2,
+    ASRelativeLayoutSpecPositionEnd = 2,
 };
 
-/** How much space the spec will take up. */
+/** How much space the spec will take up. 
+ *  Note: To use the ASRelativeLayoutSpecSizingOptionDefault sizing option in Swift use []
+ */
 typedef NS_OPTIONS(NSUInteger, ASRelativeLayoutSpecSizingOption) {
     /** The spec will take up the maximum size possible */
-    ASRelativeLayoutSpecSizingOptionDefault = 1 << 0,
+    ASRelativeLayoutSpecSizingOptionDefault,
     /** The spec will take up the minimum size possible along the X axis */
-    ASRelativeLayoutSpecSizingOptionMinimumWidth = 1 << 1,
+    ASRelativeLayoutSpecSizingOptionMinimumWidth = 1 << 0,
     /** The spec will take up the minimum size possible along the Y axis */
-    ASRelativeLayoutSpecSizingOptionMinimumHeight = 1 << 2,
+    ASRelativeLayoutSpecSizingOptionMinimumHeight = 1 << 1,
     /** Convenience option to take up the minimum size along both the X and Y axis */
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Layout/ASRelativeLayoutSpec.h
```diff
@@ -15,21 +15,21 @@
...
 typedef NS_OPTIONS(NSUInteger, ASRelativeLayoutSpecPosition) {
     /** The child is positioned at point 0 relatively to the layout axis (ie left / top most) */
-    ASRelativeLayoutSpecPositionStart = 0,
+    ASRelativeLayoutSpecPositionStart = 1 << 0,
     /** The child is centered along the specified axis */
-    ASRelativeLayoutSpecPositionCenter = 1 << 0,
+    ASRelativeLayoutSpecPositionCenter = 1 << 1,
     /** The child is positioned at the maximum point of the layout axis (ie right / bottom most) */
-    ASRelativeLayoutSpecPositionEnd = 1 << 1,
+    ASRelativeLayoutSpecPositionEnd = 1 << 2,
```
Adlai-Holler: Wow! Since you can't specify multiple options here, this should actually be an `NS_ENUM`, with `=1`, `=2`, `=3` etc.

AsyncDisplayKit/Layout/ASRelativeLayoutSpec.h
```diff
@@ -15,21 +15,21 @@
...
 };
 
 /** How much space the spec will take up. */
 typedef NS_OPTIONS(NSUInteger, ASRelativeLayoutSpecSizingOption) {
     /** The spec will take up the maximum size possible */
-    ASRelativeLayoutSpecSizingOptionDefault,
+    ASRelativeLayoutSpecSizingOptionDefault = 1 << 0,
```
Adlai-Holler: Discussing this more, changing the default value here to 1 creates an undefined case for Swift users. In Swift, it should always be valid to provide `[]` for an `OptionSet`. If they do that here, then we have an undefined value.

I think the best thing we can do is to restore the original values in here, and add documentation that in Swift, pass `[]` to get `ASRelativeLayoutSpecSizingOptionDefault`. Apple has made exceptions in their own API to import 0-value options (e.g. `UIViewAutoResizing`) but we're not so lucky…

AsyncDisplayKit/Layout/ASRelativeLayoutSpec.h
lappp9: Yeah, I actually had that realization last night right before bed, lol

AsyncDisplayKit/Layout/ASRelativeLayoutSpec.h
lappp9: Ah, interesting/bummer, sounds good

lappp9: Maybe this is a bad idea, but couldn't this also be defined as an NS_ENUM  
```
typedef NS_ENUM(NSUInteger, ASRelativeLayoutSpecSizing) {
  ASRelativeLayoutSpecSizingDefault               = 0, 
  ASRelativeLayoutSpecSizingMinimumWidth  = 1,   
  ASRelativeLayoutSpecSizingMinimumHeight = 2, 
  ASRelativeLayoutSpecSizingMinimumSize     = 3,
};
```
Though I'm sure that would need some changes in ASRelativeSpec to make work.

lappp9:
@Adlai-Holler Re-posting this comment here in case it would have been missed in the "outdated changes" section:

Maybe this is a bad idea, but couldn't this also be defined as an NS_ENUM  
```
typedef NS_ENUM(NSUInteger, ASRelativeLayoutSpecSizing) {
  ASRelativeLayoutSpecSizingDefault               = 0, 
  ASRelativeLayoutSpecSizingMinimumWidth  = 1,   
  ASRelativeLayoutSpecSizingMinimumHeight = 2, 
  ASRelativeLayoutSpecSizingMinimumSize     = 3,
};
```
Though I'm sure that would need some changes in ASRelativeSpec to make work.

maicki:
@lappp9 Hey Luke, thanks for working on that. There are more substantial problems (in terms of layout logic as it was originally kind of copied from `ASCenterLayoutSpec`) with how `ASRelativeLayoutSpec` is implemented that I'm currently looking into and will follow up with something pretty soon.

cc @appleguy @Adlai-Holler

appleguy:
@maicki should we merge this before your diff?

maicki:
@appleguy No, this change will be included in my diff. Furthermore we are not able to merge it as it needs further adjustments in `ASRelativeLayoutSpec` itself as well as the snapshot tests.

maicki:
This PR has the change in there as well as some fixes that are necessary: #2393

PR #2386, opened 2016-10-14
[ASNetworkImageNode] Test and Improve the Progress Block Handling

Adlai-Holler:
I found the old logic hard to reason about, and I had no confidence that it behaved ideally. In fact, specifically when the download identifier changed the old download's progress block was not unbound which could have led to drawing the wrong image.

In practice that rarely matters, but this adds our first `ASNetworkImageNode` tests which we desperately need, fixes that issue, and makes the progress image binding logic much much more reliable.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -38,6 +38,8 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 
   NSUUID *_cacheUUID;
   id _downloadIdentifier;
+  // The download identifier that we have set a progress block on, if any.
+  id _downloadIdentifierForProgressBlock;
 
   BOOL _imageLoaded;
   CGFloat _currentImageQuality;
@@ -222,16 +224,12 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 
 - (void)setShouldRenderProgressImages:(BOOL)shouldRenderProgressImages
 {
-  __instanceLock__.lock();
+  ASDN::MutexLocker l(__instanceLock__);
   if (shouldRenderProgressImages == _shouldRenderProgressImages) {
-    __instanceLock__.unlock();
     return;
   }
-  
   _shouldRenderProgressImages = shouldRenderProgressImages;
-  
-  
-  __instanceLock__.unlock();
+
   [self _updateProgressImageBlockOnDownloaderIfNeeded];
 }
 
@@ -283,9 +281,9 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 - (void)didEnterVisibleState
 {
   [super didEnterVisibleState];
-  
+  ASDN::MutexLocker l(__instanceLock__);
+
   if (_downloaderFlags.downloaderImplementsSetPriority) {
-    ASDN::MutexLocker l(__instanceLock__);
     if (_downloadIdentifier != nil) {
       [_downloader setPriority:ASImageDownloaderPriorityVisible withDownloadIdentifier:_downloadIdentifier];
     }
@@ -297,9 +295,9 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 - (void)didExitVisibleState
 {
   [super didExitVisibleState];
-  
+  ASDN::MutexLocker l(__instanceLock__);
+
   if (_downloaderFlags.downloaderImplementsSetPriority) {
-    ASDN::MutexLocker l(__instanceLock__);
     if (_downloadIdentifier != nil) {
       [_downloader setPriority:ASImageDownloaderPriorityPreload withDownloadIdentifier:_downloadIdentifier];
     }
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
For review, your best bet is a split-diff. I inserted a method `handleProgressImage` just before the `_updateProgressImageBlock` method and GitHub's unified diff is confused.

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -350,41 +348,56 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
   return [super calculateSizeThatFits:constrainedSize];
 }
 
-#pragma mark - Private methods -- only call with lock.
+#pragma mark - Private methods, safe to call without lock
 
-- (void)_updateProgressImageBlockOnDownloaderIfNeeded
+- (void)handleProgressImage:(UIImage *)progressImage progress:(CGFloat)progress downloadIdentifier:(nullable id)downloadIdentifier
```
appleguy: If this is private, should it be prefixed with _?

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -350,41 +348,56 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
...
+  });
+}
 
-  if (!_downloaderFlags.downloaderImplementsSetProgress || _downloadIdentifier == nil) {
+#pragma mark - Private methods, call with lock held
+
+- (void)_updateProgressImageBlockOnDownloaderIfNeeded
+{
+  // If the downloader doesn't do progress, we are done.
+  if (_downloaderFlags.downloaderImplementsSetProgress == NO) {
```
appleguy: Does this not need the || _downloadIdentifier == nil check that previously existed?

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -350,41 +348,56 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
...
-  
-  BOOL shouldRenderProgressImages = _shouldRenderProgressImages;
-  ASInterfaceState interfaceState = self.interfaceState;
+  // Getting a result back for a different download identifier, download must not have been successfully canceled
+  if (ASObjectIsEqual(_downloadIdentifier, downloadIdentifier) == NO && downloadIdentifier != nil) {
+    return;
+  }
+  self.image = progressImage;
+  dispatch_async(dispatch_get_main_queue(), ^{
+    // See comment in -displayDidFinish for why this must be dispatched to main
```
appleguy: I wonder if there is any other solution that could avoid this.

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -350,41 +348,56 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
...
+    [_downloader setProgressImageBlock:nil callbackQueue:dispatch_get_main_queue() withDownloadIdentifier:oldDownloadIDForProgressBlock];
   }
-  [_downloader setProgressImageBlock:progress callbackQueue:dispatch_get_main_queue() withDownloadIdentifier:_downloadIdentifier];
+
+  // Bind to the current download.
+  if (newDownloadIDForProgressBlock != nil) {
+    __weak __typeof(self) weakSelf = self;
+    [_downloader setProgressImageBlock:^(UIImage * _Nonnull progressImage, CGFloat progress, id  _Nullable downloadIdentifier) {
+      [weakSelf handleProgressImage:progressImage progress:progress downloadIdentifier:downloadIdentifier];
+    } callbackQueue:dispatch_get_main_queue() withDownloadIdentifier:newDownloadIDForProgressBlock];
```
appleguy: @Adlai-Holler  ah, this is very interesting to me...see comments in #cx-ios for the amount of delay that occurs by sending UI loading-critical things through the main thread (independent of time they take to execute - queuing time).

It would be highly desirable to make it possible to do this on a different queue.  Would it also allow the previous main queue usage to be changed, if we created a single global serial queue for handling communications from the image downloader?

AsyncDisplayKit/ASNetworkImageNode.mm
Adlai-Holler: I didn't document it, but my original intention here is that underscored methods should assume the lock is held. It's not *really* private since it's called into from the downloader. I'll add some kind of documentation.

AsyncDisplayKit/ASNetworkImageNode.mm
Adlai-Holler: No. Even if `_downloadIdentifier == nil` we still want to continue. We may need to disconnect from the prior download identifier.

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -350,41 +348,56 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
...
+
+  // Unbind from the previous download.
+  if (oldDownloadIDForProgressBlock != nil) {
+    [_downloader setProgressImageBlock:nil callbackQueue:dispatch_get_main_queue() withDownloadIdentifier:oldDownloadIDForProgressBlock];
   }
-  [_downloader setProgressImageBlock:progress callbackQueue:dispatch_get_main_queue() withDownloadIdentifier:_downloadIdentifier];
+
+  // Bind to the current download.
+  if (newDownloadIDForProgressBlock != nil) {
+    __weak __typeof(self) weakSelf = self;
```
Adlai-Holler: Yes but `instancetype` can't be used in this context. It can only be used as the return type for an instance method.

AsyncDisplayKit/ASNetworkImageNode.mm
Adlai-Holler: Sure, but I don't want to go down that rabbit hole right now. Let's file a task!

AsyncDisplayKitTests/ASNetworkImageNodeTests.m
```diff
@@ -0,0 +1,95 @@
...
+  node.URL = [NSURL URLWithString:@"http://imageA"];
+  [[downloader expect] setProgressImageBlock:[OCMArg isNotNil] callbackQueue:OCMOCK_ANY withDownloadIdentifier:@0];
+  [node enterInterfaceState:ASInterfaceStateInHierarchy];
+  [downloader verify];
+  [[downloader expect] setProgressImageBlock:[OCMArg isNil] callbackQueue:OCMOCK_ANY withDownloadIdentifier:@0];
+  [node exitInterfaceState:ASInterfaceStateInHierarchy];
+  [downloader verify];
+}
+
+- (void)testThatProgressBlockIsSetAndClearedCorrectlyOnChangeURL
```
Adlai-Holler: Yes, `testThatProgressBlockIsSetAndClearedCorrectlyOnChangeURL` failed.

hannahmbanana:
Restarted the build. I'm getting a failed unit test (probably) related to this PR.

PR #2387, opened 2016-10-14
[Layout] Introduce layoutElementThatFits:

maicki:
With the changes we did for 2.0 we are now able to return an object that conforms to the `ASLayoutElement` protocol from `layoutSpecThatFits:`. Unfortunately `layoutSpecThatFits:` only allows to return objects from  type `ASLayoutSpec`, so just one kind of class of objects that currently conforms to the `ASLayoutElement` protocol can be returned.

With this PR we would introduce a new method with the name `layoutElementThatFits:` and it would be possible to return any object that conforms to the `ASLayoutElement` protocol than.

What would be the benefits?
For now this would allow us to remove the `ASWrapperLayoutSpec` class completely before we even introduce it. E.g in Pinterest we could change the following code:
```Objc
{
    self.controlNode.layoutPosition = CGPointZero;
    self.controlNode.sizeRange = ASRelativeSizeRangeMakeWithExactCGSize(…);
    return [ASStaticLayoutSpec staticLayoutSpecWithChildren:@[self.controlNode]];
}
```
to currently:
```Objc
{
  self.controlNode.style.preferredSize = CGSizeMake(…);
  return [ASWrapperLayoutSpec wrapperWithLayoutElement:self.controlNode];
}
```

and with `layoutElementThatFits:` support:
```Objc
- (id<ASLayoutElement>)layoutElementThatFits:(ASSizeRange)constrainedSize
{
   self.controlNode.style.preferredSize = CGSizeMake(…);
   return self.controlNode;
}
```
or even:
```Objc
- (id<ASLayoutElement>)layoutElementThatFits:(ASSizeRange)constrainedSize
{
  return [self.controlNode styledWithBlock:^(ASLayoutElementStyle *style) {
    style.preferredSize = CGSizeMake(...);
  }]
}
```
In the future the idea and idea would be to also allow to return an `ASLayoutElement` that involves `self` in `layoutElementThatFits:` to be able to remove the internal `finalLayoutable` property that was kind of a hack …

What should we do with `layoutSpecThatFits:`?
I think it should be deprecated as it’s mostly a search and replace

What should we do with `layoutSpecBlock`?
`layoutSpecBlock` is a difficult topic and I don’t like the current idea around it, as it does not provide subclass / chaining support with other `layoutSpecBlocks` or `layoutSpecThatFits`. I tend to to also deprecate it and for now advice to create a subclass and implement `layoutElementThatFits:`.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -63,9 +63,6 @@ typedef struct {
 + (BOOL)suppressesInvalidCollectionUpdateExceptions AS_WARN_UNUSED_RESULT;
 + (void)setSuppressesInvalidCollectionUpdateExceptions:(BOOL)suppresses;
 
-/** @name Layout */
-
-
 /**
  * @abstract Recursively ensures node and all subnodes are displayed.
  * @see Full documentation in ASDisplayNode+FrameworkPrivate.h
@@ -88,6 +85,23 @@ typedef struct {
 @property (nonatomic, copy, nullable) ASDisplayNodeContextModifier didDisplayNodeContentWithRenderingContext;
 
 /**
+ * @abstract Return a layout element that describes the layout of the receiver and its children.
+ *
+ * @param constrainedSize The minimum and maximum sizes the receiver should fit in.
+ *
+ * @discussion Subclasses that override should expect this method to be called on a non-main thread. The returned layout element
+ * is used to calculate an ASLayout and cached by ASDisplayNode for quick access during -layout. Other expensive work that needs to
+ * be done before display can be performed here, and using ivars to cache any valuable intermediate results is
+ * encouraged.
+ *
+ * @note This method should not be called directly outside of ASDisplayNode; use -measure: or -calculatedLayout instead.
+ *
+ * @warning Subclasses that implement -layoutElementThatFits: must not also implement -layoutSpecThatFits: or use
+ * .layoutSpecBlock. Doing so will trigger an exception.
+ */
+- (id<ASLayoutElement>)layoutElementThatFits:(ASSizeRange)constrainedSize;
+
+/**
  * @abstract A bitmask representing which actions (layout spec, layout generation) should be measured.
  */
 @property (nonatomic, assign) ASDisplayNodePerformanceMeasurementOptions measurementOptions;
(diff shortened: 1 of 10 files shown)


appleguy:
@maicki Nice to have this route available!  Any comments you can add here on what you think we should do?  This would be a pretty significant change, but it's the right time to consider it.

Is there any other advantage besides avoiding the return [ASLayoutSpec layoutSpecWithNode:self] line, and replacing it with return self; ?

I do think that 'size to match parent' easy path is very useful.  If that's the only advantage though, looking at the two names:

- (id<ASLayoutElement>)layoutElementThatFits:
- (ASLayoutSpec *)layoutSpecThatFits:

The layout spec method feels easier to understand.  This isn't just because of the protocol / return value, but because layoutSpec seems to clearly name it is the method to specify layout.  A layoutElement is a bit more vague of a connection - does a layoutElement just exist within a layout, or does it _define_ the layout?

Open to all thoughts and ideas here; let's decide this week.  @garrettmoon @Adlai-Holler @hannahmbanana  @nguyenhuy

maicki:
@appleguy This PR is still in development that's the reason there is no description yet. I thought about it today and I would like to go the route to actually deprecate `layoutSpecThatFits:`, but we should talk about it if we would like to consider adding this to the 2.0 roadmap, what I'm not really sure about.

maicki:
@appleguy Updated comments

maicki:
@appleguy I moved `layoutElementThatFits:` to a beta header. I still think this is the right approach to go for the future as it would allow developers to return any object that conforms to `ASLayoutElement` (program to  interface not implementations). I don't know if we should do more API changes for 2.0 for this so I moved it to a beta header for now ...

Furthermore this PR includes some cleanup internally that would be great to have too.

@Adlai-Holler @appleguy Ready for review.

maicki:
@Adlai-Holler @appleguy What do you guys think of merging that in soon (after rebasing) and disable `layoutElementThatFits:` for now in the public API. Besides introducing `layoutElementThatFits:` it also has some nice code cleanup that would be good to have in.

Adlai-Holler:
Geez this is a tough one. Now that we have ASWrapperLayoutSpec, which is very easy to understand, I lean toward keeping `layoutSpecThatFits` as the central layout spec method. I do want the code cleanup, but we should avoid bringing in the underlying changes like `_layoutElementThatFits` if we're not going to use it. Do you feel strongly about this @maicki ?

maicki:
@Adlai-Holler Valid point and I'm unsure myself if this would harm the API more as it would improve it. As #2469 just landed let me put up a new PR (or change the PR) with just the code cleanup.

maicki:
Will be replaced by: #2480 that has the cleanup that was also in this PR but without adding `layoutElementThatFits:`

PR #2388, opened 2016-10-15
[ASCellNode] Add subclass method cache; skip visibility rect calculations in the common case.

appleguy:
Profiling showed this area as the most expensive part of ASRangeController propogation of .interfaceState, when scrolling an ASTableView.

This is very much on the critical path for sustained 60FPS in these views.

Before **This is filtered to only the method in question**, meaning a total of 1.4% impact on main with a single moderate-speed swipe.  It is also triggered in a different runloop stage offscreen:
screen shot 2016-10-14 at 8 09 29 pm

After *not directly comparable* but shows the absence of the method:
screen shot 2016-10-14 at 8 36 10 pm

### DIFF
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -11,6 +11,7 @@
 #import "ASCellNode+Internal.h"
 
 #import "ASEqualityHelpers.h"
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Excellent! Love the use of +initialize to keep the set of classes up to date.

PR #2389, opened 2016-10-15
[ASDisplayNode] Return constrainedSize in ASDisplayNode

maicki:
This change will return `constrainedSize` from ASDisplayNode in `calculateSizeThatFits:` instead of CGSizeZero. The reason for that is that the `constrainedSize` usually is within the range of the `style.preferredSize`. The old behavior did not consider that.

* Removes the invalid `constrainedSize` check within `ASNetworkImageNode`.
* Fix tests: As ASDisplayNode does not return CGSizeZero anymore we have to give the display nodes we use in tests and are involving a stack spec an intrinsic content size.,

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2491,7 +2491,7 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 {
   __ASDisplayNodeCheckForLayoutMethodOverrides;
 
-  return CGSizeZero;
+  return constrainedSize;
 }
 
 - (id<ASLayoutElement>)_layoutElementThatFits:(ASSizeRange)constrainedSize
--- a/AsyncDisplayKit/ASImageNode.mm
+++ b/AsyncDisplayKit/ASImageNode.mm
@@ -190,7 +190,7 @@ struct ASImageNodeDrawParameters {
   ASDN::MutexLocker l(__instanceLock__);
 
   if (_image == nil) {
-    return constrainedSize;
+    return [super calculateSizeThatFits:constrainedSize];
   }
 
   return _image.size;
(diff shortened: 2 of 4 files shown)


maicki:
@garrettmoon @Adlai-Holler Ready for review

@Adlai-Holler Maybe worth it for this behavior to add a unit test in #2386?

maicki:
The alternative would be to remove the whole thing as it's wrong behavior in the first place. I don't think @appleguy will too happy about that though as you know we will get a lot of complaints.

maicki:
@garrettmoon @Adlai-Holler After looking into it a bit more we should add the assertion we have in here as the whole purpose of the network image node is to load an image from the network with a given URL. That said I'm not really sure if we should actually remove the whole check after 2.0. It kind of reflects the bahavior on the web if the img tag has a src but it's not fully loaded yet.

I tend to either add the assertion from this pull request and remove the comment about removing this check after 2.0 *OR* remove it all completely now.

I think we should make a call soon and either merge this PR or close it based on the decision.

Adlai-Holler:
I think we should remove it all completely now – it should be valid to have network image nodes without URLs.

maicki:
@Adlai-Holler You think we should remove the URL check of the whole check?

Adlai-Holler:
@maicki I was thinking we remove the override entirely – `ASImageNode`'s implementation of `calculateSizeThatFits:` seems correct to me. If we have a size from our image, use it. If we don't, just use the constrained size. Does that make sense?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2491,7 +2491,7 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 {
   __ASDisplayNodeCheckForLayoutMethodOverrides;
 
-  return CGSizeZero;
+  return constrainedSize;
```
Adlai-Holler: Should we check if this is a valid size before returning it? Seems like we should return CGSizeZero if it's invalid.

maicki:
Superseeded by: #2505

PR #2392, opened 2016-10-16
[ASTextNode] Add Fast-Paths For Text Measurement and Drawing

Adlai-Holler:
There's currently one failing test with this, and I don't have the energy to investigate it right now. From the looks of it, it's not meaningful and just a hassle.

Here's the results of running the LayoutKit benchmark on my iOS 9.3 iPod touch 5g, newly ported to LayoutKit 3.0 and Swift 3:

screen shot 2016-10-15 at 7 03 21 pm


appleguy:
@Adlai-Holler Awesome, thanks for creating this on a Saturday!  The graph is really cool :).

Is the latest benchmark on your LayoutKit branch?  As we know, this ops/sec test is not exactly the most important metric for user-percieved performance, but still I'm curious to try out the latest on a tri-core iPad Air 2 and look at the key sources of contention.

Adlai-Holler:
@appleguy Yep! Latest benchmark is at https://github.com/Adlai-Holler/LayoutKit/tree/AddASDK just point the Podfile as you desire.

appleguy:
Cool, I have some interesting profiling results.  We are spending an astounding amount of time dealing with the weak variable .layoutElement on ASLayout.

We're also triggering creation of the new .style object even when it is not needed / set to all-default values, but this is nowhere near as significant as the weak variable and general object cost of ASLayout.

This is an **inverted** tree showing the cost of weak, primarily layoutElement, on the powerful A8X over a short period of ~2 seconds of intense work.  The figures are pretty astounding!  The folded aggregations that do exceed weak represent the cost of object allocations in general, but unsure how actionable that is.

If we found an alternative data structure to perform the ASLayout -> ASLayoutElement weak mapping, it could pay off hugely, because of the extreme commonality of these objects (during both initial layout calculation and then flattening).  If we could ensure safety of unsafe_unretained (maybe by having the ASLayoutElement clear its own back-pointer on dealloc), that's one fix.  Perhaps only during the initial layout tree creation, before the longer-lived flattened version is created, there may be a guarantee that the ASLayoutElement will live as long as the ASLayout.

screen shot 2016-10-16 at 12 03 55 am

Adlai-Holler:
@appleguy @maicki Awesome find! I notice that in `filteredNodeLayoutTree` we copy all of the display node layouts and change `flattened` and `position` on the copy. 

It seems likely that we can skip this copy. The old `position` value is probably not important to anyone. And `flattened` is only read on the layout's sublayouts in that same method, so unless we have a circular dependency the flag could probably be removed entirely.

Thoughts?

maicki:
@appleguy @Adlai-Holler Another approach is to finally create a new private (or layoutspec / subclass) header for `ASLayout` that we use internally in ASDK that allows us to change the properties like "flatten" and "position" of the layout without having to make an explicit copy. Furthermore this would allow us to declare the position of the `ASLayout` in the public header finally as `readonly`.

Creating the internal header would allow us to get rid of that extra creation of the `ASLayout` object.

appleguy:
Re-ran the benchmark with some of the other optimizations I've been working on locally:

screen shot 2016-10-16 at 7 08 34 pm

Adlai-Holler:
@appleguy  🙌 Awesome! I saw your email about `weak` and I'll be contemplating it.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -147,7 +148,18 @@ - (void)_calculateSize
...
+  BOOL defaultTruncation = _attributes.avoidTailTruncationSet == nil && [_attributes.truncationAttributedString.string isEqualToString:@"\u2026"];
+  BOOL doesNoExclusion = _attributes.exclusionPaths.count == 0;
+  if (doesNotScale && defaultTruncation && doesNoExclusion) {
+    CGSize baseSize = [_attributes.attributedString boundingRectWithSize:_constrainedSize options:NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingTruncatesLastVisibleLine context:nil].size;
+    _calculatedSize = [self.shadower outsetSizeWithInsetSize:baseSize];
+    return;
+  }
+
   __block NSTextStorage *scaledTextStorage = nil;
   BOOL isScaled = [self isScaled];
```
appleguy: @Adlai-Holler Is this variable different or can be shared with `doesNotScale`?

Also, I think _currentScaleFactor always *starts* at 1 — but if pointScaleFactors has other values, and the content doesn't fit the constraint, then other scales are tried.  Is it correct that if _currentScaleFactor is 1 on this pass, that we are not using scaling at all?

Best option would be to find one of our uses of pointScaleFactors in Pinterest and test this with a hardcoded string that triggers the scaling.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -147,7 +148,18 @@ - (void)_calculateSize
...
   }
-  
+
+  // If we do not scale, do exclusion, or do custom truncation, we should just use TextKit for a fast-path.
+  BOOL doesNotScale = _currentScaleFactor == 1;
+  // NOTE: This code does not correctly handle if they set `…` with different attributes.
+  BOOL defaultTruncation = _attributes.avoidTailTruncationSet == nil && [_attributes.truncationAttributedString.string isEqualToString:@"\u2026"];
+  BOOL doesNoExclusion = _attributes.exclusionPaths.count == 0;
+  if (doesNotScale && defaultTruncation && doesNoExclusion) {
+    CGSize baseSize = [_attributes.attributedString boundingRectWithSize:_constrainedSize options:NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingTruncatesLastVisibleLine context:nil].size;
```
appleguy: @Adlai-Holler does the constrainedSize need to apply the content inset values?  I'm not sure exactly what the semantics of the new inset property are - they may be unrelated to sizing.

appleguy: @Adlai-Holler could you add a comment about the options chosen here, and maybe break them into a local variable for a shorter line length?  I'm not familiar with LineFragmentOrigin / TruncatesLastVisible, and it would be helpful to describe why these options are chosen.

appleguy: We should try running Pinterest in the CJK / scaled languages (Japanese is a good one for testing).  I think Garrett is familiar with what to look for, if Ricky isn't back on Monday.

AsyncDisplayKitTests/ASTextNodeSnapshotTests.m
```diff
@@ -72,7 +72,21 @@ - (void)testTextContainerInsetHighlight
...
+}
+
+- (void)testThatTextKitTruncationMatchesStyle
+{
+  NSAttributedString *str = [[NSAttributedString alloc] initWithString:@"Quality is Important" attributes:@{ NSForegroundColorAttributeName: [UIColor blueColor], NSFontAttributeName: [UIFont italicSystemFontOfSize:24] }];
+  CGRect bounds = CGRectMake(0, 0, 110, 30);
+  UIGraphicsBeginImageContextWithOptions(bounds.size, YES, 1);
+  [[UIColor whiteColor] setFill];
+  UIRectFill(bounds);
+  [str drawWithRect:bounds options:NSStringDrawingTruncatesLastVisibleLine | NSStringDrawingUsesLineFragmentOrigin context:nil];
```
appleguy: Not a big deal, but the options are provided in flipped order compared to the measurement step.

Is this the same command we've already been using for drawing?  I assume so.  Nice test!

appleguy:
test this please

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -161,15 +188,13 @@ - (void)_calculateSize
...
+
   // Force glyph generation and layout, which may not have happened yet (and isn't triggered by
   // -usedRectForTextContainer:).
   [[self context] performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
     [layoutManager ensureLayoutForTextContainer:textContainer];
-  }];
-  
-  CGRect constrainedRect = {CGPointZero, _constrainedSize};
-  __block CGRect boundingRect;
-  [[self context] performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
```
Adlai-Holler: We were calling `performBlockWithLock…` twice in a row, without doing any work in-between!

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -203,42 +238,50 @@ - (void)drawInContext:(CGContextRef)context bounds:(CGRect)bounds;
     [self _calculateSize];
   }
 
+  bounds = CGRectIntersection(bounds, { .size = _constrainedSize });
```
Adlai-Holler: This wasn't needed before, because the slow-path drawing actually doesn't use `bounds`. It uses `textContainer.size` which got updated when the constrained size changed to account for the text inset. Since the fast-path does use `bounds`, this is necessary. We could absolutely refactor this, to have the constrained (inset) bounds passed into this method but time is of the essence.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -203,42 +238,50 @@ - (void)drawInContext:(CGContextRef)context bounds:(CGRect)bounds;
...
   [[self shadower] setShadowInContext:context];
   UIGraphicsPushContext(context);
 
   LOG(@"%@, shadowInsetBounds = %@",self, NSStringFromCGRect(shadowInsetBounds));
-  
-  [[self context] performBlockWithLockedTextKitComponents:^(NSLayoutManager *layoutManager, NSTextStorage *textStorage, NSTextContainer *textContainer) {
-    
-    NSTextStorage *scaledTextStorage = nil;
-    BOOL isScaled = [self isScaled];
+  BOOL isScaled = [self isScaled];
```
Adlai-Holler: Use split view here. I had to change the indentation of the slow-path rendering code because early return wasn't an attractive option, since we need to pop the context and restore the state. So `if-else` it is, and GitHub isn't as smart as Phabricator with indentation-only changes.

Adlai-Holler:
@appleguy Alright we're ready to rick! Jenkins failed because it couldn't find `PINCache` when building the `ASCollectionView` example. Strange. But the unit tests on Jenkins passed!

Adlai-Holler:
@appleguy The `NSStringDrawingContext` documentation makes it seem like it's just for adjusting scale factor. [Not so!](https://github.com/nst/iOS-Runtime-Headers/blob/master/PrivateFrameworks/UIFoundation.framework/NSStringDrawingContext.h)

appleguy:
test this please

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -144,9 +162,18 @@ - (void)_calculateSize
   if (isinf(_constrainedSize.width) == NO && [_attributes.pointSizeScaleFactors count] > 0) {
     _currentScaleFactor = [[self fontSizeAdjuster] scaleFactor];
   }
-  
-  __block NSTextStorage *scaledTextStorage = nil;
+
+  // If we do not scale, do exclusion, or do custom truncation, we should just use TextKit for a fast-path.
```
appleguy: Is it "just use TextKit", or more like "Just use NSAttributedString's built-in sizing?"  Or is this implemented in Foundation or UIKit as a category?

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -144,9 +162,18 @@ - (void)_calculateSize
...
   BOOL isScaled = [self isScaled];
```
appleguy: Not new with this diff, but should this be scalingEnabled or requiresScaling?  isScaled sounds like it already knows it will have to be scaled down.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -144,9 +162,18 @@ - (void)_calculateSize
...
+  if (isScaled == NO && self.usesCustomTruncation == NO && self.usesExclusionPaths == NO) {
```
appleguy: Is there a point at which we can lock to collect these various values for usesCustomTruncation, isScaled, etc with one lock event?  Or are all of these methods assumed to be called when locked anyway?

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -144,9 +162,18 @@ - (void)_calculateSize
...
+    CGRect rect = [_attributes.attributedString boundingRectWithSize:_constrainedSize options:NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingTruncatesLastVisibleLine context:self.stringDrawingContext];
```
appleguy: Is it worth checking if attributedString is nil?  I think recent LLVM makes it safe to use the struct returned by messaging nil, but historically this was not safe, and I'm not totally sure.

appleguy:
Significant improvements!  Before is Run 1, after is Run 2.  Scrolling the same board, but there are suggestions at the bottom so workload is not identical in size - but close.  Compare percentage instead of ms.

screen shot 2016-10-21 at 8 23 14 pm
screen shot 2016-10-21 at 8 21 19 pm
screen shot 2016-10-21 at 8 20 16 pm
screen shot 2016-10-21 at 8 18 07 pm
screen shot 2016-10-21 at 8 16 15 pm

Adlai-Holler:
Note: This diff has an issue where numberOfLines is not respected. Will attempt to test & address today.

PR #2393, opened 2016-10-16
[ASRelativeLayoutSpec] Fix ASRelatativeLayoutSpec behavior for ASRelativeLayoutSpecPositionStart

maicki:
Currently if no `ASRelativeLayoutSpecPosition` is given `ASRelativeLayoutSpecPositionStart` is assumed. This unfortunately has some problems as it was ported over from `ASCenterLayoutSpec` 

The problem is that if no position for one of the axis of `ASCenterLayoutSpec` was given the spec automatically tried to size the children to the min constrained size of that axis. As the spec was kind of ported to `ASRelativeLayoutSpec` this behavior was introduced for axis that have `ASRelativeLayoutSpecPositionStart` assigned. This PR addresses this by introducing `ASRelativeLayoutSpecPositionNone` that has the same behavior as no position in `ASCenterLayoutSpec` and `ASRelativeLayoutSpecPositionStart` will shrink the size of the child to fit and move it to the start of the axis.

Fore reference this code produce the screenshots below before and after the fix:
```Objc
{
    _buttonNode.style.preferredSize = CGSizeMake(100.0, 30.0);
    return
     [ASRelativeLayoutSpec
      relativePositionLayoutSpecWithHorizontalPosition:ASRelativeLayoutSpecPositionStart
      verticalPosition:ASRelativeLayoutSpecPositionCenter
      sizingOption:ASRelativeLayoutSpecSizingOptionMinimumSize
      child:_buttonNode];
}
```

Before:
screen shot 2016-10-15 at 7 53 24 pm

After:
screen shot 2016-10-15 at 7 51 54 pm

Sample project:
[Archive.zip](https://github.com/facebook/AsyncDisplayKit/files/531566/Archive.zip)

**TODO:**
- [x] Update `ASRelativeLayoutSpec` screenshots for tests


maicki:
To be on par with the behavior of `ASAbsoluteLayoutSpec`, `ASRatioLayoutSpec` etc. behavior as well as to support `ASCenterLayoutSpec` we need to the `None` option to tell the spec to stretch the child in the `None` axis to as far as possible.

maicki:
@Adlai-Holler Updated screenshots

PR #2394, opened 2016-10-16
[ASAbsoluteLayoutSpec] Add ASAbsoluteLayoutSpecSizing

maicki:
Add ability for ASAbsoluteLayoutSpec to grow as far as possible or shrink to fit it's children

### DIFF
--- a/AsyncDisplayKit/Layout/ASAbsoluteLayoutSpec.h
+++ b/AsyncDisplayKit/Layout/ASAbsoluteLayoutSpec.h
@@ -10,14 +10,31 @@
 
 #import <AsyncDisplayKit/ASLayoutSpec.h>
 
+/** How much space the spec will take up. */
+typedef NS_ENUM(NSInteger, ASAbsoluteLayoutSpecSizing) {
+  /** The spec will take up the maximum size possible. */
+  ASAbsoluteLayoutSpecSizingDefault,
+  /** Computes a size for the spec that is the union of all childrens' frames. */
+  ASAbsoluteLayoutSpecSizingSizeToFit,
+};
+
 NS_ASSUME_NONNULL_BEGIN
 
 /**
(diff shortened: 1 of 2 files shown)


maicki:
It does not change the behavior of `ASAbsoluteLayoutSpec` if you don't see it as a find a replace for `ASStaticLayoutSpec`. We need to add that to the conversion guide to set the flag if you want the behavior from `ASStaticLayoutSpec`.

`ASAbsoluteLayoutSpec` actually works now the same way in terms of sizing like `ASCenterLayoutSpec` and other layout specs. It just adds the ability to size itself to fit it's content.

AsyncDisplayKit/Layout/ASAbsoluteLayoutSpec.h
```diff
@@ -10,16 +10,33 @@
...
 /**
+ How much space will the spec taken up
+ */
+@property (nonatomic, assign) ASAbsoluteLayoutSpecSizing sizing;
+
+/**
+ @param sizing How much space the spec will take up
+ @param children Children to be positioned at fixed positions
+ */
++ (instancetype)absoluteLayoutSpecWithSizing:(ASAbsoluteLayoutSpecSizing)sizing children:(NSArray<id<ASLayoutElement>> *)children AS_WARN_UNUSED_RESULT;
```
garrettmoon: should this be `absoluteLayoutSpecWithChildren:sizing:`? To match the next one?

maicki: All the other layout spec have the children at the end and I think it's the right approach also here.

garrettmoon: That's sounds right to me.

PR #2395, opened 2016-10-16
[ASLayoutElement] Move ASDisplayNode's .name property to ASLayoutElement (renamed to .debugName)

hannahmbanana:
Improves ascii art debugging - configurable `-asciiArtName` return value.

image


maicki:
Agree let's add the name to the `ASDisplayNode+Deprecated.h` header for 2

AsyncDisplayKit/ASDisplayNode+Deprecated.h
```diff
@@ -15,6 +15,14 @@
 @interface ASDisplayNode (Deprecated)
 
 /**
+ * @abstract The name of this node, which will be displayed in `description`. The default value is nil.
+ * 
+ * @deprecated Deprecated in version 2.0: Use .debugName instead. This value will display in 
+ * results of the -asciiArtString method (@see ASLayoutElementAsciiArtProtocol).
+ */
+@property (nullable, nonatomic, copy) NSString *name;
```
appleguy: does this need to have an NS_DEPRECATED or something?  There might be an ASDK version of that annotation too.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3488,6 +3488,20 @@ - (UIView *)preferredFocusedView
 
 #pragma mark - Deprecated
 
+- (NSString *)name
+{
+  ASDN::MutexLocker l(__instanceLock__);
```
appleguy: rewrite this to do return self.debugName in one line - then the locking can be done in debugName accessor

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3488,6 +3488,20 @@ - (UIView *)preferredFocusedView
...
+  return _debugName;
+}
+
+- (void)setName:(NSString *)name
+{
+  ASDN::MutexLocker l(__instanceLock__);
```
appleguy: same, rewrite this to be self.debugName = name —locking and equality check handled in the setter.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3551,12 +3565,16 @@ - (NSString *)_recursiveDescriptionHelperWithIndent:(NSString *)indent
...
-    return [ASLayoutSpec asciiArtStringForChildren:@[] parentName:[self asciiArtName]];
+  return [ASLayoutSpec asciiArtStringForChildren:@[] parentName:[self asciiArtName]];
 }
 
 - (NSString *)asciiArtName
 {
-    return NSStringFromClass([self class]);
+  NSString *string = NSStringFromClass([self class]);
+  if (_debugName) {
+    string = [string stringByAppendingString:[NSString stringWithFormat:@" (debugName = %@)",_debugName]];
```
appleguy: Suggest maybe put in "\"%@\"" to have only the _debugName shown directly in quotes - to shorten the asciiArt constructs that are created

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -18,6 +18,7 @@ @implementation ASLayoutSpec
 
 // Dynamic properties for ASLayoutElements
 @dynamic layoutElementType;
+@synthesize debugName = _debugName;
```
appleguy: I think you probably don't need this line, because you implemented the getter and setter method.  See if it compiles, since you do need it to create the _debugName instance variable, but I think it will do that from the @property alone.

AsyncDisplayKit/ASDisplayNode+Deprecated.h
hannahmbanana: eek. good catch!

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -56,6 +57,20 @@ - (BOOL)canLayoutAsynchronous
   return YES;
 }
 
+- (NSString *)debugName
```
appleguy: Move this to the bottom of the file, it's mostly boilerplate / debugging feature that should be out of the way if possible

AsyncDisplayKit/Layout/ASLayoutSpec.mm
hannahmbanana: It doesn't compile without.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3309,8 +3309,8 @@ - (void)setInHierarchy:(BOOL)inHierarchy
 {
   NSMutableArray<NSDictionary *> *result = [NSMutableArray array];
   
-  if (self.name.length > 0) {
-    [result addObject:@{ @"name" : ASStringWithQuotesIfMultiword(self.name)}];
+  if (self.debugName.length > 0) {
+    [result addObject:@{ @"debugName" : ASStringWithQuotesIfMultiword(self.debugName)}];
```
appleguy: I didn't even know we had this Multiword function, nice :)

PR #2397, opened 2016-10-16
[ASLayout - Performance] Avoid weak ivar access & method call to check layoutElementType.

appleguy:
This shows up in benchmarking profiles for layout-bound workloads.

@maicki @Adlai-Holler @garrettmoon 

screen shot 2016-10-16 at 11 55 42 am

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayout.mm
+++ b/AsyncDisplayKit/Layout/ASLayout.mm
@@ -40,7 +40,9 @@ static inline NSString * descriptionIndents(NSUInteger indents)
 }
 
 @interface ASLayout () <ASDescriptionProvider>
-
+{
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -68,6 +70,7 @@ - (instancetype)initWithLayoutElement:(id<ASLayoutElement>)layoutElement
 #endif
     
     _layoutElement = layoutElement;
+    _layoutElementType = layoutElement.layoutElementType;
```
maicki: Nit but maybe we could add a comment why we do that in this case. Otherwise in the future maybe we will look over it and wonder why we not reading it directly in the `type` method

PR #2398, opened 2016-10-16
[Layout] Don't Copy Layout Tree During Filtering

Adlai-Holler:
I haven't profiled this yet, but it should be very safe and quite a nice performance win.

As @appleguy identified in #2392, there's a lot of headroom in terms of reducing object allocations and `weak` object handling.

I removed the `flattened` flag because I don't believe it serves a purpose. During the filtering, we only travel _down_ the tree. In the prior implementation, we only marked the newly-copied ASLayout objects as `flattened` so it's impossible that we would encounter a flattened layout as we traverse down the sublayouts.

After this change, the `position` value of the sublayouts will be changed by calling `filteredNodeLayoutTree`. I think this is fine. Nobody cares about the old `position` values.

This needs some profiling!

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayout.mm
+++ b/AsyncDisplayKit/Layout/ASLayout.mm
@@ -40,12 +40,6 @@ static inline NSString * descriptionIndents(NSUInteger indents)
 }
 
 @interface ASLayout () <ASDescriptionProvider>
-
-/**
- * A boolean describing if the current layout has been flattened.
- */
-@property (nonatomic, getter=isFlattened) BOOL flattened;
-
 @end
 
 @implementation ASLayout
@@ -84,7 +78,6 @@ static inline NSString * descriptionIndents(NSUInteger indents)
     }
 
     _sublayouts = sublayouts != nil ? [sublayouts copy] : @[];
-    _flattened = NO;
   }
   return self;
 }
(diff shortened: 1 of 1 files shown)


maicki:
As I said already in here: https://github.com/facebook/AsyncDisplayKit/pull/2392#issuecomment-254068139

I would propose to not go this route and instead add an internal header where we expose properties for our use internally. The clear benefit of that would be to still have the `flattened` property if needed and move the `position` property to readonly in the public header.

appleguy:
@maicki Could we still have position be readonly externally, but writeable internally?  I don't have an opinion yet on if this change is the right direction, although reducing object allocations will definitely be valuable here.

maicki:
@appleguy Yes! The position will be `readonly` public no problem and internally we can expose properties of `ASLayout` that we need `readwrite` like `position`

Adlai-Holler:
@maicki Adding an internal header to make `position` `readwrite` seems right, but `flattened` is absolutely useless and should be removed. 

I've modified the diff to lock `position` and noted a potential race condition. We could build basically a stochastic thrash unit test that attempts to reproduce the issue I described but the risk is very low.

Adding the internal header could be accomplished in a follow-up diff.

appleguy:
Profiling suggests that this does help quite a lot - I think we should take this change, and adapt it further as needed.  @maicki I think you know this code the best of us, so feel free to change this again in the future (maybe not a top priority now).

The layoutElement weak pointer is still one of the top causes of contention even after this change, but it is in a more reasonable range.  If we can think of a way to avoid using weak - some other data store that even if it behaves similarly, does not have lock contention against other weak variables- it would be worth doing.

appleguy:
test this please

Adlai-Holler:
Superseded by #2404

PR #2399, opened 2016-10-16
[ASRunLoopQueue - Performance] Add ASDeallocQueue for efficient object teardown.

appleguy:
This measurably reduces block overhead and context switching.  In the layout benchmark, it increases ops/s while actually reducing CPU utilization.

This suggests that we are now at a lock-bounded local maximum, at least for tri-core devices.

@maicki @Adlai-Holler @garrettmoon first two screenshots give context on the nature of the costs before, and the new one is after (ms are comparable in this case).  Utilization peaked at 300% previously and now is closer to 210%.

screen shot 2016-10-16 at 12 27 00 pm

screen shot 2016-10-16 at 12 27 16 pm

screen shot 2016-10-16 at 1 30 19 pm


appleguy:
Interesting how many additional context switches there were, over the same 3 second period, before this change.  There are probably complex factors involved and comparing a 3 second period may not be the best way - but I anchored from the same point at the end of the benchmark (before display processes begin), and continued 3 seconds prior.

Old:
screen shot 2016-10-16 at 1 55 38 pm

New:
screen shot 2016-10-16 at 1 56 49 pm

AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -25,6 +26,34 @@ static void runLoopSourceCallback(void *info) {
...
+- (void)main
+{
+  [[NSRunLoop currentRunLoop] runUntilDate:[NSDate distantFuture]];
+}
+
+- (ASRunLoopQueue *)deallocQueue
+{
+  if (_deallocQueue == nil) {
+    NSRunLoop *runLoop = [NSRunLoop currentRunLoop];
+    _deallocQueue = [[ASRunLoopQueue alloc] initWithRunLoop:[runLoop getCFRunLoop] andHandler:nil];
```
Adlai-Holler: Could this use `CFRunLoopGetCurrent`?

appleguy: Sure, good suggestion

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -127,17 +128,12 @@ - (void)setConstrainedSize:(CGSize)constrainedSize
     // truncater do it's job again for the new constrained size. This is necessary as after a truncation did happen
     // the context would use the truncated string and not the original string to truncate based on the new
     // constrained size
-    __block ASTextKitContext *ctx = _context;
-    __block ASTextKitTailTruncater *tru = _truncater;
-    __block ASTextKitFontSizeAdjuster *adj = _fontSizeAdjuster;
+    [[ASRunLoopQueue sharedDeallocationQueue] enqueue:_context];
```
Adlai-Holler: Nit: Might be worth reading the deallocation queue only once here? Could we make a variant of the `enqueue` method that takes a variadic list of arguments?

appleguy:
Oh, also, forgot to include the ops/s values.  Yes it would be great to support variable arguments.  I'll make the change to get the queue only once.

screen shot 2016-10-16 at 2 04 15 pm

Adlai-Holler:
@appleguy Want to remove the old `ASPerformBlockOnDeallocationQueue` while we're at it? No reason to keep it around.

Adlai-Holler:
Nice, looks like a solid 15% increase, basically for free!

AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -25,6 +26,41 @@ static void runLoopSourceCallback(void *info) {
...
+
+@implementation ASDeallocThread
+
+- (void)main
+{
+  self.name = @"ASDeallocThread";
+  [self deallocQueue];
+  while (YES) {
+    // This method will still return once all sources have finished.
+    [[NSRunLoop currentRunLoop] runUntilDate:[NSDate distantFuture]];
```
appleguy: @maicki @Adlai-Holler looking for tips here.  I can't get this code right and it's driving me nuts.  Even though this line is called repeatedly, it does NOT result in the running of the observer's handler block.

Am I missing something obvious?  Is there a better way to keep the thread alive and runloop idling while still allowing the thread to sleep most of the time?

I'm considering ditching the runloop and just having the timer to periodically check the queue.  In this case, a separate class from ASRunloopQueue (but sharing the same file due to related purpose) would be simplest.  Especially if we can get a lockless queue in place.

maicki: Did you look into `ASRunLoopQueue` how we handle it? Especially look at the `_runLoopObserver` / `CFRunLoopObserverRef`. I know it's not a subclass of `NSThread` but I think you can model it similar to it.

appleguy: I think I figured it out.  CFRunloopRun() is actually, apparently, totally different than ANY option available on NSRunloop.  NSRunloop run will return, but CFRunloopRun will continue "indefinitely" per the documentation.  What a sham :-P

appleguy: @maicki my current implementation is using the runloop management in ASRunLoopQueue.  The NSThread exists so it has a single-threaded place to perform the work, without the deallocations going onto the main queue runloop.  I don't think I can get a runloop reference to provide to the ASRunLoopQueue unless I have the thread...

appleguy:
Much annoyance later, I'm ready to land this rewritten version.

@Adlai-Holler I turned off event logging by default even in DEBUG.  For the TableView example, we were exceeding 1GB of memory at peak (even without this change - it's not the queuing).  With those objects disabled, it's closer to 300MB.  I didn't look into it exhaustively, but there's a whole lot of the tracer objects :).

screen shot 2016-10-16 at 5 10 20 pm

Adlai-Holler:
@appleguy I don't agree that we should disable event tracing in debug.

- You don't know you need event logging until you do. If you stumble upon some rare crash/race condition, you can't relaunch the app to turn it on.
- This is a wacky synthetic benchmark that creates and measures as many nodes as possible all at once, so the performance implications of event tracing in this context shouldn't drive our decisions.

Adlai-Holler:
^ That said, the current log capacity is set to 20 events which I knew was quite liberal. It might be reasonable to reduce that significantly.

appleguy:
@Adlai-Holler With 284,000 events live at one time, is it possible they're staying around longer than expected?  I can leave it on, but am concerned about how it made up for seemingly 700MB - maybe it caused other large objects to stay alive?

appleguy:
test this please

appleguy:
@Adlai-Holler also, thanks for the clear feedback re: disabling by default.  Your reasoning is sound as usual, and I've reverted that side-effect of this change.

I agree we don't need to be too concerned with a synthetic benchmark; however, since running as DEBUG on a device is not too uncommon for Pinterest development, if the memory overhead reaches similar levels for complex UI flows then we should consider making it simulator-only (or reducing the log size some other way)

PR #2402, opened 2016-10-17
[Example] Rewrite `layoutSpecThatFits:` in ASDKgram

maicki:
The goal of this change is to have an extended example that show's what the Layout API is able to. 

The style of composing layout elements in `layoutSpecThatFits:` is most of the time matter of taste so this change can be controversial, but as said it's more for demonstrating what's possible now.


maicki:
@Adlai-Holler @appleguy @hannahmbanana  Any thoughts on merging? Otherwise let's close it if you think it's not worth it ...

maicki:
@Adlai-Holler Rebased

PR #2406, opened 2016-10-17
[ASNetworkImageNode] Make Test More Stable

Adlai-Holler:
This test is designed to ensure that network image nodes set and clear their progress blocks when the node becomes visible/invisible. This avoids extra progressive image renders.

Previously, the test had a race condition where the download might not have started (it starts asynchronously) before the node becomes visible and we verify that the progress image block is set.

Now the test waits for the download to start before making the node visible/invisible.


garrettmoon:
Awesome, LGTM!

appleguy:
Thanks!

Adlai-Holler:
Unit tests have passed. Merging so we can get started rebasing open PRs.

PR #2408, opened 2016-10-17
[Layout] Check Layout Spec Tree for Duplicated Elements

Adlai-Holler:
Currently we have a major issue where, if the user returns a layout spec tree that contains duplicated elements (nodes or specs) there will be a hard crash when the layout gets applied.

This diff adds an assertion that the tree contains no duplicated elements. In production, we will use an empty `ASLayoutSpec` in this case.

This is an extremely performance-sensitive method. It avoids allocating a result set unless needed, and it avoids membership checking in favor of checking `count`.

I considered using `NSHashTable` with `NSHashTableObjectPointerPersonality` to avoid the retains and hashing etc but avoided it because of [this table](https://www.objc.io/issues/7-foundation/collections/#performance-characteristics-of-nshashtable). It's possible that the table they used was configured differently, but it's more important to keep moving.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -36,6 +36,10 @@
 #import "ASWeakProxy.h"
 #import "ASLayoutSpecPrivate.h"
 
+#if DEBUG
+  #define AS_DEDUPE_LAYOUT_SPEC_TREE 1
+#endif
+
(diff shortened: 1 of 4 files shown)


Adlai-Holler:
Testing with Pinterest while the CI runs

Adlai-Holler:
Yeah @garrettmoon on second thought, let's disable this in production for now. If we see a bunch of crashes that seem like they'd be related, we'll consider turning it on. Updating the diff.

PR #2411, opened 2016-10-18
[ASLayoutPlayground] Some fixes

george-gw:
* Fixed crash when tapping descender.
* Fixed setting the item to inspect.
* Fixed button states in inspector node.
* Added sliders for spacingBefore, spacingAfter, ascender.

### DIFF
--- a/AsyncDisplayKit/ASLayoutElementInspectorCell.m
+++ b/AsyncDisplayKit/ASLayoutElementInspectorCell.m
@@ -84,7 +84,7 @@ __weak static ASLayoutElementInspectorCell *__currentlyOpenedCell = nil;
 {
   ASStackLayoutSpec *horizontalSpec = [ASStackLayoutSpec horizontalStackLayoutSpec];
   horizontalSpec.children           = @[_buttonNode, _textNode];
-  horizontalSpec.style.flexGrow           = YES;
+  horizontalSpec.style.flexGrow     = 1.0;
   horizontalSpec.alignItems         = ASStackLayoutAlignItemsCenter;
   horizontalSpec.justifyContent     = ASStackLayoutJustifyContentSpaceBetween;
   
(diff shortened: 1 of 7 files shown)


AsyncDisplayKit/ASLayoutElementInspectorCell.m
```diff
@@ -197,23 +222,23 @@ - (void)buttonTapped:(ASButtonNode *)sender
...
+    sender.selected = NO;
     return;
   }
   
 //  NSUInteger currentAlignSelfValue;
 //  NSUInteger nextAlignSelfValue;
 //  CGFloat    newValue;
-  
+
+  sender.selected = !sender.selected;
```
appleguy: Should this actually be toggled for the properties that aren't treated like booleans, like AlignSelf?  This code was written when flexGrow and flexShrink were actual BOOLs, but even with the 0.0 / 1.0 implementation they make sense to toggle in the "on/off" sort of way.  We could make UISwitches as an alternative idea.

george-gw: The toggle existed in descender before, but the behavior was inconsistent:
- Select descender [descender button state: selected]
- Select flexGrow [descender button state: selected]
- Select descender [descender button state: deselected]

Based on the initial code, I figured that the goal was to toggle the option on while it's being edited (and then off when editing is done), as well as keep it on for boolean values that are "YES".

The code that I pushed maintains a more consistent button state for that type of behavior.

george-gw:
@appleguy This is simply more consistent based on the behavior that I understood from the initial code.

The toggle is just a temporary visual representation to show the user what field they're changing, I think it might be useful to some, the design could definitely be nicer, but I doubt that is the purpose of this project at this time :)

I will submit another PR for the Swift playground!

appleguy:
@george-gw awesome, thanks!!  The initial code was extremely rough / never really finished, so I support your change and thoughtful consideration of how it should work.

AsyncDisplayKit/ASLayoutElementInspectorCell.m
```diff
@@ -61,20 +61,30 @@ - (instancetype)initWithProperty:(ASLayoutElementPropertyType)property layoutEle
...
     _textNode2.attributedText = [ASLayoutElementInspectorCell propertyValueDetailAttributedStringForProperty:property withLayoutElement:layoutElement];
     
   }
   return self;
 }
 
+- (void)updateButtonStateForProperty:(ASLayoutElementPropertyType)property withLayoutElement:(id<ASLayoutElement>)layoutElement
+{
+  if (property == ASLayoutElementPropertyFlexGrow) {
+    _buttonNode.selected = layoutElement.style.flexGrow;
```
maicki: Be aware flexGrow and FlexShrink are no boolean values anymore.

george-gw: Agreed, this was just a cut/paste from previous code that was based on the old API.
Since it's already merged, let me know if you would like me to submit another PR to address this.

PR #2412, opened 2016-10-18
Add PINCache subspec for PINRemoteImage.

smyrgl:
Currently the Cocoapods podspec fails to build because of the dependency on PINCache which this change resolves.


garrettmoon:
@smyrgl this isn't quite right, beta 4 had breaking changes, I'm working on a PR right this second.

smyrgl:
@garrettmoon Yeah I noticed, I updated the PR.

garrettmoon:
https://github.com/facebook/AsyncDisplayKit/pull/2414

Feel free to update your PR and I'll close mine, but that method is optional, so we need to check for that. Also, I wouldn't add the version requirement to PINCache subspec, it doesn't need to be defined more than once?

smyrgl:
@garrettmoon No yours is fine, closing mine.

garrettmoon:
Thank you so much for hopping on this quickly! @maicki and I were just commenting on how awesome it is that the community quickly fixes issues we (me) cause!

PR #2413, opened 2016-10-18
[ASLayoutElementStyle] ASLayoutElementStyle improvements

maicki:
- Move `ASLayoutElement` properties to use std::atomic
- Remove some `objc_msgSend` and locking overhead for sizing properties

@appleguy Let's give it a try and let me know

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayoutElement.mm
+++ b/AsyncDisplayKit/Layout/ASLayoutElement.mm
@@ -243,22 +243,33 @@ do {\
 
 #pragma mark - ASLayoutElementStyleSizeHelpers
 
+// We explicitly not call the setter for (max/min) width and height to avoid locking overhead
+
 - (void)setPreferredSize:(CGSize)preferredSize
 {
-  self.width = ASDimensionMakeWithPoints(preferredSize.width);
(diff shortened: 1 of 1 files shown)


appleguy:
@maicki very interesting test, thanks!  This definitely had the intended benefit.  I think we could land this; I'm excited about during further work here, but let's hold off until next week / after the migration is finalized.

+ @Adlai-Holler, @garrettmoon for context.  I modified the layout benchmark to be more discrete / reproducible, by changing the array that specifies view counts to be [100, 100, 100].

Running this benchmark on my iPhone 7 / A10, the cost associated with accessing the Style properties _other than .size, which is still locked_ has disappeared.  Accessing the style object itself also is still locked and is the largest single locking cost.

Images below:
1. Before cost (**73ms**) for properties on the style object (run 2 of 2 was the before / without the patch)
2. After cost (**17ms**) for properties on the style object (run 1 of 2)
2. Cost (**64ms**) (same before / after) of accessing the style object itself, through the lock.

screen shot 2016-10-18 at 11 11 30 am
screen shot 2016-10-18 at 11 14 11 am
screen shot 2016-10-18 at 11 16 53 am

Adlai-Holler:
The CI failures are due to the PINCache thing which was addressed this morning.

PR #2414, opened 2016-10-18
Upgrades to the latest PINRemoteImage

garrettmoon:


Adlai-Holler:
Nice, merging! Let's get any affected diffs rebased.

PR #2416, opened 2016-10-18
[ASCollectionNode] Add waitUntilAllUpdatesAreCommitted Method

Adlai-Holler:
Can't believe I forgot this.

We had previously planned to update the name of this method, but since this particular change is urgent for Pinterest, and since the ideal name might change in light of the recent behavior change of `nodeForRowAtIndexPath:`, this diff does not do that.


Adlai-Holler:
Merging this so we can get it to Pinterest.

PR #2426, opened 2016-10-19
[ASLayoutSpec] Add ASStaticLayoutSpec back without deprecation warning

maicki:
For the transition phase let's add `ASStaticLayoutSpec` back without deprecation warnings.


Adlai-Holler:
I reverted this.

PR #2427, opened 2016-10-19
[ASLayoutElementStyle] Add preferredSize as getter to ASLayoutElementStyle

maicki:
After getting some feedback it seems like the `preferredSize` is read more often than expected. The goal of not adding it as a getter in the first place was to prevent an unexpected behavior if the `width` and `height` of the style is not of dimension points and to give the responsibility to the developer to be sure that the right dimensions are set to the `width` and `height`. Furthermore it should give the developers the opportunity to think what the `width` and `height` `ASDimension` actually means ...

... however this PR add's a getter for `preferredSize` to `ASLayoutElementStyle` and throws an assertion, if `width` or `height` of the `ASLayoutElementStyle` are not of unit points.

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayoutElement.h
+++ b/AsyncDisplayKit/Layout/ASLayoutElement.h
@@ -250,10 +250,11 @@ extern NSString * const ASLayoutElementStyleLayoutPositionProperty;
  * @discussion This method is optional, but one of either preferredSize or preferredLayoutSize is required
  * for nodes that either have no intrinsic content size or 
  * should be laid out at a different size than its intrinsic content size. For example, this property could be 
- * set on an ASImageNode to display at a size different from the underlying image size. 
+ * set on an ASImageNode to display at a size different from the underlying image size.
+ *
+ * @warning Calling the getter when the size's width or height are relative will cause an assert.
  */
 @property (nonatomic, assign) CGSize preferredSize;
-- (CGSize)preferredSize UNAVAILABLE_ATTRIBUTE;
 
  /**
  * @abstract An optional property that provides a minimum size bound for a layout element. If provided, this restriction will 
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Layout/ASLayoutElement.mm
```diff
@@ -254,6 +254,22 @@ - (void)setPreferredSize:(CGSize)preferredSize
...
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  if (_size.width.unit != ASDimensionUnitPoints) {
+    NSCAssert(NO, @"preferredSize.width needs to be in ASDimensionUnitPoints");
+    return CGSizeZero;
+  }
+  
+  if (_size.height.unit != ASDimensionUnitPoints) {
+    NSCAssert(NO, @"preferredSize.height needs to be in ASDimensionUnitPoints");
+    return CGSizeZero;
```
Adlai-Holler: Let's change the copy to make it clear _why_ we had to assert, and what they should do to fix it.

`Cannot get preferredSize of element with fractional height. Element: <ASTextNode: 0xFFFFFFFF>, height: 25%.`

Use the `delegate` property to populate the element string.

AsyncDisplayKit/Layout/ASLayoutElement.h
```diff
@@ -253,7 +253,6 @@ extern NSString * const ASLayoutElementStyleLayoutPositionProperty;
  * set on an ASImageNode to display at a size different from the underlying image size.
```
garrettmoon: Can we add an `@warning calling the getter to this method when the size's width or height are relative will cause an assert.`

maicki: Good suggestion will add

PR #2432, opened 2016-10-19
[Operations] Fix Xcode 7 Build Failing Due to Missing Signposts Header

### LINKED ISSUE #2429: Project Does Not Build on Xcode 7
This should be pretty easy to fix – some of the ksdebug stuff is missing. So we should be able to use `#if __has_include("<sys/kdebug_signpost.h>")` to compile that conditionally.

This is important since Xcode 8 has issues with CI and with syntax highlighting, plus some large-scale Swift apps still haven't migrated believe it or not.

Fixing this may be a precursor to #2418.

Adlai-Holler:
Resolves #2429 . Very simple.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -23,8 +23,6 @@
 #import "ASInternalHelpers.h"
 #import "ASCellNode+Internal.h"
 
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Merging this since it's very low-impact and developing other fixes on Xcode 8 is such a pain.

PR #2438, opened 2016-10-19
Remove Explicit Calls to -copy on Blocks

Adlai-Holler:
I've confirmed that the optimizer does not remove these calls automatically, even though they're unnecessary.

Under ARC, block copying is done automatically when required. So calling `-copy` is purely wasted resources.

http://clang.llvm.org/docs/AutomaticReferenceCounting.html#id50

> With the exception of retains done as part of initializing a `__strong` parameter variable or reading a `__weak` variable, whenever these semantics call for retaining a value of block-pointer type, it has the effect of a `Block_copy`. The optimizer may remove such copies when it sees that the result is used only as an argument to a call.

Details at http://stackoverflow.com/a/23352604

I was also able to make the display completion block static, since it's pure. Not huge, but hey it's a freebie!


Adlai-Holler:
@appleguy @maicki @johnepinterest ready for review!

PR #2445, opened 2016-10-20
Update Flow Layout Inspector Delegate Assertion Test for New Selector

Adlai-Holler:
Follow-up to #2444


Adlai-Holler:
Merging now so that more branches don't get opened with the failing test.

PR #2449, opened 2016-10-21
[ASTableNode] expose ASTableView's -visibleNodes

hannahmbanana:


hannahmbanana:
@Adlai-Holler: another review please? :)

AsyncDisplayKit/ASCollectionNode.mm
```diff
@@ -287,7 +287,8 @@ - (NSInteger)numberOfSections
 
 - (NSArray<__kindof ASCellNode *> *)visibleNodes
 {
-  return [self.view visibleNodes];
+  ASDisplayNodeAssertMainThread();
+  return self.isNodeLoaded ? [self.view visibleNodes] : @[];
```
hannahmbanana: @Adlai-Holler: preference for nil / @[] here?

AsyncDisplayKit/ASTableNode.h
```diff
@@ -287,6 +287,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSInteger)tableNode:(ASTableNode *)tableNode numberOfRowsInSection:(NSInteger)section;
 
 /**
+ * Similar to -visibleCells.
+ *
+ * @return an array containing the nodes being displayed on screen. This must be called on the main thread.
+ */
+- (NSArray<__kindof ASCellNode *> *)visibleNodes AS_WARN_UNUSED_RESULT;
```
maicki: I would suggest we declare it as readonly property.

AsyncDisplayKit/ASCollectionNode.mm
maicki: I think we should return an empty array like `visibleCells` on UITableView does

AsyncDisplayKit/ASTableNode.h
Adlai-Holler: Interesting. UITableView declares `visibleNodes` as a property, `UICollectionView` declares it as a property. I suppose we should just mirror their declarations so that Swift migrations don't have a senseless removal of `()`.

hannahmbanana:
@Adlai-Holler @maicki - needs final review

AsyncDisplayKit/ASCollectionNode.h
```diff
@@ -291,7 +291,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Similar to -visibleCells.
  *
- * @return an array containing the nodes being displayed on screen.
+ * @return an array containing the nodes being displayed on screen. This must be called on the main thread.
  */
 - (NSArray<__kindof ASCellNode *> *)visibleNodes AS_WARN_UNUSED_RESULT;
```
maicki: We should also move this to a readonly property. `UICollectionView` declared it previously as methods but going forward  also as property:
```
// UICollectionView.h
@property (nonatomic, readonly) NSArray<__kindof UICollectionViewCell *> *visibleCells;
@property (nonatomic, readonly) NSArray<NSIndexPath *> *indexPathsForVisibleItems;
```

Adlai-Holler: Ah good catch! They changed it in iOS 10 SDK. Never seen them do that before.

PR #2451, opened 2016-10-21
[ASCollectionNode] add -indexPathsForVisibleItems

hannahmbanana:
Any reason we don't want to expose this on ASCollectionView and pass to the ASCollectionNode?

### DIFF
--- a/AsyncDisplayKit/ASCollectionNode.h
+++ b/AsyncDisplayKit/ASCollectionNode.h
@@ -313,6 +313,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable NSIndexPath *)indexPathForNode:(ASCellNode *)cellNode AS_WARN_UNUSED_RESULT;
 
+/**
+ * Retrieve the index paths of all visible items.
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASCollectionNode.h
```diff
@@ -313,6 +313,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable NSIndexPath *)indexPathForNode:(ASCellNode *)cellNode AS_WARN_UNUSED_RESULT;
 
+/**
+ * Retrieve the index paths of all visible items.
+ *
+ * @return an array containing the index paths of all visible items. This must be called on the main thread.
+ */
+- (NSArray<__kindof NSIndexPath *> *)indexPathsForVisibleItems AS_WARN_UNUSED_RESULT;
```
maicki: It think we should have a readonly property for this:

```
@property(nonatomic, readonly) NSArray<__kindof NSIndexPath *> *indexPathsForVisibleItems;
```

PR #2454, opened 2016-10-21
[Project Config] switch ios-snapshot-test-case back to main repo

romankl:
The latest version of ios-snapshot-test-case (v2.1.4) brought back the "old" deployment target
`s.ios.deployment_target  = '7.0'`.


appleguy:
@romankl cool, thank you!

Issue #2428, opened 2016-10-19
ASControlNode does not call actions in the same order as they were added


tamaskarai:
Hi,

I recently discovered that the ASControlNode class stores the target actions pairs associated with an event in an NSSet structure. Iterating through a NSSet won't guarantee that the selectors will be called in the same order as they were added.
### Use case

I'd like to make sure to call an event handler that logs analytics before all other event handlers are called. UIControl stores the target action pairs in an NSArray, so it is possible to manipulate the order of event handlers.

If you agree with this then I'm happy to implement the changes needed to store target action pairs in an array rather then a set.
### How to test
1. Download example app: https://github.com/tamaskarai/AsyncDisplayKitTestApp
2. Run pod install
3. Run UI tests
4. Check tests: `testUIKitOrder` test will pass, the `testAsyncDisplayKitOrder` will fail, although the actions are added in the same order
### Versions

AsyncDisplayKit 1.9.91

Thanks,
Tamas

Adlai-Holler:
@tamaskarai Thanks for a thorough, clean issue report! We'd be glad if you implemented this change, and ideally added a unit test to ASControlNode to ensure the behavior stays fixed.

Obviously it's important that the targets are not retained – our best bet may be to follow `UIControl`'s model and create e.g. [ASControlTargetAction](https://github.com/nst/iOS-Runtime-Headers/blob/master/Frameworks/UIKit.framework/UIControlTargetAction.h) objects.

tamaskarai:
Hi @Adlai-Holler ,

Thank you for confirming this issue. I'll create a pull request and link this issue report once I'm ready with the implementation. Also, thanks for the implementation tips!

Tamas

Adlai-Holler:
Closed in #2456.

PR #2456, opened 2016-10-21
[ASControlNode] Invoke target actions in the same order as they were added

tamaskarai:
Issue #2428 implementation

In the end I chose to store `ASControlTargetAction` objects in a `NSMutableArray` instead of storing C structs in a `std::vector`.

Let me know if you need me to change anything on the implementation.


tamaskarai:
@hannahmbanana @Adlai-Holler Thank you for your support. Meantime I managed to fix all CI issues.

AsyncDisplayKit/Private/ASControlTargetAction.h
```diff
@@ -0,0 +1,29 @@
...
+
+/** 
+ The action to be called on the registered target.
+ */
+@property (nonatomic, readwrite, assign) SEL action;
+
+/**
+ Event handler target. The specified action will be called on this object.
+ */
+@property (nonatomic, readwrite, strong) id target;
```
Adlai-Holler: We should store this `weak` to prevent retain cycles.

It's important that we find a way to distinguish between a target-action that was created with `nil` target, so it should travel up the responder chain, and one that was created with a target but the target has deallocated. We could use `NSNull` to signify the former, but that's likely to be surprising for future developers. A better move might be to add a `BOOL createdWithNoTarget` flag to make the distinction.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -280,30 +281,51 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
...
+              actionString = NSStringFromSelector(action);
+            
+            if ([NSStringFromSelector(targetAction.action) isEqualToString:actionString])
+              shouldRemove = YES;
+          }
+          
+          if (shouldRemove)
+            continue;
+          
+          [newEventTargetActionArray addObject:targetAction];
```
Adlai-Holler: The logic here is a bit lengthy. Instead, how about overriding `isEqual` in `ASControlTargetAction` and then using `indexOfObject:` or `NSMutableOrderedSet` instead?

AsyncDisplayKit/Private/ASControlTargetAction.h
```diff
@@ -0,0 +1,29 @@
...
+//  This source code is licensed under the BSD-style license found in the
+//  LICENSE file in the root directory of this source tree. An additional grant
+//  of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ @abstract ASControlTargetAction stores target action pairs registered for specific ASControlNodeEvent values.
+ ASControlNode This class stores target action pairs similar to the UIControlTargetAction class.
```
Adlai-Holler: Just to be sure, let's remove the private Apple class name. They don't like those.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -280,30 +281,51 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
...
-        // Create the dispatch table for this event.
-        eventDispatchTable = [NSMapTable weakToStrongObjectsMapTable];
-        if (eventKey) {
-          [_controlEventDispatchTable setObject:eventDispatchTable forKey:eventKey];
+        // Create a target action array for this event.
+        eventTargetActionArray = [[NSMutableArray alloc] init];
+      }
+      else {
+        NSMutableArray *newEventTargetActionArray = [[NSMutableArray alloc] init];
+        NSString *actionString;
```
tamaskarai: @Adlai-Holler Sure, I've fixed it.

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @Adlai-Holler I agree, it's a bit lengthy. Overriding the `isEqual:` is a great idea, although checking the `indexOfObject:` might not work, as it returns an NSUInteger rather than a nullable number, so it's not possible to determine whether the actual target action pair was part of the original set or not. Calling the `containsObject:` then `removeObject:` methods would work but they will check the `isEqual:` twice, so I keep using a foreach cycle and call the `isEquals:` once.

AsyncDisplayKit/Private/ASControlTargetAction.h
tamaskarai: @Adlai-Holler I agree.

AsyncDisplayKit/Private/ASControlTargetAction.h
tamaskarai: @Adlai-Holler I've added the `createdWithNoTarget` BOOL flag and I've removed converting the nil target to NSNull object. Now if the target is nil and the `createdWithNoTarget` flag is true, then the action should travel up in the responder chain.

tamaskarai:
Hi @Adlai-Holler ! I implemented the changes you've requested and answered on your comments. Would you mind reviewing it again?  Thanks!

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -316,13 +329,28 @@ - (NSArray *)actionsForTarget:(id)target forControlEvent:(ASControlNodeEvent)con
...
+  for (ASControlTargetAction *targetAction in eventTargetActionArray) {
+    if (target && targetAction.target && target != targetAction.target) {
+      continue;
+    }
+    
+    if (!target && !targetAction.createdWithNoTarget) {
+      continue;
+    }
+    
+    [actions addObject:NSStringFromSelector(targetAction.action)];
```
Adlai-Holler: This logic is a little tough for my pea-brain to handle. Would this logic be equivalent:

```objectivec
if ((target == nil && targetAction.createdWithNoTarget) || (target != nil && target == targetAction.target)) {
  [actions addObject:NSStringFromSelector(targetAction.action)];
}
```

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -354,44 +382,47 @@ - (void)removeTarget:(id)target action:(SEL)action forControlEvents:(ASControlNo
...
+      NSMutableOrderedSet *newEventTargetActionArray;
+      
+      for (ASControlTargetAction *targetAction in eventTargetActionArray) {
+        BOOL shouldRemove = NO;
+        
+        // Unlike addTarget:, if target is nil here we remove all targets with action.
+        // If target is not nil then check whether it's the same target
+        if (!target || targetAction.target == target) {
+          if (!action) {
+            shouldRemove = YES;
```
Adlai-Holler: Can we simplify this section by using:

```
NSIndexSet *indexes = [eventTargetActionArray indexesOfObjectsPassingTest:];
[eventTargetActionArray removeObjectsAtIndexes:indexes];
```

Note that you can compare `SEL` using `==`. So `targetAction.action == action` is valid.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -280,30 +277,46 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
...
+          }
+          
+          [newEventTargetActionArray addObject:storedTargetAction];
+        }
+        
+        if (!newEventTargetActionArray) {
+          newEventTargetActionArray = [[NSMutableOrderedSet alloc] init];
         }
+        
+        eventTargetActionArray = newEventTargetActionArray;
```
tamaskarai: @Adlai-Holler Yes, I totally forgot that now we're using `NSMutableOrderedSet`, so removing and adding an item will do that same.

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @Adlai-Holler Yep, it's equivalent.

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @Adlai-Holler I ended up using `filterUsingPredicate:`. The result is the same, but I'm happy to rewrite it using your suggestion.

tamaskarai:
@Adlai-Holler Thanks for the feedback on how to further improve the readability of this code. I didn't really use that all operations made on an `NSMutableOrderedSet` makes sure to store the same object only once. Most of the complex logic is now simplified.

tamaskarai:
retest this please

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -278,32 +275,28 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
...
-      {
-        // Nope. Create an action set for it.
-        targetActions = [[NSMutableSet alloc] initWithCapacity:kASControlNodeActionDispatchTableInitialCapacity]; // enough to handle common types without re-hashing the dictionary when adding entries.
-        [eventDispatchTable setObject:targetActions forKey:target];
+      
+      // Remove any prior target-action pair for this event, as UIKit does.
+      [eventTargetActionArray removeObject:targetAtion];
+      
+      // Register the new target-action as the last one to be sent.
+      [eventTargetActionArray addObject:targetAtion];
```
Adlai-Holler: Now that we're manually removing any prior target-action, we can switch back to using `NSMutableArray` which is a little faster.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -278,32 +275,28 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
   _ASEnumerateControlEventsIncludedInMaskWithBlock(controlEventMask, ^
     (ASControlNodeEvent controlEvent)
     {
+      // Create new target action pair
+      ASControlTargetAction *targetAtion = [[ASControlTargetAction alloc] init];
+      targetAtion.action = action;
+      targetAtion.target = target;
```
Adlai-Holler: Just a typo here – `targetAtion`

tamaskarai: @Adlai-Holler Woops, fixed!

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @Adlai-Holler Yes, you're right. There's no need to use `NSMutableOrderedSet` anymore.

Adlai-Holler:
Jenkins, test this please

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -278,32 +275,28 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
   _ASEnumerateControlEventsIncludedInMaskWithBlock(controlEventMask, ^
     (ASControlNodeEvent controlEvent)
     {
+      // Create new target action pair
+      ASControlTargetAction *targetAction = [[ASControlTargetAction alloc] init];
```
Adlai-Holler: Let's save resources, by sharing this targetAction for all events in the mask, by creating the object outside of the enumeration.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -253,10 +254,6 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
   NSParameterAssert(controlEventMask != 0);
   
   ASDN::MutexLocker l(_controlLock);
-
```
Adlai-Holler: We need to update the documentation for `_controlEventDispatchTable`. I think something like:

```objectivec
// ASControlEvent -> [TargetAction0, TargetAction1, ... ]
NSMutableDictionary<id<NSCopying>, NSMutableArray<ASControlTargetAction *> *> *_controlEventDispatchTable;
```

is plenty.

Adlai-Holler:
Note: Tested this thoroughly by hand and it's working great!

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @Adlai-Holler I agree, let's create the object outside the block.

tamaskarai:
Thanks @Adlai-Holler for helping me implement this feature, you're review comments were super useful! We're looking forward to contributing to this awesome product more as our team is going to use AsyncDisplayKit for more features in the future.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -273,37 +259,33 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
...
-      // Have we seen this target before for this event?
-      NSMutableSet *targetActions = [eventDispatchTable objectForKey:target];
-      if (!targetActions)
-      {
-        // Nope. Create an action set for it.
-        targetActions = [[NSMutableSet alloc] initWithCapacity:kASControlNodeActionDispatchTableInitialCapacity]; // enough to handle common types without re-hashing the dictionary when adding entries.
-        [eventDispatchTable setObject:targetActions forKey:target];
+      
+      // Remove any prior target-action pair for this event, as UIKit does.
+      [eventTargetActionArray removeObject:targetAction];
```
appleguy: @tamaskarai it would be better to do removeObjectIdenticalTo:, I think.  Not a critical issue, but this code is much more performance-sensitive than you might think — because all ASImageNodes, ASTextNodes, ASButtonNodes are ASControlNodes and they are invoked heavily during scrolling.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -273,37 +259,33 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
...
+      
+      // Register the new target-action as the last one to be sent.
+      [eventTargetActionArray addObject:targetAction];
+      
+      if (eventKey) {
```
appleguy: Should we early return above or skip the other work, maybe throw an assertion if this is nil?

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -316,13 +298,22 @@ - (NSArray *)actionsForTarget:(id)target forControlEvent:(ASControlNodeEvent)con
...
+  }
 
-  // Return the actions for this target.
-  return [eventDispatchTable objectForKey:target];
+  NSMutableArray *actions = [[NSMutableArray alloc] init];
+  
+  // Collect all actions for this target.
+  for (ASControlTargetAction *targetAction in eventTargetActionArray) {
+    if ((target == nil && targetAction.createdWithNoTarget) || (target != nil && target == targetAction.target)) {
+      [actions addObject:NSStringFromSelector(targetAction.action)];
```
appleguy: I think this would change the performance characteristics of this code quite a bit - won't this method get called on the main thread when event delivery is happening?  It would be better to build this upon adding the action, which may occur on a background thread, instead of always on the main thread.

AsyncDisplayKit/Private/ASControlTargetAction.m
```diff
@@ -0,0 +1,66 @@
...
+
+#import "ASControlTargetAction.h"
+
+@implementation ASControlTargetAction
+{
+  id _target;
+  BOOL _createdWithNoTarget;
+}
+
+- (void)setTarget:(id)target {
```
appleguy: @tamaskarai methods in this file should match the style of having method open braces on the next line.  Control statements have them on the same line though.

AsyncDisplayKit/Private/ASControlTargetAction.m
```diff
@@ -0,0 +1,66 @@
...
+    _createdWithNoTarget = YES;
+  }
+}
+
+- (id)target {
+  return _target;
+}
+
+- (BOOL)isEqual:(id)object {
+  if (![object isKindOfClass:[ASControlTargetAction class]]) {
```
appleguy: This seems pretty complex.  Can you remove some of this logic by starting out with ASObjectIsEqual() ?

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @appleguy As far as I know `removeObject:` calls the `isEqual:` function on each object while `removeObjectIdenticalTo:` just removes the exact same object. I ran tests using `removeObjectIdenticalTo:` instead of `removeObject:` and both the `testDuplicateEntriesWithoutTarget` and `testDuplicateEntriesWithTarget` ASControlNodeTests were failing then. Am I missing something here?

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @appleguy Throwing an assertion here if `eventKey` is nil is a good idea.

AsyncDisplayKit/ASControlNode.mm
tamaskarai: @appleguy If I understand you correctly then you're suggesting to maintain a dictionary where all the actions for a target stored upon adding a target action pair? This could definitely improve the performance of this code.

AsyncDisplayKit/ASControlNode.mm
```diff
@@ -354,44 +345,28 @@ - (void)removeTarget:(id)target action:(SEL)action forControlEvents:(ASControlNo
...
-          if ([targetActions count] == 0)
-          {
-            [eventDispatchTable removeObjectForKey:aTarget];
-
-            // If there are no targets for this event anymore, remove it.
-            if ([eventDispatchTable count] == 0)
-              [_controlEventDispatchTable removeObjectForKey:eventKey];
+      }
+      
+      NSPredicate *filterPredicate = [NSPredicate predicateWithBlock:^BOOL(ASControlTargetAction *_Nullable evaluatedObject, NSDictionary<NSString *,id> * _Nullable bindings) {
```
tamaskarai: @appleguy Yeah, @Adlai-Holler suggested me that there's a function called `indexOfObjectsPassingTest:` which returns the indexes of requested objects. Then the `removeObjectsAtIndexes:` could be used to remove those items. Do you think calling these have a better performance then creating a predicate and call the `filterUsingPredicate:` function?

AsyncDisplayKit/Private/ASControlTargetAction.m
tamaskarai: @appleguy I agree, I'm happy to fix the formatting issues.

AsyncDisplayKit/Private/ASControlTargetAction.m
tamaskarai: @appleguy Sure, I can use `ASObjectsIsEquals()` here to reduce the complexity, thanks for mentioning.

tamaskarai:
@appleguy Thank you for suggesting such great improvements. I'll go through these and implement them. I'll create a PR once I'm ready.

PR #2466, opened 2016-10-24
[ASPagerNode] Implement New ASCollectionDataSource/Delegate Methods

hannahmbanana:


AsyncDisplayKit/ASPagerNode.m
```diff
@@ -14,7 +14,6 @@
 #import "ASDelegateProxy.h"
 #import "ASDisplayNode+Subclasses.h"
 #import "ASPagerFlowLayout.h"
-#import "ASCollectionView+Undeprecated.h"
```
hannahmbanana: does the absence of this line mean that the collectionView methods cannot be called?

Adlai-Holler: Nope, this is good.

Issue #2447, opened 2016-10-20
Add indexPath Property to ASCellNode


Adlai-Holler:
This would be extremely convenient:

```objectivec
/**
 * The current index path of this cell node, or @c nil if this node is
 * not a valid item inside a table node or collection node.
 *
 * @note This property must be accessed on the main thread.
 */
@property (nonatomic, readonly, nullable) NSIndexPath *indexPath;
```

There is already implementation for this, inside the `-propertiesForDebugDescription` method, which finds the owning ASTableNode/ASCollectionNode, and finds the index path for this node. It should be as simple as extracting that logic out into a new method.

We should add a unit test that loads the cell node, asserts that its index path is correct, then deletes an item at a lower index from the collection, asserts that its `item` decreased by one, then deletes the section containing the item and asserts that `indexPath` now returns `nil`.

PR #2468, opened 2016-10-24
[ASCellNode] Added indexPath property.

george-gw:
refs #2447 

[ASCellNode]:
- Added indexPath property.
- Added a test case.

### DIFF
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -108,6 +108,14 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  */
 @property (nonatomic, assign, getter=isHighlighted) BOOL highlighted;
 
+/**
+ * The current index path of this cell node, or @c nil if this node is
+ * not a valid item inside a table node or collection node.
+ *
+ * @note This property must be accessed on the main thread.
+ */
+@property (nonatomic, readonly, nullable) NSIndexPath *indexPath;
+
 /*
  * ASCellNode must forward touch events in order for UITableView and UICollectionView tap handling to work. Overriding
  * these methods (e.g. for highlighting) requires the super method be called.
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -114,7 +114,7 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  *
  * @note This property must be accessed on the main thread.
  */
-@property (nonatomic, readonly, nullable) NSIndexPath *indexPath;
+@property (nonatomic, strong, readonly, nullable) NSIndexPath *indexPath;
 
 /*
  * ASCellNode must forward touch events in order for UITableView and UICollectionView tap handling to work. Overriding
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -116,6 +116,12 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  */
 @property (nonatomic, readonly, nullable) NSIndexPath *indexPath;
 
+/**
+ * The owning node (ASCollectionNode/ASTableNode) of this cell node, or @c nil if this node is
+ * not a valid item inside a table node or collection node or if those nodes are nil.
+ */
+@property (weak, nonatomic, readonly, nullable) ASDisplayNode *owningNode;
+
 /*
  * ASCellNode must forward touch events in order for UITableView and UICollectionView tap handling to work. Overriding
  * these methods (e.g. for highlighting) requires the super method be called.
--- a/AsyncDisplayKit/ASCellNode.h
+++ b/AsyncDisplayKit/ASCellNode.h
@@ -114,7 +114,7 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  *
  * @note This property must be accessed on the main thread.
  */
-@property (nonatomic, strong, readonly, nullable) NSIndexPath *indexPath;
+@property (nonatomic, readonly, nullable) NSIndexPath *indexPath;
 
 /*
  * ASCellNode must forward touch events in order for UITableView and UICollectionView tap handling to work. Overriding
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -40,7 +40,7 @@
   struct {
     unsigned int isTableNode:1;
     unsigned int isCollectionNode:1;
-  } _scrollViewType;
+  } _owningNodeType;
 
 }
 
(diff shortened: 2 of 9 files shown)


AsyncDisplayKit/ASCellNode.h
```diff
@@ -108,6 +108,14 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  */
 @property (nonatomic, assign, getter=isHighlighted) BOOL highlighted;
 
+/**
+ * The current index path of this cell node, or @c nil if this node is
+ * not a valid item inside a table node or collection node.
+ *
+ * @note This property must be accessed on the main thread.
+ */
+@property (nonatomic, readonly, nullable) NSIndexPath *indexPath;
```
maicki: Please add `strong`

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -204,6 +204,26 @@ - (void)__setHighlightedFromUIKit:(BOOL)highlighted;
...
 }
 
+- (NSIndexPath *)indexPath
+{
+  ASDisplayNodeAssertMainThread();
+
+  UIScrollView *scrollView = self.scrollView;
+
+  ASDisplayNode *owningNode = scrollView.asyncdisplaykit_node;
+  if ([owningNode isKindOfClass:[ASCollectionNode class]]) {
```
maicki: The isKindOfClass: calls are pretty heavy ... can we move this logic in setScrollView: and cache via a flags property? Look into `ASCollectionView` / `ASTableView` how we check if delegate methods are implemented.

george-gw: Done

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -204,6 +244,26 @@ - (void)__setHighlightedFromUIKit:(BOOL)highlighted;
...
+
+  UIScrollView *scrollView = self.scrollView;
+
+  ASDisplayNode *owningNode = scrollView.asyncdisplaykit_node;
+  if (_scrollViewType.isTableView) {
+    return [(ASTableView *)owningNode indexPathForNode:self];
+  } else if (_scrollViewType.isTableNode) {
+    return [(ASTableNode *)owningNode indexPathForNode:self];
+  } else if (_scrollViewType.isCollectionView) {
+    return [(ASCollectionView *)owningNode indexPathForNode:self];
```
Adlai-Holler: `scrollView indexPathForNode:`

That said, I think we should actually remove support for ASCollectionView and ASTableView from here, and only work with the ASTableNode and ASCollectionNode. 

The rationale here is,
- The implementation is much simpler.
- The views return index paths in the view-space, which are pretty useless to developers.

AsyncDisplayKit/ASCellNode.h
```diff
@@ -108,6 +108,14 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  */
 @property (nonatomic, assign, getter=isHighlighted) BOOL highlighted;
 
+/**
+ * The current index path of this cell node, or @c nil if this node is
+ * not a valid item inside a table node or collection node.
+ *
+ * @note This property must be accessed on the main thread.
+ */
+@property (nonatomic, strong, readonly, nullable) NSIndexPath *indexPath;
```
Adlai-Holler: No need for `strong` here. Computed properties having no storage qualifier is a pattern e.g. `UICollectionView.visibleCells`.

AsyncDisplayKit/ASCellNode.mm
```diff
@@ -186,6 +194,38 @@ - (void)setHighlighted:(BOOL)highlighted
...
+    _scrollViewType.isTableView      = 0;
+    _scrollViewType.isTableNode      = 0;
+    _scrollViewType.isCollectionView = 0;
+    _scrollViewType.isCollectionNode = 1;
+  } else {
+    _scrollViewType.isTableView      = 0;
+    _scrollViewType.isTableNode      = 0;
+    _scrollViewType.isCollectionView = 0;
+    _scrollViewType.isCollectionNode = 0;
+  }
```
maicki: No need for setting the flags one after another to `0`. Just add at the beginning of the if else block:

```Objc
memset(&_scrollViewType, 0, sizeof(_scrollViewType));
```

 and in the if else set the once to `1`. You can than also remove the else block completely

Adlai-Holler:
The CI failure is spurious and can be ignored.

AsyncDisplayKit/ASCellNode.mm
george-gw: `scrollView` would return ASCollectionView or ASTableView, we need the owningNode which is the corresponding ASCollectionNode or ASTableNode.

Aside from that I removed support for ASCollectionView and ASTableView as requested.

george-gw:
A note: `scrollView`is only non-nil when the cell is displayed:

```
- (void)collectionView:(UICollectionView *)collectionView willDisplayCell:(_ASCollectionViewCell *)cell forItemAtIndexPath:(NSIndexPath *)indexPath
{
  ASCellNode *cellNode = [cell node];
  cellNode.scrollView = collectionView;
```

```
- (void)collectionView:(UICollectionView *)collectionView didEndDisplayingCell:(_ASCollectionViewCell *)cell forItemAtIndexPath:(NSIndexPath *)indexPath
{
  ASCellNode *cellNode = [cell node];
  [...]
  cellNode.scrollView = nil;
```

Should this case be handled in some way? If so, how would you recommend doing that?

Adlai-Holler:
@george-gw You're absolutely right. 

One solution would be to add a weak reference to the owning `ASTableNode/ASCollectionNode` to `ASIndexedNodeContext` (passed into `init`) and to `ASCellNode` (publicly `readonly`, privately `readwrite` in `ASCellNode+Internal.h`). When `ASIndexedNodeContext` creates the node, we can attach that pointer to the cell node. 

I think that should work pretty nicely. Thoughts?

george-gw:
@Adlai-Holler I changed the code to better fit our needs based on your suggestions, I ran all the test cases and they all succeeded, I added a few more checks in the indexPath consistency test to check for nodes that aren't displayed and it seems to work fine.

Adlai-Holler:
test this please

george-gw:
@Adlai-Holler There is a test case for indexPath consistency, it tests different aspects. Do you have a different kind of test you'd like me to write?

Adlai-Holler:
@george-gw Oh those are just the magic words that you use to get our Jenkins CI to run tests. Your test looks good.

george-gw:
@Adlai-Holler oh haha! 👍

Adlai-Holler:
@george-gw Take a look at #2493. Now that `ASCellNode` has `indexPath`, that custom subclass will not automatically synthesize an ivar for it. Could you modify the example code to get Jenkins passing?

george-gw:
@Adlai-Holler which would you prefer? changing the property name to `currentIndexPath` or just doing something like: `@synthesize indexPath=__indexPath;`?

Adlai-Holler:
add to whitelist and test this please

Adlai-Holler:
Thanks @george-gw. Resolves #2447. Here, we, go!

hannahmbanana:
@Adlai-Holler @george-gw: should we rename the `indexPath` property to something more unique? This currently conflicts with Pinterest code.

PR #2470, opened 2016-10-24
[ASStackLayoutSpec] Performance improvements

maicki:
Moves layout calculations of the stack layout spec in the C++ layer of the layout spec implementation. Previously we used references of `ASLayout` to as for specific properties like `layoutElement` and `style` or `size` on the `ASLlayoutElement`.

Now we create a C++ wrapper around the `ASLayout` before the actual layout calculations happening and use this to access the common components to reduce the overhead of Objc message calling.

- Add `ASStackLayoutSpecItem ` and don't separate between unpositioned and positioned items anymore. This didn't work properly previously in the first place as the `ASLayout` that was included was basically shared, so on ASDK side having them was basically nonsense. It would have make sense if we would have create specific `ASLayout` objects for unpositioned and positioned items by copying the old one. In that case the immutability of the `ASLayout` would have make sense.
- Reduce overhead of calling `.style` `.style.size` on `ASLayoutElement`s
- Bypass baseline calculation if not needed

### DIFF
--- a/AsyncDisplayKit/Layout/ASStackLayoutSpec.mm
+++ b/AsyncDisplayKit/Layout/ASStackLayoutSpec.mm
@@ -133,7 +133,10 @@
     return {child, style, style.size};
   });
   
-  ASStackLayoutSpecStyle style = {.direction = _direction, .spacing = _spacing, .justifyContent = _justifyContent, .alignItems = _alignItems, .baselineRelativeArrangement = _baselineRelativeArrangement};
+  const ASStackLayoutSpecStyle style = {.direction = _direction, .spacing = _spacing, .justifyContent = _justifyContent, .alignItems = _alignItems, .baselineRelativeArrangement = _baselineRelativeArrangement};
+  const BOOL needsBaselinePass = _baselineRelativeArrangement ||
+                                 _alignItems == ASStackLayoutAlignItemsBaselineFirst ||
+                                 _alignItems == ASStackLayoutAlignItemsBaselineLast;
   
   const auto unpositionedLayout = ASStackUnpositionedLayout::compute(stackChildren, style, constrainedSize);
   const auto positionedLayout = ASStackPositionedLayout::compute(unpositionedLayout, style, constrainedSize);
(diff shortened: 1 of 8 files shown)


maicki:
@rcancro Hey Ricky, could you please give that a first review pass. Furthermore do know of any other case where we could bypass the baseline pass? Thanks!

AsyncDisplayKit/Layout/ASStackLayoutSpec.mm
```diff
@@ -120,43 +122,62 @@ - (void)setBaselineRelativeArrangement:(BOOL)baselineRelativeArrangement
...
+  // Accessing the style and size property is pretty costly we create layout spec children we use to figure
+  // out the layout for each child
+  const auto stackChildren = AS::map(children, [&](const id<ASLayoutElement> child) -> ASStackLayoutSpecChild {
+    ASLayoutElementStyle *style = child.style;
+    return {child, style, style.size};
+  });
   
-  ASStackLayoutSpecStyle style = {.direction = _direction, .spacing = _spacing, .justifyContent = _justifyContent, .alignItems = _alignItems, .baselineRelativeArrangement = _baselineRelativeArrangement};
-  BOOL needsBaselinePass = _baselineRelativeArrangement || _alignItems == ASStackLayoutAlignItemsBaselineFirst || _alignItems == ASStackLayoutAlignItemsBaselineLast;
+  // TOOD: Does this needs locking?
```
maicki: We should add a lock here for accessing `_direction`, `_spacing`, etc.

AsyncDisplayKit/Layout/ASStackLayoutSpec.mm
```diff
@@ -120,43 +122,62 @@ - (void)setBaselineRelativeArrangement:(BOOL)baselineRelativeArrangement
...
+  // Figure out if a baseline pass is really needed
+  const BOOL directionIsVertical = (style.direction == ASStackLayoutDirectionVertical);
+  const BOOL needsBaselineAlignment = ASStackBaselinePositionedLayout::needsBaselineAlignment(style);
+  const BOOL needsBaselinePositioning = (directionIsVertical == NO || needsBaselineAlignment == YES);
+  
+  NSMutableArray *sublayouts = [NSMutableArray array];
   CGSize finalSize = CGSizeZero;
-  NSArray *sublayouts = nil;
+  if (needsBaselinePositioning) {
+    // Regardless of whether or not this stack aligns to baseline, we need to let ASStackBaselinePositionedLayout::compute find the max ascender
```
rcancro: This comment is a little misleading now. Let's change it to something like:

All horizontal stacks, regardless of whether or not they are baseline aligned, should go through a baseline computation. They could be used in another horizontal stack that is baseline aligned and will need to have computed the proper ascender/descender. 

Vertical stacks do not need to go through this computation since we can easily compute ascender/descender by looking at their first/last child's ascender/descender.

AsyncDisplayKit/Private/ASStackBaselinePositionedLayout.mm
```diff
@@ -13,31 +13,31 @@
...
+      return l.layout.size.height + l.child.style.descender;
     default:
       return 0;
   }
-  
 }
 
 static CGFloat baselineOffset(const ASStackLayoutSpecStyle &style,
-                              const ASLayout *l,
+                              const ASStackLayoutSpecItem &l,
```
rcancro: while we are here, we might as well give this parameter a better name than `l`

AsyncDisplayKit/Private/ASStackBaselinePositionedLayout.mm
```diff
@@ -48,18 +48,27 @@ static CGFloat baselineOffset(const ASStackLayoutSpecStyle &style,
   return 0;
 }
 
-static CGFloat maxDimensionForLayout(const ASLayout *l,
-                                     const ASStackLayoutSpecStyle &style)
+static CGFloat maxDimensionForItem(const ASStackLayoutSpecItem &l,
```
rcancro: `l` -> `item`

AsyncDisplayKit/Private/ASStackBaselinePositionedLayout.mm
```diff
@@ -48,18 +48,27 @@ static CGFloat baselineOffset(const ASStackLayoutSpecStyle &style,
...
+         style.alignItems == ASStackLayoutAlignItemsBaselineLast;
+}
+
 ASStackBaselinePositionedLayout ASStackBaselinePositionedLayout::compute(const ASStackPositionedLayout &positionedLayout,
                                                                          const ASStackLayoutSpecStyle &style,
                                                                          const ASSizeRange &constrainedSize)
 {
+  const auto stackedChildren = positionedLayout.items;
+  
   /* Step 1: Look at each child and determine the distance from the top of the child node it's baseline.
```
rcancro: Can you fix my comment and change `child node it's` to `child node to its`

appleguy:
This does appear to be a significant performance improvement — Great work @maicki !!  This makes a big difference on not only our benchmarks, but also real-world app workloads!

screen shot 2016-10-28 at 8 32 41 pm
screen shot 2016-10-28 at 8 30 13 pm

appleguy:
Benchmark went from:

Old, ASDK: 1.46 ops/sec (100 iterations - worst case for ASDK perf)
New, ASDK: 2.26 ops/sec (100 iterations)
Autolayout: 0.26 ops/sec (100 iterations)
LayoutKit: 2.56 ops/sec (100 iterations)

ASDK has substantial perf gains over LayoutKit in all of the more normal use cases, especially because it does not actually wait on / flush all of the layout at once as this benchmark tests.

In the new version, here are some optimization targets:

screen shot 2016-10-29 at 6 10 43 pm
screen shot 2016-10-29 at 6 06 08 pm
screen shot 2016-10-29 at 6 05 27 pm

AsyncDisplayKit/Layout/ASStackLayoutSpec.mm
```diff
@@ -120,43 +122,66 @@ - (void)setBaselineRelativeArrangement:(BOOL)baselineRelativeArrangement
...
+  NSMutableArray *sublayouts = [NSMutableArray array];
   CGSize finalSize = CGSizeZero;
-  NSArray *sublayouts = nil;
+  if (needsBaselinePositioning) {
+    // All horizontal stacks, regardless of whether or not they are baseline aligned, should go through a baseline
+    // computation. They could be used in another horizontal stack that is baseline aligned and will need to have
+    // computed the proper ascender/descender.
+    
+    // Vertical stacks do not need to go through this computation since we can easily compute ascender/descender by
+    //looking at their first/last child's ascender/descender.
```
appleguy: Space after //

maicki:
@rcancro @appleguy Thanks for reviewing! I will let the variable name for now as it is as we use this name all in the stack layout spec components from the beginning on. We could address it in a follow up PR, it would add some noise that is not really related to the task to this PR.

Will merge as soon as CI finished.

Issue #2471, opened 2016-10-24
Layout Inspector scrollableDirections May Be Called when Not Implemented


Adlai-Holler:
The method was recently added in #2381 . 

![screen shot 2016-10-24 at 9 47 28 am](https://cloud.githubusercontent.com/assets/2466893/19654912/0c270c4c-99cf-11e6-90ba-bafd9dd33c5d.png)
![screen shot 2016-10-24 at 9 47 44 am](https://cloud.githubusercontent.com/assets/2466893/19654913/0c3d45b6-99cf-11e6-999e-330f26bb26a4.png)

```
(lldb) po self->_layoutInspectorFlags.scrollableDirections
<nil>
(lldb) po self->_layoutInspectorFlags.didChangeCollectionViewDataSource
<nil>
(lldb) po self->_layoutInspectorFlags.didChangeCollectionViewDelegate
1
```

garrettmoon:
Ok, I haven't gotten to the bottom of it, but one issue is we have a getter that can generate a new layoutInspector if _layoutInspector is nil. In that case, if the layoutInspector is nil'd out without updating flags, you could check the flags, see it responds to scrollable directions (which wouldn't matter if it was nil because messaging nil should be safe), but then calling the getter would get you a new layoutInspector to message which may have changed its flags.

garrettmoon:
Oooh, ok, _layoutInspector is weak, if it gets nil'd out due to that, the above case could happen.

PR #2473, opened 2016-10-24
[ASCollectionView] layoutInspector can be nil'd at anytime making flags unsafe

garrettmoon:
Hopefully addresses #2471 

I'm actually not sure if even this is safe as layoutInspector could
be nil'd out *after* we check it here…


garrettmoon:
Yeah, that seems better, will update.

garrettmoon:
Thinking about this a bit more, is there any concern about the site effect of creating a layoutInspector if there isn't one on every call to scrollableDirections?

Adlai-Holler:
@garrettmoon We won't – in the getter we will only ever create one default layout inspector.

garrettmoon:
test this please

PR #2478, opened 2016-10-25
[ASDimension] Add ASLayoutSizeAuto

maicki:
Add `ASLayoutSizeAuto` to easier check if the `layoutSize` of an `ASLayoutElement` is set to auto width and height.


maicki:
@Adlai-Holler Could you give that a look please

Issue #2457, opened 2016-10-21
Table & Collection Node Should Enqueue Initial Reload If Needed


Adlai-Holler:
For instance, if you set up a data source and do:

```objectivec
ASTableNode *node = [[ASTableNode alloc] init];
NSInteger sectionCount = node.numberOfSections;
```

You will get 0 back because the initial reload has not been enqueued. This is surprising and inconvenient. There is a property `ASDataController.initialReloadDataHasBeenCalled` that could be used to call `reloadData` when the user accesses these properties so that they return the correct values.

The following methods at the very least should do this:
- numberOfSections
- numberOfRowsInSection
- nodeForRowAtIndexPath

PR #2479, opened 2016-10-26
Table & Collection Node Should Enqueue Initial Reload If Needed

george-gw:
refs #2457

- Call reloadData if necessary when accessing specific methods.
- Added test case.

NB: Since the nodes aren't visible at this time, you won't be able to get all the correct information from the node, for instance:

```
NSIndexPath *indexPath = [NSIndexPath indexPathForItem:0 inSection:0];
ASCellNode *node = [cn nodeForItemAtIndexPath:indexPath];
```

In this case `node.indexPath`will actually be `nil` even if the node itself is correct. Take a look at #2468.

### DIFF
--- a/AsyncDisplayKit/ASCollectionNode.mm
+++ b/AsyncDisplayKit/ASCollectionNode.mm
@@ -337,13 +337,22 @@
 
 #pragma mark - Querying Data
 
+- (void)reloadDataInitiallyIfNeeded
+{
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Such a huge improvement in a small diff. And a unit test too! Nice @george-gw .

Note: Tested with Pinterest and we're looking good.

Issue #2483, opened 2016-10-26
Return CGSizeZero for style.preferredSize if width and height are not of unit type ASDimensionUnitFraction


maicki:
If `width` and `height` of an ASLayoutElement is not of unit type `ASDimensionUnitPoints` we throw an assertion for accessing the `style.preferredSize`.

The initial idea was to push the responsibility that the `preferredSize` was set before accessing it. After using it a bit more in production code we decided to change the logic around accessing it a bit.

Instead of throwing an assertion if the unit type is not `ASDimensionUnitPoints` we will throw an assertion if the unit type is `ASDimensionUnitFraction`. If the unit type is `ASDimensionUnitPoints` or `ASDimensionUnitAuto` we will use the value of the `width` and `height` dimension to create the size to return.

PR #2484, opened 2016-10-26
[Layout] Only throw if width or height are fractions in preferredSize getter

maicki:
If width and height of an `ASLayoutElement` is not of unit type `ASDimensionUnitPoints` we throw an assertion for accessing the `style.preferredSize`.

Instead of throwing an assertion if the unit type is not `ASDimensionUnitPoints` we will throw an assertion if the unit type is `ASDimensionUnitFraction`. If the unit type is `ASDimensionUnitPoints` or `ASDimensionUnitAuto` we will use the value of the `width` and `height` dimension to create the size to return.

Addresses: #2483


maicki:
@Adlai-Holler @appleguy Ready for review

PR #2492, opened 2016-10-27
Workaround iOS 7 Early-Truncation Bug

Adlai-Holler:
When using the new fast path in text node under iOS 7, the text may be truncated on the second-to-last line in some cases.

Adding a small fudge factor to the drawing height (in iOS 7) works around the issue.

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -265,7 +265,12 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
   // If we use default options, we can use NSAttributedString for a
   // fast path.
   if (self.canUseFastPath) {
-    [_attributes.attributedString drawWithRect:shadowInsetBounds options:NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingTruncatesLastVisibleLine context:self.stringDrawingContext];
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -265,7 +265,12 @@ - (void)drawInContext:(CGContextRef)context bounds:(CGRect)bounds;
   // If we use default options, we can use NSAttributedString for a
   // fast path.
   if (self.canUseFastPath) {
-    [_attributes.attributedString drawWithRect:shadowInsetBounds options:NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingTruncatesLastVisibleLine context:self.stringDrawingContext];
+    CGRect drawingBounds = shadowInsetBounds;
+    // Add a fudge-factor to the height, to workaround a bug in iOS 7
+    if (AS_AT_LEAST_IOS8 == NO) {
+      drawingBounds.size.height += 3;
```
garrettmoon: Does this need to be scaled in anyway depending on font size?

Adlai-Holler: I'm not sure. This amount was enough to cover all the common cases in Pinterest. I wish I had time to understand the _reason_ for the inconsistency – considering that the bounding rect we're using was given to us by the exact same attributed string – but alas I don't. =/

PR #2496, opened 2016-10-27
Reduce Traffic to Table/Collection View When Updating

### LINKED ISSUE #2487: reloadData Should Perform the Entire Reload in One Operation
Currently calling `reloadData` causes separate UIKit updates:

```objectivec
[super deleteRows…]
[super deleteSections…]
[super insertSections…]
[super insertRows…]
```

The ideal call pattern is:

```objectivec
[super beginUpdates]
[super deleteSections:{0 - oldSectionCount}];
[super insertSections:{0 - newSectionCount}];
[super endUpdates]
```

This would greatly improve performance and eliminate the possibility that the user sees a frame in-between operations. A similar visual effect can currently achieved by calling `[tableNode reloadData]; [tableNode waitUntilAllUpdatesAreCommitted]`. Since the reload code path is also used during the initial data load, any performance gains made here would apply to all table / collection nodes.

Adlai-Holler:
Resolves #2487.

- In `reloadData`, skip deleting the rows. Deleting the sections is plenty.
- In `reloadData`, batch the deleteSections and insertSections calls together for efficiency.
- When deleting/reloading sections, skip deleting the rows. We know the rows are gone.

I've tested this with Pinterest and it's working very well!

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -636,9 +636,6 @@ NSString * const ASDataControllerRowNodeKind = @"_ASDataControllerRowNodeKind";
 
     // remove elements
     LOG(@"Edit Transaction - deleteSections: %@", sections);
-    NSArray *indexPaths = ASIndexPathsForMultidimensionalArrayAtIndexSet(_editingNodes[ASDataControllerRowNodeKind], sections);
-    
-    [self _deleteNodesAtIndexPaths:indexPaths withAnimationOptions:animationOptions];
     [self _deleteSectionsAtIndexSet:sections withAnimationOptions:animationOptions];
   });
 }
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -624,9 +635,6 @@ - (void)deleteSections:(NSIndexSet *)sections withAnimationOptions:(ASDataContro
 
     // remove elements
     LOG(@"Edit Transaction - deleteSections: %@", sections);
-    NSArray *indexPaths = ASIndexPathsForMultidimensionalArrayAtIndexSet(_editingNodes[ASDataControllerRowNodeKind], sections);
```
appleguy: These two lines being deleted is a happy sight!

Adlai-Holler:
Hell yeah!

PR #2503, opened 2016-10-28
[ASDimension] We need to ensure point dimensions are valid for sizing.

garrettmoon:
It's a programming error to set an infinite height or width with a point value. If we assert here the programmer is more likely to be able to easily address the issue.


garrettmoon:
Looks like we're doing better handling at the ASLayout creation layer in #2505 however I think this is still valid as it throws the error at the source of the issue.

maicki:
@garrettmoon Ok let's get that in as soon as Jenkins finished

PR #2505, opened 2016-10-28
[ASDisplayNode] Proper handling of `constrainedSize`

maicki:
This change will return constrainedSize from ASDisplayNode in calculateSizeThatFits: instead of CGSizeZero. The reason for that is that the constrainedSize usually is within the range of the style.preferredSize. The old behavior did not consider that.

* Change check for valid size in `ASLayout` to check for sizing instead of for layout
* Removes the invalid constrainedSize check within ASNetworkImageNode.
* Fix tests: As ASDisplayNode does not return CGSizeZero anymore we have to give the display nodes we use in tests and are involving a stack spec an intrinsic content size.

Replaces: #2503 and #2389

### DIFF
--- a/AsyncDisplayKitTests/ASDisplayNodeTests.m
+++ b/AsyncDisplayKitTests/ASDisplayNodeTests.m
@@ -2159,7 +2159,7 @@ static bool stringContainsPointer(NSString *description, id p) {
     return [ASWrapperLayoutSpec wrapperWithLayoutElement:displayNode];
   };
   
-  XCTAssertThrows([node layoutThatFits:ASSizeRangeMake(CGSizeMake(0, CGFLOAT_MAX))]);
+  XCTAssertThrows([node layoutThatFits:ASSizeRangeMake(CGSizeMake(0, FLT_MAX))]);
   
   // This dance is necessary as we would assert in case we create an ASDimension that is not real numbers
   ASDimension width = displayNode.style.width;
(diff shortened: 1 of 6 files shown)


AsyncDisplayKitTests/ASDisplayNodeTests.m
```diff
@@ -2151,4 +2151,21 @@ - (void)testThatHavingTheSameNodeTwiceInALayoutSpecCausesExceptionOnLayoutCalcul
...
 
+- (void)testThatLayoutWithInvalidSizeCausesException
+{
+  ASDisplayNode *displayNode = [[ASDisplayNode alloc] init];
+  ASDisplayNode *node = [[ASDisplayNode alloc] init];
+  node.layoutSpecBlock = ^ASLayoutSpec *(ASDisplayNode *node, ASSizeRange constrainedSize) {
+    return [ASWrapperLayoutSpec wrapperWithLayoutElement:displayNode];
+  };
+  
+  XCTAssertThrows([node layoutThatFits:ASSizeRangeMake(CGSizeMake(0, CGFLOAT_MAX))]);
```
garrettmoon: Maybe check with FLT_MAX? ;D

Adlai-Holler:
@maicki Would it be worth adding an assertion in `-[ASDisplayNode calculateSizeThatFits:]`, that the constrained size is valid? It might help debugging. Cannot calculate size of node because constrained size is infinite and node does not override -calculateSizeThatFits:. Try setting style.preferredSize on the node. Node: %@

maicki:
Follow Up: #2507

PR #2506, opened 2016-10-28
Improve Table Performance, Especially When Using Section Index Bar

### LINKED ISSUE #2501: Table View Added Overhead when Using Section Index
When using a section index with a table node, there is extra overhead because the constrained size used to measure the node in advance doesn't match the content size of the resulting cell.

For instance, if the table is 320 wide, and the index is 15 wide, the node will be initially measured against 320 and when it appears on screen it will have to be remeasured against 305. The remeasure also requires a table view -beginUpdates/-endUpdates call which also takes time.

I already have a diff for this that @chrisdanford has tested and it works great!

Adlai-Holler:
This is a diff that I developed with @chrisdanford – it resolves #2501.

- Only call `beginUpdates/endUpdates` if the node's height changed as a result of remeasuring against content view.
- Subtract section index frame from bounds-width when determining the row width. Otherwise, when the cell comes on screen, it's going to have to synchronously remeasure.
- Fix an unrelated issue where if you call `ASTableNode.numberOfSections` before the first layout, it'll reload data with `_nodesConstrainedWidth = 0` and all the rows will get measured against 0 which is less than ideal.

@chrisdanford Would you be willing to download this as a patch and try it against Pinterest to make sure that I haven't regressed anything since the last profiling?

### DIFF
--- a/AsyncDisplayKit/ASTableNode.mm
+++ b/AsyncDisplayKit/ASTableNode.mm
@@ -389,18 +389,22 @@ ASEnvironmentCollectionTableSetEnvironmentState(_environmentStateLock)
 - (void)reloadDataInitiallyIfNeeded
 {
   if (!self.dataController.initialReloadDataHasBeenCalled) {
-    [self reloadData];
+    // Note: Just calling reloadData isn't enough here – we need to
+    // ensure that _nodesConstrainedWidth is updated first.
+    [self.view layoutIfNeeded];
   }
 }
 
 - (NSInteger)numberOfRowsInSection:(NSInteger)section
 {
+  ASDisplayNodeAssertMainThread();
   [self reloadDataInitiallyIfNeeded];
   return [self.dataController numberOfRowsInSection:section];
 }
 
 - (NSInteger)numberOfSections
 {
+  ASDisplayNodeAssertMainThread();
   [self reloadDataInitiallyIfNeeded];
   return [self.dataController numberOfSections];
 }
--- a/AsyncDisplayKit/ASTableNode.mm
+++ b/AsyncDisplayKit/ASTableNode.mm
@@ -388,6 +388,7 @@ ASEnvironmentCollectionTableSetEnvironmentState(_environmentStateLock)
 
 - (void)reloadDataInitiallyIfNeeded
 {
+  ASDisplayNodeAssertMainThread();
   if (!self.dataController.initialReloadDataHasBeenCalled) {
     // Note: Just calling reloadData isn't enough here – we need to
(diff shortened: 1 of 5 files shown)


AsyncDisplayKit/ASTableNode.mm
```diff
@@ -378,18 +378,22 @@ - (void)scrollToRowAtIndexPath:(NSIndexPath *)indexPath atScrollPosition:(UITabl
 - (void)reloadDataInitiallyIfNeeded
 {
   if (!self.dataController.initialReloadDataHasBeenCalled) {
-    [self reloadData];
```
appleguy: Can we put an assertMainThread here?

AsyncDisplayKit/ASTableView.mm
```diff
@@ -1584,6 +1596,31 @@ - (void)clearFetchedData
...
+  UIView *indexView = _sectionIndexView;
+  if (indexView.superview == self) {
+    return indexView.frame.size.width;
+  }
+
+  CGRect bounds = self.bounds;
+  for (UIView *view in self.subviews) {
+    CGRect frame = view.frame;
+    // Section index is right-aligned and less than half-width.
+    if (CGRectGetMaxX(frame) == CGRectGetMaxX(bounds) && frame.size.width * 2 < bounds.size.width) {
```
appleguy: This seems pretty questionable... what about other subviews people may add to a table view, and manage with scrollViewDidScroll?

I think in the future, it would be more ideal to call super layoutSubviews in the layout method and then capture the bounds on a .contentView inside an _ASTableViewCell, which is sized to reflect these things - as well as delete control, resize handle, etc.  This would be tricky to say the least, but would be more UIKit-compliant if possible.

Adlai-Holler: Yeah it's not great.

chrisdanford:
Very obvious improvement when I compare the "AHTablePerformance" and "releases/p6.14" branches.

Steps:
- cold start, wait for CPU to settle
- on the first HF pin, context menu -> Pin
- wait for AddPinVC to load and CPU to settle
- scroll continuously from top to bottom of table view

With iPod 5G:
Before
screen shot 2016-10-28 at 5 28 51 pm

After
screen shot 2016-10-28 at 5 32 53 pm

Adlai-Holler:
Thanks guys! Since this has such a substantial performance improvement (and it adds tests!) I'm going to merge this. In the future I'd love to find a more direct approach, but reading a cell's content width and applying it to all other cells is fraught with other issues so let's keep brainstorming and revisit this later.

PR #2511, opened 2016-10-30
Share a Singleton Shadower for Unshadowed Text, Fix Issue

### LINKED ISSUE #2509: Optimize Text Node by Only Creating Shadowers If Needed
This should be very easy. We create shadowers for every text kit renderer, which can [add up quickly](https://github.com/facebook/AsyncDisplayKit/pull/2470#issuecomment-257126231).

Adlai-Holler:
Resolves #2509 

This also fixes an issue where the shadowing was applied _twice_ – once in ASTextNode.drawRect:withParameters: and again inside ASTextKitRenderer. Along with the small resource penalty, this caused drawing metrics to be wrong in some cases for shadowed text.

Ready for review @appleguy @maicki @nguyenhuy !


appleguy:
@Adlai-Holler This is a great change - profiling definitely did show shadower objects created in significant numbers, sometimes resulting in runtime lock contention.  I don't recall the numbers, but this could be a 1-2% improvement in text handling!

PR #2516, opened 2016-10-31
Invalidate TextNode Renderer Instead of Changing constrainedSize

### LINKED ISSUE #2515: Text Node maximumNumberOfLines May Not Be Honored if Constrained Width Changes from Infinite to Finite
In `ASTextKitRenderer`, we check the current constrained width against infinity, when deciding whether to apply the `maximumNumberOfLines` constraint to the string drawing context used by the fast-path. If the constrained width later becomes finite, we will not check it again and update the property.

This can cause text nodes to go beyond the maximum number of lines.

Adlai-Holler:
Resolves #2515 

Instead of fixing the immediate issue, I decided to make a more substantial change and remove `setConstrainedSize:` from the renderer.

The renderer is meant to be one-shot, and this bug would never have cropped up if we had stuck to that. Changing the constrained size basically gutted the entire renderer anyway, so this change hardly matters.

I've tested this with Pinterest and it resolves the issue nicely.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -398,7 +398,9 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
   
   _constrainedSize = constrainedSize;
   
-  [self _invalidateRenderer];
+  if (_renderer != nil && CGSizeEqualToSize(constrainedSize, _renderer.constrainedSize) == NO) {
+    [self _invalidateRenderer];
+  }
 
   [self setNeedsDisplay];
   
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASTextNode.mm
```diff
@@ -403,9 +398,7 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
   
   _constrainedSize = constrainedSize;
   
-  // Instead of invalidating the renderer, in case this is a new call with a different constrained size,
-  // just update the size of the NSTextContainer that is owned by the renderer's internal context object.
-  [self _renderer].constrainedSize = _constrainedSize;
+  [self _invalidateRenderer];
```
garrettmoon: Can / should this be wrapped in a size changed check?

Adlai-Holler:
Thanks for the review @garrettmoon!

PR #2518, opened 2016-10-31
[ASDisplayNode] Remove thread affinity for _displaysAsynchronously

### LINKED ISSUE #2517: [ASDisplayNode] _displaysAsynchronously is not safe to call off the main thread
We're seeing an assertion in `_displaysAsynchronously` for calling it off the main thread. I believe the assertion is incorrect and we should instead be holding the `__instanceLock__`. Currently, `_pendingLayoutTransitionDidComplete` calls it and that method should be able to run off the main thread I believe.

maicki:
Resolves #2517

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1242,7 +1242,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
  */
 - (BOOL)_displaysAsynchronously
 {
-  ASDisplayNodeAssertThreadAffinity(self);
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1242,7 +1242,6 @@ - (BOOL)displaysAsynchronously
  */
 - (BOOL)_displaysAsynchronously
 {
-  ASDisplayNodeAssertThreadAffinity(self);
   return _flags.synchronous == NO && _flags.displaysAsynchronously;
```
garrettmoon: Needs a lock too, no?

garrettmoon: Nm, this one specifically avoids locking.

PR #2519, opened 2016-10-31
Deprecations should be re-enabled. If there are specific deprecations…

garrettmoon:
… we'd like to hold off on, we should address those specifically.


hannahmbanana:
LGTM!

PR #2520, opened 2016-11-01
[Example] Fix Swift Example

maicki:


examples/Swift/Sample/TailLoadingCellNode.swift
```diff
@@ -38,6 +39,9 @@ final class TailLoadingCellNode: ASCellNode {
   }
 
   override func layoutSpecThatFits(constrainedSize: ASSizeRange) -> ASLayoutSpec {
+    
+    spinner.style.flexBasis = ASDimensionMakeWithPoints(0.0)
```
garrettmoon: Mind educating me on why this needs to be zero points?

maicki: As the node is wrapping a UIView component it cannot provide it's own intrinsic content size, so for flexing we need to give a minimum size to it.

PR #2522, opened 2016-11-01
[ASDisplayNode Fix using UIActivityIndicatorView provided via view block

maicki:
While going through our Swift example I discovered some strange behavior if a UIActivityIndicatorView is wrapped in a node via a view block. If the node is get a specific size via `style.preferredSize` it will just show up as a black box.
After looking into it a bit more it seems it has the same problems as an `UIImageView` and it’s not safe to clear the contents of the layer of the view and recreate it by calling `setNeedsDisplay`.

@appleguy @garrettmoon  @Adlai-Holler  Ready for review


nguyenhuy:
LGTM!

PR #2523, opened 2016-11-01
[Debugging] Fixed asciiArtString.

### LINKED ISSUE #2500: [Debugging] -asciiArtString assertion when reading from bg thread
`-asciiArtString` sometimes depends on reading a property from the view/layer causing an assertion failure when called from a background thread. We should either make the assertion message (below) more clear what's going on or fix this. 

image

george-gw:
I was debugging using asciiArtString but I found some inconsistencies and a crash. This PR fixes that, if it's possible @hannahmbanana could you check if this fixes #2500 since I don't have your test case?


Adlai-Holler:
Beautiful, thanks @george-gw !

PR #2535, opened 2016-11-02
[ASDisplayNode] Consolidate and cleanup subnode hierarchy management

maicki:
Subnode hierarchy management within `ASDisplayNode` uses different paths for adding, modifying the subnode hierarchy. While looking into that it was harder to understand what is actually going on in  each phase.

Let's consolidate and cleanup the situation and have one central place to add and modify the subnode hierarchy.

As it's also a hot path let's remove the overhead of creating a block while dispatching adding / removing views / layers as they should run on the main thread anyway and we have them so there are not problems in production.

@Adlai-Holler @appleguy @nguyenhuy Ready for review

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1641,13 +1641,120 @@ static inline CATransform3D _calculateTransformFromReferenceToTarget(ASDisplayNo
 
 #pragma mark - Managing the Node Hierarchy
 
-static bool disableNotificationsForMovingBetweenParents(ASDisplayNode *from, ASDisplayNode *to)
-{
+ASDISPLAYNODE_INLINE bool shouldDisableNotificationsForMovingBetweenParents(ASDisplayNode *from, ASDisplayNode *to) {
   if (!from || !to) return NO;
   if (from->_flags.synchronous) return NO;
   if (to->_flags.synchronous) return NO;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1641,130 +1641,162 @@ - (void)didDisplayAsyncLayer:(_ASDisplayLayer *)layer
...
-    if (_layer) {
-      ASDisplayNodeCAssertMainThread();
+  // Don't bother inserting the view/layer if in a rasterized subtree, because there are no layers in the hierarchy and
+  // none of this could possibly work.
+  if (nodeIsInRasterizedTree(self) == NO && self.nodeLoaded) {
+    // If node is loaded insert the subnode otherwise wait until the node get's loaded
+    if (ASDisplayNodeThreadIsMain()) {
+      [self _insertSubnodeSubviewOrSublayer:subnode atIndex:sublayerIndex];
+    } else {
+      ASPerformBlockOnMainThread(^{
```
Adlai-Holler: `ASPerformBlockOnMainThread` runs synchronously if we're already on main so we can remove the `if`. Same down in `_removeFromSupernode`.

maicki:
@Adlai-Holler Addressed all comments

Adlai-Holler:
Nice!

PR #2538, opened 2016-11-02
[Layout] Coalescing setNeedsLayout calls

maicki:
## Current Problems

1. Currently `-[ASDisplayNode setNeedsLayout]` changes the node's size.  That's very different from the UIKit-equivalent method; UIKit never changes the view's size whose `-[UIView layoutSubviews]` is being invoked (its subviews might change sizes, but never the view itself, unless you do it yourself explicitly in the subclass). 
2. The behavior in 1) is troublesome because sometimes the the root-node should not resize itself to the size that's returned by its `-[ASDisplayNode layoutThatFits:]` (or `-[UIView sizeThatFits:]` in UIKit is just that) that just about fits the content - that's the definition by UIKit anyway; it can be different from the size that actually want the node to be. If the root node is a full-screen node with subnodes, it can happen that the fullscreen node's frames changes based on the content, but as the intention for  this node was to be a full-screen node this should never happen. If a node should resize to just fit its content, there should be a method like `[UIView sizeToFit]` ; it's not expected that `-[ASDisplayNode setNeedsLayout]` to be doing this job.
3. Calling `-[ASDisplayNode layoutThatFits:]` explicitly before calling `-[ASDisplayNode setNeedsLayout]`, then `-[ASDisplayNode setNeedsLayout]` just discards the old layout and calls `-[ASDisplayNode layoutThatFits:]` on its own again.

##  Planned Changes

1. **The size of the node is expected to be handled / updated by the node's owner / supernode.**
2. `-[ASDisplaynode layoutThatFits:]`  Does not cache any layout anymore. It performs a measurement pass and calculates a layout based on the given constrained size. This will be assigned as `pendingLayout` and returned.
3. `-[ASDisplayNode setNeedsLayout]`  marks the receive's layout as invalid. This allows layout to happen in the next layout pass before the drawing cycle happens.
4. Layout of subnodes are happening in the layout pass (`-[ASDisplayNode __layout]`) based on `-[ASDisplayNode bounds]`  as `constrainedSize`.
5. An `-[ASDisplayNode bounds]` change can has multiple routes:
  1. Can trigger a measurement pass (`-[ASDisplayNode layoutThatFits:]`) in the next layout pass (`-[ASDisplayNode __layout]`) in case the node's bounds does not equal the cached layout's constrained size or the layout was marked invalid before.
  2. Should invalidate the layout of itself if the bounds changed  or was explicitly set on, as well as invalidate the layout of the supernode. This is the behavior of UIKit. By changing the `frame` / `boudns` of a UIView, `layoutSublayersOfLayer` and eventually `layoutSubviews`  will be called on the superview and the view.
6. `-[ASDisplayNode layoutIfNeeded]` should be added  lays out the subnodes immediately and does not wait until the next layout pass is happening.
7. `-[ASDisplayNode sizeToFit]` resizes and moves the receiver node so it just encloses its subnodes.
8. `-[ASDisplayNode invalidateCalculatedLayout]` should be removed (deprecated) as it's basically the same as `-[ASDisplayNode setNeedsLayout]` now.

## TODOS:
- [ ] Splitting up measurement and layout pass
- [ ] Automatic cell sizing in `ASTableView`
- [ ] Automatic cell sizing in `ASCollectionView`
- [ ] Search for `TODO: coalayout:` in code base for open tasks left
- [ ] Add / Adjust unit tests
- [ ] Change sample apps
- [ ] Move layout code to a central place like `ASDisplayNode+Layout.mm`

### DIFF
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -121,9 +121,6 @@ static NSMutableSet *__cellClassesForVisibilityNotifications = nil; // See +init
 {
   CGSize oldSize = self.bounds.size;
   [super __setNeedsLayout];
-    
-  // TODO: coalayout: We have to get rid of that
-  [super __layoutIfNeeded];
   
   //Adding this lock because lock used to be held when this method was called. Not sure if it's necessary for
   //didRelayoutFromOldSize:toNewSize:
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -119,8 +119,11 @@ static NSMutableSet *__cellClassesForVisibilityNotifications = nil; // See +init
 
 - (void)__setNeedsLayout
 {
-  CGSize oldSize = self.calculatedSize;
+  CGSize oldSize = self.bounds.size;
   [super __setNeedsLayout];
+    
+  // TODO: coalayout: We have to get rid of that
+  [super __layoutIfNeeded];
   
   //Adding this lock because lock used to be held when this method was called. Not sure if it's necessary for
   //didRelayoutFromOldSize:toNewSize:
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -755,7 +755,15 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   }
 
   // TODO: coalayout: Cache the layout as pending layout
-  return [self calculateLayoutThatFits:constrainedSize restrictedToSize:self.style.size relativeToParentSize:parentSize];
+  ASLayout *layout = [self calculateLayoutThatFits:constrainedSize restrictedToSize:self.style.size relativeToParentSize:parentSize];
+  _pendingDisplayNodeLayout = std::make_shared<ASDisplayNodeLayout>(
+    layout,
+    constrainedSize,
+    parentSize
+  );
+  
+  // Reuse of already calculated layout is not possible create a new one based on the constrained and parent size
+  return layout;
 }
 
 /**
@@ -864,7 +872,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   __instanceLock__.lock();
   // Setup new constrained and parent size based on bounds
   CGSize parentSize = bounds.size;
-  ASSizeRange constrainedSize = ASSizeRangeMake(parentSize);
+  ASSizeRange constrainedSize = ASSizeRangeMake(parentSize, parentSize);
   
   // Check if we have to create a layout before the layout pass can happen
   CGSize calculatedDisplayNodeLayoutSize = _calculatedDisplayNodeLayout->layout.size;
@@ -891,6 +899,12 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     constrainedSize,
     parentSize
   );
+  
+  calculatedDisplayNodeLayoutSize = _pendingDisplayNodeLayout->layout.size;
+  if (CGSizeEqualToSize(calculatedDisplayNodeLayoutSize, bounds.size) && _pendingDisplayNodeLayout->isDirty() == NO) {
+    pendingLayout = _pendingDisplayNodeLayout;
+  }
+  
 
   _pendingLayoutTransition = [[ASLayoutTransition alloc] initWithNode:self
                                                         pendingLayout:pendingLayout
@@ -2507,6 +2521,7 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
       ASDisplayNodeAssert(layoutSpec.isMutable, @"Node %@ returned layout spec %@ that has already been used. Layout specs should always be regenerated.", self, layoutSpec);
     }
     
+    ASDisplayNodeAssert(layoutSpec.isMutable, @"Node %@ returned layout spec %@ that has already been used. Layout specs should always be regenerated.", self, layoutSpec);
     layoutSpec.parent = self;
     layoutSpec.isMutable = NO;
   }
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -799,6 +799,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
  */
 - (BOOL)__needsLayout
 {
+  ASDN::MutexLocker l(__instanceLock__);
+
   return _calculatedDisplayNodeLayout->isDirty();
 }
 
@@ -811,22 +813,15 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   ASDisplayNodeAssertThreadAffinity(self);
   
   __instanceLock__.lock();
-  
-  // Invalidate the current calculated layout so in the next layout pass it get's recalculated
-  [self invalidateCalculatedLayout];
-  
-  // Walk the node's supernodes until a ancestor node is found that does not require layout
-  if ([_supernode __needsLayout]) {
-    ASDisplayNode *supernode = _supernode;
-    __instanceLock__.unlock();
-    [_supernode __layoutIfNeeded];
-    return;
-  }
-
-  // Layout all subviews starting from the first node that needs layout
-  [self __layout];
-  
+  ASDisplayNode *supernode = _supernode;
   __instanceLock__.unlock();
+
+  if ([supernode __needsLayout]) {
+    [supernode __layoutIfNeeded];
+  } else {
+    // Layout all subviews starting from the first node that needs layout
+    [self __layout];
+  }
 }
 
 /**
@@ -846,38 +841,34 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     return;
   }
 
-  // Node needs to be measured in case the bounds changed from the current calculated layout
+  // Node needs to be measured and calculates a new calculatedLayout if needed in case the bounds changed
   [self calculateLayoutIfNecessaryWithBounds:bounds];
   
   // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
   // was created
   if ([self _shouldHavePlaceholderLayer]) {
     [self _setupPlaceholderLayerIfNeeded];
+    _placeholderLayer.frame = bounds;
   }
-  _placeholderLayer.frame = bounds;
   
-  // Layout all subnodes based on the layout determined before
+  // Layout all subnodes based on the calculatedLayout
   [self layout];
   [self layoutDidFinish];
 }
 
-/*
+/**
  * Executes a measure pass if necessary based on the current bounds as constrained size
  */
 - (void)calculateLayoutIfNecessaryWithBounds:(CGRect)bounds
 {
+  __instanceLock__.lock();
   // Setup new constrained and parent size based on bounds
-  ASSizeRange constrainedSize = ASSizeRangeMake(bounds.size);
-  CGSize parentSize = constrainedSize.max;
+  CGSize parentSize = bounds.size;
+  ASSizeRange constrainedSize = ASSizeRangeMake(parentSize);
   
   // Check if we have to create a layout before the layout pass can happen
-  CGSize calculatedDisplayNodeLayoutSize = CGSizeZero;
-  BOOL needsCalculateLayout = NO;
-  {
-    ASDN::MutexLocker l(__instanceLock__);
-    calculatedDisplayNodeLayoutSize = _calculatedDisplayNodeLayout->layout.size;
-    needsCalculateLayout = [self shouldCalculateLayoutWithConstrainedSize:constrainedSize parentSize:parentSize];
-  }
+  CGSize calculatedDisplayNodeLayoutSize = _calculatedDisplayNodeLayout->layout.size;
+  BOOL needsCalculateLayout = [self shouldCalculateLayoutWithConstrainedSize:constrainedSize parentSize:parentSize];
   
   // Check if we can reuse the calculated layout for the layout pass
   if (CGSizeEqualToSize(calculatedDisplayNodeLayoutSize, CGSizeZero) == NO &&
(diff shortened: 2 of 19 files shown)


AsyncDisplayKit/ASCellNode.mm
```diff
@@ -119,8 +119,11 @@ - (void)layoutDidFinish
 
 - (void)__setNeedsLayout
 {
-  CGSize oldSize = self.calculatedSize;
+  CGSize oldSize = self.bounds.size;
```
appleguy: Do we require this is called on the main thread?  If so, maybe we should assert that.  If not, maybe we should read self.threadsafeBounds.size

maicki: I'm not pretty sure if we have to do that necessarily as the layout pass in `__layout` needs to happen on the main thread. Changed to `_threadSafeBounds` though, as we hold the lock and it's a hot path.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -747,39 +758,30 @@ - (ASLayout *)layoutThatFits:(ASSizeRange)constrainedSize
 #pragma clang diagnostic pop
 }
 
+/**
+ * Calculates a new or reuses the already calculated layout based on given constrained and parent size
```
appleguy: Optional: for in-implementation comments, maybe don't need header-style comment format.
Optional: Later, maybe we should introduce a +Layout.mm category implementation file for Layout and Transition-related code.  Probably not worth doing right now, but it would be exciting to have all this important code tightly located together.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+/**
+ * Returns a BOOL indicating whether the layer has been marked as needing a layout update.
+ */
+- (BOOL)__needsLayout
+{
+  return _calculatedDisplayNodeLayout->isDirty();
+}
+
+/**
+ * The node's supernodes are traversed until a ancestor node is found that does not require layout. Then layout
```
appleguy: This seems like the right traversal behavior to me.  I'm nearly certain that UIKit / CA also do this - that is why they internally have a different method called layoutBelowIfNeeded, since that version doesn't consider upward-traversal.  (we shouldn't add layoutBelow now, hopefully we never need it :) )

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+ * is performed on the entire node-tree beneath that ancestor
+ */
+- (void)__layoutIfNeeded
+{
+  ASDisplayNodeAssertThreadAffinity(self);
+  
+  __instanceLock__.lock();
+  
+  // Invalidate the current calculated layout so in the next layout pass it get's recalculated
+  [self invalidateCalculatedLayout];
```
appleguy: This can now be deleted in your next change

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+  
+  // Walk the node's supernodes until a ancestor node is found that does not require layout
+  if ([_supernode __needsLayout]) {
+    ASDisplayNode *supernode = _supernode;
```
appleguy: Should we create the local supernode pointer outside the if?  I'm not totally sure if this would be net-better (since it would add some retain / release calls, but would reduce weak accesses in the case the branch is taken).  Not critical either way.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+    __instanceLock__.unlock();
+    [_supernode __layoutIfNeeded];
+    return;
+  }
```
appleguy: What do you think, should we make this an if-else clause and remove the early return?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+  ASDisplayNodeAssertMainThread();
+  ASDN::MutexLocker l(__instanceLock__);
+  CGRect bounds = _threadSafeBounds;
+  
+  // Prevent zero bounds subnode adjustments
+  if (CGRectEqualToRect(bounds, CGRectZero)) {
+    // Performing layout on a zero-bounds view often results in frame calculations with negative sizes after applying
+    // margins, which will cause layoutThatFits: on subnodes to assert.
+    LOG(@"Warning: No size given for node before node was trying to layout itself: %@. Please provide a frame for the node.", self);
+    return;
```
appleguy: Note for future iteration: we may want some way to mark "needs layout on bounds change, once" — basically, to re-set the needs layout flag when the bounds becomes nonzero.  Otherwise, it may not match UIKit / CA behavior - I believe those will call layoutSublayers even on a zero-sized item.  It makes sense that we block this because of the ASLayoutSpec's deep reliance on parent / constrained sizes, though.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+
+  // Node needs to be measured in case the bounds changed from the current calculated layout
+  [self calculateLayoutIfNecessaryWithBounds:bounds];
+  
+  // Handle placeholder layer creation in case the size of the node changed after the initial placeholder layer
+  // was created
+  if ([self _shouldHavePlaceholderLayer]) {
+    [self _setupPlaceholderLayerIfNeeded];
+  }
+  _placeholderLayer.frame = bounds;
```
appleguy: Probably not any cause for problems, but should we just put _placeholderLayer.frame = inside the if statement?  Is that sufficient?  It would read more logically even though it would otherwise be a nil-no-op.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+  
+  // Layout all subnodes based on the layout determined before
```
appleguy: Maybe say "set the calculated values on all subnodes based on _pendingLayout" (or, is it _cachedLayout?)  Would be good to name where the layout determined before is stored.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+}
+
+/*
+ * Executes a measure pass if necessary based on the current bounds as constrained size
+ */
+- (void)calculateLayoutIfNecessaryWithBounds:(CGRect)bounds
+{
+  // Setup new constrained and parent size based on bounds
+  ASSizeRange constrainedSize = ASSizeRangeMake(bounds.size);
+  CGSize parentSize = constrainedSize.max;
```
appleguy: I think this would read a bit more logically as:
parentSize = bounds.size;
constrainedSize = SizeMake(parentSize)

Rather than unpacking from the SizeRange to get the input value back.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+  
+  // Check if we have to create a layout before the layout pass can happen
+  CGSize calculatedDisplayNodeLayoutSize = CGSizeZero;
+  BOOL needsCalculateLayout = NO;
+  {
+    ASDN::MutexLocker l(__instanceLock__);
```
appleguy: Optional: Due to being extremely hot path, let's use .lock() and .unlock() as we've found valuable with past profiling of these paths (micro-optimization)

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1400,53 +1532,6 @@ - (void)recursivelyDisplayImmediately
...
-  if (! CGSizeEqualToSize(oldSize, newSize)) {
-    self.bounds = (CGRect){ oldBounds.origin, newSize };
-    
-    // Frame's origin must be preserved. Since it is computed from bounds size, anchorPoint
-    // and position (see frame setter in ASDisplayNode+UIViewBridge), position needs to be adjusted.
-    CGPoint anchorPoint = self.anchorPoint;
-    CGPoint oldPosition = self.position;
-    CGFloat xDelta = (newSize.width - oldSize.width) * anchorPoint.x;
-    CGFloat yDelta = (newSize.height - oldSize.height) * anchorPoint.y;
-    self.position = CGPointMake(oldPosition.x + xDelta, oldPosition.y + yDelta);
```
appleguy: This decomposition into bounds and position is pretty important.  Also, adjusting the position on bounds change is important behavior that can be hard to detect when it's missing.  Is this code supposed to be moved to a new location?

AsyncDisplayKit/Details/_ASDisplayLayer.mm
```diff
@@ -122,6 +122,12 @@ - (void)setNeedsLayout
   ASDisplayNodeAssertMainThread();
   [super setNeedsLayout];
 }
+
+- (void)layoutIfNeeded
+{
+  ASDisplayNodeAssertMainThread();
+  [super layoutIfNeeded];
```
appleguy: Why not also trampoline this one with perform-on-main, if we do for the View class?

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -354,6 +354,34 @@ - (void)setNeedsLayout
   }
 }
 
+- (void)layoutIfNeeded
+{
+  _bridge_prologue_write;
+  BOOL shouldApply = ASDisplayNodeShouldApplyBridgedWriteToView(self);
+  if (shouldApply) {
+    // The node is loaded and we're on main.
+    // Quite the opposite of setNeedsDisplay, we must call __setNeedsLayout before messaging
```
appleguy: For later - comment still mentions setNeedsDisplay

appleguy:
This is one of the most exciting diffs of the year!  Thank you Michael for the great work :)

It's definitely uncertain as to whether this should target the 2.0 release, so we should make sure to focus on issues that are definitely needed for 2.0.  This would be one of the best most valuable stretch goals for the release, but we should decide next week whether it is practical.

AsyncDisplayKit/ASDisplayNode.mm
maicki: In the current behavior the node is changing it's `bounds` within `__setNeedsLayout`. As this is the responsibility of the supernode now, the node will not change the `bounds` of itself anymore, it will happen due to `setFrame:` via the supernode and we do this adjustment of the position in `setFrame:` in `ASDisplayNode+UIViewBridge`.

This is I think one of the most important changes that will happen within this diff as the bounds change was the main difference between the UIKit and ASDK behavior.

AsyncDisplayKit/ASDisplayNode.mm
maicki: Yep will delete.

AsyncDisplayKit/ASDisplayNode.mm
maicki: In general I like early returns, but let's change it to an if-else to be consistent.

AsyncDisplayKit/ASDisplayNode.mm
maicki: Changed

AsyncDisplayKit/ASDisplayNode.mm
maicki: Yep changed to `.lock()` and `.unlock()`

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -796,17 +798,142 @@ - (BOOL)shouldCalculateLayoutWithConstrainedSize:(ASSizeRange)constrainedSize pa
...
+  if (CGSizeEqualToSize(calculatedDisplayNodeLayoutSize, CGSizeZero) == NO &&
+      CGSizeEqualToSize(calculatedDisplayNodeLayoutSize, bounds.size) == YES &&
+      needsCalculateLayout == NO
+      ) {
+    // We can use the calculated layout for the layout subnodes so just return
+    return;
+  }
+  
+  // Cancel layout transition that is in flight
+  [self cancelLayoutTransition];
```
maicki: I agree. I changed to hold the `__instanceLock__`  to the whole method as we also create the layout transition within this method and assign it to an instance method. That said all methods we call within within will grab the lock so there would be some improvements possible.

AsyncDisplayKit/Details/_ASDisplayLayer.mm
maicki: I made `layoutIfNeeded` as well as `setNeedsLayout` safe on `_ASDisplayLayer.m` and `_ASDisplayView.m`. They both trampoline to the main thread now.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
maicki: Changed comment.

AsyncDisplayKit/ASDisplayNode.mm
maicki: The `bounds` change of a note is actually an interesting problem ... if you change the frame of a `UIView` it also invalidates the layout of the superview. We should have the same behavior in this case. The question is just how do we prevent any infinite loops as we set the frames of the subnodes within `__layoutSublayouts` that would invalidate the supernode. I guess we only invalidate the layout if the `bounds` actually changes.

garrettmoon:
test this please

appleguy:
test this please

maicki:
This was the first initial approach for coalescing layout passes and reduce the overhead of `setNeedsLayout` calls. I will close this issue for now and further work is going on within #2597

PR #2539, opened 2016-11-03
[ASControlNode] Implementation needs to match decleration.

garrettmoon:
This could have been… bad. It's amazing that Xcode doesn't get angry about this :(


appleguy:
@tamaskarai :)  Be careful out there!

tamaskarai:
Ohh, it's a bad mistake. Thanks @garrettmoon @appleguy for spotting and fixing this!

Issue #2536, opened 2016-11-02
[Discussion] Solutions to eliminate flashes caused by reloadData


nguyenhuy:
So I’ve been thinking about ways to eliminate, or at least minimize, flashes caused by `reloadData`. This is one of the top complaints from developers that are new to the framework. I have a couple of solutions in mind and would love to start a discussion here:

- Cell node prioritization:
  - I reckon one of the main causes of flashing is due to the fact that new cell nodes are initialized and laid out linearly, meaning they are handled from first to last index. Instead, we should have prioritized them based on the ranges. Nodes in visible range should be handled first, followed by the ones in fetch data range and so on. We can even block main thread while dealing with the highest priority nodes (i.e. visible ones).
  - One possible issue with this technique is that it needs to maintain content offset of the scroll view. In other words, we need to keep a static viewport while deleting/inserting nodes. 
  - This solution maybe easier to implement once virtualization is supported, or maybe a step toward virtualization.

- Another way to solve this problem is hiding external components (esp. the underlying table/collection view) from the fact that `reloadData`is broken down into a series of deletions and insertions. What it means is that, while making changes to the data set, we don't notify `ASDataControllerDelegate` at all. Then once we finished loading all cells, the delegate will be notified about a single atomic reload operation.

I believe either of these solutions can eliminate the flash. Although it will make developers less likely to do batch update :(

cc @Adlai-Holler @maicki @garrettmoon @appleguy

garrettmoon:
@nguyenhuy does @Adlai-Holler 's PR address this? Can it be closed?

nguyenhuy:
@garrettmoon: it does partially address this issue. But I'm having an improvement in mind, will report back this week.

hannahmbanana:
@nguyenhuy @Adlai-Holler - any updates on this issue? Should this be revisited now that we have IGListKit support?

nguyenhuy:
#3017 implements the 2nd approach mentioned above. Flashes are gone now.

shellhue:
i don't think flashed are gone. i create a simple table node with a cell containing just a image node and text node. If i call [tableNode reloadData], sometimes there is no flash, but for half of the time flash happens. I am at the most commit on master branch.

nguyenhuy:
@shellhue Are you seeing flashes at the image and text nodes level? Could you please post a video? There is a chance that we're talking about different kinds of flashes here :)

shellhue:
[...]

PR #2541, opened 2016-11-03
Reduce Flashing & Optimize ASTableNode/ASCollectionNode Data Loading

Adlai-Holler:
Issuing updates to `UICollectionView` is one of the most expensive operations we perform.

This diff combines the section-delete, section-insert, and first-batch-of-rows-insert operations during initial data load & reload data. This will fix flashing in cases where the first batch of items covers the visible area, and it will reduce the total number of UIKit update operations we perform.

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -431,6 +431,15 @@ NSString * const ASDataControllerRowNodeKind = @"_ASDataControllerRowNodeKind";
   dispatch_group_async(_editingTransactionGroup, _editingTransactionQueue, ^{
     LOG(@"Edit Transaction - reloadData");
     
+    /**
+     * Leave the current data in the collection view until the first batch of nodes are laid out.
+     * Once the first batch is laid out, in one operation, replace all the sections and insert
+     * the first batch of items.
+     *
+     * We previously would replace all the sections immediately, and then start adding items as they
+     * were laid out. This resulted in more traffic to the UICollectionView and it also caused all the
+     * section headers to bunch up until the items come and fill out the sections.
+     */
     __block BOOL isFirstBatch = YES;
     [self batchLayoutNodesFromContexts:newContexts batchCompletion:^(NSArray<ASCellNode *> *nodes, NSArray<NSIndexPath *> *indexPaths) {
       if (isFirstBatch) {
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -140,6 +140,11 @@ NSString * const ASDataControllerRowNodeKind = @"_ASDataControllerRowNodeKind";
 #if AS_MEASURE_AVOIDED_DATACONTROLLER_WORK
     [ASDataController _expectToInsertNodes:contexts.count];
 #endif
+  
+  if (contexts.count == 0) {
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -140,6 +140,11 @@ - (void)batchLayoutNodesFromContexts:(NSArray<ASIndexedNodeContext *> *)contexts
 #if AS_MEASURE_AVOIDED_DATACONTROLLER_WORK
     [ASDataController _expectToInsertNodes:contexts.count];
 #endif
+  
+  if (contexts.count == 0) {
+    batchCompletionHandler(@[], @[]);
```
appleguy: Is a comment justified for this one?  I assume this is like zero insertions or deletions, and ultimately calls back to...which code?

If you'd rather the reader follow the code, I trust your judgement.  Just a bit mysterious in a diff review that it needs an early return handler.

Adlai-Holler:
To summarize the nature of this change:

Adlai-Holler:
@nguyenhuy cc #2536 not resolved but relevant. Could you give this a post-commit review?

@garrettmoon Can we get this a lodestone pass when it gets synced to Pinterest?

nguyenhuy:
@Adlai-Holler: Awesome PR! It's a much simpler approach to fix flashes. 

I think we can improve this technique to cover cases in which the visible patch is not the first patch. We can queue up all laid out nodes and insert them all at once when the visible patch is ready. What do you think?

Adlai-Holler:
@nguyenhuy I think that is the ideal approach, but we have an architectural issue to resolve before then – how, given a user that implements UICollectionViewLayout, do we know where that visible patch is?

That is, if the user doesn't get a chance to layout nodes until the nodes are measured and inserted (since UICollectionViewLayout talks directly to UICollectionView), how can we wait until they are laid out to insert them? And how can we let the user drive the measurement process instead of us measuring batches of nodes and THEN handing them to the user to layout?

One good solution might be to create a new protocol such as `ASCollectionNodeLayout` that resembles `UICollectionViewLayout`. The user provides us this object, and we talk to it, and we provide our own `UICollectionViewLayout` that vends data buffered from their async layout.

This new async layout would break the circular dependency (insertion [depends on] layout [depends on] insertion). That is, the async layout would find out about new items before the UICollectionView does. It would stay in sync with the data source, which would be another huge win for reliability, since layouts often talk directly to delegates (such as `collectionView:layout:sizeForItemAtIndexPath`). This async layout object (which may well have to run on the main thread, we'll see) will be responsible for allocating & measuring nodes only as it needs to (which is a huge win for the photo grid use case). 

It would be easy enough for us to ask this new async layout object for `layoutAttributesForElementsInRect:` given the screen + some buffer in case they're scrolling. We can't do this today, because the elements aren't yet inserted in the collection view.

Thoughts @nguyenhuy?

nguyenhuy:
@Adlai-Holler: Yeah, great question on determining visible nodes. I was thinking maybe we can rely on the previous visible paths and assume that the new ones are at least very close. It's definitely the case if items have the same size. I can put up a short working prototype with that assumption.

That being said, your idea is more generalized and should cover all cases. But it looks like a big change. After a discussion with @garrettmoon, @appleguy and @maicki, we agreed that we don't have time to make it to 2.0. @garrettmoon will move #2612 to later milestones.

Issue #2540, opened 2016-11-03
[Examples] CustomCollectionView - constrainedSizeForNodeAtIndexPath is not called while rotation


Baevra:
I'm trying to use custom collectionViewLayout with layoutInspector like in sample app i have similar rotation problem, after two rotation constrainedSizeForNodeAtIndexPath not called in portrait mode
ASDK 2.0 beta 1

hannahmbanana:
@maicki: In case it wasn't clear, the example project collectionViewLayout shows this bug.

hannahmbanana:
@Baevra - Great catch! See the linked PR for an explanation of the issue.

You can resolve the issue by having your layoutInspector implement `- (ASScrollDirection)scrollableDirections`.

Baevra:
@hannahmbanana It works great with scrollableDirections! As always thanks for efficiency!!!

PR #2543, opened 2016-11-03
[ASCollectionViewLayoutInspecting] make -scrollableDirections  required

hannahmbanana:
Thanks to @Baevra for catching this (#2540) and @Adlai-Holler for debugging this with me!

**Issue:**
Rotation from Portrait to Landscape in CustomCollectionView example does not trigger the layout inspector's `-constrainedSizeForNodeAtIndexPath` method. 

**Explanation:**
The collectionView's `scrollableDirections` was getting set to all directions because the bounds size changed before the content size changed. Going from portrait to landscape, the bounds size changed from a smaller width to a larger width, triggering a relayout of all nodes. Going from landscape to portrait, the bounds size changed from a larger width to a smaller width, causing the data controller to think that it was scrollable in the horizontal direction as well. 

Since laying out all nodes is expensive, we only do this if the bounds changes in the non-scrollable direction.  If, for example, a vertical flow layout has its height changed due to a status bar appearance update, we do not need to relayout all nodes. 

**Solution:**
make `-scrollableDirections` a required method for ASCollectionViewLayoutInspecting

- [x] update CustomCollectionView example
- [x] make scrollableDirections required for ASCollectionViewLayoutInspecting object
- [x] remove ASCollectionView's implementation

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1115,27 +1115,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   return direction;
 }
 
-- (ASScrollDirection)scrollableDirections
-{
-  //Cache results of layoutInspector to ensure flags are up to date if getter lazily loads a new one.
-  id<ASCollectionViewLayoutInspecting> layoutInspector = self.layoutInspector;
-  if (_layoutInspectorFlags.scrollableDirections) {
-    return [layoutInspector scrollableDirections];
-  } else {
-    ASScrollDirection scrollableDirection = ASScrollDirectionNone;
-    CGFloat totalContentWidth = self.contentSize.width + self.contentInset.left + self.contentInset.right;
-    CGFloat totalContentHeight = self.contentSize.height + self.contentInset.top + self.contentInset.bottom;
-    
-    if (self.alwaysBounceHorizontal || totalContentWidth > self.bounds.size.width) { // Can scroll horizontally.
-      scrollableDirection |= ASScrollDirectionHorizontalDirections;
-    }
-    if (self.alwaysBounceVertical || totalContentHeight > self.bounds.size.height) { // Can scroll vertically.
-      scrollableDirection |= ASScrollDirectionVerticalDirections;
-    }
-    return scrollableDirection;
-  }
-}
-
 - (ASScrollDirection)flowLayoutScrollableDirections:(UICollectionViewFlowLayout *)flowLayout {
   return (flowLayout.scrollDirection == UICollectionViewScrollDirectionHorizontal) ? ASScrollDirectionHorizontalDirections : ASScrollDirectionVerticalDirections;
 }
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -212,7 +212,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   struct {
     unsigned int didChangeCollectionViewDataSource:1;
     unsigned int didChangeCollectionViewDelegate:1;
-	unsigned int scrollableDirections:1;
   } _layoutInspectorFlags;
 }
 
@@ -515,7 +514,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   _layoutInspectorFlags.didChangeCollectionViewDataSource = [_layoutInspector respondsToSelector:@selector(didChangeCollectionViewDataSource:)];
   _layoutInspectorFlags.didChangeCollectionViewDelegate = [_layoutInspector respondsToSelector:@selector(didChangeCollectionViewDelegate:)];
-  _layoutInspectorFlags.scrollableDirections = [_layoutInspector respondsToSelector:@selector(scrollableDirections)];
 }
 
 - (void)setTuningParameters:(ASRangeTuningParameters)tuningParameters forRangeType:(ASLayoutRangeType)rangeType
@@ -1113,6 +1111,11 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   }
   
   return direction;
+}
+
+- (ASScrollDirection)scrollableDirections
+{
+  return [self.layoutInspector scrollableDirections];
 }
 
 - (ASScrollDirection)flowLayoutScrollableDirections:(UICollectionViewFlowLayout *)flowLayout {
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -79,6 +81,16 @@ - (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSize
   return NodeConstrainedSizeForScrollDirection(collectionView);
 }
 
+- (ASScrollDirection)scrollableDirections
+{
+  if (_delegateFlags.implementsScrollableDirections) {
```
garrettmoon: Is this necessary? Won't compilation fail now when conforming to the protocol?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1117,23 +1115,7 @@ - (ASScrollDirection)_scrollDirectionForVelocity:(CGPoint)scrollVelocity
...
-    CGFloat totalContentHeight = self.contentSize.height + self.contentInset.top + self.contentInset.bottom;
-    
-    if (self.alwaysBounceHorizontal || totalContentWidth > self.bounds.size.width) { // Can scroll horizontally.
-      scrollableDirection |= ASScrollDirectionHorizontalDirections;
-    }
-    if (self.alwaysBounceVertical || totalContentHeight > self.bounds.size.height) { // Can scroll vertically.
-      scrollableDirection |= ASScrollDirectionVerticalDirections;
-    }
-    return scrollableDirection;
-  }
```
hannahmbanana: not 100% sure about about moving this. It seems important.

hannahmbanana:
For reference, it looks like this area was changed recently in #2381

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1117,23 +1115,7 @@ - (ASScrollDirection)_scrollDirectionForVelocity:(CGPoint)scrollVelocity
...
+  return [self.layoutInspector scrollableDirections];
```
appleguy: Might be good to add an AssertNotNil for the self.layoutInspector to ensure this object is always set.  I think the current code does ensure it's always set, but the assertion might help catch any future changes there.

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -79,6 +79,12 @@ - (ASSizeRange)collectionView:(ASCollectionView *)collectionView constrainedSize
   return NodeConstrainedSizeForScrollDirection(collectionView);
 }
 
+- (ASScrollDirection)scrollableDirections
```
appleguy: I see, this is actually in ASCollectionViewLayoutInspector (not Flow) — the file just happens to contain both objects, and is named after the Flow variant instead of the more generic name (kinda weird, but predates this diff).

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1117,23 +1115,7 @@ - (ASScrollDirection)_scrollDirectionForVelocity:(CGPoint)scrollVelocity
 
 - (ASScrollDirection)scrollableDirections
 {
-  //Cache results of layoutInspector to ensure flags are up to date if getter lazily loads a new one.
-  id<ASCollectionViewLayoutInspecting> layoutInspector = self.layoutInspector;
-  if (_layoutInspectorFlags.scrollableDirections) {
-    return [layoutInspector scrollableDirections];
-  } else {
```
appleguy: My main question about this diff is whether this logic below, in the Else clause, should live somewhere — either in the ASCollectionViewLayoutInspector base class, or in the Flow layout version.  Or is it really no longer needed?

nguyenhuy:
Another way to fix this is storing the scroll direction after its first (and only) calculation. Thoughts?

appleguy:
@nguyenhuy could you elaborate?  Doesn't this potentially change as the bounds, contentSize, or contentInset changes?

I'm not too familiar with the core issue that was discovered (that lead to this diff), but it certainly would be nice if scrollableDirections weren't required to be overridden by all subclasses.  If the base class could provide an implementation that works for most cases — like relying on contentSize, which has to be set for scrolling to work properly anyway — that would be great!

nguyenhuy:
@appleguy: I meant we can just cached the directions once computed, since for most layouts, those directions are constants. We can provide a simple mechanism to invalidate the cache if needed: https://github.com/nguyenhuy/AsyncDisplayKit/commit/12aa5f8bdb66d47813b292c5ad829587561b21d6

AsyncDisplayKit/ASCollectionView.mm
garrettmoon: I'm of the opinion the code was dangerous. It's correct most of the time, but not when the bound size has changed and relayout to update contentSize hasn't occurred yet (this is what started us on this path, we were checking scrollable directions to see if we should invalidate all the layout).

AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
garrettmoon: Yeah, we should definitely pull this out separately: https://github.com/facebook/AsyncDisplayKit/issues/2560

garrettmoon:
@nguyenhuy I think it would be simpler to require implementation rather than to require invalidation. Requiring it will make compilation impossible. Requiring invalidation will mean users will need to learn to know to do or suffer with strange bugs.

nguyenhuy:
@garrettmoon to clarify, only layouts that have different directions under different conditions (say, vertical in portrait mode and horizontal in landscape) need to invalidate. 

That's said, after thinking about it again, I agree with you that it's better to require implementation. It's a bit inconvenient to developers but necessary to ensure correctness.

hannahmbanana:
@Adlai-Holler @appleguy @garrettmoon - Adding the assertion causes some of the `ASCollectionView` tests to fail. It looks like the issue is that ASCollectionView's `initWithFrame:` triggers ASCollectionView `didChangeBoundsWithOldValue:`. This method uses the `layoutInspector` getter which relies on the layout (which hasn't been set yet). 

What's the correct solution here? 

screen shot 2016-11-06 at 3 36 46 pm

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1672,6 +1655,9 @@ - (void)didMoveToWindow
  */
 - (void)layer:(CALayer *)layer didChangeBoundsWithOldValue:(CGRect)oldBounds newValue:(CGRect)newBounds
 {
+  if (self.collectionViewLayout == nil) {
+    return;
+  }
```
hannahmbanana: it fixes the tests, but is this the best way?

Adlai-Holler: @hannahmbanana Yes I believe this is perfectly reasonable.

Adlai-Holler:
Nice! Here we go!

PR #2545, opened 2016-11-04
[ASDisplayNode] Trampoline to main to make contents check safe

garrettmoon:
Also took out the check for isMain because ASPerformBlockOnMainThread performs the same check.


maicki:
Maybe we should trampoline the whole `_completePendingLayoutTransition` method. Thoughts?

PR #2547, opened 2016-11-04
[ASLayoutSpecBlock] I know we love blocks but cannot remember them ever

maicki:
Let's add a copy and paste way of an `ASLayoutSpecBlock` to the documentation.


garrettmoon:
Nice :)

appleguy:
Awesome lol, I have never been able to remember blocks either.  Makes me feel better that I'm not alone <3

PR #2550, opened 2016-11-05
[ASTextNode] Ensure that isTruncated computes the correct value when sizing fast-path is used.

### LINKED ISSUE #2504: -[ASTextNode isTruncated] must check truncation manually if sizing fast-path is engaged (or value may be incorrectly NO)
If a text node is rendered using the new fast-path, the `isTruncated` flag will return `NO` even when the text is truncated.

To fix this, we could add an `isTruncated` method onto the renderer. If the renderer is slow-path, then the value that we return today (`firstVisibleRange.length < _attributedText.length`) is returned. For the fast-path case, we can run another sizing pass, with unbounded height, and see if the returned height is greater than the height it returned when bounded. We cache this value in the renderer for future accesses.

connor:
This is an attempt to fix #2504.

@Adlai-Holler what do you think?

@garrettmoon I tried setting the context to be `self.stringDrawingContext` but that didn't work – it only worked as expected when the context is `nil`.

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
@@ -299,6 +299,16 @@ static NSCharacterSet *_defaultAvoidTruncationCharacterSet()
   return lineCount;
 }
 
+- (BOOL)isTruncated
+{
+  if (self.canUseFastPath) {
+    CGRect boundedRect = [_attributes.attributedString boundingRectWithSize:CGSizeMake(_constrainedSize.width, CGFLOAT_MAX) options:NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingTruncatesLastVisibleLine context:nil];
+    return boundedRect.size.height > _constrainedSize.height;
+  } else {
+    return self.firstVisibleRange.length < _attributes.attributedString.length;
+  t }
+}
+
 - (std::vector<NSRange>)visibleRanges
 {
   ASTextKitTailTruncater *truncater = [self truncater];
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -299,6 +299,16 @@ - (NSUInteger)lineCount
   return lineCount;
 }
 
+- (BOOL)isTruncated
+{
+  if (self.canUseFastPath) {
+    CGRect boundedRect = [_attributes.attributedString boundingRectWithSize:CGSizeMake(_constrainedSize.width, CGFLOAT_MAX) options:NSStringDrawingUsesLineFragmentOrigin | NSStringDrawingTruncatesLastVisibleLine context:nil];
```
appleguy: Could you break this method call across a couple lines because of the long arguments?  Aligning on : should look nice :)

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -299,6 +299,16 @@ - (NSUInteger)lineCount
...
+    return boundedRect.size.height > _constrainedSize.height;
```
appleguy: It looks like this check is safe even if attributedString is nil - a CGRectZero will be returned and this should always pass as "nope, not truncated" even if constrainedSize.height is zero.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -299,6 +299,16 @@ - (NSUInteger)lineCount
...
+  } else {
+    return self.firstVisibleRange.length < _attributes.attributedString.length;
+  t }
```
appleguy: It looks like there's a spare character here that needs to be deleted

appleguy:
@connor looks great!  I think we can merge this.  Needs one more commit of formatting cleanup (a spare character tripped up the build, and a stylistic thing)

connor:
Thanks @appleguy! Comments addressed.

Adlai-Holler:
Alright thanks @connor let's take this to the hoop!

PR #2554, opened 2016-11-05
[ASLayoutSpec] Initial commit to support visualizing layout specs (with Playground app).

appleguy:
This code has been waiting on a branch for a few months now.  Although there is a lot that needs to be improved, I think the core concepts are good and should be pulled into master so I don't need to rebase it anymore :).

For a project of this size and scope, with integration points in code that changes frequently, having it in master will allow us to collaborate on the task of making layout visualization a full-fledged feature.  It's not too far from being possible to visualize AND inspect layout specs even in real applications.

Thanks to @hannahmbanana for making the first version of this, and for @nguyenhuy who did the ASLayoutContext integrations.

Some of the core ideas:

- Visualizer Node and Inspector Node must exist in ASDK framework itself, to allow apps to inspect their own layouts in the future.
- Other components that are specific to the Playground app are only in that app.
- Once this and the LayoutSpecExamples PR lands, @hannahmbanana can work on reducing duplication between the apps.  They could share one directory of source files for the demonstration layouts, so the Playground can visualize these and showcase the inspector features without duplicating their code.


appleguy:
This diff include's @george-gw's recent commit to the branch.  Thanks George!  https://github.com/facebook/AsyncDisplayKit/pull/2411

appleguy:
Here's the current state of the UI.  It's actually quite useful to be able to tap around and see the description, as well as all of the most important ASLayoutElementStyle properties show up in the Inspector!

screen shot 2016-11-05 at 6 41 54 pm

AsyncDisplayKit/Debug/ASLayoutElementInspectorNode.m
```diff
@@ -0,0 +1,507 @@
...
+  
+  NSDictionary *attributes = @{NSForegroundColorAttributeName : [UIColor whiteColor],
+                               NSFontAttributeName : [UIFont fontWithName:@"Menlo-Bold" size:12]};
+  headerTitle.attributedText = [[NSAttributedString alloc] initWithString:title attributes:attributes];
+  
+  return headerTitle;
+}
+
+
+//- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)constrainedSize
```
appleguy: This particular file has a bunch of commented-out code, but I'd like to leave it in place for this initial commit.  Like many of our +Beta or +Debug headers, it's still a work in progress.

The most important thing to its success in becoming an official feature is allowing time to focus on improvements rather than the continued complex rebasing it has required.

Fortunately, the value of an Inspector like this isn't speculative - we can be quite confident that this will be one of our most powerful debugging (and educational) features once it is finished.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -60,7 +70,34 @@ - (BOOL)canLayoutAsynchronous
...
 {
-  return self;
+  if (ASLayoutElementGetCurrentContext().needsVisualizeNode && !self.neverShouldVisualize) {
+    return [[ASLayoutSpecVisualizerNode alloc] initWithLayoutSpec:self];
+  } else {
+    return self;
+  }
+}
+
+- (void)recursivelySetShouldVisualize:(BOOL)visualize
```
appleguy: If there is a DEBUG type flag that works in the ASDK framework and is toggled by the default Xcode configuration for most apps (I'm not sure there is)...then we could wrap methods like this to be debug-only.  We can probably clean up the count check at the bottom, but it doesn't pose a risk to production code.

AsyncDisplayKit/Layout/ASLayoutSpec.mm
```diff
@@ -111,8 +148,11 @@ - (void)setChild:(id<ASLayoutElement>)child
 {
   ASDisplayNodeAssert(self.isMutable, @"Cannot set properties when layout spec is not mutable");
   ASDisplayNodeAssert(_childrenArray.count < 2, @"This layout spec does not support more than one child. Use the setChildren: or the setChild:AtIndex: API");
-  
+ 
   if (child) {
+    if ([child isKindOfClass:[ASLayoutSpec class]]) {
```
appleguy: Note to self - this can be changed to adopt the new .layoutElementType.

appleguy:
test this please

appleguy:
Thanks for the review Adlai!  I've triple-checked this and minimized the change on critical paths - I think it's good enough to integrate.  I'd like to continue cleaning it up soon though, and hopefully we can showcase this feature for 2.1.

maicki:
We should create a Github issue to clean it up. We should especially add it behind a flag. We added some unnecessary overhead to it that 99% of time needs not to be run (e.g. `setShouldVisualize` calls). 

Furthermore we should move all of that into it's own category: `ASLayoutSpec (LayoutVisualization)`.

@hannahmbanana could you create an issue for that?

Issue #2528, opened 2016-11-02
[ASPagerNode] implement -indexOfPageWithNode:


hannahmbanana:

lappp9:
https://github.com/facebook/AsyncDisplayKit/pull/2561

Adlai-Holler:
Resolved in #2561

PR #2561, opened 2016-11-07
Added -indexOfPageWithNode: method to ASPagerNode

lappp9:
Decided to knock https://github.com/facebook/AsyncDisplayKit/issues/2528 out
@hannahmbanana


AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1546,8 +1549,6 @@
 		058D0A01195D050800B7D73C /* Private */ = {
 			isa = PBXGroup;
 			children = (
-				9F98C0231DBDF2A300476D92 /* ASControlTargetAction.h */,
-				9F98C0241DBDF2A300476D92 /* ASControlTargetAction.m */,
```
Adlai-Holler: I've seen Xcode 8.1 introduce this, along with moving the `ASDisplayNode+Deprecated.h` change above. I don't know why it insists on doing that, but let's go ahead and land this so that we don't have to keep manually reverting it.

PR #2565, opened 2016-11-07
Include latest version of PINRemoteImage and PINCache

garrettmoon:


garrettmoon:
FYI, PINRemoteImage beta 6 has = (not >=) PINCache beta 2

PR #2566, opened 2016-11-07
Add a Declarative API for Collection Node (standard dataSource still available).

Adlai-Holler:
UICollectionView/UITableView API is a pain to work with. Does the message `Invalid number of items in section 0. The number of items in an existing section after the update (10) must be equal to the number of items before the update (6), plus or minus the number inserted or deleted (3 inserted, 0 deleted)` look familiar?

#### Before:

```objectivec
- (NSInteger)numberOfSectionsInCollectionNode:(ASCollectionNode *)collectionNode
{
  if (self.showRecentSearches) {
    return 2;
  } else {
    return 1;
  }
}

- (NSInteger)collectionNode:(ASCollectionNode *)collectionNode numberOfItemsInSection:(NSInteger)section
{
  if (self.showRecentSearches && section == 0) {
    return self.recentSearches.count;
  } else {
    return self.searchResults.count + (self.searchIsLoading ? 1 : 0); // Account for loading spinner
  }
}

- (ASCellNodeBlock)collectionNode:(ASCollectionNode *)collectionNode nodeBlockForItemAtIndexPath:(NSIndexPath *)indexPath
{
  if (self.showRecentSearches && indexPath.section == 0) {
    EXRecentSearch *s = self.recentSearches[indexPath.item];
    return ^{
      return [EXRecentSearchNode nodeWithSearch:s];
    };
  } else if (self.searchIsLoading) {
    return ^{
      return [[EXLoadingNode alloc] init];
    }
  } else {
    EXSearchResult *r = self.searchResults[indexPath.item];
    return ^{
      return [EXSearchResultNode nodeWithResult:r];
    }
  } else {
    NSAssert(NO, @"Can't provide node block for item at index path: %@", indexPath);
  }
}

- (void)searchDidReceiveResults:(NSArray *)newResults
{
  NSInteger oldResultCount = self.searchResults.count;
  [self.searchResults addObjectsFromArray:newResults];
  NSInteger newResultCount = self.searchResults.count;
  NSInteger resultsSection = self.showRecentSearches ? 1 : 0;
  [self.collectionNode performBatchUpdates:^{
    // Delete the loading spinner.
    [self.collectionNode deleteItemsAtIndexPaths:@[ [NSIndexPath indexPathForItem:oldResultCount inSection:resultsSection];
    NSMutableArray *indexPaths = [NSMutableArray array];
    for (NSInteger i = oldResultCount; i < newResultCount; i++) {
      [indexPaths addObject:[NSIndexPath indexPathForItem:oldResultCount inSection:resultsSection]];
    }
    [self.collectionNode insertItemsAtIndexPaths:indexPaths];
  } completion:nil];
}
```

#### After:

```objectivec
- (ASCollectionData *)dataForCollectionNode:(ASCollectionNode *)collectionNode
{
  ASCollectionData *data = [collectionNode createNewData];
  if (self.showRecentSearches) {
    [data addSectionWithIdentifier:@"recentSearches" block:^(ASCollectionData *data) {
      for (EXRecentSearch *s in self.recentSearches) {
        // Assume namespacedIdentifier is e.g. "EXRecentSearch.18797555"
        [data addItemWithIdentifier:s.namespacedIdentifier nodeBlock:^{
          return [EXRecentSearchNode nodeWithSearch:s];
        }];
      }
    }];
  }
  [data addSectionWithIdentifier:@"searchResults" block:^(ASCollectionData *data) {
    for (EXSearchResult *r in self.searchResults) {
      [data addItemWithIdentifier:r.namespacedIdentifier nodeBlock:^{
        return [EXSearchResultNode nodeWithResult:r];
      }];
    }
    if (self.searchIsLoading) {
      [data addItemWithIdentifier:@"resultsSpinner" nodeBlock:^{
        return [[EXLoadingNode alloc] init];
      }];
    }
  }];
  // There are also APIs for getting more fine-grained control, i.e.
  // getting an NSMutableArray of items/sections that you can do with as you please.
}

- (void)searchDidReceiveResults:(NSArray *)newResults
{
  [self.searchResults addObjectsFromArray:newResults];
  [self.collectionNode performBatchUpdate:nil completion:nil]; // Could be some new method, or -reloadData.
}
```

#### How it Works

If the data source implements the new method, all imperative updates are ignored. Whenever an update is submitted, the method is called. It retrieves a data object that contains all of the existing items and sections in the collection node, by identifier. It configures the data set how it wants it, we do a nested diff against the old data, and we update if needed.

#### Performance Considerations

- The diffing algorithm in this diff has been vastly optimized. I scrolled down through Pinterest's home feed a long way (30 seconds' worth) on my iPhone 6. Using the original diffing implementation, it spent about 1100ms diffing. After my optimizations, the number was around 33ms, which was small compared to the amount of time spent in UIKit machinery. 
- If the user adds an item with an identifier that already existed, then no allocations occur. If they provide the nodeBlock argument inline, then no `Block_copy` happens either. I've added documentation indicating that you will get somewhat improved performance by specifying the node block inline, rather than calling `[self nodeBlockForItemAtRow:]` or similar.
- Users may generate their identifiers lazily i.e. `[NSString stringWithFormat:@"%@.%@", model.class, model.identifier]`. This is unfortunate, but acceptable – string allocations are crazy fast. For cases with extremely high item counts and extremely common updates, we can advise users to store their identifiers, which really isn't hard.
- In the future, we might do some or all of the diffing asynchronously, but the cost-benefit doesn't justify that at this time.
  - We would have to be sure to wait for the update to complete if the user called `nodeForItemAtIndexPath:` or `numberOfItemsInSection` or `indexPathForNode:` or really any ASCollectionNode API.
  - We should call `dataForCollectionNode:` on the main thread because users are going to use main-thread-affine view controller flags, and their upstream data sources are main-thread-affine as well.

### Further Considerations

- How should a user trigger an update?
  - Option 1: Calling `reloadData` or ending a batch update always triggers the incremental update.
  - Option 2: A new method e.g. `performUpdateWithCompletion:` triggers the incremental update, and `reloadData` always invalidates everything.
  - Advocacy: Option 1.
    - Users are trained to throw `reloadData` when they aren't sure how to commit an update.
    - I can't think of a use case for "throw away all the nodes that exists for all the identifiers and then reload." I'm sure one exists out there, but this could be achieved by another means (see below).
- How should a user indicate they want to invalidate the existing node for a given identifier/section?
  - Consider the use case: a user taps on a pin, and now we want to change that pin to be represented by a different class of node.
  - If this method exists, it should live on the collection node itself, not on the data. They should be able to call it inside of `dataForCollectionNode:` as well as in arbitrary regions of their code , which might be slightly tricky.
  - Option 1: They can't. If you want to get a new node, change the identifier.
  - Option 2: Add a method like `invalidateElementWithIdentifier:` to ASCollectionNode that will pipe the invalidation through to both the current data, and to the data that is currently being configured.
- How should we gather the constrained size? What if they give us a different constrained size for an already-measured node?
  - I believe the correct solution is for the layout inspector (or the asynchronous layout object that will likely replace it) to decide the constrained size on its own.
  - If that object wants to reach out to the data source or delegate (through a separate protocol a la `UICollectionViewDelegateFlowLayout`) to get advisement about the constrained size of an element, then it can. ASDK doesn't need to be involved.
  - For today, I advocate for leaving the constrained-size gathering system as-is. It is sufficient for our purposes, it is likely to change, and in my opinion, it is very unlikely to require ASDK to gather constrained sizes for all items upfront.

#### Wrapping Up

This diff also updates the examples to use the new API. They all run nicely. Once this API is approved, I'll apply it also to ASTableNode. Please be generous with your feedback!
@maicki @garrettmoon @appleguy @nguyenhuy @levi @hannahmbanana @rcancro @Yue-Wang-Google @george-gw

### DIFF
--- a/AsyncDisplayKit/ASCollectionData.h
+++ b/AsyncDisplayKit/ASCollectionData.h
@@ -1,0 +1,131 @@
+//
+//  ASCollectionItem.h
+//  AsyncDisplayKit
+//
+//  Created by Adlai Holler on 11/4/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#pragma once
+#import <Foundation/Foundation.h>
+#import <AsyncDisplayKit/ASDimension.h>
+
+@class ASCellNode;
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef NSString * ASSectionIdentifier;
+typedef NSString * ASItemIdentifier;
+typedef NSString * ASSupplementaryElementKind;
+
+/**
+ * ASCellNode creation block. Used to lazily create the ASCellNode instance for an item.
+ */
+typedef ASCellNode * _Nonnull(^ASCellNodeBlock)();
+
+@protocol ASCollectionItem <NSObject>
+
+/**
+ * The identifier for the item.
+ */
+@property (nonatomic, strong, readonly) ASItemIdentifier identifier;
+
+@end
+
+@protocol ASCollectionSection <NSObject>
+
+/**
+ * The identifier for the section.
+ */
+@property (nonatomic, strong, readonly) ASSectionIdentifier identifier;
+
+/**
+ * The items in the section. You can use this property to get fine-grained control over the collection data.
+ */
+@property (nonatomic, strong, readonly) NSMutableArray<id<ASCollectionItem>> *mutableItems;
+
+@end
+
+/**
+ * An object used to construct the data set for a collection
+ * or table node. 
+ * When the collection needs to update its data, it calls @c dataForCollectionNode:
+ * on the data source. The data source calls @c createNewData on the collection node,
+ * configures the data object, and returns it.
+ */
+@interface ASCollectionData : NSObject
+
+/**
+ * Appends a section to the collection.
+ *
+ * @param identifier The identifier for the section.
+ * @param block A block in which to configure the section.
+ *
+ * @warning It is an error to call this method with an identifier that is associated with another section.
+ */
+- (void)addSectionWithIdentifier:(ASSectionIdentifier)identifier
+                           block:(__attribute((noescape)) void(^)(ASCollectionData * data))block;
+
+/**
+ * Adds an item to the current section. If this method is called outside of an addSectionWithIdentifier:
+ * block, a default section will be created and used.
+ *
+ * @param identifier The identifier for the new item.
+ * @param nodeBlock A block that will be used to construct the node for the item.
+ *
+ * @note If an item already exists with this identifier, the node block will be ignored.
+ * @note It is more performant to specify the node block inline here, rather than calling a method that generates it, but the difference is not prohibitive.
+ */
+- (void)addItemWithIdentifier:(ASItemIdentifier)identifier
+                    nodeBlock:(ASCellNodeBlock)nodeBlock;
+
+/**
+ * Adds a supplementary element to the current section. If this method is called outside of an addSectionWithIdentifier:
+ * block, a default section will be created and used.
+ *
+ * @param elementKind The kind of supplementary element to add.
+ * @param identifier The identifier for the new item.
+ * @param index The index for the supplementary element. This can be @c NSNotFound.
+ * @param nodeBlock A block that will be used to construct the node for the supplementary element.
+ *
+ * @note If an element already exists with this identifier, the node block will be ignored.
+ * @note It is more performant to specify the node block inline here, rather than calling a method that generates it, but the difference is not prohibitive.
+ */
+- (void)addSupplementaryElementOfKind:(NSString *)elementKind
+                       withIdentifier:(ASItemIdentifier)identifier
+                                index:(NSInteger)index
+                            nodeBlock:(ASCellNodeBlock)nodeBlock;
+
+/**
+ * Finds or creates an item with the given identifier.
+ *
+ * @param identifier The identifier for the item.
+ * @param nodeBlock A block that will be used to construct the node for the item.
+ *
+ * You can use this method to get more fine-grained control over the collection data.
+ * The returned item can be inserted into the @c mutableItems of an @c ASCollectionSection.
+ *
+ * Note that if an item already exists with this identifier, the node block will be ignored.
+ * @note It is more performant to specify the node block inline here, rather than calling a method that generates it, but the difference is not prohibitive.
+ */
+- (id<ASCollectionItem>)itemWithIdentifier:(ASItemIdentifier)identifier
+                                 nodeBlock:(ASCellNodeBlock)nodeBlock;
+
+/**
+ * Finds or creates a section with the given identifier.
+ *
+ * @param identifier The identifier for the section.
+ *
+ * You can use this method to get more fine-grained control over the collection data.
+ * The returned section can be inserted into @c mutableSections.
+ */
+- (id<ASCollectionSection>)sectionWithIdentifier:(ASSectionIdentifier)identifier;
+
+/**
+ * The sections in the collection. You can use this property to get fine-grained control over the collection data.
+ */
+@property (nonatomic, strong, readonly) NSMutableArray<id<ASCollectionSection>> *mutableSections;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/AsyncDisplayKit/ASCollectionNode.mm
+++ b/AsyncDisplayKit/ASCollectionNode.mm
@@ -22,6 +22,9 @@
 #import "ASSectionContext.h"
 #import "ASCollectionDataController.h"
 #import "ASCollectionView+Undeprecated.h"
+#import "ASDisplayNodeInternal.h"
+#import "NSArray+Diffing.h"
+#import "ASCollectionDataInternal.h"
 
 #pragma mark - _ASCollectionPendingState
 
@@ -457,6 +460,16 @@
 {
   ASDisplayNodeAssertMainThread();
   return [self.dataController contextForSection:section];
+}
+
+- (ASCollectionData *)createNewData
+{
+  ASCollectionData *data = [[ASCollectionData alloc] initWithReusableContentFromCompletedData:self.dataController.previousData];
+  __weak __typeof(self) weakSelf = self;
+  data.postNodeBlock = ^(ASCellNode *node) {
+    [weakSelf.view didCreateNode:node];
+  };
+  return data;
 }
 
 #pragma mark - Editing
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -26,6 +26,8 @@
 #import "ASCollectionViewLayoutFacilitatorProtocol.h"
 #import "ASSectionContext.h"
 #import "ASCollectionView+Undeprecated.h"
+#import "ASCollectionInternal.h"
+#import "ASCollectionDataInternal.h"
 
 /**
  * A macro to get self.collectionNode and assign it to a local variable, or return
@@ -226,6 +228,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     unsigned int numberOfSectionsInCollectionNode:1;
     unsigned int collectionNodeNumberOfItemsInSection:1;
     unsigned int collectionNodeContextForSection:1;
+    unsigned int dataForCollectionNode:1;
   } _asyncDataSourceFlags;
   
   struct {
@@ -233,8 +236,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     unsigned int didChangeCollectionViewDelegate:1;
   } _layoutInspectorFlags;
 }
-
-@property (nonatomic, weak)   ASCollectionNode *collectionNode;
 
 @end
 
@@ -414,6 +415,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     _asyncDataSource = asyncDataSource;
     _proxyDataSource = [[ASCollectionViewProxy alloc] initWithTarget:_asyncDataSource interceptor:self];
     
+    _asyncDataSourceFlags.dataForCollectionNode = [_asyncDataSource respondsToSelector:@selector(dataForCollectionNode:)];
     _asyncDataSourceFlags.collectionViewNodeForItem = [_asyncDataSource respondsToSelector:@selector(collectionView:nodeForItemAtIndexPath:)];
     _asyncDataSourceFlags.collectionViewNodeBlockForItem = [_asyncDataSource respondsToSelector:@selector(collectionView:nodeBlockForItemAtIndexPath:)];
     _asyncDataSourceFlags.numberOfSectionsInCollectionView = [_asyncDataSource respondsToSelector:@selector(numberOfSectionsInCollectionView:)];
@@ -428,11 +430,33 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     _asyncDataSourceFlags.collectionNodeNodeForSupplementaryElement = [_asyncDataSource respondsToSelector:@selector(collectionNode:nodeForSupplementaryElementOfKind:atIndexPath:)];
 
 
-    ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNumberOfItemsInSection || _asyncDataSourceFlags.collectionViewNumberOfItemsInSection, @"Data source must implement collectionNode:numberOfItemsInSection:");
-    ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNodeBlockForItem
-                        || _asyncDataSourceFlags.collectionNodeNodeForItem
-                        || _asyncDataSourceFlags.collectionViewNodeBlockForItem
-                        || _asyncDataSourceFlags.collectionViewNodeForItem, @"Data source must implement collectionNode:nodeBlockForItemAtIndexPath: or collectionNode:nodeForItemAtIndexPath:");
+    // Make assertions about what they implement, unless they are suppressed.
+    if (self.test_suppressCallbackImplementationAssertions == NO) {
+      // If they implement the functional method, they should not implement the classic ones.
+      if (_asyncDataSourceFlags.dataForCollectionNode) {
+        SEL legacySelectors[] = {
+          @selector(numberOfSectionsInCollectionNode:),
+          @selector(numberOfSectionsInCollectionView:),
+          @selector(collectionNode:nodeForItemAtIndexPath:),
+          @selector(collectionView:nodeForItemAtIndexPath:),
+          @selector(collectionNode:nodeBlockForItemAtIndexPath:),
+          @selector(collectionView:nodeBlockForItemAtIndexPath:),
+          @selector(collectionNode:numberOfItemsInSection:),
+          @selector(collectionView:numberOfItemsInSection:),
+          @selector(collectionNode:nodeForSupplementaryElementOfKind:atIndexPath:),
+          @selector(collectionView:nodeForSupplementaryElementOfKind:atIndexPath:)
+        };
+        for (SEL s : legacySelectors) {
+          ASDisplayNodeAssert([_asyncDataSource respondsToSelector:s] == NO, @"If data source implements dataForCollectionNode:, it should not implement %@", NSStringFromSelector(s));
+        }
+      } else {
+        ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNumberOfItemsInSection || _asyncDataSourceFlags.collectionViewNumberOfItemsInSection, @"Data source must implement collectionNode:numberOfItemsInSection:");
+        ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNodeBlockForItem
+                            || _asyncDataSourceFlags.collectionNodeNodeForItem
+                            || _asyncDataSourceFlags.collectionViewNodeBlockForItem
+                            || _asyncDataSourceFlags.collectionViewNodeForItem, @"Data source must implement collectionNode:nodeBlockForItemAtIndexPath: or collectionNode:nodeForItemAtIndexPath:");
+      }
+    }
   }
   
   super.dataSource = (id<UICollectionViewDataSource>)_proxyDataSource;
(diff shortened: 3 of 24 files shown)


Adlai-Holler:
On second thought about "how to trigger updates," I change my stance and we should go option 2 because `reloadData` usually means "update without animation." This update will have animation. The current implementation in this patch does reloads the old way – to trigger the incremental update, just submit any update to the collection view.

chrisdanford:
Looking awesome!  Some thoughts:

- Instead of ASCollectionNode asking for new ASCollectionData synchronously on Main, would it make any more sense to have either a setter interface (`setCollectionData:` or `setCollectionData:complete:`) or some kind of continuation convention like `computeDataForCollectionNode:completeBlock:`?  Here's a situation where I think this might be useful:`dataForCollectionNode:` methods seem like they'll inevitably be too slow to run on main for collections of > 1,000 items.  If `dataForCollectionNode:` is synchronous and on Main, application code may try to precompute ASCollectionData values off of Main ahead of the `dataForCollectionNode:` call.  However, ASCollectionNode's ability to ingest new data is throttled all the UIKit stuff that needs to happen.  I worry that applications will precompute new ASCollectionData values on on small changes but that many of the precomputed values will go unused because ASCollectionNode isn't ready to consume them before the next change in data.  A setter interface wouldn't prescribe where/when the computation of ASCollectionData needs to happen, or the continuation interface would ensure that ASCollectionData is computed only when it's about to be used but still allow for that computation to be off of Main.

> Users are trained to throw reloadData when they aren't sure how to commit an update

- Would it make sense to remove the ability to call `reloadData`?  I would think that the entire point of diffing would be to discourage/disallow manual edits/reloads because they're error prone or lead to sloppy patterns like `reloadData` or `reloadSection` being called for small changes.
- Would it make any more sense for this to be a wrapper around an ASCollectionNode?  Only thinking that this seems like a clean layer on top an IndexPath-based interface, and if you want to disallow manual edits/reloads then hiding the methods entirely seems like a friendlier interface than making methods behave differently depending on whether it's in "declarative mode".  

> Option 2: Add a method like invalidateElementWithIdentifier: to ASCollectionNode that will pipe the invalidation through to both the current data, and to the data that is currently being configured.

- I worry that allowing such a pattern defeats the purpose of diffing.  Option 1 sounds much better to me because I feel like Option 2 will be misused.

Adlai-Holler:
Test this please

Adlai-Holler:
Test this please

Adlai-Holler:
@chrisdanford Thanks for the feedback!

- I absolutely want to avoid a setter interface, for the same reason that React and others do. Usually _one_ object knows how the screen should look, and so we should always ask that object when we need an update. Otherwise developers will have to wire up all that on their side, which will get ugly.
- A continuation interface is interesting but for the reasons that I mentioned in "Performance Considerations" our updates must be synchronous. I know it's a hard pill to swallow but otherwise we simply cannot ensure that we're in sync with the user's data source.
- Good point that we need to adjust the API or add assertions to prevent users from doing what you described and attempting to pre-compute data and return the.
  - I considered making the API something like `collectionNode:configureNewData:` and pass in the data object. I couldn't think of a great name for it so I bailed, but as I type it out right now it sounds great. Thoughts?
- As I mentioned in "Performance Considerations" this solution is performant even for thousands of items. With more detail:
  - For items that existed in the previous data, `addItemWithIdentifier:` is:
    - 1x `[NSDictionary objectForKey:]`
    - 1x `[NSMutableArray addObject:]`
    - 1x `[NSMutableDictionary setObject:forKey:]`
  - For new items, `addItemWithIdentifier:` is that, plus:
    - 1x `[ASCollectionItem alloc] init]`
    - 1x `Block_copy`
- I'm interested in disallowing `reloadData` or having it go the diffing route when possible. I think the latter is the way to go.
- I don't think we should make this a wrapper – I think this should be the _de facto_ way to use ASCollectionNode. If we can't make it fast and flexible enough for that, then we should simply abandon it.
- (On the problems with `invalidateItemWithIdentifier:`) Good point. We could roll with this as-is, and if we discover a use case that requires it we can revisit.

george-gw:
I think the solution is super useful. @Adlai-Holler, thanks for introducing this!

In order to remove the need for `performBatchUpdates`, we still have to deal with the `update` calls. The current implementation doesn't cater for that, and I think that developers would assume that it somehow does.

For this case to be covered, I would think that having something like the `invalidateItemWithIdentifier:` method is needed. But perhaps we could call this method `updateItemsWithIdentifier(s):` instead. Alternatively, we could add a `shouldUpdate:` bool (or `forceUpdate:`) to the `addItemWithIdentifier` method.

What do you think?

hannahmbanana:
+ @lappp9 @austinlouden

lappp9:
This looks super cool 👍

Personally, I agree that option 2 for performing an update sounds better.  It seems like having a custom name implies that something special is happening, whereas I kind of expect `-reloadData` to work basically like  collectionView's `-reloadData`.

Eke:
This will definitely make my life easier. Great PR !

ay8s:
This looks awesome @Adlai-Holler. We've had a couple of cases where we've faced issues with indexPaths especially around reordering updates within our queues. 

We took a similar approach and used the diff method within ASDK to setup the batch updates. But that still seems to cause some issues every so often. Think we can move to this approach pretty quickly within Buffer for some wider testing.

Agreed that Option 2 sounds better. Would be great to see this on ASTableNode!

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -535,16 +548,31 @@ - (void)waitUntilAllUpdatesAreCommitted
...
+
+- (void)_fetchDataFromSourceIfNeeded
+{
+  if (_dataFromDataSourceIsValid) {
+    return;
+  }
+
+  if (self.supportsDeclarativeData) {
+    _data = [self.dataSource dataForDataController:self];
+    [_data markCompleted];
```
levi: We should update the layout spec API to be the same. Currently, setting `isMutable` to `NO`.

codeOfRobin:
Hey @Adlai-Holler ! How similar is the diffing algorithm you've used to the one used in [IGListKit](https://github.com/instagram/IGListKit) ?

AsyncDisplayKit/ASCollectionData.h
```diff
@@ -0,0 +1,131 @@
...
+#import <Foundation/Foundation.h>
+#import <AsyncDisplayKit/ASDimension.h>
+
+@class ASCellNode;
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef NSString * ASSectionIdentifier;
+typedef NSString * ASItemIdentifier;
+typedef NSString * ASSupplementaryElementKind;
```
appleguy: Is it important these are strings, or could they be NSCopying or some other more lenient protocol?  Strings are fine if needed, just thinking there could be incremental value in it being more open if it would be supported easily.

AsyncDisplayKit/ASCollectionData.h
```diff
@@ -0,0 +1,131 @@
...
+@end
+
+/**
+ * An object used to construct the data set for a collection
+ * or table node. 
+ * When the collection needs to update its data, it calls @c dataForCollectionNode:
+ * on the data source. The data source calls @c createNewData on the collection node,
+ * configures the data object, and returns it.
+ */
+@interface ASCollectionData : NSObject
```
appleguy: All of the class and type names in this file are quite nice IMO :)

AsyncDisplayKit/ASCollectionData.h
```diff
@@ -0,0 +1,131 @@
...
+ * block, a default section will be created and used.
+ *
+ * @param identifier The identifier for the new item.
+ * @param nodeBlock A block that will be used to construct the node for the item.
+ *
+ * @note If an item already exists with this identifier, the node block will be ignored.
+ * @note It is more performant to specify the node block inline here, rather than calling a method that generates it, but the difference is not prohibitive.
+ */
+- (void)addItemWithIdentifier:(ASItemIdentifier)identifier
+                    nodeBlock:(ASCellNodeBlock)nodeBlock;
```
appleguy: The fact that this is a "nodeBlock:" does seem to suggest value in calling the other block something like dataBlock or otherwise.

AsyncDisplayKit/ASCollectionData.h
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Finds or creates a section with the given identifier.
+ *
+ * @param identifier The identifier for the section.
+ *
+ * You can use this method to get more fine-grained control over the collection data.
+ * The returned section can be inserted into @c mutableSections.
+ */
+- (id<ASCollectionSection>)sectionWithIdentifier:(ASSectionIdentifier)identifier;
```
appleguy: Is this more like sectionForIdentifier: ?  This sounds a bit like a generator method that would create a section given an identifier, even if it didn't exist previously.  Not definitive feedback though.

AsyncDisplayKit/ASCollectionNode.mm
```diff
@@ -459,6 +462,16 @@ - (nullable UICollectionViewCell *)cellForItemAtIndexPath:(NSIndexPath *)indexPa
   return [self.dataController contextForSection:section];
 }
 
+- (ASCollectionData *)createNewData
+{
+  ASCollectionData *data = [[ASCollectionData alloc] initWithReusableContentFromCompletedData:self.dataController.previousData];
+  __weak __typeof(self) weakSelf = self;
+  data.postNodeBlock = ^(ASCellNode *node) {
+    [weakSelf.view didCreateNode:node];
```
appleguy: Probably add assert main thread in this block for good measure, and assert isNodeLoaded?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -428,11 +430,33 @@ - (void)setAsyncDataSource:(id<ASCollectionDataSource>)asyncDataSource
...
-    ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNumberOfItemsInSection || _asyncDataSourceFlags.collectionViewNumberOfItemsInSection, @"Data source must implement collectionNode:numberOfItemsInSection:");
-    ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNodeBlockForItem
-                        || _asyncDataSourceFlags.collectionNodeNodeForItem
-                        || _asyncDataSourceFlags.collectionViewNodeBlockForItem
-                        || _asyncDataSourceFlags.collectionViewNodeForItem, @"Data source must implement collectionNode:nodeBlockForItemAtIndexPath: or collectionNode:nodeForItemAtIndexPath:");
+    // Make assertions about what they implement, unless they are suppressed.
+    if (self.test_suppressCallbackImplementationAssertions == NO) {
+      // If they implement the functional method, they should not implement the classic ones.
+      if (_asyncDataSourceFlags.dataForCollectionNode) {
+        SEL legacySelectors[] = {
```
appleguy: This is cute, but I'm not sure we should call it legacy yet :).  Maybe imperative?  If you want to separate out the deprecated ones, those are legacy at this point :)

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1335,17 +1381,21 @@ - (ASCellNodeBlock)dataController:(ASDataController *)dataController nodeBlockAt
...
+    [weakSelf didCreateNode:node];
     return node;
   };
   return block;
 }
 
+- (void)didCreateNode:(ASCellNode *)node
+{
+  [node enterHierarchyState:ASHierarchyStateRangeManaged];
+  if (node.interactionDelegate == nil) {
```
appleguy: Is this check important?  Should it check != self?  I wonder what happens if people are reusing ASCellNodes across different data controllers - this would be bad, but ideally either explicitly / immediately bad (or works!), rather than implicitly / subtly bad :)

AsyncDisplayKit/Details/ASChangeSetDataController.mm
```diff
@@ -102,6 +110,42 @@ - (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
...
+{
+  ASDisplayNodeAssertNotNil(_changeSet, nil);
+
+  ASCollectionData * oldData = self.currentData;
+  [self invalidateDataSourceData];
+  ASCollectionData * data = self.currentData;
+  NSIndexSet *insertedSections = nil, *deletedSections = nil;
+  NSArray<NSIndexPath *> *insertedItems = nil, *deletedItems = nil;
+  [oldData.mutableSections asdk_nestedDiffWithArray:data.mutableSections
+                                insertedSections:&insertedSections
```
appleguy: MOAR SPACE to line up :

AsyncDisplayKit/Details/ASCollectionDataController.mm
```diff
@@ -237,7 +238,23 @@ - (void)_populateSupplementaryNodesOfKind:(NSString *)kind withSections:(NSIndex
...
+      // supplementary element kind in this section, and make an array of supplementaries up to that index.
+      // Unfortunately this means we cannot support indexless supplementary elements.
+      NSUInteger itemCount = 0;
+      if (self.supportsDeclarativeData) {
+        for (NSNumber *indexObj in self.currentData.sectionsInternal[sec].supplementaryElements[kind]) {
+          NSInteger index = indexObj.integerValue;
+          if (index == NSNotFound) {
+            ASDisplayNodeFailAssert(@"Supplementary elements with no index are not currently supported. Support is planned for the near future.");
+          }
+          itemCount = MAX(itemCount, index + 1);
```
appleguy: I'm not really sure how to verify this logic is correct.  Seems like a real pain in the ass, will be great to tear apart the supplementary support and rebuild it next year :).  Do we have sufficient test cases to vet this?

AsyncDisplayKit/Details/ASDataController.h
```diff
@@ -118,6 +121,29 @@ FOUNDATION_EXPORT NSString * const ASDataControllerRowNodeKind;
...
+ * The valid ASCollectionData most recently fetched from the data source, if functional data
+ * updating is active.
+ * 
+ * If the data has been invalidated, this property will request new data from the data source before returning.
+ */
+@property (nonatomic, readonly, strong, nullable) ASCollectionData * currentData;
+
+/**
+ * The ASCollectionData most recently fetched from the data source, if functional data
+ * updating is active.
```
appleguy: "... nil otherwise."

AsyncDisplayKit/Details/ASDataController.mm
appleguy: @levi's isMutable idea is a good one -- e.g. we should check if a _data has been marked as completed before using it in other contexts.  No matter what the docs say, people will cache and reuse them :)

AsyncDisplayKit/Private/ASCollectionDataInternal.h
```diff
@@ -0,0 +1,76 @@
...
+@interface ASCollectionSectionImpl : NSObject <ASCollectionSection, ASDescriptionProvider, NSCopying>
+
+- (instancetype)initWithIdentifier:(ASSectionIdentifier)identifier;
+
+// Private properties
+@property (nonatomic, strong, readonly) NSArray<ASCollectionItemImpl *> *itemsInternal;
+@property (nonatomic, strong, readonly) NSMutableDictionary<NSString *, NSMutableDictionary<NSNumber *, ASCollectionItemImpl *> *> *supplementaryElements;
+@end
+
+@interface ASCollectionData () <ASDescriptionProvider>
```
appleguy: Since this is an extension of the actual ASCollectionData object — can we do the same for ASCollectionItem, and provide a base class or class extension to implement the internal functionality rather than a separate object?

AsyncDisplayKit/Private/ASCollectionDataInternal.mm
```diff
@@ -0,0 +1,320 @@
...
+#pragma mark - Convenience Builders (Public)
+
+- (void)addSectionWithIdentifier:(ASSectionIdentifier)identifier block:(void (^)(ASCollectionData * _Nonnull))block
+{
+  if (_currentSection != nil) {
+    ASDisplayNodeFailAssert(@"Call to %@ must not be inside an addSection: block.", NSStringFromSelector(_cmd));
+    return;
+  }
+
+  _currentSection = (ASCollectionSectionImpl *)[self sectionWithIdentifier:identifier];
```
appleguy: Same question about SectionImpl, if it's possible to unify with the public Section object and only expose limited API.

AsyncDisplayKit/Private/ASCollectionDataInternal.mm
```diff
@@ -0,0 +1,320 @@
...
+
+- (ASCollectionItemImpl *)_itemWithIdentifier:(ASItemIdentifier)identifier nodeBlock:(nonnull ASCellNodeBlock)nodeBlock
+{
+  ASCollectionItemImpl *item = _itemsDict[identifier];
+  if (item == nil) {
+    void (^postNodeBlock)(ASCellNode *) = _postNodeBlock;
+    item = [[ASCollectionItemImpl alloc] initWithIdentifier:identifier nodeBlock:^{
+      ASCellNode *node = nodeBlock();
+      if (postNodeBlock != nil) {
+        postNodeBlock(node);
```
appleguy: is postNodeBlock always / only internal?  The name seems a bit ambiguous -- is it postNodeAllocationBlock?  nodeSetupBlock?

AsyncDisplayKitTests/ASCollectionViewTests.mm
```diff
@@ -56,36 +58,92 @@ @implementation ASTestSectionContext
...
+@property (nonatomic, strong) NSMutableArray<NSString *> *sections;
+@property (nonatomic, strong) NSMutableArray<NSMutableArray<NSString *> *> *items;
 @end
 
 @implementation ASCollectionViewTestDelegate {
-  @package
-  std::vector<NSInteger> _itemCounts;
 }
 
 - (id)initWithNumberOfSections:(NSInteger)numberOfSections numberOfItemsInSection:(NSInteger)numberOfItemsInSection {
```
appleguy: Brace style :) (optional)

AsyncDisplayKit/ASCollectionData.h
george-gw: You're right @appleguy which makes its name relevant, as this is exactly what the method does.
The first line of the comment states:

> Finds or **creates** a section with the given identifier.

Based on its implementation, it seems that it does in fact create a section if the identifier does not exist, it just doesn't add it to the `mutableSections`array.

Adlai-Holler:
I'm very excited about making this happen, but this change should have more thorough testing & discussion before we bring it live. If we find extra time, we can bring this back into 2.0 but by default let's call it 2.1.

Adlai-Holler:
This is powerful, but it will be superseded by support for IGListKit who are working tirelessly to make diffing data sources easy.

PR #2570, opened 2016-11-08
Improve Event Tracing Quality, Performance

### LINKED ISSUE #2564: [ASTraceEvent] In debug only, trace events cause large amounts of memory to be retained.
@Adlai-Holler we should fix this in the next week - it causes memory usage to grow to problematic levels even on the simulator.  In this screenshot, it's ASDKgram, without any user interaction occurring - memory growth proceeds rapidly.

screen shot 2016-11-07 at 1 18 44 pm

Adlai-Holler:
Resolves #2564 by a couple means:

- Reduce default buffer length from 20 to 5.
- Create a plain old `NSArray<NSString *>` rather than using `[NSThread callStackSymbols]` directly – the latter returns a private class called `__NSBacktraceArray` that allocates 2KB of memory and retains objects from each frame (register contents?).
- Combine interface state changes into one event instead of separate events for each flag.
- Combine subnode insertions/removals caused by layout transitions into one event, instead of separate events for each.


appleguy:
test this please

appleguy:
Nice improvements, thanks @Adlai-Holler!

Issue #2140, opened 2016-08-25
truncation string is not working reliablely with NSLineBreakByTruncatingTail


Yue-Wang-Google:
Test case: 

Apply the following patch to the Kittens example:

```
--- /Users/wangyue/Downloads/AsyncDisplayKit-master/examples/Kittens/Sample/KittenNode.mm   2016-08-25 10:03:06.000000000 -0700
+++ Sample/KittenNode.mm    2016-08-25 13:00:22.000000000 -0700
@@ -98,8 +98,12 @@

   // lorem ipsum text, plus some nice styling
   _textNode = [[ASTextNode alloc] init];
+  _textNode.truncationMode = NSLineBreakByTruncatingTail;
+  _textNode.maximumNumberOfLines = 2;
   _textNode.attributedString = [[NSAttributedString alloc] initWithString:[self kittyIpsum]
                                                                attributes:[self textStyle]];
+  _textNode.truncationAttributedText = [[NSAttributedString alloc] initWithString:@"\u2026 More"
+                                                                       attributes:[self textStyle]];
   [self addSubnode:_textNode];

   // hairline cell separator
```

See the first node  in the following screenshot. 

![screen shot 2016-08-25 at 1 04 44 pm](https://cloud.githubusercontent.com/assets/20689178/17984049/8ae77e7a-6ac4-11e6-8eec-dd5b11771363.png)

Yue-Wang-Google:
OK it seems ASDK calculate the truncation point either too early or too late. When using NSLineBreakByTruncatingTail, seems the line truncates too late. When using without, it happens too early. In a few cases two or three more words can be crammed to the last line, but ASDK just truncate a lot earlier.

abbasmousavi:
I have the same problem, It seems the line truncates too late, When using ```NSLineBreakByTruncatingTail``` and truncates too early when using ```NSLineBreakByWordWrapping```.

Also when ```NSLineBreakByTruncatingTail```  breaks the line too late, it displays default truncate message instead of ```truncationAttributedText``` or ``` additionalTruncationMessage```

abbasmousavi:
@appleguy Could you please take a look at this issue?

jdee:
[...]

PR #2581, opened 2016-11-09
Disable Text Node Fast Path

Adlai-Holler:
As fast as this is, it's unfortunately crashing sometimes in production, inside `boundingRectWithSize:`.

This also reveals an issue with our classic (slow-path) truncation. Before this diff, both `testFastPathTruncationWorks` and `testSlowPathTruncationWorks` were using the fast-path due to an incorrectly formed test. 

I fixed `testSlowPathTruncationWorks` to actually use the slow path, and it reveals that the text is truncated too early. Investigation reveals that two truncation passes occur. The first pass calculates the correct truncation index, but then the second pass computes the index just before it. This helps us isolate #2140 .


Adlai-Holler:
Ran the tests locally. This is pretty urgent so I'm going to land it and damn the consequences!

jdee:
This looks like a plausible fix for #2140. I wanted to try it out with master, but I forgot that master is version 2.0. Are there plans to roll this fix into a 1.9.93 or something?

maicki:
@jdee Hey, we are currently hard working on 2.0 and further fixes on the text component of ASDK. Currently there are no concrete plans to roll that into 1.9.93 though.

jdee:
@maicki Thanks for the info. It's a fairly minor bug in a production app that uses ASDK almost exclusively for the UI. Updating to 2.0 looks painful. I guess we won't have much choice but to invest significant effort at some point. I'm sure it will improve a lot of things. Maybe there is/will be a guide or other assistance once we get there.

At least this explains why it works as long as the text includes highlighted links, and why my attempt to insert an invisible link didn't fix anything. Presumably you're checking whether the styling varies and optimizing when it doesn't. But now that I understand that, maybe I can find another workaround in 1.9.92.

Thanks again.

maicki:
@jdee In terms of updating to 2.0. We just updated our documentation to include guides to upgrade to 2.0: http://asyncdisplaykit.org/docs/adoption-guide-2-0-beta1.html hope that helps.

PR #2582, opened 2016-11-09
Fix Issue in ASPagerNode Flags Handling, Restore Disabled Test

Adlai-Holler:
We disabled an ASViewController test because it was failing after #2561 landed.

It turns out the cause of the failure is that the pager node tests set the application's key window and when the ASViewController test updated the key window, the previous one would be released and torn down. The pager node would have its data source set to `nil`, and would throw an exception because `nil` doesn't implement either `nodeBlockAtIndex:` or `nodeAtIndex:`.

This diff updates the flags handling for `ASPagerNode` pretty holistically, handling the `nil` case, and restores the affected view controller test.

### DIFF
--- a/AsyncDisplayKit/ASPagerNode.m
+++ b/AsyncDisplayKit/ASPagerNode.m
@@ -14,6 +14,8 @@
 #import "ASDelegateProxy.h"
 #import "ASDisplayNode+Subclasses.h"
 #import "ASPagerFlowLayout.h"
+#import "ASAssert.h"
+#import "ASCellNode.h"
 
 @interface ASPagerNode () <ASCollectionDataSource, ASCollectionDelegate, ASCollectionViewDelegateFlowLayout, ASDelegateProxyInterceptor>
 {
@@ -23,6 +25,7 @@
   ASPagerNodeProxy *_proxyDataSource;
   struct {
     unsigned nodeBlockAtIndex:1;
+    unsigned nodeAtIndex:1;
   } _pagerDataSourceFlags;
 
   __weak id <ASPagerDelegate> _pagerDelegate;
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASPagerNode.m
```diff
@@ -21,11 +21,15 @@ @interface ASPagerNode () <ASCollectionDataSource, ASCollectionDelegate, ASColle
 
   __weak id <ASPagerDataSource> _pagerDataSource;
   ASPagerNodeProxy *_proxyDataSource;
-  BOOL _pagerDataSourceImplementsNodeBlockAtIndex;
+  struct {
+    unsigned nodeBlockAtIndex:1;
```
appleguy: Nice style!

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -159,9 +163,13 @@ - (void)setDataSource:(id <ASPagerDataSource>)dataSource
...
     
-    _pagerDataSourceImplementsNodeBlockAtIndex = [_pagerDataSource respondsToSelector:@selector(pagerNode:nodeBlockAtIndex:)];
-    // Data source must implement pagerNode:nodeBlockAtIndex: or pagerNode:nodeAtIndex:
-    ASDisplayNodeAssertTrue(_pagerDataSourceImplementsNodeBlockAtIndex || [_pagerDataSource respondsToSelector:@selector(pagerNode:nodeAtIndex:)]);
+    if (dataSource == nil) {
+      memset(&_pagerDataSourceFlags, 0, sizeof(_pagerDataSourceFlags));
+    } else {
+      _pagerDataSourceFlags.nodeBlockAtIndex = [_pagerDataSource respondsToSelector:@selector(pagerNode:nodeBlockAtIndex:)];
+      // Data source must implement pagerNode:nodeBlockAtIndex: or pagerNode:nodeAtIndex:
+      ASDisplayNodeAssertTrue(_pagerDataSourceFlags.nodeBlockAtIndex || [_pagerDataSource respondsToSelector:@selector(pagerNode:nodeAtIndex:)]);
```
appleguy: @Adlai-Holler what happens if the data source does not implement either of these, in production?  Can we guard against this by setting it to nil or ensuring that nodeAtIndex: isn't called later?

PR #2589, opened 2016-11-09
Improve Automatic Subnode Management Node Ordering

### LINKED ISSUE #2549: [AutomaticSubnodeMgmt] node ordering issue when using ASBackgroundLayoutSpec/ASOverlayLayoutSpec
I have node which has automaticallyManagesSubnodes=true. I have this spec in my code which works as intended.
`ASBackgroundLayoutSpec(child: self.tableNode, background: self.switchNode)`
It puts switch node behind table node. But if i change it like that.
`ASBackgroundLayoutSpec(child: self.tableNode, background: ASInsetLayoutSpec(insets: UIEdgeInsetsMake(1,1,1,1), child: self.switchNode))`
Then table node is behind switch node. Same issue happens if i use ASOverlayLayoutSpec.
I use 2.0 beta.

Adlai-Holler:
Resolves #2549 

This diff changes the flattening traversal from breadth-first to depth-first, so that the resulting node array will have an ordering that matches the user's expectations, and adds some unit tests which failed before the change.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2184,7 +2184,7 @@ ASDISPLAYNODE_INLINE BOOL nodeIsInRasterizedTree(ASDisplayNode *node) {
 - (NSArray *)subnodes
 {
   ASDN::MutexLocker l(__instanceLock__);
-  return [_subnodes copy];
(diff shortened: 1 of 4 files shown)


levi:
Looks good! Thanks for taking the time to look at this issue.

PR #2592, opened 2016-11-10
[Examples] ASViewController Improvements

maicki:
Some random ASViewController example improvements while looking into some changes around that.


appleguy:
@maicki nice improvements indeed!

PR #2601, opened 2016-11-14
[ASTextNode] Set flexShrink to 1.0 on Text Nodes by Default

Adlai-Holler:
The vast majority of use cases prefer this behavior. We previously held off on doing this for the sake of consistency across the framework. When the impact is this low, user convenience should precede consistency of that kind. See `UIImageView` changing the default value for `userInteractionEnabled`.

It's unfortunate that we create the style object for all text nodes in this diff. We could do something more elaborate in the future, by creating a "default style object" system where we share singleton style objects instead of using `nil`. The cost is acceptably small for this diff.

### DIFF
--- a/AsyncDisplayKit/ASTextNode.mm
+++ b/AsyncDisplayKit/ASTextNode.mm
@@ -117,6 +117,10 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
     self.opaque = NO;
     self.backgroundColor = [UIColor clearColor];
 
+    // Enable flexShrink by default on text nodes. Text is naturally reflowable and
+    // the vast majority of use cases prefer this behavior.
+    self.style.flexShrink = 1.0;
+
     self.linkAttributeNames = DefaultLinkAttributeNames;
 
     // Accessibility
--- a/AsyncDisplayKit/Layout/ASStackLayoutElement.h
+++ b/AsyncDisplayKit/Layout/ASStackLayoutElement.h
@@ -40,7 +40,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * @abstract If the sum of childrens' stack dimensions is greater than the maximum size, how much should this component shrink?
  * This value represents the "flex shrink factor" and determines how much this component should shink in relation to
- * other flexible children.
+ * other flexible children. Defaults to @c 0.0 except for @c ASTextNode which defaults to @c 1.0.
  */
 @property (nonatomic, readwrite) CGFloat flexShrink;
 
--- a/examples_extra/ASLayoutSpecPlayground-Swift/Sample/Sample.playground/Pages/StackLayout.xcplaygroundpage/Contents.swift
+++ b/examples_extra/ASLayoutSpecPlayground-Swift/Sample/Sample.playground/Pages/StackLayout.xcplaygroundpage/Contents.swift
@@ -7,8 +7,9 @@ import AsyncDisplayKit
 extension StackLayout {
 
   override public func layoutSpecThatFits(_ constrainedSize: ASSizeRange) -> ASLayoutSpec {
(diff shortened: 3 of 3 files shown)


Adlai-Holler:
Michael has just made the excellent point, that this will not have the desired effect in cases where the user has an inset layout spec or something wrapped around the text node. In the future we may be able to propagate this value outward automatically.

Would love comments from the community @george-gw @gazreese @austinlouden @tomizimobile @lappp9 @appleguy everyone join up!

hannahmbanana:
Context copied from a separate convo:
- The default value for flex shrink in the CSS spec is 1.0 for all elements. All of the other libraries (css-layout, Componentkit, react native) the default value is 0 for flex shrink.

Thoughts:
- is it better to have unexpected shrinking than content overflowing for text elements?
- would it be better for consistency to keep all nodes to have a flex shrink default values 0?
- setting flex shrink to 1 is not always the full solution. If a text node is wrapped in an inset spec, then the inset spec needs flex shrink set on it as well.

lappp9:
Personally, I would expect text to shrink by default in my layouts and like the comparison to UIImageView having userInteractionEnabled set to YES by default.  

Oftentimes, if there's an element with a flexible size in the layouts I've worked on it ends up being things like long usernames or content that can be expanded.  This is just anecdotal to what I've seen though so maybe people would prefer the overflow by default.

nguyenhuy:
I think it's good to have text nodes to shrink by default. Couple of follow ups:
- How do we better highlight this exception? Maybe add it to [FAQ](http://asyncdisplaykit.org/docs/faq.html)?
- Once this is merged, we should prioritize implementing outward propagation of style properties.

lappp9:
@nguyenhuy It might be worth just putting a disclaimer on the ASTextNode page itself?  Really, it might be nice to have a little "Layout Tips" section for a couple of the nodes highlighting whether they have an intrinsic content size or any other good to know things.

Adlai-Holler:
This seems to have a decent base of support. 

@garrettmoon This may affect existing layouts in Pinterest. Do you think we should take any extra steps to vet & notify?

@hannahmbanana If this lands, would you be willing to add this to the FAQ and the ASTextNode pages?

hannahmbanana:
@Adlai-Holler - has this been tried with Pinterest yet? If not, I could check it out.

Adlai-Holler:
That would be great 👌

hannahmbanana:
A quick test didn't find anything, but @garrettmoon will have more insight into what this could potentially break.

maicki:
I'm still the opinion this will introduce more confusion than it helps 😬

@hannahmbanana if you look into Pinterest code also look how often we actually add the textnode directly to a stack spec and how often we wrap it into another spec and put this inside a stack steck.

george-gw:
@Adlai-Holler could you also change a small thing in: [ASLayoutSpecPlayground-Swift example](https://github.com/facebook/AsyncDisplayKit/blob/master/examples_extra/ASLayoutSpecPlayground-Swift/Sample/Sample.playground/Pages/StackLayout.xcplaygroundpage/Contents.swift )

change:
```
// Try commenting out the flexShrink to see its consequences.
subtitleNode.style.flexShrink = 1.0
```
to:
```
// Try commenting out the flexShrink to see its consequences.
subtitleNode.style.flexShrink = 0.0
```

garrettmoon:
@Adlai-Holler @hannahmbanana I don't have any additional insight into what this would break, but we'll need to communicate this far in advance of landing, then land right after a branch.

garrettmoon:
I also am worried that this will be confusing without the additional "propagating out" changes.

Adlai-Holler:
This diff is still pretty controversial – let's hold off on it for the time being, in light of the concern @garrettmoon and @maicki raised about propagation. I've linked the propagation issue.

PR #2609, opened 2016-11-15
Add method to specify the default NSURLSessionConfiguration

rahul-malik:
This will be the configuration used by the PINRemoteImageManager instance contained by ASPINRemoteImageDownloader. This will allow callers to supply their own custom
logic whether that is background downloads, custom http headers, cookie
storage, etc.

### DIFF
--- a/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
+++ b/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
@@ -77,16 +77,22 @@
 //Share image cache with sharedImageManager image cache.
 - (id <PINRemoteImageCaching>)defaultImageCache
 {
-    return [[PINRemoteImageManager sharedImageManager] cache];
+  return [[PINRemoteImageManager sharedImageManager] cache];
 }
 
+@end
+
+
+static ASPINRemoteImageDownloader *sharedDownloader = nil;
+
+@interface ASPINRemoteImageDownloader ()
 @end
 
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -77,48 +77,65 @@ @implementation ASPINRemoteImageManager
...
   static dispatch_once_t onceToken = 0;
   dispatch_once(&onceToken, ^{
     sharedDownloader = [[ASPINRemoteImageDownloader alloc] init];
   });
   return sharedDownloader;
 }
 
++ (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfiguration *)configuration
+{
+  NSAssert(sharedDownloader != nil, @"Singleton has been created and session can no longer be configured.");
```
rahul-malik: This assertion is incorrect. I'll update now

garrettmoon:
@rahul-malik would it be better if this was something that a subclass provided? That way you can guarantee it's called before the sharedDownloader is created?

rahul-malik:
@garrettmoon - How would we guarantee that the network and multiplex image nodes used the subclass instance?

garrettmoon:
I hadn't thought about that :/ ok, I think you're right this may be better.

appleguy:
@rahul-malik thanks!  Great to have this kind of customizability for all clients.

PR #2610, opened 2016-11-15
[ASDisplayNode] Placeholder contents configuration needs to check self.contents on main.

garrettmoon:


Adlai-Holler:
@maicki Exactly. Moiging.

PR #2614, opened 2016-11-16
[Examples] Added Bear Shop project

StrauneanuDimitri:
This project (in Swift 3) demonstrates different layout examples.


hannahmbanana:
@StrauneanuDimitri - thanks for submitting this! 

We had planned to consolidate all layout examples into a single app - the `LayoutSpecExamples-Swift` app. I'll take a look at this later today and see if it would be possible to incorporate into that app.

appleguy:
@StrauneanuDimitri thank you for submitting this!  We're always excited to get new example code, especially in Swift, since we don't have a ton of stuff in the new language.

Could you post a few screenshots of the app on the diff so we have it in the PR record / commit history?

It may make sense to keep this as a separate app rather than folding into the layout examples, because it has some data source supporting infra, but let's see - we might be able to put it in a tab of another app and organize the code nicely in a group.  Anything that will maximize the number of people opening the project and seeing the code, rather than having to pod install for a bunch of separate projects.

StrauneanuDimitri:
Yes.

Adlai-Holler:
Thanks for contributing @StrauneanuDimitri ! Sorry for the delay in landing this, very nifty app!

PR #2623, opened 2016-11-17
[ASDisplayNode] Convert assertion to log so developer can see layout issues onscreen.

appleguy:
In this case, the user hasn't specified enough about the layout of a node.  We default to
0, 0 - but ideally the case should not occur at all.  So it's important to help developers
detect these cases and fix them quickly.

An assertion causes developers several problems:
- They can't see important information unless an exception breakpoint is manually added.
- They can't see their layout onscreen and so visually understanding the problem is impossible.
- They can't proceed with testing to trigger other faults in the layout and are blocked fixing one at a time.

A log fixes every one of those problems:
- They can set a breakpoint on the log line very easily if desired.
- They can see their layout display and recognize the 0,0 node even more quickly than the log text information.
- They can see if multiple logs print out, or if the log only occurs for one item in a feed, or certain types --
    This helps them debug faster by knowing if the layout is always broken or certain conditions break it.

Since developing with ASLayoutSpecs is an iterative process, it's crucial that we let developers have the
freedom to experiment and test without hitting too many assertions.  Fortunately it will be impossible to ignore
these huge logs (full recursive description) and their nodes will be 0, 0 size, so they will get fixed.

### DIFF
--- a/AsyncDisplayKitTests/ASDisplayNodeLayoutTests.mm
+++ b/AsyncDisplayKitTests/ASDisplayNodeLayoutTests.mm
@@ -111,12 +111,6 @@
   };
   
   XCTAssertThrows([node layoutThatFits:ASSizeRangeMake(CGSizeMake(0, FLT_MAX))]);
-  
-  // This dance is necessary as we would assert in case we create an ASDimension that is not real numbers
-  ASDimension width = displayNode.style.width;
-  width.value = INFINITY;
-  displayNode.style.width = width;
-  XCTAssertThrows([node layoutThatFits:ASSizeRangeMake(CGSizeMake(0, 0), CGSizeMake(INFINITY, INFINITY))]);
 }
 
 - (void)testThatLayoutCreatedWithInvalidSizeCausesException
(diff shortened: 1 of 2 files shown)


appleguy:
@maicki @Adlai-Holler @nguyenhuy This is important to get into Pinterest.  Currently the third tab will crash on master (if the server response contains certain types) as well as other areas.

This is because a previously supported operation, returning [[ASCellNode alloc] init], is now a hard assertion / crash in debug that hinders development.

Note even our base Section Controller uses this return blank cell for its empty state implementation, if the delegate doesn't provide anything specific for the empty state.  There are actually 31 cases of [[ASCellNode alloc] init] in the codebase currently, but in addition to those, this is important mostly for developers iterating on new layouts locally as they are still trying to get them to work.

Once reviewed, please merge this for me - I want it to land before Beta 2 goes out.

AsyncDisplayKitTests/ASDisplayNodeLayoutTests.mm
```diff
@@ -111,12 +111,6 @@ - (void)testThatLayoutWithInvalidSizeCausesException
   };
   
   XCTAssertThrows([node layoutThatFits:ASSizeRangeMake(CGSizeMake(0, FLT_MAX))]);
-  
-  // This dance is necessary as we would assert in case we create an ASDimension that is not real numbers
-  ASDimension width = displayNode.style.width;
-  width.value = INFINITY;
-  displayNode.style.width = width;
-  XCTAssertThrows([node layoutThatFits:ASSizeRangeMake(CGSizeMake(0, 0), CGSizeMake(INFINITY, INFINITY))]);
```
appleguy: This no longer throws, but it does NSLog, and any regular developer would see the node missing in their layout (as well as a non-silenced, really huge log).

PR #2624, opened 2016-11-17
[ASCollectionNode, ASTableNode] Deprecate asyncDelegate & asyncDataSource Properties 

### LINKED ISSUE #2459: [ASCollectionNode, ASTableNode] Deprecate asyncDelegate & asyncDataSource Properties
All users should be using the `delegate` and `dataSource` properties on the node.

This should be very easy.

hannahmbanana:
Resolves #2459.


hannahmbanana:
I'm having trouble connecting to Phabricator, but I sent the diff that fixes Pinterest to Adlai.

Issue #2615, opened 2016-11-16
[ASCellNode] implement a .viewController property


MarvinNazari:
When using `ASCellNode(viewControllerBlock:` there is no way to access the `viewcontroller` afterwards, we use it in some places specially with `ASPagerNode`
this is how i do it right now
```swift
extension ASCellNode {
    var viewController: UIViewController? {
        return valueForKey("_viewController") as? UIViewController
    }
}
```

hannahmbanana:
@MarvinNazari - This is possible to implement, but has yet to happen. It would require some logic to make sure that the view controller is only accessed from the main thread. If you'd be interested in submitting a PR for this, we'd be happy to merge it.

Adlai-Holler:
Thanks for building this @MarvinNazari! Hope to see you around more!

PR #2625, opened 2016-11-17
[ASCellNode] implement a .viewController property

MarvinNazari:
#2615

### DIFF
--- a/AsyncDisplayKit/ASCellNode.mm
+++ b/AsyncDisplayKit/ASCellNode.mm
@@ -236,6 +236,12 @@ static NSMutableSet *__cellClassesForVisibilityNotifications = nil; // See +init
   return nil;
 }
 
+- (UIViewController *)viewController
+{
+  ASDisplayNodeAssertMainThread();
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASCellNode.mm
```diff
@@ -236,6 +236,12 @@ - (NSIndexPath *)indexPath
   return nil;
 }
 
+- (UIViewController *)viewController
+{
+  ASDisplayNodeAssertMainThread();
+  return _viewController;
```
Adlai-Holler: Would it be reasonable to force-load the view here, so that the cell node will run the view controller block if it hasn't already? It would be sufficient to add something like:

```objectivec
- (UIViewController *)viewController
{
  // Force the view to load so that we will create the
  // view controller if we haven't already.
  if (self.isNodeLoaded == NO) {
    [self view];
  }
  return _viewController;
}
```

MarvinNazari:
@Adlai-Holler thanks for the review. Updated. 😄

PR #2627, opened 2016-11-17
[ASStackLayoutSpec] Don’t floor space between items for ASStackLayoutJustifyContentSpaceBetween and ASStackLayoutJustifyContentSpaceAround

### LINKED ISSUE #1339: [ASLayoutSpec] ASStackLayoutJustifyContentSpaceBetween flooring causing issues
I have a ASStackLayoutSpec that contains nine views stacked with .SpaceBetween. The flooring in the following line causes enough extra spacing to make the views not perfectly aligned:
 `return stackedLayout(style, 0, floorf(violation / numOfSpacings), fmodf(violation, numOfSpacings), unpositionedLayout, constrainedSize);`
In my specific case at least, it seems to lay out perfectly like this:
`return stackedLayout(style, 0, violation / numOfSpacings, 0, unpositionedLayout, constrainedSize);`

Is this something that should be changed, or maybe a new ASStackLayoutJustifyContent case made for?

maicki:
We should not floor while calculating the space between items for `justifyContent`: `ASStackLayoutJustifyContentSpaceBetween` or `ASStackLayoutJustifyContentSpaceAround` and take the sub pixel rendering if necessary. Otherwise the space between will grow with the number of items. Good to see in the sample project from the issue below. It just mentions `ASStackLayoutJustifyContentSpaceBetween` but it actually also happens in `ASStackLayoutJustifyContentSpaceAround` so let's fix both.

Resolves #1339

@nguyenhuy @appleguy @Adlai-Holler @garrettmoon Ready for review


AsyncDisplayKit/Private/ASStackPositionedLayout.mm
```diff
@@ -118,12 +118,13 @@ static ASStackPositionedLayout stackedLayout(const ASStackLayoutSpecStyle &style
       return stackedLayout(style, violation, unpositionedLayout, constrainedSize);
     }
     case ASStackLayoutJustifyContentSpaceBetween: {
+      // Spacing between items and no spaces at the edges and even distributed
```
garrettmoon: nit: // Spacing between the items, no spaces at the edges, evenly distributed.

Issue #2560, opened 2016-11-06
[ASCollectionViewLayoutInspector] should be pulled out into it's own file.


garrettmoon:
Or at the very least, ASCollectionViewFlowLayoutInspector.* should be renamed to ASCollectionViewLayoutInspector.

garrettmoon:
I'll take this one on.

PR #2635, opened 2016-11-19
[ASCollectionView] Move ASCollectionViewLayoutInspector to its own file

garrettmoon:
Mostly a move, however one shared function becomes non-inline and I removed the assert in ASCollectionViewLayoutInspector scrollableDirections.


garrettmoon:
Addresses #2560

appleguy:
@garrettmoon thanks!  We should revisit this API soon, but breaking these up is definitely a good step.

Issue #2350, opened 2016-10-06
[ASViewController] Accessing vc.node.view doesn't go through ViewController's -loadView.


EugenePogrebnoy:
Hi everybody,

I have a problem when trying to nest two `ASViewController`s: the nodes of both of them are loaded and laid out properly, but the `viewDidLoad()` method of the child view controller is not getting called. 

Here is my code:

```
import UIKit
import AsyncDisplayKit

@UIApplicationMain
class AppDelegate: UIResponder, UIApplicationDelegate {
    var window: UIWindow?

    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
        self.window = UIWindow()
        self.window?.rootViewController = ParentViewController()
        self.window?.makeKeyAndVisible()

        return true
    }
}

class ParentViewController: ASViewController<ASDisplayNode> {

    let child = ChildViewController()

    init() {
        super.init(node: ASDisplayNode())
        self.node.backgroundColor = .red

        self.node.usesImplicitHierarchyManagement = true
        self.node.layoutSpecBlock = { [child = self.child] _ in
            let childNode = child.node
            return ASInsetLayoutSpec(insets: UIEdgeInsetsMake(20, 20, 20, 20), child: childNode)
        }

        self.addChildViewController(self.child)
        self.child.didMove(toParentViewController: self)
    }

    required init?(coder aDecoder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }

    override func viewDidLoad() {
        super.viewDidLoad()

        self.node.backgroundColor = .green
    }
}

class ChildViewController: ASViewController<ASDisplayNode> {
    init() {
        super.init(node: ASDisplayNode())
        self.node.backgroundColor = .blue
    }

    required init?(coder aDecoder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }

    override func viewDidLoad() {
        super.viewDidLoad()

        self.node.backgroundColor = .yellow
    }
}
```

![simulator screen shot 6 2016 12 53 51](https://cloud.githubusercontent.com/assets/430187/19148132/f67afaca-8bc3-11e6-9f2a-47ced969775a.png)

I expect the inner rectangle to be yellow.

[...]

Issue #2213, opened 2016-09-08
[ASViewController] Regression causing layout issues when presenting a portrait-only view controller while in landscape


GitGadd:
Presenting a portrait-only view controller while in landscape causes incorrect frames for both the presented and the presenting view controller.

It seems the view controllers are being laid out in the orientation/bounds of the other view controller, causing the portrait view controller to be laid out in landscape and vice versa.

This regression was introduced sometime after 1.9.6.

Attaching videos of presenting in 1.9.81 (the bug) as well as 1.9.6 as reference for the expected behavior. Also attaching a demo project reproducing the issue.

Steps to reproduce:
1. Add a rotatable view controller as the root view controller of the window
~~~~
 func supportedInterfaceOrientations() -> UIInterfaceOrientationMask { return .All } 
~~~~
2. Rotate the device to landscape
3. Present a portrait-only view controller over the root view controller
~~~~ 
func supportedInterfaceOrientations() -> UIInterfaceOrientationMask { return [.Portrait] }
 ~~~~

Expected: The presented view controller slides up from the portrait bottom of the device (home button). When dismissed, it slides towards the physical bottom of the device (left or right edge of the screen, depending on home button positioning)

Result: The presented view controller has an incorrect (landscape) frame with the window showing behind it due to not covering the entire screen. When dismissed, the presented view controller also has an incorrect (portrait) frame.

[ASRotatedPresentation.zip](https://github.com/facebook/AsyncDisplayKit/files/461654/ASRotatedPresentation.zip)

[ASDKRotationVideos.zip](https://github.com/facebook/AsyncDisplayKit/files/461658/ASDKRotationVideos.zip)

GitGadd:
Confirmed fixed on master! That's great news, thank you. I'll make sure to test on master before posting future issues.

hannahmbanana:
@GitGadd: no worries! I will use your examples to add unit tests to prevent future regressions. 

FYI - there have been a few breaking API changes on master so far that are [listed here](http://asyncdisplaykit.org/docs/upgrading.html)

GitGadd:
This just happened to me again while building against master.

It only happens in a very specific case where I am overlaying the presenting view controller's node with two nodes at the same time.

The layout spec of the presenting ASViewController's node:

~~~~
override func layoutSpecThatFits(_ constrainedSize: ASSizeRange) -> ASLayoutSpec {
[...]

Issue #2085, opened 2016-08-17
ASViewController does not resize when contained in a Container View


dmcrodrigues:
This seems similar/related to #1959 but this happens consistently in both iOS 8 and 9. I also tested with both 1.9.80, 1.9.81 and master and the issue happens consistently.

An `ASViewController` added as child view controller in a simple `UIViewController` which also is a child view controller but contained in a Container View is not resized at all, it keeps the original dimensions of the size class.

A workaround to force the resize is overriding this methods in the parent view controller and tweak the instance of `ASViewController`.

``` swift
override func traitCollectionDidChange(previousTraitCollection: UITraitCollection?) {
     super.traitCollectionDidChange(previousTraitCollection)
     viewController.traitCollectionDidChange(previousTraitCollection)
}

override func viewDidLayoutSubviews() {
     super.viewDidLayoutSubviews()
     viewController.view.frame = view.frame
     viewController.view.setNeedsLayout()
}
```

I'm attaching a sample project where the issue can be observed, the red view controller (`ASViewController`) should overlap completely his parent, the green one, but it doesn't.

![simulator screen shot 17 aug 2016 13 01 53](https://cloud.githubusercontent.com/assets/1098602/17735556/cab4a9d8-647a-11e6-8c80-74d81aef18e9.png)

If we set `ContainedViewController` as the _Initial View Controller_ in the Storyboard, and that way removing the Container View from the equation, the red view controller (`ASViewController`) is properly resized to match the size of the parent.

The sample project can be found here:
[StoryboardContainedEmbed.zip](https://github.com/facebook/AsyncDisplayKit/files/422505/StoryboardContainedEmbed.zip)

tungvoduc:
I have the same problem. I have a custom view controller container that contains two ASViewController instances as child view controllers. Changing child view controllers' frames in viewWillLayoutSubviews used to work but not anymore after version 2.0-beta.1.

hannahmbanana:
This may be resolved by #2736.

maicki:
I just tested this with `master` and it appears to be fixed with the recent changes we did around `ASViewController` and the layout based on the bounds.

I will close this issue for now please reopen if you think it's not fixed in your case.

hannahmbanana:
cc @dmcrodrigues

PR #2636, opened 2016-11-20
[ASViewController] Don’t use the view of the root node for the view of the ASViewController

maicki:
Due to the fact we use an `ASDisplayNode`’s view as view of the `ASViewController` we ran into a couple of problems in the past as well as we have a couple of open issues (see issues below). Especially around sizing and layout, we work against UIKit instead of working with UIKit.

### Changes
- We should change from having the view of the display node as ASViewController.view to add the node as subview to the ASViewController.view
- Stretch the node always the full size of the ASViewController.view in the layout pass
- Remove `nodeConstrainedSize`

### Open Questions
- `ASViewController` containment: Big question mark and that's the reason why we have this PR still in development stage is the handling of view controller containment. A good example is within here: https://gist.github.com/maicki/45a33341af35d6a04dbb8df7db13f42e . The problem within here is that the node hierarchy will not reflect the view hierarchy anymore. @Adlai-Holler, @appleguy any thoughts on it?

### Related issues
- ASViewController/ASPagerNode layout bug since version 1.9.81 https://github.com/facebook/AsyncDisplayKit/issues/1949 (fixed)
- ASViewController does not resize when contained in a Container View https://github.com/facebook/AsyncDisplayKit/issues/2085 (fixed)
- [ASViewController] Accessing vc.node.view doesn't go through ViewController's -loadView. (fixed) https://github.com/facebook/AsyncDisplayKit/issues/2350
- Accepts edgesForExtendedLayout without any hacks (fixed without special code) https://github.com/facebook/AsyncDisplayKit/pull/2123
- [ASViewController] Regression causing layout issues when presenting a portrait-only view controller while in landscape (fixed without special code) https://github.com/facebook/AsyncDisplayKit/issues/2213
- [ASViewController] iOS 8 + UIWindow rotation problem (fixed without special code) https://github.com/facebook/AsyncDisplayKit/issues/1905
- [ASViewController] Modal presented ASViewController don't rotate on iOS 8.3 (fixed without special code) https://github.com/facebook/AsyncDisplayKit/issues/1810

### DIFF
--- a/AsyncDisplayKit/ASViewController.h
+++ b/AsyncDisplayKit/ASViewController.h
@@ -64,16 +64,6 @@ typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitWindowSizeBlock)(C
 // Refer to examples/SynchronousConcurrency, AsyncViewController.m
 @property (nonatomic, assign) BOOL neverShowPlaceholders;
 
-
-/**
- * The constrained size used to measure the backing node.
- *
- * @discussion Defaults to providing a size range that uses the view controller view's bounds as
- * both the min and max definitions. Override this method to provide a custom size range to the
- * backing node.
- */
-- (ASSizeRange)nodeConstrainedSize AS_WARN_UNUSED_RESULT;
-
 @end
 
 @interface ASViewController (ASRangeControllerUpdateRangeProtocol)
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Nice! This essentially resolves https://github.com/facebook/AsyncDisplayKit/issues/2350 as well, since now using `vc.node.view` and `vc.view` interchangeably is no longer valid. Some users may misunderstand this, but I think that's acceptable.

appleguy:
@maicki once you resolve the merge conflicts, excited to see this one green! √

maicki:
@appleguy Rebased and merge conflict resolved

maicki:
@Adlai-Holler Some answers:
- We don't rely on a `setNeedsLayout` call. What is your thinking behind it?
- It should work with `initWithViewBlock:` nodes as the view will be assigned as view to the node. Layout does not rely on any `setNeedsLayout` call anymore. We call `layoutThatFits:` in `viewWillLayoutSubview` or if the `constrainedSize` will change due to frame changes we use the frame within the node to layout the subnodes if a layout pass happens.
- I changed to set the autoresizing mask to `[.FlexibleWidth, .FlexibleHeight]`. We still call `layoutThatFits:` in `viewWillLayoutSubviews` to prepare the node for the layout pass otherwise we just delay the measure pass to the layout pass. It should not have any overhead as if the frame didn't change it's a no op. We could push this layout async in the future, but I actually already plan to look into that doing it within `ASDisplayNode` by default.

Adlai-Holler:
Hmmm what if we keep the current behavior, but always use exactly `view.bounds.size` as nodeConstrainedSize? It seems like `UIViewController` is designed to naturally support you providing your own view, as long as you don't manually change its frame.

appleguy:
@Adlai-Holler @maicki I think that last statement is correct:

- UIViewController is most definitely designed for users providing their own view.  This is of course what -loadView is for.
- This works with UITableView, UICollectionView, but also many other custom view subclasses.  It should work with _ASDisplayView too :).
- Indeed, the main difference is sizing / layout.  UIViewController expects to set the bounds on the view, just like the parent of a UIView would; also just like a regular UIView, it should never change its own bounds size.

Adlai-Holler:
Superseded by #2653

PR #2637, opened 2016-11-20
[ASDisplayNode] Re-add layout transition check for invalid constrained size

maicki:
Using CGSizeZero for the sizeRange can cause negative values in client layout code. Most likely called transitionLayout: without providing a size, before first layout pass.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -845,6 +845,12 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
                 measurementCompletion:(void(^)())completion
 {
   ASDisplayNodeAssertMainThread();
+  
+  if (constrainedSize.max.width <= 0.0 || constrainedSize.max.height <= 0.0) {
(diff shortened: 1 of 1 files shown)


appleguy:
Nice, this is a good improvement.  We still have an occasional layout issue but this fixes an assertion which is more serious.

We should consider if non-animated transitionLayout: calls that hit this case can be diverted to a -setNeedsLayout instead, as during coalescing they will probably get the size they need.  However, the measurementCompletion block makes this a bit tougher.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -856,9 +867,6 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
     ASDN::MutexLocker l(__instanceLock__);
     ASDisplayNodeAssert(ASHierarchyStateIncludesLayoutPending(_hierarchyState) == NO, @"Can't start a transition when one of the supernodes is performing one.");
   }
-
-  // Invalidate the current layout to be able to measure a new layout based onthe given constrained size
-  [self setNeedsLayout];
```
appleguy: @maicki I thin this line is still needed in the case this method is called directly without going through the one above?

PR #2642, opened 2016-11-21
[scottg reviewing] Replace fetch data with preload terminology

### LINKED ISSUE #2611: [Polish] Update comments and internal implementation from Fetch Data -> Preload terminology.
@nguyenhuy, could you rename anything that isn't being used to support the deprecated method?  We can probably leave the ASViewController internal call as it is for now as we don't have a recursive Preload method.

screen shot 2016-11-14 at 7 03 34 pm
screen shot 2016-11-14 at 7 03 14 pm

nguyenhuy:
- Deprecate `-fetchData` and `-clearFetchedData` in favor of `-didEnterPreloadState` and `-didExitPreloadState`.
- `-didEnterPreloadState` and `-didExitPreloadState` call the deprecated `-fetchData` and `-clearFetchedData` methods if they are overriden.
- Move `-setNeedsPreload`, `-recursivelyPreload` and `-recursivelyClearPreloadedData` to `ASDisplayNode+FrameworkPrivate.h`. Those methods now call `-didEnterPreloadState` and `-didExitPreloadState`.
- Update internal implementation, comments and tests

Resolves #2611

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1675,18 +1675,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   _nextLayoutInvalidationStyle = invalidationStyle;
 }
 
-#pragma mark - Memory Management
-
-- (void)clearContents
-{
-  [_rangeController clearContents];
-}
-
-- (void)clearFetchedData
-{
-  [_rangeController clearFetchedData];
-}
-
 #pragma mark - _ASDisplayView behavior substitutions
 // Need these to drive interfaceState so we know when we are visible, if not nested in another range-managing element.
 // Because our superclass is a true UIKit class, we cannot also subclass _ASDisplayView.
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -324,23 +324,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly, assign, getter=isInHierarchy) BOOL inHierarchy;
 
 /**
- * @abstract Indicates that the node should preload any external data, such as images.
- *
- * @discussion Subclasses may override this method to be notified when they should begin to preload. Fetching
- * should be done asynchronously. The node is also responsible for managing the memory of any data.
- * The data may be remote and accessed via the network, but could also be a local database query.
- */
-- (void)preload ASDISPLAYNODE_REQUIRES_SUPER;
-
-/**
- * Provides an opportunity to clear any preloaded data (e.g. remote / network or database-queried) on the current node.
- *
- * @discussion This will not clear data recursively for all subnodes. Either call -recursivelyClearPreloadedData or
- * selectively clear preloaded data.
- */
-- (void)clearPreloadedData ASDISPLAYNODE_REQUIRES_SUPER;
-
-/**
  * Provides an opportunity to clear backing store and other memory-intensive intermediates, such as text layout managers
  * on the current node.
  *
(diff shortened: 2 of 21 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1675,18 +1675,6 @@ - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged
...
-- (void)clearContents
-{
-  [_rangeController clearContents];
-}
-
-- (void)clearFetchedData
-{
-  [_rangeController clearFetchedData];
-}
-
```
nguyenhuy: These methods are not declared in ASCollectionView.h

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -324,21 +324,21 @@ NS_ASSUME_NONNULL_BEGIN
...
- * Provides an opportunity to clear any fetched data (e.g. remote / network or database-queried) on the current node.
+ * Provides an opportunity to clear any preloaded data (e.g. remote / network or database-queried) on the current node.
  *
- * @discussion This will not clear data recursively for all subnodes. Either call -recursivelyClearFetchedData or
- * selectively clear fetched data.
+ * @discussion This will not clear data recursively for all subnodes. Either call -recursivelyClearPreloadedData or
+ * selectively clear preloaded data.
  */
-- (void)clearFetchedData ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)clearPreloadedData ASDISPLAYNODE_REQUIRES_SUPER;
```
appleguy: Should we get rid of this method?  Is didExitPreloadState enough?  I think it probably is.

AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -324,21 +324,21 @@ NS_ASSUME_NONNULL_BEGIN
...
- * @abstract Indicates that the node should fetch any external data, such as images.
+ * @abstract Indicates that the node should preload any external data, such as images.
  *
- * @discussion Subclasses may override this method to be notified when they should begin to fetch data. Fetching
+ * @discussion Subclasses may override this method to be notified when they should begin to preload. Fetching
  * should be done asynchronously. The node is also responsible for managing the memory of any data.
  * The data may be remote and accessed via the network, but could also be a local database query.
  */
-- (void)fetchData ASDISPLAYNODE_REQUIRES_SUPER;
+- (void)preload ASDISPLAYNODE_REQUIRES_SUPER;
```
appleguy: This one is trickier than I thought -- because of setNeedsDataFetch.  I think we should probably eliminate this method for now, and just call didEnterPreloadState again (even if imbalanced) when setNeedsDataFetch is called.  Let's deprecate setNeedsDataFetch and circle back to a cleaner API for "re-fetching" when we have more time (and we'll add that better API before finalizing removal of the deprecated method).

appleguy:
This is very close, thanks Huy!  Yes, having setNeedsPreload be in +FrameworkPrivate seems good for now.

Do you think we can avoid exposing -preload and -clearPreloadedData as public overrides?  Can we drive the didEnter and didExit methods instead, although this is a bit less conceptually clean, to give us more flexibility to improve the API later?

nguyenhuy:
@appleguy I pushed a new commit that addresses your comments. Yes, we can drive `-didEnterPreloadState` and `-didExitPreloadState` instead. It results in a cleaner API surface. We will need to revisit `-setNeedsPreload`, `-recursivelyPreload` and `-recursivelyClearPreloadedData` after 2.0, although they are "framework private" now.

nguyenhuy:
@appleguy Btw, looks like pinterest-jenkins is taking holidays. Machines don't have holidays!

PR #2643, opened 2016-11-21
[ASTextNode] Remove caching of constrained size within ASTextNode in layout pass

maicki:
`ASTextNode` relies on the `_constrainedSize` property within `ASTextNode` across layout and display pass. Furthermore we cache the `ASTextKitRenderer` based on the constrained size and invalidate it in certain cases. Managing the creation and invalidation of the renderer is a huge overhead within `ASTextNode` and also results in bugs like calling `calculateSizeThatFits:` is not without any side effects and can cause misbehavior if called multiple times before drawing with a different bounds size as returned from it.

This PR has multiple purposes. It should move the `ASTextKitRenderer` to a thread safe components that can be used across multiple `ASTextNode's` as well as cache a specific amount of renderer within `ASTextNode` to be able to reuse one that is used multiple times across layout and display pass without having to invalidate it.

We already had a similar PR merged a couple of month ago, but the problem with it was that we didn't move over the `ASTextKitRenderer` to be an immutable component that can be used across threads. This PR will add this.

## Changes
- [x] Remove `_constrainedSize` dependency within `ASTextNode`
- [x] Make `ASTextKitRenderer` usable across multiple Threads / `ASTextNode`'s: This means instead of creating the internal components and size calculation lazily we create the components and compute the size immediately to have an immutable internal state.
- [x] Cache `ASTextKitRenderer` across `ASTextNode`s
- [x] Fix failing tests due to `containerInset` setting on `ASTextNode`


maicki:
@garrettmoon Ready for first review

appleguy:
@maicki Thanks!  Since this one is pretty detailed, could you review it in-person with Adlai tomorrow (or by VC?)  He studied the code closely last week and knows it more accurately than I do, so if you are both aligned on the strategy, let's move forward and get as much testing as possible (testing beats an extended period of review for this kind of change IMO).

Adlai-Holler:
@maicki Could you look at why the new snapshots are different? The new ones seem to have extra space on the right side of the text…

maicki:
@Adlai-Holler I briefly talked with @garrettmoon as the `ASTextNode` calculates it's size now on the bounds this is the new calculation of the `ASTextNode` and the snapshot needs to be updated

Adlai-Holler:
@maicki "Calculates its size now on the bounds" what do you mean? It looks like the new image is larger than the old one, and the text is slightly off-center?

Adlai-Holler:
This change to text node needs more testing before we bring it live – moving out of the 2.0.0 milestone.

maicki:
I looked into the failing test a bit more and the reason why the images are differ is that previously we don't use the truncated string for calculating the highlighted ranges, within this change as we truncate immediately after creating the renderer we highlight based on the truncated sting and so highlighted rects differ.

Adlai-Holler:
@maicki OK cool, let's update the reference image for the affected test by adding `self.recordMode = YES` at the top of the method and running it.

maicki:
@Adlai-Holler Rebased and added new reference image

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -49,62 +48,32 @@ - (instancetype)initWithTextKitAttributes:(const ASTextKitAttributes &)attribute
...
                                                    exclusionPaths:attributes.exclusionPaths
                                                   constrainedSize:shadowConstrainedSize];
+    
+    NSCharacterSet *avoidTailTruncationSet = attributes.avoidTailTruncationSet ? : _defaultAvoidTruncationCharacterSet();
+    _truncater = [[ASTextKitTailTruncater alloc] initWithContext:[self context]
+                                      truncationAttributedString:attributes.truncationAttributedString
+                                          avoidTailTruncationSet:avoidTailTruncationSet];
+    
+    // Calcualate size immediately
+    [self _calculateSize];
```
Adlai-Holler: We need to set up the font size adjuster.

AsyncDisplayKit/TextKit/ASTextKitRenderer.mm
```diff
@@ -49,62 +48,32 @@ - (instancetype)initWithTextKitAttributes:(const ASTextKitAttributes &)attribute
...
+        // We must inset the constrained size by the size of the shadower.
     CGSize shadowConstrainedSize = [[self shadower] insetSizeWithConstrainedSize:_constrainedSize];
+    
     _context = [[ASTextKitContext alloc] initWithAttributedString:attributes.attributedString
                                                     lineBreakMode:attributes.lineBreakMode
                                              maximumNumberOfLines:attributes.maximumNumberOfLines
                                                    exclusionPaths:attributes.exclusionPaths
                                                   constrainedSize:shadowConstrainedSize];
+    
+    NSCharacterSet *avoidTailTruncationSet = attributes.avoidTailTruncationSet ? : _defaultAvoidTruncationCharacterSet();
```
Adlai-Holler: The indentation in this section is off.

AsyncDisplayKit/ASTextNode.mm
```diff
@@ -50,6 +50,77 @@
...
+   __rendererCache = [[NSCache alloc] init];
+   __rendererCache.countLimit = 500; // 500 renders cache
+ });
+ return __rendererCache;
+}
+
+/**
+ The concept here is that neither the node nor layout should ever have a strong reference to the renderer object.
+ This is to reduce memory load when loading thousands and thousands of text nodes into memory at once. Instead
+ we maintain a LRU renderer cache that is queried via stack-allocated keys.
```
Adlai-Holler: We have to remove the "stack-allocated keys" bit. 

I know you're already aware of this, but just to comment for any other readers, I think the cost of allocating these key-objects is acceptable for the time-being. In the future we can optimize this later if we find it to be a major source of contention, but this is the correct design and the implementation is simple & reliable.

maicki:
@Adlai-Holler Addressed comments

@Adlai-Holler @appleguy Ready for another review and testing cc @garrettmoon

PR #2626, opened 2016-11-17
[ASVideoNode] Prevent premature placeholder image clear when setting the video asset

master-nevi:
Prevents flashing the background color in between the placeholder and the first frame of the video.

For more info refer to issue: https://github.com/facebook/AsyncDisplayKit/issues/2329

Issue #2329, opened 2016-09-30
[ASVideoNode] Video node flickers when asset is set after the thumbnail image is set


master-nevi:
Our app is using ASVideoNode with much success except for one hiccup which has been around since v1.9.81 . Currently our app gets and sets the video asset URLs on demand in order to avoid unnecessary network calls made by AVFoundation when creating AVAssets. Until the video plays we use a combination of `[ASNetworkImageNode defaultImage]` and `[ASNetworkImageNode URL]` for remote video thumbnails as well as `[ASImageNode image]` for local video thumbnails on `ASVideoNode`. 

Now as soon as we set the asset, internally `[ASVideoNode clearFetchedData]` is then called which clears the entire node momentarily until the video begins to play; which looks like a flicker.

This is demonstrated on my branch below in the video example project. In this project, the remote thumbnails are set immediately while the video assets don't get set until after a delay to simulate our scenario: 
https://github.com/master-nevi/AsyncDisplayKit/tree/dr/1.9.90-video-flicker/examples/Videos

Video demo: https://drive.google.com/file/d/0BxSzpqJ8X6gyQzVwZDZPdmo0cWs/view?usp=sharing

Our project has been able to work around this by only clearing the asset when a new asset is set, not the ASNetworkImageNode as demonstrated in this commit:
https://github.com/master-nevi/AsyncDisplayKit/commit/c05ff568f8307df1fcc8ffe7926c234edbedc13b

Please let me know if there's a better work around for this. We're willing to put time and effort in making sure this is right; please advise.

master-nevi:
Video demo: https://drive.google.com/file/d/0BxSzpqJ8X6gyQzVwZDZPdmo0cWs/view?usp=sharing

master-nevi:
Adding @appleguy @maxsz or @binl as they're mentioned to have AVFoundation expertise.

hannahmbanana:
@master-nevi: Thanks for the very detailed description! A quick question: could you provide the backtrace of when the clearFetchedData is called before the video is played? Is this something you are triggering, or is it being called by the framework? 

As a short term fix, you may be able to avoid this by setting defaultImage to be the local video frame whenever you have a local frame available. Use it for the generic thumbnail if that is all you have, or override it with a locally cached image if available. 

There is a known issue with the .image property being cleared if it is not the same as represented as the URL (#2275), because ASNetworkImageNode was originally designed for these two to be related.

master-nevi:
[...]

PR #2644, opened 2016-11-21
[ASVideoNode] Second attempt - Prevent premature placeholder image clear when setting the video asset

master-nevi:
Fix tests referenced in reverted pull request:
Reference https://github.com/facebook/AsyncDisplayKit/pull/2626


AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -505,7 +508,7 @@ - (AVAsset *)asset
 
 - (void)_setAndFetchAsset:(AVAsset *)asset url:(NSURL *)assetURL
 {
-  [self clearFetchedData];
+  [self _clearPlayer]; // Clear the player but not the underlying ASNetworkImageNode to avoid clearing the placeholder image and showing the background before the video starts.
   _asset = asset;
```
appleguy: @master-nevi could you describe a bit more the intended behavior here?

Should we only avoid clearing the placeholder in certain cases?  If you change the asset to a different one that has an unrelated contents, then it may be correct to clear the placeholder.

For example, should we check if URL (placeholder URL) is set?  And then, if it is NOT set, maybe clear the placeholder here?

I think the unit test should not need to call recursivelyClearFetchedData to get the right behavior, particularly if it is using the codepath where the asset's first frame is used as the thumbnail rather than a separate URL.

master-nevi: Ahh I didn't realized that a placeholder image was generated if ASVideoNode.URL was nil. That makes sense to check if URL is set:
```objective-c
- (void)_setAndFetchAsset:(AVAsset *)asset url:(NSURL *)assetURL
{
  if (self.URL != nil) {
    [self _clearPlayer]; // Clear the player but not the underlying ASNetworkImageNode to avoid clearing the placeholder image and showing the background before the video starts.
  } else {
    [self clearFetchedData]; // Clear player and underlying ASNetworkImageNode with generated placeholder
  }
  _asset = asset;
  _assetURL = assetURL;
  [self setNeedsDataFetch];
}
```


If we did it this way the tests should pass without modification.

master-nevi: Let me know if the code above is acceptable and I'll make the change.

master-nevi:
Rather than trying to patch this, it's probably more effective to do what was originally suggested here: https://github.com/facebook/AsyncDisplayKit/issues/2329#issuecomment-261432437 as use a separate node for the thumbnail.

PR #2645, opened 2016-11-21
Fix memory leaks in ASCollectionNode and ASTableNode

nguyenhuy:
- Don't capture the nodes in their view blocks just for getting their event logs
- Get the logs outside and pass them into the blocks instead


nguyenhuy:
Wrong fix. The event log is captured too early and not available yet. Closing.

PR #2646, opened 2016-11-21
Fix self capture in ASCollectionNode

austinlouden:


garrettmoon:
@austinlouden you cool with this one instead https://github.com/facebook/AsyncDisplayKit/pull/2645 ? @nguyenhuy also was fixing at the same time :)

garrettmoon:
Actually I think this is the correct patch. Can you apply the same thing to `ASTableNode`?

nguyenhuy:
@austinlouden Thanks for working on this!

garrettmoon:
test this please

PR #2648, opened 2016-11-22
[ASTableView] Not animated for end updates if animation is already going on within ASTableView

### LINKED ISSUE #1361: [ASTableView] Strange artifacts when changing the device orientation
Hi there,

I am seeing strange artifacts in `ASTableView` when changing the device orientation. There are some white boxes appearing on top of the content during the rotation animation. I have uploaded a video that shows what happens:

https://www.dropbox.com/s/wks3zy6kl1q5kj2/Untitled.mov?dl=0

Is there something I am doing wrong? I have seen the same effect in the kittens sample.

Thanks in advance!

maicki:
Fix strange artifacts in ASTableView are visible when changing the device orientation or other animations are going on while calling endUpdates

Resolves #1361

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -21,6 +21,7 @@
 #import "ASInternalHelpers.h"
 #import "ASLayout.h"
 #import "_ASDisplayLayer.h"
+#import "_ASCoreAnimationExtras.h"
 #import "ASTableNode.h"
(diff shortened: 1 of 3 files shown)


maicki:
@garrettmoon @appleguy @Adlai-Holler Ready for review

maicki:
@Adlai-Holler Do we have consider the same situation as above in `- [ASTableView requeryNodeHeights]`

Issue #2634, opened 2016-11-19
[ASCollectionView] Layout Inspector needs to fail earlier if not set up; ideally be provided by Layout object.


appleguy:
On some closeup views in Pinterest, we are hitting this assertion.  It's due to recent changes to require that scrollableDirections be overridden for any collection with a custom layout.  We do have a custom layout inspector in Pinterest but it seems it may not always be set.

Question is, why would we create this base class object at all if we know it will hit an assertion?  Can we throw the assertion at an earlier stage that is easier to debug, like maybe if a collection becomes visible without its layout inspector?

Two possible directions for improving how layoutInspector is set:

1. Instead of waiting for an external user to set the layoutInspector, can we take in a class name for it at initialization time?  In Pinterest, we have several places that set the .layoutInspector property but it seems easy to forget this line.

2. If a custom layout inspector always needs to be used for a particular custom layout, can we make it a category method on UICollectionViewLayout, and in the method check for class FlowLayout to create a default one + allow overrides in subclasses to return their custom one like Masonry needs to?

```
    } else {
      // Register the default layout inspector delegate for custom collection view layouts
      _defaultLayoutInspector = [[ASCollectionViewLayoutInspector alloc] initWithCollectionView:self];
    }

```
![screen shot 2016-11-19 at 2 01 58 pm](https://cloud.githubusercontent.com/assets/565251/20458712/c93686f0-ae60-11e6-819a-a53b2815e44b.png)

garrettmoon:
I believe it's valid to setup a collectionView without a layout inspector, correct me if I'm wrong @Adlai-Holler ?

Adlai-Holler:
The current layout inspector API requires a collection _view_. We can either modify the API to target collection _node_ or leave things as they are. Now that that assertion is removed and functionality is restored, we should focus on the rest of the 2.0 milestone and replace the layout inspector with a node-based async layout object afterwards.

Adlai-Holler:
Phew. Finally done in #3003

PR #2649, opened 2016-11-23
[ASCollectionNode] Modify Initialization API

Adlai-Holler:
- Remove `initWithFrame:collectionViewLayout:` and replace with `initWithCollectionViewLayout:`. This is more inline with `ASTableNode` and `ASDisplayNode`, rather than `UIView`.
- Add `initWithCollectionViewLayout:layoutInspector:` for when they have a layout inspector to provide. It is still acceptable to set the layout inspector later.

Note that the layout inspector API is likely to be replaced by a single asynchronous layout object, that serves as both our equivalent to `UICollectionViewLayout` and the layout inspector.

### DIFF
--- a/AsyncDisplayKit/ASCollectionNode.h
+++ b/AsyncDisplayKit/ASCollectionNode.h
@@ -33,19 +33,19 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @discussion Initializes and returns a newly allocated collection node object with the specified layout.
  *
- * @param layout The layout object to use for organizing items. The collection view stores a strong reference to the specified object. Must not be nil.
+ * @param layout The layout object to use for organizing items. The collection node stores a strong reference to the specified object. Must not be nil.
  */
 - (instancetype)initWithCollectionViewLayout:(UICollectionViewLayout *)layout;
 
 /**
  * Initializes an ASCollectionNode
  *
- * @discussion Initializes and returns a newly allocated collection node object with the specified frame and layout.
+ * @discussion Initializes and returns a newly allocated collection node object with the specified layout.
  *
- * @param frame The frame rectangle for the collection view, measured in points. The origin of the frame is relative to the superview in which you plan to add it. This frame is passed to the superclass during initialization.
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASCollectionNode.h
```diff
@@ -33,19 +33,19 @@ NS_ASSUME_NONNULL_BEGIN
...
- * @discussion Initializes and returns a newly allocated collection node object with the specified frame and layout.
+ * @discussion Initializes and returns a newly allocated collection node object with the specified layout.
  *
- * @param frame The frame rectangle for the collection view, measured in points. The origin of the frame is relative to the superview in which you plan to add it. This frame is passed to the superclass during initialization.
- * @param layout The layout object to use for organizing items. The collection view stores a strong reference to the specified object. Must not be nil.
+ * @param layout The layout object to use for organizing items. The collection node stores a strong reference to the specified object. Must not be nil.
+ * @param layoutInspector The layout inspector to use, or nil to use the ASDK default layout inspector.
  */
-- (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout;
+- (instancetype)initWithCollectionViewLayout:(UICollectionViewLayout *)layout layoutInspector:(nullable id<ASCollectionViewLayoutInspecting>)layoutInspector;
```
appleguy: @Adlai-Holler this may sound silly, but what do you think about including an initializer right below this, without another block comment, that is -initWithCollectionViewFlowLayout: ?  I think that would be more clear for folks who use the common case of a flow layout and don't know what this layoutInspector thing is -- and flow layouts most definitely aren't going away, at least not before the layoutInspector is :)

Adlai-Holler: True, I see how that would help users feel comfortable picking their initializer, but I think the one above this (`initWithCollectionViewLayout:`) is enough. I'm going to go ahead and land this, if you feel strongly about the flow layout initializer, let me know and we'll drop it in.

Adlai-Holler:
I've reverted this. See comments in #2634.

PR #2650, opened 2016-11-23
Expose internal singleton method as a class method vs instance method in ASPINRemoteImageDownloader

rahul-malik:
Currently setting an NSURLSessionConfiguration through ASPINRemoteImageManager
will crash. 

- PINRemoteImage is not linked with ASDK so it makes it difficult to make changes in this file without adding it manually into the Podfile.  
- Also I noticed the compiler will yell at you for `unrecognized selector` calls if you use the actual class name (`ASPINRemoteImageDownloader`) vs `[self class]`. I expected the build should fail since we should create the build in configurations where PINRemoteImage is available or unavailable.

### DIFF
--- a/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
+++ b/AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
@@ -103,15 +103,10 @@ static ASPINRemoteImageDownloader *sharedDownloader = nil;
 + (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfiguration *)configuration
 {
   NSAssert(sharedDownloader == nil, @"Singleton has been created and session can no longer be configured.");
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Details/ASPINRemoteImageDownloader.m
```diff
@@ -103,15 +103,10 @@ + (instancetype)sharedDownloader
 + (void)setSharedImageManagerWithConfiguration:(nullable NSURLSessionConfiguration *)configuration
 {
   NSAssert(sharedDownloader == nil, @"Singleton has been created and session can no longer be configured.");
-  __unused PINRemoteImageManager *sharedManager = [[self class] sharedPINRemoteImageManagerWithConfiguration:configuration];
+  __unused PINRemoteImageManager *sharedManager = [ASPINRemoteImageDownloader sharedPINRemoteImageManagerWithConfiguration:configuration];
```
appleguy: ideally use "self" here?

Issue #2620, opened 2016-11-17
Avoid creating placeholder image on the main thread.


garrettmoon:
@appleguy has said that #2610 will likely cause a performance regression because placeholderImage is now being created on the main thread. We'll need to figure out how to avoid that while still checking the contents on the main thread *or* creating a thread-safe contents.

appleguy:
So the great mystery is why this code worked since at least March without issues...in any case I will put up a PR shortly that should resolve the main performance issue.  This area has been due for some refactoring for some time, and my PR won't change that, but it'll allow us to read .contents in the case the node isn't loaded (since it's safe to do so at that time, and almost always will be nil).

appleguy:
Diff is up, we can close this when #2651 is landed.

appleguy:
This is fixed.

PR #2651, opened 2016-11-23
[ASDisplayNode] Create placeholders off main when possible.

appleguy:
This is a fix for https://github.com/facebook/AsyncDisplayKit/issues/2620,
which is a followup to https://github.com/facebook/AsyncDisplayKit/pull/2610

This will ensure thread affinity is respected for .contents but still run
placeholderImage off-main when possible (node isn't loaded).

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1306,21 +1306,27 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   // We generate placeholders at measureWithSizeRange: time so that a node is guaranteed to have a placeholder ready to go.
   // This is also because measurement is usually asynchronous, but placeholders need to be set up synchronously.
   // First measurement is guaranteed to be before the node is onscreen, so we can create the image async. but still have it appear sync.
-  if (_placeholderEnabled && [self _displaysAsynchronously]) {
+  if (_placeholderEnabled && !_placeholderImage && [self _displaysAsynchronously]) {
     
     // Zero-sized nodes do not require a placeholder.
     ASLayout *layout = _calculatedDisplayNodeLayout->layout;
     CGSize layoutSize = (layout ? layout.size : CGSizeZero);
-    if (CGSizeEqualToSize(layoutSize, CGSizeZero)) {
+    if (layoutSize.width * layoutSize.height <= 0.0) {
       return;
     }
-
-    if (!_placeholderImage) {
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1306,21 +1306,27 @@ - (void)_pendingLayoutTransitionDidComplete
...
       return;
     }
-
-    if (!_placeholderImage) {
+    
+    // If we've displayed our contents, we don't need a placeholder.
+    // Contents is a thread-affined property and can't be read off main after loading.
+    if (self.isNodeLoaded) {
       ASPerformBlockOnMainThread(^{
         if (self.contents == nil) {
```
maicki: Can we move the `self.contents` block higher or do we have to check it in here after running on main? Something like:

```Objc
if (self.contents == nil) {
  if (self.isNodeLoaded) {
    ASPerformBlockOnMainThread(_placeholderImage = [self placeholderImage];)
  } else {
   _placeholderImage = [self placeholderImage];
  }
}
```

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1306,21 +1306,27 @@ - (void)_pendingLayoutTransitionDidComplete
...
   // This is also because measurement is usually asynchronous, but placeholders need to be set up synchronously.
   // First measurement is guaranteed to be before the node is onscreen, so we can create the image async. but still have it appear sync.
-  if (_placeholderEnabled && [self _displaysAsynchronously]) {
+  if (_placeholderEnabled && !_placeholderImage && [self _displaysAsynchronously]) {
     
     // Zero-sized nodes do not require a placeholder.
     ASLayout *layout = _calculatedDisplayNodeLayout->layout;
     CGSize layoutSize = (layout ? layout.size : CGSizeZero);
-    if (CGSizeEqualToSize(layoutSize, CGSizeZero)) {
+    if (layoutSize.width * layoutSize.height <= 0.0) {
```
maicki: Can you addd a `()` around `layoutSize.width * layoutSize.height` just to make it a bit more clear

AsyncDisplayKit/ASDisplayNode.mm
Adlai-Holler: @maicki We can't read `contents` off the main thread =/

appleguy:
In a bit of a rush for Thanksgiving -- I agree it would be good to add parentheses when we have a moment to do so.

PR #2655, opened 2016-11-23
Re-add NSURLSessionConfiguration support and fix crashes

rahul-malik:
Tested locally with the Pinterest app. 

The issue that caused this to be revered in the first place was the value for `alternativeRepresentationProvider` was set to the `ASPINRemoteImageDownloader` class instead of an instance. This slipped when converting the singleton method to a class method.

Included in this PR:
- Revert "Revert "Expose internal singleton method as a class method vs instance method.""
- Fix assignment of alternativeRepresentationProvider to be the singleton
  instance of ASPINRemoteImageDownloader.


Adlai-Holler:
Nice!

PR #2657, opened 2016-11-25
[Layout Transition] Layout Transition Fixes

maicki:
Calling `layoutIfNeeded` before actually starting the layout transition pcoess can cause a measurement pass with the already new layout description to happen. This means instead of animating the nodes in and out as well as the position of nodes, it will just apply the new layout immediately without any animation.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -933,7 +933,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   ASDisplayNodeAssertMainThread();
 
   [self setNeedsLayout];
-  [self.view layoutIfNeeded];
     
   [self transitionLayoutWithSizeRange:[self _locked_constrainedSizeForLayoutPass]
                              animated:animated
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -319,7 +319,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   
   _defaultLayoutTransitionDuration = 0.2;
   _defaultLayoutTransitionDelay = 0.0;
-  _defaultLayoutTransitionOptions = UIViewAnimationOptionBeginFromCurrentState;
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -878,12 +878,13 @@ - (ASLayout *)layoutThatFits:(ASSizeRange)constrainedSize parentSize:(CGSize)par
 {
   ASDN::MutexLocker l(__instanceLock__);
  
-  // If multiple layout transitions are in progress it can happen that an invalid one is still trying to do a measurement
-  // before it get's cancelled. In this case we should not touch any layout and return a no op layout
+  // If one or multiple layout transitions are in flight it still can happen that size layout informationen are requested
```
Adlai-Holler: `that layout information is requested`

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1202,7 +1206,7 @@ - (void)animateLayoutTransition:(id<ASContextTransitioning>)context
     insertedSubnode.alpha = 0;
   }
 
-  [UIView animateWithDuration:self.defaultLayoutTransitionDuration delay:self.defaultLayoutTransitionDelay options:self.defaultLayoutTransitionOptions animations:^{
+  [UIView animateWithDuration:self.defaultLayoutTransitionDuration delay:self.defaultLayoutTransitionDelay options:kNilOptions animations:^{
```
Adlai-Holler: Why this change?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1051,6 +1049,12 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
       
       // Kick off animating the layout transition
       [self animateLayoutTransition:_pendingLayoutTransitionContext];
+        
+      // Mark transaction as finished
+      [self _finishOrCancelTransition];
+      
+      // As we prevented to do a layout pass while the layout transition was in flight
+      [self setNeedsLayout];
```
Adlai-Holler: Do we necessarily want to do this every time? Assuming it results in the same final layout, the UIKit layout pass that comes as a result of this, won't interfere with the inflight animations?

garrettmoon:
Merging now to get this in for testing. CI test fail is unrelated.

PR #2664, opened 2016-11-27
This patch enables memory warning support on ASRangeController by def…

garrettmoon:
…ault

It also fixes a couple subtle bugs:
1. If a range controller update was in flight and you were manually setting the range
mode, you could actually clear out the range mode by calling updateCurrentRangeWithMode:
This is fixed by setting _didUpdateCurrentRange = YES if _needsRangeUpdate is YES.
2. Calling setNeedsUpdate after calling updateCurrentRangeWithMode: can cause the
current range mode to get cleared out. This is because updateCurrentRangeWithMode:
will not set _didUpdateCurrentRange if _currentRangeMode == rangeMode.

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeController.mm
+++ b/AsyncDisplayKit/Details/ASRangeController.mm
@@ -36,7 +36,7 @@
   BOOL _layoutControllerImplementsSetViewportSize;
   NSSet<NSIndexPath *> *_allPreviousIndexPaths;
   ASLayoutRangeMode _currentRangeMode;
-  BOOL _didUpdateCurrentRange;
+  BOOL _preserveCurrentRangeMode;
   BOOL _didRegisterForNodeDisplayNotifications;
   CFTimeInterval _pendingDisplayNodesTimestamp;
   
@@ -62,7 +62,7 @@ static UIApplicationState __ApplicationState = UIApplicationStateActive;
   
   _rangeIsValid = YES;
   _currentRangeMode = ASLayoutRangeModeInvalid;
-  _didUpdateCurrentRange = NO;
+  _preserveCurrentRangeMode = NO;
   
   [[[self class] allRangeControllersWeakSet] addObject:self];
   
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -145,6 +149,10 @@ - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
     _didUpdateCurrentRange = YES;
 
     [self setNeedsUpdate];
+  } else if (_needsRangeUpdate) {
+    // If _needsRangeUpdate is YES, but we don't set didUpdateCurrentRange, it
+    // will be overridden.
+    _didUpdateCurrentRange = YES;
```
garrettmoon: Should we just *always* set this? Otherwise if we don't and someone calls setNeedsUpdate, our mode will be cleared out.

appleguy: I suppose so, yes.  We should also rename the instance variable to something more descriptive like, _preserveRangeMode or _rangeModeSetManually

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -532,7 +540,7 @@ + (void)registerSharedApplicationNotifications
   [center addObserver:self selector:@selector(willEnterForeground:) name:UIApplicationWillEnterForegroundNotification object:nil];
 }
 
-static ASLayoutRangeMode __rangeModeForMemoryWarnings = ASLayoutRangeModeVisibleOnly;
+static ASLayoutRangeMode __rangeModeForMemoryWarnings = ASLayoutRangeModeLowMemory;
```
appleguy: Could you file a 2.1 or later task to revisit this?  The ideal behavior would be VisibleOnly for things with visibilityDistance of 0 or 1, but to use Low Memory if visibilityDistance >= 2 — but then, once something becomes 2 -> 1, it needs to automatically trigger going from LowMemory to VisibleOnly.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -145,6 +149,10 @@ - (void)updateCurrentRangeWithMode:(ASLayoutRangeMode)rangeMode
     _didUpdateCurrentRange = YES;
```
appleguy: Does this preserve manual mode logic work if a controller is hidden deep in a stack (e.g. visibilityDistance 4), but one of its data models updates and triggers a slight height resize or an item to be added to a collection?  If that happened and it re-flows its ranges, would it then kick back into Minimum or Normal mode?  Even though it may seem minor, such behavior could result in really impactful memory problems depending on how an app listens to model updates.

garrettmoon: The answer is: I'm not sure.

We likely need to audit this system and bring visibilityDepth knowledge to the range controller post 2.0

Issue #2665, opened 2016-11-28
Examples Build Issue


Adlai-Holler:
http://jenkinsmaster-vpc-007.ec2.pin220.com/job/AsyncDisplayKit/350/console

`xcodebuild: error: The workspace named "Sample" does not contain a scheme named "Sample". The "-list" option can be used to find the names of the schemes in the workspace.`

Could you take a look?

hannahmbanana:
Fixed!

PR #2669, opened 2016-11-28
Use the Method Form for UIViewController.viewLoaded – Maintains Xcode Backwards Compatibility

Adlai-Holler:
Syntax highlighting is still broken in Xcode 8 so I sometimes prefer using Xcode 7.3.1. Previously `isViewLoaded` was a method, not a property, and this uses that method explicitly.


Adlai-Holler:
CI failure is unrelated, is being investigated in https://github.com/facebook/AsyncDisplayKit/issues/2665

PR #2675, opened 2016-11-29
Improve Batch Fetching Call Pattern to Reduce Fetch Calls

Adlai-Holler:
- Add a (previously-failing) unit test that when the content starts off not filling the screen, and then a batch fetch is issued that results in the new content filling the screen, we do not start another batch fetch.
- Remove `didCompleteUpdatesInRangeController:`
  - This required callback is to inform ASCollectionView/ASTableView that the range controller has updated the ranges of visible nodes.
  - Its name is easily confused with `rangeController:didEndUpdatesAnimated:completion:`
  - It was previously used to schedule a batch fetch check when the ranges were updated.
  - Batch fetching here has two issues:
    - It doesn't really make sense. The range controller updating the interface state of nodes has no impact on whether it is appropriate to batch fetch or not.
    - It may cause batch fetches to be issued at bad times. When we issue an update to UICollectionView, if the update caused the visible cells to change, then we synchronously update the range controller (in order to make our API guarantees about visibility). However, UICollectionView does not immediately update its `contentSize` and so we may _think_ we're still near the end, when in reality a new batch of content has just arrived and we do not need to fetch again.
- Pass `self.scrollDirection` for the `scrollDirection` param, rather than passing `self.scrollableDirections` (!!) 
- Add a new `scrollableDirections` arg to the batch checking algorithm, so that even if the user is not scrolling (content-not-filling-bounds case) we can still use the correct axes when computing e.g. "total length."
- Update the batch fetching algorithm to use `CGRectIsEmpty` rather than comparing the rect to zero. It is common for UICollectionView to have e.g. `(0, -64, 0, 0)` early in its lifecycle due to contentInset.
- Update batch checking algorithm to correctly issue a batch update in the small-content case even if not scrolling.
  - This was previously working due to us passing `self.scrollableDirections` rather than `self.scrollDirection`. Two wrongs made a right in this isolated case 😛 .
- Update the prior unit tests to be compatible with the new code.

I am about to start testing this with Pinterest, but I'm very confident in the code and I think it's important to land this ASAP. @garrettmoon

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1080,6 +1080,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   ASInterfaceState interfaceState = [self interfaceStateForRangeController:_rangeController];
   if (ASInterfaceStateIncludesVisible(interfaceState)) {
     [_rangeController updateCurrentRangeWithMode:ASLayoutRangeModeFull];
+    [self _checkForBatchFetching];
   }
   
   for (_ASCollectionViewCell *collectionCell in _cellsForVisibilityUpdates) {
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1103,7 +1103,7 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 
   if (targetContentOffset != NULL) {
     ASDisplayNodeAssert(_batchContext != nil, @"Batch context should exist");
-    [self _beginBatchFetchingIfNeededWithScrollView:self forScrollDirection:[self scrollDirection] contentOffset:*targetContentOffset];
+    [self _beginBatchFetchingIfNeededWithContentOffset:*targetContentOffset];
   }
   
   if (_asyncDelegateFlags.scrollViewWillEndDragging) {
@@ -1261,12 +1261,12 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
     return;
   }
   
-  [self _beginBatchFetchingIfNeededWithScrollView:self forScrollDirection:[self scrollableDirections] contentOffset:self.contentOffset];
+  [self _beginBatchFetchingIfNeededWithContentOffset:self.contentOffset];
 }
 
-- (void)_beginBatchFetchingIfNeededWithScrollView:(UIScrollView<ASBatchFetchingScrollView> *)scrollView forScrollDirection:(ASScrollDirection)scrollDirection contentOffset:(CGPoint)contentOffset
+- (void)_beginBatchFetchingIfNeededWithContentOffset:(CGPoint)contentOffset
 {
-  if (ASDisplayShouldFetchBatchForScrollView(self, scrollDirection, contentOffset)) {
+  if (ASDisplayShouldFetchBatchForScrollView(self, self.scrollDirection, self.scrollableDirections, contentOffset)) {
     [self _beginBatchFetching];
   }
 }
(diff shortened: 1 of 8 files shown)


Adlai-Holler:
Testing with Pinterest went great! Let's move forward on this.

AsyncDisplayKit/Details/ASRangeController.h
```diff
@@ -160,13 +160,6 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)rangeController:(ASRangeController * )rangeController didEndUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion;
 
 /**
- * Completed updates to cell node addition and removal.
- *
- * @param rangeController Sender.
- */
-- (void)didCompleteUpdatesInRangeController:(ASRangeController *)rangeController;
```
garrettmoon: I doubt this was used publicly, but do you think we should deprecate instead of just removing as it was exposed?

AsyncDisplayKit/Private/ASBatchFetching.m
```diff
@@ -37,19 +38,14 @@ BOOL ASDisplayShouldFetchBatchForContext(ASBatchContext *context,
...
   // No fetching for null states
-  if (leadingScreens <= 0.0 || CGRectEqualToRect(bounds, CGRectZero)) {
+  if (leadingScreens <= 0.0 || CGRectIsEmpty(bounds)) {
     return NO;
   }
 
   CGFloat viewLength, offset, contentLength;
 
-  if (ASScrollDirectionContainsDown(scrollDirection)) {
+  if (ASScrollDirectionContainsVerticalDirection(scrollableDirections)) {
```
garrettmoon: Not that it worked before, and I'm not sure how it would, but does this preclude batch fetching from working in collection views which scroll in both directions?

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -174,6 +174,15 @@ @interface ASCollectionView () <ASRangeControllerDataSource, ASRangeControllerDe
...
   BOOL _superIsPendingDataLoad;
+
+  /**
+   * It's important that we always check for batch fetching at least once, but also
+   * that we do not check for batch fetching for empty updates (as that may cause an infinite
+   * loop of batch fetching, where the batch completes and performBatchUpdates: is called without
+   * actually making any changes.) So to handle the case where a collection is completely empty
+   * (0 sections) we always check at least once after each update (initial reload is the first update.)
+   */
+  BOOL _hasEverCheckedForBatchFetchingDueToUpdate;
```
appleguy: @Adlai-Holler in a future factoring, this suggests we may want to put Fetching logic in some kind of encapsulation (like an ASBatchFetcher object) to maintain this state and simplify calls to check — so we can use it in table / collection the same way, and maybe offer it as a low-level API for custom scroll view implementers.

AsyncDisplayKit/Private/ASBatchFetching.m
appleguy: Good observation, although I think in both cases we would have preferred vertical and not done any horizontal.  For the current API limitations I think this is OK.

AsyncDisplayKit/Private/ASBatchFetching.m
```diff
@@ -61,9 +57,14 @@ BOOL ASDisplayShouldFetchBatchForContext(ASBatchContext *context,
 
   // target offset will always be 0 if the content size is smaller than the viewport
   BOOL hasSmallContent = offset == 0.0 && contentLength < viewLength;
+  if (hasSmallContent) {
```
appleguy: This is a descriptive variable name, I think it would be fine to use as return (hasSmallContent || (isScrollingTowardEnd && remaining <= trigger))

appleguy:
@Adlai-Holler a very important change, great attention to detail here!

PR #2676, opened 2016-11-29
Revert constrainedSize changes that should have never made it in

maicki:


appleguy:
@maicki thanks, this looks correct to me!

Issue #2658, opened 2016-11-25
[ASCellNode] didExitVisibleState is not called when a visible cell is deleted


flovouin:
Hi guys,

I've had an issue for some time now. I recently upgraded to `master`, but the problem is still here: `didExistVisibleState` (and the previous visibility method for < 2.0) is not called on a visible `ASCellNode` when it is deleted using the `ASCollectionNode` (and previously `ASCollectionView`) method `deleteItemsAtIndexPaths`.
When scrolling, the visibility methods are called correctly. When presenting / pushing another controller, the behaviour is also correct. It only happens when deleting visible cells. Please note that no other data management method is called either (e.g. `clearFetchedData`). As the cell nodes in my application are pretty heavy, I would very much like to be notified when they are permanently deleted.

I made a simple test project that demonstrates the issue: https://github.com/flovouin/ASDK-CellVisibleState
Running the code should output in the console:
```
Adding lots of cells
Removing cells
Showing another controller
Goodbye from cell at index 0!
Goodbye from a cell that was no longer in the collection!
Goodbye from a cell that was no longer in the collection!
Goodbye from a cell that was no longer in the collection!
Goodbye from a cell that was no longer in the collection!
Goodbye from a cell that was no longer in the collection!
Goodbye from a cell that was no longer in the collection!
```
The visibility method for the removed cells is only called once the whole collection disappears. I would expect the following output:
```
Adding lots of cells
Removing cells
Goodbye from cell at index 0!
Goodbye from cell at index 1!
Goodbye from cell at index 2!
Goodbye from cell at index 3!
Goodbye from cell at index 4!
Goodbye from cell at index 5!
Showing another controller
Goodbye from cell at index 0!
```

If you need anything else, don't hesitate to ask.

Cheers,

Flo

appleguy:
@garrettmoon this seems very important to fix, because we rely on these calls for accurate logging.  This would probably be P1 if we had more time.

We may need to punt 2.0.1, but do you think we should spend an hour or two investigating a minimal fix before making this call?

hannahmbanana:
ASRangeControllerLogging:

```
2016-11-27 15:07:18.623 VisibleStateTest[42768:1989050] <ASCollectionNode: 0x7f82b2e027b0> - visible: YES
Adding lots of cells
[...]

PR #2680, opened 2016-11-29
[WIP] [ASCellNode] Remove view from superview if the UICollectionViewCell is not displayed any more

maicki:
`UICollectionViewCell`s are usually reused and not deallocated. Currently the `ASCellNode` view only get's removed from it's superview if the `UICollectionViewCell` get's reused or the `UICollectionView` get's deallocated. The problem with this if `UICollectionViewCell`s get removed, but not immediately reused. All of the `ASCellNode`'s still stay alive as the view is still a subview of the `UICollectionView` content view and the `ASCellNode` view holds a back pointer to the `ASCellNode`. One of the problems with this is that visibility calls are not called or later called e.g. if `UICollectionView` gets deallocated.

To be able to solve this we should remove the `ASCellNode` view from the superview within `...didEndDisplayingCell:..` in `ASCollectionView` and `ASTableView`.

Resolves: #2658

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -885,6 +885,10 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   cellNode.scrollView = nil;
   cell.layoutAttributes = nil;
+
+  // It's important to remove our view now, so that the node will be released immediately.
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
I know that the 2.0 launch is resource-intensive so @garrettmoon how should we move forward with this?

Adlai-Holler:
Superseded

PR #2684, opened 2016-11-30
Remove Deleted Cell Node's View from Cell Immediately

### LINKED ISSUE #2658: [ASCellNode] didExitVisibleState is not called when a visible cell is deleted

Adlai-Holler:
Supersedes #2680 resolves #2658 This has been tested and will get merged after CI.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -931,6 +931,12 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   
   cellNode.scrollView = nil;
   cell.layoutAttributes = nil;
+
+  // If this item was deleted, we should go ahead and remove it from the cell's content view so that it will be marked
+  // invisible immediately. The range controller currently does not manage the interface state for nodes that have been deleted.
+  if (cellNode.indexPath == nil) {
+    [cellNode.view removeFromSuperview];
+  }
 }
 
 - (void)collectionView:(UICollectionView *)collectionView willDisplaySupplementaryView:(UICollectionReusableView *)view forElementKind:(NSString *)elementKind atIndexPath:(NSIndexPath *)indexPath
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -898,6 +898,12 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   [_cellsForVisibilityUpdates removeObject:cell];
   
   cellNode.scrollView = nil;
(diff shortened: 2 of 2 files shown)


garrettmoon:
test this please

hannahmbanana:
test this please

appleguy:
test this please

Adlai-Holler:
The call pattern that this new test case exercises is very interesting and should be investigated. Landing a change this significant, this late in the game for 2.0 is unwise and not justified by the benefits. Punting to 2.0.1 so I can investigate more thoroughly.

appleguy:
10:00:25 Failing tests:
10:00:25 	-[ASTableViewTests testIssue2252]
10:00:25 ** TEST FAILED **

Adlai-Holler:
This shouldn't be landed. Given what the development of the thrash tester has revealed, we should instead change how we drive the cell node visibility bit – either have the range controller remember which nodes _were_ visible and go back and clear them when it picks a new batch, or drive them directly by `willDisplayCell:` and `didEndDisplayingCell:` callbacks. Probably the former.

PR #2686, opened 2016-12-01
Fixed strong self dereferences

ngoctr:


appleguy:
@ngoctr thanks a lot for looking at these details!  Glad to have you as part of our community :)

Adlai-Holler:
Hi @ngoctr, thanks for the diff! As scary as it looks, `__typeof__(self)` doesn't retain `self` – it's evaluated by the compiler and replaced with the corresponding type like `ASDisplayNode *`. The code before the diff is simpler, so we should leave it as-is. Thanks again for looking out – cheers!

ngoctr:
Thanks @Adlai-Holler for the tip! Would be happy to be a part of the ASDK community ;-)

PR #2689, opened 2016-12-01
2611 rename fetch data

garrettmoon:
Rebasing for Huy

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1681,18 +1681,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
   _nextLayoutInvalidationStyle = invalidationStyle;
 }
 
-#pragma mark - Memory Management
-
-- (void)clearContents
-{
-  [_rangeController clearContents];
-}
-
-- (void)clearFetchedData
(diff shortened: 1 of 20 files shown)


garrettmoon:
This is a rebase of https://github.com/facebook/AsyncDisplayKit/pull/2642

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1681,18 +1681,6 @@ - (void)nodeDidRelayout:(ASCellNode *)node sizeChanged:(BOOL)sizeChanged
   _nextLayoutInvalidationStyle = invalidationStyle;
 }
 
-#pragma mark - Memory Management
-
-- (void)clearContents
-{
-  [_rangeController clearContents];
```
appleguy: @garrettmoon I thought these methods could get called by exiting the window -- I think we probably still need to support this for 2.0 compatibility, unless apps don't compile when they try to create an ASCV directly?  Is that direct-CV-init just deprecated, or gone?

appleguy:
@garrettmoon looks good to me - only wondering about ASTableView and ASCollectionView's (node-free) memory management behavior.  I believe they listen to memory warnings and ask the range controller to clear if they are not visible.

Issue #2673, opened 2016-11-28
[Layout Transition] When creating a new node in layoutSpecThatFits:, an animated transition may crash if no reference held to it.


ay8s:
While not recommended if you setup a node within layoutSpecThatFits: within a ASTableNode ASCellNode the project with crash with an assertion ('A flattened layout must consist exclusively of node sublayouts').

Removing the ASImageNode within the Sample to be setup before layoutSpecThatFits: resolves the issue. Other cases where this may be done is with hairline separators within cell nodes.

Scott suggested a relatively small change like making a strong retain in one of the transition objects should fix this up.

Sample Project: https://cl.ly/3m3p07361V2Z

garrettmoon:
We have a good idea for fixing this, but it's not going to make 2.0, moving to 2.0.1

PR #2694, opened 2016-12-02
[Layout] Don't crash if layout elements are created in `layoutSpecThatFits:`

maicki:
Referencing `ASLayoutElement`'s in `layoutSpecThatFits:` without holding a strong reference can to lead crashes if the measurement pass happens in a background thread and applied later in a layout pass. 

To assure that layout elements that are referenced in sublayouts life while the root layout is alive we should hold a strong reference to the sub layout layout elements while the root layout is alive.

Resolves: #2673

### DIFF
--- a/AsyncDisplayKit/Layout/ASLayout.mm
+++ b/AsyncDisplayKit/Layout/ASLayout.mm
@@ -50,7 +50,7 @@ static inline NSString * descriptionIndents(NSUInteger indents)
 /*
  * Caches all sublayouts if set to YES or destroys the sublayout cache if set to NO. Defaults to YES
  */
-@property (nonatomic, assign) BOOL cacheSublayoutLayoutElements;
+@property (nonatomic, assign) BOOL retainSublayoutLayoutElements;
 
 /**
  * Array for explicitly retain sublayout layout elements in case they are created and references in layoutSpecThatFits: and no one else will hold a strong reference on it
@@ -99,7 +99,7 @@ static inline NSString * descriptionIndents(NSUInteger indents)
 
     _sublayouts = sublayouts != nil ? [sublayouts copy] : @[];
     _flattened = NO;
-    _cacheSublayoutLayoutElements = NO;
+    _retainSublayoutLayoutElements = NO;
   }
   
   return self;
@@ -152,18 +152,21 @@ static inline NSString * descriptionIndents(NSUInteger indents)
 
 #pragma mark - Sublayout Elements Caching
 
-- (void)setCacheSublayoutLayoutElements:(BOOL)cacheSublayoutLayoutElements
+- (void)setRetainSublayoutLayoutElements:(BOOL)retainSublayoutLayoutElements
 {
-  _cacheSublayoutLayoutElements = cacheSublayoutLayoutElements;
-  
-  if (cacheSublayoutLayoutElements == NO) {
-    _sublayoutLayoutElements = nil;
-  } else {
-    // Add sublayouts layout elements to an internal array to retain it while the layout lives
-    if (_sublayouts.count > 0) {
-      _sublayoutLayoutElements = [NSMutableArray array];
-      for (ASLayout *sublayout in _sublayouts) {
-        [_sublayoutLayoutElements addObject:sublayout.layoutElement];
+  if (_retainSublayoutLayoutElements != retainSublayoutLayoutElements) {
+    _retainSublayoutLayoutElements = retainSublayoutLayoutElements;
+    
+    if (retainSublayoutLayoutElements == NO) {
+      _sublayoutLayoutElements = nil;
+    } else {
+      // Add sublayouts layout elements to an internal array to retain it while the layout lives
+      NSUInteger sublayoutCount = _sublayouts.count;
+      if (sublayoutCount > 0) {
+        _sublayoutLayoutElements = [NSMutableArray arrayWithCapacity:sublayoutCount];
+        for (ASLayout *sublayout in _sublayouts) {
+          [_sublayoutLayoutElements addObject:sublayout.layoutElement];
+        }
       }
     }
   }
@@ -191,7 +194,6 @@ static inline NSString * descriptionIndents(NSUInteger indents)
     if (self != context.layout && context.layout.type == ASLayoutElementTypeDisplayNode) {
       ASLayout *layout = [ASLayout layoutWithLayout:context.layout position:context.absolutePosition];
       layout.flattened = YES;
-      layout.cacheSublayoutLayoutElements = YES;
       [flattenedSublayouts addObject:layout];
     }
     
@@ -205,7 +207,7 @@ static inline NSString * descriptionIndents(NSUInteger indents)
   }
   
   ASLayout *layout = [ASLayout layoutWithLayoutElement:_layoutElement size:_size position:CGPointZero sublayouts:flattenedSublayouts];
-  layout.cacheSublayoutLayoutElements = YES;
+  layout.retainSublayoutLayoutElements = YES;
   return layout;
 }
 
--- a/AsyncDisplayKit/Layout/ASLayout.mm
+++ b/AsyncDisplayKit/Layout/ASLayout.mm
@@ -94,15 +94,8 @@ static inline NSString * descriptionIndents(NSUInteger indents)
 
     _sublayouts = sublayouts != nil ? [sublayouts copy] : @[];
     _flattened = NO;
-    
-    // Add sublayouts layout elements to an internal array to retain it while the layout lives
-    if (sublayouts.count > 0) {
-      _sublayoutLayoutElements = [NSMutableArray array];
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -88,6 +94,14 @@ - (instancetype)initWithLayoutElement:(id<ASLayoutElement>)layoutElement
 
     _sublayouts = sublayouts != nil ? [sublayouts copy] : @[];
     _flattened = NO;
+    
+    // Add sublayouts layout elements to an internal array to retain it while the layout lives
+    if (sublayouts.count > 0) {
```
maicki: We could do some optimization here to only retain the sub layout elements if the layout element type is a node:

```objc
...
if (_layoutElementType == ASLayoutElementTypeDisplayNode && sublayouts.count > 0) {
...
```

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -47,6 +47,11 @@ @interface ASLayout () <ASDescriptionProvider>
  */
 @property (nonatomic, getter=isFlattened) BOOL flattened;
 
+/**
+ * Array for explicitly retain sublayout layout elements in case they are created and references in layoutSpecThatFits: and no one else will hold a strong reference on it
+ */
+@property (nonatomic, strong) NSMutableArray *sublayoutLayoutElements;
```
appleguy: @maicki what do you think about making this private / an internal ivar?

Should we do the optimization to check layoutElementType, since it's quick?

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -88,6 +94,14 @@ - (instancetype)initWithLayoutElement:(id<ASLayoutElement>)layoutElement
 
     _sublayouts = sublayouts != nil ? [sublayouts copy] : @[];
     _flattened = NO;
+    
+    // Add sublayouts layout elements to an internal array to retain it while the layout lives
+    if (sublayouts.count > 0) {
+      _sublayoutLayoutElements = [NSMutableArray array];
+      for (ASLayout *sublayout in sublayouts) {
+        [_sublayoutLayoutElements addObject:sublayout.layoutElement];
```
appleguy: @maicki where do we release these elements to break the retain cycle?

AsyncDisplayKit/Layout/ASLayout.mm
appleguy: @maicki this is pretty awesome!  Would it be possible to put this retain operation into the setIsFlattened:YES?  Then we could assert that all sublayouts are nodes, and it would avoid any array and retain overhead for the many ASLayoutSpecs that are created during the measurement process.

The code that would trigger it is below, but I think the retain operation would happen in the flattened setter.  I think this would still make sense from a logical / usage perspective, certainly because this is a) an internal behavior and b) all externally accessible long-lived ASLayouts are flattened.

       ASLayout *layout = [ASLayout layoutWithLayout:context.layout position:context.absolutePosition];
       layout.flattened = YES;

The one question is: what happens if a user calls layoutThatFits: and saves a reference to the resulting ASLayout?  Do we always flatten those layouts?  If so, we're good.

ngoctr:
Thanks for working on this. It will help a lot.

AsyncDisplayKit/Layout/ASLayout.mm
nguyenhuy: Another (better?) way is to trigger strong refences gathering here: https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASLayout.mm#L174

AsyncDisplayKit/Layout/ASLayout.mm
appleguy: We could add the type <ASLayoutElement> here

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -98,12 +105,41 @@ - (instancetype)init
   return [self init];
 }
 
+#pragma mark - 
+
+- (void)cacheSublayouts
```
appleguy: Since this functionality is only used in one place, and it would be unusual / maybe dangerous to call more than once — what do you think about putting these lines directly in the initializer method that uses them?

Otherwise, let's make it a property -- and in setCacheSublayouts: we can either set up the array or destroy it.

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -157,7 +194,7 @@ - (ASLayout *)filteredNodeLayoutTree
     queue.pop_front();
 
     if (self != context.layout && context.layout.type == ASLayoutElementTypeDisplayNode) {
-      ASLayout *layout = [ASLayout layoutWithLayout:context.layout position:context.absolutePosition];
+      ASLayout *layout = [ASLayout layoutWithLayout:context.layout position:context.absolutePosition ];
```
appleguy: Remove the extra space added at end

appleguy:
@maicki to @nguyenhuy 's point -- I think it is true this may extend node lifecycle, and we should at least schedule another 2.0.1 task to address that.

One strategy would be to schedule a runloop queue item to clear the pointer to the ASLayout, if by the time the queue item runs, it still should be cleared.  This is similar to how we call didExitDisplayState (clearContents) on NON-range managed nodes -- we schedule it the moment they're removed from the parent, but we re-check once the block runs, and sometimes they've re-entered a new parent by that time so we skip the clearing.

That said, the main concern would be ultra high volumes of block dispatches and retain / release if this happened on every node.  So as long as we make sure that, in practice, it doesn't occur when a full node hierarchy is being destroyed (e.g. the parent of a large tree is released) -- and we have some way to detect this case (e.g. "removeSupernodeDueToDyingParent" - and not schedule the cleanup in that path)...then it should be a good solution for the more rare, manual removeFromSupernode calls.

appleguy:
@maicki after you address my two other comments, feel  free to land this and do another diff next week with followup for the memory lifetime issue - I don't think it is a blocker.

maicki:
@appleguy Can you get it another look. I decided to move a property for the time for caching the sublayouts as we called the `cacheSublayouts` in a class initializer and so we had to use `->` to access or set the instance variables of the layout. I also removed a couple of added class initializer now to reduce the changes.

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -137,6 +150,25 @@ + (instancetype)layoutWithLayout:(ASLayout *)layout position:(CGPoint)position
                             sublayouts:layout.sublayouts];
 }
 
+#pragma mark - Sublayout Elements Caching
+
+- (void)setCacheSublayoutLayoutElements:(BOOL)cacheSublayoutLayoutElements
+{
+  _cacheSublayoutLayoutElements = cacheSublayoutLayoutElements;
```
appleguy: It's probably not crucial, but we may want to short-circuit if the same value is set.

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -137,6 +150,25 @@ + (instancetype)layoutWithLayout:(ASLayout *)layout position:(CGPoint)position
...
+  
+  if (cacheSublayoutLayoutElements == NO) {
+    _sublayoutLayoutElements = nil;
+  } else {
+    // Add sublayouts layout elements to an internal array to retain it while the layout lives
+    if (_sublayouts.count > 0) {
+      _sublayoutLayoutElements = [NSMutableArray array];
```
appleguy: ultra-minor -- NSUInteger sublayoutCount = _sublayouts.count, if (sublayoutCount > 0), arrayWithCapacity:_sublayouts.count

AsyncDisplayKit/Layout/ASLayout.mm
```diff
@@ -159,6 +191,7 @@ - (ASLayout *)filteredNodeLayoutTree
     if (self != context.layout && context.layout.type == ASLayoutElementTypeDisplayNode) {
       ASLayout *layout = [ASLayout layoutWithLayout:context.layout position:context.absolutePosition];
       layout.flattened = YES;
+      layout.cacheSublayoutLayoutElements = YES;
```
appleguy: Nice.  This probably makes sense to keep like this, although we could make it implied by flattened = YES.

Should we call it .retainsSublayoutElements?  It's maybe not as clear as SublayoutLayoutElements, but using Retains instead of Cache and shortening it will probably make it understandable still.

nguyenhuy: I'm not sure why we need this. Retaining "grand" subnodes should be the job of each subnode, no?

maicki: @nguyenhuy Good catch

Issue #2699, opened 2016-12-02
[ASNetworkImageNode] Add assertion against externally setting .image (or make read-only)


appleguy:
It's a common (and easy-to-make) mistake to set the .image of a NetworkImageNode.  This doesn't work well because it gets cleared in clearFetchedData.

Devs should use defaultImage instead.  We should:

- For now add an assertion since I don't think it's ever valuable to set .image currently
- Consider if we should file a task to make this behavior more automatic / correct, e.g. either setting defaultImage for the user or changing the implementation to not clear .image if there isn't a URL.

appleguy:
This will be useful for the occasional Pinterest developer too, with PIImageNode.

PR #2708, opened 2016-12-04
[ASImageNode] Add handling of setting .image in specific ASImageNode subclasses

maicki:
We had some confusion in our Slack channel as well as a P1 bug in Pinterest due to setting an image directly to the `image` property of an `ASNetworkImageNode` that get's blown away if the node get's into the `didExitPreloadState`. This PR addresses this issue for `ASNetworkImageNode` in there we make the following behavior changes:

- Setting an image directly will set a flag: `_imageWasSetExternally` and the node will act like a plain `ASImageNode`
- In `didExitPreloadState` we check if the `_imageWasSetExternally == true` and just return and don't blow away the image anymore. Same as the `ASImageNode`
- Setting an URL aftewards will transform the node to be an image node that manages the image property internally based on the `defaultImage` and the image that was fetched based on the `URL` property.

I tried to reduce the changes within the core classes as much as possible ... but unfortunately we still have to add some clutter ...

Resolves: #2699


AsyncDisplayKit/ASImageNode.mm
```diff
@@ -200,6 +200,11 @@ - (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
 
 - (void)setImage:(UIImage *)image
 {
+  [self __setImage:image];
+}
+
+- (void)__setImage:(UIImage *)image
```
garrettmoon: Can we add a comment: "This is exposed to library subclasses, i.e. ASNetworkImageNode, ASMultiplexImageNode and ASVideoNode"

AsyncDisplayKit/ASMultiplexImageNode.mm
```diff
@@ -325,6 +326,12 @@ - (void)didExitVisibleState
 
 #pragma mark - Core
 
+- (void)setImage:(UIImage *)image
+{
+  ASDisplayNodeAssert(NO, @"Setting the image directly to an ASMultiplexImageNode is not allowed.");
```
garrettmoon: Should the assert explain why?

"Setting the image directly on an ASMultiplexImageNode is unsafe. It will be cleared in didExitPreloadRange and will have no way to restore in didEnterPreloadRange"

Adlai-Holler:
As we discussed, let's remove the assert and do one of the following:

- Implement `setImage:` by calling `setDefaultImage:` and implement `-image` by returning `super.image ?: self.defaultImage`. Possibly deprecate `.defaultImage` externally.
- Add a flag like `imageWasSetExternally` and check that flag before clearing the image when exiting the preload range.

I advocate for the first one, but I don't feel strongly about it. I think it's reasonable to say that if a URL _and_ an image are set, the node should prefer the remote image.

Adlai-Holler:
Actually on second thought, I advocate for the second option. If someone sets an image on the node externally, we should prefer that and perhaps not even both loading the image from URL.

This way the developer has finer control over how the node appears. If I set an image on it, that image will get shown full stop.

maicki:
Ok let's roll with option two for now.

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -116,6 +119,32 @@ - (void)dealloc
...
+    return;
+  }
+  __instanceLock__.unlock();
+  
+  [self __setImage:image];
+}
+
 - (void)setURL:(NSURL *)URL
 {
   [self setURL:URL resetToDefault:YES];
```
garrettmoon: Should setURL clear out _imageWasSetExternally?

AsyncDisplayKit/ASImageNode.mm
maicki: Added comment

AsyncDisplayKit/ASNetworkImageNode.mm
maicki: It will clear it out now yes

AsyncDisplayKit/Private/ASImageNode+Private.h
```diff
@@ -0,0 +1,27 @@
...
+
+@interface ASImageNode (Private)
+
+/*
+ * Set the image property of the ASImageNode. Subclasses like ASNetworkImageNode do not allow setting the
+ * image property directly and throw an assertion. There still needs to be a way for subclasses of
+ * ASNetworkImageNode to set the image.
+ *
+ * This is exposed to library subclasses, i.e. ASNetworkImageNode, ASMultiplexImageNode and ASVideoNode for setting
+ * the image directly without going throug the setter of the superclass
```
garrettmoon: total nit: through

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -506,10 +521,15 @@ - (AVAsset *)asset
...
 {
-  [self didExitPreloadState];
-  _asset = asset;
-  _assetURL = assetURL;
-  [self setNeedsPreload];
+  [self _clearVideoNode];
+  
+  {
+    ASDN::MutexLocker l(__instanceLock__);
+    _asset = asset;
```
garrettmoon: Ha, nice, fix some threading issues while you're here :)

maicki:
@garrettmoon @appleguy Ready for another round. As already mentioned not really beautiful and unfortunately adds some clutter to the core classes :(

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -116,6 +118,32 @@ - (void)dealloc
...
 #pragma mark - Public methods -- must lock
 
+/// Setter for public image property. It has the side effect to set an internal _imageWasSetExternally that prevents setting an image internally. Setting an image internally should happen with the _setImage: method
+- (void)setImage:(UIImage *)image
+{
+  __instanceLock__.lock();
+  
+#ifdef DEBUG
+  if (_URL != nil) {
+    NSLog(@"Setting the image directly on an %@ and setting and setting an URL is not supported. If you want to use a placeholder image please use defaultImage .", NSStringFromClass([self class]));
```
Adlai-Holler: Is this necessary? If not, I think we should support this condition. Think for example of using a network image node in a transition, and so you might want to override its image with some other one temporarily, to achieve a visual effect.

maicki:
@garrettmoon @Adlai-Holler @appleguy  Ready for review.

Issue #2709, opened 2016-12-05
ASViewController flashes while pushing 


Baevra:
You can reproduce this in ASDKgram example, just push another copy of controller

maicki:
Thanks for reporting @Baevra. This is a regression that was introduced with #2642 and #2689 and is in 2.0rc-1. cc @garrettmoon

PR #2710, opened 2016-12-05
[ASDisplayNode] Fix flickering for nodes that support range managed interface state

maicki:
This bug was initially introduced in #2130 as moved the check for `supportsRangeManagedInterfaceState` for `clearFetchedData` from `setInterfaceState` to `didExitPreloadState` within #2130

Let's get the old behavior back and move the `supportsRangeManagedInterfaceState` check back to `setInterfaceStateState` as it was before all of this renaming PRs. This also does allows us to not have to introduce some internal `_didExitPreloadState` method.

Resolves: #2709


garrettmoon:
Bummer! This is exactly the code that @nguyenhuy put in and I took out because I didn't understand it and seemed like additional behavior changes.

I wish there was a cleaner fix for this though, I'm worried about not calling didExitPreloadState at all. Specifically, I'm worried about people expecting this to be called and reporting bugs.

maicki:
@garrettmoon It's the same behavior we had we just moved in #2130 so there should be no regression

PR #2715, opened 2016-12-05
Ensure Batch Fetching Includes Leading Screens, Even if Not Scrolling

### LINKED ISSUE #2714: Batch Fetching May Not Retrigger Reliably if First Fetch Isn't Enough
We should keep batch fetching until our content extends sufficiently far beyond the scrollable area.

There are cases (no details yet) where the first fetch comes back with ~1 screenful of content, and we don't trigger another batch fetch. Very likely related to #2675 based on the timing of this showing up.

Adlai-Holler:
Resolves #2714.

- If they are visible, batch fetch until we have the leading screens, unless they're scrolling toward the head.
- If they are not visible, batch fetch until we have the visible area.

### DIFF
--- a/AsyncDisplayKit/Private/ASBatchFetching.m
+++ b/AsyncDisplayKit/Private/ASBatchFetching.m
@@ -61,10 +61,14 @@ BOOL ASDisplayShouldFetchBatchForContext(ASBatchContext *context,
     return YES;
   }
 
-  BOOL isScrollingTowardEnd = (ASScrollDirectionContainsDown(scrollDirection) || ASScrollDirectionContainsRight(scrollDirection));
+  // If they are scrolling toward the head of content, don't batch fetch.
+  BOOL isScrollingTowardHead = (ASScrollDirectionContainsUp(scrollDirection) || ASScrollDirectionContainsLeft(scrollDirection));
+  if (isScrollingTowardHead) {
+    return NO;
+  }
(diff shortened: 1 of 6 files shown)


AsyncDisplayKit/Private/ASBatchFetching.m
```diff
@@ -55,16 +57,25 @@ BOOL ASDisplayShouldFetchBatchForContext(ASBatchContext *context,
...
 
-  BOOL isScrollingTowardEnd = (ASScrollDirectionContainsDown(scrollDirection) || ASScrollDirectionContainsRight(scrollDirection));
+  // If we are not visible, but we do have enough content to fill visible area,
+  // don't batch fetch.
+  if (visible == NO) {
+    return NO;
+  }
+
+  // If they are scrolling toward the head of content, don't batch fetch.
+  BOOL isScrollingTowardHead = (ASScrollDirectionContainsUp(scrollDirection) || ASScrollDirectionContainsLeft(scrollDirection));
```
garrettmoon: My only concern is that we're only supporting batch fetching to the bottom and the right; has this always been the case?

Adlai-Holler: Yep it has always been the case. We'd like to support head fetching in the future, but it'll take a lot more work.

PR #2728, opened 2016-12-07
Add lint mode to build.sh

garrettmoon:


garrettmoon:
This should prevent us from running into lint errors when we go to publish in cocoapods

PR #2729, opened 2016-12-07
[Layout Transition] Reset the alpha value for inserted and removed subnode to initial value

maicki:
After removing a subnode via the animated layout transition we should set the alpha value back to the old value.

The reasoning behind is that if the subnode would be added back via a next non animated layout transition (e.g. instead of calling `transitionLayout...`, call `setNeedsLayout` it would still have an alpha value of 0 and although measured the right size it not show up. Furthermore we should also do the same for inserted nodes.


maicki:
I had a quick chat with @garrettmoon and what we actually should do is in the case of fading in as well as fading out cache the previous alpha value and restore it to that way.

maicki:
@garrettmoon @Adlai-Holler @appleguy  Ready for next review

appleguy:
test this please

appleguy:
test this please

PR #2731, opened 2016-12-08
Improve Handling of Rasterized Node Interface & Hierarchy States

### LINKED ISSUE #2205: [ASDisplayNode] Rasterized Nodes should get Interface State Updates
I just clone AsyncDisplayKit (master commit:2bfeb6d). 
Run the example ASDKgram. The simple pan to scroll get a crash every time.  

```
2016-09-07 13:58:34.293 Sample[544:100609] *** Assertion failure in -[ASTextNode dealloc], /Users/xinli/faceu/temp/AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:393
2016-09-07 13:58:34.299 Sample[544:100609] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Node should always be marked invisible before deallocating; interfaceState: 15, <ASTextNode: 0x12f9e2200; text = "View all 59 comments"; truncation string = "…"; frame = (null); renderer = 0x12f68f400>'
*** First throw call stack:
(0x182016db0 0x18167bf80 0x182016c80 0x18299c154 0x1000d7a54 0x1001930e8 0x1000bc86c 0x182007f2c 0x1000428cc 0x100042470 0x1000514e0 0x1000512b0 0x100047d7c 0x100048b74 0x10045da7c 0x10045da3c 0x1004634e4 0x181fccd50 0x181fcabb8 0x181ef4c50 0x1837dc088 0x1871de088 0x10003f908 0x181a928b8)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

screen shot 2016-09-07 at 2 04 40 pm

Adlai-Holler:
This resolves #2205 and restores the rasterization used in ASDKgram.

All of the test cases added in this diff were failing before the diff.

`shouldRasterizeDescendants` has been moved to a Beta header while we continue to develop it. If in the future we decide that its performance benefits don't justify its maintenance cost, we may deprecate it and later remove it.

In addition to fixing those behaviors, this diff
- Removes `-deallocSafeSupernode` since it's literally the exact same code as `-supernode` and updates the documentation for the latter to specify that it needs to be dealloc-safe.
- Removes cases where we were calling `ASPerformBlockOnMainThread` in cases where, in the same method, we had already asserted that we were on the main thread.
- Adds a handy little macro `ASSetDebugNames(...)` that captures the current class name so that if you're in `MYButtonNode` and you call `ASSetDebugNames(self.titleNode, _subtitleNode)` you'll get the names set to `MYButtonNode.titleNode` and `MYButtonNode.subtitleNode`. The macro has no effect in production.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Beta.h
+++ b/AsyncDisplayKit/ASDisplayNode+Beta.h
@@ -123,6 +123,28 @@ typedef struct {
  */
 + (void)setRangeModeForMemoryWarnings:(ASLayoutRangeMode)rangeMode;
 
+/**
+ * @abstract Whether to draw all descendant nodes' layers/views into this node's layer/view's backing store.
+ *
+ * @discussion
+ * When set to YES, causes all descendant nodes' layers/views to be drawn directly into this node's layer/view's backing
+ * store.  Defaults to NO.
+ *
+ * If a node's descendants are static (never animated or never change attributes after creation) then that node is a
+ * good candidate for rasterization.  Rasterizing descendants has two main benefits:
+ * 1) Backing stores for descendant layers are not created.  Instead the layers are drawn directly into the rasterized
+ * container.  This can save a great deal of memory.
+ * 2) Since the entire subtree is drawn into one backing store, compositing and blending are eliminated in that subtree
+ * which can help improve animation/scrolling/etc performance.
+ *
+ * Rasterization does not currently support descendants with transform, sublayerTransform, or alpha. Those properties
+ * will be ignored when rasterizing descendants.
+ *
+ * Note: this has nothing to do with -[CALayer shouldRasterize], which doesn't work with ASDisplayNode's asynchronous
+ * rendering model.
+ */
+@property (nonatomic, assign) BOOL shouldRasterizeDescendants;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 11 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2471,7 +2461,7 @@ - (void)__setSupernode:(ASDisplayNode *)newSupernode
             
             // Propagate down the new pending transition id
             ASDisplayNodePerformBlockOnEverySubnode(self, NO, ^(ASDisplayNode * _Nonnull node) {
-              node.pendingTransitionID = _pendingTransitionID;
+              node.pendingTransitionID = pendingTransitionId;
```
Adlai-Holler: This was just a minor performance get, by using the local variable rather than the instance variable.

AsyncDisplayKit/ASDisplayNode+Beta.h
```diff
@@ -123,6 +123,28 @@ typedef struct {
  */
 + (void)setRangeModeForMemoryWarnings:(ASLayoutRangeMode)rangeMode;
 
+/**
```
appleguy: Maybe add a note of some kind here - "Rasterization is a powerful optimization feature, but can sometimes have unpredictable effects.  We have this option in +Beta to reflect the fact that some rarer edge cases may cause issues.  Be sure to test any configurations you use Rasterization with."

Issue #1949, opened 2016-07-19
ASViewController/ASPagerNode layout bug since version 1.9.81


flovouin:
Hi guys,

Since upgrading to 1.9.81 I've got a weird bug that appears in only one of my view controllers: when it's pushed on a `UINavigationController`, it is first correctly laid out, but then when the push animation finishes, some nodes inside are slightly shifted up, leaving a blank at the bottom of the screen.
A few things to note:
- The affected nodes are inside a horizontal `ASPagerNode`. Other nodes in the hierarchy doesn't seem to be affected.
- The pager node itself fills the whole screen, so it seems correctly laid out (this is what I want). Only its children are incorrect.

This "bug" seems to be related to #1860: it is observed in 82233af but not in 33620b9. It's probably a combinaison of a change in ASDK and something I'm doing wrong. I'm trying to put up a small demo project to reproduce this, but it's not finished yet. In the mean time, do you have any ideas what could be causing this behaviour, given the changes in #1860? Anything I could try?

Cheers,
Flo

flovouin:
Unfortunately I'm not able to reproduce the bug in a demo project, however I have a bit more details:
<img width="665" alt="screen shot 2016-07-20 at 10 53 03 am" src="https://cloud.githubusercontent.com/assets/5155149/16982486/5de42740-4e70-11e6-8f54-40a4ecea0a36.png">
This is the view hierarchy with clipped content shown. Filled with blue is the `ASPagerNode`. The half-transparent rectangle is the `_ASCollectionViewCell` (for which I probably don't have access?). Finally the transparent rectangle with blue borders is my custom `ASCellNode`.
My `ASCellNode` has the correct size that I set using `constrainedSizeForNodeAtIndexPath`, however its frame doesn't have the correct origin, which is aligned with the `_ASCollectionViewCell`. Any idea why the `_ASCollectionViewCell` would be higher (and taller) than the `ASPagerNode` itself?

Thanks!

koheku:
I have a similar bug : an `ASPagerNode` whose children are laid out differently since I updated to `1.9.81`. It looks like there are top/bottom insets surrounding its children. Will report back if I find anything !
![simulator screen shot 20 jul 2016 12 10 42](https://cloud.githubusercontent.com/assets/1695036/16984702/9731c0cc-4e73-11e6-92c2-efc17fed39a2.png)
[...]

PR #2736, opened 2016-12-09
[ASPagerNode] Fix content insets are wrong in pager node if root node of ASViewController and transition back

maicki:
This PR will fix the problem that if pushed back to a full screen pager node the UICollectionViewCell that contains the ASCellNode will have an y value of -64.

It does not fix the ASPagerFlowLayout message that comes up. To fix that we will add some more documentation that view controller should use `self.automaticallyAdjustScrollViewInsets = NO` if the pager node is used full screen.

Addresses: #1949

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -1232,10 +1232,6 @@ static NSString * const kCellReuseIdentifier = @"_ASCollectionViewCell";
 
 - (void)layoutSubviews
 {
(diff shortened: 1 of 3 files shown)


maicki:
test this please

maicki:
@garrettmoon @appleguy @Adlai-Holler Ready for review

garrettmoon:
test this please

appleguy:
I remember writing this line, and this is definitely a big improvement!  To do it after [super layoutSubviews].

PR #2743, opened 2016-12-09
[Examples] Some smaller SocialAppLayout example improvements

### LINKED ISSUE #2741: Social App Layout: flexGrow does not propagate upwards as advertised
In the Social App Layout example app, it says in the layoutSpecThatFits section:

    // NOTE: This inset is not actually required by the layout, but is an example of the upward propogation of layoutable
    // properties.  Specifically, .flexGrow from the child is transferred to the inset spec so they can expand together.
    // Without this capability, it would be required to set insetSpacer.flexGrow = 1.0;

However, this is not the case.  When running the app, the space between the `usernameNode` and the `viaNode/timeNode` does not expand.

To achieve the expected behaviour, one must add `insetSpacer.style.flexGrow = 1.0`, which means that the spacer's flexGrow property is not correctly propagated to its containing layoutSpec.

maicki:
Resolves #2741

### DIFF
--- a/examples/SocialAppLayout/Sample/PostNode.m
+++ b/examples/SocialAppLayout/Sample/PostNode.m
@@ -212,16 +212,8 @@
     ASLayoutSpec *spacer = [[ASLayoutSpec alloc] init];
     spacer.style.flexGrow = 1.0;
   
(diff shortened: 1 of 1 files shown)


garrettmoon:
test this please

PR #2746, opened 2016-12-10
[ASDisplayNode] Remove node from supernode before adding it as a subview

rcancro:
If a node is being added as a subview to a UIVIew, we must make sure to remove it as a subnode of its supernode.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -3857,6 +3857,9 @@ static const char *ASDisplayNodeAssociatedNodeKey = "ASAssociatedNode";
     if (selfNode) {
       [selfNode addSubnode:subnode];
     } else {
+      if (subnode.supernode) {
+        [subnode removeFromSupernode];
+      }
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -3857,6 +3857,9 @@ - (void)addSubnode:(ASDisplayNode *)subnode
     if (selfNode) {
       [selfNode addSubnode:subnode];
     } else {
+      if (subnode.supernode) {
```
appleguy: @rcancro thanks for submitting this!  I'm curious how you found it.

Could you also fix the CALayer -addSubnode: category method that is just beneath this?

garrettmoon:
test this please

rcancro:
I found it when working on a custom transition. I noticed that if I added an ASDisplayNode to the transition context's `containerView` that the ASDisplaynode did not appear unless I explicitly called `removeFromSupernode`

appleguy:
@rcancro phenomenal fix, thank you!!

PR #2751, opened 2016-12-11
[ASControlNode] Add ASControlNodeEventValueChanged to support creating switches, sliders, etc.

appleguy:
There is a use for this in Pinterest -- however, I've also seen external requests for it, and it's quite generally useful.

This does increase the number of iterations we check for the mask, but this should be instantaneous at runtime.  It scans across some currently unused values, but this could also allow third parties to add their own if they want to be especially crafty and life on the edge :).

### DIFF
--- a/AsyncDisplayKit/ASControlNode.h
+++ b/AsyncDisplayKit/ASControlNode.h
@@ -34,6 +34,8 @@ typedef NS_OPTIONS(NSUInteger, ASControlNodeEvent)
   ASControlNodeEventTouchUpOutside    = 1 << 5,
   /** A system event canceling the current touches for the control node. */
   ASControlNodeEventTouchCancel       = 1 << 6,
+  /** A system event triggered when controls like switches, slides, etc change state. */
+  ASControlNodeEventValueChanged      = 1 << 12,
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASControlNode.h
```diff
@@ -34,6 +34,8 @@ typedef NS_OPTIONS(NSUInteger, ASControlNodeEvent)
   ASControlNodeEventTouchUpOutside    = 1 << 5,
   /** A system event canceling the current touches for the control node. */
   ASControlNodeEventTouchCancel       = 1 << 6,
+  /** A system event triggered when controls like switches, slides, etc change state. */
+  ASControlNodeEventValueChanged      = 1 << 12,
```
appleguy: This number, << 12, matches what Apple uses for UIControl (as do the others).

appleguy:
test this please

hannahmbanana:
@appleguy - could you merge this? I would like to get this into the sync candidate branch today.

hannahmbanana:
Thank you for adding this for me @appleguy!

Issue #780, opened 2015-10-27
[ASScrollNode] Enable layout specs to calculate & set contentSize separately from bounds.


levi:
Seems like a match made in heaven, however, the constrained size range used by `measureWithSizeRange` is constrained to the size of the view bounds from the view controller. I have been thinking about splitting the measurement and layout pass of ASScrollNode into two parts:
1. `layoutSpecThatFits:` measures the frame size of the scroll node, constraining the scroll view's content size to a specified view port (it's currently already doing this).
2. `layoutContentSpecThatFits:` the exact same behavior of `layoutSpecThatFits:`, but is responsible for returning a layout spec that measures the `contentSize` of the backing scroll view. In order for this to function correctly, a method `calculateContentLayoutThatFits:`, which is called before `calculateLayoutThatFits:`, performs the exact same behavior, but is an override point for subclasses to modify the passed `ASSizeRange`. The most common instance of a scroll view is fixed width, but vertically scrolling content. Overriding `calculateContentLayoutThatFits:` and modifying the size range max to `CGSize(width: constrainedSize.max.width, height: CGFloat.max)` would then measure the contentSize to achieve this common scrolling behavior.

The results of `calculateContentLayoutThatFits:` set the `contentSize` on the layout pass and also set the frames of the scroll view's children. 

I wanted to bring this up before putting up a diff because it adds an extra API step that is not common amongst other ASDisplayNodes. Moreover, it will require `_flags.isMeasured` to be exposed somewhere to achieve the same behavior as measure/layout on `ASDisplayNode`. Open to feedback and suggestions. If it's not initially clear what I'm talking about, I'll put up a quick diff to better explain my intentions.

appleguy:
To me this seems like a good idea, although a bit dicy haha :).

This would only work for scrollers where you can position everything they'll contain up-front, and thus size them together too.

Are you able to share any specifics on your use case?  That would help us assess the very best API direction (e.g. @nguyenhuy & @eanagel & @rcancro)

levi:
Yes, dicy is a good description :smile:. Open to more flexible solutions that allow for easy integration of ASScrollNode with ASViewController. I'm going to dig into the ASViewController implementation to see if there is a way in which we can order the layout logic for subclasses to control the measure pass more specifically without duplicating a lot of the internal logic.

appleguy:
[...]

PR #2753, opened 2016-12-11
[ASScrollNode] Support for automaticallyManagesContentSize, adopting the ASLayoutSpec's size as the scrollable contentSize.

appleguy:
[ASScrollNode] Support for automaticallyManagesContentSize, adopting the ASLayoutSpec's size as the scrollable contentSize.

This feature has been desired for a long time, and has turned out to be phenomenally useful and easy to use.

It works well either for a blank ASScrollNode with .layoutSpecBlock set on it, or a subclass of ASScrollNode with a more traditional layoutSpecThatFits: implementation.  With this approach there is no need to capture the layout size, use an Absolute layout spec as a wrapper, or set contentSize anywhere in the code and it will update as the layout changes!

There is no automatic management of contentInset, but it would make sense to add this with keyboard listeners in the future.

I'm including in 2.0.1 for now because this is already working, a small addition, and should be easy to land.  It's a pretty material improvement for some key use cases with Stacks, and will help us answer questions related to scroll nodes with confidence.


appleguy:
This will solve https://github.com/facebook/AsyncDisplayKit/issues/780 .

@maicki very excited to get you feedback here on Monday!

appleguy:
test this please

appleguy:
FYI -- because this was needed for one of Hannah's diffs in Pinterest, I've pulled it into p6.17 -- although that branch has not yet been synced into Pinterest.  So if we make changes to this, let's also update p6.17.

AsyncDisplayKit/ASScrollNode.mm
```diff
@@ -0,0 +1,69 @@
...
+
+- (instancetype)init
+{
+  return [super initWithViewBlock:^UIView *{ return [[ASScrollView alloc] init]; }];
+}
+
+- (void)layout
+{
+  [super layout];
+  ASDN::MutexLocker l(__instanceLock__);
```
maicki: Are we sure this will not add any deadlock situation?

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -198,8 +198,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 #define __ASDisplayNodeCheckForLayoutMethodOverrides \
     ASDisplayNodeAssert(_layoutSpecBlock != NULL || \
     ((ASDisplayNodeSubclassOverridesSelector(self.class, @selector(calculateSizeThatFits:)) ? 1 : 0) \
-    + (ASDisplayNodeSubclassOverridesSelector(self.class, @selector(layoutSpecThatFits:)) ? 1 : 0) \
-    + (ASDisplayNodeSubclassOverridesSelector(self.class, @selector(calculateLayoutThatFits:)) ? 1 : 0)) <= 1, \
+    + (ASDisplayNodeSubclassOverridesSelector(self.class, @selector(layoutSpecThatFits:)) ? 1 : 0)) <= 1, \
```
maicki: `ASDisplayNode` should ideally not overwrite `calculateLayoutThatFits:` we should add some other mechanism via maybe a whitelist that allows internal classes to overwrite the method and other subclasses not.

AsyncDisplayKit/ASScrollNode.mm
```diff
@@ -0,0 +1,69 @@
...
+    CGSize contentSize = _contentCalculatedSizeFromLayout;
+    if (ASIsCGSizeValidForLayout(contentSize) == NO) {
+      NSLog(@"%@ calculated a size in its layout spec that can't be applied to .contentSize: %@. Applying CGSizeZero instead.", self, NSStringFromCGSize(contentSize));
+      contentSize = CGSizeZero;
+    }
+    self.view.contentSize = _contentCalculatedSizeFromLayout;
+  }
+}
+
+- (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
```
maicki: Can you move `calculateLayoutThatFits:` above `layout` please

AsyncDisplayKit/ASScrollNode.mm
```diff
@@ -0,0 +1,69 @@
...
+{
+  ASLayout *layout = [super calculateLayoutThatFits:constrainedSize];
+  ASDN::MutexLocker l(__instanceLock__);
```
maicki: Can this introduce some deadlock situation?

AsyncDisplayKit/ASScrollNode.mm
```diff
@@ -0,0 +1,69 @@
...
+  if (_automaticallyManagesContentSize) {
+    _contentCalculatedSizeFromLayout = layout.size;
+    if (ASIsCGSizeValidForLayout(constrainedSize.max)) {
+      layout = [ASLayout layoutWithLayoutElement:self size:constrainedSize.max sublayouts:layout.sublayouts];
```
maicki: You have to set the position to `CGSizeZero`

AsyncDisplayKit/ASScrollNode.mm
maicki: What about looking into overwriting `- (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize restrictedToSize:(ASLayoutElementSize)size relativeToParentSize:(CGSize)parentSize;` Here you get the parent size passed in and use that instead of `constrainedSize.max`. You also don't have to remove the subclass overwrite above.

maicki:
@appleguy Could we also get some unit test for that?

AsyncDisplayKit/ASScrollNode.mm
appleguy: @maicki don't we lock in super layout?  if so, then it should be safe?

AsyncDisplayKit/ASScrollNode.mm
appleguy: Good suggestion, was also thinking about that but seemed like a private method -- I'll add it to frameworkprivate

AsyncDisplayKit/ASScrollNode.mm
appleguy: @maicki same, doesn't the super method grab the same lock?

PR #2754, opened 2016-12-11
[ASDisplayNode] Ensure all subclasses are using base class __instanceLock__ and not re-defining their own.

appleguy:
This also moves the @package definition of the instance variable to +FrameworkPrivate instead of Internal.h, because Internal.h should ideally not be used outside of the ASDisplayNode file setup.

This has greatly reduced the number of imports of Internal.h.


appleguy:
Unfortunately this has created a ton of issues where the Tests files need to become .mm to import +FrameworkPrivate now.

@Adlai-Holler, @maicki how do you feel about this?  I tried renaming a lot of them, and it isn't enough -- it needs to be basically all of them.  I think we should either standardize on this or we need to have a more careful way of avoiding this.

I tried using #ifdef __cplusplus to gate the __instanceLock__ ivar definition, but that didn't seem to work, for unknown reasons.

If anyone has time to investigate this, it would be appreciated.  Unless we can jointly come up with a strategy for how to resolve this, I'm not sure what the best approach is...

maicki:
@appleguy I don't know why we should move the instance lock? We could create it's own header for lock related ivars and only import it if we need it. What do you think about that? Maybe it will also cleanup some ...

It makes sense to have it in an internal header and import it when we need it in subclasses. I agree we should not define different ivars for subclasses of `ASDisplayNode` though.

garrettmoon:
test this please

PR #2761, opened 2016-12-13
Fixed Carthage build error

hartbit:
Carthage fails to build because it finds a framework shared scheme in `examples_extra/ASLayoutSpecPlayground-Swift/Sample.xcodeproj`. I un-shared it to hide it from Carthage.


Adlai-Holler:
Ah nice catch @hartbit 

Jenkins test this please.

appleguy:
Thanks a lot @hartbit, very glad to have your help!

PR #2764, opened 2016-12-13
[ASThread][ASDisplayNode] Detect and avoid deadlocks caused by upward lock gathering in didEnter/Exit states

### LINKED ISSUE #2605: [ASDisplayNode] Lock entanglement can occur due to -didEnterVisibleState activity.
This looks likely to be caused by the new locking around the ASLayoutElementStyle property.  It's important to note that the previous layout properties did not have locks, as entanglement issues similar to this were observed.

We need to investigate this for 2.0 to assess how common it is and whether it can be worked around.  Moving to more atomic scalar values for the layout configuration would help, and we may be able to find an alternate synchronization option for style — e.g. perhaps disallow writing the property itself externally for 2.0, since this use case is currently rare.

https://fabric.io/pinterest6/ios/apps/pinterest/issues/57dbb2df0aeb16625b51df2c/sessions/ecd258e4b7d74712884ee3d6fa794631_9c4482a0a27d4b29972e29526f069c01_0_v1?

@maicki would know this area well - however, he is sprinting on some awesome changes in the layout engine currently.  @nguyenhuy, do you have cycles to investigate this in the next week?  A first look, even without a fix, could be a valuable next step.

nguyenhuy:
This PR adds a mechanism that detects if a thread tries to access supernode of a node while holding its recursive lock (e.g early sign of upward lock gathering). To detect that, I eventually had to record mutex's owner and level of the ownership. The new code is turned off by default to avoid rather significant overheads.

This PR also changes the locking strategy in ASDisplayNode to avoid situations in which we hold a node's instance lock and call methods outside the scope of `self` (for example, [subnode __setSuperNode:self]`) or didEnter/Exit state. These methods can trigger methods that are meant to be overridden by developers in which they walk up the node tree and grab other locks and potentially cause deadlocks (like #2605). Moreover, they perform expensive tasks that cause the lock to be held for too long.

The code paths changed by the PR are well covered by our test suite. I also tested this patch using a small sample project. Moreover, I'm gonna test it with Pinterest master.

I believe this PR should fix #2605 and other deadlocks with similar traces.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -135,28 +135,30 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 
 - (void)setURL:(NSURL *)URL resetToDefault:(BOOL)reset
 {
-  ASDN::MutexLocker l(__instanceLock__);
-
-  _imageWasSetExternally = NO;
-
-  if (ASObjectIsEqual(URL, _URL)) {
-    return;
-  }
-
-  [self _cancelImageDownload];
-  _imageLoaded = NO;
-
-  _URL = URL;
-
-  BOOL hasURL = _URL == nil;
-  if (reset || hasURL) {
-    [self _setImage:_defaultImage];
-    /* We want to maintain the order that currentImageQuality is set regardless of the calling thread,
-     so always use a dispatch_async to ensure that we queue the operations in the correct order.
-     (see comment in displayDidFinish) */
-    dispatch_async(dispatch_get_main_queue(), ^{
-      self.currentImageQuality = hasURL ? 0.0 : 1.0;
-    });
+  {
+    ASDN::MutexLocker l(__instanceLock__);
+    
+    _imageWasSetExternally = NO;
+    
+    if (ASObjectIsEqual(URL, _URL)) {
+      return;
+    }
+    
+    [self _cancelImageDownload];
+    _imageLoaded = NO;
+    
+    _URL = URL;
+    
+    BOOL hasURL = _URL == nil;
+    if (reset || hasURL) {
+      [self _setImage:_defaultImage];
+      /* We want to maintain the order that currentImageQuality is set regardless of the calling thread,
+       so always use a dispatch_async to ensure that we queue the operations in the correct order.
+       (see comment in displayDidFinish) */
+      dispatch_async(dispatch_get_main_queue(), ^{
+        self.currentImageQuality = hasURL ? 0.0 : 1.0;
+      });
+    }
   }
 
   [self setNeedsPreload];
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -487,7 +487,7 @@ static NSString * const kRate = @"rate";
 
 - (void)setAsset:(AVAsset *)asset
 {
-  if (ASAssetIsEqual(asset, _asset) == NO) {
+  if (ASAssetIsEqual(asset, self.asset) == NO) {
     [self setAndFetchAsset:asset url:nil];
   }
 }
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -627,8 +627,9 @@ static NSString * const kRate = @"rate";
   if (_playerNode == nil) {
     _playerNode = [self constructPlayerNode];
 
-    [self addSubnode:_playerNode];
-
+    __instanceLock__.unlock();
+      [self addSubnode:_playerNode];
+    __instanceLock__.lock();
       
     [self setNeedsLayout];
   }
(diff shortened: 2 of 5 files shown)


AsyncDisplayKit/Details/ASThread.h
```diff
@@ -29,6 +29,7 @@ static inline BOOL ASDisplayNodeThreadIsMain()
 #ifdef __cplusplus
 
 #define TIME_LOCKER 0
+#define CHECK_LOCKING_SAFETY DEBUG
```
nguyenhuy: This might be a bit too aggressive because lots of warnings will be printed due to our current way of building logging context (PIContextLogger.m). Let me know if you think we should disable this by default.

nguyenhuy:
This PR is ready for the first round of reviews. Very interested in your feedbacks, @maicki @appleguy @Adlai-Holler @garrettmoon @levi @hannahmbanana. @hannahmbanana Really appreciate if you could help me to test this with Pinterest master. Thank you all!

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -135,28 +135,30 @@ - (void)setURL:(NSURL *)URL
...
-  BOOL hasURL = _URL == nil;
-  if (reset || hasURL) {
-    [self _setImage:_defaultImage];
-    /* We want to maintain the order that currentImageQuality is set regardless of the calling thread,
-     so always use a dispatch_async to ensure that we queue the operations in the correct order.
-     (see comment in displayDidFinish) */
-    dispatch_async(dispatch_get_main_queue(), ^{
-      self.currentImageQuality = hasURL ? 0.0 : 1.0;
-    });
+  {
```
garrettmoon: Just so I understand, is this so that the lock is released when ASDisplayNode calls recursively preload?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -467,10 +467,8 @@ - (void)setPlayerState:(ASVideoNodePlayerState)playerState
 
 - (void)setAssetURL:(NSURL *)assetURL
 {
-  ASDN::MutexLocker l(__instanceLock__);
-
   if (ASObjectIsEqual(assetURL, self.assetURL) == NO) {
```
garrettmoon: This change allows there to be a small possibility that setAndFetchAsset will be called when assetURL and self.assetURL are equal (because the lock is released after the check). Is that safe?

AsyncDisplayKit/ASVideoNode.mm
```diff
@@ -489,10 +487,8 @@ - (NSURL *)assetURL
 
 - (void)setAsset:(AVAsset *)asset
 {
-  ASDN::MutexLocker l(__instanceLock__);
-  
   if (ASAssetIsEqual(asset, _asset) == NO) {
```
garrettmoon: should this be self.asset instead of _asset? Or do we need to lock around the check?

AsyncDisplayKit/Details/ASThread.h
```diff
@@ -29,6 +29,7 @@ static inline BOOL ASDisplayNodeThreadIsMain()
 #ifdef __cplusplus
 
 #define TIME_LOCKER 0
+#define CHECK_LOCKING_SAFETY 0
```
garrettmoon: This is neat! It would be really cool if you could write up a comment explaining how to use it and how it can be used to find threading issues.

AsyncDisplayKit/ASVideoNode.mm
nguyenhuy: Fixed

AsyncDisplayKit/ASVideoNode.mm
nguyenhuy: Yes, there is a small chance that it happens.

I've just discovered that these properties are always set on main. I'm asking people to see if the same restriction applies to their getters. If it's also the case, we can replace these lockings with main thread assertions!

nguyenhuy:
@garrettmoon Thank you for your review. I addressed all comments, except one.

appleguy:
@nguyenhuy very impressive approach, thank you for working on this!  Quite valuable as an investment in the long term future of the locking strategy.

I'm pretty happy to see where locking is these days with __instanceLock__ having almost eliminated deadlocking cases, and this will help bring us very close to zero (besides cases where the CPU is truly tied up for a long time, like the forced recursive display when a large display queue exists).

I'll try to review this later in the week, but please don't wait for me to land it (especially if either Michael or Adlai can review, then Garrett + either one of their accepts should have authority to land this -- in keeping with our diff review guidelines, this kind of change should require two core team accepts).

nguyenhuy:
I tested this diff with Pinterest master over the weekend and noticed that video nodes flicker frequently while I'm scrolling. Investigating now and putting this PR back to [WIP].

AsyncDisplayKit/Details/ASThread.h
```diff
@@ -174,21 +185,51 @@ namespace ASDN {
...
       ASDISPLAYNODE_THREAD_ASSERT_ON_ERROR(pthread_mutex_lock (this->mutex()));
+#if CHECK_LOCKING_SAFETY
+      mach_port_t thread_id = pthread_mach_thread_np(pthread_self());
+      if (thread_id != _owner) {
+        assert(0 == _owner);
+        assert(0 == _count);
+        _owner = thread_id;
+      } else {
+        assert(_count > 0);
+      }
```
Adlai-Holler: (Follow-up task:) It would be great to have some comments to annotate this.

PR #2765, opened 2016-12-13
Remove Vestigial Code Supporting Removed Methods

Adlai-Holler:
When we removed some long-deprecated image downloader & cache methods, we didn't do a very thorough job of cleaning up. For example, now that there is only one primary image downloader method, we can mark it required and remove the silly assertion message `"You must implement either downloadImageWithURL:callbackQueue:downloadProgress:completion:."`

This ungunkifies our code, and lets the compiler enforce protocol requirements for us.


Adlai-Holler:
@garrettmoon Thoughts?

garrettmoon:
Thanks for cleaning up after me :3

PR #2771, opened 2016-12-14
[ASDisplayNode] Always remove associated objects on main

### LINKED ISSUE #2767: UIAXDelegateClearer Deallocating Off-Main Causes Crash in Production
This is necessary, because in addition to ivars, associated objects may also reference UIKit or other non-threadsafe object types.

E.g.:

https://fabric.io/pinterest6/ios/apps/pinterest/issues/5761fc0bffcdc042505bed8d/sessions/a7874f480ee94e9280789970aa097999_69c983cec93d43df944ccea216356c5e_0_v1

pasted image at 2016_12_13 06_26 pm

nguyenhuy:
This should fix #2767.

### DIFF
--- a/AsyncDisplayKit/ASRunLoopQueue.mm
+++ b/AsyncDisplayKit/ASRunLoopQueue.mm
@@ -12,11 +12,13 @@
 
 #import "ASRunLoopQueue.h"
 #import "ASThread.h"
+#import "ASInternalHelpers.h"
 #import "ASLog.h"
 
 #import <cstdlib>
 #import <deque>
 #import <vector>
+#import <objc/runtime.h>
 
 #define ASRunLoopQueueLoggingEnabled 0
 
@@ -48,9 +50,22 @@ static void runLoopSourceCallback(void *info) {
 
 - (void)releaseObjectInBackground:(id)object
 {
-  _queueLock.lock();
-  _queue.push_back(object);
-  _queueLock.unlock();
+  __weak __typeof__(self) weakSelf = self;
+  ASPerformBlockOnMainThread(^{
+    // Some associated objects cannot be deallocated in background (#2767). Always remove them on main instead.
+    objc_removeAssociatedObjects(object);
+    
+    __typeof__(self) strongSelf = weakSelf;
+    if (strongSelf) {
+      // It's important to push this object to dealloc queue right after the above removal, in the same run loop.
+      // Otherwise the queue may be scheduled to consume it before the removal occurs.
+      strongSelf->_queueLock.lock();
+      strongSelf->_queue.push_back(object);
+      strongSelf->_queueLock.unlock();
+    } else {
+      // Dealloc queue is gone. The object is released now.
+    }
+  });
 }
 
 - (void)threadMain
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -48,9 +50,22 @@ + (instancetype)sharedDeallocationQueue
...
-  _queueLock.lock();
-  _queue.push_back(object);
-  _queueLock.unlock();
+  __weak __typeof__(self) weakSelf = self;
+  ASPerformBlockOnMainThread(^{
+    // Some associated objects cannot be deallocated in background (#2767). Always remove them on main instead.
+    objc_removeAssociatedObjects(object);
+    
+    __typeof__(self) strongSelf = weakSelf;
+    if (strongSelf) {
```
Adlai-Holler: Since this is such a hot code path, and since in practice the queue is going to be long-lived, let's avoid hitting the weak table here and just retain `self` to reduce contention.

Adlai-Holler:
@appleguy @nguyenhuy This will resolve the issue, but it feels kind of hacky.

Since the issue we're addressing only presents under iOS 8, should we have two separate code paths? This introduces pretty significant overhead into a highly-optimized method.

We have no guarantee that all nodes will go through this system – for example, in Pinterest some node classes have code in `-dealloc` that enqueues subnodes onto an application-level deallocation queue.

Let's remove that product code and roll with this. If other problems present in the future, we'll go from there.

nguyenhuy:
@Adlai-Holler I've just pushed a new commit that avoid strongify/weakify. While it is true that this fix introduces overhead, the biggest performance impact will be in case that `-releaseObjectInBackground:` is called in a background thread. This case is exactly why we need this fix in the first place, so I think it's a fair trade-off between safety and performance. In case `-releaseObjectInBackground:` is called on main, the overhead is not that bad as `ASDisplayNodeThreadIsMain()` is quite performant and is used everywhere anyway.

I'm not sure why this system doesn't catch subnodes that are enqueued onto an application-level queue, since `-releaseObjectInBackground:` is the only public API of ASDeallocQueue. I may be missing something obvious here. Could you please elaborate?

Last but not least, while it is true that the crash we're seeing in Pinterest app only occurs on iOS 8, I'm worrying that we haven't achieved correctness in the current implementation. It is entirely possible for developers to lend a node to UIKit classes and then UIKit associates some unsafe objects to the node. Making a node a UIScrollViewDelegate for example, although its side effect can only be seen in iOS 8. So still think a fix is necessary, now or later.

Adlai-Holler:
@nguyenhuy In fact, as I think about this, this unfortunately won't fix the majority of cases.

For instance, let's say you have a view controller with `vcNode` and one if its subnodes, `subnode`. Let's say `subnode` becomes the scroll view's delegate (which is likely). When the view controller is deallocated (on main) it will enqueue `vcNode` onto the deallocation queue, but `subnode` will not have its associated objects cleared.

Unfortunately this code path just isn't used for most nodes. It's used as the _jumping off point_ to trigger entire hierarchies to be deallocated.

nguyenhuy:
@Adlai-Holler Oh yes, that's true. And we can't do something like this in `-dealloc` because we can't retain `self`:
```
  ASPerformBlockOnMainThread(^{
    objc_removeAssociatedObjects(self);
  });
```

Looks like the only way to fix this is to have a blacklist of associated keys. Gonna close this PR. Thank you!

PR #2773, opened 2016-12-14
Safer checks in ASDisplayNode’s setFrame before assigning bounds and position

rmls:
Even with application checks (for Inf and NaN) before assigning a frame to a node, it is still possible to have the newPosition containing NaN values. Since there is still some logic being done in `setFrame:` where other values are considered to calculate the new position and new bounds, it is possible that the NaN is being generated here.

This PR is meant to prevent that from happening by checking the final values before assigning them.

### DIFF
--- a/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
+++ b/AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
@@ -255,12 +255,8 @@ if (shouldApply) { _layer.layerProperty = (layerValueExpr); } else { ASDisplayNo
       CGPoint newPosition = CGPointZero;
       ASBoundsAndPositionForFrame(rect, origin, anchorPoint, &newBounds, &newPosition);
 
-      if (ASIsCGRectValidForLayout(newBounds) == NO) {
-        ASDisplayNodeAssert(NO, @"-[ASDisplayNode setFrame:] - The new bounds are invalid and unsafe to be set.");
-        return;
-      }
-      if (ASIsCGPointValidForLayout(newPosition) == NO) {
-        ASDisplayNodeAssert(NO, @"-[ASDisplayNode setFrame:] - The new position is invalid and unsafe to set.");
+      if (ASIsCGRectValidForLayout(newBounds) == NO || ASIsCGPointValidForLayout(newPosition) == NO) {
+        ASDisplayNodeFailAssert(@"-[ASDisplayNode setFrame:] - The new frame (%@) is invalid and unsafe to be set.", NSStringFromCGRect(rect));
         return;
       }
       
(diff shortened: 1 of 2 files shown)


garrettmoon:
@maicki mind taking a look at this? Do we have any concerns that other machinery might rely on this and crashing is better than an inconsistent state?

maicki:
In my opinion we actually should crash in production, but definitely assert in development, so bugs are actually discovered in production and not just a visible glitch.

The reasoning behind that is that we currently see that in Pinterest: In development for some reason we don't get into the state that we crash due to a nan value in setting a frame, so the assertion / crash will not hit in development, but in production it crashes. With this PR it would not crash in production and "just" a visual glitch and we would almost never find out that there is actually something going on the product code.

That's my thoughts around it.

AsyncDisplayKit/Private/ASDisplayNode+UIViewBridge.mm
```diff
@@ -254,6 +254,15 @@ - (void)setFrame:(CGRect)rect
       CGPoint newPosition = CGPointZero;
       ASBoundsAndPositionForFrame(rect, origin, anchorPoint, &newBounds, &newPosition);
 
+      if (ASIsCGRectValidForLayout(newBounds) == NO) {
+        ASDisplayNodeAssert(NO, @"-[ASDisplayNode setFrame:] - The new bounds are invalid and unsafe to be set.");
```
appleguy: This is a great change, thank you @rmls!!  Could you make a small modification by including the values being attempted to set?

I think you could do this like: ... setFrame:%@ ...", NSStringFromCGRect(newBounds)

appleguy: Actually, do this: consolidate to 1 if (bounds valid == NO || position valid == NO), and use just one assertion.  Since the user is providing a value to setFrame:, we can use the log with StringFromCGRect(rect), rather than breaking it into bounds / position for the log itself.

appleguy:
Regarding crash vs assertion.  I think we need to develop a better way to log assertions in production, but I disagree they should be crashes by default.  IMO we should develop a new macro for this case and offer two things to app developers:

- A #define switch to turn these special types of logs into hard crashes in production (off by default)
- An override where they can log nonfatal, e.g. with Fabric, so that the logs are still visible and problems are discovered in production.  This could be registering a block, like +[ASDisplayNode setNonfatalHandlerBlock:] or something.

I think this is fairly important as a default, because the vast majority of these cases happen with newly-developed code and have a relatively minor problem when the frame is not set.  In most cases, teams (at any company) would prefer to get /some/ experiment data rather than have to hotfix the app to stop crashes or get initial data collection.

@maicki @Adlai-Holler @nguyenhuy @garrettmoon this is a great discussion for the team to have so that we can set a standard operating practice around error handling of this sort.

rmls:
@appleguy Thank you for the feedback. I think this is ready to be looked at again.

garrettmoon:
I think what Scott is suggesting probably has the best balance here. I agree there are definitely concerns around reducing visibility if we don't crash, but maybe we can have our cake and eat it too! I've opened an issue (https://github.com/facebook/AsyncDisplayKit/issues/2807) to address this so that we can hopefully move forward with this.

Adlai-Holler:

Issue #2756, opened 2016-12-12
Some images are lower resolution (recent regression)


appleguy:
In Pinterest, when in Closeup using latest ASDK master, some images are lower resolution.  It reproduces 100%.

@garrettmoon to investigate.

appleguy:
It is conceivable this is related to:

"Hey guys, I was on `4355f4d` and tried upgrading to `2.0`, but it looks like https://github.com/facebook/AsyncDisplayKit/commit/5b80a641afae4cee6dcf0e4cf5a829dffeb1e8d6 messes with `UIView drawViewHierarchyInRect` and turns my view screenshots to black. (edited)

[12:07]  
(As I am only 3 commits away from `2.0`, one being only comments, and one being an example project, `5b80a6` looks like the ideal suspect)
"

garrettmoon:
Welp! Turns out it wasn't ASDK after all!

garrettmoon:
Ok, turns out it was us *and* ASDK.

PR #2774, opened 2016-12-14
[ASNetworkImageNode] Don't behave like an image node if URL is set.

garrettmoon:
After much discussion I think this is the correct behavior. It seems
like it's much more likely to be the expected behavior but still enables
you to use the network image node like an regular image node.

Resolves https://github.com/facebook/AsyncDisplayKit/issues/2756

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -109,16 +109,12 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 
 #pragma mark - Public methods -- must lock
 
-/// Setter for public image property. It has the side effect to set an internal _imageWasSetExternally that prevents setting an image internally. Setting an image internally should happen with the _setImage: method
+/// Setter for public image property. It has the side effect of setting an internal _imageWasSetExternally that prevents setting an image internally. Setting an image internally should happen with the _setImage: method
 - (void)setImage:(UIImage *)image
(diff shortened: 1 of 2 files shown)


garrettmoon:
@maicki @appleguy @Adlai-Holler would appreciate your eyes on this :)

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -109,16 +109,12 @@ - (void)dealloc
...
 - (void)setImage:(UIImage *)image
 {
   ASDN::MutexLocker l(__instanceLock__);
   
-  _imageWasSetExternally = (image != nil);
-  if (_imageWasSetExternally) {
-    [self _cancelDownloadAndClearImage];
-    _URL = nil;
-  }
+  _imageWasSetExternally = (image != nil && _URL != nil);
```
maicki: I think this change would make the `ASNetworkImage` back to the old unpredictables behavior. The behavior of the network image node would be based on the order of setting the properties. The behavior would depend on setting the URL before or after setting the image. Without this change it behaved like either an image node or a network image node, now it's kind of a hybrid thing ...

appleguy: @maicki I don't think that's quite right, due to the code in setURL:  The problem case would seem to be when you do setImage:, and then later do setURL:, this code will do _imageWasSetExternally = YES; however, that's not the case because setURL: will clear the flag.

PR #2781, opened 2016-12-15
[ASNetworkImageNode] Had the conditional in ASNetworkImageNode wrong

garrettmoon:


appleguy:
Sweet, sorry I didn't catch this!  Was just testing it in Pinterest...

Issue #2513, opened 2016-10-31
[PR Proposal] ASVideoNode performance improvement


wokalski:
I have recently came across the slow `AVPlayer` issue that has been discussed here and on the slack channel. I created a player using `AVSampleBufferDisplayLayer`  which runs on a background queue and it works pretty good. I could submit a PR which changes `AVPlayer` in `ASVideoNode` to a substantially faster solution. Please note though that even though it's faster it is less predictable (runs on a background queue, built quickly) but stable (as in doesn't crash from what I've seen). 

Let me know what you think

appleguy:
@wokalski Hey there, thanks for investigating this and suggesting it!

We are very interested in higher performance implementations for ASVideoNode.  There are some hacky techniques to do this with AVPlayer, but not all of those techniques can be upstreamed, leaving the public version suboptimal for real apps.

What iOS version does your new implementation work on?  What are the less predictable elements — does framerate of playback vary, or any issues with audio support?

I would suggest publishing either a new mode of operation for ASVideoNode that we may enable to be the default later (perhaps even for the 2.0 release), or creating an ASVideoNodeBeta that we could cut over to later (we've done this before with ASRangeController being rewritten).  This would allow us to accept the changes into master as quickly as possible, and get other developers testing and moving it forward, without exposing all users to the initial risk.

wokalski:
thanks for the detailed reply. 

I built a *prototype app* (keep this in mind) for a client. It had a list of videos looped simultaneously **without audio**. 720p 3-4 played at the same time (playback activated when a cell enters the screen).  I don't know if it matches your AVPlayer experience but `play` and `pause` blocked main thread for long periods. 

I used `AVSampleBufferDisplayLayer` (iOS 8+). It uses a background queue to request next frames (so to speak). This states the solution and yields numerous doubts about *predictability*.

1. Does framerate vary?
Since everything happens on a background queue, the predictability of the solution depends on the resources available. For instance with 3-4 1080p videos played simultaneously on iPhone 6 the issue was noticeable (I don't have numbers). Dropping frames were also visible when I used the device as a hotspot and recorded the screen. However in general it worked perfectly and I believe the issues above are understandable and should be handled on application level. (don't play 4 videos simultaneously if you don't want dropped frames)

2. Audio support
I didn't use audio but it shouldn't be a problem. We specify the timing properties of a layer. Somebody suggested using `CMAudioClock` to time it correctly but I didn't need it (no audio)

[...]

PR #2783, opened 2016-12-15
[WIP] Async video player

wokalski:
This pull request contains async video player. It follows #2513. 

In short, after merging it, there should be `ASVideoNodeBeta` with satisfactory (more on that below) set of features.

- [x] Play/pause
- [x] Setting playback rate
- [ ] Delegate
- [ ] Seeking (tricky - a naive implementation causes a ~second long delay)
- [ ] `ASVideoNodeBeta` implementation (planning to copy-paste `ASVideoNode` for most part)
- [ ] Sound (tricky - timing)
- [ ] Thumbnail generation
- [ ] Videos from remote sources

I marked two of them _tricky_. 

- The issue with seeking is that in the naive implementation, I would rebuild the whole reader stack. It results in an awful (~1 second) delay. I think it's resolved internally both at [FB](https://github.com/iAdam1n/App-Headers/blob/0ffb7404b66881bcb18be8636fd121ad7c82f6b1/Facebook/v62.0/FBRCTVideoReader.h#L38) and [Instagram](https://github.com/snower0428/jb/blob/5ccb15ae99b9bffd15fc961ad27b9953b42b8eec/Dump/instagram.h/IGVideoReader.h#L27) so feel free to give me a hint 😉.
- Audio - similarly as above. I am not an `AVFoundation` expert so I'm looking for best ways to sync audio with videos.

---

Alternatively we might not implement any of the tricky ones because the main use case for this is 60 fps scrolling. You usually don't need audio and additional playback controls for that.

Please do let me know your thoughts and I'll proceed accordingly.


kharmabum:
@wokalski have any ideas on where to start for support for remote sources? I can try tackling it.

wokalski:
Good question. It depends on the kind of stream you want to support. There's a lot information online about it if you look for `AVSampleBufferDisplayLayer`. Feel free to take over this PR. I'm not going to have time to finish it, yet I think it'd be super useful for some people.

appleguy:
@wokalski wow, cool, thanks for thinking to contribute this!  We will probably have to close this out, but I think we could open an issue to hold onto a link to the PR so it can be revived when someone next wants to invest in it.

wokalski:
Yeah, I am not going to have time for this but from what I've seen there's some demand for this feature.

Issue #2777, opened 2016-12-15
[ASEditableTextNode] Maximum number of lines to display


leotumwattana:
## Request
I'm looking for a way to limit **ASEditableTextNode** so that it will size and display a maximum number of lines. 

## Behaviour
For example let's say `editableTextNode.maximumLinesToDisplay = 6`

This should then size the editableTextNode to a maximum of 6 lines. More specifically:

When there is only 3 lines worth of text, then it should size to three lines. However, when there are 8 lines worth of text, then it should size to six lines and become scrollable for users to see the 7th and 8th lines.

## Additional Notes:
Simply setting `editableTextNode.textView.textContainer.maximumNumberOfLines = 6` will not work in this case, as although this will limit the textView to render only up to six lines, it will also not allow users to continue typing and see beyond those six lines.

Calculating using `font.lineHeight` may not work either, as that will unlikely take into account paragraph settings, etc.

What I did with plain UIKit is to use a TextKit stack to calculate the height of the UITextView using `layoutManager.lineFragmentRect`

Thanks in advance!

EviluS:
@leotumwattana In one of our projects I had to implement a similar feature. The behavior was a single line of editable text node and when you start typing it will grow up to three lines. When it exceeds the max number of lines it will stop growing and start scrolling. 

The implementation is:
- set a preferred frame size for the editable text node (defining the max height)
- adopt the `ASEditableTextNodeDelegate`
- implement the `editableTextNodeDidUpdateText:` method
- create a variable in which you will save the calculated content size of the editable text node
- in each call of the `editableTextNodeDidUpdateText:` check whether the text has grown and relayout the text node.

Basic example:
- Lets say that we need maximum of 3 lines and assume that the point size of the font is 20px so our max frame is 60px height.
- If we have 1 line we will layout our node to fit.
- if we have more than 2 lines we will limit our node to the max height so that it can enable its scrolling.

Sorry in advance if my implementation is not explained clearly. There are probably better implementations but this one works for men. If you need any additional help you can contact me in the ASDK Slack I am `gpetrov` there.

leotumwattana:
@EviluS Thanks. I'm thinking of making it a little more seamless, since this really does sound like something more than I alone will use. As @appleguy mentioned in his talks, `ASKD` is now going beyond just providing performance, but also convenience for framework users.

Looking through ASEditableTextNode.mm, I'm thinking of extending `ASTextKitComponents` with something like:

```
[...]

PR #2786, opened 2016-12-16
[ASEditableTextNode] Maximum number of lines to display (#2777)

leotumwattana:
This adds the property `. maximumLinesToDisplay` to `ASEditableTextNode` which will limit the size calculation of the `editableTextNode` to the specified number of lines.

Calculation is done using TextKit's `lineFragment` to ensure height calculation is accurate even under mixed font conditions.

Addresses Issue: #2777

### DIFF
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -243,7 +243,6 @@
   CGSize textSize;
   
   if (_maximumLinesToDisplay > 0) {
-    // *** TODO: My version of calculation
     textSize = [displayedComponents sizeForConstrainedWidth:constrainedSize.width
                                         forMaxNumberOfLines: _maximumLinesToDisplay];
   } else {
--- a/AsyncDisplayKit/ASEditableTextNode.mm
+++ b/AsyncDisplayKit/ASEditableTextNode.mm
@@ -476,7 +476,10 @@
   [_textKitComponents.layoutManager invalidateLayoutForCharacterRange:NSMakeRange(0, [_textKitComponents.textStorage length]) actualCharacterRange:NULL];
 
   // When you type beyond UITextView's bounds it scrolls you down a line. We need to remain at the top.
-  [_textKitComponents.textView setContentOffset:CGPointZero animated:NO];
+  // [Question] Is there a good reason why we are forcing content offset to top? Why not
+  // keep at the current offset of the text view?
+  CGPoint offset = _textKitComponents.textView.contentOffset;
+  [_textKitComponents.textView setContentOffset:offset animated:NO];
 }
 
 #pragma mark - Keyboard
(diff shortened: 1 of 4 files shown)


appleguy:
@leotumwattana this seems really great, thank you for writing the patch!!  All these improvements are making me especially excited to see UITextField support be added soon ... not sure who will pick up that task, but it will be really wonderful when we have a type enum of some kind and can also rely on the better performance & special features of UITextField without developers having to switch out the classes manually.

appleguy:
@Adlai-Holler or @maicki should review this one.  Please ping us on Slack if you don't hear back by mid-next-week and I'll take a deeper pass myself to approve.

leotumwattana:
@appleguy Thanks for this really great framework. I've been delving more and more into it these past few days to see how and if we can adopt it to our app. Falling in ❤️  with it. Looking forward to seeing this framework continue to grow and to see if my friends and I will be able to contribute to it more in the future. Cheers!

@Adlai-Holler @maicki: Do let me know your comments and to let me know if I need to make any changes for this patch. 👍

Adlai-Holler:
@leotumwattana Hi! I'm really glad you're enjoying the framework! Would you be willing to re-open this PR with `master` as the base branch, rather than `2.0`?

AsyncDisplayKit/ASEditableTextNode.mm
```diff
@@ -460,7 +476,10 @@ - (void)_layoutTextView
   [_textKitComponents.layoutManager invalidateLayoutForCharacterRange:NSMakeRange(0, [_textKitComponents.textStorage length]) actualCharacterRange:NULL];
 
   // When you type beyond UITextView's bounds it scrolls you down a line. We need to remain at the top.
-  [_textKitComponents.textView setContentOffset:CGPointZero animated:NO];
+  // [Question - leotumwattana] Is there a good reason why we are forcing content offset to top? Why not
+  // keep at the current offset of the text view?
+  CGPoint offset = _textKitComponents.textView.contentOffset;
+  [_textKitComponents.textView setContentOffset:offset animated:NO];
```
Adlai-Holler: This change seems reasonable to me but I'd rather separate this change out from this PR, since:

- UITextView behavior is weird and sometimes counterintuitive.
- One change per PR makes future debugging easier.
- As written, it seems like this line won't have any effect. It's basically: `textVIew.contentOffset = textView.contentOffset`.

leotumwattana:
@Adlai-Holler: Sure thing. It's getting pretty late into the night here in HK. Let me separate into different PR and base it off master and resubmit.

AsyncDisplayKit/ASEditableTextNode.mm
leotumwattana: Yes, it does look strange, but what I've found is that without this strange `textView.contentOffset = textView.contentOffset` the textView jumps up/down during editing under certain cases when the editableTextNode is scrollable due to `.maximumLinesToDisplay`. 

Whereas the original code (below) will cause the editableTextNode to jump up to top:
```
[_textKitComponents.textView setContentOffset:CGPointZero animated: NO];
```

I'll add an additional comment when I resubmit this as two separate PR.

leotumwattana:
Just submitted a separate PR #2867 in replacement for this PR. So closing.

Issue #2752, opened 2016-12-11
[ASDimension] Public header has a lot of definitions around ASLayoutElementSize (internal-only)


appleguy:
ASDimension.h is a much longer header than it should be for such a core, simple type.  At the least, we should move down any less-crucial definitions below the basics.  I'd say these are the basics, along with a few of their helper functions:

```
typedef NS_ENUM(NSInteger, ASDimensionUnit) {
  /** This indicates "I have no opinion" and may be resolved in whatever way makes most sense given the circumstances. */
  ASDimensionUnitAuto,
  /** Just a number. It will always resolve to exactly this amount. This is the default type. */
  ASDimensionUnitPoints,
  /** Multiplied to a provided parent amount to resolve a final amount. */
  ASDimensionUnitFraction,
};

typedef struct {
  ASDimensionUnit unit;
  CGFloat value;
} ASDimension;

/**
 * Expresses an inclusive range of sizes. Used to provide a simple constraint to layout.
 */
typedef struct {
  CGSize min;
  CGSize max;
} ASSizeRange;
```

But then, even before you get to a function about how to create any of these, we also see the much larger ASLayoutElementSize.  This is a very elegant struct and works ideally for internal purposes for now, but shouldn't be exposed yet, particularly because we don't have good documentation for what it's for.

```
/**
 * A struct specifying a ASLayoutElement's size. Example:
 *
 *  ASLayoutElementSize size = (ASLayoutElementSize){
 *    .width = ASDimensionMakeWithFraction(0.25),
 *    .maxWidth = ASDimensionMakeWithPoints(200),
 *    .minHeight = ASDimensionMakeWithFraction(0.50)
 *  };
 *
 *  Description: <ASLayoutElementSize: exact={25%, Auto}, min={Auto, 50%}, max={200pt, Auto}>
 *
 */
typedef struct {
  ASDimension width;
  ASDimension height;
  ASDimension minWidth;
  ASDimension maxWidth;
  ASDimension minHeight;
  ASDimension maxHeight;
} ASLayoutElementSize;
```

Lastly, these methods at the top are unclear in their differences.  one uses CGFLOAT and the other FLT, but it's unclear what the practical differences are of that, especially from names that are only Size vs Layout.  They are currently the very first thing listed in the header, which also seems a bit too prominent for these declarations, but if they were reduced to 2 (and should be) -- then keeping them relatively prominent somewhere in the file is worthwhile to encourage their usage.

```
[...]

PR #2794, opened 2016-12-17
[ASDimension] Simplify header file to focus on the most essential types for app developers.

maicki:
Public header of `ASDimension` has a lot of definitions e.g. around `ASLayoutElementSize`. Let's cleanup and get it into a better structure.

- Move `ASLayoutElementSize` to `ASDimensionInternal`
- Remove helper categories for `NSNumber`: They where never 100% thought through so let's get rid of them for now
- Move all deprecated code into `ASDimensionDeprecated`

Resolves: #2752


AsyncDisplayKit/AsyncDisplayKit.h
```diff
@@ -48,6 +48,8 @@
 
 #import <AsyncDisplayKit/ASLayout.h>
 #import <AsyncDisplayKit/ASDimension.h>
+#import <AsyncDisplayKit/ASDimensionInternal.h>
```
appleguy: Internal files shouldn't be included in the umbrella; OK to remove in another diff though!

AsyncDisplayKit/Layout/ASLayoutElement.h
```diff
@@ -8,7 +8,7 @@
 //  of patent rights can be found in the PATENTS file in the same directory.
 //
 
-#import <AsyncDisplayKit/ASDimension.h>
+#import <AsyncDisplayKit/ASDimensionInternal.h>
```
appleguy: Similar as the Umbrella, if it is true that this file can be imported externally (which I think that it can) -- then ideally, the *Internal.h would be imported only in the .m/.mm file.

appleguy:
After we look at the build failure, I'd suggest landing this before fixing the Internal.h imports, so that we can get ready to launch 2.0.1.  We can always fix the Internal.h in 2.1.  However, it would be nice to start a more consistent style around Internal.h naming (ASDisplayNodeInternal was off to a bad start!).  In fact, we should consider if ASDimensionPrivate is a better name since it's intended for use across the framework, whereas Internal is normally used for within one class.

appleguy:
test this please

appleguy:
@maicki thanks again for working on this :).  I can't see what the cause of the failure is, but let's land this as soon as you have the opportunity.

maicki:
@garrettmoon @appleguy Fixed the build errors. Let's get that in for 2.0.2.

PR #2795, opened 2016-12-18
[AS{Overlay,Background}LayoutSpec] Don't allow nil layout elements

### LINKED ISSUE #2747: [ASBackgroundLayoutSpec] Fix to not allow a nil node
Running into this assertion https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASLayoutSpec+Subclasses.mm#L101 when setting nil as the background element in an ASBackgroundLayoutSpec. 

The documentation says background is allowed to be nil: https://github.com/facebook/AsyncDisplayKit/blob/master/AsyncDisplayKit/Layout/ASBackgroundLayoutSpec.h#L29

I assume `ASOverlayLayoutSpec` has the same issue.

If the background element shouldn't be nil, the method signature should be updated to take in only nonnull values for background and the initializer should also assert from receiving a nil value same as the child param.

hannahmbanana:
Resolves #2747.

Michael and I couldn't think of a use case for allowing an `ASBackgroundLayoutSpec` with a nil background layout element. cc @nguyenhuy

### DIFF
--- a/AsyncDisplayKit/Layout/ASBackgroundLayoutSpec.h
+++ b/AsyncDisplayKit/Layout/ASBackgroundLayoutSpec.h
@@ -20,15 +20,15 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  * Background layoutElement for this layout spec
  */
-@property (nullable, nonatomic, strong) id<ASLayoutElement> background;
+@property (nonatomic, strong) id<ASLayoutElement> background;
 
 /**
  * Creates and returns an ASBackgroundLayoutSpec object
  *
  * @param child A child that is laid out to determine the size of this spec.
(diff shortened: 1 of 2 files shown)


master-nevi:
@hannahmbanana shouldn't we also add in some assertions too? It may not be as important for newer ASDK adopters, but for code that is being migrated to v2, the assertions I added [here](https://github.com/master-nevi/AsyncDisplayKit/commit/1f6ec77871c7623a8a0d14b12c1203fb86171db3) helped a lot in determining the cause of https://github.com/facebook/AsyncDisplayKit/issues/2747

hannahmbanana:
@master-nevi - would you be willing to put up a PR? I just noticed your PR above that I totally missed the first time I read this. :'(

master-nevi:
@hannahmbanana PR opened!

PR #2798, opened 2016-12-19
Master

GR-harry:
[tvOS] Add default focus states to ASControlNode & ASImageNode.


hannahmbanana:
@GR-harry - did you create this PR accidentally? I can't quite tell what's going on here. :) It looks like your comment references #1273.

PR #2799, opened 2016-12-19
Support visibility events in ASScrollNode

### LINKED ISSUE #2791: ASScrollNode doesn't get visibility events
Scott: "the reason for this is that the view of the node is a UIScrollView and not an _ASDisplayView.  That's the case even though the +layerClass is specially overridden to provide _ASDisplayLayer, but the _ASDisplayView is apparently the only thing driving the visibility callbacks.

[thinks for a bit] :)

I think the right fix is to make ASScrollNode behave as though it were a layer-backed node inside ASDisplayNode.  There is a way that layerBacked nodes work with this, which is the CALayer "actionForKey:" orderIn/orderOut calls."

nguyenhuy:
This should fix #2791.

### DIFF
--- a/AsyncDisplayKit/ASScrollNode.mm
+++ b/AsyncDisplayKit/ASScrollNode.mm
@@ -9,7 +9,9 @@
 //
 
 #import "ASScrollNode.h"
+#import "ASDisplayNodeExtras.h"
 #import "ASDisplayNode+FrameworkSubclasses.h"
+#import "ASDisplayNode+FrameworkPrivate.h"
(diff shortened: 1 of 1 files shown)


nguyenhuy:
cc @hannahmbanana, @appleguy, @garrettmoon. I think this change is safe. Would be great if it can make to 2.0.1 because it helps @hannahmbanana to avoid a workaround in her Pinterest code.

hannahmbanana:
Thank you @nguyenhuy! I tested it out with my code and it works for me. Thanks again!

Issue #2436, opened 2016-10-19
Table View Thrash Tests Should Be Enabled


Adlai-Holler:
Currently `ASTableViewThrashTests.testThrashingWildly` is disabled because, when run alongside other tests, we see `Node should always be marked invisible before deallocating` issues. 

In #2435, we can go the easy route if we solve this issue and re-enable `ASTableViewThrashTests.testThrashingWildly`. The hard route is pretty brutal.

Adlai-Holler:
The first step is to use the exception breakpoint to find out where that exception is thrown when running all unit tests with Cmd+U. It is thrown from other test methods, which implies that the table view or window from the thrash testing is not cleaned up properly.

One initial approach is to look at how the window used in the thrash tests is torn down, and try explicitly removing the table view from the window before letting the window deallocate.

Adlai-Holler:
After a good bit of work on this yesterday, I have gotten the test method to always fail _inside the test case_ if it's going to fail. Unfortunately the thrash tests still contain surplus cases of "Node must be marked invisible before deallocation" and so I probably can't reenable them today. Later today I'll open a diff that modifies the test method to make sure it fails inside the test case. 

Removing from the milestone since the code isn't public-facing and the underlying issues I've exposed probably won't be fixed today.

Adlai-Holler:
Resolved in linked PRs

Issue #2203, opened 2016-09-06
ASTableView crashes with assertion "Node should always be marked invisible before deallocating" depending on header height.


mikhailmulyar:
Hi. I have simple ASTableNode showing sectioned list. There is a really strange behaviour. App stops with assertion "Node should always be marked invisible before deallocating" in a second after table content shows when reallocing my custom ASCellNode. There is nothing happening at the moment, no updates or reloads.
Most strange thing is that crash depends on tableView header height. If i return 15 in tableView(tableView: UITableView, viewForHeaderInSection section: Int) app runs ok. If i return 18 it crashes. So i guess it is something related to how cells positioned in the view.
Here is the screen at crash state. There is searchBar - 44 height, headers are 18, and cell nodes are 41.
I'm using latest master version.
![simulator screen shot 07 sep 2016 00 55 34](https://cloud.githubusercontent.com/assets/2234720/18292442/340d30de-749e-11e6-9215-30219a418a4a.png)

hannahmbanana:
@mikhailmulyar: Thanks for reporting this! This issue looks similar to #2205, which @Adlai-Holler is planning on investigating soon.

Adlai-Holler:
Hi @mikhailmulyar I'm investigating this for you. I've got a couple quick questions:

- Are any of your nodes set with `shouldRasterizeDescendants=YES`? Any of them `layerBacked=YES`?
- How are your section headers working? Are you implementing `viewForHeaderInSection` and returning `cellNode.view`?
- Which node is hitting the assertion failure? A node for a row, a node for a section, or one of the subnodes?

I know those assertion failures can be extremely irritating so I'll do my best to get this fixed the right way!

mikhailmulyar:
Hi @Adlai-Holler. 
1) I've used shouldRasterizeDescendants in ASCellNode. However when i received bug, i'm pretty sure i've tried to turn it off, and it did not change anything
2) Yes. I'm returning cellNode.view and storing them in array. However it does not matter, since i've completely commented viewForHeaderInSection and only returned height in heightForHeaderInSection. The screenshot is made in that state, as you can see headers are empty, there is nothing at all.
3) Custom subclass of ASCellNode for row.

Adlai-Holler:
@mikhailmulyar Hi! We've been super busy preparing for the 2.0 launch but this is still an important issue for us. 

Curious if there has been any change in behavior from the framework, or have you found a workaround that unblocks you on this issue?

[...]

PR #2800, opened 2016-12-19
Fix Issue Where Thrash Tester Would Throw Exceptions During Other Tests

Adlai-Holler:
CC #2436 #2203 #2658 
This diff ensures that all cell nodes are deallocated before the thrash test method returns. 

We now have another good test case for cell node visibility issues linked above. Unfortunately, that test case is currently failing, so the test has to remain disabled.


Adlai-Holler:
It's not important that this is landed before or after the 2.0.1 release, since it doesn't touch public-facing code in a significant way.

Issue #2802, opened 2016-12-19
[Swift] ASDisplayNode deallocation introspection may crash with Swift generic types.


haritowa:
ASDealloc crash when try to find ivars, that need dealloc on main thread for swift generic type, which contains props of generic type, for example:
```
class RightImageTrackCellNode<AdapterType: ImageTrackCellAdapter>{
    private var adapter: AdapterType!
    …
}
```
where `ImageTrackCellAdapter` is swift protocol.

My explanation may look complicated, but example extremely simple. When `ivar_getTypeEncoding` invoked on adapter ivar it returns `nil`, which leads to crash

![pasted image at 2016_12_19 10_39 pm](https://cloud.githubusercontent.com/assets/6107211/21328299/471a6a5a-c644-11e6-9b63-f9bdf300d10a.png)

PR #2803, opened 2016-12-19
[ASDisplayNode Deallocation] Fix for crash when certain types are used as instance variables.

haritowa:
Fix [this](https://github.com/facebook/AsyncDisplayKit/issues/2802) issue

### DIFF
--- a/AsyncDisplayKit/Private/ASInternalHelpers.m
+++ b/AsyncDisplayKit/Private/ASInternalHelpers.m
@@ -107,7 +107,7 @@ BOOL ASClassRequiresMainThreadDeallocation(Class class)
 Class _Nullable ASGetClassFromType(const char *type)
 {
   // Class types all start with @"
-  if (strncmp(type, "@\"", 2) != 0) {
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Private/ASInternalHelpers.m
```diff
@@ -107,7 +107,7 @@ BOOL ASClassRequiresMainThreadDeallocation(Class class)
 Class _Nullable ASGetClassFromType(const char *type)
 {
   // Class types all start with @"
-  if (strncmp(type, "@\"", 2) != 0) {
+  if (type == nil || strncmp(type, "@\"", 2) != 0) {
```
Adlai-Holler: It's super minor point, but we should use `NULL` instead of `nil` here since it's not an object. 🤓

haritowa: Done, looks better?)

Adlai-Holler:
@haritowa Thanks!

PR #2805, opened 2016-12-19
Range Controller Explicitly Marks Disappeared Nodes as Invisible

Adlai-Holler:
This should resolve a huge swath of our cell node visibility issues, such as #2436 #2203 #2658.

### DIFF
--- a/AsyncDisplayKit/Details/ASRangeController.mm
+++ b/AsyncDisplayKit/Details/ASRangeController.mm
@@ -170,6 +170,24 @@ static UIApplicationState __ApplicationState = UIApplicationStateActive;
   }
 }
 
+// Clear the visible bit from any nodes that disappeared since last update.
+// Currently we guarantee that nodes will not be marked visible when deallocated,
+// but it's OK to be in e.g. the preload range. So for the visible bit specifically,
+// we add this extra mechanism to account for e.g. deleted items.
+//
+// NOTE: There is a minor risk here, if a node is transferred from one range controller
+// to another before the first rc updates and clears the node out of this set. It's a pretty
+// wild scenario that I doubt happens in practice.
+- (void)_setVisibleNodes:(ASWeakSet *)newVisibleNodes
+{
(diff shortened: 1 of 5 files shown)


Adlai-Holler:
I've confirmed that when paired with #2800, it fixes the dealloc-while-visible cases we were seeing in the thrash testers.

AsyncDisplayKit/Details/ASWeakSet.m
```diff
@@ -22,7 +22,7 @@ - (instancetype)init
 {
   self = [super init];
   if (self) {
-    _hashTable = [NSHashTable weakObjectsHashTable];
```
appleguy: So did this previously de-duplicate objects by using isEqual?  slow and shitty!  lol

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -34,6 +34,7 @@ @interface ASRangeController ()
   BOOL _layoutControllerImplementsSetVisibleIndexPaths;
   BOOL _layoutControllerImplementsSetViewportSize;
   NSSet<NSIndexPath *> *_allPreviousIndexPaths;
+  ASWeakSet<ASCellNode *> *_visibleNodes;
```
appleguy: Interesting.  There used to be tracking like this, and when the class was rewritten to be more functional, I forgot about the deletion case :).  I might argue that we should try to handle deletions explicitly, because we could also call the other two range exit methods (although this will incur cost, I do fear that some clients may depend on this for cache invalidation or such).

That said, this is an excellent solution to an important problem, and any further changes would be prospective improvements.  Thanks @Adlai-Holler for working on this challenging issue and the tricky testing that comes with developing such a change!

Should we file an issue for writing a unit test to expose the problem from before this diff?

Adlai-Holler:
Thanks for the awesome feedback @appleguy ! Added a (previously-failing) unit test and managed to catch an unexpected case (deleting the last item in a collection view) and now that's handled too. Time to merge!

PR #2808, opened 2016-12-20
Range Controller Uses Last Scroll Direction, Defaults to Down/Right

### LINKED ISSUE #2806: Range Controller Should Default to Down/Right Before User Scrolls
Currently the logic inside `ASDirectionalScreenfulBufferVertical` means that if the user isn't scrolling, we will point the leading range upward which is wacky.

What we would do ideally is to use the "last-known scroll direction" for the user, with a default of down/right, so that ranges point in the same direction until the user actually scrolls in some other direction.

Adlai-Holler:
Resolves #2806

### DIFF
--- a/AsyncDisplayKit/Details/ASFlowLayoutController.mm
+++ b/AsyncDisplayKit/Details/ASFlowLayoutController.mm
@@ -57,17 +57,9 @@
   ASRangeTuningParameters      tuningParameters  = [self tuningParametersForRangeMode:rangeMode rangeType:rangeType];
 
   if (_layoutDirection == ASFlowLayoutDirectionHorizontal) {
-    ASDisplayNodeAssert(scrollDirection == ASScrollDirectionNone ||
-                        scrollDirection == ASScrollDirectionLeft ||
-                        scrollDirection == ASScrollDirectionRight, @"Invalid scroll direction");
-
     viewportDirectionalSize = viewportSize.width;
     directionalBuffer = ASDirectionalScreenfulBufferHorizontal(scrollDirection, tuningParameters);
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Details/ASFlowLayoutController.mm
```diff
@@ -57,17 +57,9 @@ - (NSSet *)indexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(
   ASRangeTuningParameters      tuningParameters  = [self tuningParametersForRangeMode:rangeMode rangeType:rangeType];
 
   if (_layoutDirection == ASFlowLayoutDirectionHorizontal) {
-    ASDisplayNodeAssert(scrollDirection == ASScrollDirectionNone ||
-                        scrollDirection == ASScrollDirectionLeft ||
-                        scrollDirection == ASScrollDirectionRight, @"Invalid scroll direction");
-
```
Adlai-Holler: Needed to remove these because the range controller asks the layout controller about `[.down, .right]` before the user's first scroll.

It's a little bit hacky, I know, but two excuses:
- The alternative would be for the range controller to have a new code path to determine the scrollable directions. We could do it, but it's added code.
- These assertions have to go anyway. If I use a vertical flow layout, and I set `alwaysBounceHorizontally` and scroll sideways/diagonally that's a valid usage of flow layout and it would fail this assertion.

Adlai-Holler:
I'm gonna go ahead and land this because I consider it pretty low-risk, I need it for some further work I'm doing in a different area, and everyone is doing Christmas!

PR #2810, opened 2016-12-20
[ASViewController] Support optional node

### LINKED ISSUE #2760: [ASViewController] Add unit test or sample app that uses an ASVC without self.node, using .view manually.
Because Objective-C and Swift only support Single Inheritance, it's critical that ASViewController can be used as a BASE class in apps that have only partial usage of ASDK.

Most apps have some base VC class, like PIBaseViewController, so the only way to adopt ASVC is to make their base VC inherit from it.  This means that it is crucial that ASVC work with:

-initWithCoder / nib
-manual / regular usage of the .view property

This should be possibly to achieve by simply gating any ASVC behaviors to do nothing at all if self.node == nil.  Then we should make sure the documentation highlights that this is safe and supported.

nguyenhuy:
This PR makes ASViewController's node optional. This allows developers to use ASViewController like a normal UIViewController. More importantly, it can be used as a base class for view controllers among which some use a node hierarchy and some don't. 

Resolve #2760.

### DIFF
--- a/AsyncDisplayKit/ASViewController.mm
+++ b/AsyncDisplayKit/ASViewController.mm
@@ -24,6 +24,7 @@
 
 @implementation ASViewController
 {
+  BOOL _nodeProvided;
   BOOL _ensureDisplayed;
   BOOL _automaticallyAdjustRangeModeBasedOnViewEvents;
   BOOL _parentManagesVisibilityDepth;
@@ -38,9 +39,7 @@
     return nil;
   }
   
-  _node = [[ASDisplayNode alloc] initWithViewBlock:^UIView * _Nonnull{
-    return [[UIView alloc] init];
-  }];
+  _node = [[ASDisplayNode alloc] init];
   [self _initializeInstance];
   
   return self;
@@ -52,9 +51,7 @@
     return nil;
   }
   
-  _node = [[ASDisplayNode alloc] initWithViewBlock:^UIView * _Nonnull{
-    return [[UIView alloc] init];
-  }];
+  _node = [[ASDisplayNode alloc] init];
   [self _initializeInstance];
   
   return self;
@@ -66,6 +63,7 @@
     return nil;
   }
   
+  _nodeProvided = YES;
   _node = node;
   [self _initializeInstance];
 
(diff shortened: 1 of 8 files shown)


hannahmbanana:
This is really awesome. I've heard that ASViewController's `initWithNode:` pattern is the most confusing aspect of ASDK.

maicki:
@nguyenhuy After looking over this PR it's a really nice improvement idea, but I'm not pretty sure if we should introduce having to test if the `_node` is `nil` in almost every method. Especially going forward we would need to be careful to always consider that.

What about having something like a `ASViewControllerNullNode` that would act like a proxy and this would be used as root node for the `ASViewController` and has a reference to the view controller if no node is given? In that null node we would just not do anything I guess in case we call method's like `layoutThatFits:` but if the view is accessed we actually would return the view from the view controller. I would have to lookup in the `ASViewController` implementation what we exactly would need, but I think that would be a bit _cleaner_ solution.

nguyenhuy:
@maicki Thank you for the feedbacks. I've came up with a simpler fix. In case users don't provide ASViewController with a node (by using `-initWithNibName:bundle:` or `-initWithCoder:`, I create a default node that simply loads an UIView. And so the node is non-optional as it has always been. This introduces a small overhead but helps reduce complexity of ASViewController. The overhead might even encourage people to opt-in to a node backed hierarchy and take full advantage of ASViewController.

I tested this approach with ASDKgram and didn't find any problem. Given the fact that we currently don't even support `-initWithNibName:bundle:` or `-initWithCoder:`, this change shouldn't cause any new regression on existing codebases and thus it should be safe.

AsyncDisplayKit/ASViewController.mm
```diff
@@ -34,14 +34,30 @@ @implementation ASViewController
...
 {
-  ASDisplayNodeAssert(NO, @"ASViewController requires using -initWithNode:");
-  return [self initWithNode:[[ASDisplayNode alloc] init]];
+  if (!(self = [super initWithCoder:aDecoder])) {
+    return nil;
+  }
+  
+  _node = [[ASDisplayNode alloc] initWithViewBlock:^UIView * _Nonnull{
+    return [[UIView alloc] init];
+  }];
```
Adlai-Holler: We should use simply `[[ASDisplayNode alloc] init]` here because using a view block results in a wrapper node, which has limited capacity compared to a node using `_ASDisplayView`.

Adlai-Holler:
It looks like this has been updated by @maicki in the linked PR. Closing.

PR #2814, opened 2016-12-20
These messages must be called on main.

### LINKED ISSUE #2813: [ASNetworkImageNode] Setting url on background thread causes crash
I'm testing out an app on latest master (3e53d26686). I'm running into a crash when setting the url property from a background thread. Looks to be because the `didEnterPreloadState` asserts that it must be called from the main thread.

Here is the stack trace
```
	0   CoreFoundation                      0x000000010f6fd34b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000010ed6721e objc_exception_throw + 48
	2   CoreFoundation                      0x000000010f701442 +[NSException raise:format:arguments:] + 98
	3   Foundation                          0x000000010e8fde4d -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 195
	4   AsyncDisplayKit                     0x0000000109dc4a50 -[ASDisplayNode didEnterPreloadState] + 304
	5   AsyncDisplayKit                     0x0000000109e6df14 -[ASNetworkImageNode didEnterPreloadState] + 68
	6   AsyncDisplayKit                     0x0000000109dc4141 __35-[ASDisplayNode recursivelyPreload]_block_invoke + 81
	7   AsyncDisplayKit                     0x0000000109dd8c06 ASDisplayNodePerformBlockOnEveryNode + 1686
	8   AsyncDisplayKit                     0x0000000109dc40e1 -[ASDisplayNode recursivelyPreload] + 65
	9   AsyncDisplayKit                     0x0000000109dc408f -[ASDisplayNode setNeedsPreload] + 111
	10  AsyncDisplayKit                     0x0000000109e6c79b -[ASNetworkImageNode setURL:resetToDefault:] + 635
	11  AsyncDisplayKit                     0x0000000109e6c4e2 -[ASNetworkImageNode setURL:] + 82
```

garrettmoon:
Resolves #2813

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -3029,8 +3029,10 @@ void recursivelyTriggerDisplayForLayer(CALayer *layer, BOOL shouldBlock)
 
 - (void)recursivelyClearContents
 {
-  ASDisplayNodePerformBlockOnEveryNode(nil, self, YES, ^(ASDisplayNode * _Nonnull node) {
-    [node clearContents];
+  ASPerformBlockOnMainThread(^{
+    ASDisplayNodePerformBlockOnEveryNode(nil, self, YES, ^(ASDisplayNode * _Nonnull node) {
+      [node clearContents];
+    });
   });
 }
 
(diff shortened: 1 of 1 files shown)


garrettmoon:
@nguyenhuy does this look right to you?

garrettmoon:
@nguyenhuy it should be fine because nothing is retaining the block itself?

nguyenhuy:
@garrettmoon I was just worrying that in rare cases, we might end up keeping these nodes around for a bit longer than needed. But this is definitely an improvement.

garrettmoon:
@nguyenhuy oh good point! I wonder if they should be though?

nguyenhuy:
@garrettmoon Thinking more about it, I think we actually want them to be around. For example, an image node should be kept alive until it clears its contents (i.e cache entry, placeholder image, etc).

appleguy:
@nguyenhuy could you share with me (maybe in an email, once you're up tomorrow) what changed in this area to cause off-main thread calls to occur?  This didn't occur in practice before 2.0, right?

nguyenhuy:
@appleguy it's an external cause actually (https://github.com/facebook/AsyncDisplayKit/issues/2813). Developers can call '-[ASNetworkImageNode setURL]' in background which then triggers '-setNeeedPreload'. Since it's a legitimate use case to call '-setURL' off main, I think this is necessary.

Issue #2711, opened 2016-12-05
Random crash on node dealloc (Node should always be marked invisible before deallocating)


smyrgl:
Version: a2e75152f87687c89dd5566ecd86fac2ff5f09d1

I'm having a weird crash that is randomly occurring that seems to be inside the ADK dealloc process, the exception is `Node should always be marked invisible before deallocating.` Here's the stack trace:

```
2016-12-05 12:23:28.245 Disqus[60615:10514027] *** Assertion failure in -[Disqus.ChannelCategoryListCellNode dealloc], /Users/john/code/disqus-ios/Pods/AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:439
2016-12-05 12:23:28.256 Disqus[60615:10514027] *** Assertion failure in -[ASTextNode dealloc], /Users/john/code/disqus-ios/Pods/AsyncDisplayKit/AsyncDisplayKit/ASDisplayNode.mm:439
2016-12-05 12:23:28.302 Disqus[60615:10514027] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Node should always be marked invisible before deallocating. Node: <ASTextNode: 0x7fc9d0079000; text = "Select Start">'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000111c4e34b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00000001116ab21e objc_exception_throw + 48
	2   CoreFoundation                      0x0000000111c52442 +[NSException raise:format:arguments:] + 98
	3   Foundation                          0x000000010cfd2e4d -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 195
	4   AsyncDisplayKit                     0x000000010b14de06 -[ASDisplayNode dealloc] + 438
	5   AsyncDisplayKit                     0x000000010b267799 -[ASTextNode dealloc] + 345
	6   libobjc.A.dylib                     0x00000001116bfb12 _ZN11objc_object17sidetable_releaseEb + 212
	7   AsyncDisplayKit                     0x000000010b0b243f -[_ASDisplayView didMoveToSuperview] + 1711
	8   UIKit                               0x000000010eee49c1 __45-[UIView(Hierarchy) _postMovedFromSuperview:]_block_invoke + 932
	9   UIKit                               0x000000010eee45a2 -[UIView(Hierarchy) _postMovedFromSuperview:] + 857
[...]

PR #2815, opened 2016-12-21
[ASDataController] Ensure ASRangeController has an opportunity to update visible nodes before they're deleted by -reloadData path.

appleguy:
This change is specific to the reloadData path, which had the last-known occurrence of "deallocated while marked visible".

https://github.com/facebook/AsyncDisplayKit/issues/2711


appleguy:
The reporter could reproduce the bug and this fixed it straight away.

It makes sense because the set is weak, and the range controller doesn't have time to run before they are deallocated otherwise.  See the lower levels of the stack in the github issue.

PR #2817, opened 2016-12-21
[ASStackLayoutSpec] Clean up ASStackPositionedLayout

nguyenhuy:
- Remove `lastChildOffset`. It's unncessary after #2627.
- Remove the version of `stackedLayout()` that takes 1 less param. Isn't worth the complexity.


nguyenhuy:
cc @maicki.

PR #2834, opened 2016-12-24
[ASVideoNode] Add missing strong attributes

### LINKED ISSUE #2832: Suggesting fix an annoying build-warning?
Since my project set treat all warning as error, that's pretty annoying indeed.

```
ASVideoNode.h
@property (nonatomic) NSString *gravity;
```

```
Default property attribute 'assign' not appropriate for non-gc object
No 'assign', 'retain' or 'copy' attribute is specified - 'assign' is assumed
```

hannahmbanana:
Resolves #2832.


maicki:
@hannahmbanana Thanks for fixing! The attribute should be 'copy' though.

PR #2843, opened 2016-12-28
[ASTableNode] Fix returning nil from willSelectRowAtIndexPath

### LINKED ISSUE #2840: -tableNode:willSelectRowAtIndexPath: does not allow returning nil
If you implement `-tableNode:willSelectRowAtIndexPath:` and return `nil`, the first cell node (index path 0,0) ends up being selected, instead of preventing selection from occurring.

This doesn't match the expected UITableView behavior, where the [docs](https://developer.apple.com/reference/uikit/uitableviewdelegate/1614943-tableview?language=objc) state:

> Return `nil` if you don't want the row selected.

Stepping through the debugger shows that `ASTableView`:
1. [Asks the asyncDelegate to return an `indexPath`](https://github.com/facebook/AsyncDisplayKit/blob/e264d94dde78d9bfe98cb71be55ef14356f4a7f2/AsyncDisplayKit/ASTableView.mm#L913) - in my case this `result` is sometimes `nil`.
2. [Immediately converts the possibly-nil index path](https://github.com/facebook/AsyncDisplayKit/blob/e264d94dde78d9bfe98cb71be55ef14356f4a7f2/AsyncDisplayKit/ASTableView.mm#L914) - due to implementation details, this now comes back with a non-nil `result` pointing to index path 0,0.
3. [Table view is told to select index path 0,0](https://github.com/facebook/AsyncDisplayKit/blob/e264d94dde78d9bfe98cb71be55ef14356f4a7f2/AsyncDisplayKit/ASTableView.mm#L915).

The simplest fix would probably be to make `ASTableView` return `nil` immediately, in cases where [the call](https://github.com/facebook/AsyncDisplayKit/blob/e264d94dde78d9bfe98cb71be55ef14356f4a7f2/AsyncDisplayKit/ASTableView.mm#L913) to the `asyncDelegate` returned `nil`. In other words, keep `nil` as `nil` and return it to UIKit as-is.

### Example

I'm attempting to prevent an already-selected row from being selected again (and subsequently calling the `didSelect` delegate method).

```objc
- (NSIndexPath *)tableNode:(ASTableNode *)tableNode willSelectRowAtIndexPath:(NSIndexPath *)indexPath {
    // Check if this indexPath is already selected.
    if ([tableNode.indexPathsForSelectedRows containsObject:indexPath]) {
        return nil; // Don't allow re-selecting the same index path
    }
    return indexPath;
}
```

Issue in action:

table-reselect

Note how attempting to re-select Jim ends up selecting Meghan, instead of doing nothing.

### Workaround

Until this is fixed, I've implemented `-tableView:willSelectRowAtIndexPath:` instead, which doesn't have this issue but **is** marked as deprecated.

rewandy:
Fixes #2840

Allow implementations of `-tableNode:willSelectRowAtIndexPath:` to return `nil`,
in order to match UIKit behavior. Previously, returning `nil` caused the first
cell node to become selected.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -911,6 +911,9 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
       return indexPath;
     } else {
       result = [_asyncDelegate tableNode:tableNode willSelectRowAtIndexPath:result];
+      if (result == nil) {
+        return nil; // If consumer explicitly returned nil, just pass it as-is to the table view.
+      }
       result = [self convertIndexPathFromTableNode:result waitingIfNeeded:YES];
       return result;
     }


appleguy:
@rewandy Thanks!  Does this same method exist for collection views?  I'm not sure if it does, but if so, could you check that one too?

@Adlai-Holler would be the best person to confirm this is the right fix.  How was a nil value handled previously?  Wouldn't the convert index path method provide a nil response if it had a nil input?  That seems important to determine.

rewandy:
The `UICollectionView` equivalent is [`shouldSelectItemAtIndexPath`](https://developer.apple.com/reference/uikit/uicollectionviewdelegate/1618095-collectionview?language=objc) and ASDK's [implementation](https://github.com/facebook/AsyncDisplayKit/blob/99c5021be7fe9d6bb428e3c335ab9a11e8fc15de/AsyncDisplayKit/ASCollectionView.mm#L957) of that is much more straight-forward (just takes the `BOOL` and gives it to the collection view).

I'd definitely be interested in why a nil input to the conversion method results in a non-nil output. From what I remember during debugging, it was actually the [eventual](https://github.com/rewandy/AsyncDisplayKit/blob/96b2fd763ac190e5c35096023e372f0cd774b6a3/AsyncDisplayKit/ASTableView.mm#L594) calls into the data controller that come back with a non-nil result. I suspect that might be intentional though, so I didn't want to start changing data controller internals...

Adlai-Holler:
Hey @rewandy you're absolutely right! I dug into the calls you linked, and created a more general fix in #2856. Could you give that PR a look and tell me what you think?

PR #2844, opened 2016-12-29
[ASVideoNode] Remove duplicated _delegate ivar

nguyenhuy:
It's confusing to have 2 `_delegate` ivars at the same time: one `ASVideoNode` and another in its super class (e.g `ASNetworkImageNode`), and both of them are accessed/set by the same pair of getter/setter. It also introduces a bit of boilerplate code in `ASVideoNodeTests`.

### DIFF
--- a/AsyncDisplayKit/ASVideoNode.mm
+++ b/AsyncDisplayKit/ASVideoNode.mm
@@ -40,7 +40,6 @@ static NSString * const kRate = @"rate";
 
 @interface ASVideoNode ()
 {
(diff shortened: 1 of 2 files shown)


appleguy:
@nguyenhuy This is a good improvement!  Any idea why we don't get warnings for these kind of collisions?  Like when we declared _propertyLock overwriting the base class ones, we didn't get warnings...I wonder if _delegateFlags also exists in Network or ImageNode?

PR #2848, opened 2016-12-31
Add Support for IGListKit

Adlai-Holler:
I'm excited to announce that I've been working on this, and I think it's ready for beta testing. 

IGListKit is a great framework and this, if landed, would supersede #2566. It's true that IGListKit doesn't support table views, but the benefits outweigh that cost I believe.

- Adds a new header, `ASCollectionNode+ListKitBeta.h` containing the public API for the integration.
  - New property, `ASCollectionNode.listAdapter`. Set this to a list adapter to get IGListKit going.
- Updates the `ASDKgram` example to have a new tab that uses ASDK+IGListKit.
- Adds a new workspace containing a few basic tests for ASDK+IGListKit. More to come here.
  - New workspace is necessary because otherwise the ASDK build in the main workspace will have to always link IGListKit. I tried a bunch of approaches for keeping it all inside one workspace but trust me, they all sucked.
- Adds a new subspec `IGListKit` to the podspec.
- Adds a new mode to `build.sh` to run the ListKit tests.

### To Make it Work

Users should have their section controller conform to `ASIGListSectionType` rather than just plain `IGListSectionType` (our protocol inherits from theirs) and add `ASIGSectionControllerCellForIndexImplementation` and `ASIGSectionControllerSizeForItemImplementation` into their section controllers, and then implement `constrainedSizeForItemAtIndex:` (optionally) and `nodeBlockForItemAtIndex:`.

### Hacks

- IGListKit has a trivial subclass of UICollectionView called `IGListCollectionView` that is subclassing-restricted. They assert that you use that class. So we use the (evil) function `class_setSuperclass(ASCollectionView, IGListCollectionView)` when you link IGListKit so that we inherit from it. If they would be willing to remove the subclass restriction, we could conditionally compile the `@interface` to inherit from the one we want.


Adlai-Holler:
@levi @garrettmoon @ay8s @chrisdanford @rahul-malik

ay8s:
Will check it out @Adlai-Holler, excited about some of the stuff we can do with this in ASDK.

Currently using tables in Buffer but have the Declarative stuff using Collection Views without much hassle. Curious on the decision to move towards IGListKit over ASDK's own implementation?

Adlai-Holler:
@ay8s The decision hasn't been made – it's only a proposal right now – but using IGListKit, which is quickly being recognized by the industry as an extremely powerful and reliable framework for building data-driven collection views, will seriously cut down our maintenance/API design burden and keep the scope of AsyncDisplayKit limited, so I obviously advocate very strongly for it. The section controller architecture is an extremely useful one, and their diffing mechanism has the concept of "updating an object" where we don't. If we can integrate with IGListKit, then we can spend our resources building other great stuff

ay8s:
@Adlai-Holler Awesome 👍 Started checking it out today and will see if I can get a branch together using it alongside #2566's implementation. Totally get keeping the ASDK scope limited to focus on other areas especially when IGListKit is so powerful. Will let you know if I spot anything!

Adlai-Holler:
Awesome, thanks @ay8s for looking into it!

ay8s:
@Adlai-Holler Got started with implementing this into Buffer this morning. Just about got it working and displaying updates from the first fetch from the Buffer API. Wanted to post some early thoughts. It definitely seems like a bit more work to get up and running compared to the ASDK diff implementation, that said I'm sure the benefits down the line will make it worth it.

One thing I did spot also, the ASDKGram app seems to crash after a few page loads in the IGListKit tab. First few pages load fine then it asserts. 

Will keep at it :)

Adlai-Holler:
@ay8s Turns out the crash is https://github.com/facebook/AsyncDisplayKit/issues/2855 and not related to IGListKit (phew!). I've worked-around and pushed.

Adlai-Holler:
Hmm it seems like the tests are timing out. Test this please

Adlai-Holler:
@ay8s What are your thoughts? Have you had a chance to give it a spin?

ay8s:
@Adlai-Holler Going to run through it a little more, right now got it populating the collection view but not tried inserting/removing items just yet to see how that performs.

ocrickard:
This is sooo awesome.

jessesquires:
Nice! 💯  🎉

PR #2856, opened 2017-01-03
ASDataController: Correctly Handle Nil IndexPath in nodeForIndexPath: Methods

Adlai-Holler:
This fixes #2843 which @rewandy raised in a more general way.


rewandy:
Awesome! That definitely makes sense to me, and seems to be a more "correct" fix to my original issue.

I didn't have the confidence & inside knowledge to make `ASDataController` do what you did, so thanks very much for that! I guess it was a bug this whole time.

PR #2859, opened 2017-01-04
[UIImage+ASDKAdditions] Add convenient as_imageNamed methods

nguyenhuy:
Originates from `pi_imageNamed`. Credit goes to @garrettmoon and @rcancro. 

cc @appleguy and @rewandy.

### DIFF
--- a/AsyncDisplayKit/UIImage+ASConvenience.m
+++ b/AsyncDisplayKit/UIImage+ASConvenience.m
@@ -122,4 +122,47 @@
   return result;
 }
 
+#pragma mark - as_imageNamed
+
+UIImage *cachedImageNamed(NSString *imageName, UITraitCollection *traitCollection)
+{
+  static NSCache *imageCache = nil;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    // Because NSCache responds to memory warnings, we do not need an explicit limit.
+    // all of these objects contain compressed image data and are relatively small
+    // compared to the backing stores of text and image views.
+    imageCache = [[NSCache alloc] init];
+  });
+  
+  UIImage *image = nil;
+  if ([imageName length] > 0) {
+    NSString *imageKey = imageName;
+    if (traitCollection) {
+      char imageKeyBuffer[256];
+      snprintf(imageKeyBuffer, sizeof(imageKeyBuffer), "%s|%ld|%ld", imageName.UTF8String, (long)traitCollection.horizontalSizeClass, (long)traitCollection.verticalSizeClass);
+      imageKey = [NSString stringWithUTF8String:imageKeyBuffer];
+    }
+    
+    image = [imageCache objectForKey:imageKey];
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/UIImage+ASConvenience.m
```diff
@@ -122,4 +122,47 @@ + (UIImage *)as_resizableRoundedImageWithCornerRadius:(CGFloat)cornerRadius
...
+    imageCache = [[NSCache alloc] init];
+  });
+  
+  UIImage *image = nil;
+  if ([imageName length] > 0) {
+    NSString *imageKey = imageName;
+    if (traitCollection) {
+      char imageKeyBuffer[256];
+      snprintf(imageKeyBuffer, sizeof(imageKeyBuffer), "%s|%ld|%ld", imageName.UTF8String, (long)traitCollection.horizontalSizeClass, (long)traitCollection.verticalSizeClass);
+      imageKey = [NSString stringWithUTF8String:imageKeyBuffer];
```
Adlai-Holler: This is pretty indirect. Wouldn't this work?

`NSString *imageKey = [NSString stringWithFormat:@"%@|%ld|%ld", imageName, (long)traitCollection.horizontalSizeClass, (long)traitCollection.verticalSizeClass];`

garrettmoon: @Adlai-Holler This was put in specifically because it was targeted as a performance issue.

Adlai-Holler: Hmm … OK.

Adlai-Holler:
Merging!

nguyenhuy:
Cool, thanks @Adlai-Holler, @garrettmoon!

hovox:
As I know UIImage's `imageNamed` method already has caching implemented in it ?
Do we have any example projects or tests presenting this solution is faster then just using `imageNamed ` ?

garrettmoon:
@hovox no but it'd be great if you would be willing to write some! If I remember correctly, the reasoning for this being faster than the default implementations has to do with trait collections. Essentially using the method that doesn't specify a trait collection or using the one that does and not passing in nil is quite slow. So this caching layer protects agains that. But I could be wrong, it's been a while :)

hovox:
@garrettmoon Thanks for explanation.

jesseluo:
Hi, is this method thread safe in iOS 8+?
As method document this is thread safe, but imageNamed: is not thread safe in iOS 8. So I wonder if I have some misunderstanding here.

garrettmoon:
@jesseluo I *believe* it's thread safe if you don't provide a trait collection.

jesseluo:
@garrettmoon But in fact it's not. I used this method and caused about 2~3% crash rate on iOS 8
image

maicki:
@jesseluo Hey @jesseluo, can you please open an issue and if possible provide some stack traces? Thanks!

Issue #2841, opened 2016-12-28
custom Layout Transition doesn't change node size


hashemp206:
I have created a custom `ASDisplayNode` wich  displays some tags and their descriptions. for layout transition I have used Layout Transition API. using version `2.0-beta.1`
when user selects a tag, selected tag grows and it displays tags description.

![taglistold](https://cloud.githubusercontent.com/assets/3206748/21520812/babe6960-cd0c-11e6-845c-7c33b70c847e.gif)

now I have updated ASyncDisplayKit to `2.0.1` but resulting behavior is corrupted: (for simplicity I have hidden to description text node)

<img width="479" alt="screen shot 2016-12-28 at 3 31 23 pm" src="https://cloud.githubusercontent.com/assets/3206748/21521511/daec14de-cd12-11e6-9bdd-04ba1cefb845.png">

![taglistanimated](https://cloud.githubusercontent.com/assets/3206748/21520837/e95efe2e-cd0c-11e6-9e83-56244ee79ee8.gif)
node hierarchy: AScellNode -> TagListNode -> TagNode
<img width="619" alt="screen shot 2016-12-28 at 3 18 14 pm" src="https://cloud.githubusercontent.com/assets/3206748/21521289/e85f7e6e-cd10-11e6-9e70-1de1c6b4b9af.png">

for the below one by selecting a tag, I just call `setNeedsLayout()` on tags container node and it behaves correctly with no animation:
![taglist](https://cloud.githubusercontent.com/assets/3206748/21520935/9b7c5e12-cd0d-11e6-8eba-b0c674cac429.gif)

TagNode Layout Spec:

``` swift
override func layoutSpecThatFits(_ constrainedSize: ASSizeRange) -> ASLayoutSpec {  
        var children: [ASLayoutElement] = [imageNode]
        if open {
            descriptionNode.isHidden = false
            children.insert(descriptionNode, at: 0)
        } else{
            descriptionNode.isHidden = true
        }
        
        descriptionNode.style.spacingAfter = inset
        let horizontalStack = ASStackLayoutSpec(direction: .horizontal, spacing: 0, justifyContent: .center, alignItems: .center, children: children)
        
        let insetNode = ASInsetLayoutSpec(insets: UIEdgeInsetsMake(0, inset, 0, inset), child: horizontalStack)
        return insetNode;
    }
```

by tapping a tag,  tagNode `onStateChangeBlock` gets called in TagListNode:

``` swift
[...]

Issue #2816, opened 2016-12-21
[Layout Transition API] V2.0.1 breaks the layout transition API in certain cases


master-nevi:
I was able to modify the sample project [ASDKgram](https://github.com/facebook/AsyncDisplayKit/tree/master/examples/ASDKgram) to reproduce the layout transition API bug.
Branch: https://github.com/master-nevi/AsyncDisplayKit/commits/layout-api-broken
Commit: https://github.com/master-nevi/AsyncDisplayKit/commit/248de21bbbe3a8c9dc16ca77a7d85704f4a87cd6

**Steps to repro:**
1) Tap on the comments node. This should trigger a layout transition similar to what's in my project.
- Expect a red node to animate above the comments and push the comments down.
- What actually happens is the comments move down, but the red node is never put into the view hierarchy

Now if you revert the placement of `[self _finishOrCancelTransition];` like it was in v2.0 it will work again. It appears that by delaying the call to `[self _finishOrCancelTransition];`, it thwarts one of the layout passes which places the red node.

**v2.0.1**
![screen shot 2016-12-20 at 10 55 55 pm](https://cloud.githubusercontent.com/assets/987706/21380096/8a414c2c-c707-11e6-930a-e5b5050e5301.png)

**v2.0.0**
![screen shot 2016-12-20 at 10 57 47 pm](https://cloud.githubusercontent.com/assets/987706/21380136/c0f36250-c707-11e6-8593-8ea6650c1ba8.png)

PR #2862, opened 2017-01-04
[WIP][ASLayoutTransition] Relayout a node if one of its subnodes has a new size after layout transition

nguyenhuy:
- Revert 7872cfb because a transition isn't complete until its CAAnimation finishes. If `_transitionInProgress` flag is turned off before that, any layout pass occurs afterward, but before `-transitionContext:didComplete:` is called, can erase all pending states of the transition, causing old subviews to remain in the hierarchy forever.
- If a layout transition causes the node to have a different size, notify its supernode to resize. Currently we're already doing this with ASCellNode via measurement completion block, but not with normal ASDisplayNode. ASScrollNode in particular isn't notified and thus doesn't update its content size.

@maicki, @garrettmoon, @appleguy, @Adlai-Holler: Would love to have your feedbacks here.

@hannahmbanana Please help me to test this diff with Pinterest master. I already did that, as well as with ASDKLayoutTransition example and the sample projects in #2841 and #2816. However, with all the branch and context switchings, I may overlooked something. Gonna test it more tomorrow!

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1072,9 +1072,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
       
       // Kick off animating the layout transition
       [self animateLayoutTransition:_pendingLayoutTransitionContext];
-      
-      // Mark transaction as finished
-      [self _finishOrCancelTransition];
     });
   };
   
@@ -1298,6 +1295,9 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   _pendingLayoutTransitionContext = nil;
 
   [self _pendingLayoutTransitionDidComplete];
+    
+  // Mark transaction as finished
+  [self _finishOrCancelTransition];
 }
 
 #pragma mark - Layout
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1030,10 +1041,10 @@ - (void)transitionLayoutWithSizeRange:(ASSizeRange)constrainedSize
...
-      // Grab __instanceLock__ here to make sure this transition isn't invalidated
-      // right after it passed the validation test and before it proceeds
-      ASDN::MutexLocker l(__instanceLock__);
-
+      // Hold onto __instanceLock__ until at least after the animation of this transtion kicked off,
+      // to make sure this transition isn't invalidated after it passed the abort test below
+      __instanceLock__.lock();
+      
       if ([self _shouldAbortTransitionWithID:transitionID]) {
         return;
```
garrettmoon: Should this unlock here before the return?

nguyenhuy: Good catch! It's very hard to hit this line but it's a legitimate mistake of mine!

hannahmbanana:
I tested this out on my product code and it doesn't work in my case. Huy has the repro steps and will double check to see if it's a bug on my side or ASDK's side.

AsyncDisplayKit/ASDisplayNode.mm
nguyenhuy: Fixed.

nguyenhuy:
Closing this in favor of an overhaul of the layout system.

Issue #698, opened 2015-10-02
[ASDataController] "Move" Operations Animate Incorrectly (Delete + Insert)


Adlai-Holler:
This is a minor issue but I want to get it on the board while we're thinking through table view updates in #684 / #693. It results from us splitting moves into delete+insert in ASDataController.

I believe this issue is important because it highlights a really common architectural problem when dealing with table views: interpreting changes merely as array updates and discarding change metadata. For example, interpreting `move` as `delete,insert` or interpreting a `reload section` as `remove all items, add all items` which ignores section metadata like header views.

Since our updates are already required to be on the main thread, maybe we should split each intercepted call into two parts: the intact update, which we forward to the table view, and the array update where we interpret the update as above (e.g. `move -> delete + insert`).

Naturally all this applies to collection views as well.

appleguy:
@nguyenhuy @levi @eanagel have any of you thought about how we might fix this, or encountered it in your apps?  I'm not certain how high priority this is compared to the many other improvements / refinements we might implement, especially since ASDataController is now substantially improved and appears to be mostly stable (no crashes in Pinterest after a week of heavy public use).  Still though, it would be really great to ensure ASDK remains super animation-friendly.

smyrgl:
@appleguy @Adlai-Holler  I was about to open a new issue for this but I found this while searching so I decided to drop my own take on this here.  

So here is the problem as I see it with this behavior: because ADK treats moves as delete->insert that means that there is no way to tell in the context of a custom UICollectionViewLayout that this is a move.  When I perform a move op `prepareForCollectionViewUpdates:` is called on my UICollectionViewLayout with an update array with a pair of updates with the action set to .Delete and .Insert.  That means that all context is lost.  So lets say I wanted to do something like this:

```Swift
    override func initialLayoutAttributesForAppearingItemAtIndexPath(itemIndexPath: NSIndexPath) -> UICollectionViewLayoutAttributes? {
        if let attributes = layoutAttributesForItemAtIndexPath(itemIndexPath)?.copy() as? UICollectionViewLayoutAttributes {
            let returnAttrs = attributes
            var returnFrame = attributes.frame
            returnFrame.origin.y -= returnFrame.height
            returnAttrs.frame = returnFrame
            returnAttrs.alpha = 0.1
            return returnAttrs
        }
        return super.initialLayoutAttributesForAppearingItemAtIndexPath(itemIndexPath)
    }
    
[...]

PR #2863, opened 2017-01-04
Remove Unused Implementations of moveItem: and moveSection:

Adlai-Holler:
Since moves are composite operations (like `reloadItems` and `reloadSections`), they must be handled by `ASChangeSetDataController` and they _are_. The `ASDataController`-level implementations are completely unused and don't fit into the current architectural model for the two classes. In preparation for #698, which we'll hopefully implement soon, we should remove this code to reduce confusion and complexity.


Adlai-Holler:
Since this doesn't affect behavior at all, I'll land once CI passes.

harryworld:
I was going to ask whether these methods are deprecated. Thanks for the PR to clarify.

PR #2864, opened 2017-01-04
[ASNetworkImageNode] Don't lock while calling downloader

garrettmoon:
Addresses #2785
To avoid performance issues, we should avoid locking the downloader.
To achieve this we need to do some kinda gross things. Essentially
the cost is the code is more complex and potentially far less performant
in edge cases. In testing, edge cases are nearly never hit, but I'm not
sure how good I feel about the cost in code complexity. This exacerbates
the locking issues in ASNetworkImageNode:

1. There is no convention for which methods lock.
2. There's no indication which vars are only set on init and therefore
safe to access except in the class extension definition.

### DIFF
--- a/AsyncDisplayKit/ASNetworkImageNode.mm
+++ b/AsyncDisplayKit/ASNetworkImageNode.mm
@@ -326,8 +326,8 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
     }
   }
   
-  if (_downloadIdentifier != nil) {
-    [_downloader setPriority:ASImageDownloaderPriorityPreload withDownloadIdentifier:_downloadIdentifier];
+  if (downloadIdentifier != nil) {
+    [_downloader setPriority:ASImageDownloaderPriorityPreload withDownloadIdentifier:downloadIdentifier];
   }
   
   [self _updateProgressImageBlockOnDownloaderIfNeeded];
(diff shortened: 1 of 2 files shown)


garrettmoon:
Putting this up for discussion. If it seems like the direction we want to go in I'll attempt to write a couple unit tests to exercise the failure cases (when the URL changes in the download case or download identifier changes in the update progress block case).

AsyncDisplayKit/ASNetworkImageNode.mm
```diff
@@ -293,28 +295,41 @@ - (void)displayWillStartAsynchronously:(BOOL)asynchronously
...
-    if (_downloadIdentifier != nil) {
-      [_downloader setPriority:ASImageDownloaderPriorityPreload withDownloadIdentifier:_downloadIdentifier];
+  {
+    ASDN::MutexLocker l(__instanceLock__);
+    if (_downloaderFlags.downloaderImplementsSetPriority) {
+      downloadIdentifier = _downloadIdentifier;
     }
   }
   
+  if (_downloadIdentifier != nil) {
```
Adlai-Holler: Hmm it looks like we're accessing this ivar without the lock held

garrettmoon: Good catch! This was supposed to be downloadIdentifier, not _downloadIdentifier.

Adlai-Holler:
Here we go!

Issue #2845, opened 2016-12-30
[ASCollectionNode] Move Items Animation


leotumwattana:
We've been rebuilding parts of our app as a feasibility test for using ASDK. So far ASDK has been awesome and we've resolved most of the issues except for one thing. The `moveItem:` animation for ASCollectionNode.

I've done a little reading through past issues (most notable #698) and noticed that right now, `moveItem:` is done as a `delete + insert`. However, in our app, the `moveItem:` animation is actually quite critical to help inform our users to changes. A simple fade-out-fade-in animation provides a lot less visual cue to the eyes. 

I understand #698 is marked for 2.1. Anxiously waiting to see this enhancement so we can start using ASDK in our app. Thanks very much in advance!

garrettmoon:
@leotumwattana it sounds like this is a duplicate of #698 so I'm going to close this.

garrettmoon:
@leotumwattana sorry if that sounded callous, we really appreciate the enthusiasm! However having one issue makes it easier to track and get stuff done :)

leotumwattana:
@garrettmoon: No worries. Let's go with #698.

PR #2868, opened 2017-01-05
[ASDataController] Move animates using UIKit native behaviour

harryworld:
Hi @Adlai-Holler ,

I would like to seek your comments on this PR, regarding `moveItem:` animation concern specified in #698 #2845 

As my colleague @leotumwattana mentioned in #2845 , we are seeking a way to use UICollectionView native `move` behaviour, which provides a visual cue to users that the item is moved from one position to another.

I developed the work only on `ASCollectionView` and `moveItem:` by far. As I am new to the framework (2 days so far), I would like to seek your advices on whether I am heading the right direction in implementing such enhancement.

Addresses: #698 
I'm also including a sample project [here](https://github.com/harryworld/ASDKMoveItems) to demonstrate the updated behaviour.

Thanks so much.

- [ ] Support ASCollectionNode
  - [ ] Section
  - [x] Row
- [ ] Support ASTableNode
  - [ ] Section
  - [ ] Row
- [ ] Decouple / Replicate _ASHierarchyItemChange for `from` and `to` indexPaths
- [ ] Documentation
- [ ] Tests

@appleguy @garrettmoon @nguyenhuy @levi

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -448,7 +448,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   _flags.isDeallocating = YES;
 
   // Synchronous nodes may not be able to call the hierarchy notifications, so only enforce for regular nodes.
-  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating. Node: %@", self);
+  // FIXME: move UICollectionViewCell out of visible screen will trigger dealloc
+//  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating. Node: %@", self);
   
   self.asyncLayer.asyncDelegate = nil;
   _view.asyncdisplaykit_node = nil;
--- a/AsyncDisplayKit/Details/ASChangeSetDataController.mm
+++ b/AsyncDisplayKit/Details/ASChangeSetDataController.mm
@@ -88,7 +88,7 @@
     [super beginUpdates];
     
     for (_ASHierarchyMoveItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
-      [super moveRowAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath withAnimationOptions:change.animationOptions];
+      [super moveNodeFromIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath withAnimationOptions:change.animationOptions];
     }
     
     for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeDelete]) {
--- a/AsyncDisplayKit/Details/ASChangeSetDataController.mm
+++ b/AsyncDisplayKit/Details/ASChangeSetDataController.mm
@@ -81,7 +81,15 @@
     
     ASDataControllerLogEvent(self, @"triggeredUpdate: %@", _changeSet);
     
+    // TODO: move
+    [super prepareMoveItemChanges:[_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]];
+    
+    // Prepare UIKit batch updates
     [super beginUpdates];
+    
+    for (_ASHierarchyMoveItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
+      [super moveRowAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath withAnimationOptions:change.animationOptions];
+    }
     
     for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeDelete]) {
       [super deleteRowsAtIndexPaths:change.indexPaths withAnimationOptions:change.animationOptions];
@@ -99,10 +107,6 @@
       [super insertRowsAtIndexPaths:change.indexPaths withAnimationOptions:change.animationOptions];
     }
     
-    for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
-      [super moveRowAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath withAnimationOptions:change.animationOptions];
-    }
-
 #if ASEVENTLOG_ENABLE
     NSString *changeSetDescription = ASObjectDescriptionMakeTiny(_changeSet);
     batchCompletion = ^(BOOL finished) {
@@ -205,6 +209,8 @@
   ASDisplayNodeAssertMainThread();
   [self beginUpdates];
   [_changeSet moveItemFromIndexPath:indexPath toIndexPath:newIndexPath animationOptions:animationOptions];
+  [_changeSet deleteItems:@[indexPath] animationOptions:animationOptions];
+  [_changeSet insertItems:@[newIndexPath] animationOptions:animationOptions];
   [self endUpdates];
 }
 
--- a/AsyncDisplayKit/Details/ASChangeSetDataController.mm
+++ b/AsyncDisplayKit/Details/ASChangeSetDataController.mm
@@ -98,6 +98,10 @@
     for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeInsert]) {
       [super insertRowsAtIndexPaths:change.indexPaths withAnimationOptions:change.animationOptions];
     }
+    
+    for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
+      [super moveRowAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath withAnimationOptions:change.animationOptions];
+    }
 
 #if ASEVENTLOG_ENABLE
     NSString *changeSetDescription = ASObjectDescriptionMakeTiny(_changeSet);
@@ -200,8 +204,7 @@
 {
   ASDisplayNodeAssertMainThread();
   [self beginUpdates];
-  [_changeSet deleteItems:@[indexPath] animationOptions:animationOptions];
-  [_changeSet insertItems:@[newIndexPath] animationOptions:animationOptions];
+  [_changeSet moveItemFromIndexPath:indexPath toIndexPath:newIndexPath animationOptions:animationOptions];
   [self endUpdates];
 }
 
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -263,10 +263,22 @@ NSString * const ASCollectionInvalidUpdateException = @"ASCollectionInvalidUpdat
   // Deep copy is critical here, or future edits to the sub-arrays will pollute state between _editing and _complete on different threads.
   NSMutableArray *completedNodes = ASTwoDimensionalArrayDeepMutableCopy(editingNodes);
   
+  // Filter out insert indexPaths in Edit Queue
+  NSMutableArray *newNodes = [[NSMutableArray alloc] initWithCapacity:nodes.count];
+  NSMutableArray *newIndexPaths = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
+  
+  for (int i = 0; i < indexPaths.count; i++) {
+    if (![_moveToIndexPaths containsObject:indexPaths[i]]) {
+      [newNodes addObject:nodes[i]];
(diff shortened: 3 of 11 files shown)


AsyncDisplayKit/Details/ASChangeSetDataController.mm
```diff
@@ -98,6 +98,10 @@ - (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
     for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeInsert]) {
       [super insertRowsAtIndexPaths:change.indexPaths withAnimationOptions:change.animationOptions];
     }
+    
+    for (_ASHierarchyItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
+      [super moveRowAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath withAnimationOptions:change.animationOptions];
+    }
```
Adlai-Holler: The tricky bit here is, we need to grab the "move from" objects before we do any deleting/inserting, then delete them along with any other delete operations, then re-insert them along with any other insert operations.

According to the [rules of batch updating](https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/TableView_iPhone/ManageInsertDeleteRow/ManageInsertDeleteRow.html#//apple_ref/doc/uid/TP40007451-CH10-SW17), we interpret delete's and move-from's as "before the update" indexes and move-to's as "after the update" indexes.

For example, imagine we start with [A, B, C] and we receive changeset [ delete(0), move(1, 1) ]. The correct way to process this is:

```
// Gather moved items
id movedItem = items[1]; // [A, B, C]
// Deletes & move-froms, in descending order.
items.delete(1) // [ A, C ]
items.delete(0) // [ C ]
// Inserts & move-tos, in ascending order.
items.insert(1, movedItem) // [ C, B ]
```

If we process it like the code above, we get:

```
items.delete(0) // [ B, C ]
items.move(1, 1) // [ B, C ]
```

It's tricky! So ASDataController should not implement `moveItem:` but instead should support `insertItemWithNode:` or something similar I think.

Adlai-Holler:
Excellent start! This is coming together fast, and I really appreciate it!

AsyncDisplayKit/Details/ASChangeSetDataController.mm
Adlai-Holler: Here's some sample code demonstrating it using UICollectionView. Ignore the project name 🙂
[MoveExample.zip](https://github.com/facebook/AsyncDisplayKit/files/688555/MoveExample.zip)

AsyncDisplayKit/Details/ASChangeSetDataController.mm
```diff
@@ -205,6 +209,8 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
   ASDisplayNodeAssertMainThread();
   [self beginUpdates];
   [_changeSet moveItemFromIndexPath:indexPath toIndexPath:newIndexPath animationOptions:animationOptions];
+  [_changeSet deleteItems:@[indexPath] animationOptions:animationOptions];
+  [_changeSet insertItems:@[newIndexPath] animationOptions:animationOptions];
   [self endUpdates];
```
harryworld: I made use of the original idea of split here, reusing the existing `delete` + `insert` logic.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -361,8 +360,19 @@ - (void)_insertNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAni
...
+    NSMutableArray *newIndexPaths = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
+    
+    for (int i = 0; i < indexPaths.count; i++) {
+      if (![_moveToIndexPaths containsObject:indexPaths[i]]) {
+        [newNodes addObject:nodes[i]];
+        [newIndexPaths addObject:indexPaths[i]];
+      }
+    }
+    
     if (_delegateDidInsertNodes)
```
harryworld: Very ugly here. It is trying to filter out the Move `from` (delete) and `to` (insert), and send them to UICollectionView move animation call.

Advices please.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -448,7 +448,8 @@ - (void)dealloc
   _flags.isDeallocating = YES;
 
   // Synchronous nodes may not be able to call the hierarchy notifications, so only enforce for regular nodes.
-  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating. Node: %@", self);
+  // FIXME: move UICollectionViewCell out of visible screen will trigger dealloc
+//  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating. Node: %@", self);
```
harryworld: There is an issue with `Display` state upon moving a cell out of window. The system deallocates before the state was cleaned up. Please advice.

AsyncDisplayKit/Details/ASChangeSetDataController.mm
```diff
@@ -81,8 +81,16 @@ - (void)endUpdatesAnimated:(BOOL)animated completion:(void (^)(BOOL))completion
...
+    // TODO: move
+    [super prepareMoveItemChanges:[_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]];
+    
+    // Prepare UIKit batch updates
     [super beginUpdates];
     
+    for (_ASHierarchyMoveItemChange *change in [_changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
+      [super moveRowAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath withAnimationOptions:change.animationOptions];
+    }
+
```
Adlai-Holler: Hmm this still won't work, I believe. For the reasons in [this comment](https://github.com/facebook/AsyncDisplayKit/pull/2868#discussion_r94865245), `ASDataController` simply _cannot implement_ `moveRowAtIndexPath:` because the move has to be decomposed into delete and insert, which have to be performed separately, each with all the other operations of the same kind.

I believe what we should do instead is modify ASDataController to have a method like `insertRowsAtIndexPaths:withExistingNodes:…` that takes a `NSDictionary<NSIndexPath, ASIndexedNodeContext>`. Before the update, we build that dictionary by going through all the `_ASHierarchyMoveItemChange` and using the `from` index path to fetch the node context and the `to` index path to act as the key into the dictionary.

Then inside `insertRows`, the data controller will always check the provided dictionary to see if there's already a context for the given index path, and use that if possible.

* Note we could use ASCellNodes directly, or contexts, whichever is more convenient.

Does that make sense?

harryworld: It sounds more like a rename to the method, and I understand there might be ambiguity. I will change it, but some clarifications go on.

Yet, since I'm using the existing logic to insert the `move-to` nodes with the new node block at the toIndexPaths of the updated data source, there is no need to store them in a dictionary and reuse.

In fact, I think as move is decomposed to `delete` and `insert`, they shall be processed together with other deletes and inserts in desc and asc ways respectively. If we isolate move-related methods like the one you suggested, it might break the rules of batching.

What do you think?

Adlai-Holler: I agree, they should be processed along with other inserts and deletes. We will, however, have to _not_ tell UIKit about those inserts/deletes and instead tell UIKit about the move. That may require a somewhat radical change in how we inform UIKit about updates, which I'm open to.

harryworld: Agree, that's what my current implementation does. Let me first move away from the ambiguous `moveRowAtIndexPath:` method, and tidy up a bit and ask for your review again.

harryworld:
After testing with our projects, the mechanism of simply re-creating nodes would not work, because the moved view is not able to link to newly created node.

Instead, I changed the logic to move the item in `nodeContext` and `completedNodes` arrays.

Besides, I am renaming the method.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -62,8 +63,13 @@ @interface ASDataController () {
 
   BOOL _delegateDidInsertNodes;
   BOOL _delegateDidDeleteNodes;
+  BOOL _delegateDidMoveNode;
   BOOL _delegateDidInsertSections;
   BOOL _delegateDidDeleteSections;
+  
+  NSMutableArray *_moveFromIndexPaths;
+  NSMutableArray *_moveToIndexPaths;
+  NSMutableDictionary *_moveToNodeContextsDict;
```
Adlai-Holler: Let's add type annotations and documentation here. What are these used for, what queue are the modified on, etc.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -747,13 +852,21 @@ - (void)insertRowsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDat
...
   
   for (NSIndexPath *indexPath in sortedIndexPaths) {
-    ASCellNodeBlock nodeBlock = [_dataSource dataController:self nodeBlockAtIndexPath:indexPath];
-    ASSizeRange constrainedSize = [self constrainedSizeForNodeOfKind:ASDataControllerRowNodeKind atIndexPath:indexPath];
-    [contexts addObject:[[ASIndexedNodeContext alloc] initWithNodeBlock:nodeBlock
-                                                              indexPath:indexPath
-                                               supplementaryElementKind:nil
-                                                        constrainedSize:constrainedSize
-                                                            environment:environment]];
+    if ([_moveToIndexPaths containsObject:indexPath]) {
```
Adlai-Holler: This'll be `N^2` so let's replace by:

```objc
if (ASIndexedNodeContext *context = _moveToNodeContextsDict[indexPath])
```

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -284,6 +303,60 @@ - (void)deleteNodesOfKind:(NSString *)kind atIndexPaths:(NSArray *)indexPaths co
...
+    _moveToNodeContextsDict = [NSMutableDictionary dictionaryWithCapacity:items.count];
+    
+    for (_ASHierarchyMoveItemChange *change in items) {
+      [_moveFromIndexPaths addObject:change.fromIndexPath];
+      [_moveToIndexPaths addObject:change.toIndexPath];
+      [_moveIndexPathPairs setObject:change.toIndexPath forKey:change.fromIndexPath];
+    }
+    
+    for (int i = 0; i < items.count; i++) {
+      NSIndexPath *toIndexPath = [_moveIndexPathPairs objectForKey:_moveFromIndexPaths[i]];
```
Adlai-Holler: We can combine this with the for-loop above, I believe, and reduce duplicate accesses.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -357,8 +430,35 @@ - (void)_deleteNodesAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASD
   [self deleteNodesOfKind:ASDataControllerRowNodeKind atIndexPaths:indexPaths completion:^(NSArray *nodes, NSArray *indexPaths) {
     ASDisplayNodeAssertMainThread();
     
+    // TODO - Filter out Move indexPaths
+    NSMutableArray *newNodes = [[NSMutableArray alloc] initWithCapacity:nodes.count];
+    NSMutableArray *newIndexPaths = [[NSMutableArray alloc] initWithCapacity:indexPaths.count];
+    
+    for (int i = 0; i < indexPaths.count; i++) {
+      if (![_moveFromIndexPaths containsObject:indexPaths[i]]) {
```
Adlai-Holler: `_moveFromIndexPaths` is modified on the editing queue, but it's being read here from the main queue. Would it make sense to put these data structures inside the `_ASHierarchyChangeSet` and then pass that object into these methods as a `context` object? 

That would cut down on the state inside the data controller.

harryworld: Agree with moving to `_ASHierarchyChangeSet` as the primary place handling changing data, and I want to clarify couple points before the move.

1. Should the `move` data set be consolidated on the editing queue? As it seems in `reload` case, it was split into `insert` and `delete` on the main queue.
2. Would it make sense to override some methods in `ASChangeSetDataController` to handle the `move` behaviour? (For instance, `insertNodes:` and `deleteNodesOfKind:`)
3. Instead of passing `context`, these methods can access `_changeSet` directly inside `ASChangeSetDataController`

Adlai-Holler:
@harryworld Since your changes are dependent on @nguyenhuy 's in-progress data controller changes, and so the final outcome of native moves will probably come from a separate PR, I'm closing this for the time being. If I've got something wrong, feel free to reopen at your discretion!

PR #2883, opened 2017-01-10
Account for Cell Separators when Reporting Table View Row Height

Adlai-Holler:
UITableView actually wants you to include the separator height in `heightForRowAtIndexPath:`. Which is terrible! But if you want a content view of height `44` you need to return `44.5` on a 2x screen if you have separators turned on.


Adlai-Holler:
@garrettmoon @rcancro

PR #2886, opened 2017-01-11
[Nitpick] Update CocoaPod badges to be dynamic.

aaronschubert0:
This updates the values automatically, which is perfect to reflect the growth of ASDK. I've wanted to do this for ages but the shield.io API didn't have the support up till now. We lose some granularity but I think this is worth it.

Maybe @hannahmbanana or @nguyenhuy could take a quick look. Sorry it's been a while since I've kept up to date with the project, not sure if @appleguy is still directly involved.


Adlai-Holler:
Huy you raised a good point about the badges being links. Without derailing that, I'm going to land this because it's so _good_!

aaronschubert0:
Awesome thanks guys! @Adlai-Holler @nguyenhuy

hannahmbanana:
@aaronschubert0 - this is super awesome! thank you!

PR #2889, opened 2017-01-11
Use Native convertRect:, convertPoint: Methods to Handle Nil Cases

### LINKED ISSUE #2888: [ASDisplayNode] convertPoint:toNode: doesn't go to UIWindow when passed nil

Adlai-Holler:
Resolves #2888 @hannahmbanana 

I didn't know this, but `convertPoint:toLayer:nil` behaves a lot differently from `convertPoint:toView:nil` and so it's important that we call the appropriate method for the appropriate case.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1806,10 +1806,18 @@ static inline CATransform3D _calculateTransformFromReferenceToTarget(ASDisplayNo
 {
   ASDisplayNodeAssertThreadAffinity(self);
   
+  /**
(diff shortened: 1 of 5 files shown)


garrettmoon:
I'm not positive this is an issue, but CALayer and UIView document different behavior when the view / layer is nil.

CALayer:
"If you specify nil for the layer parameter, this method returns the original point subtracted from the layer's frame's origin."

UIView:
"If view is nil, this method instead converts to window base coordinates."

Adlai-Holler:
@garrettmoon It's true the behavior changes, but I think it's reasonable for us to do the layer-thing for layer-backed nodes, and the view-thing for view-backed nodes.

PR #2890, opened 2017-01-12
Need an autorelease pool!

garrettmoon:


Adlai-Holler:
Hey so I thought that all threads spawn with an autorelease pool already? What's the deal here?

garrettmoon:
Ok, I think this is wrong. I think the issue is that threadMain never exits and therefore the autorelease pool created in it never drains.

garrettmoon:
Ok, apparently threadMain doesn't exit by design so I think this is the correct solution :)

Issue #2837, opened 2016-12-25
[Table / Collection] Add BOOL option to flip everything for "upward-loading" apps.


appleguy:
This has become one of the most commonly-asked questions on Slack.  In addition to the benefit from having a simple answer / documentation callout for the feature, the complexity is low enough to provide this kind of BOOL.  The use case common enough — all comment / messaging views — that it makes sense to do if we can implement it with ~20 lines of code or less.

All this option would do:
- Provide BOOL on Table or Collection
- When enabled, it flips 180º every ASCellNode that is returned (by applying a transform to it.  It also sets its own transform to flip 180º.
- That's it!  It would be reasonable to assert / not allow changing the property after content has started to load (or even, when the node's view exists).  The 90% value is captured in this case.  Any more dynamic cases should be handled by apps directly.

> Hello, I’m just new to this trying explore ASDK. I have some questions about `head fetching` which got closed - https://github.com/facebook/AsyncDisplayKit/pull/1141
> 
> Is ASDK currently doesn’t support head fetching out of the box?
> If we want that functionality is there any suggestion on how to build it on top of ASDK? (edited)
> 
> appleguy [9:46 AM]  
> @sangbill what kind of UI were you thinking about?  Pull to refresh, or automatic?
> 
> sainttail [9:46 AM]  
> automatic like chat
> 
> appleguy [9:47 AM]  
> well, the first fetch will be automatic if you just use batch fetching
> 
> [9:47]  
> beyond that though, you'd need some kind of polling system or a server push system (persistent connection), which you have to manage yourself at the network layer in your app
> 
> [9:47]  
> but it's always nice to rely on batch fetching for both scrolling back in history (Pagination), as well as the first load of a screen as you're arriving at it
> 
> [9:48]  
> so it's only updating with new content while the user is currently on a screen that you need to handle manually
> 
> sainttail [9:49 AM]  
> doesn’t batch fetching support only one direction right now. which is scrolling from top to bottom. For bottom to top (load history) it isn’t support yet?
> 
> appleguy [9:51 AM]  
[...]

PR #2891, opened 2017-01-12
[ASTableNode/ASCollectionNode] Inversion

aaronschubert0:
Implements #2837 on ASTableNode and ASCollectionNode. I've tried to keep the API minimal. Also added an example app, since the user will have to adjust the contentInset of their ASTable/CollectionNode for the inversion to work properly. 

Documentation: https://github.com/AsyncDisplayKit/Documentation/pull/78

@appleguy


hannahmbanana:
Thank you for including a sample project and documentation @aaronschubert0!

AsyncDisplayKit/ASCollectionNode.mm
```diff
@@ -217,6 +218,12 @@ - (_ASCollectionPendingState *)pendingState
   return _pendingState;
 }
 
+- (void)setInverted:(BOOL)inverted
+{
+  self.view.dataController.inverted = inverted;
+  self.transform = inverted ? CATransform3DMakeRotation(M_PI, 0, 0, 1.0) : CATransform3DIdentity;
```
Adlai-Holler: Let's use `CATransform3DMakeScale(1, -1, 1)` instead of rotation. It's cleaner because π can't be represented perfectly.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -912,7 +912,9 @@ - (ASCellNode *)nodeAtIndexPath:(NSIndexPath *)indexPath
       context = completedNodesSection[row];
     }
   }
-  
+  if (_inverted) {
+    context.node.transform = CATransform3DMakeRotation(M_PI, 0, 0, 1.0);
```
Adlai-Holler: It might be better to put this code in `-[ASCollectionView/ASTableView dataController:nodeBlockAtIndexPath:]` near the bottom along with the other customization that gets applied to the node, so that it's run exactly once per node, so the data controller doesn't have to be aware of the inversion.

Adlai-Holler:
Jenkins, test this please

hannahmbanana:
Build error:

```
AsyncDisplayKit/ASCollectionNode.mm:221:1: ivar '_inverted' which backs the property is not referenced in this property's accessor [-Werror,-Wunused-property-ivar]
11:22:06 
11:22:06 - (void)setInverted:(BOOL)inverted
```

AsyncDisplayKit/Details/ASDataController.mm
aaronschubert0: Yeah that makes sense, I had opted to avoid this initially so that I didn't have to duplicate code, but for the sake of two extra lines of code, it would feel nice not to have the ASDataController be involved.

AsyncDisplayKit/ASCollectionNode.mm
aaronschubert0: Awesome, will change it to that, thanks.

aaronschubert0:
@Adlai-Holler Made those changes!

hovox:
+1 For this.

davitPiloyan93:
+1 For this.

NikoSahradyan:
+1 For this. good job

Adlai-Holler:
Jenkins, add to whitelist

AsyncDisplayKit/ASCollectionNode.mm
```diff
@@ -217,6 +218,12 @@ - (_ASCollectionPendingState *)pendingState
   return _pendingState;
 }
 
+- (void)setInverted:(BOOL)inverted
+{
+  self.view.inverted = inverted;
+  self.transform = inverted ? CATransform3DMakeScale(1, -1, 1)  : CATransform3DIdentity;
```
Adlai-Holler: We need to implement this flag the same way we implement `delegate` – pass to the pending state or to the view, and that'll make the unused `_inverted` ivar disappear.

hasmikvardanyan:
+1

aaronschubert0:
@Adlai-Holler Added the logic to the pending state instead, should be good to go now.

AsyncDisplayKit/ASCollectionNode.mm
```diff
@@ -217,6 +220,17 @@ - (_ASCollectionPendingState *)pendingState
...
+- (void)setInverted:(BOOL)inverted
+{
+  self.transform = inverted ? CATransform3DMakeScale(1, -1, 1)  : CATransform3DIdentity;
+  if ([self pendingState]) {
+    _pendingState.inverted = inverted;
+  } else {
+    ASDisplayNodeAssert([self isNodeLoaded], @"ASCollectionNode should be loaded if pendingState doesn't exist");
+    self.view.inverted = inverted;
+  }
+}
```
Adlai-Holler: You'll need to override the accessor `- (BOOL)inverted` as well (see `-delegate`). The idea is to get rid of the synthesized `_inverted` ivar, because it's not used and that's making the compiler 💣

aaronschubert0: Gotcha, I've added the accessor

Adlai-Holler:
Awesome, thanks! Here we go folks!

PR #2892, opened 2017-01-12
[ASStackLayoutSpec] Refactor baseline alignment algorithm

### LINKED ISSUE #1604: [ASStackLayoutSpec] Baseline Alignment + Space Between Justification Does Not Space Correctly
When using a stack layout spec, when `.BaselineFirst` alignment is used in combination with `.SpaceBetween` justification, the spacing isn't correct – the justification is instead like `.Start`. 

Here's a reduced case: https://github.com/Adlai-Holler/ASDKBaselineSpaceBetweenIssue

| .SpaceBetween + .Start | .SpaceBetween + .BaselineFirst |
| --- | --- |
| img_1537 | img_1536 |

nguyenhuy:
The current baseline alignment algorithm is implemented as an extra step that is executed after our main stack algorithm finished. While it has served us well for a long time, this implementation has a few problems that should be addressed:
- It diverges from CSS Flexbox algorithm ([here](https://www.w3.org/TR/css-flexbox-1/#layout-algorithm)).
- It's not extensible. For example, implementing flex-wrap requires adding a new multi-line logic to multiple places.
- It's hard to maintain. Changes in the main algorithm need to be ported to the baseline algorithm as well. Failing to do will likely introduce bugs that only occur when the new changes are used together with baseline alignment. #1604 is an example.
- It doesn't correctly handle stretched items ([failed snapshot test](https://github.com/nguyenhuy/AsyncDisplayKit/commit/3d8337c762159ab26becdeb7e8d810dca8b0ea94#diff-406d9e98f5ddec85da07c6fa2c49d603)). This problem happens because these items are stretched in the main algorithm, before baseline alignment is applied and increases the cross size of the stack. Those stretched items are also baseline-aligned!

This PR does a few things:
- It merges the two algorithms into one, following the CSS Flexbox algorithm. 
- Fixes #1604.
- Adds some snapshot test cases to make sure bugs exist in the old implementation are gone.
- Removes unused baselineRelativeArrangement property.
- Other small changes.
- It is the foundation of a flex-wrap implementation that I'm working on.


nguyenhuy:
Since this PRs touches a core component of the layout system, I really appreciate reviews from @maicki, @rcancro, @appleguy, @garrettmoon and @Adlai-Holler.

nguyenhuy:
Hm, the new snapshot test cases are failing on Jenkins. They pass locally! Now sure what is going on and how to see image differences on a Jenkins machine?

nguyenhuy:
Tests are all green now! It turned out text rasterization (antialiasing in particular) works a bit different on iOS 10 and that causes my new snapshot test cases which were captured on iOS 9 to fail. 

I will try to clean up our reference images tomorrow in a separate PR.

nguyenhuy:
@Adlai-Holler: Great catch on the misalignment. It's indeed a pre-existing issue. These snapshots were generated based on the current implementation on master. 

It has something to do with the baseline difference between the texts that should be floored to 0 but instead (indirectly) ceiled to 1 pixel. I'm investigating this but don't think it should block this PR.

Adlai-Holler:
OK I created the linked issue to track the rounding error. It's time to merge this! Awesome work @nguyenhuy – improvements in many dimensions with a single diff ✨

PR #2906, opened 2017-01-18
Minor Collection Simplifications

Adlai-Holler:
I found these while working on IGListKit support, and wanted to break them out into their own PR:

- Always use the same reuse identifier. As documented, UICV separates supplementaries by kind so we don't have to.
- Remove unused code from `ASCollectionViewFlowLayoutInspector`

### DIFF
--- a/AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
+++ b/AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
@@ -29,10 +29,6 @@
     unsigned int implementsConstrainedSizeForNodeAtIndexPathDeprecated:1;
     unsigned int implementsConstrainedSizeForItemAtIndexPath:1;
   } _delegateFlags;
-  
-  struct {
-    unsigned int implementsNumberOfSectionsInCollectionView:1;
-  } _dataSourceFlags;
 }
 
 #pragma mark Lifecycle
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/Details/ASCollectionViewFlowLayoutInspector.m
```diff
@@ -67,11 +63,7 @@ - (void)didChangeCollectionViewDelegate:(id<ASCollectionDelegate>)delegate;
 
 - (void)didChangeCollectionViewDataSource:(id<ASCollectionDataSource>)dataSource
 {
-  if (dataSource == nil) {
-    memset(&_dataSourceFlags, 0, sizeof(_dataSourceFlags));
-  } else {
-    _dataSourceFlags.implementsNumberOfSectionsInCollectionView = [dataSource respondsToSelector:@selector(numberOfSectionsInCollectionView:)];
-  }
+  // nop
```
nguyenhuy: Since this delegate method is optional, maybe we should just remove it here and [here](https://github.com/facebook/AsyncDisplayKit/pull/2906/files#diff-7187b8a17e0b7d3d8ec80608f32ac09eR43)?

Adlai-Holler: Good point! I did that, and added `AS_SUBCLASSING_RESTRICTED` to this class.

This is a breaking change! But subclassing this was never intended in the first place, and now we've just made it formal.

Adlai-Holler:
@nguyenhuy Ah! I meant "as documented in this PR", the UICollectionView behavior here is undocumented but unlikely to change.

PR #2913, opened 2017-01-19
ASLayoutSpec and ASDisplayNode should respect ASLayoutElementSize passed to them

nguyenhuy:
- The size is passed as a param of `-calculateLayoutThatFits:restrictedToSize:relativeToParentSize:`.
- It's almost always the same as `self.style.size`, but it should be used nevertheless as part of the API contract.

Would love to have a review from @maicki. Thanks!


nguyenhuy:
Actually the current code is fine. Closing...

PR #2914, opened 2017-01-19
[ASStackLayoutSpec] Implement flex wrap

### LINKED ISSUE #1213: [ASStackLayoutSpec] Add flex-wrap support (& back-port to ComponentKit or React Native?)
Hi,
Is there any way to achieve flex-wrap behaviour using existing stack layout spec?

nguyenhuy:
## Author: Thanh Huy Nguyen

Currently it's quite common for users to have layouts that are similar to collection view's flow layout or layouts that represent multiple horizontal lines stacked vertically. The latter can be achieved using nested horizontal stacks inside a vertical stack. While nesting works, it requires users to set flex properties on certain layout elements (i.e the direct children of a containing stack). Also, it's not trivial to figure out that you need to set `alignSelf 
= .Stretch` on a horizontal stack in order for it to fill the horizontal space of its containing vertical stack. In general, nested stacks are hard to understand and maintain. Flex wrap is a simpler alternative to achieve the same results.

Furthermore, soon we will need an async collection view flow layout object. It is simply a thin layer around a multi-line stack layout spec! We can even make `ASStackLayoutSpec` to layout cell nodes on multiple threads.

This diff is in a pretty good shape that I think it can take the first round of quick reviews. After #2892, adding flex wrap is easier. We essentially need to introduce a concept of lines (which also represents columns/pages in case it's a vertical stack). Then we need a new logic that collects items into lines and finally handle line alignment.

The default value is no wrap, so stacks currently used by users are simply 1 line stacks. There are fast paths that handle a 1 line stack and the code path is generally the same for this case. All existing snapshot test cases also ensure that this PR doesn't introduce any behaviour changes to this case.

Remaining tasks for this PR:
- [x] More testing
- [x] Another round of code clean up
- [x] Documentation

Resolves #1213.


AsyncDisplayKit/Private/ASStackUnpositionedLayout.mm
```diff
@@ -16,6 +16,8 @@
 #import "ASLayoutSpecUtilities.h"
```
nguyenhuy: GitHub always manages to collapse the most important file within a PR. Love it!

davitPiloyan93:
+1 For this,thanks @nguyenhuy.

QuynhNguyen:
`The latter can be achieved using nested horizontal stacks inside a vertical stack`
Found that out the hard way. Wasted few hours until I realized that. Thanks for implementing this!

AsyncDisplayKit/Layout/ASStackLayoutDefines.h
```diff
@@ -85,6 +85,22 @@ typedef NS_ENUM(NSUInteger, ASStackLayoutAlignSelf) {
   ASStackLayoutAlignSelfStretch,
 };
 
+// TODO documentation
+typedef NS_ENUM(NSUInteger, ASStackLayoutFlexWrap) {
+  ASStackLayoutFlexWrapNoWrap,
+  ASStackLayoutFlexWrapWrap,
```
QuynhNguyen: How about `ASStackLayoutFlexWrapDisable` and ``ASStackLayoutFlexWrapEnable` to avoid the double `Wrap`?

nguyenhuy: I'm not entirely happy with these names. However, I also want to stick to CSS Flexbox spec in which these values are called "wrap" and "no-wrap". Yoga uses the same convention (https://github.com/facebook/yoga/blob/master/yoga/YGEnums.h#L89).

nguyenhuy:
@Adlai-Holler, @maicki, @appleguy, @garrettmoon: This PR is ready. Appreciate your reviews here. Thank you!

AsyncDisplayKit.xcodeproj/xcshareddata/xcschemes/AsyncDisplayKit.xcscheme
```diff
@@ -75,6 +75,13 @@
...
          </BuildableReference>
       </MacroExpansion>
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "FB_REFERENCE_IMAGE_DIR"
+            value = "$(SOURCE_ROOT)/$(PROJECT_NAME)Tests/ReferenceImages"
+            isEnabled = "YES">
+         </EnvironmentVariable>
+      </EnvironmentVariables>
       <AdditionalOptions>
```
nguyenhuy: Somehow my XCode doesn't recognize `FB_REFERENCE_IMAGE_DIR` preprocessor macro. I'm switching to declare it in our scheme. It's the recommended way and a bit easier to enable/disable.

Adlai-Holler (approved):
Beautiful! Love the ASCII art in the documentation too!

nguyenhuy:
Woohoo, thanks for reviewing this, @Adlai-Holler!

PR #2915, opened 2017-01-19
[Buck] use snapshot_reference_images_path instead of copying images to XCTest bundle

nguyenhuy:
Previously, I found it difficult to reliably set a correct `FB_REFERENCE_IMAGE_DIR` to our test runner. So I told Buck to copy these images to our XCTest bundle and use them there. And because of that, it's not straight-forward to record and save new snapshots.

Thanks to https://github.com/facebook/buck/pull/1082 and https://github.com/facebook/buck/pull/1106, it's now possible to config the reference images path directly inside `apple_test` target. This PR does just that!


Adlai-Holler:
Nice Huy, way to get facebook/buck#1106 in the bank too!

PR #2922, opened 2017-01-24
[ASLayout] Consider the dedupe compiler flag to check for duplicate elements in a subtree

maicki:
We are currently check for duplicate layout elements in a subtree although the compiler flag to check it is not defined / enabled. We should consider the compiler flag so in production no extra work is done.


maicki:
cc @garrettmoon

PR #2923, opened 2017-01-24
[ASDataController] Simplify data controller

### LINKED ISSUE #2916: Simplify Data Controller Classes 1: Remove ASChangeSetDataController
This year I'd like to achieve some _significant_ refactoring to our data controller class cluster. It is a crucial piece of our architecture, and there's a lot that we can improve about it:

This first task is to remove `ASChangeSetDataController`. Currently, this class receives calls like `insertItemsAtIndexPaths:` and `deleteItemsAtIndexPaths:` etc and buffers them into an `_ASHierarchyChangeSet` (does not call super). Then, during `endUpdates`, it reads all the changes out of the hierarchy change set and makes the same insert/delete calls, _in safe order_ up to its parent `ASDataController`. 

This causes a number of headaches, because:
- It obscures what's really going on when you update a collection: you _transmit_ a changeset across a bunch of calls, and then you actually commit it in `-endUpdates`. So once they commit the change, we should forget all about the big stream of `reload/insert/delete` calls and think of it as just one change that we need to process.
- `ASDataController` has no context around the editing calls it receives. It just gets a stream of insert/delete calls during `endUpdates`, and for each one we switch queues and do all this stuff.
- Since it lacks context, its power is really limited, which is bad because it's an important class.
- It's very weird for a subclass to suppress, and then reorder, and then submit `super` calls for the calls it receives. It makes it tricky to nail down just what each method means!

So let's remove it, and add a single editing method to `ASDataController` like `updateWithChangeSet:`, and have `ASCollectionDataController` inherit directly from `ASDataController`. In the future we'll combine `ASCollectionDataController` and `ASDataController` since they duplicate so much functionality.

I've pushed a branch `SimplifyDataController` that is partially complete, but I'm swamped with work right now so if @nguyenhuy you would be willing to commandeer it I would appreciate it.

nguyenhuy:
@Adlai-Holler: Appreciate your review here.

Closes #2916.


AsyncDisplayKit/ASCollectionNode+Beta.h
```diff
@@ -17,12 +17,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout layoutFacilitator:(nullable id<ASCollectionViewLayoutFacilitatorProtocol>)layoutFacilitator;
 
-- (void)beginUpdates ASDISPLAYNODE_DEPRECATED_MSG("Use -performBatchUpdates:completion: instead.");
-
-- (void)endUpdatesAnimated:(BOOL)animated ASDISPLAYNODE_DEPRECATED_MSG("Use -performBatchUpdates:completion: instead.");
-
-- (void)endUpdatesAnimated:(BOOL)animated completion:(nullable void (^)(BOOL))completion ASDISPLAYNODE_DEPRECATED_MSG("Use -performBatchUpdates:completion: instead.");
```
Adlai-Holler: Unfortunately Pinterest still relies on this API so we can't remove it just yet.

nguyenhuy: @Adlai-Holler: I was hoping to reduce code complexity in ASCollectionView by removing these methods. I added them back, turns out it's not that bad.

PR #2926, opened 2017-01-24
[NSArray+Diffing] Use heap-based allocation for array diff.

appleguy:
This avoids compiler warnings when using the strictest clang settings.
It also ensures that stack overflow can't occur even with the largest datasets.

pasted image at 2017_01_23 08_16 pm

### DIFF
--- a/AsyncDisplayKit/Details/NSArray+Diffing.m
+++ b/AsyncDisplayKit/Details/NSArray+Diffing.m
@@ -60,18 +60,19 @@
   NSInteger arrayCount = array.count;
   
   // Allocate the diff map in the heap so we don't blow the stack for large arrays.
-  NSInteger (*lengths)[arrayCount+1] = NULL;
-  size_t lengthsSize = ((selfCount+1) * sizeof(*lengths));
-  // Would rather use initWithCapacity: to skip the zeroing, but TECHNICALLY
-  // `mutableBytes` is only guaranteed to be non-NULL if the data object has a non-zero length.
-  NS_VALID_UNTIL_END_OF_SCOPE NSMutableData *lengthsData = [[NSMutableData alloc] initWithLength:lengthsSize];
-  lengths = lengthsData.mutableBytes;
+  NSInteger **lengths = NULL;
+  lengths = (NSInteger **)malloc(sizeof(NSInteger*) * (selfCount+1));
   if (lengths == NULL) {
-    ASDisplayNodeFailAssert(@"Failed to allocate memory for diffing with size %tu", lengthsSize);
+    ASDisplayNodeFailAssert(@"Failed to allocate memory for diffing");
     return nil;
   }
   
   for (NSInteger i = 0; i <= selfCount; i++) {
+    lengths[i] = (NSInteger *)malloc(sizeof(NSInteger) * (arrayCount+1));
+    if (lengths[i] == NULL) {
+      ASDisplayNodeFailAssert(@"Failed to allocate memory for diffing");
+      return nil;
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
Just saw your comments on Slack! Wild that a cast won't fix it! What compiler setting should I use to trigger? Just curious to play with it myself, cool with landing the diff.

Adlai-Holler:
@appleguy The only way I have been able to trigger that warning is to use the C++ compiler to compile `NSArray+Diffing.m`. The cast is invalid under C++ because arrays aren't exactly pointers.

So we have three options:
- A. Land this diff
   - Pro: Easy!
   - Con: Lots of malloc/free calls
   - Con: Leaves the door open for future incompatibilities between your C++ build environment and our other `.m` files.
- B. Open a new diff that makes `NSArray+Diffing.mm` into a C++ file that uses a C++ heap-allocated matrix a la http://stackoverflow.com/a/21036413 except on the heap.
   - Pro: Clean! One call, no memory manipulation.
   - Con: Bigger change, more C++.
   - Con: Leaves the door open for future incompatibilities between your C++ build environment and our other `.m` files.
- C. Modify the build system you're using to ensure it compiles `.m` files using a C compiler.
   - Pro: No code change.
   - Pro: Formally correct – other incompatibilities may arise if `.m` files are compiled under C++
   - Con: May be a huge hassle for you.

If C is reasonably easy, I think we should go with that. Otherwise I think we should go with A.

Adlai-Holler:

@appleguy Any thoughts on the difficulty of getting your build environment to use the C compiler on this file?

appleguy:
@Adlai-Holler Will look into this soon.  Thanks a lot for your detailed presentation of options and being open to how this is solved -- I agree that figuring out the build config would be ideal now that there is a clear lead on why the error was occurring at all.

appleguy:
Is it an option to do the more-C++ approach with a simple vector, which is commonly used in other places in the codebase?

appleguy:
@Adlai-Holler As far as I can tell, the only way to force the C compiler to be used there would be to create a separate Bazel build target -- an actually separate static library -- so it wouldn't actually address the problem with other files.  I'm definitely worried about this because I didn't realize it was treating it differently than the mainline build.

However, since this has never occurred before (at least for any existing code), it raises the question: should we ever write C code that doesn't compile under C++?  In terms of portability, and in the spirit of the project's already-strict warning settings, _and_ given the widespread usage of .mm -- it seems like we should strive for C++ compatibility.

Let me know your thoughts on this.  I could close this PR if you see a better way to do it, or update it to use a vector or something else.  For the meantime, this will be the only "large" divergence in our fork at the moment, which is a pretty good place to be.  It'll be nice to eliminate, but isn't urgent.

maicki:
@Adlai-Holler @appleguy Let's try to land a solution soon so there are no problems for other developers who face the same problem with their build infra.

I personally would tend to go with the C++ approach.

Re project warning settings: We don't have a strict warning setting at the moment and I would see improve our warning game pretty soon.

appleguy:
Thanks for the input @maicki !  I would **love** to see some really pedantic warning settings enabled :).

Adlai-Holler:
OK since I think cases like this will be rare, let's ignore the idea of other possible incompatibilities when our files are compiled under C++, and go with option **B** to use a vector. @appleguy Are you willing to update the implementation and rename to `.mm`?

Adlai-Holler:
On second thought, we've got more important things to do. We can refactor/optimize this later as needed. Landing this diff.

PR #2932, opened 2017-01-25
[AsyncDisplayKit] Fix a few comment typos and minor compiler strictness warnings.

appleguy:


AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -15,7 +15,7 @@
 #import "ASDisplayNodeExtras.h"
 #import "ASDisplayNodeInternal.h" // Required for interfaceState and hierarchyState setter methods.
 #import "ASInternalHelpers.h"
-#import "ASMultiDimensionalArrayUtils.h"
+#import "ASMultidimensionalArrayUtils.h"
```
garrettmoon: This is weird, did anything actually change here?

appleguy: Yes, capitalization :).  It turns out the file is lowercase d in dimensional, and so this fails to compile on case-sensitive filesystems, which are required for the build environment it's being used in.

PR #2936, opened 2017-01-26
[Documentation] Improve readability of NS_OPTIONS in Swift

hannahmbanana:


maicki:
@hannahmbanana If we remove the comments directly from the enums will we loose the Xcode documentation in ObjC if you option-click on it within a source code?

maicki:
@hannahmbanana Can you please also rebase your branch on top of master. It seems like you merged in master a couple of times instead of rebasing.

hannahmbanana:
@maicki - interesting point about losing the option-click documentation. Shall we abandon this idea then?

Adlai-Holler:
I think we should _only_ have the `None` option specified in the main NS_OPTIONS documentation, and only if the behavior of none isn't obvious.

Adlai-Holler:
Nice improvement!

PR #2940, opened 2017-01-27
[ASLayoutTransition] Remove taking a snapshot in the default layout transition code

maicki:
cc @garrettmoon

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -1633,6 +1633,10 @@ ASLayoutElementFinalLayoutElementDefault
     insertedSubnode.frame = [context finalFrameForNode:insertedSubnode];
     insertedSubnode.alpha = 0;
   }
+  
+  // Adjust groupOpacity for animation
+  BOOL originAllowsGroupOpacity = node.allowsGroupOpacity;
+  node.allowsGroupOpacity = YES;
 
   [UIView animateWithDuration:self.defaultLayoutTransitionDuration delay:self.defaultLayoutTransitionDelay options:self.defaultLayoutTransitionOptions animations:^{
     // Fade removed subnodes and views out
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1611,15 +1610,6 @@ - (void)animateLayoutTransition:(id<ASContextTransitioning>)context
       CGRect fromFrame = [context initialFrameForNode:subnode];
       CGRect toFrame = [context finalFrameForNode:subnode];
       if (CGSizeEqualToSize(fromFrame.size, toFrame.size) == NO) {
-        // To crossfade resized subnodes, show a snapshot of it on top.
-        // The node itself can then be treated as a newly-inserted one.
-        UIView *snapshotView = [subnode.view snapshotViewAfterScreenUpdates:YES];
```
appleguy: I think even if we remove this entirely, it is better than before -- however, it would be great to support this as an option because there are some nice use cases.  How about including it in the same place that the upward-propagating "subnode PREVENTS ASCellNode from animation" option?  @nguyenhuy would know the name for this.  But, the idea would be, by default there are no snapshots...unless one subnode declares that it requires snapshots to transition, in which case the entire transition would be snapshotted?

Alternatively, we could have every node expose a .shouldSnapshotTransitions BOOL, and do some kind of performoneverysubnode call here.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -1612,15 +1611,6 @@ - (void)animateLayoutTransition:(id<ASContextTransitioning>)context
...
-        snapshotView.frame = [context initialFrameForNode:subnode];
-        snapshotView.alpha = 1;
-        
-        [node.view insertSubview:snapshotView aboveSubview:subnode.view];
-        [removedViews addObject:snapshotView];
-        
         [insertedSubnodes addObject:subnode];
```
appleguy: @maicki somewhere around here, let's enable groupOpacity at least on the root node, and then set it back to its previous value at the end of the animation.

Thanks a lot for putting this up.  If you don't have time to do the groupOpacity thing now, that's totally fine, but would you be OK with landing this patch as-is?  That will help with my merge conflicts but also allow our developers on Slack to use it more easily.

The two follow ups I remember (for another diff / maybe an issue):
- Allow snapshotting as some kind of mode, suggest it be controlled in the same way that subnodes of a cell can require "no animations" from their parent.
- Always enable groupOpacity when not snapshotting

maicki: @appleguy You are totally right!! I would like to add enable and disabling `groupOpacity` tonight / tomorrow morning into this PR and than let's land it.

Let's address the follow up steps in another PR.

maicki:
@appleguy Just added `groupOpacity` handling. Let me know what you think. Thanks!

PR #2941, opened 2017-01-28
[Layout] Improve Layout System Abstraction

maicki:
Currently our layout system is pretty coupled to the overall ASDK system. This PR is the first step to loosen up this coupling.

For now it:
1. Cleans up the protocol situation around `ASLayoutElement`.
2. Removes the dependency on `ASEnvironment` and introduces trait collection support that would allow us to more easier compile out in certain cases.
3. Move `ASLayoutElementStyleExtensions` out from `ASEnvironment`

## TODO:
- [x] Move `ASLayoutElementTraitCollection` related code to `ASTraitCollection`
- [x] Locking situation for layout element extensions in `_extensions`
- [x] Reenable logging in `ASLayoutElement`
- [x] Re-add removed `ASEnvironmentTraitCollection` as deprecated
- [x] Remove all `ASEnvironment` related files from project
- [x] Rename `ASLayoutElementTraitCollection` to `ASPrimitiveTraitCollection`


appleguy:
test this please

appleguy:
@maicki do you really think we should remove all of ASEnvironment?  Although adoption was not completed, I feel it would be *very* valuable (and is great, flexible code) if we can get ASInterfaceState and ASHierarchyState into the ASEnvironment struct, as was originally intended / in-progress when we had to jump to other projects.

One reason I say this is because the integration idea I have for Yoga will almost certainly be able to use ASEnvironment as a central mechanism for keeping the Yoga tree up to date with the ASDK children tree.

maicki:
@appleguy Yes for now we should get rid of `ASEnvironment` it is trivial to write this kind of tree traversal without having to hook into another system like `ASEnvironment`. Look how I replaced the trait collection system that used to use `ASEnvironment`.

maicki:
Ready for first review. CC @nguyenhuy

Adlai-Holler:
Nice! Thoughts:

- Since we don't support macOS and there aren't plans to do so, I think we shouldn't land any OSX-oriented code.
- Some of the names are verbose. How about:
   - `ASLayoutElementTraitCollection` -> `ASPrimitiveTraitCollection`. This conveys that it is a struct version of the class `ASTraitCollection`.
   - `ASLayoutElementTraitEnvironment` -> `ASTraitEnvironment`. This mirrors the `UITraitEnvironment` protocol and the word `Trait` is unambiguous within ASDK.
   - `progagateNewLayoutElementTraitCollection` -> `propagateNewTraitCollection`. Since the format used for propagation doesn't matter, we can omit it from the name.
   - `-layoutElementTraitCollection` -> `-primitiveTraitCollection`

@maicki

maicki:
@Adlai-Holler 
- I think it would be valuable to let the macOS related code in there as the layout system can be used on macOS without having the coupling with ASDK
-  Agree with naming of trait collections is not optimal. Will change the naming.

maicki:
@Adlai-Holler @appleguy @garrettmoon Ready for next review. Thanks!

maicki:
@Adlai-Holler Just rebased can you take another look. Thanks!

AsyncDisplayKit/Details/ASTraitCollection.h
```diff
@@ -8,12 +8,134 @@
...
+/// For backward compatibility reasons we redefine the old layout element trait collection struct name
+#define ASEnvironmentTraitCollection ASPrimitiveTraitCollection
+#define ASEnvironmentTraitCollectionMakeDefault ASPrimitiveTraitCollectionMakeDefault
+
+ASDISPLAYNODE_EXTERN_C_END
+
+/**
+ * Abstraction on top of UITraitCollection for propagation within AsyncDisplayKit-Layout
+ */
+@protocol ASPrimitiveTraitEnvironment <NSObject>
```
Adlai-Holler: Let's name this protocol `ASTraitEnvironment`.

AsyncDisplayKit/ASViewController.mm
```diff
@@ -102,8 +101,8 @@ - (void)loadView
   
   // ensure that self.node has a valid trait collection before a subclass's implementation of viewDidLoad.
   // Any subnodes added in viewDidLoad will then inherit the proper environment.
-  ASEnvironmentTraitCollection traitCollection = [self environmentTraitCollectionForUITraitCollection:self.traitCollection];
-  [self progagateNewEnvironmentTraitCollection:traitCollection];
+  ASPrimitiveTraitCollection traitCollection = [self primitiveTraitCollectionForUITraitCollection:self.traitCollection];
+  [self progagateNewPrimitiveTraitCollection:traitCollection];
```
Adlai-Holler: Nit: let's call this method `-propagateNewTraitCollection:`.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2487,7 +2488,8 @@ - (void)__setSupernode:(ASDisplayNode *)newSupernode
       }
       
       // Now that we have a supernode, propagate its traits to self.
-      ASEnvironmentStatePropagateDown(self, [newSupernode environmentTraitCollection]);
+      ASPrimitiveTraitCollectionPropagateDown(self, newSupernode.primitiveTraitCollection);
```
Adlai-Holler: Let's call this `ASTraitCollectionPropagateDown`. In cases where there won't be ambiguity, we should omit the word `Primitive` because it's non-essential.

maicki:
@Adlai-Holler Addressed your comments ready for next review. Thanks!

PR #2942, opened 2017-01-28
IGListKit Support II: Electric Boogaloo

Adlai-Holler:
This supersedes #2848 which was a bit hacky and messy.

- Add `ASCollectionDataSourceInterop` and `ASCollectionDelegateInterop` protocols to formalize what hooks we need to support IGListKit.
- Add `IGListAdapterBasedDataSource` to connect IGListKit to AsyncDisplayKit using the existing dataSource/delegate architecture.
- Beef up ASDKgram to use a header node on the IGListKit tab.

To use IGListKit:
- Have your section controllers conform to `ASSectionController`. 
- Have your supplementary view sources conform to `ASSupplementaryNodeSource`.
- On your `IGListAdapter`, call `-setASDKCollectionNode:` to make the connection.

Note: Use the provided implementations from `ASIGListKitMethodImplementations.h` to cover required IGListKit methods that don't apply when used with ASDK.

@ay8s @appleguy @nguyenhuy @maicki @garrettmoon @jessesquires @ocrickard

### DIFF
--- a/ASDKListKit/ASDKListKitTests/ASListKitTests.m
+++ b/ASDKListKit/ASDKListKitTests/ASListKitTests.m
@@ -1,0 +1,107 @@
+//
+//  ASListKitTests.m
+//  AsyncDisplayKit
+//
+//  Created by Adlai Holler on 12/25/16.
+//  Copyright © 2016 Facebook. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+#import <AsyncDisplayKit/AsyncDisplayKit.h>
+#import "ASListKitTestAdapterDataSource.h"
+#import "ASXCTExtensions.h"
+#import <JGMethodSwizzler/JGMethodSwizzler.h>
+
+@interface ASListKitTests : XCTestCase
+
+@property (nonatomic, strong) ASCollectionNode *collectionNode;
+@property (nonatomic, strong) UICollectionView *collectionView;
+@property (nonatomic, strong) IGListAdapter *adapter;
+@property (nonatomic, strong) ASListKitTestAdapterDataSource *dataSource;
+@property (nonatomic, strong) UICollectionViewFlowLayout *layout;
+@property (nonatomic, strong) UIWindow *window;
+@property (nonatomic) NSInteger reloadDataCount;
+
+@end
+
+@implementation ASListKitTests
+
+- (void)setUp {
+  [super setUp];
+
+  [ASCollectionView swizzleInstanceMethod:@selector(reloadData) withReplacement:JGMethodReplacementProviderBlock {
+    return JGMethodReplacement(void, ASCollectionView *) {
+      JGOriginalImplementation(void);
+      _reloadDataCount++;
+    };
+  }];
+
+  self.window = [[UIWindow alloc] initWithFrame:CGRectMake(0, 0, 100, 100)];
+
+  self.layout = [[UICollectionViewFlowLayout alloc] init];
+  self.collectionNode = [[ASCollectionNode alloc] initWithCollectionViewLayout:self.layout];
+  self.collectionNode.frame = self.window.bounds;
+  self.collectionView = self.collectionNode.view;
+
+  [self.window addSubnode:self.collectionNode];
+
+  IGListAdapterUpdater *updater = [[IGListAdapterUpdater alloc] init];
+
+  self.dataSource = [[ASListKitTestAdapterDataSource alloc] init];
+  self.adapter = [[IGListAdapter alloc] initWithUpdater:updater
+                                         viewController:nil
+                                       workingRangeSize:0];
+  self.adapter.dataSource = self.dataSource;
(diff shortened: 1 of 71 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -389,11 +389,10 @@ - (void)onDidLoad:(ASDisplayNodeDidLoadBlock)body
...
   if ([self _isNodeLoaded]) {
-    ASDisplayNodeFailAssert(@"Attempt to call %@ on node after it was loaded. Node: %@", NSStringFromSelector(_cmd), self);
-    return;
-  }
-  
-  if (_onDidLoadBlocks == nil) {
+    ASDisplayNodeAssertThreadAffinity(self);
+    ASDN::MutexUnlocker l(__instanceLock__);
+    body(self);
+  } else if (_onDidLoadBlocks == nil) {
```
Adlai-Holler: Changed this behavior to call `-didLoad` before any `onDidLoadBlocks`, and no longer to throw if they call on main on a loaded node (just runs block).

ASDKListKit/ASDKListKitTests/ASListKitTests.m
```diff
@@ -0,0 +1,107 @@
...
+@property (nonatomic, strong) ASListKitTestAdapterDataSource *dataSource;
+@property (nonatomic, strong) UICollectionViewFlowLayout *layout;
+@property (nonatomic, strong) UIWindow *window;
+@property (nonatomic) NSInteger reloadDataCount;
+
+@end
+
+@implementation ASListKitTests
+
+- (void)setUp {
```
appleguy: nit: The default-inserted methods use the brace-at-end-of-name style, so we should bump it down.

ASDKListKit/ASDKListKitTests/ASListTestObject.m
```diff
@@ -0,0 +1,46 @@
...
+    _value = value;
+  }
+  return self;
+}
+
+- (instancetype)copyWithZone:(NSZone *)zone {
+  return [[ASListTestObject alloc] initWithKey:self.key value:self.value];
+}
+
+
```
appleguy: nit: Extra return here.  Is there a standard on protocol names?  Would it make sense for the pragma mark to say <IGListDiffable> or <IGListDiffable> protocol ?

AsyncDisplayKit/ASCollectionNode.mm
```diff
@@ -26,8 +27,8 @@
 #pragma mark - _ASCollectionPendingState
 
 @interface _ASCollectionPendingState : NSObject
-@property (weak, nonatomic) id <ASCollectionDelegate>   delegate;
-@property (weak, nonatomic) id <ASCollectionDataSource> dataSource;
+@property (weak, nonatomic) id <ASCollectionDelegate>    delegate;
```
appleguy: Revert this change unless you see a benefit

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -27,6 +28,8 @@
 #import "ASSectionContext.h"
 #import "ASCollectionView+Undeprecated.h"
 #import "_ASHierarchyChangeSet.h"
+#import <objc/runtime.h>
```
appleguy: Put runtime.h at the bottom (or top).  This imports list should have its order cleaned up (alphabetized or grouped by functional area)

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -464,8 +470,8 @@ - (void)setAsyncDataSource:(id<ASCollectionDataSource>)asyncDataSource
   if (asyncDataSource == nil) {
     _asyncDataSource = nil;
     _proxyDataSource = _isDeallocating ? nil : [[ASCollectionViewProxy alloc] initWithTarget:nil interceptor:self];
-    
-    memset(&_asyncDataSourceFlags, 0, sizeof(_asyncDataSourceFlags));
+    _asyncDataSourceFlags = {};
+
```
appleguy: Should not have space after.  I'm split on this change -- I think the memset is much more explicit, at least since I'm not sure it is widespread knowledge that {} will zero memory for the specific size of a struct type (particularly in an equality like this, where we need to overwrite existing values, as opposed to initialization-time).

jessesquires:
💯  🎉

AsyncDisplayKit/ASCollectionView.mm
Adlai-Holler: I've applied all of your suggestions save this one, with the rationale that:

- This is a higher level of abstraction (symbol vs memory)
- If a dev doesn't understand what `= {}` means then they should look it up and learn, tough love 💪

Adlai-Holler:
Here we go!

rnystrom:
So awesome, great work @Adlai-Holler, going to have to play around w/ this!

Also 👋  @appleguy ❤️, we've come full circle!

PR #2943, opened 2017-01-28
[Layout] Add extensibility support to ASLayoutElementStyle

maicki:
With the introduction of `ASLayoutElementStyle` that combines all style related properties for a `ASLayoutElement` we need to provide some kind of extensibility support. This is important as it should be possible to add custom layout specs and extend the `ASLayoutElementStyle ` for usage within the custom layout specs.

This PR relies on #2941 to land before.

The general idea is outline in this gist via examples: https://gist.github.com/maicki/2630b51f62cbedc0cb34e0bcac96d6c9

Original diff: #2294

## TODO:
- [ ] Rebase on top of #2941 if this landed on master


appleguy:
@maicki Also, thanks for putting this up so quickly!  It will be very useful.

I'd be happy to help review & collaborate on the Layout cleanup diff this depends on, both because that cleanup is a great step in itself, and also because it will allow us to land this and close out two PRs that make nice steps forward :).

I will very likely be able to take on the Yoga integration test, although due to an offsite next week and other projects at the moment, I think it will take until the first week of February to start.  I may get further delayed, but I think it's surprisingly feasible that we might have a prototype by mid-Feb!

maicki:
Closing this as new implementation based on master is in #2975

PR #2945, opened 2017-01-28
[ASNodeController] First implementation of node controller class

hannahmbanana:
Paired programming with @appleguy. Example project - ASDKgram modified to use a node controller:

[ASDKgram-nodeController.zip](https://github.com/facebook/AsyncDisplayKit/files/736714/ASDKgram-nodeController.zip)

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Subclasses.h
+++ b/AsyncDisplayKit/ASDisplayNode+Subclasses.h
@@ -39,7 +39,61 @@ NS_ASSUME_NONNULL_BEGIN
  * variables.
  */
 
-@interface ASDisplayNode (Subclassing)
+@protocol ASInterfaceState <NSObject>
+@required
+
+/**
+ * @abstract Called whenever any bit in the ASInterfaceState bitfield is changed.
+ * @discussion Subclasses may use this to monitor when they become visible, should free cached data, and much more.
+ * @see ASInterfaceState
+ */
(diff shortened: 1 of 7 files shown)


AsyncDisplayKit/ASDisplayNode+Subclasses.h
```diff
@@ -39,7 +39,61 @@ NS_ASSUME_NONNULL_BEGIN
  * variables.
  */
 
-@interface ASDisplayNode (Subclassing)
+@protocol ASInterfaceState <NSObject>
```
Adlai-Holler: Let's name this e.g. `ASInterfaceStateDelegate`. It's OK for something to be its own delegate so this term applies both to nodes and node controllers.

AsyncDisplayKit/ASNodeController.mm
```diff
@@ -0,0 +1,57 @@
...
+  if (_node == nil) {
+    [self loadNode];
+  }
+  return _node;
+}
+
+-(void)setNode:(ASDisplayNode *)node
+{
+  _node = node;
+  node.interfaceStateDelegate = self;
```
Adlai-Holler: We should unhook from our old node `_node.interfaceStateDelegate = nil`.

Adlai-Holler:
@hannahmbanana Could you rebase? Then we'll land!

Adlai-Holler:
Rebased! Here we go!

hannahmbanana:
Opened Issue #2964 to track Adlai's suggestions.

PR #2948, opened 2017-01-29
[ASCollectionNode] Add support for using UICollectionViewCells

hannahmbanana:
Paired programming with @appleguy.

This method offers compatibility with synchronous, standard UICollectionViewCell objects. These cells will **not** have the performance benefits of ASCellNodes (like preloading, async layout, and async drawing) - even when mixed within the same ASCollectionNode.

In order to use this method, you must:
1. Implement it on your ASCollectionDataSource object.
2. Call registerCellClass: on the collectionNode.view (in viewDidLoad, or register an onDidLoad: block).
3. Return nil from the nodeBlockForItem...: or nodeForItem...: method. NOTE: it is an error to return nil from within a nodeBlock, if you have returned a nodeBlock object.
4. Lastly, you must implement a method to provide the size for the cell. There are two ways this is done:
4a. UICollectionViewFlowLayout (incl. ASPagerNode). Implement collectionNode:constrainedSizeForItemAtIndexPath:.
4b. Custom collection layouts. Set .view.layoutInspector and have it implement collectionView:constrainedSizeForNodeAtIndexPath:.

For an example of using this method with all steps above (including a custom layout, 4b.), see the app in examples/CustomCollectionView and enable kShowUICollectionViewCells = YES (example project included in this PR).

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -213,8 +213,10 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     unsigned int scrollViewWillBeginDragging:1;
     unsigned int scrollViewDidEndDragging:1;
     unsigned int scrollViewWillEndDragging:1;
+    unsigned int collectionViewWillDisplayCellForItem:1;
     unsigned int collectionViewWillDisplayNodeForItem:1;
     unsigned int collectionViewWillDisplayNodeForItemDeprecated:1;
+    unsigned int collectionViewDidEndDisplayingCellForItem:1;
     unsigned int collectionViewDidEndDisplayingNodeForItem:1;
     unsigned int collectionViewShouldSelectItem:1;
     unsigned int collectionViewDidSelectItem:1;
@@ -250,6 +252,7 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   struct {
     unsigned int collectionViewNodeForItem:1;
     unsigned int collectionViewNodeBlockForItem:1;
+    unsigned int collectionViewCellForItem:1;
     unsigned int collectionViewNodeForSupplementaryElement:1;
     unsigned int numberOfSectionsInCollectionView:1;
     unsigned int collectionViewNumberOfItemsInSection:1;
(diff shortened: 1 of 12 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -948,23 +954,30 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
 - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cellForItemAtIndexPath:(NSIndexPath *)indexPath
 {
-  _ASCollectionViewCell *cell = [self dequeueReusableCellWithReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
-  
   ASCellNode *node = [self nodeForItemAtIndexPath:indexPath];
-  cell.node = node;
-  [_rangeController configureContentView:cell.contentView forCellNode:node];
+  UICollectionViewCell *cell = nil;
   
-  if (!AS_AT_LEAST_IOS8) {
```
hannahmbanana: getting rid of this because we're dropping iOS 7 support

hannahmbanana:
cc @KieranLafferty

hannahmbanana:
Jenkins test this please

QuynhNguyen:
Just wondering, what are we getting out of this? From the description, it seems like using ASCellNode over UICollectionViewCell is a no brainer.

hannahmbanana:
@QuynhNguyen - great question! 

`ASCellNode` is always preferred. However, for code bases that have hundreds or more custom `UICollectionViewCells` subclasses, this feature will allow them to port to ASDK more easily, and then slowly convert their backlog of cells to be `ASCellNode` subclasses. 

This change allows mixing both those un-ported `UICollectionViewCell`s and `ASCellNode`s without a performance penalty for the `UIKitCells`. In addition, it still provides of `ASCellNode`, including automatic preload, asynchronous measurement and asynchronous rendering. Because of this difference in performance, app developers will clearly see the value of porting cells to `ASCellNode` and be able to do so over time. 

That said, I think @Adlai-Holler implemented this feature in #2942 so we may not need this PR anymore.

Adlai-Holler:
I'm closing this PR – supporting this use case will be much different in light of #2942 so let's take a fresh stab at it. If you disagree and want to fix forward from this PR, feel free to reopen.

PR #2966, opened 2017-02-01
[ASNodeController] Move ASNodeController to beta header

maicki:
`ASNodeController` is an experimental state at the moment where we don't know what implications and future potential it could have. Let's put it behind a beta header.


garrettmoon:
This seems like a pretty good idea to me. I'm not sure we have a full picture of what we'd like the node controller to be yet and we should have that before exposing this outside a beta header.

hannahmbanana:
@maicki - Excellent idea. Beta didn't occur to me when adding the file :P

maicki:
@Adlai-Holler Yep just changed to `.m`. Will merge after CI passes.

PR #2969, opened 2017-02-02
Improve TableLayoutController: Simplify, Fix Crash & Improve Perf

### LINKED ISSUE #2823: [ASFlowLayoutController] Crash in findIndexPathAtDistance:fromIndexPath:
https://www.fabric.io/pinterest6/ios/apps/pinterest/issues/579b21f7ffcdc04250aeefcf
Fatal Exception: NSRangeException
0  CoreFoundation                 0x19255c1c0 __exceptionPreprocess
1  libobjc.A.dylib                0x190f9455c objc_exception_throw
2  CoreFoundation                 0x1924384ac -[__NSArrayM removeObjectAtIndex:]
3  Pinterest                      0x10083fd88 -[ASFlowLayoutController findIndexPathAtDistance:fromIndexPath:] (ASFlowLayoutController.mm:133)
4  Pinterest                      0x10083f810 -[ASFlowLayoutController indexPathsForScrolling:rangeMode:rangeType:] (ASFlowLayoutController.mm:78)
5  Pinterest                      0x10085fd3c -[ASRangeController _updateVisibleNodeIndexPaths] (ASRangeController.mm:228)
6  Pinterest                      0x10085f8a4 __35-[ASRangeController setNeedsUpdate]_block_invoke (ASRangeController.mm:127)
7  libdispatch.dylib              0x1913e5200 _dispatch_call_block_and_release
8  libdispatch.dylib              0x1913e51c0 _dispatch_client_callout
9  libdispatch.dylib              0x1913e9d6c _dispatch_main_queue_callback_4CF
10 CoreFoundation                 0x192509f2c __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
11 CoreFoundation                 0x192507b18 __CFRunLoopRun
12 CoreFoundation                 0x192436048 CFRunLoopRunSpecific
13 GraphicsServices               0x193eb9198 GSEventRunModal
14 UIKit                          0x198409628 -[UIApplication _run]
15 UIKit                          0x198404360 UIApplicationMain
16 Pinterest                      0x100058aa4 main (main.m:47)
17 libdispatch.dylib              0x1914185b8 (Missing)

Adlai-Holler:
This class was previously named `ASFlowLayoutController` and intended to be used with collection view flow layouts as well as tables. But it wasn't ever used with flow layouts, and it wouldn't work with flow layouts.

Resolves #2823 

So:
- Replace `ASFlowLayoutController` with `ASTableLayoutController` which is more purpose-built.
- Use `indexPathsForRowsInRect:` to find target index paths directly, rather than going one-by-one summing up node heights. Much more direct, and accounts for section headers/footers correctly.
- Since we no longer go one-by-one through items, we don't create any more index paths than we need.
- That means there's no longer any reason to have our `ASIndexPath` struct.
- Between the fact that it's unused, and since NSIndexPaths use tagged pointers on iPhone 5S+, I've removed `ASIndexPath`. RIP, you served us well.

I've tested this by hand with ASDKgram and the debugger and everything is working as expected. I think this is a significant improvement on multiple dimensions.


AsyncDisplayKit/Details/ASTableLayoutController.m
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * Solution: Make a target rect, starting at height 100, and if we don't
+   * find any rows, keep doubling its height and searching again. In practice,
+   * this will virtually always find a row on the first try (unless you have a 
+   * section header more than 100 tall and we land near the middle.)
+   */
+  NSIndexPath *result = nil;
+  for (CGRect targetRect = CGRectMake(0, targetY, 1, kInitialRowSearchHeight);
+       result == nil;
+       targetRect = CGRectInset(targetRect, 0, -CGRectGetHeight(targetRect) / 2.0)) {
```
nguyenhuy: Shouldn't we advance `targetRect` only in the same direction as `distance`? Please correct me if I'm wrong, but if we expand it on both direction, I think it's possible that we'll return an index path of `start - 1` even though the distance is positive (and the other way around).

appleguy:
@Adlai-Holler awesome!  Indeed the original did not support true flow layouts.  As a further step — especially as an implementation of an ASCollectionViewFlowLayout that can do async computation — what do you think about using Yoga as a C-level implementation of the flow algorithm + caching the entire layout in C, and then generating and/or caching the layout elements as they are requested?

garrettmoon:
@appleguy are you suggesting we make Yoga a required dependency?

AsyncDisplayKit/Details/ASTableLayoutController.m
Adlai-Holler: Good point – we should probably err on the side of _more_ than ~2 screenfuls rather than less, so I've updated it to always search at the furthest extent in the aiming direction.

Adlai-Holler:
@appleguy I love the idea – once we have our own layout protocol we'll need a flow layout implementation, and that would be a great way to drive it. For the current API, UICollectionViewFlowLayout is absolutely sufficient IMO.

AsyncDisplayKit/Details/ASTableLayoutController.m
```diff
@@ -0,0 +1,200 @@
...
+   */
+  NSIndexPath *result = nil;
+  for (CGRect searchRect = [ASTableLayoutController initialSearchRectDownward:downward targetY:targetY];
+       // continue while result is nil
+       result == nil;
+       // grow search rect after each loop
+       searchRect = [ASTableLayoutController growSearchRect:searchRect downward:downward]) {
+    NSArray *rows = [_tableView indexPathsForRowsInRect:searchRect];
+    if (downward) {
+      result = [ASTableLayoutController fastArrayMax:rows];
```
nguyenhuy: I actually think we should honor the distance and grab the min index path here.

nguyenhuy:
@appleguy Caching the flow layout will be great! Although I'm not sure if we want to make Yoga as a required dependency. Please correct me if I'm wrong, but I believe our ASStackLayoutSpec - once its flex-wrap implementation is merged (#2914) - will serve this purpose quite well for the time being, at least until we decide to use Yoga as our main layout system which will probably no happen any time soon.

AsyncDisplayKit/Details/ASTableLayoutController.m
Adlai-Holler: Ah great point! Will make the change.

Adlai-Holler:
Thanks @appleguy and @nguyenhuy for your reviews! Comments addressed, merging…

PR #2973, opened 2017-02-03
Force ASPagerNode Pages to Fill Pager Bounds

Adlai-Holler:
- Instead of using `[CGSizeZero, pagerNode.bounds.size]` as the size range for pages, use exactly `pagerNode.bounds.size`. Much of the logic in `ASPagerNode` depends on the pages filling the bounds, and we've had issues in Pinterest where pages did not fill the bounds and that was very unexpected and annoying.
- Deprecate `pagerNode:constrainedSizeForNodeAtIndex:` with no replacement. As said above, the only appropriate size range for a page in a pager node is exactly `pagerNode.bounds.size` which is now the default behavior.

### DIFF
--- a/AsyncDisplayKit/ASPagerNode.h
+++ b/AsyncDisplayKit/ASPagerNode.h
@@ -66,7 +66,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @param index The index of the node.
  * @return A constrained size range for layout the node at this index.
  */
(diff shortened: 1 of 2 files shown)


Adlai-Holler:
Yeah for example `-currentPageIndex` and `-scrollToPage` won't work if your pages aren't full-sized. If we want to support partially-sized pages in the future, we should allocate resources for it and make sure that in the common case (full-sized pages) users don't have to resort to overriding an obscure delegate method.

PR #2987, opened 2017-02-06
[ASLayoutSpec] Remove legacy and unused parent property

maicki:


AsyncDisplayKit/Layout/ASLayoutSpec.h
```diff
@@ -26,11 +26,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign) BOOL isMutable;
 
 /**
- * Parent of the layout spec
- */
-@property (nullable, nonatomic, weak) id<ASLayoutElement> parent;
```
garrettmoon: Does this need to be deprecated?

maicki: It was never used outside of the framework (and actually in the framework it was not used either if you want to be precise). You can argue it was publicly available, but there was no reason to use it in the first place.

garrettmoon: Ok, it's pretty unlikely anyone was using this.

PR #2988, opened 2017-02-06
Restore IGListKit Support

Adlai-Holler:
We recently accidentally broke IGListKit support when we added support for full nodeless interop.

This diff ensures that both modes (IGListKit and full passthrough) work correctly.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -173,8 +173,11 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     unsigned int collectionNodeWillDisplaySupplementaryElement:1;
     unsigned int collectionNodeDidEndDisplayingSupplementaryElement:1;
     unsigned int shouldBatchFetchForCollectionNode:1;
-    // Whether this delegate conforms to ASCollectionDataSourceInterop
+
+    // Interop flags
     unsigned int interop:1;
+    unsigned int interopWillDisplayCell:1;
+    unsigned int interopDidEndDisplayingCell:1;
   } _asyncDelegateFlags;
   
   struct {
@@ -193,6 +196,10 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 
     // Whether this data source conforms to ASCollectionDataSourceInterop
     unsigned int interop:1;
+    // Whether this interop data source returns YES from +dequeuesCellsForNodeBackedItems
+    unsigned int interopAlwaysDequeue:1;
+    // Whether this interop data source implements viewForSupplementaryElementOfKind:
+    unsigned int interopViewForSupplementaryElement:1;
   } _asyncDataSourceFlags;
   
   struct {
@@ -411,7 +418,13 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     _asyncDataSourceFlags.collectionNodeContextForSection = [_asyncDataSource respondsToSelector:@selector(collectionNode:contextForSection:)];
     _asyncDataSourceFlags.collectionNodeNodeForSupplementaryElement = [_asyncDataSource respondsToSelector:@selector(collectionNode:nodeForSupplementaryElementOfKind:atIndexPath:)];
     _asyncDataSourceFlags.collectionNodeSupplementaryElementKindsInSection = [_asyncDataSource respondsToSelector:@selector(collectionNode:supplementaryElementKindsInSection:)];
+
     _asyncDataSourceFlags.interop = [_asyncDataSource conformsToProtocol:@protocol(ASCollectionDataSourceInterop)];
+    if (_asyncDataSourceFlags.interop) {
+      id<ASCollectionDataSourceInterop> interopDataSource = (id<ASCollectionDataSourceInterop>)_asyncDataSource;
+      _asyncDataSourceFlags.interopAlwaysDequeue = [[interopDataSource class] respondsToSelector:@selector(dequeuesCellsForNodeBackedItems)] && [[interopDataSource class] dequeuesCellsForNodeBackedItems];
+      _asyncDataSourceFlags.interopViewForSupplementaryElement = [interopDataSource respondsToSelector:@selector(collectionView:viewForSupplementaryElementOfKind:atIndexPath:)];
+    }
 
     ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNumberOfItemsInSection || _asyncDataSourceFlags.collectionViewNumberOfItemsInSection, @"Data source must implement collectionNode:numberOfItemsInSection:");
     ASDisplayNodeAssert(_asyncDataSourceFlags.collectionNodeNodeBlockForItem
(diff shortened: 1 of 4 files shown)


AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -906,25 +912,31 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
...
+      return cell;
+    }
   } else {
     ASDisplayNodeAssert(node != nil, @"Cell node should exist. indexPath = %@, collectionDataSource = %@", indexPath, self);
     cell = [self dequeueReusableCellWithReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
-    [(_ASCollectionViewCell *)cell setNode:node];
-    [_rangeController configureContentView:cell.contentView forCellNode:node];
   }
+
+  [(_ASCollectionViewCell *)cell setNode:node];
```
maicki: Can we be sure that it's always a `_ASCollectionViewCell`?

Adlai-Holler:
It's complicated. IGListKit never uses UIKit cells. But we still have to dequeue cells from IGListKit, before installing our cell nodes inside them. There was a fundamental misunderstanding about how the Interop protocol worked.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -411,7 +418,13 @@ - (void)setAsyncDataSource:(id<ASCollectionDataSource>)asyncDataSource
     _asyncDataSourceFlags.collectionNodeContextForSection = [_asyncDataSource respondsToSelector:@selector(collectionNode:contextForSection:)];
     _asyncDataSourceFlags.collectionNodeNodeForSupplementaryElement = [_asyncDataSource respondsToSelector:@selector(collectionNode:nodeForSupplementaryElementOfKind:atIndexPath:)];
     _asyncDataSourceFlags.collectionNodeSupplementaryElementKindsInSection = [_asyncDataSource respondsToSelector:@selector(collectionNode:supplementaryElementKindsInSection:)];
+
     _asyncDataSourceFlags.interop = [_asyncDataSource conformsToProtocol:@protocol(ASCollectionDataSourceInterop)];
+    if (_asyncDataSourceFlags.interop) {
```
appleguy: Does this work if you don't implement viewForSupplementaryElement?  I think we won't pass the conformsToProtocol: check if the datasource doesn't implement that separately.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -411,7 +418,13 @@ - (void)setAsyncDataSource:(id<ASCollectionDataSource>)asyncDataSource
...
+      id<ASCollectionDataSourceInterop> interopDataSource = (id<ASCollectionDataSourceInterop>)_asyncDataSource;
```
appleguy: Probably not needed to do this cast, because all the operations are dynamic / runtime methods

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -490,6 +503,11 @@ - (void)setAsyncDelegate:(id<ASCollectionDelegate>)asyncDelegate
     _asyncDelegateFlags.collectionNodeCanPerformActionForItem = [_asyncDelegate respondsToSelector:@selector(collectionNode:canPerformAction:forItemAtIndexPath:sender:)];
     _asyncDelegateFlags.collectionNodePerformActionForItem = [_asyncDelegate respondsToSelector:@selector(collectionNode:performAction:forItemAtIndexPath:sender:)];
     _asyncDelegateFlags.interop = [_asyncDelegate conformsToProtocol:@protocol(ASCollectionDelegateInterop)];
+    if (_asyncDelegateFlags.interop) {
```
appleguy: Same regarding the cast -- as well as the case of only one of willDisplay vs didEndDisplaying being implemented.

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -882,20 +900,26 @@ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollection
...
 {
+  if ([_registeredSupplementaryKinds containsObject:kind] == NO) {
+    [self registerSupplementaryNodeOfKind:kind];
+  }
+  
   UICollectionReusableView *view = nil;
   ASCellNode *node = [_dataController supplementaryNodeOfKind:kind atIndexPath:indexPath];
 
-  if (node.shouldUseUIKitCell && _asyncDataSourceFlags.interop) {
+  BOOL shouldDequeueExternally = _asyncDataSourceFlags.interopViewForSupplementaryElement && (_asyncDataSourceFlags.interopAlwaysDequeue || node.shouldUseUIKitCell);
```
appleguy: Great, this seems right -- though complicated, it satisfies some pretty tricky use cases.  It will be a happy day to find a further simplification to this...

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -882,20 +900,26 @@ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollection
...
-    if ([_registeredSupplementaryKinds containsObject:kind] == NO) {
-      [self registerSupplementaryNodeOfKind:kind];
-    }
     view = [self dequeueReusableSupplementaryViewOfKind:kind withReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
-    if (node) {
-      [_rangeController configureContentView:view forCellNode:node];
-    }
+  }
+
+  if (!node.shouldUseUIKitCell) {
```
appleguy: Totally stylistic, but might suggest using == NO in these cases for more explicitness / slightly improved readability

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -906,25 +930,32 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
   UICollectionViewCell *cell = nil;
   ASCellNode *node = [self nodeForItemAtIndexPath:indexPath];
 
-  if (node.shouldUseUIKitCell && _asyncDataSourceFlags.interop) {
+  BOOL shouldDequeueExternally = _asyncDataSourceFlags.interopAlwaysDequeue || (_asyncDataSourceFlags.interop && node.shouldUseUIKitCell);
```
appleguy: It's a bit weird how the logic here is flipped, but I think it is correct

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -906,25 +930,32 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
...
     cell = [(id<ASCollectionDataSourceInterop>)_asyncDataSource collectionView:collectionView cellForItemAtIndexPath:indexPath];
   } else {
-    ASDisplayNodeAssert(node != nil, @"Cell node should exist. indexPath = %@, collectionDataSource = %@", indexPath, self);
     cell = [self dequeueReusableCellWithReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
-    [(_ASCollectionViewCell *)cell setNode:node];
+  }
+
+  ASDisplayNodeAssert(node != nil, @"Cell node should exist. indexPath = %@, collectionDataSource = %@", indexPath, self);
+
+  if (_ASCollectionViewCell *asCell = ASDynamicCast(cell, _ASCollectionViewCell)) {
```
appleguy: Nice usage of ASDynamicCast!  Slightly less of a fan of doing the assignment inside the (), but I'm guessing you love it :) (I'm not opposed to it landing)

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -906,25 +930,32 @@ - (UICollectionViewCell *)collectionView:(UICollectionView *)collectionView cell
...
 - (void)collectionView:(UICollectionView *)collectionView willDisplayCell:(_ASCollectionViewCell *)cell forItemAtIndexPath:(NSIndexPath *)indexPath
 {
+  if (_asyncDelegateFlags.interopWillDisplayCell) {
+    [(id <ASCollectionDelegateInterop>)_asyncDelegate collectionView:collectionView willDisplayCell:cell forItemAtIndexPath:indexPath];
+  }
+
   // Since _ASCollectionViewCell is not available for subclassing, this is faster than isKindOfClass:
   // We must exit early here, because only _ASCollectionViewCell implements the -node accessor method.
   if ([cell class] != [_ASCollectionViewCell class]) {
-    if (_asyncDelegateFlags.interop) {
```
appleguy: The theory of having this here is that we shouldn't pester the UIKit method in cases where only a node exists.  At least for the use case I have in mind, this would be preferable and indeed may be important.  Would it work to use the value of +always... as an additional check on whether the UICollectionView method is called in the _ASCollectionViewCell case?

PR #2989, opened 2017-02-06
Fix a warning preventing pod spec validation.

garrettmoon:


garrettmoon:
Gonna merge because CI is being fixed and this is blocking the 2.1 release.

Issue #2990, opened 2017-02-06
Remove unused visualizing layout spec code


maicki:
Currently we have code around visualizing layout spec code that is not used, but needs to be considered somehow on every change we would like to do.

I would to start a discussion and propose to go on with cleaning up the code base by removing the visualizing code.

This does not mean this is the death of this feature as we would like to find a better way to integrate it that would not interfere so deeply with how the framework is working internally.

Related PR: #2991

hannahmbanana:
Two comments from me on this:
- I think the ability to "visualize" layout specs is still very badly needed by the community (as well as Pinterest). However, my attempt to create a layout visualization feature is only half-baked and not ready for prime-time. I don't think I can bring the ideal feature to fruition, but maybe you or @nguyenhuy could do so now, or in the future. 
- I understand your concern around maintaining a system that isn't actively used and support removing this for now and revisiting with a new / modified approach later on.

hannahmbanana:
I'm going to close this issue and move my comments to the PR.

PR #2991, opened 2017-02-06
[Debug] Remove not used code around visualizing layout specs

maicki:
Related Issue: #2990


appleguy:
@maicki I think the thought behind this is good.  I won't be upset if the team decides to land this — however, I think there is a case to be considered before removing this entirely.

1. Let's consider how much it interferes with refactoring.  In particular, I don't think we should remove the Inspector or Example code (but just move them to an experimental/ folder instead).  It probably makes more sense to remove the ASDisplayNode level integration, but that doesn't mean that all code which is otherwise "out of the way" also must go.
2. I think we all agree that debugging tools around layout, especially visualization, would be one of the most incredible capabilities that has ever been added to ASDK...and, for better or worse, it is very much **needed** as developers struggle every day on this.  Many hours lost trying to figure out tiny issues.
3. Since the visualization support does work, but it's mostly the editor / example environment that wasn't complete, IMO we should put it on ice until it is revived.  This is on the assumption that whether it's next month, or a year from now, it's almost certain that we'll eventually want this functionality.

That said, I wouldn't suggest that anyone spend many hours carefully preserving the functionality.  Instead, do you think there's a reasonable way to "move aside" any code that's in the way, and then break the integration with ASDisplayNode?

hannahmbanana:
You have my blessing to remove it! :)

maicki:
@appleguy @Adlai-Holler @hannahmbanana @nguyenhuy Thank you for all of your inputs. I will merge this in as soon as CI is passing.

Issue #2955, opened 2017-01-31
[ASCollectionNode] Cells must relayout if bounds changes after -reloadData called with zero-size bounds.


appleguy:
Currently, this flow occurs:

1. ASCollectionNode with UICollectionViewFlowLayout, OR with any layoutInspector that depends on collectionView.bounds.size within its constrainedSizeForItem: method.  This is common, e.g. using the bounds width as the constraint.
2. App calls reloadData manually during initialization, at any point before layout occurs.  bounds is CGRectZero at this time.
3. Cell measurement proceeds with 0 width max constraint, and 0, INF height constraint.  **Problem A:** This measurement should be skipped, because the max constraint has zero x/y area.
4. layout then occurs on the UICollectionView and sets its bounds to a non-zero size.
5. **Problem B** Due to some kind of optimization, the ASCellNodes are not asked to layout again at the new size, and so don't show up onscreen besides 0, 0 size (but in the view hierarchy)

appleguy:
@garrettmoon feel free to schedule this whenever it is feasible for the team.  I have some other bug fixes coming for ASCollection*, but probably won't get to this one which would be really great to fix.

There's a related problem where if a 0,0 CGExactSize is passed as the constrainedSize for a node, some kind of incorrect value gets into UICollectionView and causes the "infinite spinning / livelock" issue that CGFLOAT_MAX does...

Adlai-Holler:
Resolved in the linked PRs. Great observations @appleguy these are major problems!

PR #3000, opened 2017-02-07
Avoid Measuring Cell Nodes with Empty Size Range

Adlai-Holler:
Addresses "A" from #2955

### DIFF
--- a/AsyncDisplayKit/Layout/ASDimension.h
+++ b/AsyncDisplayKit/Layout/ASDimension.h
@@ -212,6 +212,14 @@ extern ASSizeRange const ASSizeRangeZero;
 extern ASSizeRange const ASSizeRangeUnconstrained;
 
 /**
+ * Returns whether a size range has zero max width or height.
+ */
+ASDISPLAYNODE_INLINE AS_WARN_UNUSED_RESULT BOOL ASSizeRangeIsEmpty(ASSizeRange sizeRange)
+{
+  return (sizeRange.max.width == 0 || sizeRange.max.height == 0);
+}
+
+/**
  * Creates an ASSizeRange with provided min and max size.
  */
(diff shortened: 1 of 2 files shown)


AsyncDisplayKit/Layout/ASDimension.h
```diff
@@ -212,6 +212,14 @@ extern ASSizeRange const ASSizeRangeZero;
 extern ASSizeRange const ASSizeRangeUnconstrained;
 
 /**
+ * Returns whether a size range has zero max width or height.
+ */
+ASDISPLAYNODE_INLINE AS_WARN_UNUSED_RESULT BOOL ASSizeRangeIsEmpty(ASSizeRange sizeRange)
+{
+  return (sizeRange.max.width == 0 || sizeRange.max.height == 0);
```
appleguy: Some suggestions on how to make these checks more effective:

- These numbers can be (and have seen them be) less than zero sometimes, due to client code insetting bounds that is often transiently zero sized:
- Let's consider using (value < FLT_EPSILON) as I think this is more formally correct.  This is probably less crucial for 0.0 itself, since floats can represent it exactly.
- One step further than FLT_EPSILON -- since sometimes "dirty", low precision math is used to come up with layout values, I wouldn't be surprised if some apps pass in a value like 0.00272 here occasionally.  I wonder if we should make it (value < 0.1), which while arbitrary, is most likely to help more cases than it hurts (any case that has an intentional bounds of 0.1 points is definitely unusual!)

Adlai-Holler: Although these muddy our abstraction and leave us on the hook to handle invalid layout values from users in the future, the cost of being lenient here is probably justified. Let's go with it for now.

In the future, maybe we will go further and add an `ASSizeRangeNormalize` function that ensures everything is positive and than `max > min` etc.

Issue #2851, opened 2017-01-03
[ASScrollNode] Sizing bug


jennibunny:
Hello,

`ASDisplayNode` is inside `ASScrollNode`.

I want to resize `ASDisplayNode height` based on its subnodes and `ASScrollNode contentSize` based on that `ASDisplayNode`.

8ofproject:
I have the same issue.
Cant find a way to make ASScrollNode's child node bigger then ASScrollNode itself to enable scrolling (all scrolling-bouncing properties set to true).
ASTableNode works, but too complex for the task.
ASDK 2.0.1
Can make example project if necessary.

8ofproject:
The issue looks related to https://github.com/facebook/AsyncDisplayKit/issues/2853

8ofproject:
Example project:
look at _HowJobDoneViewController.swift_
[ScrollExample.zip](https://github.com/facebook/AsyncDisplayKit/files/691700/ScrollExample.zip)

hashemp206:
thanks for uploading sample project. I checked, it seem `_contentCalculatedSizeFromLayout` in ASScrollNode that used for `contentSize` of scrollView, has wrong size! but @hannahmbanana needs more investigation.

`ASScrollNode`:
<img width="928" alt="screen shot 2017-01-08 at 2 44 25 pm" src="https://cloud.githubusercontent.com/assets/3206748/21749498/b5c73ac6-d5b4-11e6-8163-b937f02d7ca5.png">.

custom child node of ASScrollNode:
<img width="1099" alt="screen shot 2017-01-08 at 3 13 26 pm" src="https://cloud.githubusercontent.com/assets/3206748/21749520/29f85826-d5b5-11e6-9f29-2466252dcf25.png">

hannahmbanana:
@jennibunny @hashemp206 @8ofproject - does this commit address your bug?
e264d94

hashemp206:
@hannahmbanana thanks, commit doesn't fixed the issue. here is the sample project: https://github.com/hashemp206/ASScrollNode-Example

8ofproject:
@hannahmbanana I can confirm that nothing has changed (unfortunately).

davertay:
+1 for same issue. Example project that one would expect to calculate the correct contentSize height but doesn't: https://github.com/davertay/asdkScrollExample

garrettmoon:
Going to close this out since it's community debugging and hasn't been commented on in a while.

PR #3001, opened 2017-02-08
[ASScrollNode] Add .scrollDirection property so that internal content sizing can be easily "unlimited" in the direction of scrolling.

appleguy:


rurza:
This fixes issue when setting automaticallyManagesContentSize didn't work as expected.
I tried with different sizes for preferredSize, implementation of nodeConstrainedSize in VC. Nothing.
With this PR just set scrollDirection. 
I think that setting scrollDirection should set automaticallyManagesContentSize though.

davertay:
This works for my example where I have an ASScrollNode with a simple vertical ASStackLayoutSpec containing nodes all set to 100% width: https://github.com/davertay/asdkScrollExample

Without this change the contentSize was always being set to the view bounds, even when overriding nodeConstrainedSize in the ASViewController.

Note that this probably also fixes #2851

appleguy:
Discussed in Slack.  Thanks for the reviews everyone!

PR #3007, opened 2017-02-09
Route setDelegate: to setAsyncDelegate:

### LINKED ISSUE #2245: Allow Setting ASCollectionView.delegate for Interop with TLYShyNavBar etc.
There are several 3rd party components that require access to scrollview delegates directly or indirectly (you just pass a reference to the scrollview). Categories include "Pull to Refresh" and "Hide Navbar on scroll down" components, such as: TLYShyNavBar, BLKFlexibleHeightBar, AMScrollingNavbar...etc

Currently we get the following error:

```
Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'ASCollectionView uses asyncDelegate, not UICollectionView's delegate property.'
```

Issues opened because of this limitation:
https://github.com/telly/TLYShyNavBar/issues/72
https://github.com/telly/TLYShyNavBar/issues/158

Furthermore, It's not feasible for these 3rd party components to support asyncDelegate.

One suggestion is to support switching between asyncDelegate and delegate by passing a configuration, or by optionally chaining the two.

Chat log with @maicki:

```
simo [12:25 PM]  
asyncDelegate vs delegate
perhaps something that could be addressed in ASDK 2
so currently I get the following error "ASCollectionView uses asyncDelegate, not UICollectionView's delegate property."
It’s not because I’m explicitly accessing .delegate. It’s because I’m binding the ASCollectionNode scroll view to a plugin that does it under the hood `shyNavBarManager.scrollView = collectionView.view`

I’m using this: ` var collectionView: ASCollectionNode `

maicki [12:30 PM]  
You could subclass `ASCollectionView` and handle connect `asyncDelegate <-> delegate`
we had something similar in the past, just didn’t remember exactly how we solved it

simo [12:32 PM]  
so my thought was, why not continue to support  access to .delegate. We know there are tons of 3rd party components that can’t support ASDK for this. Encourage asyncDelegate, but allow access to delegate

maicki [12:33 PM]  
Currently the reason we don’t allow it is that we have a proxy between `delegate <-> asyncDelegate` (edited)

simo [12:33 PM]  
ah ok. It seems like if a 3rd party component wanted to support asyncDelegate, they would have a dependency on ASDK which is not recommended for standalone components, as it could create version mismatch

maicki [12:35 PM]  
As said I think you could actually support it if you plugin a proxy between `setDelegate -> asyncDelegate`. Basically it would be `collectionView.delegate -> proxy -> collectionView.asyncDelegate -> proxy -> super.delegate`. You could archive that I guess by overwritting `setDelegate:` in your `ASCollectionView` subclass

Does that makes sense?

simo [12:36 PM]  
what does `proxy` represent?

maicki [12:37 PM]  
Look into how `NSProxy` is working

[12:37]  
It basically passes through all of the methods to the target (edited)

[12:38]  
We use it as we have to intercept certain methods that get’s called from the `UICollectionView` (edited)

simo [12:39 PM]  
yeah, I’m not there yet knowledge-wise but I’ll look. perhaps I’ll find something of interest in ASDK code base

maicki [12:39 PM]  
Look into `ASDelegateProxy` and `ASCollectionViewProxy` (edited)

[12:39]  
There you can see which methods we intercept

[12:40]  
Just for further undestanding we intercept, e.g. `scrollViewWillEndDragging:withVelocity:targetContentOffset:` to support batch fetching (edited)

simo [12:41 PM]  
I see. does subclassing not allow you to override the behavior? not the case with delegates?

maicki [12:42 PM]  
No as we want to intercept certain `delegate` / `dataSource`, subclassing will not help in this case (edited)

[12:43]  
@simo If you would like to see a change for that behavior I encourage you to open a Github Issue. (edited)

simo [12:44 PM]  
Yeah, it would really be nice for you guys to think of ways to support it, perhaps through configuration.

maicki [12:44 PM]  
As said we are open for discussion, just open a Github Issue with some examples or even better some ideas around how we could support it
```

Adlai-Holler:
Resolves #2245 and improves interop with frameworks like TLYShyNavBar which set the delegate of the table/collection view.

We do not guarantee that we support these other frameworks, but in cases like this where the cost is low, we can help.

### DIFF
--- a/examples/CustomCollectionView-Swift/Sample/MosaicCollectionViewLayout.swift
+++ b/examples/CustomCollectionView-Swift/Sample/MosaicCollectionViewLayout.swift
@@ -229,17 +229,6 @@ class MosaicCollectionViewLayoutInspector: NSObject, ASCollectionViewLayoutInspe
   }
   
   /**
-   * Asks the inspector for the number of supplementary sections in the collection view for the given kind.
-   */
-  func collectionView(_ collectionView: ASCollectionView, numberOfSectionsForSupplementaryNodeOfKind kind: String) -> UInt {
-    if (kind == UICollectionElementKindSectionHeader) {
-      return UInt((collectionView.dataSource?.numberOfSections!(in: collectionView))!)
-    } else {
-      return 0
-    }
-  }
-  
-  /**
    * Asks the inspector for the number of supplementary views for the given kind in the specified section.
    */
   func collectionView(_ collectionView: ASCollectionView, supplementaryNodesOfKind kind: String, inSection section: UInt) -> UInt {
@@ -251,6 +240,6 @@ class MosaicCollectionViewLayoutInspector: NSObject, ASCollectionViewLayoutInspe
   }
   
   func scrollableDirections() -> ASScrollDirection {
-    return ASScrollDirectionVerticalDirections;
+    return ASScrollDirectionVerticalDirections
   }
 }
--- a/AsyncDisplayKit/ASCollectionView.h
+++ b/AsyncDisplayKit/ASCollectionView.h
@@ -125,6 +125,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface ASCollectionView (Deprecated)
 
+@property (nonatomic, weak) id<UICollectionViewDelegate> delegate AS_UNAVAILABLE("Use ASCollectionNode.delegate instead.");
+@property (nonatomic, weak) id<UICollectionViewDataSource> dataSource AS_UNAVAILABLE("Use ASCollectionNode.dataSource instead.");
+
 /**
  * The object that acts as the asynchronous delegate of the collection view
  *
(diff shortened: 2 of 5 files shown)


examples/CustomCollectionView-Swift/Sample/MosaicCollectionViewLayout.swift
```diff
@@ -229,17 +229,6 @@ class MosaicCollectionViewLayoutInspector: NSObject, ASCollectionViewLayoutInspe
   }
   
   /**
-   * Asks the inspector for the number of supplementary sections in the collection view for the given kind.
-   */
-  func collectionView(_ collectionView: ASCollectionView, numberOfSectionsForSupplementaryNodeOfKind kind: String) -> UInt {
```
Adlai-Holler: This protocol method is deprecated and no longer called, and it calls `ASCollectionView.dataSource` which is now deprecated.

PR #3021, opened 2017-02-13
[ASViewController] Support optional node

### LINKED ISSUE #2760: [ASViewController] Add unit test or sample app that uses an ASVC without self.node, using .view manually.

maicki:
Thanks to @nguyenhuy for the initial work! Unfortunately there were a couple of bigger changes since the last commit of the initial PR so I rebased and also cleaned it up a bit. Will replace #2810

This PR makes ASViewController's node optional. This allows developers to use ASViewController like a normal UIViewController. More importantly, it can be used as a base class for view controllers among which some use a node hierarchy and some don't.

Resolve #2760.

### DIFF
--- a/AsyncDisplayKit/ASViewController.h
+++ b/AsyncDisplayKit/ASViewController.h
@@ -41,10 +41,14 @@ typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitWindowSizeBlock)(C
  */
 - (instancetype)initWithNode:(DisplayNodeType)node;
 
+NS_ASSUME_NONNULL_END
+
 /**
  * @return node Returns the ASDisplayNode which provides the backing view to the view controller.
  */
-@property (nonatomic, strong, readonly) DisplayNodeType node;
+@property (nonatomic, strong, readonly, null_unspecified) DisplayNodeType node;
+
+NS_ASSUME_NONNULL_BEGIN
 
 /**
  * Set this block to customize the ASDisplayTraits returned when the VC transitions to the given traitCollection.
--- a/AsyncDisplayKit/ASViewController.h
+++ b/AsyncDisplayKit/ASViewController.h
@@ -21,20 +21,25 @@ NS_ASSUME_NONNULL_BEGIN
 typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitCollectionBlock)(UITraitCollection *traitCollection);
 typedef ASTraitCollection * _Nonnull (^ASDisplayTraitsForTraitWindowSizeBlock)(CGSize windowSize);
 
+/**
+ * ASViewController allows you to have a completely node backed hierarchy. It automatically
+ * handles @c ASVisibilityDepth, automatic range mode and propogating @c ASDisplayTraits to contained nodes.
+ *
+ * You can opt-out of node backed hierarchy and use it like a normal UIViewController.
+ * More importantly, you can use it as a base class for all of your view controllers among which some use a node hierarchy and some don't.
+ * See examples/ASDKgram project for actual implementation.
+ */
 @interface ASViewController<__covariant DisplayNodeType : ASDisplayNode *> : UIViewController <ASVisibilityDepth>
 
 /**
- * ASViewController Designated initializer.
- *
- * @discussion ASViewController allows you to have a completely node backed heirarchy. It automatically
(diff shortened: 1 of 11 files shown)


AsyncDisplayKit/ASViewController.h
```diff
@@ -21,20 +21,25 @@ NS_ASSUME_NONNULL_BEGIN
...
  * @param node An ASDisplayNode which will provide the root view (self.view)
  * @return An ASViewController instance whose root view will be backed by the provided ASDisplayNode.
  *
  * @see ASVisibilityDepth
  */
-- (instancetype)initWithNode:(DisplayNodeType)node NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithNode:(DisplayNodeType)node;
 
 /**
  * @return node Returns the ASDisplayNode which provides the backing view to the view controller.
```
Adlai-Holler: We should mark the `_node` property `nullable` instead of using a dummy node. This will be breaking API for Swift, but it's cleaner and avoids the need for `_providedNode`.

AsyncDisplayKit/ASViewController.mm
```diff
@@ -82,13 +101,16 @@ - (void)dealloc
...
   // will have a greater performance benefit than the impact of this transient view.
   [super loadView];
+  
+  if (!_nodeProvided) { return; }
+  
+  ASDisplayNodeAssertTrue(!_node.layerBacked);
+  
   UIView *view = self.view;
   CGRect frame = view.frame;
   UIViewAutoresizing autoresizingMask = view.autoresizingMask;
```
Adlai-Holler: So if you call `-setNode:` after the view is loaded, won't we be in chaos? `self.node` will get updated, but the node won't get hosted in the view hierarchy.

AsyncDisplayKit/ASViewController.h
maicki: I'm not sold on having a nil node within the view controller. If we would go ahead with that, we always would have to check for if the node is nil on every call within `ASViewController` and also all subclasses need to consider that (Swift needs to unwrap). I know you can call methods on nil within `Objective-C` ...

We will have an overhead with having around a temporary node that's true ...

AsyncDisplayKit/ASViewController.mm
maicki: The node property is `readonly`. It should be not possible to set a node after a view controller is initialized.

AsyncDisplayKit/ASViewController.h
Adlai-Holler: I'm not concerned with overhead – it's simply incorrect to have a node around that you never touch or add to the window hierarchy (because we wrap it in this `_didProvideNode` flag). That's not any safer than having a nil node.

AsyncDisplayKit/ASViewController.mm
Adlai-Holler: Ah OK I had assumed this diff made it `readwrite` to appease the community members who wanted to use nodes but didn't want to use `initWithNode:`. Cool.

Adlai-Holler:
(Mirrored from our discussion on Slack) The move here is to use an implicitly-unwrapped optional for the `node` property. If they don't give us one, it's `nil`. If they unwrap `vc.node` in Swift (if they assume an ASViewController has a node) they get a crash. These crashes should be pretty easy to debug, and very easy to prevent (`guard let node = vc.node else { assertionFailure("Expected vc to have a node"); return ; }`). 

This way everyone can tell whether an `ASViewController` has a node or not, but Swift users won't have to constantly unwrap their node properties manually.

maicki:
Pushed a new change that changes the nil node behavior based on @Adlai-Holler and my Slack conversation.

maicki:
Thanks @Adlai-Holler and @nguyenhuy for working together on that.

Issue #2894, opened 2017-01-13
ASTextNode sizing issue for CJK language(caused by TextKit bug, and there's a quick fix)


yxztj:
This should be a TextKit bug, based on which ASTextNode is built.

To repro:
1. add a simple UILabel and a ASTextNode
2. fill them with the same attributedString
3. overlap each other and see if them coincide, which they should

## Repro
(I've created a [sample project](https://github.com/yxztj/TextKitSizingIssue), render the string is UILabel in blue and ASTextNode in black. Each view has a border with the same color as their content)
Everything looks good for English:
<img width="132" alt="screen shot 2017-01-13 at 9 42 49 am" src="https://cloud.githubusercontent.com/assets/3334458/21915402/af5e9c38-d974-11e6-9577-baf14ccc521c.png">
The UILabel and ASTextNode coincide completely. Since the black ASTextNode is drawn upon the blue UILabel, we can barely see the latter one.

When it comes to Chinese, we can see that the 2 strings don't fully coincide. The blue one goes a little bit down, and has a larger frame than black one.
<img width="147" alt="screen shot 2017-01-13 at 9 39 44 am" src="https://cloud.githubusercontent.com/assets/3334458/21915571/a597beb8-d975-11e6-9868-5a35ad9397c0.png">

The same goes for Korean:
<img width="178" alt="screen shot 2017-01-13 at 9 39 24 am" src="https://cloud.githubusercontent.com/assets/3334458/21915577/b180f37a-d975-11e6-8f39-64e8807ba04e.png">

The black frame is 6pt short in height if you print in console, while the height of blue frame exactly matches lineHeight property of the font. So the black frame is wrong, seems it incorrectly matches the pointSize of the font.
For English string, both blue and black frame matches lineHeight of the font.

## Quick Fix
Fortunately if you add a "useless" line of code right after [_textStorage addLayoutManager] in ASTextKitContext.mm, all issues go away:
```
[_textStorage setAttributedString:attributedString];
```
_testStorage should have just been initialized with the attributedString provided, so setting the same string again should be useless. But it does magically fix the sizing issue:
<img width="143" alt="screen shot 2017-01-13 at 9 55 27 am" src="https://cloud.githubusercontent.com/assets/3334458/21915676/73e09f38-d976-11e6-923b-c364d73d6b75.png">

Sadly Apple has been ignoring my bug report for months.

Also the issue looks very similar to #1013 , hope the fix also applies to that.
[...]

PR #3026, opened 2017-02-14
[TextKit] Fix text layout issue for CJK laungages

yxztj:
A quick fix for CJK language layout issue, which is supposed to bring no side effect. Has been tested in our production for one month.
See https://github.com/facebook/AsyncDisplayKit/issues/2894

### DIFF
--- a/AsyncDisplayKit/TextKit/ASTextKitContext.mm
+++ b/AsyncDisplayKit/TextKit/ASTextKitContext.mm
@@ -39,12 +39,13 @@
     __instanceLock__ = std::make_shared<ASDN::Mutex>();
     
     // Create the TextKit component stack with our default configuration.
-    _textStorage = (attributedString ? [[NSTextStorage alloc] initWithAttributedString:attributedString] : [[NSTextStorage alloc] init]);
+    
+    _textStorage = [[NSTextStorage alloc] init];
     _layoutManager = [[ASLayoutManager alloc] init];
     _layoutManager.usesFontLeading = NO;
     [_textStorage addLayoutManager:_layoutManager];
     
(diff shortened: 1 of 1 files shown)


yxztj:
Just signed the CLA.

Adlai-Holler:
@yxztj Thanks for the diff! As a potentially smaller change, is it sufficient to instead call one of these:

- `ensureAttributesAreFixedInRange:`
- `fixAttributesInRange:`
- `invalidateAttributesInRange:`

If not, it would be great to see profiling results from this change if possible. Text nodes are some of the most common nodes that are used, and changing the attributed string of `NSTextStorage` could be a quite expensive operation.

yxztj:
@Adlai-Holler Thanks for the suggestions. I tried the 3 methods you suggested but none of them worked..

After some profiling(copy and run the related code, from _textStorage init to setAttributedString 30000 times), setting attributedString both in init and later is doubling the execution time than before.

So I changed the code: always init _textStorage with no parameter, and set attributedString later. Running the profiling code this time I can see the execution time is very close to before.

appleguy:
@yxztj awesome, thanks for investigating this issue and then also guiding your work via profiling!  An allstar contributor :), really appreciate how you've helped out over time.

Both @Adlai-Holler and @maicki are the de-facto experts on ASTextNode these days, so I'll leave it to them to make the call on safety of taking this change.  However, we're all interested in seeing improvements to a few nagging correctness issues that still affect the component and so will seriously consider any PRs in this area.

One last suggestion: consider adding this to examples_extra/TextStressTest?  Not a blocker for this PR, just a suggested followup.  I've been trying to collect our known ASTextNode bugs there and then the fixes so we can run the app to check them.  Of course it would be better to add snapshot tests for all of these, but either one would be a great step.

Adlai-Holler:
Great work @yxztj, thanks!

PR #3028, opened 2017-02-14
Fix Pager Node Issues

### LINKED ISSUE #2974: Pager Flow Layout is Not Aware of Programmatic Scrolls
`ASPagerFlowLayout` has a `currentIndexPath` property that it tries to keep up-to-date using the `targetContentOffsetForProposedContentOffset:` methods from `UICollectionViewLayout`.

The pager layout uses this index path to reset the `contentOffset` during animated bounds changes (rotation) so we stay looking at the same page.

Unfortunately, this method is not called for programmatic scrolls – only for updates, user scrolls, and animated bounds changes. 

In Pinterest, we have a screen where the header recedes when you scroll. If you programmatically change to a different page, and then trigger the header to recede, the pager flow layout detects an animated bounds change and scrolls over to its "currentIndexPath" which is now incorrect because we just programmatically scrolled away.

Two steps:
- Add a `scrollViewDidScroll:` call to the pager flow layout, and drive all `_currentIndexPath` updates from this call. This is the only hook we _always_ get from UIScrollView on scroll.
- Ignore animated bounds changes that only affect height, since all pagers scroll horizontally.

Adlai-Holler:
Resolves #2974 and other major pager node rotation/content inset issues.

- Deprecates `ASCollectionView.zeroContentInsets`.
  - This flag was added as a hack during development of `ASPagerNode` to "undo" the effects of `UIViewController.automaticallyAdjustsScrollViewInsets`.
  - The flag adds `contentInsets = .zero` into `layoutSubviews`, after `[super layoutSubviews]`
  - This means our `contentInsets` is thrashing around constantly.
  - The view controller _also_ sets `contentOffset` during its insets adjustment, which we didn't account for.
- Adds a flag `ASPagerNode.allowsAutomaticInsetsAdjustment`.
  - Defaults to `NO` since you virtually NEVER want UIVC to adjust pager node insets.
  - If `NO`, finds the owning view controller of the pager and sets `automaticallyAdjustsScrollViewInsets=NO`, and logs a message.
  - In the future, we may remove the automatic setting of this property and just log the message, so users retain control of their view controllers' flags.
  - If they for some reason _want_ automatic insets adjustment from their VC, they can set `allowsAutomaticInsetsAdjustment=YES` on their pager node.
- Adds a public category method `-[UIResponder asdk_associatedViewController]` that safely travels up the responder chain and finds the nearest view controller. This is used above, and it may very well be handy for users especially when debugging. I also put the view controller in the node's `-debugDescription` for fun. I could have left it out to keep this diff smaller.
- Adds a private class `ASResponderChainEnumerator` that is used in implementing above, and may come in handy in other places later.
- Replaces `ASPagerFlowLayout.currentIndexPath` with `ASPagerFlowLayout.currentCellNode` so that we can stay pointing at the same node across inserts/deletes.
- Makes `ASPagerFlowLayout` use `-shouldInvalidateLayoutForBoundsChange:` – which is like the `UICollectionViewLayout` version of `scrollViewDidScroll:` – to update its `currentIndexPath`, rather than using `targetContentOffsetForProposedContentOffset:withScrollingVelocity:`, which is not called during programmatic scrolls.
- When a collection view's bounds change, removes the `performBatchUpdates:` that we placed around `_relayoutAllNodes`. The stated reason for the batch updates was to get the relayout into the data controller's transaction pipeline, but it already _goes_ in that pipeline internally, and we don't need to submit an empty update to trigger a layout invalidation, since we're already manually invalidating the layout on the next line. This avoids triggering `UICollectionViewLayout.targetContentOffsetForProposedContentOffset:` needlessly, which happens on every collection view update, even empty ones.
- Fixes indentation throughout `ASPagerFlowLayout.m`.

In short:
Before:

After:

Note the flashing may look ugly here, but it's intentionally added by flow layout because in a real app, the layout transition is usually going to be bad.

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -26,6 +26,7 @@
 #import <AsyncDisplayKit/_ASCollectionViewCell.h>
 #import <AsyncDisplayKit/_ASDisplayLayer.h>
 #import <AsyncDisplayKit/ASCollectionViewLayoutFacilitatorProtocol.h>
+#import <AsyncDisplayKit/ASPagerNode.h>
 #import <AsyncDisplayKit/ASSectionContext.h>
 #import <AsyncDisplayKit/ASCollectionView+Undeprecated.h>
 #import <AsyncDisplayKit/_ASHierarchyChangeSet.h>
@@ -93,6 +94,8 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   NSMutableSet *_registeredSupplementaryKinds;
   
   CGPoint _deceleratingVelocity;
+
+  BOOL _zeroContentInsets;
   
   ASCollectionViewInvalidationStyle _nextLayoutInvalidationStyle;
   
@@ -582,6 +585,16 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
 - (ASRangeTuningParameters)tuningParametersForRangeMode:(ASLayoutRangeMode)rangeMode rangeType:(ASLayoutRangeType)rangeType
 {
   return [_rangeController tuningParametersForRangeMode:rangeMode rangeType:rangeType];
+}
+
+- (void)setZeroContentInsets:(BOOL)zeroContentInsets
+{
+  _zeroContentInsets = zeroContentInsets;
+}
+
+- (BOOL)zeroContentInsets
+{
+  return _zeroContentInsets;
 }
 
 - (CGSize)calculatedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath
@@ -1911,19 +1924,11 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
     BOOL changedInNonScrollingDirection = (fixedHorizontally && newBounds.size.width != lastUsedSize.width) || (fixedVertically && newBounds.size.height != lastUsedSize.height);
 
     if (changedInNonScrollingDirection) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-      // This actually doesn't perform an animation, but prevents the transaction block from being processed in the
-      // data controller's prevent animation block that would interrupt an interrupted relayout happening in an animation block
-      // ie. ASCollectionView bounds change on rotation or multi-tasking split view resize.
-      [self performBatchAnimated:YES updates:^{
-        [_dataController relayoutAllNodes];
-      } completion:nil];
+      [_dataController relayoutAllNodes];
+      [_dataController waitUntilAllUpdatesAreCommitted];
       // We need to ensure the size requery is done before we update our layout.
-      [self waitUntilAllUpdatesAreCommitted];
       [self.collectionViewLayout invalidateLayout];
     }
-#pragma clang diagnostic pop
   }
 }
 
--- a/AsyncDisplayKit/ASPagerFlowLayout.m
+++ b/AsyncDisplayKit/ASPagerFlowLayout.m
@@ -11,62 +11,51 @@
 //
 
 #import <AsyncDisplayKit/ASPagerFlowLayout.h>
+#import <AsyncDisplayKit/ASCellNode.h>
+#import <AsyncDisplayKit/ASCollectionView.h>
 
 @interface ASPagerFlowLayout () {
-  BOOL _didRotate;
-  CGRect _cachedCollectionViewBounds;
-  NSIndexPath *_currentIndexPath;
+  __weak ASCellNode *_currentCellNode;
 }
 
 @end
 
 @implementation ASPagerFlowLayout
 
-- (CGPoint)targetContentOffsetForProposedContentOffset:(CGPoint)proposedContentOffset withScrollingVelocity:(CGPoint)velocity
+- (ASCollectionView *)asCollectionView
 {
-  NSInteger currentPage = ceil(proposedContentOffset.x / self.collectionView.bounds.size.width);
-  _currentIndexPath = [NSIndexPath indexPathForItem:currentPage inSection:0];
-  
-  return [super targetContentOffsetForProposedContentOffset:proposedContentOffset withScrollingVelocity:velocity];
+  // Dynamic cast is too slow and not worth it.
+  return (ASCollectionView *)self.collectionView;
 }
 
-
-- (void)prepareForAnimatedBoundsChange:(CGRect)oldBounds
+- (void)prepareLayout
 {
-  // Cache the current page if a rotation did happen. This happens before the rotation animation
-  // is occuring and the bounds changed so we use this as an opportunity to cache the current index path
-  if (_cachedCollectionViewBounds.size.width != self.collectionView.bounds.size.width) {
-      _cachedCollectionViewBounds = self.collectionView.bounds;
-        
-    // Figurring out current page based on the old bounds visible space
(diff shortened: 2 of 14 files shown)


Adlai-Holler:
I know this diff is huge, and I attempted to decompose it into smaller updates but to get the correct pager behavior, all of these changes were necessary (excluding e.g. adding the view controller into node debugDescription which was just for fun and seems very low-risk).

AsyncDisplayKit/ASCollectionView.mm
```diff
@@ -1911,19 +1924,11 @@ - (void)layer:(CALayer *)layer didChangeBoundsWithOldValue:(CGRect)oldBounds new
...
     if (changedInNonScrollingDirection) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-      // This actually doesn't perform an animation, but prevents the transaction block from being processed in the
-      // data controller's prevent animation block that would interrupt an interrupted relayout happening in an animation block
-      // ie. ASCollectionView bounds change on rotation or multi-tasking split view resize.
-      [self performBatchAnimated:YES updates:^{
-        [_dataController relayoutAllNodes];
-      } completion:nil];
+      [_dataController relayoutAllNodes];
```
garrettmoon: So much clearer.

AsyncDisplayKit/ASPagerFlowLayout.m
```diff
@@ -75,19 +56,49 @@ - (CGPoint)_targetContentOffsetForItemAtIndexPath:(NSIndexPath *)indexPath propo
...
+  // isn't sufficient here! It's broken!
+  NSArray *animKeys = self.collectionView.layer.animationKeys;
+  for (NSString *key in animKeys) {
+    if ([key hasPrefix:@"bounds"]) {
+      return;
+    }
+  }
+  
+  CGRect bounds = self.collectionView.bounds;
+  CGRect rect = CGRectMake(CGRectGetMidX(bounds), CGRectGetMidY(bounds), 1, 1);
```
garrettmoon: Probably doesn't matter, but should this possibly be `CGRectMake(CGRectGetMidX(bounds) - 0.5, CGRectGetMidY(bounds) - 0.5, 1, 1);`

garrettmoon: I assume it doesn't because the pager should have cells that fill the view?

AsyncDisplayKit/ASPagerNode.h
```diff
@@ -120,6 +120,24 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *
+ * @discussion ASPagerNode should usually not have its content insets automatically adjusted
+ * because it scrolls horizontally, and flow layout will log errors because the pages
+ * do not fit between the top & bottom insets of the collection view.
+ *
+ * The default value is NO, which means that ASPagerNode expects that its view controller will
+ * have automaticallyAdjustsScrollViewInsets=NO.
+ *
+ * If this property is NO, but your view controller has automaticallyAdjustsScrollViewInsets=YES,
+ * the pager node will set the property to NO and log a warning message. In the future,
```
garrettmoon: This isn't 100% clear to me. It will set the property on the view controller to NO? Will it **leave** the pageNode property NO? Or, will it actually set the pagerNode property to YES?

AsyncDisplayKit/ASPagerNode.m
```diff
@@ -211,4 +208,21 @@ - (void)proxyTargetHasDeallocated:(ASDelegateProxy *)proxy
   [self setDelegate:nil];
 }
 
+- (void)didEnterVisibleState
+{
+	[super didEnterVisibleState];
+
+	// Check that our view controller does not automatically set our content insets
+	// It would be better to have a -didEnterHierarchy hook to put this in, but
```
garrettmoon: Will `willEnterHierarchy` not suffice?

AsyncDisplayKit/UIResponder+AsyncDisplayKit.m
```diff
@@ -0,0 +1,31 @@
...
+#import <AsyncDisplayKit/ASResponderChainEnumerator.h>
+
+@implementation UIResponder (AsyncDisplayKit)
+
+- (__kindof UIViewController *)asdk_associatedViewController
+{
+  ASDisplayNodeAssertMainThread();
+  
+  for (UIResponder *responder in [self asdk_responderChainEnumerator]) {
+    UIViewController *vc = ASDynamicCast(responder, UIViewController);
```
garrettmoon: Is this documented behavior? A responder will have it's viewController in the responder chain prior to any other possible view controllers at all times? If not, maybe a comment indicating it's been tested to be true in practice? If so a link would be awesome :)

examples/PagerNode/Sample/PageNode.m
```diff
@@ -21,9 +21,9 @@
 
 @implementation PageNode
 
-- (ASLayout *)calculateLayoutThatFits:(ASSizeRange)constrainedSize
+- (CGSize)calculateSizeThatFits:(CGSize)constrainedSize
```
garrettmoon: `@note Subclasses that override are committed to manual layout. Therefore, -layout: must be overriden to layout all subnodes or subviews.`

Not needed in this case?

AsyncDisplayKit/ASPagerFlowLayout.m
Adlai-Holler: Right – "government work" level precision is enough here 🏗

AsyncDisplayKit/ASPagerNode.h
Adlai-Holler: It will leave the pager property (as NO) but it'll automatically set NO on the view controller (for backwards compatibility) and log a message that it did so. In the future, we want users to be responsible for configuring their own view controllers, so we'll just log the message.

AsyncDisplayKit/ASPagerNode.m
Adlai-Holler: No unfortunately, because we're not yet in the hierarchy =(

AsyncDisplayKit/UIResponder+AsyncDisplayKit.m
Adlai-Holler: Hmm I'll leave the link out because the best I could do is a general link to an overview of the responder chain. But responder chain always goes `aView -> aView.superview -> aView.superview.viewController -> aView.superview.viewController.view.superview` etc, which is documented.

examples/PagerNode/Sample/PageNode.m
Adlai-Holler: Right because we have no subnodes.

AsyncDisplayKit/ASPagerNode.m
garrettmoon: Interesting, the isInHierarchy flag is set before willEnterHierarchy is called…

AsyncDisplayKit/UIResponder+AsyncDisplayKit.m
garrettmoon: 👍

AsyncDisplayKit/ASPagerNode.h
garrettmoon: Maybe just a slight adjustment to your comment then: `If this property is NO, but your view controller has automaticallyAdjustsScrollViewInsets=YES, the pager node will set the property **on the view controller** to NO and log a warning message.`

garrettmoon: But without the stars :)

PR #3031, opened 2017-02-15
Don't Skip Remeasurement On First Layout Pass

Adlai-Holler:
Fixes the rest of #2955 

The logic behind this flag is pretty nonsensical. The only reason we would ever get `layoutSubviews` before our size is "final" is if someone calls `layoutIfNeeded` specifically on us, which they might do I guess.

But in the normal case, we get `layoutSubviews` after our superview gets `layoutSubviews` and sets our frame to e.g. `(375, 667)` or whatever our final frame is. Silly silly silly!

### DIFF
--- a/AsyncDisplayKit/ASCollectionView.mm
+++ b/AsyncDisplayKit/ASCollectionView.mm
@@ -89,7 +89,6 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
   ASBatchContext *_batchContext;
   
   CGSize _lastBoundsSizeUsedForMeasuringNodes;
-  BOOL _ignoreNextBoundsSizeChangeForMeasuringNodes;
   
   NSMutableSet *_registeredSupplementaryKinds;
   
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Note: I removed `testRelayoutAllNodesWithZeroSizeInitially`.

The justification is, this test is kind of misleading… in the normal flow:

> Have initial size `(0,0)`
> Core Animation `layoutSubviews` down from the window to your superview
> Superview gives you a real size `(375, 667)`
> You get `layoutSubviews`
> We call `relayoutAllNodes` but there are no nodes yet
> We call `[super layoutSubviews]`
> Table view calls `[self reloadData]`
> We start loading the initial data

The unit test failed because `relayoutAllNodes` was called, but there were no nodes because **the initial data load happens after that**. Now take the case where you manually call `reloadData` before first `layoutSubviews`:

> Have initial size `(0,0)`
> Someone calls `reloadData`
> We allocate but do not measure the nodes.
> Core Animation `layoutSubviews` down from the window to your superview
> Superview gives you a real size `(375, 667)`
> You get `layoutSubviews`
> We call `relayoutAllNodes` and remeasure all nodes
> We call `[super layoutSubviews]`

PR #3017, opened 2017-02-11
[ASDataController] Refactor ASDataController

nguyenhuy:
## Author: Thanh Huy Nguyen

This PR has 2 main goals:
- Simplify `ASDataController` and overhaul the way it processes change sets.
  - `ASDataController` now processes a change set in 3 steps:
    - Synchronizes its internal states with data source on main thread.
    - Switches to a background queue to layout new nodes.
    - Switches back to main to forward the change set to UIKit.
  - Each step consumes the **whole** change set before switching to another thread for the next step. The data shared between these steps are immutable. This helps to guarantee thread safety and data consistency.
  - Each step may interpret a change set differently. The first step decomposes reloads into deletes and inserts, while the last step forwards the change set as-is. As a result, the implementation of `-reloadData` is much simpler. We finally can destroy everything at once and repopulate the whole data set. And flashes should be gone now (#2536). Furthermore, it's much easier to support native moves (#698, #2868) and reloads (#469).
- Merge `ASCollectionDataController` into `ASDataController`. 
  - `ASCollectionDataController` subclasses `ASDataController` to support supplementary views. In order to achieve that, it requires many hooks during the update process. Those hooks add complexity to `ASDataController`. Moreover, `ASCollectionDataController` doesn't handle supplementary node contexts correctly. It doesn't persistently store those contexts which is important for upcoming features (i.e node virtualization and measurement range).
  - Even after #1629 and #1773, repopulating supplementary nodes still doesn't work perfectly. After merging the 2 classes, it's much easier to rewrite this part.

Remaining tasks:
- [x] Fix failing unit tests
- [x] More testing on various projects
- [x] Revisit TODOs
- [x] Support reloads (#469) natively
- [ ] Support moves (#698, #2868) natively
- [ ] Revisit `-[ASDataController relayoutAllNodes]` and its callees.
- [ ] Consider changing `_nodeContexts` to be a two dimension array of dictionaries (instead of a dictionary of two dimensional arrays).

This PR is still a WIP and contains many TODOs. However, I'd appreciate any early feedbacks and help for testing. Thanks!

Ticket: https://github.com/TextureGroup/Texture/issues/186

PR #3033, opened 2017-02-15
Majorly Improve automaticallyAdjustsContentOffset

### LINKED ISSUE #2908: ASTableview automaticallyAdjustsContentOffset incorrect when inserting sections with headers/footers
Hello. I noticed that `automaticallyAdjustsContentOffset` on ASTableView is not working correctly, when inserting sections with headers/footers.

I have created a test project for this. https://github.com/ArmandsB/AsyncDisplayKitTest
Basically, when inserting new sections at top, which have header/footers - offset is being calculated incorrectly. When manually add that extra offset, everything is fine.

Adlai-Holler:
- Only do it for non-animated updates. For animated updates, as I mention in the comments, we can't synchronize with table view's update animation and you see an ugly wave.
  - This is **breaking API** but we need to do it. No user would ever prefer to see this effect.
- Switch our tracking system to be much simpler and to account for section headers/footers/etc.
- Add a unit test for content offset adjusting

Old system:
> Update about to start
> Save first visible index path
> Update inserted rows
> Sum up all the nodes' `calculatedSize.height`
> Update deleted rows
> Subtract all the nodes' `calculatedSize.height`
> Update about to end
> Find the row at the saved index path (!!! may not be correct)
> Add the summed-up value to contentOffset.y

New system:
> Update about to start
> Save the first visible cell node and its y-offset from the top of viewport.
> Update about to end
> Find the index path for that cell node, and scroll so that its y-offset is what it was before

This removes a ton of logic where we sum up `cellNode.calculatedSize` and makes sure we handle things like section headers/footers etc.

I also removed some now-unused arguments from `ASDataControllerDelegate` and `ASRangeControllerDelegate`.

Resolves #2908

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -1432,7 +1432,7 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   }
 }
 
-- (void)rangeController:(ASRangeController *)rangeController didInsertNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
+- (void)rangeController:(ASRangeController *)rangeController didInsertItemsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
 {
   ASDisplayNodeAssertMainThread();
   LOG(@"UITableView insertRows:%ld rows", indexPaths.count);
@@ -1454,7 +1454,7 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   });
 }
 
-- (void)rangeController:(ASRangeController *)rangeController didDeleteNodes:(NSArray *)nodes atIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
+- (void)rangeController:(ASRangeController *)rangeController didDeleteItemsAtIndexPaths:(NSArray *)indexPaths withAnimationOptions:(ASDataControllerAnimationOptions)animationOptions
 {
   ASDisplayNodeAssertMainThread();
   LOG(@"UITableView deleteRows:%ld rows", indexPaths.count);
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -128,8 +128,10 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 
   NSIndexPath *_pendingVisibleIndexPath;
 
-  NSIndexPath *_contentOffsetAdjustmentTopVisibleRow;
-  CGFloat _contentOffsetAdjustment;
+  // The top cell node that was visible before the update.
+  __weak ASCellNode *_contentOffsetAdjustmentTopVisibleNode;
+  // The y-offset of the top visible row's origin before the update.
+  CGFloat _contentOffsetAdjustmentTopVisibleNodeOffset;
   
   CGPoint _deceleratingVelocity;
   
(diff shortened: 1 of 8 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -110,21 +105,6 @@ - (instancetype)init
...
-    return;
-  }
-  
-  _delegate = delegate;
-  
-  // Interrogate our delegate to understand its capabilities, optimizing away expensive respondsToSelector: calls later.
-  _delegateDidInsertNodes     = [_delegate respondsToSelector:@selector(dataController:didInsertNodes:atIndexPaths:withAnimationOptions:)];
-  _delegateDidDeleteNodes     = [_delegate respondsToSelector:@selector(dataController:didDeleteNodes:atIndexPaths:withAnimationOptions:)];
-  _delegateDidInsertSections  = [_delegate respondsToSelector:@selector(dataController:didInsertSections:atIndexSet:withAnimationOptions:)];
-  _delegateDidDeleteSections  = [_delegate respondsToSelector:@selector(dataController:didDeleteSectionsAtIndexSet:withAnimationOptions:)];
```
Adlai-Holler: All this can be removed because these methods are required i.e. enforced by compiler. Sorry @nguyenhuy if this requires a rebase but it should be simple if so!

Adlai-Holler:
Jenkins, test this please.
Looks like a (new) spurious failure… 

```
/Users/engineering/jenkins_root/workspace/AsyncDisplayKit-PRs/examples/ASDKTube/Sample/WindowWithStatusBarUnderlay.h:20:9: could not acquire lock file for module 'UIKit': failed to create link /Users/engineering/ASDKDerivedData/ModuleCache/1OROSK0MU46A0/UIKit-1V5UHAPTOD24G.pcm.lock to /Users/engineering/ASDKDerivedData/ModuleCache/1OROSK0MU46A0/UIKit-1V5UHAPTOD24G.pcm.lock-229256f2: No such file or directory
```

AsyncDisplayKit/ASTableView.mm
```diff
@@ -778,53 +780,41 @@ - (void)moveRowAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)n
...
-    NSIndexPath *indexPath = indexPaths[index];
-    if ([indexPath compare:top] <= 0) { // if this row is before or equal to the topmost visible row, make adjustments...
-      ASCellNode *cellNode = nodes[index];
-      adjustment += cellNode.calculatedSize.height * dir;
-      if (indexPath.section == top.section) {
-        top = [NSIndexPath indexPathForRow:top.row+dir inSection:top.section];
-      }
-    }
+  
+  // We can't do this if we didn't have a top visible row before.
```
nguyenhuy: It's great that this edge case is handled. 

Non-blocking feedback: I'm wondering if we should store 1 or maybe even 2 more nodes? That way we would rarely give up because all of the anchor nodes were gone. The last visible node and the one in the middle are good candidates for this purpose.

AsyncDisplayKit/ASTableView.mm
```diff
@@ -1462,13 +1452,9 @@ - (void)rangeController:(ASRangeController *)rangeController didInsertNodes:(NSA
     }
     [self _scheduleCheckForBatchFetchingForNumberOfChanges:indexPaths.count];
   });
-
-  if (_automaticallyAdjustsContentOffset) {
-    [self adjustContentOffsetWithNodes:nodes atIndexPaths:indexPaths inserting:YES];
-  }
```
nguyenhuy: I guess these lines are deleted based on the assumption that edits are always done in batches and so offset adjustment can be solely done in `-rangeController: didEndUpdatesAnimated:completion:`?

I honestly think we should embrace this "always in batches" strategy to the point that there should be only one delegate method `-rangeController:didUpdateWithChangeSet:`. It will allow us to greatly simplify ASCollectionView and ASTableView. Currently those classes need many internal states to deal with the possibility that updates can be done separately.

AsyncDisplayKit/Details/ASDataController.mm
nguyenhuy: I also deleted this in [#3017](https://github.com/facebook/AsyncDisplayKit/pull/3017/files#diff-c03b807fdd44839fd88691746d9dcc27L120). I will rebase with joy if necessary!

AsyncDisplayKit/ASTableView.mm
Adlai-Holler: Good call! For the time being, let's leave it as-is to reduce complexity since cases where they delete the anchor row will be somewhat rare, and I have a suspicion that not many people are using this flag.

garrettmoon:
retest this please

garrettmoon:
retest this please

PR #3035, opened 2017-02-15
ASTextFieldNode

kyleshank:
ASTextFieldNode is a convenient wrapper for UITextField.

ASEditableTextNode works for many situations but its behavior for single line text input isn't ideal.  Secure text entry also doesn't appear to work on latest for ASEditableTextNode and UITextView in the latest version of iOS.

An example is included to demonstrate usage.


Adlai-Holler:
Hi @kyleshank , thanks for opening this diff! In order to provide full and long-lasting support for a text field node, which is super valuable, we need to do a ton of work including introducing a new class of pending state object and build support for bridging fields like `font` for reads _and_ writes, and avoid duplicating the portions of code that are concerned with `UITextInputTraits` etc.

This is a huge contribution and I sincerely hope to see you share more with us, but right now we can't devote time to integrating this particular one. If you want to talk about developing the framework, we're always in open discussion at the `#development` channel on http://asyncdisplaykit.org/slack . Cheers man!

kyleshank:
Joined the slack.  If anyone is interested in using ASTextFieldNode it is available as a framework here:
https://github.com/pocketlabs/ASTextFieldNode

Adlai-Holler:
👌

PR #3061, opened 2017-02-22
Promote ASIndexedNodeContext -> ASCollectionElement

Adlai-Holler:
- Rename `ASIndexedNodeContext -> ASCollectionElement`. This context is not "indexed" anymore – it may live at different indexes in different collection data sets.
- Remove the now-unused (and terribly misleading) `indexPath` field from the object. This was always destined to happen, and now it is.

Ready for review!


AsyncDisplayKit/Details/ASCollectionItem.mm
```diff
@@ -51,10 +49,9 @@ - (ASCellNode *)node
     ASCellNode *node = _nodeBlock();
     _nodeBlock = nil;
     if (node == nil) {
-      ASDisplayNodeFailAssert(@"Node block returned nil node! Index path: %@", _indexPath);
+      ASDisplayNodeFailAssert(@"Node block returned nil node!");
       node = [[ASCellNode alloc] init];
     }
-    node.cachedIndexPath = _indexPath;
```
Adlai-Holler: The loss of this encache is notable, but it will probably have no runtime impact. This was always a hack, and we'll find a better way to quickly look up the index path for a cell node (probably using a `ASCellNode -> NSIndexPath` map table in the data controller).

AsyncDisplayKit.xcodeproj/project.pbxproj
```diff
@@ -1749,7 +1749,6 @@
 				058D0A39195D057000B7D73C /* ASDisplayNodeAppearanceTests.m in Sources */,
 				CCB2F34D1D63CCC6004E6DE9 /* ASDisplayNodeSnapshotTests.m in Sources */,
 				AE6987C11DD04E1000B9E458 /* ASPagerNodeTests.m in Sources */,
-				699B83511E3C1BA500433FA4 /* ASLayoutSpecTests.m in Sources */,
```
Adlai-Holler: This seems to come up every time I open the project in Xcode. Unless someone objects, let's let this ride-along with this diff.

PR #3062, opened 2017-02-23
Prevent Adding View-Backed Nodes to Layer-Backed Hierarchies

### LINKED ISSUE #3038: Rotating ASDKgram Causes Crash
You have to first scroll down a few photos, and then rotate.

Reported by @ay8s on Slack.

Adlai-Holler:
Resolves #3038

Doing this will prevent required callbacks like `didMoveToWindow` from getting sent to our view.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2255,7 +2255,7 @@ static const char *ASDisplayNodeDrawingPriorityKey = "ASDrawingPriority";
   }
 
   ASDisplayNodeAssert(_flags.layerBacked, @"We shouldn't get called back here if there is no layer");
-  return (id)kCFNull;
+  return nil;
 }
 
 #pragma mark - Error Handling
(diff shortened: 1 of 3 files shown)


AsyncDisplayKitTests/ASDisplayNodeTests.mm
```diff
@@ -1098,33 +1098,34 @@ - (void)testSubnodes
...
+- (void)checkReplaceSubnodeLoaded:(BOOL)loaded layerBacked:(BOOL)isLayerBacked
 {
   DeclareNodeNamed(parent);
   DeclareNodeNamed(a);
   DeclareNodeNamed(b);
   DeclareNodeNamed(c);
+  DeclareNodeNamed(d);
 
-  for (ASDisplayNode *n in @[parent, a, b, c]) {
+  for (ASDisplayNode *n in @[parent, a, b, c, d]) {
```
Adlai-Holler: These tests forgot to set `layerBacked` on the new subnode `d` so they failed our new assertion. The method name is also super confusing so I updated it.

AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2551,6 +2551,11 @@ - (void)_insertSubnode:(ASDisplayNode *)subnode atSubnodeIndex:(NSInteger)subnod
     return;
   }
   
+  if (self.layerBacked && !subnode.layerBacked) {
+    ASDisplayNodeFailAssert(@"Cannot add a view-backed node as a subnode of a layer-backed node. Supernode: %@, subnode: %@", self, subnode);
```
garrettmoon: I swear this type of check existed before?

Adlai-Holler: I think so too, but the test that failed post-change is super old so maybe not… ¯\_(ツ)_/¯

appleguy: yeah, this is really weird.  The check did exist before, but I'm not sure how long ago.  We should be pretty worried it vanished lol.

PR #3063, opened 2017-02-23
Clean Up actionForLayer:forKey:

Adlai-Holler:
- Fix comments
- Return nil

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode.mm
+++ b/AsyncDisplayKit/ASDisplayNode.mm
@@ -2245,7 +2245,7 @@ static const char *ASDisplayNodeDrawingPriorityKey = "ASDrawingPriority";
 
 #pragma mark <CALayerDelegate>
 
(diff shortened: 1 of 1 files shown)


AsyncDisplayKit/ASDisplayNode.mm
```diff
@@ -2254,8 +2254,8 @@ - (void)subnodeDisplayDidFinish:(ASDisplayNode *)subnode
     [self __exitHierarchy];
   }
 
-  ASDisplayNodeAssert(_flags.layerBacked, @"We shouldn't get called back here if there is no layer");
-  return (id)kCFNull;
+  ASDisplayNodeAssert(_flags.layerBacked, @"We shouldn't get called back here unless we are layer-backed.");
+  return nil;
```
Adlai-Holler: The different between returning `kCFNull` and `nil` is, the latter is simpler and the former explicitly prevents core animation from calling `+defaultActionForKey:` on the layer, which we want it to do.

PR #3066, opened 2017-02-23
[ASCollectionView/ASTableView] Support native item and section reloads

### LINKED ISSUE #469: reloadSections(_, withRowAnimation:_) does not reload the section header view
I am using an `ASTableView` that consists of multiple sections. Each section consists of a reusable `UITableViewHeaderFooterView` and a list of rows. In the past, in order to update data inside the section header view, I would call `reloadSections` on the table view for the sections I wanted to update and it would update but in ASDK this is not the case. 

The table view delegate callback: `func tableView(tableView: UITableView, viewForHeaderInSection section: Int) -> UIView?` is not invoked which is telling me only the section contents are reloaded, not the header.

Is this a bug or is there a different way (short of reloading the entire table) to reload the section header view?

nguyenhuy:
After #3017, it's much easier to support native reloads now. Since updates are already broken down to inserts and deletes by `_ASHierarchyChangeSet`, `ASDataController` can consume these inserts and deletes as is, and then `ASCollectionView`/`ASTableView` forward the original changes to `UICollectionView`/`UITableView`. This resolves #469. 

Demo:
- [Reload item](https://cl.ly/jIkf)
- [Reload section](https://cl.ly/jIC7)

### DIFF
--- a/AsyncDisplayKit/Details/ASDataController.mm
+++ b/AsyncDisplayKit/Details/ASDataController.mm
@@ -512,6 +512,9 @@ typedef void (^ASDataControllerCompletionBlock)(NSArray<ASCollectionElement *> *
     return;
   }
   
+  // TODO if the change set includes solely section reloads that together are equivalent to reloadData (i.e reload the only section),
+  // do a reloadData here as an optimization.
+  
   if (changeSet.includesReloadData) {
     [_sections removeAllObjects];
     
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -512,6 +512,9 @@ - (void)_updateSectionContextsWithChangeSet:(_ASHierarchyChangeSet *)changeSet
     return;
   }
   
+  // TODO if the change set includes solely section reloads that together are equivalent to reloadData (i.e reload the only section),
+  // do a reloadData here as an optimization.
+
```
Adlai-Holler: Since `reloadData` is never animated, but section reloads may be, we probably won't want to build this optimization.

PR #3071, opened 2017-02-23
ASDataController: Eliminate Mutable State

Adlai-Holler:
- Before data controller had one mutable & one immutable representation of the data.
- Now data controller has just two immutable data representations: `completedElements` and `pendingElements`
- Mutable representation is never saved – it's created in `updateWithChangeSet:` and provided to methods that are expected to modify it, and then immediately cloned.

This makes the data controller much easier to reason about, and sets the stage for building a custom immutable representation of the state-of-the-world for data controller.

### DIFF
--- a/AsyncDisplayKit/ASTableView.mm
+++ b/AsyncDisplayKit/ASTableView.mm
@@ -128,6 +128,10 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 
   NSIndexPath *_pendingVisibleIndexPath;
 
+  // When we update our data controller in response to an interactive move,
+  // we don't want to tell the table view about the change (it knows!)
+  BOOL _updatingInResponseToInteractiveMove;
+
   // The top cell node that was visible before the update.
   __weak ASCellNode *_contentOffsetAdjustmentTopVisibleNode;
   // The y-offset of the top visible row's origin before the update.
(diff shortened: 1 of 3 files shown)


AsyncDisplayKit/ASTableView.mm
```diff
@@ -128,6 +128,10 @@ @interface ASTableView () <ASRangeControllerDataSource, ASRangeControllerDelegat
 
   NSIndexPath *_pendingVisibleIndexPath;
 
+  // When we update our data controller in response to an interactive move,
+  // we don't want to tell the table view about the change (it knows!)
+  BOOL _updatingInResponseToInteractiveMove;
```
Adlai-Holler: Before, ASTableView had a backdoor into data controller that let it handle interactive moves by blindly modifying the `completedNodes`.

That backdoor is closed now! Instead, table view issues an update through the normal data controller mechanism, and uses an internal flag to say "oh! This update was in response to an interactive move. I don't need to tell `super` about it, because `super` already knows.

AsyncDisplayKit/Details/ASDataController.mm
```diff
@@ -839,26 +842,34 @@ - (ASCellNode *)supplementaryNodeOfKind:(NSString *)kind atIndexPath:(NSIndexPat
...
 
 //TODO Move this to somewhere else?
-+ (ASNodeContextTwoDimensionalDictionary *)deepImmutableCopyOfElementsDictionary:(ASNodeContextTwoDimensionalDictionary *)originalDict
++ (ASCollectionElementTwoDimensionalDictionary *)deepImmutableCopyOfElementsDictionary:(ASCollectionElementTwoDimensionalDictionary *)originalDict
+{
+  NSMutableDictionary *deepCopy = [NSMutableDictionary dictionaryWithCapacity:originalDict.count];
+  [originalDict enumerateKeysAndObjectsUsingBlock:^(NSString * _Nonnull kind, ASCollectionElementTwoDimensionalArray * _Nonnull obj, BOOL * _Nonnull stop) {
+    deepCopy[kind] = [[NSArray alloc] initWithArray:obj copyItems:YES];
+  }];
+  return deepCopy;
```
garrettmoon: This is returning a mutable dictionary, is that intended or should it return `[deepCopy copy]`?

PR #3072, opened 2017-02-24
[ASRunloopQueue] Improve logging output of runloop queue when enabled.

appleguy:


AsyncDisplayKit/ASRunLoopQueue.mm
```diff
@@ -65,17 +66,18 @@ - (void)threadMain
     // 100ms timer.  No resources are wasted in between, as the thread sleeps, and each check is fast.
     // This time is fast enough for most use cases without excessive churn.
     CFRunLoopTimerRef timer = CFRunLoopTimerCreateWithHandler(NULL, -1, 0.1, 0, 0, ^(CFRunLoopTimerRef timer) {
+      weakSelf->_queueLock.lock();
```
appleguy: Moving this up here because this turns pretty frequently, and we do not need to create / destroy an autorelease pool on a turn unless items are removed from the queue (indeed, unless some object is retained / autoreleased).  Since all these calls are C++ there should be no autorelease possible.

PR #3073, opened 2017-02-24
ASDataController: Adopt ASElementMap

Adlai-Holler:
- Add a new immutable class `ASElementMap` that gives a state-of-the-world for a collection/table.
- Have `ASDataController` vend two maps: one visible and one pending.
- Have `ASTableView ASCollectionView ASRangeController` use the maps directly to answer questions like `numberOfSections` etc.
- Remove `cachedIndexPath` and `supplementaryElementKind` fields from `ASCellNode+Internal`. Add new `collectionElement` field to it.
  - Index path can be efficiently looked up from a given element map.
  - Supplementary element kind can be looked up from new `collectionElement` field.
  - `collectionElement` field added on assumption that if you return an existing cell node in a node block, then it no longer participates in the old collection view.

Plenty more work to do from here! At least now we can reliably move collection view data states around.

### DIFF
--- a/AsyncDisplayKit/ASCellNode+Internal.h
+++ b/AsyncDisplayKit/ASCellNode+Internal.h
@@ -14,6 +14,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class ASCollectionElement;
+
 @protocol ASCellNodeInteractionDelegate <NSObject>
 
 /**
@@ -58,10 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, strong, nullable) UICollectionViewLayoutAttributes *layoutAttributes;
 
-/// readwrite variant of the readonly public property.
-@property (nonatomic, copy, nullable) NSString *supplementaryElementKind;
-
-@property (nonatomic, copy, nullable) NSIndexPath *cachedIndexPath;
+@property (weak, nullable) ASCollectionElement *collectionElement;
 
 @property (nonatomic, weak, nullable) ASDisplayNode *owningNode;
 
--- a/AsyncDisplayKit/Details/ASDataController.h
+++ b/AsyncDisplayKit/Details/ASDataController.h
@@ -117,8 +117,14 @@ AS_SUBCLASSING_RESTRICTED
 
 - (instancetype)initWithDataSource:(id<ASDataControllerSource>)dataSource eventLog:(nullable ASEventLog *)eventLog NS_DESIGNATED_INITIALIZER;
 
+/**
+ * The map that is currently displayed. The "UIKit index space."
+ */
 @property (nonatomic, strong, readonly) ASElementMap *visibleMap;
 
+/**
+ * The map that is currently being prepared for display. The "data source index space."
+ */
 @property (nonatomic, strong, readonly) ASElementMap *pendingMap;
 
 /**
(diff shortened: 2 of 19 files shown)


AsyncDisplayKit/ASCellNode+Internal.h
```diff
@@ -58,10 +60,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, strong, nullable) UICollectionViewLayoutAttributes *layoutAttributes;
 
-/// readwrite variant of the readonly public property.
-@property (nonatomic, copy, nullable) NSString *supplementaryElementKind;
-
-@property (nonatomic, copy, nullable) NSIndexPath *cachedIndexPath;
+@property (weak, nullable) ASCollectionElement *collectionElement;
```
garrettmoon: Does the collectionElement hold a strong reference to its cellNode? Shouldn't be the other way around?

AsyncDisplayKit/Details/ASDataController.h
```diff
@@ -116,6 +117,16 @@ extern NSString * const ASCollectionInvalidUpdateException;
 - (instancetype)initWithDataSource:(id<ASDataControllerSource>)dataSource eventLog:(nullable ASEventLog *)eventLog NS_DESIGNATED_INITIALIZER;
 
 /**
+ * The map that is currently displayed. The "UIKit index space."
+ */
+@property (nonatomic, strong, readonly) ASElementMap *visibleMap;
```
garrettmoon: I wonder if there's a better name for this… I feel like completeMap is wrong but possibly is less confusing (maybe it's just confusing in a different way). Possibly uiSpaceMap (looks terrible)?

garrettmoon: maybe viewMap?

AsyncDisplayKit/Private/ASElementMap.m
```diff
@@ -0,0 +1,151 @@
...
+    for (NSArray *section in _sectionsOfItems) {
+      NSInteger i = 0;
+      for (ASCollectionElement *element in section) {
+        NSIndexPath *indexPath = [NSIndexPath indexPathForItem:i inSection:s];
+        [_elementToIndexPathMap setObject:indexPath forKey:element];
+        i++;
+      }
+      s++;
+    }
+    for (NSDictionary *supsOfKind in [_supplementaryElements objectEnumerator]) {
```
garrettmoon: Super super nit, feel free to ignore, I'd spell this out; it's the objective-c way :)

AsyncDisplayKit/ASCellNode+Internal.h
Adlai-Holler: True – right now the Element owns the Node. That way node stays alive as long as someone has a reference to its collection element, but if you take a node and put it somewhere else, then we let the element die if nobody else retains it.

AsyncDisplayKit/Details/ASDataController.h
Adlai-Holler: True. Let's leave the name only because this property is doomed anyway – it's here right now to keep this diff small. In a follow-up diff coming soon, I'm going to have the DataController generate this and then hand it out through its delegate, rather than saving it. The data controller shouldn't really care what's "visible", it's just a processing unit at this point.

Issue #2616, opened 2016-11-16
NavigationController is nil inside ASCellNode


MarvinNazari:
When using `ASCellNode(viewControllerBlock:` the viewController dosent inherit the `navigationController` so it returns nil, this is when i use `ASPagerNode`

using `1.9.90`

hannahmbanana:
@MarvinNazari - This is also possible to implement, but has been something we've needed ourselves yet.

MarvinNazari:
@hannahmbanana Im not really sure how can i implement this, i think the problem is that when adding a view controller inside a another view controller we need to use `addChildViewController` and `didMoveToParentViewController` but thats not possible inside a `ASCellNode`
Any idea?

Adlai-Holler:
@MarvinNazari What we need in order pull this off, is:

- In `willDisplayCell:`, get the view controller that owns the `ASCollectionView` (you can do this by repeatedly calling `-nextResponder` until a view controller is returned) and call `addChildViewController` and `didMoveToParentViewController:` with the cell node's view controller.
- In `didEndDisplayingCell:`, call `willMoveToParentViewController:nil` and then `removeFromParentViewController` on the cell node's view controller.

Since `ASTableView` and `ASCollectionView` will both have to do this, you may want to create methods like `-[ASCellNode associateWithViewControllerForView:(UIView *)]` `-[ASCellNode disassociateFromViewController` to hold the duplicated code.

Andy1984:
Same problem. You need to `addChildViewController` in `- (ASCellNode *)pagerNode:(ASPagerNode *)pagerNode nodeAtIndex:(NSInteger)index`

garrettmoon:
Going to close this as it seems like it was addressed. If not, please feel free to open a new issue in the Texture repo: https://github.com/texturegroup/texture/issues

PR #3076, opened 2017-02-24
Add a Convenience Property to Get Owning View Controller

Adlai-Holler:
This could be tremendously useful in debugging. The responder chain is underrated. Sad!

I chose to add a new category `Convenience` to ASDisplayNode, so that we can start separating out core functionality from stuff like this.

### DIFF
--- a/AsyncDisplayKit/ASDisplayNode+Convenience.m
+++ b/AsyncDisplayKit/ASDisplayNode+Convenience.m
@@ -1,0 +1,39 @@
+//
+//  ASDisplayNode+Convenience.m
+//  AsyncDisplayKit
+//
+//  Created by Adlai Holler on 2/24/17.
+//  Copyright © 2017 Facebook. All rights reserved.
+//
+
+#import "ASDisplayNode+Convenience.h"
+
+#import <UIKit/UIViewController.h>
+
+#import <AsyncDisplayKit/ASDisplayNodeExtras.h>
+#import <AsyncDisplayKit/ASResponderChainEnumerator.h>
+
+@implementation ASDisplayNode (Convenience)
+
+- (__kindof UIViewController *)closestViewController
+{
+  ASDisplayNodeAssertMainThread();
+  
+  // Careful not to trigger node loading here.
+  if (!self.nodeLoaded) {
+    return nil;
+  }
+
+  // Get the closest view.
+  UIView *view = ASFindViewOfLayer(self.layer);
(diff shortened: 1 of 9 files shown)


AsyncDisplayKit/ASDisplayNodeExtras.mm
```diff
@@ -251,13 +251,19 @@ static inline BOOL _ASDisplayNodeIsAncestorOfDisplayNode(ASDisplayNode *possible
 
 extern UIWindow * _Nullable ASFindWindowOfLayer(CALayer *layer)
 {
+  UIView *view = ASFindViewOfLayer(layer);
```
appleguy: Doesn't this need to ascend the superlayer path if there is no corresponding view at this level, but a parent is inside of a window?

Adlai-Holler: `ASFindViewOfLayer` does that walking. I considered calling it `ASFindClosestViewOfLayer` of similar but decided to use the cleaner name and rely on documentation 😛

nguyenhuy: I tend to like `ASFindClosestViewOfLayer` actually. Definitely not a blocking issue though.

nguyenhuy: I tend to like `ASFindClosestViewOfLayer` actually. Definitely not a blocking issue though.

MarvinNazari:
@Adlai-Holler this PR makes #2616 really easy to implement.

AsyncDisplayKit/ASDisplayNode+Convenience.m
```diff
@@ -0,0 +1,39 @@
...
+#import <AsyncDisplayKit/ASResponderChainEnumerator.h>
+
+@implementation ASDisplayNode (Convenience)
+
+- (__kindof UIViewController *)closestViewController
+{
+  ASDisplayNodeAssertMainThread();
+  
+  // Careful not to trigger node loading here.
+  if (!self.nodeLoaded) {
```
maicki: Should we document that if the node was not loaded we return nil.

AsyncDisplayKit/ASDisplayNodeExtras.mm
maicki: +1 ASFindClosestViewOfLayer

PR #3080, opened 2017-02-25
Clean Up Many Files

Adlai-Holler:
I was prospecting some new ideas and encountered some surprising issues.

- Make `ASRangeController.h` and the ASLayoutController headers non-public. This is technically **breaking API** but we shouldn't treat it as such. These classes were never meant for public access.
- Move the `shouldShowRangeDebugOverlay` property over to `ASDisplayNode` from `ASRangeController`. Also technically **breaking API** but we shouldn't treat it as such. This method is called only specifically for debugging, and it's not worth leaving `ASRangeController` public just for this one property.
- Fix broken file ref to `AsyncDisplayKit+Debug.h/.m`
- Move internal range controller debugging methods out of public header `AsyncDisplayKit+Debug.h`
- Remove `ASLayoutRangeModeCount` which was being used as an "invalid" `ASLayoutRangeMode` _and_ as the number of valid range modes (4).
  - I get why this was done. It's nice how it "auto-updates" as we add range modes.
  - But it's confusing as hell and incorrect. It's not a range mode. It's an integer. There is no "Count" range mode.
  - I added a `ASLayoutRangeModeUnspecified = -1` to represent a range mode that is unknown.
- Make `ASLayoutRangeTypeCount` an actual integer, rather than an `ASLayoutRangeType`. Again, cute, I get it, great job.


maicki:
@Adlai-Holler If you fiddle around with the project anyway can you move the `Base` folder from the root directory into AsyncDisplayKit please.

Adlai-Holler:
@maicki Good call. I went ahead and made some more file shuffles:
- `/AsyncDisplayKit` -> `/Source`
- `/AsyncDisplayKitTests` -> `/Tests`
- `/AsyncDisplayKitTestHost` -> `/Test/TestHost`

nguyenhuy:
Cool! Thanks for updating BUCK file as well!

3a4oT:
@Adlai-Holler  Hey! Have you pushed new 2.2 podspec to the trunk?

nguyenhuy:
@3a4oT 2.2 has not been pushed to CocoaPods yet. You can still try it using this: `pod 'AsyncDisplayKit', :git => 'https://github.com/facebook/AsyncDisplayKit.git', :branch => '2.2'`

3a4oT:
@nguyenhuy thanks. Do we know a release date?

nguyenhuy:
@3a4oT It will be publish within 24 hours. Btw, you can get way faster answers if you reach us on our Slack channel: http://asyncdisplaykit.org/slack.html

Issue #3053, opened 2017-02-20
_ASCollectionViewCell inherits content node properties like backgroundColor? 


skensell:
Hey Guys, I'm in the process of upgrading from 1.9 to 2.1, and this is the first in hopefully not a long sequence of bugs which I'll report:

My issue is that I have an `ASCellNode` which sets `backgroundColor` and `cornerRadius` like so:

```Obj-C
self.backgroundColor = [[UIColor blackColor] colorWithAlphaComponent:0.25]; // in init
self.cornerRadius = self.bounds.size.height / 2.0f; // in layoutDidFinish
```
and then it seems the `backgroundColor` is transferred to the backing cell in `[_ASCollectionViewCell setNode:]` while the `cornerRadius` is not. I don't understand why _any_ property is being transferred to the backing cell and rather you just let the content node make display decisions.

It results in a funny picture like this:
<img width="85" alt="screen shot 2017-02-20 at 9 55 33 pm" src="https://cloud.githubusercontent.com/assets/3983827/23141575/b5d52c38-f7b7-11e6-84dc-23aba6298e8a.png">

You can see that there are a few problems:

1. The `backgroundColor` of the cell and the node blend to form a darker shade than initially desired.
2. The cell does not inherit the `cornerRadius` property from the node, and appears as a square.

To solve this, I was considering

Workaround 1: Add a `contentNode` subnode to each of my `ASCellNode`s and only modify _its_ `cornerRadius` and `backgroundColor`.
Workaround 2: Hack the `_ASCollectionViewCell` to have `[UIColor clearColor]`.

I went with Workaround 2, which I'm not very proud of:

```Obj-C
- (void)collectionNode:(ASCollectionNode *)collectionNode willDisplayItemWithNode:(ASCellNode *)node {
    UICollectionViewCell* cell = nil;
    UIView* view = node.isNodeLoaded ? node.view : nil;
    while (view && ![view isKindOfClass:[UICollectionViewCell class]]) {
        view = view.superview;
    }
    if (view) {
        cell = (UICollectionViewCell*)view;
    }
    cell.backgroundColor = [UIColor clearColor];
}
```

IMO, it would make more sense if `_ASCollectionViewCell` just had `[UIColor clearColor]` as its background and it shouldn't inherit any properties from the contentNode.  After looking around, I noticed `_ASTableViewCell` has the same issue.  Is this intentional? What's the reasoning?

maicki:
[...]

Issue #3044, opened 2017-02-17
Strange ASCellNode background in 2.1


acedened:
ASCellNode in 2.1 has strange background in addition to the one that I have specified in code. This background doesn't respond to frame or cornerRadius changes. Everything was fine in 2.0.1.

Here is the screenshot of 2.1:
![wrong](https://cloud.githubusercontent.com/assets/7037300/23075081/3a552fa8-f54c-11e6-9ee5-18db6e8eb745.png)

This is how it was in 2.0.1:
![fine](https://cloud.githubusercontent.com/assets/7037300/23075131/662ad006-f54c-11e6-9cd1-dd5623e7b553.png)

maicki:
This can be related to: #3053

garrettmoon:
It looks like this was fixed, if not, do you mind reopening an issue on the Texture repo? https://github.com/texturegroup/texture/issues

PR #3081, opened 2017-02-26
Fix cell selection state issues

### LINKED ISSUE #3037: Breaking ASTableView regression in 2.0.2 - Cell selection state
Sometime between `2.0.1` and `2.0.2`, an issue related to (I think) cell reuse was introduced. 

Previously, cells correctly maintained their selection state after being reused, but as of `2.0.2`, the selection state of reused cells is "leaking" into new unrelated cells. This breaks some pretty important functionality, since selection state is no longer tracked and represented on screen accurately.

The best way to describe this is with the GIFs below:

### Correct Selection (2.0.1)

cell-edit-state-correct

### Broken Selection (2.0.2)

cell-edit-state-wrong
cell-selection-state-wrong

I wasn't able to pinpoint which commit caused this, but hopefully someone more aware of the recent changes can shed some light.

rewandy:
Resolves #3037

This is a do-over of c85aa11, with the following differences:
- Fixed `_ASTableViewCell` properties **always** being mutated to match corresponding `ASCellNode` passthrough properties, even when the hosted `ASCellNode` is currently `nil`. Now only inheriting passthrough properties when there’s an actual instance of `ASCellNode` to read them from.
- Corrected spelling of `ASCellNode`’s `separatorInset` property.
- Reverted `_ASCollectionViewCell` inheriting the hosted cell node’s `backgroundColor` & `clipsToBounds` properties. This seems to be surprising and unwanted behavior to some, as seen in #3053 and #3044.
- Moved passthrough of `UITableViewCell`’s `selectedBackgroundView` into `_ASTableViewCell`’s `-setNode:` method for consistency.

### DIFF
--- a/Source/ASCellNode.h
+++ b/Source/ASCellNode.h
@@ -182,10 +182,10 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  */
 @property (nonatomic) UITableViewCellAccessoryType accessoryType;
 
-/* @abstract The seperator inset of the cell seperator line
+/* @abstract The inset of the cell separator line
  * ASTableView uses these properties when configuring UITableViewCells that host ASCellNodes.
  */
-@property (nonatomic) UIEdgeInsets seperatorInset;
+@property (nonatomic) UIEdgeInsets separatorInset;
 
 @end
 
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -79,11 +79,19 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 {
   _node = node;
   
-  self.backgroundColor = node.backgroundColor;
-  self.selectionStyle = node.selectionStyle;
-  self.accessoryType = node.accessoryType;
-  self.separatorInset = node.seperatorInset;
-  self.clipsToBounds = node.clipsToBounds;
+  if (node) {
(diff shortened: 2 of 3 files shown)


Source/ASCellNode.h
```diff
@@ -182,10 +182,10 @@ typedef NS_ENUM(NSUInteger, ASCellNodeVisibilityEvent) {
  */
 @property (nonatomic) UITableViewCellAccessoryType accessoryType;
 
-/* @abstract The seperator inset of the cell seperator line
+/* @abstract The separatorInset inset of the cell separatorInset line
```
Adlai-Holler: Just `separator`

Source/ASTableView.mm
```diff
@@ -864,7 +872,6 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
     [_rangeController configureContentView:cell.contentView forCellNode:node];
 
     cell.node = node;
-    cell.selectedBackgroundView = node.selectedBackgroundView;
```
Adlai-Holler: Nice, how'd that get there 😅

Source/ASCellNode.h
rewandy: Oops, too much copy-paste here. Good catch!

rewandy:
Thanks for the review @Adlai-Holler! Much appreciated.

That one comment should now be addressed.

PR #3088, opened 2017-02-27
prevent crash on 'Expected index path of length 2. Index path: (null)'

newbdez33:


Adlai-Holler:
Hi @newbdez33 thanks for the diff, but we took another fix for this issue from #3090. Hope to see you around more!

PR #3091, opened 2017-02-27
Clean up snapshot test cases

nguyenhuy:
- Drop 32bit snapshots
- Update iOS 9 snapshots and move some others to the correct directory
- Remove `setUp` overrides in subclasses of `ASLayoutSpecSnapshotTestCase` that set `recordMode` flag. Doing so in multiple places make it difficult to enable/dis
able the flag for all layout spec test cases. It can always be overridden locally if need
to.
- Fix `testThatOnDidLoadThrowsIfCalledOnLoadedOffMain` of `ASDisplayNodeTests`
- Fix BUCK build


Tests/ASDisplayNodeTests.mm
```diff
@@ -2149,12 +2149,9 @@ - (void)testThatOnDidLoadThrowsIfCalledOnLoadedOffMain
 {
   ASTestDisplayNode *node = [[ASTestDisplayNode alloc] init];
   [node view];
-  dispatch_semaphore_t sem = dispatch_semaphore_create(0);
-  [NSThread detachNewThreadWithBlock:^{
```
nguyenhuy: This test case was failing on iOS 9 because `detachNewThreadWithBlock` is available since iOS 10.

PR #3092, opened 2017-02-27
Allow users to set child directly on overlay and background specs

nguyenhuy:
Current calling `layoutSpec.child` on a background or overlay layout spec would cause assertion: 
```
Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'This layout spec does not support more than one child. Use the setChildren: or the setChild:AtIndex: API'
```

This is inconvenient to users because it forbids them from doing something like below and instead have to rely on initializers of these specs:
```Objective-C
    ASOverlayLayoutSpec *containerSpec = [[ASOverlayLayoutSpec alloc] init];
    containerSpec.overlay = titleSpec;
    containerSpec.child = self.imageNode;
```

[Reported within Pinterest](https://pinterest.slack.com/archives/ios-help/p1487814286000839). Inconvenient no more!

### DIFF
--- a/Source/Layout/ASBackgroundLayoutSpec.mm
+++ b/Source/Layout/ASBackgroundLayoutSpec.mm
@@ -33,9 +33,7 @@ static NSUInteger const kBackgroundChildIndex = 1;
   if (!(self = [super init])) {
     return nil;
   }
-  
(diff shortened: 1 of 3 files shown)


yans:
looks good, thanks for making this change @nguyenhuy

PR #3093, opened 2017-02-27
[ASVideoPlayerNode] Always load asset when it enters preload state

### LINKED ISSUE #3042: [ASVideoPlayerNode] Crash initialising videoplayer node
Im getting a crash initialising ASVideoPlayerNode inside my ASCellNode.
looking at the diff i think [this](https://github.com/facebook/AsyncDisplayKit/commit/f7a0ac9760be7dc8c532fa7f9f232c5dcf738a30#diff-d560a1b1d55d40fbeef7c555fbeb22e9) caused the problem.

cc @nguyenhuy 

[Sample code](https://github.com/facebook/AsyncDisplayKit/files/783197/ASVideoPlayerNode.Crash.zip)

```
2017-02-17 09:56:45.642 Starflow[20538:8744927] *** Assertion failure in -[ASVideoNode setAsset:], /Users/marvinnazari/Code/Starflow/iOS/Pods/AsyncDisplayKit/AsyncDisplayKit/ASVideoNode.mm:499
2017-02-17 09:56:45.718 Starflow[20538:8744927] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'This method must be called on the main thread'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001088cbd4b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00000001064c221e objc_exception_throw + 48
	2   CoreFoundation                      0x00000001088cfe42 +[NSException raise:format:arguments:] + 98
	3   Foundation                          0x000000010605766d -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 195
	4   AsyncDisplayKit                     0x000000010377f66a -[ASVideoNode setAsset:] + 298
	5   AsyncDisplayKit                     0x0000000103782b8e -[ASVideoPlayerNode _init] + 318
	6   AsyncDisplayKit                     0x0000000103781fa7 -[ASVideoPlayerNode init] + 183
	7   Starflow                            0x00000001021b43b0 _TTOFCSo17ASVideoPlayerNodecfT_S_ + 16
	8   Starflow                            0x00000001021b2a84 _TFCSo17ASVideoPlayerNodeCfT_S_ + 68
	9   Starflow                            0x0000000102271554 _TFC8Starflow5NNOdecfT8isActiveSb_S0_ + 52
	10  Starflow                            0x00000001022713c4 _TFC8Starflow5NNOdeCfT8isActiveSb_S0_ + 52
	11  Starflow                            0x0000000102271386 _TFFC8Starflow31SFVideoCollectionViewController14collectionNodeFTCSo16ASCollectionNode18nodeBlockForItemAtV10Foundation9IndexPath_FT_CSo10ASCellNodeU_FT_S4_ + 22
	12  Starflow                            0x00000001020d2d17 _TTRXFo__oCSo10ASCellNode_XFdCb__aS__ + 39
	13  AsyncDisplayKit                     0x0000000103616fe8 __56-[ASCollectionView dataController:nodeBlockAtIndexPath:]_block_invoke_3.619 + 104
	14  AsyncDisplayKit                     0x00000001036ac2af -[ASIndexedNodeContext node] + 127
	15  AsyncDisplayKit                     0x0000000103629de2 __45-[ASDataController _layoutNodesFromContexts:]_block_invoke + 162
	16  AsyncDisplayKit                     0x0000000103639810 ___ZL15ASDispatchApplymPU28objcproto17OS_dispatch_queue8NSObjectmU13block_pointerFvmE_block_invoke + 80
	17  libdispatch.dylib                   0x000000010a62d978 _dispatch_call_block_and_release + 12
	18  libdispatch.dylib                   0x000000010a6570cd _dispatch_client_callout + 8
	19  libdispatch.dylib                   0x000000010a636445 _dispatch_queue_override_invoke + 1733
	20  libdispatch.dylib                   0x000000010a638385 _dispatch_root_queue_drain + 720
	21  libdispatch.dylib                   0x000000010a638059 _dispatch_worker_thread3 + 123
	22  libsystem_pthread.dylib             0x000000010aa06712 _pthread_wqthread + 1299
	23  libsystem_pthread.dylib             0x000000010aa061ed start_wqthread + 13
)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

nguyenhuy:
Some problems in `ASVideoPlayerNode` currently:
- `_loadAssetWhenNodeBecomesVisible` isn't turned on in the empty initializer. This causes an assertion in #3042 because it's not safe to set an asset on ASVideoNode off main.
- The flag itself is troublesome. It's enabled by default, but even so, "visible" is a bit too late. The asset can be loaded as soon as the player enters preload state, which is what ASVideoNode is doing.

This PR deprecates `_loadAssetWhenNodeBecomesVisible` and fixes #3042 by letting `ASVideoPlayerNode` handles its objects automatically.

@MarvinNazari Would be great if you can comment and help me to test this PR. Thank you!

### DIFF
--- a/Source/ASVideoPlayerNode.h
+++ b/Source/ASVideoPlayerNode.h
@@ -39,7 +39,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, assign) BOOL controlsDisabled;
 
-@property (nonatomic, assign, readonly) BOOL loadAssetWhenNodeBecomesVisible;
+@property (nonatomic, assign, readonly) BOOL loadAssetWhenNodeBecomesVisible ASDISPLAYNODE_DEPRECATED_MSG("Asset is always loaded when this node enters preload state. This flag does nothing.");
 
 #pragma mark - ASVideoNode property proxy
 /**
--- a/Source/ASVideoPlayerNode.mm
+++ b/Source/ASVideoPlayerNode.mm
@@ -72,7 +72,6 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
   ASStackLayoutSpec *_controlFlexGrowSpacerSpec;
   ASDisplayNode *_spinnerNode;
 
-  BOOL _loadAssetWhenNodeBecomesVisible;
   BOOL _isSeeking;
   CMTime _duration;
 
@@ -114,7 +113,6 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
   
   _url = url;
   _asset = [AVAsset assetWithURL:_url];
-  _loadAssetWhenNodeBecomesVisible = YES;
   
   [self _init];
   
@@ -128,7 +126,6 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
   }
 
   _asset = asset;
-  _loadAssetWhenNodeBecomesVisible = YES;
 
   [self _init];
 
@@ -144,7 +141,6 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
   _asset = asset;
   _videoComposition = videoComposition;
   _audioMix = audioMix;
-  _loadAssetWhenNodeBecomesVisible = YES;
 
   [self _init];
 
@@ -159,7 +155,6 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
 
   _url = url;
   _asset = [AVAsset assetWithURL:_url];
-  _loadAssetWhenNodeBecomesVisible = loadAssetWhenNodeBecomesVisible;
 
   [self _init];
 
@@ -173,7 +168,6 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
   }
 
   _asset = asset;
-  _loadAssetWhenNodeBecomesVisible = loadAssetWhenNodeBecomesVisible;
 
   [self _init];
 
@@ -189,7 +183,6 @@ static void *ASVideoPlayerNodeContext = &ASVideoPlayerNodeContext;
   _asset = asset;
   _videoComposition = videoComposition;
   _audioMix = audioMix;
-  _loadAssetWhenNodeBecomesVisible = loadAssetWhenNodeBecomesVisible;
 
   [self _init];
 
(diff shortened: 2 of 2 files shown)


MarvinNazari:
@nguyenhuy awesome! gonna try it later tonight.

MarvinNazari:
@nguyenhuy so the crash is gone, but now it dosent fetch right when i set the url to videoNode after some delay, if you update the sample project in the issue to this you see what is happening (scroll up and down a little bit)
this happens if i do it in the `didLoad()` too

```swift
final class SFVideoCollectionNode: ASCellNode {
    
    fileprivate let videoPlayerNode: ASVideoPlayerNode
    let isActive: Bool
    
    init(isActive: Bool = false) {
        self.isActive = isActive
        videoPlayerNode = ASVideoPlayerNode()
        super.init()
        automaticallyManagesSubnodes = true
        
        fetch()
    }

    func fetch() {
        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
            self.videoPlayerNode.videoNode.assetURL = URL(string: "http://www.sample-videos.com/video/mp4/360/big_buck_bunny_360p_2mb.mp4")
        }
    }
    
    override func layoutSpecThatFits(_ constrainedSize: ASSizeRange) -> ASLayoutSpec {
        return ASRatioLayoutSpec(ratio: 9/16, child: videoPlayerNode)
    }
}

```

maicki:
@MarvinNazari It seems like that this even worked before was pure luck. Looking at the code if you not initialize the player with an assert and set an `assetURL` directly to the `videoNode` it will be blown away in `didEnterPreloadState` as the asset is unequal the asset from the `videoNode`. One way we could go around this is to check if `_asset` is nil in `didEnterPreloadState` or use a delegate method of `ASVideoNode` that will recalled if the asset is loaded and set `_asset` to the asset of the `ASVideoNode`. cc @nguyenhuy

nguyenhuy:
@MarvinNazari With the current API and implementation of `ASVideoPlayerNode`, you're not supposed to grab its video node and change the asset that way. So @maicki is right that it depended on luck to work.

Could you explain your use case a bit more? Why don't you have set `assetURL` to `ASVideoPlayerNode` in init? Do you have the URL available by that time?

MarvinNazari:
@nguyenhuy yeah so we don't have URL available at that time for some of our videos (not alla of them) so we need to do some small fetching/parsing first then we can set the URL

nguyenhuy:
@MarvinNazari OK, then we need to add new asset (and maybe assetURL) setters to ASVideoPlayerNode. The implementations of these methods would simply update the appropriate ivar and then forward to ASVideoNode if the player is already in preload state. Do you have time to put up another PR for it? I can get to it maybe next week but maybe you need it before that.

MarvinNazari:
@nguyenhuy I don't actually have time to fix it (as i'm going of work for 2 weeks from today).
So if you have time to fix it next week, that would be awesome! if not i can try to put up a PR when im back.

nguyenhuy:
Closing in favor of #3135. Thanks, @maicki!

PR #3047, opened 2017-02-18
[IGListKit] Search example

MarvinNazari:
I made a simple example using new IGListKit support in Swift similar to IGListKit search example in their repo.

PR #3094, opened 2017-02-27
Replace IGListKit Method Macros with Actual Methods

Adlai-Holler:
As mentioned in #3047 @MarvinNazari @ay8s 

Replaces the macros with class methods. I think this is a huge improvement.


Adlai-Holler:
Updated!

MarvinNazari:
@Adlai-Holler great! gonna update the search example to use these instead.

PR #3098, opened 2017-02-27
Fix Issues Repopulating Supplementary Elements

### LINKED ISSUE #3096: Cannot Gather Supplementary Kinds in Deleted Sections
https://github.com/facebook/AsyncDisplayKit/blob/master/Source/Details/ASDataController.mm#L580

We need to just blow them all away rather than trying to requery.

Adlai-Holler:
Resolves #3096 

The repopulation business is wild! We need to be careful what are old section indexes and new section indexes.


Adlai-Holler:
This makes the sample project added by @MarvinNazari work again

PR #3099, opened 2017-02-27
[ASCellNode] New way to measure ASCellNode's if layout was invalidated

maicki:


Source/Details/ASDataController.mm
```diff
@@ -617,6 +617,28 @@ - (void)_insertElementsIntoMap:(ASMutableElementMap *)map
 
 #pragma mark - Relayout
 
+- (void)relayoutNodes:(id<NSFastEnumeration>)nodes nodesSizeChanged:(NSMutableArray **)nodesSizesChanged
```
Adlai-Holler: We should use either `(out NSArray **)` or `(NSMutableArray *)` for the second param.

PR #3101, opened 2017-02-28
With CALayer, Use Its Internal Storage Instead of objc_associatedObject

Adlai-Holler:
This works since CALayer will store arbitrary key-value pairs. @appleguy you have the most context around this system. Thoughts?


Adlai-Holler:
Nah let's skip this for now. We need to profile to make a meaningful decision here – less code vs less changes is an even split.

Issue #3082, opened 2017-02-26
[ASRangeController] Union all rects to make only one call to layoutAttributesForElementsInRect:


appleguy:
UICollectionViewLayout overhead, especially creating layout attributes, has always been the most significant portion of ASRangeController's cost.  It is a much smaller cost than the many savings provided by the Collection, Table, and Pager — but their main thread footprint can contribute to frame drops on the slowest devices.

I realized this can be substantially reduced.  Instead of making 3 calls for Visible, Display, and Preload — we should union all these rects to create one large one.

- Obtain the scroll coordinates rect for each range. This depends only on bounds.origin (contentOffset) and bounds.size (viewport size), and the tuning parameters for each range.
- Union all of these rects and request the layoutAttributes for this one rect.
- With all the layoutAttributes in hand, iterate over each layoutAttribute (once).
- For each layoutAttribute, intersect its frame with ALL of the range rects, one at a time.
- For each rect it has a nonzero intersection with, add the attributes.indexPath to a set.

I think this change would largely have to be in ASCollectionLayoutController.  Although it would affect internal API, it should be a remarkably straightforward change for the magnitude of the win it can provide.  From my testing, this could save almost 10% of all time spent on the main thread.

appleguy:
@maicki @Adlai-Holler @nguyenhuy, since every one of us has worked on ASRangeController, I'm curious for your very quick thoughts: does this seem possible, or am I missing something?

Adlai-Holler:
LOL I implemented this over the weekend, I'll open a diff soon.

appleguy:
@Adlai-Holler ha, awesome cleanup!  Massive velocity :).

I also ended up doing this the day after filing...not sure which change is more efficient, but let's compare notes.

Adlai-Holler:
Resolved in #3102

PR #3102, opened 2017-02-28
[ASRangeController] Optimize calls into UICollectionViewLayout via union rect technique.

appleguy:
Details in https://github.com/facebook/AsyncDisplayKit/issues/3082

### DIFF
--- a/Source/Details/ASAbstractLayoutController.h
+++ b/Source/Details/ASAbstractLayoutController.h
@@ -31,6 +31,8 @@ ASDISPLAYNODE_EXTERN_C_END
 
 - (NSSet *)indexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(ASLayoutRangeMode)rangeMode rangeType:(ASLayoutRangeType)rangeType __unavailable;
 
+- (void)allIndexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(ASLayoutRangeMode)rangeMode displaySet:(NSSet * _Nullable * _Nullable)displaySet preloadSet:(NSSet * _Nullable * _Nullable)preloadSet __unavailable;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Source/Details/ASCollectionViewLayoutController.mm
+++ b/Source/Details/ASCollectionViewLayoutController.mm
@@ -53,6 +53,50 @@ typedef struct ASRangeGeometry ASRangeGeometry;
   return [self indexPathsForItemsWithinRangeBounds:rangeBounds];
 }
 
+- (void)allIndexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(ASLayoutRangeMode)rangeMode displaySet:(NSSet **)displaySet preloadSet:(NSSet **)preloadSet
+{
+  if (displaySet == NULL || preloadSet == NULL) {
+    return;
+  }
+  
+  ASRangeTuningParameters displayParams = [self tuningParametersForRangeMode:rangeMode rangeType:ASLayoutRangeTypeDisplay];
+  ASRangeTuningParameters preloadParams = [self tuningParametersForRangeMode:rangeMode rangeType:ASLayoutRangeTypePreload];
+  CGRect displayBounds = [self rangeBoundsWithScrollDirection:scrollDirection rangeTuningParameters:displayParams];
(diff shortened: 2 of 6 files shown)


appleguy:
@Adlai-Holler although this no longer merges cleanly (my branch point was a full 3 days ago!), I'd be happy to clean it up so it does merge cleanly onto current master.  I'm putting it up on its current base to get your input on whether it's equal or better to your current implementation, in case it saves you some polish time.

I've tested this and it seems to work quite well (including some profile-guided optimization; we're down from 11.4% of all main thread time being spent in this method, to 9.0%).

AsyncDisplayKit/Details/ASAbstractLayoutController.h
```diff
@@ -35,6 +35,8 @@ ASDISPLAYNODE_EXTERN_C_END
 
 - (NSSet *)indexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(ASLayoutRangeMode)rangeMode rangeType:(ASLayoutRangeType)rangeType __unavailable;
 
+- (void)allIndexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(ASLayoutRangeMode)rangeMode displaySet:(NSSet * _Nullable * _Nullable)displaySet preloadSet:(NSSet * _Nullable * _Nullable)preloadSet __unavailable;
```
appleguy: The compiler actually required I put in the _Nullable _Nullable or it would not assume I had any specifier there, even though there is a NONNULL assumption here.  I could clean this up somehow before landing.

AsyncDisplayKit/Details/ASRangeController.mm
```diff
@@ -255,28 +247,49 @@ - (void)_updateVisibleNodeIndexPaths
...
+                              && emptyDisplayRange == NO);
+
+  // Check if both Display and Preload are unique. If they are, we load them with a single fetch from the layout controller for performance.
+  BOOL optimizedLoadingOfBothRanges = (equalDisplayPreload == NO && equalDisplayVisible == NO && emptyDisplayRange == NO);
+
+  NSSet<NSIndexPath *> *visibleIndexPaths = [NSSet setWithArray:visibleNodePaths];
+  NSSet<NSIndexPath *> *displayIndexPaths = nil;
+  NSSet<NSIndexPath *> *preloadIndexPaths = nil;
+  
+  if (optimizedLoadingOfBothRanges) {
```
appleguy: Though a bit complex on the surface, I quite like this code, and think it is pretty readable if you parse & trust the meaning of the BOOLs.

This method should be refactored in general, and I could move some of this to another method in ASRangeController, but for now there is not much net complexity added -- just a few more lines to give more descriptive names to the BOOLs.

AsyncDisplayKit/Details/ASTableLayoutController.m
```diff
@@ -64,6 +64,17 @@ - (NSSet *)indexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(
...
 }
 
+- (void)allIndexPathsForScrolling:(ASScrollDirection)scrollDirection rangeMode:(ASLayoutRangeMode)rangeMode displaySet:(NSSet **)displaySet preloadSet:(NSSet **)preloadSet
+{
+    if (displaySet == NULL || preloadSet == NULL) {
+    return;
+  }
+
+  *displaySet = [self indexPathsForScrolling:scrollDirection rangeMode:rangeMode rangeType:ASLayoutRangeTypeDisplay];
+  *preloadSet = [self indexPathsForScrolling:scrollDirection rangeMode:rangeMode rangeType:ASLayoutRangeTypePreload];
```
appleguy: Of course, this method could have a more efficient union implementation as well.  But I figured this could wait for the new async flow layout, and as-implemented it would be tough since we need the frames of each item to decompose them from the unioned rect.  It may not even make sense to do here because this class is pretty fast at generating the paths for each rect uniquely.

Adlai-Holler:
Test failure is spurious:
```
/examples/ASViewController/Pods/Target Support Files/PINRemoteImage/PINRemoteImage-prefix.pch:2:9: could not acquire lock file for module 'UIKit': failed to create link /UIKit-1V5UHAPTOD24G.pcm.lock to UIKit-1V5UHAPTOD24G.pcm.lock-cc319ffb: No such file or directory
```

Jenkins retest this please

appleguy:
Thanks so much for getting this merged!!  Very helpful since I wasn't as familiar with the changes.

PR #3105, opened 2017-02-28
Rejigger Range Controller Rect Gathering

### LINKED ISSUE #3082: [ASRangeController] Union all rects to make only one call to layoutAttributesForElementsInRect:

Adlai-Holler:
Resolves #3082

### DIFF
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -651,15 +651,12 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 
 - (NSArray<ASCellNode *> *)visibleNodes
 {
-  NSArray *indexPaths = [self visibleNodeIndexPathsForRangeController:_rangeController];
+  NSArray *indexPaths = [self indexPathsForVisibleRows];
   
   NSMutableArray<ASCellNode *> *visibleNodes = [NSMutableArray array];
   for (NSIndexPath *indexPath in indexPaths) {
     ASCellNode *node = [self nodeForRowAtIndexPath:indexPath];
-    if (node) {
-      // It is possible for UITableView to return indexPaths before the node is completed.
-      [visibleNodes addObject:node];
-    }
+    [visibleNodes addObject:node];
   }
   
   return visibleNodes;
@@ -1366,63 +1363,9 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
     return _rangeController;
 }
 
-- (NSArray *)visibleNodeIndexPathsForRangeController:(ASRangeController *)rangeController
-{
-  ASDisplayNodeAssertMainThread();
-  
-  CGRect bounds = self.bounds;
-  // Calling indexPathsForVisibleRows will trigger UIKit to call reloadData if it never has, which can result
-  // in incorrect layout if performed at zero size.  We can use the fact that nothing can be visible at zero size to return fast.
-  if (CGRectIsEmpty(bounds)) {
-    return @[];
-  }
-
-  NSMutableArray *visibleIndexPaths = [self.indexPathsForVisibleRows mutableCopy];
-
-  [visibleIndexPaths sortUsingSelector:@selector(compare:)];
-
-  // In some cases (grouped-style tables with particular geometry) indexPathsForVisibleRows will return extra index paths.
-  // This is a very serious issue because we rely on the fact that any node that is marked Visible is hosted inside of a cell,
-  // or else we may not mark it invisible before the node is released. See testIssue2252.
-  // Calling indexPathForCell: and cellForRowAtIndexPath: are both pretty expensive – this is the quickest approach we have.
-  // It would be possible to cache this NSPredicate as an ivar, but that would require unsafeifying self and calling @c bounds
-  // for each item. Since the performance cost is pretty small, prefer simplicity.
-  if (self.style == UITableViewStyleGrouped && visibleIndexPaths.count != self.visibleCells.count) {
-    [visibleIndexPaths filterUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(NSIndexPath *indexPath, NSDictionary<NSString *,id> * _Nullable bindings) {
-      return CGRectIntersectsRect(bounds, [self rectForRowAtIndexPath:indexPath]);
-    }]];
-  }
-
-  NSIndexPath *pendingVisibleIndexPath = _pendingVisibleIndexPath;
-  if (pendingVisibleIndexPath == nil) {
-    return visibleIndexPaths;
-  }
-
-  BOOL isPendingIndexPathVisible = (NSNotFound != [visibleIndexPaths indexOfObject:pendingVisibleIndexPath inSortedRange:NSMakeRange(0, visibleIndexPaths.count) options:kNilOptions usingComparator:^(id  _Nonnull obj1, id  _Nonnull obj2) {
-    return [obj1 compare:obj2];
-  }]);
-  
-  if (isPendingIndexPathVisible) {
-    _pendingVisibleIndexPath = nil; // once it has shown up in visibleIndexPaths, we can stop tracking it
-  } else if ([self isIndexPath:visibleIndexPaths.firstObject immediateSuccessorOfIndexPath:pendingVisibleIndexPath]) {
-    [visibleIndexPaths insertObject:pendingVisibleIndexPath atIndex:0];
-  } else if ([self isIndexPath:pendingVisibleIndexPath immediateSuccessorOfIndexPath:visibleIndexPaths.lastObject]) {
-    [visibleIndexPaths addObject:pendingVisibleIndexPath];
-  } else {
-    _pendingVisibleIndexPath = nil; // not contiguous, ignore.
-  }
-  return visibleIndexPaths;
-}
-
 - (ASScrollDirection)scrollDirectionForRangeController:(ASRangeController *)rangeController
 {
   return self.scrollDirection;
-}
-
-- (CGSize)viewportSizeForRangeController:(ASRangeController *)rangeController
-{
-  ASDisplayNodeAssertMainThread();
-  return self.bounds.size;
 }
 
 - (ASInterfaceState)interfaceStateForRangeController:(ASRangeController *)rangeController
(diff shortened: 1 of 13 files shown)


Tests/ASTableViewTests.mm
```diff
@@ -772,7 +772,12 @@ - (void)testIssue2252
     }];
   });
   dispatch_resume(timer);
-  [self waitForExpectationsWithTimeout:60 handler:nil];
+  // Even if an exception is thrown, we need to cancel the timer so we don't pollute other test methods
+  @try {
+    [self waitForExpectationsWithTimeout:60 handler:nil];
+  } @finally {
+    dispatch_cancel(timer);
+  }
```
Adlai-Holler: When I was debugging this diff, I saw that this method would pollute other tests if an exception was thrown during the updating, so I added this in to ensure that we only fail once.

Source/ASTableView.mm
```diff
@@ -651,15 +651,12 @@ - (nullable NSIndexPath *)indexPathForNode:(ASCellNode *)cellNode waitingIfNeede
 
 - (NSArray<ASCellNode *> *)visibleNodes
 {
-  NSArray *indexPaths = [self visibleNodeIndexPathsForRangeController:_rangeController];
+  NSArray *indexPaths = [self indexPathsForVisibleRows];
   
   NSMutableArray<ASCellNode *> *visibleNodes = [NSMutableArray array];
   for (NSIndexPath *indexPath in indexPaths) {
     ASCellNode *node = [self nodeForRowAtIndexPath:indexPath];
-    if (node) {
```
appleguy: Do you have confidence that this change is safe?  As one data point, there were some very strange behaviors found that this code worked around; although we shouldn't fear any possible regression, I think protecting against an insert: nil crash is pretty reasonable here.

Maybe add an assertion that node should never be nil, and if it is never reported by a developer who makes more active use of ASTableView (notably @eanagel / Yahoo Messenger), then we're probably safe? :)

Source/ASTableView.mm
```diff
@@ -1366,65 +1363,11 @@ - (ASRangeController *)rangeController
     return _rangeController;
 }
 
-- (NSArray *)visibleNodeIndexPathsForRangeController:(ASRangeController *)rangeController
```
appleguy: This code was also added for similar reasons.  At one point @eanagel had a reproducible test case for the scenario this addressed.

Could we split these changes out into a diff that is separate from the optimization, to be able to better understand the cause of any ASTableView-specific regressions?  I'm thinking we should be cautious here mainly because the largest clients are not using ASTableView very much, especially in dynamic / editing environments where this issue was occurring.

These changes were able to be avoided in https://github.com/facebook/AsyncDisplayKit/pull/3102, so I think it will work from a technical perspective to separate them.

Source/Details/ASCollectionViewLayoutController.mm
```diff
@@ -42,45 +40,48 @@ - (instancetype)initWithCollectionView:(ASCollectionView *)collectionView
...
+  
+  NSMutableSet *preloadIndexPaths = [NSMutableSet set];
+  NSMutableSet *visibleIndexPaths = [NSMutableSet set];
+  NSMutableSet *displayIndexPaths = [NSMutableSet set];
+  for (UICollectionViewLayoutAttributes *attr in rawAttributes) {
+    CGRect frame = attr.frame;
+    NSIndexPath *indexPath = attr.indexPath;
+    if (CGRectIntersectsRect(frame, preloadRect)) {
+      [preloadIndexPaths addObject:indexPath];
+      if (CGRectIntersectsRect(displayRect, frame)) {
```
appleguy: This nesting logic unfortunately isn't valid.  Although it is rare, it is technically valid to have ranges that don't nest within one another.  Because all of the other code is careful to support this, we should continue doing so for now.

Source/Details/ASLayoutController.h
```diff
@@ -18,30 +18,17 @@ NS_ASSUME_NONNULL_BEGIN
...
-
-- (void)setVisibleNodeIndexPaths:(NSArray<NSIndexPath *> *)indexPaths;
-
-- (void)setViewportSize:(CGSize)viewportSize;
-- (CGSize)viewportSize;
+- (void)indexPathsForScrolling:(ASScrollDirection)scrollDirection
+                     rangeMode:(ASLayoutRangeMode)rangeMode
+             visibleIndexPaths:(out NSSet<NSIndexPath *> * _Nullable * _Nonnull)visibleIndexPaths
+             displayIndexPaths:(out NSSet<NSIndexPath *> * _Nullable * _Nonnull)displayIndexPaths
+             preloadIndexPaths:(out NSSet<NSIndexPath *> * _Nullable * _Nonnull)preloadIndexPaths;
```
appleguy: Nice "out" annotation!

Source/Details/ASRangeController.mm
```diff
@@ -211,40 +208,13 @@ - (void)_updateVisibleNodeIndexPaths
   
   ASElementMap *map = [_dataSource elementMapForRangeController:self];
 
-  // TODO: Consider if we need to use this codepath, or can rely on something more similar to the data & display ranges
-  // Example: ... = [_layoutController indexPathsForScrolling:scrollDirection rangeType:ASLayoutRangeTypeVisible];
-  NSArray<NSIndexPath *> *visibleNodePaths = [_dataSource visibleNodeIndexPathsForRangeController:self];
```
appleguy: I considered removing this in my diff, but for the historical reasons mentioned, it does add more risk.  I think we need better ASTableView tests before taking on this particular change.

The TODO written here is something I actually did attempt to do, and failed; it had to be reverted.  The comment didn't describe it, but the issue amounted to something like this: not all methods like rectForRowAtIndexPath: actually behave correctly until after a runloop has turned following edit transactions, yet there is some special handling of visibleIndexPaths that makes it more accurate than these other methods.  Then this was combined with some other workarounds as seen in ASTableView.mm

Source/Details/ASRangeController.mm
```diff
@@ -211,40 +208,13 @@ - (void)_updateVisibleNodeIndexPaths
...
-  ASWeakSet *newVisibleNodes = [[ASWeakSet alloc] init];
-
-  if (visibleNodePaths.count == 0) { // if we don't have any visibleNodes currently (scrolled before or after content)...
-    [self _setVisibleNodes:newVisibleNodes];
-    return; // don't do anything for this update, but leave _rangeIsValid == NO to make sure we update it later
-  }
-  ASProfilingSignpostStart(1, self);
```
appleguy: It is of course important that the signpost is started before the expensive indexPath gathering work, so this should be moved back.

Source/Details/ASRangeController.mm
```diff
@@ -253,30 +223,26 @@ - (void)_updateVisibleNodeIndexPaths
...
-                                                          rangeMode:rangeMode
-                                                          rangeType:ASLayoutRangeTypePreload];
-  }
   
-  ASRangeTuningParameters parametersDisplay = [_layoutController tuningParametersForRangeMode:rangeMode
-                                                                                    rangeType:ASLayoutRangeTypeDisplay];
-  if (rangeMode == ASLayoutRangeModeLowMemory) {
-    displayIndexPaths = [NSSet set];
-  } else if (ASRangeTuningParametersEqualToRangeTuningParameters(parametersDisplay, ASRangeTuningParametersZero)) {
-    displayIndexPaths = visibleIndexPaths;
```
appleguy: This deletes some fast paths that are very valuable performance optimizations, by avoiding extra set assembly & object manipulations when some sets are truly identical.  In the most main-thread-performance sensitive views, it's always made a big difference to set the tuning parameters to the same value for display and preload.

Although the core of this change does help bring down the time for generating two unique sets, I re-measured the benefit of this set-sharing after my most recent change and so included the more descriptive BOOL cases to handle this.

Because ASRangeController is, essentially, the core of AsyncDisplayKit — it's crucial that we aggressively optimize this path.  Most importantly, this path currently occurs during the same runloop as other very large operations like [UICollectionView layoutSubviews], so each ms tends to contribute directly to additional frame drops.  Combined with the high frequency of the method being called, it is still responsible for >10% of all main thread time (and a greater proportion of "time over 16.7ms")

Source/Details/ASTableLayoutController.m
```diff
@@ -32,169 +30,45 @@ - (instancetype)initWithTableView:(UITableView *)tableView
...
-    if (lastSectionWithAnyRows < 0) {
-      return nil;
-    }
-    rowCount = [_tableView numberOfRowsInSection:lastSectionWithAnyRows];
-  }
-  return [NSIndexPath indexPathForRow:rowCount - 1 inSection:lastSectionWithAnyRows];
-}
-
-// Same as valueForKeyPath:@"@min.self" but faster
-+ (nullable id)fastArrayMin:(NSArray *)array ASDISPLAYNODE_CONST
```
appleguy: @Adlai-Holler Could you add some diff comments to this change?  It's not immediately apparent why so much code is deleted here, and whether you were able to confirm it didn't regress performance to do so.  Since this code works, and is essentially end-of-life'd / slated for replacement soon, I think we should leave it in place unless the replacement has parity-or-better performance.

Adlai-Holler:
I'll decompose this into smaller diffs so they're easier to review and talk about.

PR #3107, opened 2017-02-28
[ASDisplayNode] Don't protect contentsScaleForDisplay with thread affinity

garrettmoon:
It has locking and is backed by an ivar, not UIKit.


appleguy:
@garrettmoon  in theory, this code should probably call Seth needs display… Let's not add that for now though, as it would almost certainly have performance implications?

You could check if there is an overridden display method in that case though, perhaps that is safe to do ... would be a lot less costly than trampolining a call for every node.

 I assume we set this only through the instance variable during initialization, and if there are no other calls to the setter property in the framework, it is probably safe for performance to add the gated check.

PR #3116, opened 2017-03-01
[ASStackLayoutSpec] Utilize multiple threads if the stack algorithm runs on main

nguyenhuy:
This will be helpful in many use cases, such as:
- When a node needs to force a layout pass on main before a view controller transition can start.
- Relayout all nodes.
- An upcoming async collection view flow layout, backed by a multi-line multi-threaded stack, will be able to determine constrained size of all cell nodes and drive their layout measurements on multiple threads.

Before:
before - without multi-threaded stack

After:
after - with multi-threaded stack

### DIFF
--- a/Source/Private/Layout/ASStackUnpositionedLayout.mm
+++ b/Source/Private/Layout/ASStackUnpositionedLayout.mm
@@ -13,8 +13,10 @@
 #import <tgmath.h>
 #import <numeric>
 
+#import <AsyncDisplayKit/ASDispatch.h>
 #import <AsyncDisplayKit/ASLayoutSpecUtilities.h>
 #import <AsyncDisplayKit/ASLayoutElementStylePrivate.h>
+#import <AsyncDisplayKit/ASThread.h>
 
 CGFloat const kViolationEpsilon = 0.01;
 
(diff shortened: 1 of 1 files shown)


appleguy:
I am so excited you did this, what a surprise!!!

 It has been at least a year since we first discussed this approach together, but I am so glad and truly excited to try this out now that you have created it :)

nguyenhuy:
@appleguy Yes, the idea has been haunted me at nights since then! We hosted a performance day yesterday and this is the first thing came to my mind! Definitely let me know your findings. 

My profiling showed that laying out a stack on main is so fast now, it makes async transition a tiny bit less relevant and buys us more time to overhaul the transition system.

nguyenhuy:
One more profiling. This time I modified our `example/ASCollectionView` to replace `UICollectionViewFlowLayout` with a new `ASCollectionViewLayout` that internally uses a multiline `ASStackLayoutSpec`. Stacking up 1000 text nodes took ~323ms on an iPhone 5S prior to this diff. With this diff applied, it took ~31ms!

Before:
before - ascollectionview

After:
after - ascollectionview

[ASCollectionView.zip](https://github.com/facebook/AsyncDisplayKit/files/814493/ASCollectionView.zip)

PR #3119, opened 2017-03-01
[Layout] Use rectForRowAtIndexPath: to check if a cell node size changed

maicki:
Another step to simplify our layout code within `ASDisplayNode`. After we landed the new handling of `ASCellNode` if a root node needs to do a layout pass now we try to remove the clutter around the layout transitions and how they are handling currently.

The big thing here is that instead of using the bounds of the node we will call `rectForRowAtIndexPath:` or grab the layout element in collection view to get the current height of the cell and determine based on that if we have to trigger a measure pass.

### DIFF
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -1695,6 +1695,22 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
   }
 }
 
+- (BOOL)dataController:(ASDataController *)dataController presentedSizeForElement:(ASCollectionElement *)element matchesSize:(CGSize)size
+{
+  NSIndexPath *indexPath = [self indexPathForNode:element.node];
+  CGRect rect = [self rectForRowAtIndexPath:indexPath];
+  
+  /**
+   * Weirdly enough, Apple expects the return value in tableView:heightForRowAtIndexPath: to _include_ the height
+   * of the separator, if there is one! So if rectForRow would return 44.0 we need to use 43.5.
+   */
+  if (self.separatorStyle != UITableViewCellSeparatorStyleNone) {
+    rect.size.height -= 1.0 / ASScreenScale();
+  }
+
+  return (fabs(rect.size.height - size.height) < FLT_EPSILON);
+}
+
 #pragma mark - ASDataControllerEnvironmentDelegate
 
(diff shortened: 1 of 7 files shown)


Source/Details/ASDataController.h
```diff
@@ -65,6 +65,11 @@ extern NSString * const ASCollectionInvalidUpdateException;
  */
 - (NSUInteger)numberOfSectionsInDataController:(ASDataController *)dataController;
 
+/**
+ Fetch the rect of a cell for at a given index path
+ */
+- (CGRect)dataController:(ASDataController *)dataController rectForRowAtIndexPath:(NSIndexPath *)indexPath;
```
Adlai-Holler: To avoid confusion, we should make this method more clearly related to "what is visible" and we should support supplementary elements too. Maybe:

`- (CGSize)dataController:(ASDataController *)dataController committedSizeForElement:(ASCollectionElement *)element`

Source/ASTableView.mm
```diff
@@ -1695,6 +1695,11 @@ - (NSUInteger)numberOfSectionsInDataController:(ASDataController *)dataControlle
   }
 }
 
+- (CGRect)dataController:(ASDataController *)dataController rectForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+  return [self rectForRowAtIndexPath:indexPath];
```
Adlai-Holler: We should be careful to subtract the separator height here if needed (look at `tableView:heightForRowAtIndexPath:` for info).

maicki:
@Adlai-Holler Can you give it another look please

Source/ASTableView.mm
```diff
@@ -1695,6 +1695,38 @@ - (NSUInteger)numberOfSectionsInDataController:(ASDataController *)dataControlle
...
+   * of the separator, if there is one! So if rectForRow would return 44.0 we need to use 43.5.
+   */
+  if (self.separatorStyle != UITableViewCellSeparatorStyleNone) {
+    rect.size.height -= 1.0 / ASScreenScale();
+  }
+
+  return (fabs(rect.size.height - size.height) < FLT_EPSILON);
+}
+
+- (CGSize)dataController:(ASDataController *)dataController committedSizeForElement:(nonnull ASCollectionElement *)element
```
Adlai-Holler: Remove this old one

PR #3120, opened 2017-03-01
Don't Load Table/Collection Node Just Because User Updated

Adlai-Holler:
If the node isn't loaded, then we haven't fetched any data yet so just ignore these updates.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -499,12 +499,22 @@
 
 - (void)performBatchAnimated:(BOOL)animated updates:(void (^)())updates completion:(void (^)(BOOL))completion
 {
-  [self.view performBatchAnimated:animated updates:updates completion:completion];
+  ASDisplayNodeAssertMainThread();
+  if (self.nodeLoaded) {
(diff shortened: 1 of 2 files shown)


Source/ASCollectionNode.mm
```diff
@@ -499,12 +499,22 @@ - (void)registerSupplementaryNodeOfKind:(NSString *)elementKind
...
+    if (completion) {
+      completion(YES);
+    }
+  }
 }
 
 - (void)performBatchUpdates:(void (^)())updates completion:(void (^)(BOOL))completion
 {
-  [self.view performBatchUpdates:updates completion:completion];
+  [self performBatchAnimated:YES updates:updates completion:completion];
```
nguyenhuy: Shouldn't we grab the `animated` param from `[UIView areAnimationsEnabled]` like we are currently doing in `ASCollectionView` ([here](https://github.com/facebook/AsyncDisplayKit/blob/master/Source/ASCollectionView.mm#L789)).

Adlai-Holler: Ah yes! Good call!

Issue #2762, opened 2016-12-13
Taking a screenshot of the window no longer works on master


flovouin:
Hi guys,

As mentioned on Slack, I discovered an issue while trying to upgrade from `4355f4d` to `2.0`. I just tried on `master` (`ab5627`) and the problem is still here. There is a great probability that this was introduced in `5b80a6`.

The issue is that taking a screenshot of a view (the main window actually) no longer works and returns a black image. To take the screenshot I use a (popular?) piece of code found around the interwebs:
```
UIGraphicsBeginImageContextWithOptions(view.bounds.size, true, 0.0)
view.drawViewHierarchyInRect(view.bounds, afterScreenUpdates: true)
let screenshot = UIGraphicsGetImageFromCurrentImageContext()!
UIGraphicsEndImageContext()
```
In my case, `view` is the `keyWindow` of the `sharedApplication`. I take the screenshot on the main thread, in a `viewDidLoad` of an `ASViewController` that appears short after.

Using `4355f4d`, this works perfectly and returns the expected image.

Please tell me if you need more details or a test project.

Cheers,

Flo

flovouin:
Hi @Adlai-Holler 
I still have the problem in my app, however I was unable to reproduce the issue in a simple sample app. As mentioned above, I am trying to take a screenshot of the whole window. This window contains regular `ASDisplayNode`s, as well as an `ASCollectionNode` with many `GLKView`s.
I can confirm that taking a screenshot works with 016d99 and returns a blank (black?) image with 5b80a6 onwards.
To my knowledge, I do not make use of rasterization in my app.
Would you have any pointers on what I could check to find the issue and solve this?
Cheers,
Flo

rankinit:
Seeing the same issue.

`public extension UIImage {
  public convenience init(node: ASDisplayNode) {
    UIGraphicsBeginImageContext(node.frame.size)
    node.layer.renderInContext(UIGraphicsGetCurrentContext()!)
    let image = UIGraphicsGetImageFromCurrentImageContext()
    UIGraphicsEndImageContext()
    self.init(CGImage: image!.CGImage!)
  }
  
}`

The ASImageNode and ASTextNode are missing, but ASDiplayNode's UIViews do show up in the screenshot.  I'm only taking a screenshot of a select View/Node and its SubNodes.

garrettmoon:
[...]

PR #3121, opened 2017-03-02
[ASDisplayNode] Fix node is showing up as black box if wrapping an UIActivityIndicatorView

maicki:
Set the node to not opaque if it’s wrapping an UIActivityIndicatorView. It's reproducible in the Swift example.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -587,10 +587,18 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     view = [[_viewClass alloc] init];
   }
   
-  // Update flags related to special handling of UIImageView layers. More details on the flags
-  if (_flags.synchronous && ([_viewClass isSubclassOfClass:[UIImageView class]] || [_viewClass isSubclassOfClass:[UIActivityIndicatorView class]])) {
-    _flags.canClearContentsOfLayer = NO;
-    _flags.canCallSetNeedsDisplayOfLayer = NO;
+  // Special handling of wrapping UIKit components
+  if (_flags.synchronous) {
+    // UIImageView layers. More details on the flags
+    if ([_viewClass isSubclassOfClass:[UIImageView class]]) {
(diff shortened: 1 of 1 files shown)


garrettmoon:
@maicki I never thought I'd live to see the day you'd write code like this ;)

Source/ASDisplayNode.mm
```diff
@@ -587,10 +587,18 @@ - (UIView *)_viewToLoad
...
   }
   
-  // Update flags related to special handling of UIImageView layers. More details on the flags
-  if (_flags.synchronous && ([_viewClass isSubclassOfClass:[UIImageView class]] || [_viewClass isSubclassOfClass:[UIActivityIndicatorView class]])) {
-    _flags.canClearContentsOfLayer = NO;
-    _flags.canCallSetNeedsDisplayOfLayer = NO;
+  // Special handling of wrapping UIKit components
+  if (_flags.synchronous) {
+    // UIImageView layers. More details on the flags
+    if ([_viewClass isSubclassOfClass:[UIImageView class]]) {
```
garrettmoon: Should this also check UIActivityIndicator? Or was the old cold wrong?

maicki: The old was wrong. This is not necessary for `UIIndicatorView`

flovouin:
Hi there, sorry for intruding ^^
I think it might also be the case for `UIVisualEffectView`. I had misdiagnosed an issue in #2762 with similar symptoms: getting a black box instead of the blurred background I was expecting. I had also noticed the black box for the `UIActivityIndicatorView`. In both cases, my workaround was to set the background colour of the node to `clear`.
I haven't tried to apply the changes in this PR to the visual effect view and I might be wrong, but it might be worth a look :-)

maicki:
Ok, maybe this get's slowly out of hand, but @flovouin was right and it's also necessary for `UIVisualEffectView`. I updated the diff and will merge after CI passes.

Furthermore let's close #2762 as it seems this will resolves this.

PR #3124, opened 2017-03-02
Majorly Simplify ASTableLayoutController

Adlai-Holler:
We previously:

- Waited for `setVisibleNodexIndexPaths:`.
- Found the min row and max row.
- Measured the range bounds distance from the min row and max row.
- Searched in rects at that distance away from the min row and max row, growing our search rect until we found one.

We now:

- Ask the table view for the index paths of the rows in the specified rect.


Adlai-Holler:
Since it's now unused, I removed the internal `ASLayoutController setVisibleNodeIndexPaths:` API as well.

PR #3125, opened 2017-03-02
Remove ASTableView's pendingVisibleIndexPath System

Adlai-Holler:
This was added in #430. Since then it's grown and grown in complexity probably due to other issues encountered along the way. 😱 

Back then we would flush the range controller (queued in the run loop), and now we synchronously flush it (if it's invalidated):

- At the end of the layout pass
- After calling endUpdates
- When we enter the window

The original rationale was that "we depend on -indexPathsForVisibleRows to be correct immediately after willDisplayCell:, but it isn't." But I don't think it's true that we depend on it being correct at that time… we depend on it being correct later, asynchronously. Could be interesting to see if `indexPathsForVisibleRows` calls `layoutIfNeeded` or equivalent.

@eanagel What are your thoughts on this? You mentioned that not having it caused issues with programmatic scrolling – do I need to do anything specific to repro?

### DIFF
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -136,8 +136,6 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 
   ASBatchContext *_batchContext;
 
-  NSIndexPath *_pendingVisibleIndexPath;
-
   // When we update our data controller in response to an interactive move,
   // we don't want to tell the table view about the change (it knows!)
   BOOL _updatingInResponseToInteractiveMove;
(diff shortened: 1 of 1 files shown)


eanagel:
TL;DR - Yeah I think you can get rid of this code now.

The original issue was that we expected the newly visible cell to already be in `-indexPathsForVisibleRows` during the call to `-willDisplayCell:forRowAtIndexPath:` which is almost always true, except in certain cases when scrolling programmatically. When I would see this, cells would render as blank when scrolling programmatically. It wouldn't always happen but once a cell was missed it would be blank as long as it was on screen. 

It looks like ASRangeController now delays it's updates to the next run loop. I think this would handle that edge case. And from what I can tell in this code (it's pretty complex) if it wasn't visible at that point we would eventually get it.

@appleguy As an aside we are currently testing code that allows to move to ASCollectionNode instead of ASTableNode. This is due to an amazing custom layout that @master-nevi has put together that performs virtualization without requiring changes to ASCollectionNode/View its self. (By Virtualization I mean that inserts or deletes outside the visible window don't cause any visible change to the user. To do this we adjust the content offset so the visible window doesn't change as it naturally would with a collection or table view.)

Adlai-Holler:
Good enough for me. Thanks @eanagel for the reply – we definitely don't call `indexPathsForVisibleRows` inside `willDisplayCell:` anymore. Cheers!

PR #3130, opened 2017-03-03
Introduce ASCollectionLayout and friends

nguyenhuy:
## v1.0 - Obsolete:

- `ASCollectionViewLayout` is an asynchronous and thread-safe `UICollectionViewLayout` that encapsulates its layout calculation logic into a separate thread-safe object which can be used ahead of time and/or on multiple threads.
- `ASDataController` now can prepare new layouts that are resulted from change sets before `ASCollectionView` even knows about it. By the time a change set is ready to be consumed by `ASCollectionView`, its new layout is also ready.
- A new `ASCollectionViewLayoutCalculating` protocol that is simple and generic enough that many types of calculators can be built on top. `ASCollectionViewLayoutSpecCalculator` conforms to `ASCollectionViewLayoutCalculating` protocol and can be backed by any layout spec (e.g `ASStackLayoutSpec`, etc). We can even build a `ASCollectionViewLayoutYogaCalculator` that uses Yoga internally.
- A built-in `ASCollectionViewFlowLayoutCalculator` that is a subclass of `ASCollectionViewLayoutSpecCalculator` and uses a multi-threaded multi-line `ASStackLayoutSpec` internally. The result is a performant and thread-safe flow layout calculator.
- Finally, `ASCollectionViewLayout` can be subclassed to handle a specific type of calculator with optimizations implemented based on the knowledge of such calculator. For example, `ASCollectionViewFlowLayout` can have a highly optimized implementation of `-layoutAttributesForElementsInRect:`.

## v2.0:
- `ASCollectionLayout` is an asynchronous and thread-safe `UICollectionViewLayout`. It can prepare layouts ahead of time and/or on multiple threads.
- `ASDataController` now can prepare new layouts that are resulted from change sets before `ASCollectionView` even knows about it. By the time a change set is ready to be consumed by `ASCollectionView`, its new layout is also ready.
- `ASDataControllerLayoutDelegate` is a new, simple and generic protocol. If a layout delegate is set to `ASDataController`, it will delegate all layout operations to the delegate. `ASCollectionLayout` conforms to this protocol and handles many threading issues so that its subclasses only need to implement a simple method: `-calculateLayoutForLayoutContext:`. For example, `ASCollectionFlowLayout` is a subclass of `ASCollectionLayout` that uses a multi-line concurrent `ASStackLayoutSpec` to prepare its layout. 
- For further customizations, developers are welcome to subclass `ASCollectionLayout`, or even provide an independent object that conforms to `ASDataControllerLayoutDelegate`, to handle a specific layout with optimizations implemented based on the knowledge of such layout.

## v2.0.1:
- `ASCollectionLayout` is private and expects a layout delegate that conforms to a public `ASCollectionLayoutDelegate` protocol.

## v2.0.2:
- API tweaks. Details [here]( https://github.com/facebook/AsyncDisplayKit/pull/3130/commits/9d1359ad762ae1914deb7f827ceda60c60a97504).

## Issues to be followed up later - not in the scope of this PR:
- [x] Unit tests (https://github.com/TextureGroup/Texture/pull/76): It'd be ideal to have them in this PR. But since the main purpose of this PR is proposing new APIs, the new concrete classes in this PR (e.g `ASCollectionLayout`, `ASCollectionFlowLayoutDelegate`, etc) are incomplete and writing unit tests for them is not critical at this stage.
- [ ] Better implementation of `ASCollectionFlowLayoutDelegate`: As mentioned above, the current implementation is a proof of concept and barely works :)
- [ ] Supplementary views: I think layout delegate should handle this. I'm considering adding a new method in `ASDataControllerLayoutDelegate` that is responsible for adding supplementary elements to an `ASMutableElementMap` parameter.
- [ ] Screen page map: It would be nice to have it implemented as suggested [here](https://github.com/facebook/AsyncDisplayKit/pull/3130#pullrequestreview-26144603).

Ticket: https://github.com/TextureGroup/Texture/issues/186


Source/ASCollectionView.mm
```diff
@@ -1469,6 +1485,16 @@ - (void)_beginBatchFetching
 
 #pragma mark - ASDataControllerSource
 
+- (id<ASCollectionViewLayoutCalculating>)layoutCalculatorInDataController:(ASDataController *)dataController
+{
+  return _collectionViewLayoutFlags.isASCollectionViewLayout ? ((ASCollectionViewLayout *)self.collectionViewLayout).calculator : nil;
```
nguyenhuy: These flag and cast are definitely not ideal and can be improved.

eanagel:
This is very exciting! @master-nevi - check this out.

appleguy:
@nguyenhuy you have an amazing history of seminal contributions to AsyncDisplayKit, including the original ASLayoutSpec and crucial advancements in ASDataController.  This change, along with preparatory work around flex-wrap and concurrent layout in stacks, continues this incredible tradition!

I've dreamt about achieving something like this for years, but it has never been practical due to the huge amount of work behind it.  Along with the equally necessary work from @Adlai-Holler on the collection and @maicki on the layout system, you all have made that dream a reality for thousands of iOS developers!

Flat-out inspiring.  Thank you, @maicki, @Adlai-Holler, @nguyenhuy, and Pinterest for sharing your awesome talents with the community.

Source/ASCollectionView.mm
```diff
@@ -1467,6 +1486,26 @@ - (void)_beginBatchFetching
...
+{
+  return _collectionViewLayoutFlags.providesLayoutCalculator ? [(id<ASCollectionViewLayoutCalculatorProviding>)self.collectionViewLayout collectionViewLayoutCalculator] : nil;
+}
+
+- (CGSize)sizeForCalculatingCollectionViewLayout
+{
+  return _collectionViewLayoutFlags.providesLayoutCalculator ? [(id<ASCollectionViewLayoutCalculatorProviding>)self.collectionViewLayout sizeForCalculatingCollectionViewLayout] : CGSizeZero;
+}
+
+#pragma mark - ASCollectionViewLayoutCalculatorResultConsuming
```
appleguy: The specificity of this name is nice, but it remains pretty abstract as to the purpose / function of this protocol.

Should we drop "View" from all of these names, even though it would be less similar to UICollectionView names?  Any other ideas on how to either simplify, or clarify the names?

This is mostly a question for the other reviewers and we should definitely not bog down this work in naming choice, but we're much less likely to revisit them after landing, so it's a good time to give it some thought.

Source/ASCollectionView.mm
```diff
@@ -1930,6 +1972,11 @@ - (void)layer:(CALayer *)layer didChangeBoundsWithOldValue:(CGRect)oldBounds new
   if (self.collectionViewLayout == nil) {
     return;
   }
+  if (_collectionViewLayoutFlags.providesLayoutCalculator) {
+    // TODO Maybe kick off a background calculation here and hope that the layout is ready by the time -prepareLayout
```
appleguy: One risk of this is multiple bounds changes.  We should probably schedule it to occur at the end of the runloop if anything, but that will probably mean that layout is called by then anyway.  Because the main thread runs at a higher priority, I'm not sure we would gain too much from starting it early -- although there could be other work on the main thread to run before layout is flushed.

The other question is whether there are sometimes offscreen collections that have their bounds set up, but that would be wasteful to actually run the layout for.  We could check for that here, but something to consider.

For the sake of simplicity and ensuring reliability before we introduce new layout kick-off points, it probably makes sense to defer this until testing indicates there are some reasonably-common use cases where this would provide an advantage.

Source/ASPagerFlowLayout.m
```diff
@@ -20,6 +20,7 @@ @interface ASPagerFlowLayout () {
 
 @end
 
+//TODO make this an ASCollectionViewLayout
```
appleguy: This, especially when combined with vertical flow layouts inside the pages, will be a truly revolutionary impact on performance in pin-open and pin-to-pin swiping performance.

Source/ASTableView.mm
```diff
@@ -1581,6 +1581,11 @@ - (void)rangeController:(ASRangeController *)rangeController didUpdateWithChange
 
 #pragma mark - ASDataControllerDelegate
 
+- (BOOL)dataControllerShouldAutomaticallyLayoutNodes:(ASDataController *)dataController
```
appleguy: Does this have a use case yet where NO is used?  If not, should we remove it and add it once a use case exists?

Source/Details/ASCollectionViewFlowLayoutCalculator.m
```diff
@@ -0,0 +1,35 @@
...
+//  Copyright © 2017 Facebook. All rights reserved.
+//
+
+#import <AsyncDisplayKit/ASCollectionViewFlowLayoutCalculator.h>
+#import <AsyncDisplayKit/ASCellNode.h>
+#import <AsyncDisplayKit/ASCollectionElement.h>
+#import <AsyncDisplayKit/ASElementMap.h>
+#import <AsyncDisplayKit/ASStackLayoutSpec.h>
+
+@implementation ASCollectionViewFlowLayoutCalculator
```
appleguy: This class is so impressively elegant.  It's a pure expression of what the coming together of the entire core team's work has produced.

Source/Details/ASCollectionViewFlowLayoutCalculator.m
```diff
@@ -0,0 +1,35 @@
...
+
+- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)sizeRange forElementMap:(ASElementMap *)map
+{
+  NSMutableArray<ASCellNode *> *children = [NSMutableArray array];
```
appleguy: Can the map maintain this array internally?  These operations are pretty inefficient — especially the block-based iteration will create excessive retain / release operations, and also the array being created and destroyed for just one pass.  Imagine cases where there are hundreds of nodes for example, which is common.

Source/Details/ASCollectionViewFlowLayoutCalculator.m
```diff
@@ -0,0 +1,35 @@
...
+  }];
+  
+  ASStackLayoutSpec *stackSpec = [ASStackLayoutSpec stackLayoutSpecWithDirection:ASStackLayoutDirectionHorizontal
+                                                                         spacing:0
+                                                                  justifyContent:ASStackLayoutJustifyContentStart
+                                                                      alignItems:ASStackLayoutAlignItemsStart
+                                                                        flexWrap:ASStackLayoutFlexWrapWrap
+                                                                    alignContent:ASStackLayoutAlignContentStart
+                                                                        children:children];
+  stackSpec.multithreaded = YES;
```
appleguy: different API, but we should decide if we want to use the word multithreaded here.  I think .concurrent is more appropriate; multithreaded harkens back to NSThread, whereas concurrent queues are the GCD terminology.

Source/Details/ASCollectionViewLayout.m
```diff
@@ -0,0 +1,162 @@
...
+@end
+
+@implementation ASCollectionViewLayout
+
+- (instancetype)initWithLayoutCalculator:(id<ASCollectionViewLayoutCalculating>)calculator
+{
+  self = [super init];
+  if (self) {
+    _calculator = calculator;
+    _elementToLayoutAttributesMap = [NSMapTable mapTableWithKeyOptions:(NSMapTableStrongMemory | NSMapTableObjectPointerPersonality) valueOptions:NSMapTableCopyIn];
```
appleguy: Would be great to have a comment explaining the options as they affect this use case -- in particular CopyIn, which is not common to see in my experience.

Source/Details/ASCollectionViewLayout.m
```diff
@@ -0,0 +1,162 @@
...
+  }
+  return self;
+}
+
+- (instancetype)initWithCoder:(NSCoder *)aDecoder
+{
+  return [self initWithLayoutCalculator:[[ASCollectionViewFlowLayoutCalculator alloc] init]];
```
appleguy: Is it allowed that we could just call -init here so the line isn't duplicated?

Source/Details/ASCollectionViewLayout.m
```diff
@@ -0,0 +1,162 @@
...
+    ASLayout *layout = [_calculator layoutThatFits:self.collectionView.bounds.size forElementMap:_elementMap];
+    [self _applyLayout:layout];
+  }
+}
+
+- (void)invalidateLayout
+{
+  [super invalidateLayout];
+  if (_collectionViewIsPendingLayoutUpdate == NO) {
+    _contentSize = CGSizeZero;
```
appleguy: Is this "safe" to do, in other words can we expect that UICollectionView won't call the contentSize method below until the layout is re-done?  I assume so, but just am building my own understanding of this detail.  May be worth a comment as this is a file that a reasonable number of framework users might take a peek at to try to understand.

Source/Details/ASCollectionViewLayout.m
```diff
@@ -0,0 +1,162 @@
...
+
+- (CGSize)collectionViewContentSize
+{
+  _collectionViewIsPendingLayoutUpdate = NO;
+  return _contentSize;
+}
+
+- (NSArray<UICollectionViewLayoutAttributes *> *)layoutAttributesForElementsInRect:(CGRect)rect
+{
+  NSMutableArray *attributesInRect = [NSMutableArray array];
```
appleguy: Initialize with capacity.

Enumerator seems like a great idea relative to the block-based style.  Can you use the "for (... in enumerator)" style?  I thought that was possible, but maybe not.

Source/Details/ASCollectionViewLayout.m
```diff
@@ -0,0 +1,162 @@
...
+}
+
+- (UICollectionViewLayoutAttributes *)layoutAttributesForDecorationViewOfKind:(NSString *)elementKind atIndexPath:(NSIndexPath *)indexPath
+{
+  return [self layoutAttributesForSupplementaryViewOfKind:elementKind atIndexPath:indexPath];
+}
+
+- (BOOL)shouldInvalidateLayoutForBoundsChange:(CGRect)newBounds
+{
+  if (!CGSizeEqualToSize(self.collectionView.bounds.size, newBounds.size)) {
```
appleguy: I'm not sure, but should this use sizeForCalculatingCollectionViewLayout in some way?  Maybe the API should be sizeFor...LayoutWithBounds: to make it more functional based on the bounds, and then we could compare the current size for layout with the size that would be used once the new bounds is applied.

Source/Details/ASCollectionViewLayout.m
```diff
@@ -0,0 +1,162 @@
...
+    NSIndexPath *indexPath = [_elementMap indexPathForElement:element];
+    
+    UICollectionViewLayoutAttributes *attrs;
+    if (element.supplementaryElementKind == nil) {
+      attrs = [UICollectionViewLayoutAttributes layoutAttributesForCellWithIndexPath:indexPath];
+    } else {
+      attrs = [UICollectionViewLayoutAttributes layoutAttributesForSupplementaryViewOfKind:element.supplementaryElementKind withIndexPath:indexPath];
+    }
+    
+    attrs.frame = CGRectMake(sublayout.position.x, sublayout.position.y, sublayout.size.width, sublayout.size.height);
```
appleguy: For both readability and performance, suggest using a CGPoint and CGSize local variable.

Source/Details/ASElementMap.m
```diff
@@ -133,6 +133,17 @@ - (void)enumerateUsingBlock:(void(^)(NSIndexPath *indexPath, ASCollectionElement
   }];
 }
 
+- (NSArray<ASCollectionElement *> *)filteredElementsUsingBlock:(BOOL (^)(NSIndexPath * _Nonnull, ASCollectionElement * _Nonnull, BOOL * _Nonnull))filteringBlock
+{
+  NSMutableArray<ASCollectionElement *> *elements = [NSMutableArray array];
+  [self enumerateUsingBlock:^(NSIndexPath * _Nonnull indexPath, ASCollectionElement * _Nonnull element, BOOL * _Nonnull stop) {
```
appleguy: Would be great to avoid block-based enumeration.  Especially in our highly concurrent system, these retains and releases are the #1 most common impact on performance regressions & wins — this is notable because not only do they show up prominently in time profiles, but especially the system trace shows the runtime lock contention is the biggest cause of incomplete CPU utilization.

cc @Adlai-Holler @maicki to think of ways we can solve this more systematically (if open to the idea, I think a code style with stuff like this _specifically for core code_ could be one step).

This should _obviously_ not block this diff from landing; however I think it is justified to file a followup task now if we'd like to defer addressing it.  We can spend the time to profile it, but it is a time investment to do the system tracing necessary to understand the impacts.

Source/Details/ASCollectionViewLayoutCalculating.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+- (ASLayout *)layoutThatFits:(CGSize)size forElementMap:(ASElementMap *)map;
+
+@end
+
+@protocol ASCollectionViewLayoutCalculatorProviding <NSObject>
+
+- (nullable id<ASCollectionViewLayoutCalculating>)collectionViewLayoutCalculator;
+
+- (CGSize)sizeForCalculatingCollectionViewLayout;
```
appleguy: Is there a way to use other terms in the API, like bounds or perhaps viewport?  Indeed this is not constrainedSize, but it kinda feels that way since the only places we ask for sizes from developers with layout are for constraints — so some way of disambiguating this would be nice.

boundsForCollectionLayout, viewportSizeForCollectionLayout, etc

Source/Details/ASCollectionViewLayoutCalculating.h
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param layout The new layout
+ *
+ * @param calculator The calculator used to calculate this layout
+ *
+ * @param map The element map that was used to calculate the layout
+ *
+ * @note The map might be weakly retained by the consumer for later re-calculations.
+ */
+- (void)setLayout:(ASLayout *)layout calculatedBy:(id<ASCollectionViewLayoutCalculating>)calculator forElementMap:(ASElementMap *)map;
```
appleguy: This would more naturally follow delegate conventions if lead by the calculator, e.g.

layoutCalculator:didFinishLayout:forElementMap:

appleguy (approved):
It seems like there is still work in progress here, but it certainly has my accept for whenever the core team feels ready to merge this landmark!  I'll start trying it out as soon as I hear word of that.

Source/ASTableView.mm
nguyenhuy: Ah no, this is an API I experimented with and forgot to remove :) Removing...

Source/ASCollectionView.mm
nguyenhuy: I like the idea of dropping "View" in these names as well. I'm about to experiment letting `ASTableView` to provide a calculator but doesn't consume the result. The idea is to entirely remove the layout responsibility from ASDataController and let calculators handle it.

Source/ASCollectionView.mm
nguyenhuy: Agreed, definitely not gonna work on this in this PR, if ever. It was an idea I thought might worth considering. Gonna remove the TODO for now.

Source/Details/ASCollectionViewFlowLayoutCalculator.m
nguyenhuy: concurrent it is! Thanks for suggesting!

Source/Details/ASCollectionLayoutCalculating.h
```diff
@@ -0,0 +1,64 @@
...
+ * @abstract Return a layout of elements.
+ *
+ * @param viewportSize The viewport size that the resulting layout should fit, usually equals to the bounds of the containing table/collection view.
+ *
+ * @param map An element map containings all elements to be laid out.
+ *
+ * @discussion This method can be called on multiple theads and thus must be thread-safe (and ideally, stateless).
+ * This method must block the calling thread but can dispatch to other theads to reduce the blocking time.
+ */
+- (ASLayout *)layoutThatFits:(CGSize)viewportSize forElementMap:(ASElementMap *)map;
```
Adlai-Holler: Also, should we pass some kind of context/helper object?

For example, we could remove the two arguments here, and the `sizeRangesShouldBeProvided` flag in the protocol if we gave them something like:

```objc
ASCollectionLayoutContext {
   CGSize viewportSize;
   id<ASCollectionDelegate> collectionDelegate;
   ASElementMap *map;
   - (void)parallelAllocateNodesForElements:(id<NSFastEnumeration> elements);
    // Maybe?
   - (void)setFrame:(CGRect)frame forElement:(ASCollectionElement *)element;
   - (BOOL)isCancelled;
}
```

Source/Details/ASDataController.mm
```diff
@@ -268,12 +263,14 @@ - (void)_layoutNode:(ASCellNode *)node withConstrainedSize:(ASSizeRange)constrai
...
  * @param environment The trait environment needed to initialize elements
  * @param indexPathsAreNew YES if index paths are "after the update," NO otherwise.
+ * @param shouldFetchSizeRanges Whether constrained sizes should be fetched from data source
  */
 - (void)_repopulateSupplementaryNodesIntoMap:(ASMutableElementMap *)map
              forSectionsContainingIndexPaths:(NSArray<NSIndexPath *> *)indexPaths
                                    changeSet:(_ASHierarchyChangeSet *)changeSet
                                  environment:(id<ASTraitEnvironment>)environment
                             indexPathsAreNew:(BOOL)indexPathsAreNew
+                 shouldFetchSizeRanges:(BOOL)shouldFetchSizeRanges
```
Adlai-Holler: Nit: indentation

Source/Details/ASLegacyCollectionLayoutCalculator.h
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Normally layout calculators can determine size range of all collection elements because it is the layout logic.
+ * However, prior to ASCollectionViewLayout, size ranges are provided by async delegate of ASTableView, or async delegate and layout inspector of ASCollectionView.
+ * This calculator is the bridge between the old size range API and the new layout calculating API, hence the name "legacy".
+ * Instead of deciding size ranges on its own, it uses the ones provided in collection elements.
+ *
+ * @note Since this calculator doesn't have a layout logic, it doesn't calculate positions of collection elements.
+ * Thus, the layout returned by the calculator is not meant to be consumed as is.
+ */
+@interface ASLegacyCollectionLayoutCalculator : NSObject <ASCollectionLayoutCalculating>
```
Adlai-Holler: Nit: Apply `AS_SUBCLASSING_RESTRICTED` to new classes where appropriate.

Source/Details/ASLegacyCollectionLayoutCalculator.m
```diff
@@ -0,0 +1,44 @@
...
+
+#import <AsyncDisplayKit/ASCellNode.h>
+#import <AsyncDisplayKit/ASCollectionElement.h>
+#import <AsyncDisplayKit/ASDispatch.h>
+#import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
+#import <AsyncDisplayKit/ASElementMap.h>
+#import <AsyncDisplayKit/ASLayout.h>
+#import <AsyncDisplayKit/ASLayoutSpec.h>
+
+@implementation ASLegacyCollectionLayoutCalculator
```
Adlai-Holler: Instead of adding this, could we say "we have no calculator – we are in 'legacy mode'" and have the data controller go ahead and perform the eager allocation/measurement step as before in a separate codepath? It's a little confusing that this thing doesn't really calculate any layout.

Source/Details/ASCollectionLayoutCalculating.h
Adlai-Holler: (Sorry to spam) Should this object also synchronously decide what supplementary elements exist? Invalidation is a tough question here, but as a first stab:

```objc
- (void)supplementaryElementsForMap:(ASElementMap *)map storage:(id<ASSupplementaryElementStorage>)storage {
   // code
   [storage addSupplementaryElementOfKind:kind atIndexPath:indexPath];
   // code
}
```

We would call that method synchronously after the user issues each update, so that whatever supplementary elements they specify, we can go get the nodeBlocks for them as needed. Invalidating supplementary elements is a tough question and we should probably shelve it for now. 

If we add that, then we can fully-replace the layout inspector.

Source/Details/ASCollectionLayoutSpecCalculator.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param sizeRange The size range that the resulting layout should fit. 
+ * This is the size range returned by -sizeRangeForLayoutSpecThatFits:
+ *
+ * @param map An element map containings all elements to be laid out.
+ *
+ * @discussion This method can be called on multiple theads and thus must be thread-safe (and ideally, stateless).
+ * This method must block the calling thread but can dispatch to other theads to reduce the blocking time.
+ */
+- (ASLayoutSpec *)layoutSpecThatFits:(ASSizeRange)sizeRange forElementMap:(ASElementMap *)map;
```
Adlai-Holler: I think we can eliminate this class and put the burden of doing this on the client, probably. For instance, in the flow layout case, it would amount to:

```objc
- (ASLayout *)layoutWithSize:(CGSize)size elementMap:(ASElementMap *)map
{
   ASLayoutSpec *spec = [ASStackLayoutSpec …];
   // I know I am a vertical layout, so unbounded height: 
   ASSizeRange sizeRange = ASSizeRangeMake(size.width, 0, size.width, INFINITY);
   return [spec layoutThatFits:sizeRange];
}
```

- There's not much extra code.
- Gives the calculator class control over scrollable directions.
- Eliminates a class.

Adlai-Holler (changes requested):
This is a great first-pass! Keep 'er going!

nguyenhuy:
@Adlai-Holler a few updates in this PR:
- Removed ASLegacyCollectionLayoutCalculator and let ASDataController run a slightly different code path if a calculator is not provided.
- Remove ASCollectionLayoutSpecCalculator.
- Calculators now return a new `ASCollectionLayout` object that contains a content size and an element to rect table. Layout attributes will be created on the fly.

A few things left that I need time to think about:
- Node allocation API that allows layout/layout calculator to decide which nodes to allocate.
- ASCollectionLayoutContext object. For simplicity of the first version, let's not add `-setFrame:forElement:` and `-isCancelled`.
- Supplementary elements: Ideally I want the calculator to only care about laying out elements given to it. Maybe we should introduce a new object, say `ASCollectionSupplementaryProvider` that creates all supplementaries needed for a given set of items?

With your trip coming really soon and the new code is non-intrusive enough that existing clients shouldn't be affected, does it make sense to merge this PR as its current state and improve it over time?

Source/Details/ASCollectionViewLayout.h
```diff
@@ -0,0 +1,25 @@
...
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+#import <AsyncDisplayKit/ASCollectionLayoutCalculating.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class ASElementMap, ASLayout;
+
+@interface ASCollectionViewLayout : UICollectionViewLayout <ASCollectionLayoutCalculatorProviding, ASCollectionLayoutCalculatorResultConsuming>
```
Adlai-Holler: This should be `private` and `AS_SUBCLASSING_RESTRICTED`

Source/Details/ASCollectionFlowLayoutCalculator.h
```diff
@@ -0,0 +1,16 @@
...
+//  AsyncDisplayKit
+//
+//  Created by Huy Nguyen on 28/2/17.
+//  Copyright © 2017 Facebook. All rights reserved.
+//
+
+#import <AsyncDisplayKit/ASCollectionLayoutCalculating.h>
+#import <AsyncDisplayKit/ASScrollDirection.h>
+
+@interface ASCollectionFlowLayoutCalculator : NSObject <ASCollectionLayoutCalculating>
```
Adlai-Holler: Nullability, AS_SUBCLASSING_RESTRICTED

Adlai-Holler (commented):
@nguyenhuy Great! Thoughts, sorry if they're messy, let me know what you think:

- It's going to be extremely expensive to create attributes on the fly. The range controller is going to call `layoutAttributesForElementsInRect:` every frame.
  - We could have them return us an array of layout attributes, just like UICollectionViewLayout does.
    - We'd add convenience methods to generate those from, say, an element map + a frame table or from a layout spec.
  - We could simply generate them on-demand and then store them. I think it would be safe to assume that the consumer won't mutate the attributes (UICollectionView won't, range controller won't.)
    - `ASCollectionElement -> UICollectionViewLayoutAttributes`
  - Querying these things quickly is important. UICollectionView internally uses a `screenPageMap` which is `NSMapTable<(x: NSUInteger, y: NSUInteger) -> NSSet<UICollectionViewLayoutAttributes>>` where `x` and `y` are the coordinates of a screen-sized rectangle. By finding all the screen pages that intersect a given rect, uniting the sets of attributes, and then filtering the final result, you can significantly cut down on the amount of time to request chunks of layout.
- Since supplementaries should probably be determined by this same object as well, the name `LayoutCalculator` may be too restrictive for the future, as we think about things like partial layout querying and invalidation etc. `ASCollectionLayoutDelegate` – "the object we talk to about laying out a collection" seems nice.
- In the back of our minds, we may want to look to a future where we query rects-worth of layouts at a time, much like `layoutAttributesForElementsInRect:` or `drawRect:`. We shouldn't bend over backwards to accommodate this in the architecture – since YAGNI – but we should also bear it in mind.
- Let's aim to land this into a Beta header soon, and we can tweak it as we bring it into production.

nguyenhuy:
@Adlai-Holler @appleguy I've updated most of the code in this PR to implement the "v2" API explained in the description of this PR. Please have another look!

Source/Details/ASCollectionLayout+Subclasses.h
```diff
@@ -0,0 +1,39 @@
...
+#import <AsyncDisplayKit/ASDataController+Beta.h>
+
+@class ASCollectionContentAttributes;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ASCollectionLayout () <ASDataControllerLayoutDelegate>
+
+/// The current content, if any. This property must be accessed on main thread.
+@property (nonatomic, strong, nullable) ASCollectionContentAttributes *currentContentAttributes;
```
Adlai-Holler: Are you sure we should expose this? Users can call `-collectionViewContentSize` and `-layoutAttributesForItemAtIndexPath:` etc to get this info, if they need it for some reason.

Source/Details/ASCollectionLayout.mm
```diff
@@ -0,0 +1,135 @@
...
+{
+  ASDisplayNodeAssertMainThread();
+  [super invalidateLayout];
+  _currentContentAttributes = nil;
+}
+
+- (CGSize)collectionViewContentSize
+{
+  ASDisplayNodeAssertMainThread();
+  return _currentContentAttributes.contentSize;
```
Adlai-Holler: Let's assert that `_currentContentAttributes != nil` for good measure here.

Source/Details/ASCollectionLayout.mm
```diff
@@ -0,0 +1,135 @@
...
+  return _currentContentAttributes;
+}
+
+- (void)setCurrentContentAttributes:(ASCollectionContentAttributes *)newAttrs
+{
+  ASDisplayNodeAssertMainThread();
+  if (! ASObjectIsEqual(_currentContentAttributes, newAttrs)) {
+    _currentContentAttributes = newAttrs;
+  }
+}
```
Adlai-Holler: Do we need this setter? Currently isn't called and not public.

Source/Details/ASCollectionLayout.mm
```diff
@@ -0,0 +1,135 @@
...
+  ASDataControllerLayoutContext *_layoutContextForPendingContentAttributes;
+}
+
+@end
+
+@implementation ASCollectionLayout
+
+- (instancetype)init
+{
+  return [super init];
```
Adlai-Holler: Remove this empty override

Source/Details/ASCollectionLayout.mm
```diff
@@ -0,0 +1,135 @@
...
+
+- (CGSize)collectionViewContentSize
+{
+  ASDisplayNodeAssertMainThread();
+  return _currentContentAttributes.contentSize;
+}
+
+#pragma mark - Subclass hooks
+
+- (ASCollectionContentAttributes *)calculateLayoutForLayoutContext:(ASDataControllerLayoutContext *)context
```
Adlai-Holler: `calculateLayoutWithContext:`?

Adlai-Holler (changes requested):
This is shaping up well! I think that letting them talk directly to the UICollectionView might be a mistake, at least early on. Maybe instead, they provide a `layoutDelegate` which implements `calculateLayoutWithContext:`? We can retain `layoutDelegate` for their convenience. What do you think?

nguyenhuy:
> Maybe instead, they provide a layoutDelegate which implements calculateLayoutWithContext:?

@Adlai-Holler You meant an `ASCollectionLayoutDelegate` protocol that has a required `-calculateLayoutWithContext:` method?

I think it's good idea as it forbids people from accessing the collection view directly. In addition, I'd like to leave an option for people to subclass `ASCollectionLayout` directly. What do you think?

Source/Details/ASCollectionLayoutState.h
```diff
@@ -0,0 +1,54 @@
...
+
+@class ASElementMap, ASCollectionElement, ASLayout;
+
+NS_ASSUME_NONNULL_BEGIN
+
+AS_SUBCLASSING_RESTRICTED
+@interface ASCollectionLayoutState : NSObject
+
+/// The element map used to calculate this object
+@property (nonatomic, weak, readonly) ASElementMap *elementMap;
```
Adlai-Holler: Let's make this `strong` so that this object is truly immutable.

Source/Details/ASDataController.mm
```diff
@@ -480,33 +529,57 @@ - (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet
...
-  [self _updateElementsInMap:mutableMap changeSet:changeSet owningNode:owningNode traitCollection:existingTraitCollection];
+  [self _updateElementsInMap:mutableMap changeSet:changeSet owningNode:owningNode traitCollection:existingTraitCollection shouldFetchSizeRanges:(! canDelegateLayout)];
   
   // Step 2: Clone the new data
   ASElementMap *newMap = [mutableMap copy];
-
   _pendingMap = newMap;
   
+  // Step 3: Ask layout delegate for contexts
+  ASCollectionLayoutContext *layoutContext;
```
Adlai-Holler: Initialize this to `nil` for good measure.

Source/Private/ASCollectionLayout.mm
```diff
@@ -0,0 +1,151 @@
...
+}
+
+#pragma mark - UICollectionViewLayout overrides
+
+- (void)prepareLayout
+{
+  ASDisplayNodeAssertMainThread();
+  ASCollectionLayoutContext *context =  [self layoutContextWithElementMap:_collectionNode.visibleMap];
+  
+  ASCollectionLayoutState *state;
```
Adlai-Holler: Initialize to `nil`

Source/Details/ASCollectionLayoutContext.h
```diff
@@ -0,0 +1,27 @@
...
+@class ASElementMap;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface ASCollectionLayoutContext : NSObject
+
+@property (nonatomic, assign, readonly) CGSize viewportSize;
+@property (nonatomic, weak, readonly) ASElementMap *elementMap;
+
+- (instancetype)initWithViewportSize:(CGSize)viewportSize elementMap:(ASElementMap *)map NS_DESIGNATED_INITIALIZER;
```
Adlai-Holler: What do you think about an alternative to this API:
- `LayoutContext` is subclassing-restricted and has no public initializer.
- Add an `readonly nullable id userInfo` object to it.
- Reach out to the user to get the `userInfo` and attach it to the LayoutContext that we provide them.

That was the class hierarchy is shallower and the separation of control is clearer.

Source/Details/ASCollectionLayoutDelegate.h
```diff
@@ -0,0 +1,46 @@
...
+
+@optional
+
+/**
+ * @abstract Returns a layout context for given element map that will be used to prepare a new layout.
+ * The context should return the element map and any additional information needed for the coming layout pass.
+ *
+ * @discussion This method will be called on main thread.
+ */
+- (ASCollectionLayoutContext *)layoutContextWithElementMap:(ASElementMap *)map;
```
Adlai-Holler: If you agree with https://github.com/facebook/AsyncDisplayKit/pull/3130/files#r110550765 then we'd modify this method to e.g.:

`- (nullable id)additionalInfoForLayoutWithElements:(ASElementMap *)elements`

And IMO we should make it required, so that all users know it's available. They can easily return `nil` if they have nothing else to return.

Source/Details/ASDataController.h
```diff
@@ -73,12 +70,20 @@ extern NSString * const ASCollectionInvalidUpdateException;
...
+- (ASSizeRange)dataController:(ASDataController *)dataController constrainedSizeForNodeAtIndexPath:(NSIndexPath *)indexPath;
+
 - (NSArray<NSString *> *)dataController:(ASDataController *)dataController supplementaryNodeKindsInSections:(NSIndexSet *)sections;
 
 - (NSUInteger)dataController:(ASDataController *)dataController supplementaryNodesOfKind:(NSString *)kind inSection:(NSUInteger)section;
 
 - (ASCellNodeBlock)dataController:(ASDataController *)dataController supplementaryNodeBlockOfKind:(NSString *)kind atIndexPath:(NSIndexPath *)indexPath;
 
+/**
+ The constrained size range for layout. Called only if a collection layout calculator is not provided.
```
Adlai-Holler: `calculator` -> `delegate`

Adlai-Holler (changes requested):
@nguyenhuy Left some more nitpicks and ideas, but it's almost landing time. Would you be willing to modify one of the examples to use this layout system?

Source/ASCollectionNode+Beta.h
```diff
@@ -23,8 +24,19 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (strong, nonatomic, nullable) Class collectionViewClass;
 
+/**
+ * The map that is currently displayed. The "UIKit index space.". Must be accessed on main thread.
+ */
+@property (strong, nonatomic, readonly) ASElementMap *visibleMap;
+
+- (instancetype)initWithFrame:(CGRect)frame collectionLayoutDelegate:(id<ASCollectionLayoutDelegate>)layoutDelegate layoutFacilitator:(nullable id<ASCollectionViewLayoutFacilitatorProtocol>)layoutFacilitator;
```
nguyenhuy: Removed the frame and changed to `layoutDelegate` .

Source/ASCollectionNode+Beta.h
```diff
@@ -23,8 +24,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+
 - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionViewLayout *)layout layoutFacilitator:(nullable id<ASCollectionViewLayoutFacilitatorProtocol>)layoutFacilitator;
 
+- (void)setCollectionLayoutDelegate:(nonnull id<ASCollectionLayoutDelegate>)collectionLayoutDelegate;
+
+- (nullable id<ASCollectionLayoutDelegate>)collectionLayoutDelegate;
```
nguyenhuy: Good idea! Done.

Source/Details/ASCollectionLayoutContext.m
```diff
@@ -0,0 +1,50 @@
...
+  }
+  if (! [other isKindOfClass:[ASCollectionLayoutContext class]]) {
+    return NO;
+  }
+  return [self isEqualToContext:other];
+}
+
+- (NSUInteger)hash
+{
+  return [_elementMap hash] ^ (((NSUInteger)(_viewportSize.width * 255) << 8) + (NSUInteger)(_viewportSize.height * 255));
```
nguyenhuy: Oh, I ended up using `ASIntegerArrayHash` and implemented a new `ASHashFromCGSize` function. We can make use of the new function in a couple of other places. I'll do that in another PR.

Source/Details/ASCollectionLayoutState.h
nguyenhuy: My main reason for using `weak` here and in `ASCollectionLayoutContext` is to make sure only `ASDataController` holds strong reference to the map and thus it is released as soon as `ASDataController` has a new data or it's deallocated. That being said, having a strong reference here ensures that the state is valid during its lifetime and I updated the code in favor of this. The take away is that now `visibleMap` is retained not only by `ASDataController` but also by other objects, so we need to be careful to not leak it.

Source/Details/ASDataController.mm
nguyenhuy: Done

Source/Private/ASCollectionLayout.mm
nguyenhuy: Done

Source/Details/ASCollectionLayoutContext.h
nguyenhuy: Love this idea. Implemented!

Source/Details/ASCollectionLayoutDelegate.h
nguyenhuy: ✅

Source/Details/ASDataController.h
nguyenhuy: Fixed.

Adlai-Holler (approved):
Man I'm glad to hit this approve button. You really stuck with this one, @nguyenhuy. Merge it when you feel like it.

nguyenhuy:
Wooho! Thank you, @Adlai-Holler!

PR #3140, opened 2017-03-05
Initial Work for Range Controller to Support Supplementary Elements

Adlai-Holler:
This doesn't actually change any behavior, but it lays the groundwork for range controllers to deal with supplementary elements.

After this diff, we provide elements of other kinds to the range controller, but before it starts its core work it filter-maps supplementary elements into item index paths, and continues behaving as before.

We should review this and once it's landed we can change the core range controller algorithm to deal with supplementary elements and items alike.

### DIFF
--- a/Source/ASTableView.mm
+++ b/Source/ASTableView.mm
@@ -649,18 +649,8 @@ static NSString * const kCellReuseIdentifier = @"_ASTableViewCell";
 
 - (NSArray<ASCellNode *> *)visibleNodes
 {
-  NSArray *indexPaths = [self visibleNodeIndexPathsForRangeController:_rangeController];
-  
-  NSMutableArray<ASCellNode *> *visibleNodes = [NSMutableArray array];
-  for (NSIndexPath *indexPath in indexPaths) {
-    ASCellNode *node = [self nodeForRowAtIndexPath:indexPath];
-    if (node) {
-      // It is possible for UITableView to return indexPaths before the node is completed.
-      [visibleNodes addObject:node];
-    }
-  }
-  
-  return visibleNodes;
+  NSArray<ASCollectionElement *> *elements = [self visibleElementsForRangeController:_rangeController];
+  return ASArrayByFlatMapping(elements, ASCollectionElement *e, e.node);
 }
 
 - (void)beginUpdates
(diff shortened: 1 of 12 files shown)


Source/ASTableView.mm
```diff
@@ -649,18 +649,8 @@ - (nullable NSIndexPath *)indexPathForNode:(ASCellNode *)cellNode waitingIfNeede
...
 - (NSArray<ASCellNode *> *)visibleNodes
 {
-  NSArray *indexPaths = [self visibleNodeIndexPathsForRangeController:_rangeController];
-  
-  NSMutableArray<ASCellNode *> *visibleNodes = [NSMutableArray array];
-  for (NSIndexPath *indexPath in indexPaths) {
-    ASCellNode *node = [self nodeForRowAtIndexPath:indexPath];
-    if (node) {
-      // It is possible for UITableView to return indexPaths before the node is completed.
-      [visibleNodes addObject:node];
```
Adlai-Holler: The comment here is incorrect. UITableView cannot return index paths before the node is completed.

Source/Private/ASElementMap.h
```diff
@@ -55,6 +60,11 @@ AS_SUBCLASSING_RESTRICTED
 - (nullable NSIndexPath *)indexPathForElement:(ASCollectionElement *)element;
 
 /**
+ * Returns the index path for the given element, if it represents an item. O(1)
+ */
+- (nullable NSIndexPath *)indexPathForElementIfItem:(ASCollectionElement *)element;
```
nguyenhuy: I'd prefer `-indexPathForElementIfRepresentsItem` or the like. "IfItem" is kind of confusing. I couldn't understand its purpose until I read its documentation.

nguyenhuy: I'd prefer -indexPathForElementIfRepresentsItem or the like. "IfItem" is kind of confusing. I couldn't understand its purpose until I read its documentation.

Adlai-Holler:
Thanks @nguyenhuy! Since the most recent commit was only a rename, I'll go ahead and merge.

Source/Private/ASElementMap.h
Adlai-Holler: Good call. I went with `indexPathForElementIfCell:` which I think is a lot clearer. I try to avoid using the word `Cell` in general but there is precedent in `UICollectionElementCategoryCell`.

PR #3145, opened 2017-03-07
[ASControlNode] Align touch and tracking system in ASControlNode to UIControl

### LINKED ISSUE #3139: Button node stay in highlighted state after pushing view in navigation controller
ASDK v2.2 (also tested on 2.1, 2.0.1, 2.0.2)

Case: Push view in navigation controller on ASButtonNode !TouchDown! event.

Problem: After returning back to view, from which push has been performed, the button is in highlighted state.

Sample project: https://github.com/Kuluum/ASDKButtonHighlightIssue

maicki:
Currently the touch and tracking system of ASControlNode does not align with the system of UIControl. ASControlNode does not even consider `beginTrackingWithTouch:withEvent:` or `continueTrackingWithTouch:withEvent:`.

This commit will change the ASControlNode logic around touch event handling and tracking to align with UIControl after diving a bit into the assembly and the way UIControl is working.

This PR also uncomments the assertion that if adding a target action to an `ASControlNode` that is layer backed it will throw an assertion.

Resolves #3139

### DIFF
--- a/Source/ASControlNode.mm
+++ b/Source/ASControlNode.mm
@@ -110,30 +110,42 @@ CGRect _ASControlNodeGetExpandedBounds(ASControlNode *controlNode);
   self.isAccessibilityElement = userInteractionEnabled;
 }
 
+- (void)__exitHierarchy
+{
+  [super __exitHierarchy];
+  
+  // If a control node is exit the hierarchy and is tracking we have to cancel it
+  if (self.tracking) {
+    [self _cancelTrackingForControlEvents:ASControlNodeEventTouchCancel withEvent:nil];
+  }
+}
 
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wobjc-missing-super-calls"
 
 #pragma mark - ASDisplayNode Overrides
+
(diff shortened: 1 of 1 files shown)


Source/ASControlNode.mm
```diff
@@ -230,6 +238,17 @@ - (void)touchesEnded:(NSSet *)touches withEvent:(UIEvent *)event
                           withEvent:event];
 }
 
+- (void)_cancelTrackingForControlEvents:(ASControlNodeEvent)controlEventMask withEvent:(UIEvent *)event
```
Adlai-Holler: This should probably be:

```objc
- (void)_cancelTrackingWithEvent:(UIEvent *)event
```
and then use the `ASControlNodeEventTouchCancel` literal directly here.

maicki: I had it first this way, but if you look where we call it, we actually call it also in `-touchesBegan:withEvent:` to have kind of the same event mask in this case as before.

maicki:
@Adlai-Holler Addressed the comments. Thanks!

PR #3157, opened 2017-03-09
ASMultidimensionalArrayUtils -> ASTwoDimensionalArrayUtils

Adlai-Holler:
I got kind of down the rabbit hole with this one but I think the benefits are nice!


Source/Private/ASTwoDimensionalArrayUtils.h
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Deep mutable copy of an array that contains arrays, which contain objects.  It will go one level deep into the array to copy.
+ * This method is substantially faster than the generalized version, e.g. about 10x faster, so use it whenever it fits the need.
+ */
+extern NSMutableArray<NSMutableArray *> *ASTwoDimensionalArrayDeepMutableCopy(NSArray<NSArray *> *array) AS_WARN_UNUSED_RESULT;
+
+/**
+ * Delete the elements of the mutable two-dimensional array at given index paths – sorted in descending order!
+ */
+extern void ASDeleteElementsInTwoDimensionalArrayAtIndexPaths(NSMutableArray *mutableArray, NSArray<NSIndexPath *> *indexPaths);
```
Adlai-Holler: Surprisingly, adding generics to this `mutableArray` argument causes type-mismatch warnings when this function is called with a mutable array that has a type like `NSMutableArray<NSMutableArray<ASCollectionElement *> *> *` 🤷‍♂️

Source/Private/ASTwoDimensionalArrayUtils.m
```diff
@@ -0,0 +1,82 @@
...
+  ASDisplayNodeCAssert([sortedIndexPaths isEqualToArray:indexPaths], @"Expected array of index paths to be sorted in descending order.");
+#endif
+
+  /**
+   * It is tempting to do something clever here and collect indexes into ranges or NSIndexSets
+   * but deep down, __NSArrayM only implements removeObjectAtIndex: and so doing all that extra
+   * work ends up running the same code.
+   */
+  for (NSIndexPath *indexPath in indexPaths) {
+    [mutableArray[indexPath.section] removeObjectAtIndex:indexPath.item];
```
maicki: Are we sure that always  `mutableArray.count < indexPath.section`?

PR #3161, opened 2017-03-10
[ASControlNode] ASControlNode should not receive any touches if it is disabled

### LINKED ISSUE #2234: ASButton in ASControl eat all touch event for it's area
I have a ASButton as a subview of ASControl.

I have a TouchUpInside event for ASControl
And ASButton has no event , or hidden, or userInteraction = NO,

but I got no event-callback for ASControll when touch in area of ASButton.

maicki:
Resolves #2234

### DIFF
--- a/Source/ASControlNode.mm
+++ b/Source/ASControlNode.mm
@@ -10,6 +10,7 @@
 
 #import <AsyncDisplayKit/ASControlNode.h>
 #import <AsyncDisplayKit/ASControlNode+Subclasses.h>
+#import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
 #import <AsyncDisplayKit/ASImageNode.h>
 #import <AsyncDisplayKit/AsyncDisplayKit+Debug.h>
 #import <AsyncDisplayKit/ASInternalHelpers.h>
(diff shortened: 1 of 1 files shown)


Source/ASControlNode.mm
```diff
@@ -9,6 +9,7 @@
 //
 
 #import <AsyncDisplayKit/ASControlNode.h>
+#import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
```
nguyenhuy: Please move this 1 line down.

Source/ASControlNode.mm
```diff
@@ -316,7 +329,6 @@ - (void)addTarget:(id)target action:(SEL)action forControlEvents:(ASControlNodeE
       }
     });
 
-  self.userInteractionEnabled = YES;
```
nguyenhuy: This line has a very controversial history. Let's leave it as-is and deal with it in another PR if need be.

https://github.com/facebook/AsyncDisplayKit/pull/1189

maicki:
@nguyenhuy Can you please take another look. Thanks!

PR #3166, opened 2017-03-11
flexShrinkAdjustment lambda must always return CGFloat

### LINKED ISSUE #3165: ASStackUnpositionedLayout flexShrinkAdjustment returning float rather than double
Just looking to archive a build but came across this error...
https://cl.ly/2l0H0t0C3S08

Casting the return to a double resolves the issue.

nguyenhuy:
This should fix #3165.


ay8s:
👍

PR #3167, opened 2017-03-11
When update title of a button node, need to get title node via setter

### LINKED ISSUE #3164: Button nodes not displaying text, possibly now truncated with recent change
With the latest commit 0540c863ca8d26e0aeb2c626ee1cb7391d6fc230 we're seeing that a couple of our ASButtonNode's are no longer showing text which previously displayed fine. Not sure if this is a truncation issue as spotted it mentioned in the notes or another issue introduced with that commit.

Previously: https://cl.ly/3d3O473w081G
Now: https://cl.ly/0f2s3F2T1M3A

nguyenhuy:
Since the title node is lazily created, we need to get it via its setter or risk having no title at all! 

Fixes #3164.

### DIFF
--- a/Source/ASButtonNode.mm
+++ b/Source/ASButtonNode.mm
@@ -197,13 +197,11 @@
     newTitle = _normalAttributedTitle;
   }
 
-  ASTextNode *titleNode = _titleNode;
-  
-  if ((titleNode != nil || newTitle.length > 0) && [titleNode.attributedText isEqualToAttributedString:newTitle] == NO) {
-    titleNode.attributedText = newTitle;
+  if ((_titleNode != nil || newTitle.length > 0) && [self.titleNode.attributedText isEqualToAttributedString:newTitle] == NO) {
+    _titleNode.attributedText = newTitle;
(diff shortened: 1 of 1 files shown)


Source/ASButtonNode.mm
```diff
@@ -197,13 +197,11 @@ - (void)updateTitle
     newTitle = _normalAttributedTitle;
   }
 
-  ASTextNode *titleNode = _titleNode;
-  
-  if ((titleNode != nil || newTitle.length > 0) && [titleNode.attributedText isEqualToAttributedString:newTitle] == NO) {
-    titleNode.attributedText = newTitle;
+  if ((_titleNode != nil || newTitle.length > 0) && [self.titleNode.attributedText isEqualToAttributedString:newTitle] == NO) {
```
maicki: @nguyenhuy Can you add a comment above this line *why* we call `self.titleNode` explicitly. Thanks!

nguyenhuy: @maicki Done!

PR #3169, opened 2017-03-12
Implement native move based on PR #2868

### LINKED ISSUE #698: [ASDataController] "Move" Operations Animate Incorrectly (Delete + Insert)

harryworld:
Ref to #2868, resolves #698.

This is first attempt to use NSCollectionView native move animation.

Thanks @nguyenhuy for big refactoring, this is for sure a lot more clear than before. Still appreciate for opinions on how to improve.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -1831,6 +1831,17 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
       
       __block NSUInteger numberOfUpdates = 0;
       [self _superPerformBatchUpdates:^{
+        
+        // Use native moveItem instead of Delete-Insert
+        NSMutableArray *fromIndexPaths = [NSMutableArray array];
+        NSMutableArray *toIndexPaths = [NSMutableArray array];
+        for (_ASHierarchyMoveItemChange *change in [changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
+          [super moveItemAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath];
+          [fromIndexPaths addObject:change.fromIndexPath];
+          [toIndexPaths addObject:change.toIndexPath];
+          numberOfUpdates++;
+        }
+        
         for (_ASHierarchyItemChange *change in [changeSet itemChangesOfType:_ASHierarchyChangeTypeReload]) {
           [super reloadItemsAtIndexPaths:change.indexPaths];
           numberOfUpdates++;
@@ -1842,7 +1853,14 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
         }
         
         for (_ASHierarchyItemChange *change in [changeSet itemChangesOfType:_ASHierarchyChangeTypeOriginalDelete]) {
-          [super deleteItemsAtIndexPaths:change.indexPaths];
+          NSMutableArray *actualIndexPaths = [NSMutableArray array];
+          for (NSIndexPath *indexPath in change.indexPaths) {
+            if (![fromIndexPaths containsObject:indexPath]) {
+              [actualIndexPaths addObject:indexPath];
+            }
+          }
+          
+          [super deleteItemsAtIndexPaths:actualIndexPaths];
           numberOfUpdates++;
         }
         
@@ -1857,7 +1875,14 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
         }
         
         for (_ASHierarchyItemChange *change in [changeSet itemChangesOfType:_ASHierarchyChangeTypeOriginalInsert]) {
-          [super insertItemsAtIndexPaths:change.indexPaths];
+          NSMutableArray *actualIndexPaths = [NSMutableArray array];
+          for (NSIndexPath *indexPath in change.indexPaths) {
+            if (![toIndexPaths containsObject:indexPath]) {
+              [actualIndexPaths addObject:indexPath];
+            }
+          }
+          
+          [super insertItemsAtIndexPaths:actualIndexPaths];
           numberOfUpdates++;
         }
       } completion:^(BOOL finished){
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -419,7 +419,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   _flags.isDeallocating = YES;
 
   // Synchronous nodes may not be able to call the hierarchy notifications, so only enforce for regular nodes.
-  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating. Node: %@", self);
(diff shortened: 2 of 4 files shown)


Source/ASCollectionView.mm
```diff
@@ -1831,6 +1831,17 @@ - (void)rangeController:(ASRangeController *)rangeController didUpdateWithChange
...
+        // Use native moveItem instead of Delete-Insert
+        NSMutableArray *fromIndexPaths = [NSMutableArray array];
+        NSMutableArray *toIndexPaths = [NSMutableArray array];
+        for (_ASHierarchyMoveItemChange *change in [changeSet itemChangesOfType:_ASHierarchyChangeTypeMove]) {
+          [super moveItemAtIndexPath:change.fromIndexPath toIndexPath:change.toIndexPath];
+          [fromIndexPaths addObject:change.fromIndexPath];
+          [toIndexPaths addObject:change.toIndexPath];
+          numberOfUpdates++;
+        }
+
```
nguyenhuy: To make sure the element maps of `ASDataController` are correctly synced, the steps described by @Adlai-Holler [here](https://github.com/facebook/AsyncDisplayKit/pull/2868#discussion_r94865245) should be implemented in `-ASDataController:_updateElementsInMap:changeSet:`. 

Then in `ASCollectionView` and `ASTableView`, we can simply forward each move update to `super`.

Source/Private/_ASHierarchyChangeSet.mm
```diff
@@ -300,6 +310,9 @@ - (void)moveItemAtIndexPath:(NSIndexPath *)indexPath toIndexPath:(NSIndexPath *)
   /**
    * TODO: Proper move implementation.
    */
+  [self _ensureNotCompleted];
+  _ASHierarchyMoveItemChange *change = [[_ASHierarchyMoveItemChange alloc] initWithChangeType:_ASHierarchyChangeTypeMove fromIndexPath:indexPath toIndexPath:newIndexPath animationOptions:options];
+  [_moveItemChanges addObject:change];
   [self deleteItems:@[ indexPath ] animationOptions:options];
   [self insertItems:@[ newIndexPath ] animationOptions:options];
```
harryworld: The primary reason of keeping these (for now) is that `move` nodes shall be split and processed in `delete` and `insert` during the batch process. It just skips calling the UICollectionView delete/insert, but move instead.

We can consider the implementation like reload. Yet, I'm not sure where is the appropriate place to process the split.

Source/ASDisplayNode.mm
```diff
@@ -419,7 +419,7 @@ - (void)dealloc
   _flags.isDeallocating = YES;
 
   // Synchronous nodes may not be able to call the hierarchy notifications, so only enforce for regular nodes.
-  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating. Node: %@", self);
+//  ASDisplayNodeAssert(_flags.synchronous || !ASInterfaceStateIncludesVisible(_interfaceState), @"Node should always be marked invisible before deallocating. Node: %@", self);
```
harryworld: I forgot to mention. I was getting assert error on this, after moving the nodes a few times. Appreciate if you can point out what I did wrong to hit this.

Source/ASCollectionView.mm
harryworld: This part may be irrelevant, as @Adlai-Holler was mentioning the change to dataSource, and this part is regarding to the call to UICollectionView.

Consider the comment I made above, after changing the move mechanism to something similar to current reload, we may have a part to do what @Adlai-Holler said.

Appreciate for further discussion.

nguyenhuy:
@harryworld Thank you for your reply.  It is now clear that we are not on the same page.  I'm sorry about the confusion and let me try to explain better.

First, let's limit the scope of this discussion to item moves.  It's totally fine to work on section moves later on, even in a follow up PR.

You're right that moves should be treated similar to reloads in the sense that `_ASHierarchyChangeSet` records them as-is and let other components (`ASDataController`, `ASTableView` and `ASCollectionView`) decide how to process them.  With that, I believe a valid implementation consists of 4 independent steps:
 1. Record move updates into a change set.  They should be recored as-is and not be split into inserts and deletes.  This is because a change set should be regarded as the source of truth, which sadly has not always been the case.  How an update will be processed later on is not the responsibility of this step and thus it should not care about it.
 2. Validate the updates.  In this set we care about whether the indices in all updates are valid.  As a result, for each move, we have to validate it such that:
  - Its "from" index is valid in the original index space.  It means we have to merge the index into indices of original delete updates and validate.
  - Its "to" index is valid in the future index space.  That means we put it into the list of original insert updates and validate.
 3. `ASDataController` to process moves.  The goal of this step is to make sure the internal properties of ASDataController is correctly updated.
   - It's important that for a move, we don't throw away the existing (often visible) node and load a new one which I believe is what currently happening in this PR and triggers the assertion [here](https://github.com/facebook/AsyncDisplayKit/pull/3169/files#diff-fb6dadff65468c53595c164828199c83L422).  Even if we manage to silent the assertion, we'll need to make sure the new node has the same states as the previous one.  And then I think it's inevitable that a flash will occur while the new node is being rendered.  The flash will be especially bad for an interactive move. 
   - I believe the correct implementation of this step needs to be implemented in `-[ASDataController _updateElementsInMap:changeSet:]`.  We need 2 smaller steps:
     3.1. Grab the node at the "from" index path of each move and store it into a temporary storage but use the "to" index path as key to indicate the future position of this node.  Then we put the "from" index path into a list of other deleted index paths that are originated from delete updates, sort them in descending order and remove items at these index paths from the provided element map.
     3.2. Collect all index paths from both move updates ("to" index paths) and insert updates, sort them in ascending order and insert items one by one. For each index path, either grab an existing node from the above temporary storage if available, or  populate a new one (in case of original inserts).
   - Regarding supplementary nodes, we can delete them as part of step 3.1 and repopulate as part of step 3.2.  Aggressive reloading is not ideal, but it has been the case for other type of updates and improving it is not the scope of this PR.
 4. `ASCollectionView` and `ASTableView` to forward all move updates to their super as-is.  This step is pretty straightforward.  The general rule is that we should always forward original updates.


After having a concrete implementation for native item moves, we'll be in a pretty good shape to implement section moves.  We'll also need to revisit the need for separate delete and original delete changes in `_ASHierarchyChangeSet` (as well as insert and original insert changes).

Please let me know if there is anything unclear.  As our working schedules don't overlap much, feel free to go ahead with implementing any steps and we'll discuss more on the actual code. Thank you!

PR #3172, opened 2017-03-12
[ASDisplayNode] Improve locking in ASDisplayNode

maicki:
- Add and Remove locking where it's necessary and missing or it could cause to deadlocks
- Annotate methods that needs to be called with a lock held with `_locked_` prefix
- Cleanup code around our locking guidelines in general (scoped locking and calling `lock()` and `unlock()` directly)
- Most of the locking code around layout is not touched yet that will be in a follow up PR on top of this one, but I wanted to have this two separated to have the changes as small as possible.

### DIFF
--- a/Source/ASCellNode.mm
+++ b/Source/ASCellNode.mm
@@ -118,12 +118,12 @@
   _viewControllerNode.frame = self.bounds;
 }
 
-- (void)_locked_rootNodeDidInvalidateSize
+- (void)_rootNodeDidInvalidateSize
 {
   if (_interactionDelegate != nil) {
     [_interactionDelegate nodeDidInvalidateSize:self];
   } else {
-    [super _locked_rootNodeDidInvalidateSize];
+    [super _rootNodeDidInvalidateSize];
   }
 }
 
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -729,31 +729,30 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (UIView *)view
 {
-  // TODO: locking: Should this be locked?
+  ASDisplayNodeAssertMainThread();
   
-  ASDisplayNodeAssert(!_flags.layerBacked, @"Call to -view undefined on layer-backed nodes");
-  if (_flags.layerBacked) {
+  ASDisplayNodeAssert(!self.layerBacked, @"Call to -view undefined on layer-backed nodes");
+  if (self.layerBacked) {
     return nil;
   }
 
-  if (!_view) {
-    ASDisplayNodeAssertMainThread();
+  // We don't need locking in this case as accessing the view should always happening on the main thread
+  if (_view == nil) {
     [self _loadViewOrLayerIsLayerBacked:NO];
   }
+
   return _view;
 }
 
 - (CALayer *)layer
 {
-  // TODO: locking: Should this be locked?
+  ASDisplayNodeAssertMainThread();
 
-  if (!_layer) {
-    ASDisplayNodeAssertMainThread();
-
-    if (!_flags.layerBacked) {
+  // We don't need locking in this case as accessing the view should always happening on the main thread
+  if (_layer == nil) {
+    if (!self.layerBacked) {
       return self.view.layer;
     }
-
     [self _loadViewOrLayerIsLayerBacked:YES];
   }
 
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -729,15 +729,13 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (UIView *)view
 {
-  ASDisplayNodeAssertMainThread();
-  
-  ASDisplayNodeAssert(!self.layerBacked, @"Call to -view undefined on layer-backed nodes");
-  if (self.layerBacked) {
+  ASDisplayNodeAssert(!_flags.layerBacked, @"Call to -view undefined on layer-backed nodes");
+  if (_flags.layerBacked) {
     return nil;
   }
 
-  // We don't need locking in this case as accessing the view should always happening on the main thread
   if (_view == nil) {
+    ASDisplayNodeAssertMainThread();
     [self _loadViewOrLayerIsLayerBacked:NO];
   }
 
@@ -746,13 +744,13 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (CALayer *)layer
 {
-  ASDisplayNodeAssertMainThread();
-
-  // We don't need locking in this case as accessing the view should always happening on the main thread
   if (_layer == nil) {
-    if (!self.layerBacked) {
+    ASDisplayNodeAssertMainThread();
+    
+    if (!_flags.layerBacked) {
       return self.view.layer;
     }
+    
     [self _loadViewOrLayerIsLayerBacked:YES];
   }
 
@@ -1994,8 +1992,7 @@ NSString * const ASRenderingEngineDidDisplayNodesScheduledBeforeTimestamp = @"AS
 
 - (void)recursivelyDisplayImmediately
 {
-  NSArray *subnodes = self.subnodes;
-  for (ASDisplayNode *child in subnodes) {
+  for (ASDisplayNode *child in self.subnodes) {
     [child recursivelyDisplayImmediately];
   }
   [self displayImmediately];
@@ -2636,7 +2633,7 @@ ASDISPLAYNODE_INLINE BOOL nodeIsInRasterizedTree(ASDisplayNode *node) {
       // which has some nice carry-over machinery to handle cases where we are removed from a hierarchy
       // and then added into it again shortly after.
       __instanceLock__.lock();
-      BOOL isInHierarchy = _flags .isInHierarchy;
+      BOOL isInHierarchy = _flags.isInHierarchy;
       __instanceLock__.unlock();
       
       if (parentWasOrIsRasterized && isInHierarchy) {
@@ -2796,8 +2793,7 @@ ASDISPLAYNODE_INLINE BOOL nodeIsInRasterizedTree(ASDisplayNode *node) {
 {
   ASDisplayNodeAssertMainThread();
   
-  NSArray *subnodes = self.subnodes;
-  for (ASDisplayNode *node in subnodes) {
+  for (ASDisplayNode *node in self.subnodes) {
     [self _addSubnodeSubviewOrSublayer:node];
   }
 }
@@ -2807,7 +2803,7 @@ ASDISPLAYNODE_INLINE BOOL nodeIsInRasterizedTree(ASDisplayNode *node) {
   ASDisplayNodeAssertMainThread();
   
   // Due to a bug in Apple's framework we have to use the layer index to insert a subview
-  // so just use th ecount of the sublayers to add the subnode
+  // so just use the count of the sublayers to add the subnode
   NSInteger idx = _layer.sublayers.count; // No locking is needed as it's main thread only
   [self _insertSubnodeSubviewOrSublayer:subnode atIndex:idx];
 }
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -668,7 +668,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     }
     _layer.asyncdisplaykit_node = self;
 
-    self.asyncLayer.asyncDelegate = self;
+    self._locked_asyncLayer.asyncDelegate = self;
   }
 
   {
@@ -735,6 +735,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   if (_flags.layerBacked) {
     return nil;
   }
+
   if (!_view) {
     ASDisplayNodeAssertMainThread();
     [self _loadViewOrLayerIsLayerBacked:NO];
@@ -752,8 +753,10 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     if (!_flags.layerBacked) {
       return self.view.layer;
     }
+
     [self _loadViewOrLayerIsLayerBacked:YES];
   }
+
   return _layer;
 }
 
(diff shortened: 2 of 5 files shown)


Source/ASDisplayNode.mm
```diff
@@ -729,6 +729,8 @@ - (BOOL)_isNodeLoaded
 
 - (UIView *)view
```
maicki: I think we have to add locking to this method ... any reason why we didn't have in the past?

Source/ASDisplayNode.mm
```diff
@@ -742,6 +744,8 @@ - (UIView *)view
 
 - (CALayer *)layer
```
maicki: I think we have to add locking to this method ... any reason why we didn't have in the past?

Source/ASDisplayNode.mm
```diff
@@ -3312,36 +3387,6 @@ - (void)didExitHierarchy
   }
 }
 
-- (void)_recursiveWillEnterHierarchy
```
maicki: We don't use this methods anywhere, so I just removed it.

Source/Private/ASDisplayNodeInternal.h
```diff
@@ -200,9 +200,6 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
 /// Bitmask to check which methods an object overrides.
 @property (nonatomic, assign, readonly) ASDisplayNodeMethodOverrides methodOverrides;
 
-// Swizzle to extend the builtin functionality with custom logic
-- (BOOL)__shouldLoadViewOrLayer;
```
maicki: The method is not used outside so we should not expose it.

Source/Private/ASDisplayNodeInternal.h
```diff
@@ -218,11 +215,6 @@ FOUNDATION_EXPORT NSString * const ASRenderingEngineDidDisplayNodesScheduledBefo
  */
 - (void)__layout;
 
-/*
- * Internal method to set the supernode
- */
-- (void)__setSupernode:(nullable ASDisplayNode *)supernode;
```
maicki: The method is not used outside so we should not expose it.

Source/ASDisplayNode.mm
```diff
@@ -2132,21 +2169,24 @@ - (void)recursivelyEnsureDisplaySynchronously:(BOOL)synchronously
...
 {
-  ASDN::MutexLocker l(__instanceLock__);
-  if (contentsScale != self.contentsScaleForDisplay) {
-    self.contentsScaleForDisplay = contentsScale;
-    [self setNeedsDisplay];
+  if (contentsScale == self.contentsScaleForDisplay) {
+    return;
   }
+  
+  self.contentsScaleForDisplay = contentsScale;
```
nguyenhuy: It's tempting to grab the lock here and drop down to `_contentsScaleForDisplay`.

Source/ASDisplayNode.mm
```diff
@@ -2005,8 +2033,10 @@ + (void)scheduleNodeForRecursiveDisplay:(ASDisplayNode *)node
 }
 
 /// Helper method to summarize whether or not the node run through the display process
-- (BOOL)__implementsDisplay
+- (BOOL)_implementsDisplay
 {
+  ASDN::MutexLocker l(__instanceLock__);
```
nguyenhuy: Idea: Can we make these flags (individual flag, not the whole _flags struct) atomic and avoid locking entirely?

Source/ASDisplayNode.mm
```diff
@@ -2685,10 +2743,15 @@ - (void)_insertSubnodeSubviewOrSublayer:(ASDisplayNode *)subnode atIndex:(NSInte
   // where the view and layer can change. We can avoid locking.
 
   // If we can use view API, do. Due to an apple bug, -insertSubview:atIndex: actually wants a LAYER index, which we pass in
+  __instanceLock__.lock();
+  UIView *view = _view;
+  CALayer *layer = _layer;
+  __instanceLock__.unlock();
```
nguyenhuy: As discussed above, since this method is on main (asserted few lines above), we can relax the locking requirement here.

Source/ASDisplayNode.mm
```diff
@@ -2725,16 +2788,22 @@ - (void)_addSubnode:(ASDisplayNode *)subnode
 
 - (void)_addSubnodeViewsAndLayers
 {
-  for (ASDisplayNode *node in [_subnodes copy]) {
+  __instanceLock__.lock();
+  NSArray *subnodes = [_subnodes copy];
+  __instanceLock__.unlock();
```
nguyenhuy: `self.subnodes` does the same thing.

nguyenhuy: While you're on this, could you please take a look at `recursivelyDisplayImmediately` as well?

Source/ASDisplayNode.mm
```diff
@@ -2725,16 +2788,22 @@ - (void)_addSubnode:(ASDisplayNode *)subnode
...
 - (void)_addSubnodeSubviewOrSublayer:(ASDisplayNode *)subnode
 {
-    // Due to a bug in Apple's framework we have to use the layer index to insert a subview
-    // so just use th ecount of the sublayers to add the subnode
+  // Due to a bug in Apple's framework we have to use the layer index to insert a subview
+  // so just use th ecount of the sublayers to add the subnode
+  __instanceLock__.lock();
   NSInteger idx = _layer.sublayers.count;
+  __instanceLock__.unlock();
+
```
nguyenhuy: Since `_insertSubnodeSubviewOrSublayer` below is main thread only, we should assert main in this method and avoid locking. Same for `_addSubnodeViewsAndLayers`, `_loadViewOrLayerIsLayerBacked` and related methods.

Source/ASDisplayNode.mm
nguyenhuy: 👍

Source/ASDisplayNode.mm
```diff
@@ -3670,12 +3715,17 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
 {
   // This method is only implemented on UIView on iOS 6+.
   ASDisplayNodeAssertMainThread();
+  
+  __instanceLock__.lock();
+  UIView *view = _view;
+  __instanceLock__.unlock();
```
nguyenhuy: Same deal, main thread asserted, no locking needed.

Source/ASDisplayNode.mm
```diff
@@ -3719,15 +3769,19 @@ - (void)_applyPendingStateToViewOrLayer
...
   // If no view/layer properties were set before the view/layer were created, _pendingViewState will be nil and the default values
   // for the view/layer are still valid.
-  ASDN::MutexLocker l(__instanceLock__);
 
   [self applyPendingViewState];
+  
+  __instanceLock__.lock();
+  BOOL displaySuspended = _flags.displaySuspended;
+  BOOL displaysAsynchronously = _flags.displaysAsynchronously;
+  __instanceLock__.unlock();
```
nguyenhuy: I'm convinced that we need atomic flags!

Source/ASDisplayNode.mm
```diff
@@ -3944,11 +3998,13 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state object
...
+  ASDN::MutexLocker l(__instanceLock__);
   return _primitiveTraitCollection;
 }
 
 - (void)setPrimitiveTraitCollection:(ASPrimitiveTraitCollection)traitCollection
 {
+  ASDN::MutexLocker l(__instanceLock__);
   if (ASPrimitiveTraitCollectionIsEqualToASPrimitiveTraitCollection(traitCollection, _primitiveTraitCollection) == NO) {
     _primitiveTraitCollection = traitCollection;
     ASDisplayNodeLogEvent(self, @"asyncTraitCollectionDidChange: %@", NSStringFromASPrimitiveTraitCollection(traitCollection));
```
nguyenhuy: `-asyncTraitCollectionDidChange` called 1 line below is a subclass hook :(

Source/Private/ASDisplayNodeInternal.h
nguyenhuy: 👍

Source/ASDisplayNode.mm
maicki: Yep agree this will reduce to grabbing the lock only once.

Source/ASDisplayNode.mm
maicki: And there is actually a comment about it a line above ...

Source/ASDisplayNode.mm
```diff
@@ -1849,30 +1867,31 @@ - (void)calculatedLayoutDidChange
...
+  }
 
-  ASDN::MutexLocker l(__instanceLock__);
-
-  if (_flags.displaysAsynchronously == displaysAsynchronously)
+  if (_flags.displaysAsynchronously == displaysAsynchronously) {
     return;
+  }
 
   _flags.displaysAsynchronously = displaysAsynchronously;
```
maicki: I introduced `_locked_asyncLayer` and was able to almost remove all `asyncLayer` calls as they was  called with the lock held.

Source/ASCellNode.mm
```diff
@@ -118,17 +118,21 @@ - (void)layoutDidFinish
   _viewControllerNode.frame = self.bounds;
 }
 
-- (void)_locked_rootNodeDidInvalidateSize
+- (void)_rootNodeDidInvalidateSize
```
maicki: Added

Source/ASDisplayNode.mm
```diff
@@ -1324,17 +1328,19 @@ - (void)_locked_rootNodeDidInvalidateSize
...
   // Check if the returned layout has a different size than our current bounds.
   if (CGSizeEqualToSize(boundsSizeForLayout, layout.size) == NO) {
     // If so, inform our container we need an update (e.g Table, Collection, ViewController, etc).
-    [self _locked_displayNodeDidInvalidateSizeNewSize:layout.size];
+    [self displayNodeDidInvalidateSizeNewSize:layout.size];
   }
 }
 
-- (void)_locked_displayNodeDidInvalidateSizeNewSize:(CGSize)size
+- (void)displayNodeDidInvalidateSizeNewSize:(CGSize)size
```
maicki: Done

Source/ASDisplayNode.mm
```diff
@@ -1361,14 +1367,10 @@ - (void)layout
   ASDisplayNodeAssertMainThread();
   ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__);
 
-  __instanceLock__.lock();
-  if (_calculatedDisplayNodeLayout->isDirty()) {
-    __instanceLock__.unlock();
-    return;
+  ASDN::MutexLocker l(__instanceLock__);
+  if (! _calculatedDisplayNodeLayout->isDirty()) {
+    [self _locked_layoutSublayouts];
```
maicki: Yes it was ...

Source/ASDisplayNode.mm
maicki: Yes agree also `layerBacked`. I think we should defer that to another PR as the diff is already big and `_flags` are used all over the place.

Source/ASDisplayNode.mm
maicki: @nguyenhuy I changed the implementations of `view` and `layer`. Can you please take another look over it. Not sure yet though if we should add the main thread assertion on top.

nguyenhuy: Per the documentation [here](https://github.com/facebook/AsyncDisplayKit/blob/master/Source/ASDisplayNode.h#L202), I think these getters can be called on a background thread (after the first time), given that the returned view/layer is not used on that thread. I can't think of any use case in which calling them off main is needed, but I also don't want to break the API contract.

Can we instead opt for acquiring the lock in these methods?

Source/ASDisplayNode.mm
```diff
@@ -1961,25 +1989,32 @@ - (void)displayImmediately
...
+  [self.asyncLayer displayImmediately];
 }
 
 - (void)recursivelyDisplayImmediately
 {
-  ASDN::MutexLocker l(__instanceLock__);
-  
-  for (ASDisplayNode *child in _subnodes) {
+  NSArray *subnodes = self.subnodes;
+  for (ASDisplayNode *child in subnodes) {
```
nguyenhuy: Maybe call self.subnodes directly here?

Source/ASDisplayNode.mm
```diff
@@ -2571,7 +2635,11 @@ - (void)__setSupernode:(ASDisplayNode *)newSupernode
       // Otherwise we will exit the hierarchy when our view/layer does so
       // which has some nice carry-over machinery to handle cases where we are removed from a hierarchy
       // and then added into it again shortly after.
-      if (parentWasOrIsRasterized && _flags.isInHierarchy) {
+      __instanceLock__.lock();
+      BOOL isInHierarchy = _flags .isInHierarchy;
```
nguyenhuy: Nit: No whitespace between `_flags` and `.isInHierarchy`.

Source/ASDisplayNode.mm
```diff
@@ -2725,16 +2794,21 @@ - (void)_addSubnode:(ASDisplayNode *)subnode
 
 - (void)_addSubnodeViewsAndLayers
 {
-  for (ASDisplayNode *node in [_subnodes copy]) {
+  ASDisplayNodeAssertMainThread();
+  
+  NSArray *subnodes = self.subnodes;
+  for (ASDisplayNode *node in subnodes) {
```
nguyenhuy: Same here: call `self.subnodes` directly.

Source/ASDisplayNode.mm
```diff
@@ -2725,16 +2794,21 @@ - (void)_addSubnode:(ASDisplayNode *)subnode
...
 
 - (void)_addSubnodeSubviewOrSublayer:(ASDisplayNode *)subnode
 {
-    // Due to a bug in Apple's framework we have to use the layer index to insert a subview
-    // so just use th ecount of the sublayers to add the subnode
-  NSInteger idx = _layer.sublayers.count;
+  ASDisplayNodeAssertMainThread();
+  
+  // Due to a bug in Apple's framework we have to use the layer index to insert a subview
+  // so just use th ecount of the sublayers to add the subnode
```
nguyenhuy: Nit: "so just use **the count**..."

maicki:
@nguyenhuy I removed the changes to `-view` and `-layer` for now. I'm honestly not sure what to do in there at the moment. I guess it's a hot path and the reason we don't lock in there is due to that. Otherwise it's just wrong to not lock in this case as we could access `_flags.layerBacked` and `_view` / `_layer` on a background thread. We should talk about this and follow up with a new PR around that soon. cc @garrettmoon

nguyenhuy:
@maicki Sounds good to me.

PR #3178, opened 2017-03-14
[ASCollectionView] Improve interop to silence invalid layout warnings, fix supplementary passthrough.

appleguy:
Before this change, passthrough of supplementary elements didn't work properly -- it would get rehosted into the blank ASCellNode.

Feel free to merge this if it looks good. I've tested it locally and it works well for the use case intended.


Source/ASCollectionView.mm
```diff
@@ -916,7 +916,9 @@ - (CGSize)collectionView:(UICollectionView *)collectionView layout:(UICollection
   ASCellNode *cell = [self nodeForItemAtIndexPath:indexPath];
   if (cell.shouldUseUIKitCell) {
     if ([_asyncDelegate respondsToSelector:@selector(collectionView:layout:sizeForItemAtIndexPath:)]) {
-      return [(id)_asyncDelegate collectionView:collectionView layout:collectionViewLayout sizeForItemAtIndexPath:indexPath];
+      CGSize size = [(id)_asyncDelegate collectionView:collectionView layout:collectionViewLayout sizeForItemAtIndexPath:indexPath];
+      cell.style.preferredSize = size;
```
appleguy: This structure mirrors the methods immediately adjacent for the headers and footers.

Source/ASCollectionView.mm
```diff
@@ -966,14 +968,8 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
   } else {
-    view = [self dequeueReusableSupplementaryViewOfKind:kind withReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
-  }
-
-  if (!node.shouldUseUIKitCell) {
     ASDisplayNodeAssert(node != nil, @"Supplementary node should exist.  Kind = %@, indexPath = %@, collectionDataSource = %@", kind, indexPath, self);
-  }
-
-  if (node) {
+    view = [self dequeueReusableSupplementaryViewOfKind:kind withReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
```
appleguy: Basically the change here is to only call the _rangeController configure in the case that we have a valid node to use, otherwise rely on UICollectionView doing its normal thing.

Source/ASCollectionView.mm
nguyenhuy: Should we do the same [here](https://github.com/facebook/AsyncDisplayKit/pull/3178/files?diff=split#diff-8b82b1556d1293f6f9789beb236c9271R936) and [here](https://github.com/facebook/AsyncDisplayKit/pull/3178/files?diff=split#diff-8b82b1556d1293f6f9789beb236c9271R951)?

maicki: Respect my feedback or not, but ASDK internally should never ever have to set a `preferredSize` to any layout element.

ay8s:
@appleguy Not dug too much into this but gave this branch a try to see if we saw any improvements with seeing Couldn't find 'visible' supplementary element of kind %@ at index path %@ in map %@ come up within Buffer. When using this branch it looks like our ASCellNode backed supplementary nodes no longer display. They display correctly when using ASDK from master.

appleguy:
@ay8s thanks for the feedback!  Do you see this behavior with any of the ASDK sample apps?

ay8s:
@appleguy I think if you run ASDKGram and go over to the IGListKit tab you'll see the header is missing. Should be showing "Latest Posts" from within FeedHeaderNode.

Source/ASCollectionView.mm
appleguy: @maicki could you share ideas on alternatives? Your feedback matters a lot!!

The idea here is that the app is not providing an ASCellNode at all, so it's not possible for the app to set the preferredSize. It never has access to the object.

The best possible implementation would be adjusting ASDataController to be able to use something like NSPointerArray, or [NSNull null], to indicate UICollectionView passthrough cells.  Since this isn't supported today, this ASCellNode is used as a very simple (low code complexity) mechanism to store & cache the synchronously calculated layout information.

This feature has ended up being very useful for apps looking for a migration path to AsyncDisplayKit, which is awesome considering how little code is needed to support the passthrough of a UICollectionViewCell. But we should think about whether it's worth the effort for a larger refactor to support this.

appleguy:
@ay8s super helpful, thanks! I hadn't realized that view was there -- oddly it does not appear in the normal ASDK mode, only IGListKit mode (checked on master as well).

The issue was caused by the fact that IGListKit uses the interop mode in a different way. It provides custom supplementaryViews, but still expects to mount the contents of an ASCellNode into them.

ay8s:
@appleguy I think @Adlai-Holler added it to debug an issue I was seeing with IGListKit and headers. 👍

appleguy:
@Adlai-Holler, @maicki I was able to simplify this patch; take a look when you have time (no rush).

The real issue is that the configureContentView: method is performing removeFromSuperview on everything in the cell's contentView, even if we are operating on a standard UIKit cell that doesn't have stuff inside its ASCellNode placeholder.

The IGListKit mode is currently the most "unique" behavior: it vends custom UICollectionViewCells, but _also_ wants to automatically remove all their subviews. That said, this patch makes sure to retain that behavior, and relies only on the "shouldUseUIKitCell" flag on ASCellNode. This flag is an accurate guarantee of true cell passthrough being used for that indexPath.

I heard that there might be a future change to IGListKit so it doesn't require custom cells; alternatively, a change to avoid creating ASCellNodes for UIKit passthrough would also help.

I can understand the desire to simplify this. However, the current state should be stable (there's no need to immediately refactor), and fixes the bug that prevented Supplementary View passthrough from working.

Source/ASCollectionView.mm
```diff
@@ -970,13 +972,11 @@ - (UICollectionReusableView *)collectionView:(UICollectionView *)collectionView
...
   BOOL shouldDequeueExternally = _asyncDataSourceFlags.interopViewForSupplementaryElement && (_asyncDataSourceFlags.interopAlwaysDequeue || node.shouldUseUIKitCell);
   if (shouldDequeueExternally) {
+    // This codepath is used for both IGListKit mode, and app-level UICollectionView interop.
     view = [(id<ASCollectionDataSourceInterop>)_asyncDataSource collectionView:collectionView viewForSupplementaryElementOfKind:kind atIndexPath:indexPath];
   } else {
-    view = [self dequeueReusableSupplementaryViewOfKind:kind withReuseIdentifier:kReuseIdentifier forIndexPath:indexPath];
-  }
-
-  if (!node.shouldUseUIKitCell) {
     ASDisplayNodeAssert(node != nil, @"Supplementary node should exist.  Kind = %@, indexPath = %@, collectionDataSource = %@", kind, indexPath, self);
```
Adlai-Holler: I believe this assertion should apply in all cases (normal ASDK, IGListKit, or interop). If you agree, move it outside the if-statement before landing.

appleguy: @Adlai-Holler  For some reason, I was hitting this assertion in the passthrough cases I was testing. I agree this probably shouldn't be the case, but I think it could be an artifact of issues with the supplementary node system (or potentially the test case).

I'll make a note for myself to investigate this assertion in more detail and will send another PR if I can get it re-added.

PR #3179, opened 2017-03-14
[ASDisplayNode] Add locking to view and layer in ASDisplayNode

maicki:
Currently creating a `view` / `layer` in `ASDisplayNode` can cause a race condition where it can happen the loading of the view / layer is in flight but e.g. `isNodeLoaded` would return NO although it would be ready. The idea from this PR is to make the creation of the layer / view as well as applying all pending states in one transaction so no race condition should happen anymore.

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -575,11 +575,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 
 - (BOOL)_locked_shouldLoadViewOrLayer
 {
-  if (_flags.isDeallocating) {
-    return NO;
-  }
-
-  return !(_hierarchyState & ASHierarchyStateRasterized);
+  return !_flags.isDeallocating && !(_hierarchyState & ASHierarchyStateRasterized);
 }
 
 - (UIView *)_locked_viewToLoad
@@ -667,6 +663,7 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 - (void)_didLoad
 {
   ASDisplayNodeAssertMainThread();
+  ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__);
   ASDisplayNodeLogEvent(self, @"didLoad");
   TIME_SCOPED(_debugTimeForDidLoad);
   
@@ -723,10 +720,10 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   }
 
   if (![self _locked_shouldLoadViewOrLayer]) {
-    return _view;
+    return nil;
   }
   
-  // Loading a view needs to happen on the maint thread
+  // Loading a view needs to happen on the main thread
   ASDisplayNodeAssertMainThread();
   [self _locked_loadViewOrLayerIsLayerBacked:isLayerBacked];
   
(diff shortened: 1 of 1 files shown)


Source/ASDisplayNode.mm
```diff
@@ -729,31 +729,38 @@ - (BOOL)_locked_isNodeLoaded
...
-    
+  {
+    ASDN::MutexLocker l(__instanceLock__);
+    if (_layer != nil) {
+      return _layer;
+    }
+      
     if (!_flags.layerBacked) {
-      return self.view.layer;
+      return self.view.layer; // Call view explicitly in case it needs to be
```
garrettmoon: Don't forget to finish your sandwich (sentence).

Source/ASDisplayNode.mm
```diff
@@ -729,31 +729,39 @@ - (BOOL)_locked_isNodeLoaded
...
+      return nil;
+    }
+    
+    if (_view != nil) {
+      return _view;
+    }
   }
 
+  ASDisplayNodeAssertMainThread();
+  [self _loadViewOrLayerIsLayerBacked:NO];
```
nguyenhuy: Don't we need the lock while loading view? Otherwise, if a background thread calls `isNodeLoaded`, it will get a no while the view is literally being loaded.

So this maybe:
```objective-c
- (UIView *)view
{
  ASDisplayNodeAssert(!_flags.layerBacked, @"Call to -view undefined on layer-backed nodes");
  if (_flags.layerBacked) {
    return nil;
  }
 
  if (_view != nil) {
    return _view;
  }
  
  [self _loadViewOrLayerIsLayerBacked:NO]; 
  return _view;
}
```

Source/ASDisplayNode.mm
```diff
@@ -729,29 +710,95 @@ - (BOOL)_locked_isNodeLoaded
...
+    return _view;
+  }
+  
+  // Loading a view needs to happen on the maint thread
+  ASDisplayNodeAssertMainThread();
+  [self _locked_loadViewOrLayerIsLayerBacked:isLayerBacked];
+  
+  // FIXME: Ideally we'd call this as soon as the node receives -setNeedsLayout
+  // but automatic subnode management would require us to modify the node tree
+  // in the background on a loaded node, which isn't currently supported.
```
nguyenhuy: I'm not sure how difficult it is to support modifying a node tree in background on a loaded node, but now that we have more FIXMEs like this, we should escalate it. Could you please create a new ticket for it?

Source/ASDisplayNode.mm
```diff
@@ -3761,44 +3810,48 @@ - (BOOL)pointInside:(CGPoint)point withEvent:(UIEvent *)event
...
     self._locked_asyncLayer.displaysAsynchronously = NO;
   }
-  
-  __instanceLock__.unlock();
 }
 
 - (void)applyPendingViewState
 {
   ASDisplayNodeAssertMainThread();
+
```
nguyenhuy: `ASDisplayNodeAssertLockUnownedByCurrentThread` please.

Source/ASDisplayNode.mm
maicki: https://github.com/facebook/AsyncDisplayKit/issues/3185

maicki:
@nguyenhuy Addressed your comments. Can you please take another look over it.

Issue #3019, opened 2017-02-12
Changing node's supernode unexpected behaviour.


Kuluum:
ASDK v2.2 (also tested on 2.1, 2.0, 2.0.1, 2.0.2)
Should be tested on iPhone.

Case: Trying to show different layout in portrait and landscape. In landscape image placed on root node, but in portrait it placed on proxy node that placed on root node.

Ascii pic for better understanding:

Portrait:

```
 ----ASStackLayoutSpec-----
|   ---ASDisplayNode---    |
|  |   ASImageNode     |   |
|  |    ASTextNode     |   |
|   --------------------   |
---------------------------
```

Landscape:

```
-------------ASStackLayoutSpec-----------
|                ---ASDisplayNode---    |
| ASImageNode   |   ASTextNode      |   |
|                -------------------    |
-----------------------------------------
```

Q: Why you use that proxy node? It has no sense there.
A: The real case was that proxy node was ASScrollNode or ASTableNode. So in portrait all content was scrollable but in landscape only text was scrollable, image had fixed position. In ascii pic and sample project I use simple ASDisplayNode for more clarity.

After that case was implemented I see such a problem - image node visible only in portrait. I suspect it is because of parent changing while layout, but I could be wrong.

Sample project: https://github.com/Kuluum/ASDKParrentChangingIssueSample

Kuluum:
Upd: Reproduced on 2.2

maicki:
@Kuluum Hey! Could you take me a favor and check if the linked PR will resolve your issue. It did on my side testing with your sample project, what was very valuable! Thanks!

Kuluum:
@maicki Ok, will check it ASAP :)

Kuluum:
@maicki Checked this fix in production code with real case that I have. It seems to be worked fine now! Great!

PR #3181, opened 2017-03-14
[Layout] Check in layout transition if the supernode is still the node that is executing the transition

maicki:
We have to check in subnode removals in a layout transition, if the supernode of the subnode is indeed still the node that executes the layout transition. It can happen that a node already did a layout transition and added this subnode, in this case the subnode would would be removed from the new node instead of the old node.

Resolves: #3019

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -3098,7 +3098,6 @@ ASDISPLAYNODE_INLINE BOOL nodeIsInRasterizedTree(ASDisplayNode *node) {
   [self _removeFromSupernode];
 }
 
-// NOTE: You must not called this method while holding the receiver's instance lock. This may cause deadlocks.
 - (void)_removeFromSupernode
 {
   ASDisplayNodeAssertThreadAffinity(self);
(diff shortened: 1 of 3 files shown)


Source/Private/ASLayoutTransition.mm
```diff
@@ -125,7 +125,12 @@ - (void)applySubnodeRemovals
 
   ASDisplayNodeLogEvent(_node, @"removeSubnodes: %@", _removedSubnodes);
   for (ASDisplayNode *subnode in _removedSubnodes) {
-    [subnode _removeFromSupernode];
+    // We have to check in this case if the supernode of the subnode is indeed the _node that executes a layout transition.
+    // It can happen that a node already did a layout transition and added this subnode, in this case the subnode would
+    // would be removed from the new node instead of _node
+    if (subnode.supernode == _node) {
```
nguyenhuy: The instance lock of subnode is acquired twice in this method: inside supernode getter and then inside `_removeFromSupernode`. Would it be possible for supernode to be changed during the short time gap before the 2 lock acquisitions? Should we add a new `_removeFromSupernodeIfEqualTo:(ASDisplayNode *)supernode` method? Inside that method we can just return if `_supernode != supernode`?

maicki:
@nguyenhuy Addressed comments, can you take another look over it. Thanks!

Source/ASDisplayNode.mm
```diff
@@ -3098,23 +3098,47 @@ - (void)removeFromSupernode
...
+  ASDisplayNodeAssertThreadAffinity(self);
+  ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__);
   
   __instanceLock__.lock();
-    __weak ASDisplayNode *supernode = _supernode;
-    __weak UIView *view = _view;
-    __weak CALayer *layer = _layer;
+
+  // Only remove if supernode is still the expected supernode
+  if (_supernode != nil && supernode != nil && _supernode != supernode) {
```
nguyenhuy: Please use `ASObjectIsEqual()` in `ASEqualityHelpers.h`

Source/ASDisplayNode.mm
```diff
@@ -3098,23 +3098,47 @@ - (void)removeFromSupernode
...
+  }
+  
+  __weak UIView *view = _view;
+  __weak CALayer *layer = _layer;
   __instanceLock__.unlock();
+  
+  [self removeFromSupernode:supernode view:view layer:layer];
+}
 
+- (void)removeFromSupernode:(ASDisplayNode *)supernode view:(UIView *)view layer:(CALayer *)layer
```
nguyenhuy: Nit: should we have `_` prefix to indicate this method is private?

Source/Private/ASLayoutTransition.mm
```diff
@@ -125,7 +125,10 @@ - (void)applySubnodeRemovals
 
   ASDisplayNodeLogEvent(_node, @"removeSubnodes: %@", _removedSubnodes);
   for (ASDisplayNode *subnode in _removedSubnodes) {
-    [subnode _removeFromSupernode];
+    // In this case we should only remove the subnode if it's still a subnode of the _node that executes a layout transition.
+    // It can happen that a node already did a layout transition and added this subnode, in this case the subnode would
+    // would be removed from the new node instead of _node
```
nguyenhuy: 1 last nit: 

> in this case the subnode **would would** be removed from the new node instead of _node

nguyenhuy:
@maicki LGTM, merge as will!

Issue #3177, opened 2017-03-14
When multiple ASNetworkImageNodes are loaded with the same gif image, a deadlock occurs


Finb:
Hello!~
This library is amazing, I really like it. Thanks very much for your work.

Here is a sample project
https://github.com/Finb/AsyncDisplayKitIssueDemo

AsyncDisplayKit: 2.2
PINRemoteImage: 3.0.0-beta.8
device: iPhone 6 10.2(14C92)  , simulator 10.2(14C89)
Xcode: 8.2.1

You may need to quickly scroll the TableView to see the problem

![](http://ww4.sinaimg.cn/large/0060lm7Tgy1fdmau1omvsj30oj0qvwk2.jpg
)

PR #3187, opened 2017-03-15
[ASNetworkImageNode] Improve locking

maicki:
`ASImageNode`, `ASNetworkImageNode` are pretty important classes where the current locking situation is kind of interesting. This diff tries to cleanup the whole situation a bit. I intentionally didn't refactor the classes too much to remove the noise that would make, especially methods like: `- (void)_lazilyLoadImageIfNecessary`.

This will happening in a follow up diff.

Resolves: #3177

### DIFF
--- a/Source/ASImageNode+AnimatedImage.mm
+++ b/Source/ASImageNode+AnimatedImage.mm
@@ -97,7 +97,7 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
 - (void)setCoverImageCompleted:(UIImage *)coverImage
 {
   ASDN::MutexLocker l(_animatedImageLock);
-  [self _locked_setCoverImage:coverImage];
+  [self _locked_setCoverImageCompleted:coverImage];
 }
 
 - (void)_locked_setCoverImageCompleted:(UIImage *)coverImage
--- a/Source/ASImageNode+AnimatedImage.mm
+++ b/Source/ASImageNode+AnimatedImage.mm
@@ -39,7 +39,6 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
 {
   ASDN::MutexLocker l(_animatedImageLock);
   [self _locked_setAnimatedImage:animatedImage];
-
 }
 
 - (void)_locked_setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
@@ -54,15 +53,17 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
     __weak ASImageNode *weakSelf = self;
     if ([animatedImage respondsToSelector:@selector(setCoverImageReadyCallback:)]) {
       animatedImage.coverImageReadyCallback = ^(UIImage *coverImage) {
-        [weakSelf _locked_setCoverImageCompleted:coverImage];
+        // In this case the lock is already gone we have to call the unlocked version therefore
+        [weakSelf setCoverImageCompleted:coverImage];
       };
     }
     
     if (animatedImage.playbackReady) {
-      [weakSelf _locked_animatedImageFileReady];
+      [self _locked_setShouldAnimate:YES];
     } else {
       animatedImage.playbackReadyCallback = ^{
-        [weakSelf _locked_animatedImageFileReady];
+        // In this case the lock is already gone we have to call the unlocked version therefore
+        [self setShouldAnimate:YES];
       };
     }
   }
@@ -79,13 +80,8 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
   ASDN::MutexLocker l(_animatedImageLock);
 
   _animatedImagePaused = animatedImagePaused;
-  ASPerformBlockOnMainThread(^{
-    if (animatedImagePaused) {
-      [self _locked_stopAnimating];
-    } else {
-      [self _locked_startAnimating];
-    }
-  });
+
+  [self _locked_setShouldAnimate:!animatedImagePaused];
 }
 
 - (BOOL)animatedImagePaused
@@ -149,12 +145,36 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
   _animatedImageRunLoopMode = runLoopMode;
 }
 
-- (void)_locked_animatedImageFileReady
+- (void)setShouldAnimate:(BOOL)shouldAnimate
 {
-  ASPerformBlockOnMainThread(^{
-    [self _locked_startAnimating];
-  });
+  ASDN::MutexLocker l(_animatedImageLock);
+  [self _locked_setShouldAnimate:shouldAnimate];
 }
+
+- (void)_locked_setShouldAnimate:(BOOL)shouldAnimate
+{
+  // This test is explicitly done and not ASPerformBlockOnMainThread as this would perform the block immediately
+  // on main if called on main thread and we have to call methods locked or unlocked based on which thread we are on
+  if (ASDisplayNodeThreadIsMain()) {
+    if (shouldAnimate) {
+      [self _locked_startAnimating];
+    } else {
+      [self _locked_stopAnimating];
+    }
+  } else {
+    // We have to dispatch to the main thread and call the regular methods as the lock is already gone if the
+    // block is called
+    dispatch_async(dispatch_get_main_queue(), ^{
+      if (shouldAnimate) {
+        [self startAnimating];
+      } else {
+        [self stopAnimating];
+      }
+    });
+  }
+}
+
+#pragma mark - Animating
 
 - (void)startAnimating
 {
@@ -218,6 +238,8 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
   [_animatedImage clearAnimatedImageCache];
 }
 
+#pragma mark - ASDisplayNode
+
 - (void)didEnterVisibleState
 {
   ASDisplayNodeAssertMainThread();
@@ -236,6 +258,8 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
   
   [self stopAnimating];
 }
+
+#pragma mark - Display Link Callbacks
 
 - (void)displayLinkFired:(CADisplayLink *)displayLink
 {
@@ -290,6 +314,8 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
 }
 
 @end
+
+#pragma mark - ASImageNode(AnimatedImageInvalidation)
 
 @implementation ASImageNode(AnimatedImageInvalidation)
 
--- a/Source/ASImageNode+AnimatedImage.mm
+++ b/Source/ASImageNode+AnimatedImage.mm
@@ -17,11 +17,17 @@
 #import <AsyncDisplayKit/ASDisplayNode+Subclasses.h>
 #import <AsyncDisplayKit/ASDisplayNodeExtras.h>
 #import <AsyncDisplayKit/ASEqualityHelpers.h>
+#import <AsyncDisplayKit/ASImageNode+Private.h>
 #import <AsyncDisplayKit/ASImageNode+AnimatedImagePrivate.h>
 #import <AsyncDisplayKit/ASImageProtocols.h>
 #import <AsyncDisplayKit/ASInternalHelpers.h>
 #import <AsyncDisplayKit/ASNetworkImageNode.h>
 #import <AsyncDisplayKit/ASWeakProxy.h>
+
+
+@interface ASNetworkImageNode (Private)
+- (void)_locked_setDefaultImage:(UIImage *)image;
+@end
 
 NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
 
@@ -32,6 +38,12 @@ NSString *const ASAnimatedImageDefaultRunLoopMode = NSRunLoopCommonModes;
 - (void)setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
 {
   ASDN::MutexLocker l(_animatedImageLock);
+  [self _locked_setAnimatedImage:animatedImage];
+
+}
+
+- (void)_locked_setAnimatedImage:(id <ASAnimatedImageProtocol>)animatedImage
+{
   if (ASObjectIsEqual(_animatedImage, animatedImage)) {
     return;
   }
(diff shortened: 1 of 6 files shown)


Source/ASNetworkImageNode.mm
```diff
@@ -451,18 +465,26 @@ - (void)_cancelDownloadAndClearImage
   [self _cancelImageDownload];
```
maicki: We should figure out if `_cancelDownloadAndClearImage` should happen in one transaction while holding the lock.

Source/ASNetworkImageNode.mm
```diff
@@ -185,14 +198,18 @@ - (void)setDefaultImage:(UIImage *)defaultImage
...
-      self.currentImageQuality = hasURL ? 0.0 : 1.0;
-    });
-    [self _setImage:defaultImage];
+    BOOL hasURL = (_URL == nil);
+    {
+      ASDN::MutexUnlocker l(__instanceLock__);
+      /* We want to maintain the order that currentImageQuality is set regardless of the calling thread,
+       so always use a dispatch_async to ensure that we queue the operations in the correct order.
+       (see comment in displayDidFinish) */
+      dispatch_async(dispatch_get_main_queue(), ^{
```
maicki: Looking into `setURL:resetToDefault:` - the order we set the `currentImageQuality` and call `_setImage` is different. Wondering if this was intentionally @garrettmoon ?

Source/ASNetworkImageNode.mm
garrettmoon: Yeah, I'm actually concerned about changing this behavior. I'm not sure we're consistent, but I think it makes since to be consistent and always `_cancelDownloadAndClearImage` as part of a locked transaction. We should probably add the `_locked_` prefix?

Source/ASNetworkImageNode.mm
```diff
@@ -112,16 +113,20 @@ - (void)dealloc
...
-    _URL = nil;
-  }
+  __instanceLock__.lock();
+    BOOL imageWasSetExternally = (image != nil);
+    BOOL shouldCancelAndClear = imageWasSetExternally && (imageWasSetExternally != _imageWasSetExternally);
+    _imageWasSetExternally = imageWasSetExternally;
+    if (shouldCancelAndClear) {
+      ASDisplayNodeAssertNil(_URL, @"Directly setting an image on an ASNetworkImageNode causes it to behave like an ASImageNode instead of an ASNetworkImageNode. If this is what you want, set the URL to nil first.");
+
+      __instanceLock__.unlock();
```
garrettmoon: I think we should hold the lock while calling `_cancelDownloadAndClearImage`

Source/ASNetworkImageNode.mm
```diff
@@ -149,20 +154,28 @@ - (void)setURL:(NSURL *)URL resetToDefault:(BOOL)reset
...
+    BOOL hasURL = (_URL == nil);
     if (reset || hasURL) {
-      [self _setImage:_defaultImage];
-      /* We want to maintain the order that currentImageQuality is set regardless of the calling thread,
-       so always use a dispatch_async to ensure that we queue the operations in the correct order.
-       (see comment in displayDidFinish) */
-      dispatch_async(dispatch_get_main_queue(), ^{
-        self.currentImageQuality = hasURL ? 0.0 : 1.0;
-      });
+      UIImage *defaultImage = _defaultImage;
```
garrettmoon: Do we have the lock held here?

Source/ASNetworkImageNode.mm
```diff
@@ -149,20 +154,28 @@ - (void)setURL:(NSURL *)URL resetToDefault:(BOOL)reset
...
+      {
+        ASDN::MutexUnlocker l(__instanceLock__);
+        [self _setImage:defaultImage];
```
garrettmoon: Should we have another version of this method? `_locked_setImage:` ? I think we may want to get the default image and set it in one locked transaction.

Source/ASNetworkImageNode.mm
garrettmoon: Talked with Wendy and she doesn't think there was a particular reason the ordering was different.

Source/ASNetworkImageNode.mm
```diff
@@ -279,7 +296,10 @@ - (void)displayWillStartAsynchronously:(BOOL)asynchronously
     if (_imageLoaded == NO && _URL && _downloadIdentifier == nil) {
       UIImage *result = [[_cache synchronouslyFetchedCachedImageWithURL:_URL] asdk_image];
       if (result) {
-        [self _setImage:result];
+        {
+          ASDN::MutexUnlocker u(__instanceLock__);
+          [self _setImage:result];
```
garrettmoon: I think we should use `_locked_setImage` here.

Source/ASNetworkImageNode.mm
```diff
@@ -352,32 +369,33 @@ - (void)didExitPreloadState
...
 #pragma mark - Private methods, safe to call without lock
 
 - (void)handleProgressImage:(UIImage *)progressImage progress:(CGFloat)progress downloadIdentifier:(nullable id)downloadIdentifier
 {
-  ASDN::MutexLocker l(__instanceLock__);
-  // Getting a result back for a different download identifier, download must not have been successfully canceled
-  if (ASObjectIsEqual(_downloadIdentifier, downloadIdentifier) == NO && downloadIdentifier != nil) {
-    return;
+  {
+    ASDN::MutexLocker l(__instanceLock__);
```
garrettmoon: This change seems like it likely won't cause a problem, but it seems theoretically possible. If the setImage for a progress image gets the after the setImage for the download completion we could theoretically end up with the progress image and never get the complete image.

Source/ASNetworkImageNode.mm
```diff
@@ -539,29 +566,40 @@ - (void)_downloadImageWithCompletion:(void (^)(id <ASImageContainerProtocol> ima
...
     
-    if (_URL.isFileURL) {
+    if (URL.isFileURL) {
       {
         ASDN::MutexLocker l(__instanceLock__);
 
         dispatch_async(dispatch_get_main_queue(), ^{
-          if (self.shouldCacheImage) {
-            [self _setImage:[UIImage imageNamed:_URL.path.lastPathComponent]];
+          if (_shouldCacheImage) {
```
garrettmoon: I don't think we have the lock here?

Source/ASNetworkImageNode.mm
```diff
@@ -539,29 +566,40 @@ - (void)_downloadImageWithCompletion:(void (^)(id <ASImageContainerProtocol> ima
...
+            ASDN::MutexUnlocker u(__instanceLock__);
+            [self _setImage:[UIImage imageNamed:URL.path.lastPathComponent]];
```
garrettmoon: should this be a locked version instead of unlocking?

Source/ASNetworkImageNode.mm
```diff
@@ -581,21 +619,25 @@ - (void)_lazilyLoadImageIfNecessary
             }
 
             if (animatedImage != nil) {
+              ASDN::MutexUnlocker u(__instanceLock__);
               self.animatedImage = animatedImage;
```
garrettmoon: I hate to say it but I think we should have a `_locked_` version of this too instead of unlocking…

Source/ASNetworkImageNode.mm
```diff
@@ -581,21 +619,25 @@ - (void)_lazilyLoadImageIfNecessary
...
             } else {
+              ASDN::MutexUnlocker u(__instanceLock__);
               [self _setImage:nonAnimatedImage];
```
garrettmoon: Same here.

Source/ASNetworkImageNode.mm
```diff
@@ -581,21 +619,25 @@ - (void)_lazilyLoadImageIfNecessary
...
           }
 
           _imageLoaded = YES;
+
           /* We want to maintain the order that currentImageQuality is set regardless of the calling thread,
            so always use a dispatch_async to ensure that we queue the operations in the correct order.
            (see comment in displayDidFinish) */
           dispatch_async(dispatch_get_main_queue(), ^{
-            self.currentImageQuality = 1.0;
+            _currentImageQuality = 1.0;
```
garrettmoon: We don't have the lock here, do we?

Source/ASNetworkImageNode.mm
```diff
@@ -617,8 +660,11 @@ - (void)_lazilyLoadImageIfNecessary
         if (imageContainer != nil) {
           strongSelf->_imageLoaded = YES;
           if ([imageContainer asdk_animatedImageData] && _downloaderFlags.downloaderImplementsAnimatedImage) {
-            strongSelf.animatedImage = [_downloader animatedImageWithData:[imageContainer asdk_animatedImageData]];
+            id animatedImage = [_downloader animatedImageWithData:[imageContainer asdk_animatedImageData]];
```
garrettmoon: Should we unlock before calling this? Not sure the answer. Previously we've attempted to release the lock when calling the downloader (but haven't been perfect about it obviously).

Source/ASNetworkImageNode.mm
```diff
@@ -617,8 +660,11 @@ - (void)_lazilyLoadImageIfNecessary
...
+            ASDN::MutexUnlocker u(strongSelf->__instanceLock__);
+            strongSelf.animatedImage = animatedImage;
           } else {
+            ASDN::MutexUnlocker u(strongSelf->__instanceLock__);
             [strongSelf _setImage:[imageContainer asdk_image]];
```
garrettmoon: Use a `_locked_` version?

Source/ASNetworkImageNode.mm
```diff
@@ -631,27 +677,35 @@ - (void)_lazilyLoadImageIfNecessary
...
-        _cacheUUID = cacheUUID;
+        __instanceLock__.lock();
+          _cacheUUID = cacheUUID;
+        __instanceLock__.unlock();
 
         [_cache cachedImageWithURL:_URL
                      callbackQueue:dispatch_get_main_queue()
                         completion:^(id <ASImageContainerProtocol> imageContainer) {
                           // If the cache UUID changed, that means this request was cancelled.
-                          if (!ASObjectIsEqual(_cacheUUID, cacheUUID)) {
```
garrettmoon: Good catch!

Source/ASNetworkImageNode.mm
garrettmoon: So my recommendation would be to use a `_locked_setImage` and not unlock.

Source/ASImageNode+AnimatedImage.mm
```diff
@@ -81,29 +94,37 @@ - (BOOL)animatedImagePaused
...
 {
-  BOOL setCoverImage = YES;
-  {
-    ASDN::MutexLocker l(_displayLinkLock);
-    if (_displayLink != nil && _displayLink.paused == NO) {
-      setCoverImage = NO;
-    }
-  }
+  ASDN::MutexLocker l(_animatedImageLock);
+  [self _locked_setCoverImage:coverImage];
```
Adlai-Holler: Should this be `_locked_setCoverImageCompleted:`?

maicki:
@Adlai-Holler @nguyenhuy Could you please take a another look at this. Thanks!

Source/ASNetworkImageNode.mm
```diff
@@ -394,16 +416,12 @@ - (void)_updateProgressImageBlockOnDownloaderIfNeeded
...
-    shouldRender = _shouldRenderProgressImages && ASInterfaceStateIncludesVisible(_interfaceState);
-    oldDownloadIDForProgressBlock = _downloadIdentifierForProgressBlock;
-    newDownloadIDForProgressBlock = shouldRender ? _downloadIdentifier : nil;
-  }
+  __instanceLock__.lock();
+    BOOL shouldRender = _shouldRenderProgressImages && ASInterfaceStateIncludesVisible(_interfaceState);
+    id oldDownloadIDForProgressBlock = _downloadIdentifierForProgressBlock;
+    id newDownloadIDForProgressBlock = shouldRender ? _downloadIdentifier : nil;
+    BOOL clearAndReattempt = NO;
+  __instanceLock__.unlock();
```
nguyenhuy: 👍

Source/ASNetworkImageNode.mm
```diff
@@ -446,17 +464,50 @@ - (void)_updateProgressImageBlockOnDownloaderIfNeeded
   }
 }
 
+// We don't have a _locked_cancelDownloadAndClearImage as this needs to happen in one transaction
```
nguyenhuy: Still need this?

nguyenhuy:
retest this please

Source/ASNetworkImageNode.mm
```diff
@@ -617,43 +655,49 @@ - (void)_lazilyLoadImageIfNecessary
...
         }
-        else if (error && strongSelf->_delegateFlags.delegateDidFailWithError) {
-          [strongSelf->_delegate imageNode:strongSelf didFailWithError:error];
+        else if (error && _delegateFlags.delegateDidFailWithError) {
+          ASDN::MutexUnlocker u(strongSelf->__instanceLock__);
+          [delegate imageNode:strongSelf didFailWithError:error];
         }
       };
 
       if (_cache != nil) {
```
nguyenhuy: Nit: Might worth a comment that explains why we don't need the lock to access `_cache`, i.e The ivar is set once in the initializer.

Source/ASNetworkImageNode.mm
```diff
@@ -684,7 +729,12 @@ - (void)displayDidFinish
      value stored in _currentImageQuality. */
 
     _renderedImageQuality = _currentImageQuality;
-    [self.delegate imageNodeDidFinishDecoding:self];
+ 
+    // Grab the delegate before unlocking the lock
+    __weak id<ASNetworkImageNodeDelegate> delegate = _delegate;
+
+    ASDN::MutexUnlocker u(__instanceLock__);
+    [delegate imageNodeDidFinishDecoding:self];
```
nguyenhuy: Please refactor this so that we don't need the unlocker, because by the end of this scope, we unnecessary lock and then unlock immediately.

nguyenhuy:
retest this please

maicki:
retest this please

Adlai-Holler:
Jenkins retest this please

Finb:
ASNetworkImageNode flickers when the image cache is available。

Here is a sample project
https://github.com/Finb/AsyncDisplayKitIssueDemo
You can repeatedly click the insert button to reproduce the problem. It is best to test on a real device

nguyenhuy:
cc @maicki ^

maicki:
@Finb Hey is this a regression due to this PR?

Finb:
@maicki I tested the version 2.2 without this problem。 

I set the following two properties in ImageCellNode
```Objc
self.neverShowPlaceholders = YES;
_imageNode.displaysAsynchronously = NO;
```
In version 2.2, the Insert operation does not produce flicker,
After this PR, it will first display the placeholder and then display the remote image to cause flickering

I recorded a GIF to show the problem
https://dl.dropboxusercontent.com/s/3tqq5labajhwp7h/DC13E62D-FBE3-4046-AF5A-8E6BDED9C5C3-10055-00004776455935D9.gif?dl=0

maicki:
@Finb I just looked into that and I also run it with 2.1. It has the exact same behavior nothing changed. This is the nature of `ASNetworkImageNode` in case the cell node needs a reload. It fetches the image either from the cache or the network and renders it. As you can see in this video I point to release/2.2 there is no change in behavior:

https://cl.ly/0Q2B0L070Q2Y

Issue #3112, opened 2017-03-01
Unavoidable crash when implementing ASImageCacheProtocol in Swift due to incorrect bridging


flovouin:
Hi guys,

I had asked questions about the cache / download mechanism of the `ASNetworkImageNode` a while back and successfully implemented the custom logic I needed.
I currently have an `ImageManager` class (in Swift) in my app that implements both `ASImageCacheProtocol` and `ASImageDownloaderProtocol`.
Everything works great except when `ASNetworkImageNode.URL` is null. In this case, a `nil` URL is forwarded to my implementation of `clearFetchedImageFromCacheWithURL`. The problem is that, in Swift, this method doesn't take an optional but only a defined `URL`. This makes bridging crash with the following stack trace:

```
* thread #1: tid = 0x29f98d, 0x0000000101ec0af4 libswiftFoundation.dylib`static Foundation.DateComponents._unconditionallyBridgeFromObjectiveC (Swift.Optional<__ObjC.NSDateComponents>) -> Foundation.DateComponents with unmangled suffix "_merged" + 96, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x101ec0af4)
  * frame #0: 0x0000000101ec0af4 libswiftFoundation.dylib`static Foundation.DateComponents._unconditionallyBridgeFromObjectiveC (Swift.Optional<__ObjC.NSDateComponents>) -> Foundation.DateComponents with unmangled suffix "_merged" + 96
    frame #1: 0x0000000100336f68 MyApp`@objc ImageManager.clearFetchedImageFromCache(with : URL) -> () + 64 at ImageManager.swift:0
    frame #2: 0x0000000101089b5c AsyncDisplayKit`::-[ASNetworkImageNode _cancelDownloadAndClearImage](self=0x0000000102977000, _cmd="_cancelDownloadAndClearImage") + 152 at ASNetworkImageNode.mm:438
    frame #3: 0x0000000101089058 AsyncDisplayKit`::-[ASNetworkImageNode didExitPreloadState](self=0x0000000102977000, _cmd="didExitPreloadState") + 160 at ASNetworkImageNode.mm:347
    frame #4: 0x0000000101015f84 AsyncDisplayKit`::-[ASDisplayNode setInterfaceState:](self=0x0000000102977000, _cmd="setInterfaceState:", newState=1) + 1080 at ASDisplayNode.mm:3361
[...]

PR #3190, opened 2017-03-16
[ASImageProtocols] Fix nullables to avoid bridging crashes

flovouin:
Hi guys,

Following #3112 I'm submitting this PR to solve the bridging crashes I was experiencing when implementing `ASImageCacheProtocol` in Swift.
A sample project showcasing the problem can be found here: https://github.com/flovouin/ASDK-ASImageCacheProtocolBridge
The problem is that `clearFetchedImageFromCacheWithURL` is sometimes called with a `nil` URL when it shouldn't.

So I double checked all the calls to `ASImageCacheProtocol` and `ASImageDownloaderProtocol` in the ASDK codebase and came up with the following PR.

- The first commit ensures that the image nodes never call the image protocols methods with a `nil` URL / identifier when they shouldn't. The only place where I am not entirely sure whether the URL can be `nil` is in the [ASNetworkImageNode](https://github.com/flovouin/AsyncDisplayKit/blob/d1e7759e7835521b2b1b9b00a9fae7bea7a70185/Source/ASNetworkImageNode.mm#L505). It would seem that `_downloadImageWithCompletion` is only called if `_URL != nil`, but as there's some multi thread funkyness happening, I'm not sure it is safe to assume. If someone who has more experience than me could review this part, that'd be nice :-)
- The second commit removes unnecessary checks for `nil` identifiers in image downloaders, as they are declared as `nonnull` in the protocol definitions.

@maicki was suggesting in #3112 to have `synchronouslyFetchedCachedImageWithURL` accept a `nullable` URL, which is actually already the case, although I'm not sure it really makes sense: why would you ask for the resource corresponding to a `nil` URL, especially if you already know the result? Also, this method is currently only called with a (explicitly checked) non-nil URL in the codebase. Finally, its documentation seems outdated as it references a `completion` block that does not exist in the method prototype. However I did not amend it for the time being, please tell me if you want me to do so :-) I personally would be in favour of making the URL argument `nonnull`.

This PR solves the crashes demonstrated by the sample project, but I'm willing to make more changes if you have any remarks!

Cheers,
Flo

### DIFF
--- a/Source/ASMultiplexImageNode.mm
+++ b/Source/ASMultiplexImageNode.mm
@@ -203,7 +203,10 @@ typedef void(^ASMultiplexImageLoadCompletionBlock)(UIImage *image, id imageIdent
   [self _setDownloadIdentifier:nil];
   
   if (_cacheSupportsClearing && self.loadedImageIdentifier != nil) {
-    [_cache clearFetchedImageFromCacheWithURL:[_dataSource multiplexImageNode:self URLForImageIdentifier:self.loadedImageIdentifier]];
+    NSURL *URL = [_dataSource multiplexImageNode:self URLForImageIdentifier:self.loadedImageIdentifier];
+    if (URL != nil) {
+      [_cache clearFetchedImageFromCacheWithURL:URL];
+    }
   }
 
   // setting this to nil makes the node fetch images the next time its display starts
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -450,7 +450,7 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 {
   [self _cancelImageDownload];
   [self _clearImage];
-  if (_cacheFlags.cacheSupportsClearing) {
+  if (_cacheFlags.cacheSupportsClearing && _URL != nil) {
     [_cache clearFetchedImageFromCacheWithURL:_URL];
   }
 }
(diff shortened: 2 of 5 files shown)


Source/ASMultiplexImageNode.mm
```diff
@@ -203,7 +203,10 @@ - (void)didExitPreloadState
   [self _setDownloadIdentifier:nil];
   
   if (_cacheSupportsClearing && self.loadedImageIdentifier != nil) {
-    [_cache clearFetchedImageFromCacheWithURL:[_dataSource multiplexImageNode:self URLForImageIdentifier:self.loadedImageIdentifier]];
+    NSURL *URL = [_dataSource multiplexImageNode:self URLForImageIdentifier:self.loadedImageIdentifier];
+    if (URL) {
```
maicki: Please use `URL != nil`

Source/ASNetworkImageNode.mm
```diff
@@ -527,7 +527,9 @@ - (void)_downloadImageWithCompletion:(void (^)(id <ASImageContainerProtocol> ima
     }
     
     if (cancelAndReattempt) {
-      [_downloader cancelImageDownloadForIdentifier:downloadIdentifier];
+      if (downloadIdentifier) {
```
maicki: Please use `downloadIdentifier != nil`

flovouin:
Thanks for the review @maicki ! There, I made the requested changes. If there's anything else I can do, just ask :-)

flovouin:
Thanks guys!

PR #3193, opened 2017-03-16
[ASDisplayNode] Added hooks for subclasses to respond to supernode changes.

ImJCabus:
Needs validation of `ASDisplayNode`s internal locking behaviour.


maicki:
@ImJCabus Hey! The core team had a discussion about this PR and we decided against bringing it in for now. We are trying to limit complexity in ASDisplayNode for now and there are some good ways around this (e.g. the visible state changes discussed with @appleguy in the Slack channel). Although we didn't merge this PR, we are really appreciated it and we hope we will see more in the future. Thanks!

PR #3198, opened 2017-03-18
[ASCollectionNode] Add .collectionViewClass to ASCollectionNode+Beta.

appleguy:
Although we may not want to support this property long-term, there are some usages of it
that are difficult to avoid. Specifically because the complexity it adds is both low,
and contained to a small area of the code that would be easy to remove it, it would be
great to support this.

The usage relates to apps that require the Interop protocol, and are architected to expect
a few methods / protocols being implemented on the UICollectionView class itself. It does
not directly override ASCollectionView behaviors. So hypothetically, it would also work
if it were possible to set ASCollectionView's superclass.

Instead, the app is making its own subclass descend from ASCollectionView and use the interop
APIs, even in environments where there are no ASCellNodes ever returned.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -100,6 +100,7 @@
 @interface ASCollectionNode ()
 {
   ASDN::RecursiveMutex _environmentStateLock;
+  Class _collectionViewClass;
 }
 @property (nonatomic) _ASCollectionPendingState *pendingState;
 @end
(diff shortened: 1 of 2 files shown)


Source/ASCollectionNode.mm
```diff
@@ -132,7 +147,7 @@ - (instancetype)initWithFrame:(CGRect)frame collectionViewLayout:(UICollectionVi
   ASDisplayNodeViewBlock collectionViewBlock = ^UIView *{
     // Variable will be unused if event logging is off.
     __unused __typeof__(self) strongSelf = weakSelf;
-    return [[ASCollectionView alloc] _initWithFrame:frame collectionViewLayout:layout layoutFacilitator:layoutFacilitator eventLog:ASDisplayNodeGetEventLog(strongSelf)];
+    return [[[strongSelf collectionViewClass] alloc] _initWithFrame:frame collectionViewLayout:layout layoutFacilitator:layoutFacilitator eventLog:ASDisplayNodeGetEventLog(strongSelf)];
```
nguyenhuy: The use case is legitimate to me. 

Questions: Does the class need to be accessed later on? If yes, would checking `isKindOfClass:` on `collectionNode.view.class` be sufficient? Instead of having it as a property, can it be passed to an initializer as a parameter? Something like this: `initWithFrame:collectionViewClass:collectionViewLayout: layoutFacilitator:`.

Adlai-Holler: Usually passing something like that to the initializer is better than adding a property, but since this is going into the `Beta` header, best not to touch the initialization process and keep it simple and dumb.

PR #3200, opened 2017-03-20
[ASCollectionElement] Fix trait collection handling

### LINKED ISSUE #2959: Cell nodes in processing may miss trait collection changes
https://pinterest.slack.com/archives/ios-help/p1485795980000343

From @rcancro:

> I have a question surrounding ASTableNode, its data controller and when _completedNodes in the data controller is populated. On iPad when we bring up the save dialog and tap on "New Board” we push a new VC onto the modal’s nav controller. The nav controller (since it is in a modal on iPad) has a width of 520. However, the VC doesn’t realize this until `viewWillAppear`. I’m seeing the following happen:
> 1. The VC is created with a width of 768
> 2. The VC’s tableNode creates its cells with a 768 width (I can put a breakpoint in the `nodeBlockForRowAtIndexPath` and see it being hit)
> 3. The VC gets resized to 520. This causes `asyncTraitCollectionDidChange` to be called on the table node
> 4. In `asyncTraitCollectionDidChange` the tableNode asks for its completed nodes via `[self.view.dataController completedNodes]` so it can propagate the new container size. However, `[self.view.dataController completedNodes]` returns an empty array.
> Shouldn't I expect the cell that i saw being created when my breakpoint hit to be in `_completeNodes`?

nguyenhuy:
- Don't store a trait environment directly and request it when needed. The environment is often the cotaining collection/table node. Asking it for a trait collection is an upward tree traversal and can cause deadlocks. This reverts #2963.
- If the environment changes during a change set update, notify all visible elements about the change. This fixes #2959.

### DIFF
--- a/Source/ASCollectionView.h
+++ b/Source/ASCollectionView.h
@@ -45,7 +45,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @return collectionNode The corresponding ASCollectionNode, if one exists.
  */
-@property (nonatomic, weak, readonly, nullable) ASCollectionNode *collectionNode;
+@property (nonatomic, weak, readonly) ASCollectionNode *collectionNode;
 
 /**
  * The number of screens left to scroll before the delegate -collectionView:beginBatchFetchingWithContext: is called.
--- a/Source/ASCollectionView.h
+++ b/Source/ASCollectionView.h
@@ -45,7 +45,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @return collectionNode The corresponding ASCollectionNode, if one exists.
  */
-@property (nonatomic, weak, readonly) ASCollectionNode *collectionNode;
+@property (nonatomic, weak, readonly, nullable) ASCollectionNode *collectionNode;
 
 /**
  * The number of screens left to scroll before the delegate -collectionView:beginBatchFetchingWithContext: is called.
--- a/Source/Details/ASCollectionElement.h
+++ b/Source/Details/ASCollectionElement.h
@@ -23,14 +23,14 @@ AS_SUBCLASSING_RESTRICTED
 //TODO change this to be a generic "kind" or "elementKind" that exposes `nil` for row kind
 @property (nonatomic, readonly, copy, nullable) NSString *supplementaryElementKind;
 @property (nonatomic, assign) ASSizeRange constrainedSize;
-@property (nonatomic, weak, nullable) ASDisplayNode *owningNode;
-@property (nonatomic, assign) ASPrimitiveTraitCollection primitiveTraitCollection;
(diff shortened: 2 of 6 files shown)


nguyenhuy:
@rcancro: Appreciate your inputs here as well!

Source/ASCollectionView.h
```diff
@@ -45,7 +45,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  * @return collectionNode The corresponding ASCollectionNode, if one exists.
  */
-@property (nonatomic, weak, readonly) ASCollectionNode *collectionNode;
+@property (nonatomic, weak, readonly, nullable) ASCollectionNode *collectionNode;
```
Adlai-Holler: `weak` implies `nullable`, please remove.

Source/Details/ASCollectionElement.h
```diff
@@ -21,12 +23,14 @@ AS_SUBCLASSING_RESTRICTED
 //TODO change this to be a generic "kind" or "elementKind" that exposes `nil` for row kind
 @property (nonatomic, readonly, copy, nullable) NSString *supplementaryElementKind;
 @property (nonatomic, assign) ASSizeRange constrainedSize;
-@property (weak, nonatomic) id<ASTraitEnvironment> traitEnvironment;
+@property (nonatomic, weak, nullable) ASDisplayNode *owningNode;
+@property (nonatomic, assign) ASPrimitiveTraitCollection primitiveTraitCollection;
```
Adlai-Holler: `weak` implies `nullable` and let's name the property `traitCollection` because it's shorter.

Source/Details/ASDataController.mm
```diff
@@ -496,6 +506,15 @@ - (void)updateWithChangeSet:(_ASHierarchyChangeSet *)changeSet
         _visibleMap = newMap;
 
         [_delegate dataController:self didUpdateWithChangeSet:changeSet];
+        
+        // Step 5: If the environment changed mid-update, notify all visible elements
+        __weak id<ASTraitEnvironment> newEnvironment = [self.environmentDelegate dataControllerEnvironment];
+        ASPrimitiveTraitCollection newPrimitiveTraitCollection = [newEnvironment primitiveTraitCollection];
+        if (! ASPrimitiveTraitCollectionIsEqualToASPrimitiveTraitCollection(primitiveTraitCollection, newPrimitiveTraitCollection)) {
```
Adlai-Holler: Let's just name these local variables `oldTraitCollection` and `newTraitCollection`.

nguyenhuy:
@Adlai-Holler Addressed your comments. Please take another look. Thanks!

nguyenhuy:
Thank you! Merging.

PR #3207, opened 2017-03-21
Fix the Initialization Hierarchy

Adlai-Holler:
- Make `init` the designated initializer for `ASDisplayNode`.
- Add `setViewBlock:` and `setLayerBlock:` methods so that subclasses don't have to call `[super initWithNodeBlock:]` since now they are not allowed to.
- In the future we could deprecate `initWithViewBlock:` and `initWithLayerBlock:` since they now just call `setViewBlock:` and since the `didLoad:` argument can be replaced by calling `onDidLoad:` immediately after.
- Breaks anyone who calls `[super initWithViewBlock:]` but they can easily just replace with `[super init]; [self setViewBlock:]`.

### DIFF
--- a/examples/Swift/Sample/TailLoadingCellNode.swift
+++ b/examples/Swift/Sample/TailLoadingCellNode.swift
@@ -55,7 +55,10 @@ final class SpinnerNode: ASDisplayNode {
   }
 
   override init() {
-    super.init(viewBlock: { UIActivityIndicatorView(activityIndicatorStyle: .Gray) }, didLoadBlock: nil)
+    super.init()
+    setViewBlock {
+        UIActivityIndicatorView(activityIndicatorStyle: .Gray)
+    }
     
     // Set spinner node to default size of the activitiy indicator view
     self.style.preferredSize = CGSizeMake(20.0, 20.0)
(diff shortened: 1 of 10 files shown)


examples/Swift/Sample/TailLoadingCellNode.swift
```diff
@@ -55,7 +55,10 @@ final class SpinnerNode: ASDisplayNode {
   }
 
   override init() {
-    super.init(viewBlock: { UIActivityIndicatorView(activityIndicatorStyle: .Gray) }, didLoadBlock: nil)
+    super.init()
+    setViewBlock {
+        UIActivityIndicatorView(activityIndicatorStyle: .Gray)
+    }
```
Adlai-Holler: This is a minor breaking API change for Swift. I will raise it on Slack for comment but we shouldn't fret too much.

Adlai-Holler:
I specified the designated initializer for ASTableNode. Unfortunately we can't cleanly designate one for ASCollectionNode yet because of the layoutFacilitator, which I don't want to move out of the beta header (in fact I want to deprecate it ASAP). For now let's leave ASCollectionNode with no designated initializer, since nobody is having problems with it.

PR #3208, opened 2017-03-21
Avoid multiple initial data loads being issued by UICollectionView/UITableView

nguyenhuy:
- Currently, there is a gap between the moment UICollectionView/UITableView triggers its first data load and when ASDataController finishes processing it. During this gap, the view keeps issuing "initial" loads by calling reloadData and causes its data controller to reload multiple times.
- Fix by immediately forward the first reloadData call to UICollectionView/UITableView before letting its data controller to handle it for real. During the gap, the view thinks that it loaded initial data but is empty, and thus stops triggering initial loads. Once the data controller finished loading, it will call another reloadData on the view which causes the view to swap to a correct state.

### DIFF
--- a/Source/ASCollectionView.mm
+++ b/Source/ASCollectionView.mm
@@ -141,6 +141,8 @@ static NSString * const kReuseIdentifier = @"_ASCollectionReuseIdentifier";
    */
   NSInteger _batchUpdateCount;
   
+  BOOL _initialReloadDataHasBeenCalled;
+  
   struct {
     unsigned int scrollViewDidScroll:1;
     unsigned int scrollViewWillBeginDragging:1;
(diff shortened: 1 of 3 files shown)


Tests/ASTableViewTests.mm
```diff
@@ -608,7 +608,8 @@ - (void)testThatInitialDataLoadHappensInOneShot
 
   // Assert that the beginning of the call pattern is correct.
   // There is currently noise that comes after that we will allow for this test.
-  NSArray *expectedSelectors = @[ NSStringFromSelector(@selector(reloadData)) ];
+  NSArray *expectedSelectors = @[ NSStringFromSelector(@selector(reloadData)),
+                                  NSStringFromSelector(@selector(reloadData)) ];
```
Adlai-Holler: Could you update the comments here to explain why there's 2 `reloadData` calls now?

Source/ASCollectionView.mm
```diff
@@ -317,6 +317,12 @@ - (void)reloadDataWithCompletion:(void (^)())completion
 {
   ASDisplayNodeAssertMainThread();
   
+  if (! _dataController.initialReloadDataHasBeenCalled) {
+    // If this is the first reload, forward to super immediately to prevent it from triggering more "initial" loads while our data controller is working.
+    _superIsPendingDataLoad = YES;
```
Adlai-Holler: Why set this to `YES` here? Presumably super is already pending data load.

nguyenhuy: No, it's not, but it will be right after we call `-[super reloadData]`. Turning this flag on to make sure any change set update finished before the next time super queries its data will be ignored, including the coming `reloadData` change set that we're about to fire. So this means that if `ASDataController` finishes its initial data load before the next layout pass, we'd only call `-[super reloadData]` once.

nguyenhuy:
retest this please

PR #3210, opened 2017-03-22
[ASTextNode] Reset _composedTruncationText when attributedText is updated

rcancro:
Current issue:
1) Set `attributedText` in `ASTextNode` with whiteColor
2) `_composedTruncationText` gets updated to take on this new color attribute
3) Set the `attributedText` with greenColor
4) The `_composedTruncationText` gets updated with the new attributes, but does NOT overwrite any existing attributes — i.e., does not overwrite the white color with green. Comment from `_locked_prepareTruncationStringForDrawing:`:
```
// Add any of the original string's attributes to the truncation string,
// but don't overwrite any of the truncation string's attributes
```

### DIFF
--- a/Source/ASTextNode.mm
+++ b/Source/ASTextNode.mm
@@ -392,7 +392,6 @@ static NSArray *DefaultLinkAttributeNames = @[ NSLinkAttributeName ];
       return;
     }
 
-    _composedTruncationText = DefaultTruncationAttributedString();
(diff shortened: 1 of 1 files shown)


rcancro:
Thanks for the review/suggestion, @maicki!

connor:
Just ran into this earlier today. Thanks @rcancro for fixing! ✋

PR #3211, opened 2017-03-22
Simplify composedTruncationText Logic

Adlai-Holler:
- `_locked_composedTruncationText` is a lazy getter for the ivar `_composedTruncationText`
- When it changes, call `_invalidateTruncationText`
- Never access `_composedTruncationText` ivar directly.


Source/ASTextNode.mm
```diff
@@ -177,7 +177,6 @@ - (instancetype)init
     self.needsDisplayOnBoundsChange = YES;
 
     _truncationMode = NSLineBreakByWordWrapping;
-    _composedTruncationText = DefaultTruncationAttributedString();
```
nguyenhuy: 👍

PR #3214, opened 2017-03-23
Add Support for the Run Loop Queue to be Weak

Adlai-Holler:
- Save resources when nodes are discarded before being displayed.
- Fixes crashes when table nodes/collection nodes are laid out after their dataSource/delegate are deallocated. We could fix this at a lower level, but this is a good place to do it.


Adlai-Holler:
The next step here is to cherry-pick this and see how it affects https://jira.pinadmin.com/browse/BUG-53029 . It should resolve it.

PR #3222, opened 2017-03-27
[ASControlNode]Add action with block

CaptainTeemo:
When handle actions with `ASControlNode`, it's pretty handy to deal with a block, so here comes `[someControlNode addAction:^(ASControlNode * _Nonnull sender) {
    [weakSelf doSomethingWithOrWithoutSender];
  } forControlEvents:ASControlNodeEventTouchUpInside];`

### DIFF
--- a/Source/ASControlNode.h
+++ b/Source/ASControlNode.h
@@ -14,6 +14,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@class ASControlNode;
+
 /**
   @abstract Kinds of events possible for control nodes.
   @discussion These events are identical to their UIControl counterparts.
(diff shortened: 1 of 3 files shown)


Adlai-Holler:
Jenkins retest this please

CaptainTeemo:
@Adlai-Holler Sorry about that, I made some changes, could you please review it again and thanks for your time.

appleguy:
This seems like a cool feature. However, there is a high risk for retain cycles -- is there any annotation that can allow the compiler to help warn developers of this? If so, this would be a pretty awesome convenience.

If not though, the downside risk of retain cycles in an app is so high, that we should consider if the benefit of the API is enough to justify it. Without a warning, there will definitely be developers who ship long-standing retain cycles :(

Adlai-Holler:
I did some testing, and here's the result:

The compiler will generate a warning if you call a method that starts with `set` or `add` on an object using a block that captures the object.

So the naming convention here will trigger a warning if the block obviously captures the node. 

So I'm willing to move forward, with a couple more changes:
- The method should be `-addActionForControlEvents:block:` so that Swift users get trailing closure syntax.
- The test needs to be fleshed out more – the current implementation won't work if I add multiple actions to the same node, for example.
- The signature of ASControlNodeBlock should probably be `(__kindof ASControlNode *)` so that it's clear to users that they can change the parameter type if they need to.

Thoughts @nguyenhuy @appleguy ?

garrettmoon:

Adlai-Holler:
Feel free to reopen on the new repo!

PR #3224, opened 2017-03-27
[Trait collection] Correctly handle changes that occur mid-updates

nguyenhuy:
- Currently, when there is a new trait collection, we correctly propagate it to all visible elements. However, since the propagating block is executed on main thread immediately without waiting for the background editing queue of `ASDataController`, not all elements are updated.
- Then to fix that, we updated ASDataController to handle these changes inside `updateWithChangeSet` (#3200). This works, but it doesn’t address the underlying issue.
- We now delegate the propagating task to `ASDataController` which schedule a block to its main serial queue after waiting for its background editing queue.
- Unit test included.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -219,7 +219,7 @@
 
 #pragma mark Setter / Getter
 
-// TODO: Implement this without the view.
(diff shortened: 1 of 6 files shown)


Adlai-Holler:
Jenkins retest this please

Issue #3235, opened 2017-03-31
Range Controller Updates During Deallocation Can Cause Assertion Failure


Adlai-Holler:
- ASCollectionNode in full range mode
- Data source & delegate cleared (UICollectionView calls reloadData -> get empty data)
- (No layout pass happens, old cells are still visible)
- Removed from window
- Interface state goes invisible
- Range controller schedules to observe rendering notifications #1163 <--- The problem
- Later…
- Rendering finishes
- Range controller updates
- Asks for visible elements. Gets the old index paths from last layout pass before controller dealloc.
- Data controller's element map is empty, due to reloadData
- Can't get the ASCollectionElement for the cell, fail assertion

### Small Fix

Have the range controller only listen to rendering notifications when it's NOT in full mode. This seems solid, and I'll do it in a diff soon.

### Medium Fix

Add a `layoutIfNeeded` call into `[ASCollectionView visibleElementsForRangeController:]` to ensure that the layout is up-to-date. I confirm that this fixes the issue in the limited case.

### Large Fix

Stay in closer sync with UICollectionView by only actually _committing_ reloadData when the layout pass happens. In some cases, we will need to manually trigger layout passes in order to get data preloading. @nguyenhuy toyed with this idea but had some issues – could you document them here?

Main Thread:

![screen shot 2017-03-31 at 3 12 53 pm](https://cloud.githubusercontent.com/assets/2466893/24571223/999ebecc-1624-11e7-8fbf-cd8d706a8ebc.png)

Dealloc Queue Thread:
![screen shot 2017-03-31 at 3 13 07 pm](https://cloud.githubusercontent.com/assets/2466893/24571225/9d622d50-1624-11e7-9573-5f2c4f7f3a50.png)

@maicki @appleguy as well

nguyenhuy:
@Adlai-Holler Great investigation. Regarding the fixes, I've just approved the "small" one (#3236). It should buy us some time to re-consider the "medium" one (#3237) and possibly skip it entirely in favor of the "large" and permanent fix. The medium fix doesn't address the underlying issue and introduces a dangerous example that might encourage us to call `layoutIfNeeded` whenever the layout object is out-of-sync.

[...]

PR #3237, opened 2017-03-31
ASCollectionView: Ensure Layout Is Updated in visibleElementsForRangeController:

Adlai-Holler:
Another part of #3235


Adlai-Holler:
Let's do what Huy suggested, and update `visibleMap` when UICollectionView does rather than doing this. The small fix I landed in #3236 should be sufficient to fix the issues we're seeing in production.

nguyenhuy:
👍

PR #3239, opened 2017-04-03
Add Support for IGListKit > 2.1, Without IGListCollectionView

Adlai-Holler:
In IGListKit <=2.1, you have to use IGListCollectionView. In later versions, you can just use plain UICollectionView 🙌 . This diff lets us support both cases pretty cleanly.


Adlai-Holler:
Jenkins failure is spurious. Retest this please

PR #3244, opened 2017-04-05
Standardize Dependency Flags for PINRemoteImage, IGListKit

Adlai-Holler:
We are seeing a build warning when using the IGListKit Subspec (which explicitly defines `IG_LIST_KIT` in the podspec) because that causes the `#ifndef` in the prefix header to skip defining the `IG_LIST_COLLECTION_VIEW` macro.

While I'm at it fixing that, I think we should go ahead and standardize that we always define these flags internally, by looking at which headers are available when building AsyncDisplayKit.

This diff:
- Removes the explicit define from the podspec and BUCK files.
- Removes the `#ifndef` from the prefix header.

At the end of the day, if those headers aren't available (`__has_include` would evaluate to false) then the build isn't going to work no matter what you manually define.

### DIFF
--- a/Source/AsyncDisplayKit-Prefix.pch
+++ b/Source/AsyncDisplayKit-Prefix.pch
@@ -8,11 +8,4 @@
   #import <Foundation/Foundation.h>
 #endif
 
-#define PIN_REMOTE_IMAGE __has_include(<PINRemoteImage/PINRemoteImage.h>)
-#define IG_LIST_KIT __has_include(<IGListKit/IGListKit.h>)
-
-/**
- * For IGListKit versions < 3.0, you have to use IGListCollectionView.
- * For 3.0 and later, that class is removed and you use UICollectionView.
- */
-#define IG_LIST_COLLECTION_VIEW __has_include(<IGListKit/IGListCollectionView.h>)
+#import <AsyncDisplayKit/ASAvailability.h>
--- a/Source/AsyncDisplayKit-Prefix.pch
+++ b/Source/AsyncDisplayKit-Prefix.pch
@@ -8,17 +8,8 @@
   #import <Foundation/Foundation.h>
 #endif
 
-// Some build systems (Buck, Bazel, etc) will define these flags manually if the functionality
-// is needed by the app. Carthage in particular, or if a user forgets to set the build flag, benefit from
-// checking if each flag is not defined and then setting it to whether or not the header is accessible.
-
-#ifndef PIN_REMOTE_IMAGE
-  #define PIN_REMOTE_IMAGE __has_include(<PINRemoteImage/PINRemoteImage.h>)
-#endif
-
-#ifndef IG_LIST_KIT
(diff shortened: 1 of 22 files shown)


Adlai-Holler:
Update: The macros actually need to be visible to _both_ the static library _and_ the application, since the header contents are compiled in both contexts. So I moved the macros to the public header `ASAvailability.h` and moved that public header into the prefix header (private) and the umbrella header (public).

Technically these macros should be namespaced, since they're public. so `AS_PIN_REMOTE_IMAGE` `AS_YOGA` `AS_IG_LIST_KIT` but that's a pre-existing issue and not worth the energy at this time. If someone has a conflict then we'll fix it at that time.

Adlai-Holler:
Jenkins test this please

Source/AsyncDisplayKit-Prefix.pch
```diff
@@ -8,23 +8,4 @@
...
-#ifndef IG_LIST_KIT
-  #define IG_LIST_KIT __has_include(<IGListKit/IGListKit.h>)
-
-  /**
-   * For IGListKit versions < 3.0, you have to use IGListCollectionView.
-   * For 3.0 and later, that class is removed and you use UICollectionView.
-   */
-  #define IG_LIST_COLLECTION_VIEW __has_include(<IGListKit/IGListCollectionView.h>)
-#endif
+#import <AsyncDisplayKit/ASAvailability.h>
```
appleguy: I thought the idea / main benefit of this was to not rely on the pch...indeed, there are a number of build systems where getting the pch to work correctly has been a challenge (even with Cocoapods and Carthage, I've seen issues with this in the last couple months like when working on Yoga)

What are your thoughts about keeping the explicit include in other files? If any code attempts to use one of these defines and it hasn't been defined, there should be a compiler error so it won't fail silently.

If you prefer this approach, I will try to make it work — but it's less clear to me what the benefit of moving it to ASAvailability is if the pch is still essential.

Adlai-Holler: @appleguy The macros need to be publicly visible to the application, at least under Cocoapods. I was very surprised to learn that, but it kind of makes sense – the ASDK header files may either be compiled with ASDK or with the application, depending on who imports them. So if they use those macros, the macros need to be visible in both contexts.

So moving them to ASAvailability makes them publicly visible, and moving ASAvailability to the pch makes them privately visible to all classes.

Can you elaborate on how difficult it is to get the pch configured in your build system? I'm happy to remove ASAvailability from the PCH and put back the explicit imports where needed but my (ignorant?) assumption is that any competent build system could either use the pch as configured in the ASDK build settings or allow you to manually specify a pch file (like Cocoapods).

Adlai-Holler:
Hey @appleguy I went ahead and landed so that `pod spec lint` will stop failing but please respond about the PCH issue – I'd like to use it if possible, but if a commonly-used build system truly doesn't support it, just say so and we should remove the PCH entirely rather than have some code in it and some code in ASAvailability.h.

PR #3246, opened 2017-04-05
Add download resume support

garrettmoon:
Adds resume support support for image downloads that have exited preload range (only on ASNetworkImageNode currently).

### DIFF
--- a/Source/ASNetworkImageNode.mm
+++ b/Source/ASNetworkImageNode.mm
@@ -109,7 +109,7 @@ static const CGSize kMinReleaseImageOnBackgroundSize = {20.0, 20.0};
 
 - (void)dealloc
 {
-  [self _cancelImageDownload:NO];
+  [self _cancelImageDownloadWithResumePossibility:NO];
 }
 
 #pragma mark - Public methods -- must lock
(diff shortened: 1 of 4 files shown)


Source/Details/ASImageProtocols.h
```diff
@@ -110,6 +110,17 @@ typedef NS_ENUM(NSUInteger, ASImageDownloaderPriority) {
 @optional
 
 /**
+ @abstract Cancels an image download, however indicating resume data should be stored in case of redownload.
+ @param downloadIdentifier The opaque download identifier object returned from
+ `downloadImageWithURL:callbackQueue:downloadProgressBlock:completion:`.
+ @discussion This method has no effect if `downloadIdentifier` is nil. If implemented, this method
+ may be called instead of `cancelImageDownloadForIdentifier:` in cases where ASDK believes there's a chance
+ the image download will be resumed (currently when an image exits display range). You can use this to store
```
nguyenhuy: nit: "currently when an image exits **preload** range"

Source/ASNetworkImageNode.mm
```diff
@@ -107,7 +109,7 @@ - (instancetype)init
 
 - (void)dealloc
 {
-  [self _cancelImageDownload];
+  [self _cancelImageDownload:NO];
```
nguyenhuy: The new boolean param passed to `_cancelImageDownload:` is not descriptive and thus a bit confusing.  How about `_cancelImageDownloadWithResumePossibility:`?  Same concern for `_cancelDownloadAndClearImage:` and their `_locked` variants.

PR #3254, opened 2017-04-10
Need to check availability before use of macros.

garrettmoon:


Adlai-Holler:
Hmm since `ASAvailability.h` is imported into the pre-compiled header, I guess the PCH issue @appleguy mentioned with his build system is affecting something you're using?

Maybe it's best to go ahead and remove the PCH and add these imports where necessary per #3244 ? @garrettmoon

PR #3257, opened 2017-04-10
Allow creating a ASPINRemoteImageDownloader that ignores caches

plarson:
This is a minimal API change to allow bypassing PINRemoteImage's disk caching.

### DIFF
--- a/Source/Details/ASPINRemoteImageDownloader.h
+++ b/Source/Details/ASPINRemoteImageDownloader.h
@@ -51,6 +51,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (PINRemoteImageManager *)sharedPINRemoteImageManager;
 
+/**
+ * When downloading images ignore all caches. Defaults to NO.
+ */
+@property (nonatomic, assign) BOOL shouldIgnoreCache;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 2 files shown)


garrettmoon:
@plarson can you explain the use case for this? This doesn't just bypass the disk cache but *all* caches including the memory cache.

plarson:
In our app, we have a dynamic image that maintains the same URL but the raw image still changes. I needed a way to bypass the disk cache for this single image since it's URL based. Changing our API was more work than ignoring the disk cache.

I'm found another way to work around it by using this code before setting the URL:

`         let remoteImageManager = ASPINRemoteImageDownloader.shared().sharedPINRemoteImageManager()
            let cacheKey = remoteImageManager.cacheKey(for: imageUrl, processorKey: nil)
            remoteImageManager.cache.removeObject(forKey: cacheKey, completion: nil)
`

plarson:
More broadly, it would be useful for us to have the capability to set HTTP headers and caching properties on a specific ASNetworkImageNode .

Currently, I have a subclass of PINURLSessionManager that I configure at startup as the sessionManager on the sharedRemoteImageManager. It inspects URLs for data tasks as they are created and potentially modifies the NSURLRequest before it's passed to -[PINURLSessionManager dataTaskWithRequest:completionHandler:]

garrettmoon:
@plarson I would be much more supportive of exposing the sessionManager instead of adding the ability to entirely skip the cache.

garrettmoon:

Adlai-Holler:
Diff moved!

Issue #3219, opened 2017-03-24
[RFC] The next milestones of ASCollectionNode


nguyenhuy:
I'd like to share my vision of ASCollectionNode, now that we have #3017 and #3130 in pretty good shape to form a foundation of the following projects:
1. Give back control to collection layout objects and eliminate layout inspector, layout facilitator, etc.
2. Have `ASDataController`, `ASRangeController` and `ASCollectionLayout`s to be decoupled from `ASCollectionView` and instead rely on `ASCollectionNode`. With this, we'll be able to do the initial `reloadData`, prepare collection layout and preload cells before the view is even allocated.
3. Introduce measure range (and even allocate range in the far future): `ASDataController` won't allocate and measure all cell nodes in one go, but instead delegate to its layout delegate, such as `ASCollection(Gallery|Pager)FlowLayout` (#2979).
4. Interface-state-aware `ASCollectionNode`:
    - When a collection node enters preload state, it triggers the initial `reloadData` and then preloads "visible" cells. Those cells should be in preload state and start fetching images, etc. A collection node that is in preload state should have its own visible range of 0, display range of 0 and preload range of 1.
    - When it enters display state, it allocates its collection view and renders. Its visible range at this stage is 0, render range is 1 and preload range is bigger than that.
   - When it finally enters visible range, its visible cells should all be ready and it uses the normal ranges.
   - With this system, nest collection nodes (e.g closeup) will be much better supported. Currently we can't perform initial `reloadData`, prepare the collection layouts and preload visible cells (e.g pin image) of sub-collection-cell-nodes (e.g close gallery cell) until they enter display range of the containing collection (e.g closeup gallery pager) and have their view allocated.

All of these projects won't be realized without the help and feedbacks of core team members, especially @Adlai-Holler, as well as other community members. So please leave comments!

garrettmoon:
This issue was moved to TextureGroup/Texture#209

PR #3262, opened 2017-04-12
[ASCollectionLayout] Manually set size to measured cells

nguyenhuy:
Currently, `ASDataController` sets the calculated size of a cell node right after measuring it ([here](https://github.com/facebook/AsyncDisplayKit/blob/master/Source/Details/ASDataController.mm#L195)). This behavior dates all the way back to [the first implementation](https://github.com/facebook/AsyncDisplayKit/commit/f7f5988fcd816f5f879be2c3cfcfcab501509461#diff-c03b807fdd44839fd88691746d9dcc27R301) of `ASDataController`. This is to ensure that by the time the cell needs to display, it has a valid size and the `needsLayout` flag of its layer is on. Failing to do so will cause it to skip its display pass entirely. 

ASCollectionLayout currently doesn't do this which causes some cells to be missing. This PR updates ASCollectionLayout to set the frame whenever a node's layout attributes is asked.

There are 2 other places we can fix this:
- When the cell node applies its layout attributes. This event is driven by UICollectionView and is often too late. We want all cell nodes that are not only in visible range but also display and preload ranges to have a correct size.
- When the cell node enters preload range: Inside `didEnterPreloadState`, the node needs to grab its calculated size and assumes that the size is valid. However, I'd prefer ASCollectionLayout to be in charge of these kind of decisions.

#3219


nguyenhuy:
PR ported!

PR #3263, opened 2017-04-12
[ASDisplayNode] Trigger a layout pass whenever a node enters preload state

nguyenhuy:
This ensures all subnodes have the correct size to preload their content. This is important for image nodes, as well as collection and table nodes.

For collection node, we can take a step further by telling it load initial data during preload state. Without doing this, the node often loads when it enters display state which is too late. And its cells can't even start prefetching before the loading finishes. [Here](https://www.dropbox.com/s/neqp2juv9l2zi7z/Pin-to-pin%20swiping.mp4?dl=0) is a video on pin-to-pin swiping before (left) and after (right) this diff.

Since ASDisplayNode currently doesn't have a `layoutIfNeeded` implementation, I'd have to allocate its view/layer (or call `__layout` directly) in order to trigger a layout pass. Instead, I went ahead and implement the method. It's more convenient for users and helps them to avoid prematurely allocate the backing store.

I also took the opportunity to revisit the locking situation of some methods inside `ASDisplayNode+UIViewBridge`.

### DIFF
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -207,13 +207,6 @@
   [super didEnterPreloadState];
 }
 
-- (void)didEnterDisplayState
-{
-  [super didEnterDisplayState];
-  // Since an initial data load was triggered when this node enter preloads state, wait for it to finish
-  [self waitUntilAllUpdatesAreCommitted];
-}
-
 #if ASRangeControllerLoggingEnabled
 - (void)didEnterVisibleState
 {
--- a/Source/ASCollectionNode.mm
+++ b/Source/ASCollectionNode.mm
@@ -193,16 +193,25 @@
   [self.rangeController clearContents];
 }
 
-- (void)didExitPreloadState
-{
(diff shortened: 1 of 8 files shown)


Source/ASCollectionNode.mm
```diff
@@ -193,18 +193,27 @@ - (void)clearContents
...
+  // We can get rid of this later when ASDataController, ASRangeController and the collection layout object can operate without the view.
+  [self view];
+  [super didEnterPreloadState];
+}
+
+- (void)didEnterDisplayState
+{
+  [super didEnterDisplayState];
+  // Since an initial data load was triggered when this node enter preloads state, wait for it to finish
+  [self waitUntilAllUpdatesAreCommitted];
```
garrettmoon: This is the scariest part of the change, @Adlai-Holler can you double check that this makes sense to you as well?

nguyenhuy: Actually we don't need to do this so I removed it!

PR #3270, opened 2017-04-13
Create a Pluggable "Tips" System to Help in Development

Adlai-Holler:
- To activate: `#define AS_ENABLE_TIPS=1`
- ASDK will create a new window on top of the application window.
- At the end of each run loop, tips controller goes through all nodes that appeared during the turn.
- Runs all `ASTipProvider` subclasses against them. Currently we have `ASLayerBackingTipProvider`.
- If the provider generates a tip for that node, a node is added to the overlay window.
- If the user taps on a tip, we call a custom `tipDisplayBlock` with a message. Default is to log a message using NSLog.
- You can disable tips on a per-class level. `MYDisplayNode.enableTips = NO`.
- Only requires one hook from ASDisplayNode right now – `nodeDidAppear:`.

Example Screenshot:
screen shot 2017-04-13 at 10 36 34 am

Tapping on one of the big images causes this message to get logged (or shown, if the app sets a custom `tipDisplayBlock`):

```
Enable layer backing to improve performance. Node ancestry: (
    "<PIRoundedImageNode: 0x7ff89a031c00>",
    "<PIFlexibleArticleNode: 0x7ff8984a4430>",
    "<PIWrapperCellNode: 0x7ff8985ace40>"
)
```

Caveats:
- As before, you can't have view-backed nodes inside layer-backed nodes, so you need to set `layerBacked=YES` on the leaf nodes first and work your way up.
- When tips are enabled, touches on green areas are not passed to the application window. This is annoying but not too annoying.

Results:
- No false positives in Pinterest yet, and all the green is gone in one diff!

### DIFF
--- a/Source/ASDisplayNode.mm
+++ b/Source/ASDisplayNode.mm
@@ -37,6 +37,7 @@
 #import <AsyncDisplayKit/ASTraitCollection.h>
 #import <AsyncDisplayKit/ASWeakProxy.h>
 #import <AsyncDisplayKit/ASResponderChainEnumerator.h>
+#import <AsyncDisplayKit/ASTipsController.h>
 
 #if ASDisplayNodeLoggingEnabled
   #define LOG(...) NSLog(__VA_ARGS__)
@@ -245,11 +246,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   ASScreenScale();
 }
 
-+ (BOOL)layerBackedNodesEnabled
-{
-  return YES;
-}
-
 + (Class)viewClass
 {
   return [_ASDisplayView class];
@@ -275,6 +271,8 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   _eventLog = [[ASEventLog alloc] initWithObject:self];
 #endif
   
+  _viewClass = [self.class viewClass];
+  _layerClass = [self.class layerClass];
   _contentsScaleForDisplay = ASScreenScale();
   
   _primitiveTraitCollection = ASPrimitiveTraitCollectionMakeDefault();
@@ -574,9 +572,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     _viewBlock = nil;
     _viewClass = [view class];
   } else {
-    if (!_viewClass) {
-      _viewClass = [self.class viewClass];
-    }
     view = [[_viewClass alloc] init];
   }
   
@@ -610,9 +605,6 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
     _layerBlock = nil;
     _layerClass = [layer class];
   } else {
-    if (!_layerClass) {
-      _layerClass = [self.class layerClass];
-    }
     layer = [[_layerClass alloc] init];
   }
 
@@ -805,32 +797,34 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
   return _flags.synchronous;
 }
 
-- (void)setSynchronous:(BOOL)flag
-{
-  ASDN::MutexLocker l(__instanceLock__);
-  _flags.synchronous = flag;
-}
-
 - (void)setLayerBacked:(BOOL)isLayerBacked
 {
-  if (![self.class layerBackedNodesEnabled]) {
+  // Only call this if assertions are enabled – it could be expensive.
+  ASDisplayNodeAssert(!isLayerBacked || self.supportsLayerBacking, @"Node %@ does not support layer backing.", self);
+
+  ASDN::MutexLocker l(__instanceLock__);
+  if (_flags.layerBacked == isLayerBacked) {
+    return;
+  }
+  
+  if ([self _locked_isNodeLoaded]) {
+    ASDisplayNodeFailAssert(@"Cannot change layerBacked after view/layer has loaded.");
     return;
   }
 
-  ASDN::MutexLocker l(__instanceLock__);
-  ASDisplayNodeAssert(!_view && !_layer, @"Cannot change isLayerBacked after layer or view has loaded");
-  ASDisplayNodeAssert(!_viewBlock && !_layerBlock, @"Cannot change isLayerBacked when a layer or view block is provided");
-  ASDisplayNodeAssert(!_viewClass && !_layerClass, @"Cannot change isLayerBacked when a layer or view class is provided");
-
-  if (isLayerBacked != _flags.layerBacked && !_view && !_layer) {
-    _flags.layerBacked = isLayerBacked;
-  }
+  _flags.layerBacked = isLayerBacked;
 }
 
 - (BOOL)isLayerBacked
 {
   ASDN::MutexLocker l(__instanceLock__);
   return _flags.layerBacked;
+}
+
+- (BOOL)supportsLayerBacking
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  return !_flags.synchronous && !_flags.viewEverHadAGestureRecognizerAttached && _viewClass == [_ASDisplayView class] && _layerClass == [_ASDisplayLayer class];
 }
 
 - (BOOL)shouldAnimateSizeChanges
@@ -855,6 +849,12 @@ static ASDisplayNodeMethodOverrides GetASDisplayNodeMethodOverrides(Class c)
 {
   ASDN::MutexLocker l(__instanceLock__);
   _threadSafeBounds = newBounds;
+}
+
+- (void)nodeViewDidAddGestureRecognizer
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  _flags.viewEverHadAGestureRecognizerAttached = YES;
 }
 
 #pragma mark - Layout
@@ -3656,6 +3656,9 @@ ASDISPLAYNODE_INLINE BOOL nodeIsInRasterizedTree(ASDisplayNode *node) {
   ASDisplayNodeAssertMainThread();
   ASDisplayNodeAssertLockUnownedByCurrentThread(__instanceLock__);
   [_interfaceStateDelegate didEnterVisibleState];
+#if AS_ENABLE_TIPS
+  [ASTipsController.shared nodeDidAppear:self];
+#endif
 }
 
 - (void)didExitVisibleState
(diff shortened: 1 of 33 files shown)


Source/ASDisplayNode.mm
```diff
@@ -245,11 +246,6 @@ + (void)load
   ASScreenScale();
 }
 
-+ (BOOL)layerBackedNodesEnabled
-{
-  return YES;
```
Adlai-Holler: This method was never made public or used at all. It's been replaced by the instance method `-supportsLayerBacking` now.

Source/ASDisplayNode.mm
```diff
@@ -574,10 +570,8 @@ - (UIView *)_locked_viewToLoad
     _viewBlock = nil;
     _viewClass = [view class];
   } else {
-    if (!_viewClass) {
-      _viewClass = [self.class viewClass];
-    }
-    view = [[_viewClass alloc] init];
+    Class c = _viewClass ?: [_ASDisplayView class];
+    view = [[c alloc] init];
```
Adlai-Holler: Previously `_viewClass` would be `nil` until the view loaded, then it would go to `_ASDisplayView`. Now `_viewClass` stays nil unless they used a custom view class.

Source/ASDisplayNode.mm
```diff
@@ -805,26 +797,20 @@ - (BOOL)isSynchronous
   return _flags.synchronous;
 }
 
-- (void)setSynchronous:(BOOL)flag
-{
-  ASDN::MutexLocker l(__instanceLock__);
-  _flags.synchronous = flag;
```
Adlai-Holler: This was unused so I tossed it in.

levi:
WOW THIS IS AMAZING

Tests/ASDisplayNodeTests.mm
```diff
@@ -225,13 +225,6 @@ - (void)testDefaultFirstResponderBehavior {
   XCTAssertFalse([node becomeFirstResponder]);
 }
 
-- (void)testLayerBackedFirstResponderBehavior {
-  ASTestDisplayNode *node = [[ASTestResponderNode alloc] init];
-  node.layerBacked = YES;
-  XCTAssertTrue([node canBecomeFirstResponder]);
-  XCTAssertFalse([node becomeFirstResponder]);
```
Adlai-Holler: This test now is invalid because it sets `layerBacked=YES` on a node that has a custom view class. Can't be done!

Source/ASDisplayNode.mm
```diff
@@ -574,9 +572,6 @@ - (UIView *)_locked_viewToLoad
     _viewBlock = nil;
     _viewClass = [view class];
   } else {
-    if (!_viewClass) {
-      _viewClass = [self.class viewClass];
-    }
```
Adlai-Holler: These are now established during `init` so that the values are invariant across the node's lifetime.

Source/ASDisplayNode.mm
```diff
@@ -805,26 +797,20 @@ - (BOOL)isSynchronous
...
   }
 
   ASDN::MutexLocker l(__instanceLock__);
-  ASDisplayNodeAssert(!_view && !_layer, @"Cannot change isLayerBacked after layer or view has loaded");
-  ASDisplayNodeAssert(!_viewBlock && !_layerBlock, @"Cannot change isLayerBacked when a layer or view block is provided");
-  ASDisplayNodeAssert(!_viewClass && !_layerClass, @"Cannot change isLayerBacked when a layer or view class is provided");
-
-  if (isLayerBacked != _flags.layerBacked && !_view && !_layer) {
-    _flags.layerBacked = isLayerBacked;
+  if ([self _locked_isNodeLoaded]) {
```
garrettmoon: Nice cleanup :)

Source/ASTextNode.mm
```diff
@@ -264,6 +264,28 @@ - (void)didLoad
...
 }
 
+- (BOOL)supportsLayerBacking
+{
+  if (!super.supportsLayerBacking) {
+    return NO;
+  }
+  
+  // If the text contains any links, return NO.
+  NSAttributedString *attributedText = self.attributedText;
```
garrettmoon: Is their any performance concerns that this is called when setting layer backed to YES? Is it called in other paths?

Source/Base/ASDisplayNode+Ancestry.m
```diff
@@ -0,0 +1,54 @@
...
+
+#import "ASDisplayNode+Ancestry.h"
+
+AS_SUBCLASSING_RESTRICTED
+@interface ASNodeAncestryEnumerator : NSEnumerator
+@end
+
+@implementation ASNodeAncestryEnumerator {
+  // Enumerators are one-shot optimized – avoid retains/releases/weak
+  __unsafe_unretained ASDisplayNode * _Nullable _node;
```
garrettmoon: This is a helluva optimization, worth it?

Source/Debug/AsyncDisplayKit+Tips.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+@property (class) BOOL enableTips;
+
+/**
+ * A block to be run on the main thread to show text when a tip is tapped.
+ *
+ * If nil, the default, the message is just logged to the console with the
+ * ancestry of the node.
+ */
+@property (class, nonatomic, copy, null_resettable) ASTipDisplayBlock tipDisplayBlock;
```
garrettmoon: You know all the fancy annotations.

Source/Debug/AsyncDisplayKit+Tips.m
```diff
@@ -0,0 +1,47 @@
...
++ (void)setEnableTips:(BOOL)enableTips
+{
+  objc_setAssociatedObject(self, &ASDisplayNodeEnableTipsKey, @(enableTips), OBJC_ASSOCIATION_COPY);
+}
+
++ (BOOL)enableTips
+{
+  NSNumber *result = objc_getAssociatedObject(self, &ASDisplayNodeEnableTipsKey);
+  if (result == nil) {
+    return YES;
```
garrettmoon: I may be reading this wrong, but shouldn't the default be NO?

Source/Private/ASDisplayNodeTipState.h
```diff
@@ -0,0 +1,31 @@
...
+
+AS_SUBCLASSING_RESTRICTED
+@interface ASDisplayNodeTipState : NSObject
+
+- (instancetype)initWithNode:(ASDisplayNode *)node NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)init NS_UNAVAILABLE;
+
+/// Unsafe because once the node is deallocated, we will not be able to access the tip state.
+@property (nonatomic, unsafe_unretained, readonly) ASDisplayNode *node;
```
garrettmoon: Not sure I understand why not to use weak here?

Source/Private/ASTipNode.m
```diff
@@ -0,0 +1,32 @@
...
+//  Copyright © 2017 Facebook. All rights reserved.
+//
+
+#import "ASTipNode.h"
+
+#if AS_ENABLE_TIPS
+
+// Prevent compiler from complaining about undeclared selector.
+@interface UIResponder (TipNodeActions)
+- (void)didTapTipNode:(id)sender;
```
garrettmoon: Do you mind updating the comment saying this is defined on the tip window?

Source/Private/ASTipNode.m
```diff
@@ -0,0 +1,32 @@
...
+@end
+
+@implementation ASTipNode
+
+- (instancetype)initWithTip:(ASTip *)tip
+{
+  if (self = [super init]) {
+    self.backgroundColor = [UIColor colorWithRed:0 green:0.7 blue:0.2 alpha:0.3];
+    _tip = tip;
+    [self addTarget:nil action:@selector(didTapTipNode:) forControlEvents:ASControlNodeEventTouchUpInside];
```
garrettmoon: Just curious, is this standard behavior to walk the chain if the target is nil? I see that's what it does in ASControlNode.

Source/Private/ASTipProvider.m
```diff
@@ -0,0 +1,42 @@
...
+@end
+
+@implementation ASTipProvider (Lookup)
+
++ (NSArray<ASTipProvider *> *)all
+{
+  static NSArray<ASTipProvider *> *providers;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    providers = @[ [ASLayerBackingTipProvider new] ];
```
garrettmoon: So the idea is that new tip providers would be added here? Any thoughts on if we wanted to pull these out of ASDK into a separate project?

Source/ASTextNode.mm
Adlai-Holler: Good call, will modify to only call it if assertions are on.

Source/Base/ASDisplayNode+Ancestry.m
Adlai-Holler: Hmmm under normal circumstances I would say yes, but since we allow you to modify node hierarchies from any thread, I guess we have to use `weak` here, since the "collection" we are enumerating can't be held still.

Source/Debug/AsyncDisplayKit+Tips.m
Adlai-Holler: No the default is `YES` (RTFM! 😛). The global enable is whether they define `AS_ENABLE_TIPS`. This property is only for _disabling_ tips on a per-class basis, say if we make a false-positive that they find annoying when tips are on.

Source/Private/ASDisplayNodeTipState.h
Adlai-Holler: So that the `node` property is truly immutable. If we keep tip states around after nodes are deallocated, then we should crash so that we stop doing that. This acts like an assertion, similar to Swift's `unowned` storage class.

Source/Private/ASTipNode.m
Adlai-Holler: I made a protocol to contain this method.

Source/Private/ASTipNode.m
Adlai-Holler: Yep! This is the best-kept secret of UIKit – you can set target nil and use the responder chain, instead of tightly binding all your controls to their controllers.

Source/Private/ASTipProvider.m
Adlai-Holler: We could do that in the future, and we have a ton of options. We could scan the runtime for subclasses of ASTipProvider, or we could require subclasses to register themselves, etc.

Adlai-Holler:
PR moved!

PR #3271, opened 2017-04-13
Automatically resume ASVideoNode after returning from background

plarson:
AVPlayerLayer pauses video when the application goes into the background. This change resumes the video playing when the app returns.

I'm not sure if there is a way to write tests for application state changes?

### DIFF
--- a/Source/ASVideoNode.mm
+++ b/Source/ASVideoNode.mm
@@ -711,10 +711,8 @@ static NSString * const kRate = @"rate";
 #pragma mark - Playback observers
 
 - (void)applicationDidBecomeActive:(NSNotification *)notification
-{
-  ASDN::MutexLocker l(__instanceLock__);
-  
-  if (_shouldBePlaying && ASInterfaceStateIncludesVisible(self.interfaceState)) {
+{  
+  if (self.shouldBePlaying && self.isVisible) {
     [self play];
   }
 }
--- a/Source/ASVideoNode.mm
+++ b/Source/ASVideoNode.mm
@@ -830,6 +830,9 @@ static NSString * const kRate = @"rate";
   _timeObserver = nil;
   [self removePlayerItemObservers:_currentPlayerItem];
   [self removePlayerObservers:_player];
+
+  NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+  [notificationCenter removeObserver:self name:UIApplicationDidBecomeActiveNotification object:nil];
 }
 
 @end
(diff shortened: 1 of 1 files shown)


Adlai-Holler:
The CI failure is some spurious Xcode issue. Jenkins retest this please.

Source/ASVideoNode.mm
```diff
@@ -101,6 +101,9 @@ - (instancetype)initWithCache:(id<ASImageCacheProtocol>)cache downloader:(id<ASI
   [self addTarget:self action:@selector(tapped) forControlEvents:ASControlNodeEventTouchUpInside];
   _lastPlaybackTime = kCMTimeZero;
   
+  NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+  [notificationCenter addObserver:self selector:@selector(applicationDidBecomeActive:) name:UIApplicationDidBecomeActiveNotification object:nil];
```
Adlai-Holler: Let's be sure to remove ourself as an observer in `dealloc`. Heads up, it's best to remove using the specific method `removeObserver:name:object:` rather than simply `removeObserver:` which is much slower 😱

Source/ASVideoNode.mm
```diff
@@ -707,6 +710,15 @@ - (void)resetToPlaceholder
 
 #pragma mark - Playback observers
 
+- (void)applicationDidBecomeActive:(NSNotification *)notification
+{
+  ASDN::MutexLocker l(__instanceLock__);
+  
+  if (_shouldBePlaying && ASInterfaceStateIncludesVisible(self.interfaceState)) {
```
Adlai-Holler: We should avoid calling `-interfaceState` while holding the lock, since that method takes the lock. Let's remove the explicit lock, and just call `self.shouldBePlaying` to keep things simple. We can also use `self.visible` as a shorthand for `ASInterfaceStateIncludesVisible(self.interfaceState)`.

Source/ASVideoNode.mm
plarson: Sorry, got into the habit since iOS9+ automatically deregisters, I'll add the remove step.

Source/ASVideoNode.mm
plarson: Other places in ASVideoNode seem to behave the same is why I did it this way, happy to remove it.

Source/ASVideoNode.mm
```diff
@@ -707,6 +710,13 @@ - (void)resetToPlaceholder
 
 #pragma mark - Playback observers
 
+- (void)applicationDidBecomeActive:(NSNotification *)notification
+{  
+  if (self.shouldBePlaying && self.isVisible) {
+    [self play];
```
rcancro: What if the video is paused when the user backgrounds the app? Should we save that state and only start playing on resume if the video wasn't paused?

plarson: My understanding is shouldBePlaying will be NO if the video is paused.

garrettmoon:

Adlai-Holler:
Diff has moved!

